PR #4, opened 2022-07-21
Update CONTRIBUTE.rst

artem-shelkovnikov:
Minor improvements to CONTRIBUTE.rst:
- add hyperlink to Ingestion Team
- add a comment about pinning dependencies
- add a comment about 10,001 documents

### DIFF
--- a/CONTRIBUTE.rst
+++ b/CONTRIBUTE.rst
@@ -71,11 +71,11 @@ If you want to add a new connector source, you need to:
 2. implement a class that implements all methods described in `connectors.source.BaseDataSource`
 3. add a unit test in `connectors/sources/tests <connectors/sources/tests>`_ with +90% coverage
 4. declare your connector in `config.yml <config.yml>`_ in the `sources` section
-5. declare your dependencies in `requirements.txt <requirements.txt>`_
+5. declare your dependencies in `requirements.txt <requirements.txt>`_. Make sure you pin these dependencies
 6. make sure you use an async lib for your source. If not possible, make sure you don't block the loop
 7. when possible, provide a docker image that runs the backend service, so we can test the connector
 8. if you can't provide a docker image, provide the credentials needed to run against a service
-9. The backend needs to return 10,001 documents.
+9. the test backend needs to return 10,001 documents due to internal implementation details of the connector - it batches the data into chunks 0f 10,000 items
 
 
 .. warning::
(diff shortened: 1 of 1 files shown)


CONTRIBUTE.rst
```diff
@@ -71,19 +71,19 @@ If you want to add a new connector source, you need to:
 2. implement a class that implements all methods described in `connectors.source.BaseDataSource`
 3. add a unit test in `connectors/sources/tests <connectors/sources/tests>`_ with +90% coverage
 4. declare your connector in `config.yml <config.yml>`_ in the `sources` section
-5. declare your dependencies in `requirements.txt <requirements.txt>`_
+5. declare your dependencies in `requirements.txt <requirements.txt>`_. Make sure you pin these dependencies
 6. make sure you use an async lib for your source. If not possible, make sure you don't block the loop
 7. when possible, provide a docker image that runs the backend service, so we can test the connector
 8. if you can't provide a docker image, provide the credentials needed to run against a service
-9. The backend needs to return 10,001 documents.
+9. the test backend needs to return 10,001 documents due to internal implementation details of the connector - it batches the data into chunks 0f 10,000 items
```
tarekziade, line 78: this is not the reason actually. I added 10,001 because 10k is the default size limit for pagination, so having the whole flow work for +1 seems like a decent default for folks building and trying out their connectors https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-max-result-window

CONTRIBUTE.rst
```diff
@@ -71,19 +71,19 @@ If you want to add a new connector source, you need to:
...
 
 
 .. warning::
 
    Any patch with changes outside `connectors/sources <connectors/sources>`_ or `config.yml <config.yml>`_
    and `requirements.txt <requirements.txt>`_ will be rejected.
 
-   If you need changes in the framework, reach out the Ingestion team.
+   If you need changes in the framework, reach out to the `Ingestion team <https://github.com/orgs/elastic/teams/ingestion-team/members>`_.
```
tarekziade, line 86: maybe they can add an issue here? and we can provide a template?

CONTRIBUTE.rst
artem-shelkovnikov: Oh right! I've changed the comment to:
> the test backend needs to return 10,001 documents due to 10,000 being a default size limit for Elasticsearch pagination. Having 10,001 documents returned from the test backend will help testing connector more deeply

Does it look better?

CONTRIBUTE.rst
artem-shelkovnikov: Hmmmm I think it makes sense. We have an issue template in `connectors-ruby`: https://github.com/elastic/connectors-ruby/issues/new?assignees=&labels=enhancement&template=enhancement.md&title=

Will this one work in this repo as well?

CONTRIBUTE.rst
tarekziade: thx!

CONTRIBUTE.rst
tarekziade: I think it's an excellent start

tarekziade (approved):

PR #94, opened 2022-09-30
Nightly functional tests

tarekziade:
First version of a nightly test in buildkite.

- Runs the functional test for mysql
- make sure environment variables take precedence over the config file


artem-shelkovnikov:
@tarekziade do we want to continue on this work by chance?

tarekziade:
@artem-shelkovnikov yes, we want this. It's on hold, but we do need it. This was supposed to be resumed by @seanstory afaik, but I don't know if it's still going to happen soon with all the reorgs of teams 

We can also start to push small changes until https://buildkite.com/elastic/connectors-python-nightly gets green

tarekziade:
replaced by https://github.com/elastic/connectors-python/pull/258

PR #131, opened 2022-10-17
[DO NOT MERGE] draft of rss limits

tarekziade:
## Part of https://github.com/elastic/enterprise-search-team/issues/3092

Adds:
- `max_rss`: maximum RSS size in KiB. If the service reaches that limit, gracefully exits. checked after each sync.
- `max_peak_rss`: maximum RSS size in KiB during a sync. If the service reaches that limit, stops the sync and errors out.

The first value can be used to keep the service under a RSS at all time, the second one to set an upper limit during syncs
to control memory bloats.

#### Pre-Review Checklist

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -18,6 +18,10 @@ service:
   max_errors: 20
   max_errors_span: 600
   trace_mem: false
+  # 300MB
+  max_rss: 307200
+  # 400MB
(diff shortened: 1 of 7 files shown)


tarekziade:
won't use this for now (memmon)

PR #147, opened 2022-10-27
limit the bulk queue and request in memory

tarekziade:
These improvements were done for the Network drive connector and will make sure we cap memory usage when we ingest *files*

two mechanisms:

- the async queue that piles up documents to send out will block in the next insertion until its size in MB gets down
- the bulk request that is being prepared will be sent out to Elasticsearch when the request reaches a certain size, no matter how many documents we configured per bulk. 

The latter gives us a great way to scale up and down: we can put a huge value for the number of ops for bulk requests
and adapt the request size depending on the network we have. 

#### Pre-Review Checklist

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -4,8 +4,10 @@ elasticsearch:
   password: changeme
   ssl: true
   bulk_queue_max_size: 1024
+  bulk_queue_max_mem_size: 25
   bulk_display_every: 100
-  bulk_chunk_size: 150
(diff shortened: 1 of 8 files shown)


tarekziade:
> So it will block the connector from fetching new data as well?

Correct, the coroutine that gets a document and call `queue.put` will be blocked there until the queue gets smaller 
In practice that means it will throttle calls to the connectors iterator

PR #148, opened 2022-10-27
Remove the two-phase for files

tarekziade:
Remove the `upsert` ops that are done with the result of the file download.
This two-phase is overkill and adds conflict issues. Let's just sent out the document once with everything.

#### Pre-Review Checklist


tarekziade:
the other branch was merged so the previous commits are gone now

PR #150, opened 2022-10-28
do not add a new ts if the backend provides it

tarekziade:
Not sure why that was lost at some point. We need to keep the timestamp of the document when it's provided by the connector. 

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -177,7 +177,8 @@ class Fetcher:
                 else:
                     operation = OP_INDEX
                     self.total_docs_created += 1
-                    doc["timestamp"] = iso_utc()
+                    if "timestamp" not in doc:
+                        doc["timestamp"] = iso_utc()
 
                 if lazy_download is not None:
                     data = await lazy_download(doit=True, timestamp=doc["timestamp"])
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -30,6 +30,7 @@ from connectors.utils import (
 OP_INDEX = "index"
 OP_UPSERT = "update"
 OP_DELETE = "delete"
+TIMESTAMP_FIELD = "_timestamp"
 
 
 class Bulker:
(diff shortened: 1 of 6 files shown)


connectors/byoei.py
```diff
@@ -177,7 +177,8 @@ async def get_docs(self, generator):
                 else:
                     operation = OP_INDEX
                     self.total_docs_created += 1
-                    doc["timestamp"] = iso_utc()
+                    if "timestamp" not in doc:
+                        doc["timestamp"] = iso_utc()
```
seanstory, line 181: we shouldn't set `timestamp` at all if it's not part of the doc, right? Connectors shouldn't typically be adding extra fields or metadata that aren't explicitly part of the source document. And when we do, we try to make it clear it's a system field, like with a `_` prefix.

artem-shelkovnikov: Hmmm indeed, we don't do it in Ruby service at all now and it's not described in the protocol at all. Before we've also had `last_updated` and `last_created` timestamps, but we haven't discussed these default fields at all.

I'd probably make it `_timestamp` and add to Ruby, but I think it's good to discuss it in the wider group and also determine what's the domain meaning of the `timestamp` field - what does it mean?

tarekziade: > How does the Ruby framework accomplish this? 
It does not. This feature in Python was inherited from the first mongo spacetime prototype.

It reduces by 90% the workload and time to sync when you don't have a lot of changes (a typical "warm" sync)

> Shouldn't they use the same strategy?

yes, we should eventually.

artem-shelkovnikov: > what do you mean by domain ?
I mostly use the word "domain" in DDD-ish-style connotation - logic that has a user-facing feature attached to it, as opposed to the logic that has purely engineering purpose, sorry for the confusion :) 

> How does the Ruby framework accomplish this? Shouldn't they use the same strategy?
Indeed it does not and it would be nice to implement this as well in a consistent way between Ruby and Python

artem-shelkovnikov (approved):

PR #151, opened 2022-10-31
[new connector] AWS S3

akanshi-crest:
This PR includes:

Indexing functionality for multiple buckets.
E2E test cases with docker image.

Connector Checklist:
 
- [x]  Implement a class that implements all methods described in connectors.source.BaseDataSource
 - [x] Declare your dependencies in [requirements.txt](https://github.com/elastic/connectors-python/blob/main/requirements.txt). Make sure you pin these dependencies
- [x]  When possible, provide a docker image that runs the backend service, so we can test the connector. If you can't provide a docker image, provide the credentials needed to run against an online service.
- [x]  The test backend needs to return more than 10k documents due to 10k being a default size limit for Elasticsearch pagination. Having more than 10k documents returned from the test backend will help testing connector more deeply


config.yml
```diff
@@ -23,12 +23,13 @@ service:
...
   - mysql
+  - aws
 
 # some id
 #connector_id: '1'
 
 sources:
   mongodb: connectors.sources.mongo:MongoDataSource
-  s3: connectors.sources.aws:S3DataSource
+  aws: connectors.sources.aws:S3DataSource
```
tarekziade, line 33: since we might have other connectors for AWS services, we should keep `s3` in the name

connectors/sources/aws.py
```diff
@@ -3,116 +3,223 @@
...
     def __init__(self, connector):
+        """Setup connection to the Amazon S3.
+
+        Args:
+            connector (BYOConnector): Object of the BYOConnector class.
+        """
         super().__init__(connector)
         self.session = aioboto3.Session()
         self.loop = asyncio.get_event_loop()
         set_extra_logger(aws_logger, log_level=logging.INFO, prefix="S3")
```
tarekziade, line 43: we might want to set the log level to DEBUG here

connectors/sources/aws.py
```diff
@@ -3,116 +3,223 @@
...
+                    exception.response.get("Error", {}).get("Code")
+                    == "InvalidObjectState"
+                ):
+                    logger.error(
+                        f"{filename} of {bucket} is archived and inaccessible until restored. Error: {exception}"
+                    )
+                else:
+                    logger.error(
+                        f"Something went wrong while extracting data from {filename} of {bucket}. Error: {exception}"
+                    )
```
tarekziade, line 114: don't we want to reraise the exception here?

connectors/sources/aws.py
```diff
@@ -3,116 +3,223 @@
...
     async def ping(self):
-        return True
-
-    async def _get_content(self, key, timestamp):
-        # reuse the same for all files
-
-        logger.debug(f"Downloading {key}")
+        """Verify the connection with AWS"""
+        try:
+            async with self.session.client(
```
tarekziade, line 55: you have a recurrent pattern in your code:

```
async with self.session.client(service_name="s3", config=self.config, endpoint_url=ENDPOINT_URL) as s3:
    # do something
```

let's create a `client` method in the class that is decorated with `contextlib.contextmanager` so your code can do:

```
async with self.client(**extra_options_if_needed) as s3:
    # do something
```

connectors/sources/aws.py
```diff
@@ -3,116 +3,223 @@
...
+
+        return region
+
+    def get_bucket_list(self):
+        """Returns bucket list from list_buckets response
+
+        Returns:
+            list: List of buckets
+        """
+        bucket_list = []
```
tarekziade, line 142: can be replaced with a one-liner:

```
return [bucket["Name"] for bucket in self.bucket_list["Buckets"]]
```

connectors/sources/aws.py
```diff
@@ -3,116 +3,223 @@
...
+                service_name="s3",
+                config=self.config,
+                endpoint_url=ENDPOINT_URL,
+                region_name=region_name,
+            ) as s3:
+                try:
+                    bucket_obj = await s3.Bucket(bucket)
+                    await asyncio.sleep(0)
+
+                    async for obj_summary in bucket_obj.objects.all():
```
tarekziade, line 169: let's check if `all` loads in memory all docs on big buckets (that would do a memory bloat)

connectors/sources/aws.py
```diff
@@ -3,116 +3,223 @@
...
+                "value": 5,
+                "label": "Maximum retry attempts",
+                "type": "int",
             },
-            "region": {
-                "value": "us-west-2",
-                "label": "AWS Region",
+            "connector_name": {
+                "value": "AWS Connector",
+                "label": "Friendly name for the connector",
```
tarekziade, line 222: what is the purpose of `connector_name` ? I don't see it used

connectors/sources/tests/fixtures/aws/load_storage.py
```diff
@@ -0,0 +1,31 @@
...
+        except IndexError:
+            PORT = 5000
+        s3_client = boto3.client("s3", endpoint_url=f"{ENDPOINT_URL}:{PORT}")
+        s3_client.create_bucket(Bucket=BUCKET_NAME)
+        print("Creating objects on the aws-moto server")
+        for object_id in range(0, OBJECT_COUNT):
+            s3_client.put_object(
+                Key=f"{BUCKET_NAME}/{object_id}",
+                Bucket=BUCKET_NAME,
+                Body=f"Testing object{object_id} document for bucket: {BUCKET_NAME}",
```
tarekziade, line 23: let's add bigger bodies too, like what was done with NetworkDrive so we can make sure chunk downloads etc. are exercised and we can run the ftest realistically to very the memory does not blow

connectors/sources/tests/fixtures/aws/remove.py
```diff
@@ -0,0 +1,27 @@
...
+        except IndexError:
+            PORT = 5000
+        s3_client = boto3.client("s3", endpoint_url=f"{ENDPOINT_URL}:{PORT}")
+        print("Removing data from aws-moto server.")
+        for object_id in range(0, OBJECT_COUNT):
+            s3_client.delete_object(
+                Bucket=BUCKET_NAME, Key=f"{BUCKET_NAME}/{object_id}"
+            )
+    except Exception as exception:
+        print(f"Exception: {exception}")
```
tarekziade, line 23: Why catching it? I think we want to know and stop if this fails

tarekziade (commented):
Great work! I added some comments that need to be addressed but I think it looks good in general

connectors/sources/aws.py
akanshi-crest: @tarekziade we added this friendly name of connector here  because @danajuratoni said to add this in all the connectors.

config.yml
akanshi-crest: I understand your point here but if we keep s3 as source name here then our ftest will fail becuase in rest of the code we are using key name as "aws", so either we have to change the name from aws to s3 in allover the code structure or we can directly do this a single source name change in config.yml to make it work. Let me know which one we have to opt.

connectors/sources/aws.py
akanshi-crest: @tarekziade we checked [aioboto3](https://github.com/terrycain/aioboto3/blob/main/aioboto3/resources/collection.py#:~:text=count%20%3D%200-,async,-for%20page%20in%20cast) libarary and it indirectly calls [pages](https://github.com/boto/boto3/blob/develop/boto3/resources/collection.py) method which returns a chunks of 1000 documents.
so `all` will not loads the complete data at a time in  memory.

connectors/sources/aws.py
akanshi-crest: If we raise the exception then in case of glacier bucket type we will not be able to fetch the content and we will have to stop the execution and the further data will not get indexed. Is it ok?

akanshi-crest: Yes, we can raise the exception. However there is one edge case where if the user has glacier bucket type then the lib is not able to fetch the content and raises the exception and that will stop the execution if we reraise the exception. or we conditionally continue to iterate over next bucket in case the unsupported bucket storage type.

akanshi-crest:
> Great work! I added some comments that need to be addressed but I think it looks good in general

@tarekziade I have updated the PR with the fix.

config.yml
tarekziade:
```
if we keep s3 as source name here then our ftest will fail becuase we are using folder name "aws" everywhere in the connector and in that case file/folder name and source name are not matching and throws keyerror for "s3" when running ftest,
```

I don't understand this. we have a `service_type` which is `s3` and then a FQN (Fully Qualified Name) that is another separate thing. when you use `connectors.sources.aws:S3DataSource` it will look for the `S3DataSource` class in the `connectors.sources.aws` namespace.

Currently `aws` is a file. If you make it a directory you can use this structure for example:

```
connectors/
   sources/
      aws/
          __init__.py
         s3.py 
```

and have the `S3DataSource` class in `s3.py`, and then in `__init__.py` you can import `S3DataSource`

connectors/sources/aws.py
tarekziade: in case of Glacier you are right, we can warn the user and ignore the file

connectors/sources/aws.py
tarekziade: great! let's surface the page size as an option if we want to tweak this

tarekziade:
> > Great work! I added some comments that need to be addressed but I think it looks good in general
> 
> @tarekziade I have updated the PR with the fix.

Thanks can you resolve the ones you did so we know what's left to do

akanshi-crest:
@tarekziade PR is updated and marked as resolve as well.

connectors/sources/aws.py
parth-crest: @tarekziade We checked setting the log_level as DEBUG here, but in that case suppose we have a bucket in region "us-east-1" and aws configuration default region is "ap-south-1", so we are getting a debug log for all the object exist in that bucket which says:
S3 request was previously redirected, not redirecting.

> [S3][10:26:51][DEBUG] S3 client configured for region ap-south-1 but the bucket bucket-50k is in region us-east-1; Please configure the proper region to avoid multiple unnecessary redirects and signing attempts.

Do you think it's okay to print so many logs for DEGUB?

akanshi-crest: @tarekziade I set the log_level as DEBUG here, but in that case suppose we have a bucket in region "us-east-1" and aws configuration default region is "ap-south-1", so we are getting a debug log for all the object exist in that bucket which says:
S3 request was previously redirected, not redirecting.

> [S3][10:26:51][DEBUG] S3 client configured for region ap-south-1 but the bucket bucket-50k is in region us-east-1; Please configure the proper region to avoid multiple unnecessary redirects and signing attempts.

Do you think it's okay to print so many logs for DEGUB?

tarekziade (approved):

PR #152, opened 2022-10-31
Add retry mechanism

parth-crest:
- Added custom retry mechanism to handle `ReadTimeout` issue that was resulting into Lost Connection to MySQL server.

#### Pre-Review Checklist

- [x] Added retry to handle Lost Connection to MySQL

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -26,7 +26,6 @@ DEFAULT_FETCH_SIZE = 50
 DEFAULT_RETRY_COUNT = 3
 DEFAULT_WAIT_MULTIPLIER = 2
 
-
 class MySqlDataSource(BaseDataSource):
     """Class to fetch and modify documents from MySQL server"""
 
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -24,6 +24,7 @@ QUERIES = {
 }
 DEFAULT_FETCH_SIZE = 50
 DEFAULT_RETRY_COUNT = 3
+DEFAULT_WAIT_MULTIPLIER = 2
 
 
 class MySqlDataSource(BaseDataSource):
@@ -44,6 +45,9 @@ class MySqlDataSource(BaseDataSource):
             "db": None,
             "maxsize": MAX_POOL_SIZE,
         }
+        self.retry_count = int(
+            self.configuration.get("retry_count", DEFAULT_RETRY_COUNT)
+        )
         self.connection_pool = None
 
     @classmethod
@@ -138,7 +142,7 @@ class MySqlDataSource(BaseDataSource):
         # rows_fetched: Rows fetched counter
         # cursor_position: Mocked cursor position
         rows_fetched = cursor_position = 0
-        while retry <= int(self.configuration.get("retry_count", DEFAULT_RETRY_COUNT)):
+        while retry <= self.retry_count:
             try:
                 async with self.connection_pool.acquire() as connection:
                     async with connection.cursor(aiomysql.cursors.SSCursor) as cursor:
@@ -179,11 +183,13 @@ class MySqlDataSource(BaseDataSource):
                 )
                 break
             except Exception as exception:
-                logger.exception(
-                    f"Retry count: {retry} out of {DEFAULT_RETRY_COUNT} for rows in {database}.{table}. Exception: {exception}"
+                logger.warn(
+                    f"Retry count: {retry} out of {self.retry_count} for rows in {database}.{table}. Exception: {exception}"
                 )
+                if retry == self.retry_count:
+                    raise exception
                 cursor_position = rows_fetched
-                await asyncio.sleep(2**retry)
+                await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry)
                 retry += 1
 
     async def _execute_query(self, query):
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -59,27 +59,27 @@ class MySqlDataSource(BaseDataSource):
         return {
             "host": {
                 "value": "127.0.0.1",
-                "label": "MySQL Host",
+                "label": "Host",
                 "type": "str",
             },
             "port": {
                 "value": 3306,
-                "label": "MySQL Port",
+                "label": "Port",
                 "type": "int",
             },
             "user": {
                 "value": "root",
-                "label": "MySQL Username",
+                "label": "Username",
                 "type": "str",
             },
             "password": {
                 "value": "changeme",
-                "label": "MySQL Password",
+                "label": "Password",
                 "type": "str",
             },
             "database": {
                 "value": ["customerinfo"],
-                "label": "List of MySQL Databases",
+                "label": "Databases",
                 "type": "list",
             },
             "connector_name": {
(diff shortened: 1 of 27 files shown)


connectors/sources/mysql.py
```diff
@@ -111,31 +117,74 @@ async def ping(self):
...
 
-                    if len(rows) == 0:
+        # retry: Current retry counter
+        # yield_once: Yield(fields) once flag
+        retry = yield_once = 1
+
+        # rows_fetched: Rows fetched counter
+        # cursor_position: Mocked cursor position
+        rows_fetched = cursor_position = 0
+        while retry <= int(self.configuration.get("retry_count", DEFAULT_RETRY_COUNT)):
```
tarekziade, line 141: you can store `retry_count` on the class to avoid doing this `int(self....)` on every loop

connectors/sources/mysql.py
```diff
@@ -111,31 +117,74 @@ async def ping(self):
...
-                        count = 0
-                    await asyncio.sleep(0)
+            except IndexError as exception:
+                logger.exception(
+                    f"None of rows fetched from {rows_fetched} rows in {database}.{table}. Exception: {exception}"
+                )
+                break
+            except Exception as exception:
+                logger.exception(
+                    f"Retry count: {retry} out of {DEFAULT_RETRY_COUNT} for rows in {database}.{table}. Exception: {exception}"
```
tarekziade, line 183: if `retry_count` is differerent from `DEFAULT_RETRY_COUNT` this log info won't show the current countdown

connectors/sources/mysql.py
```diff
@@ -111,31 +117,74 @@ async def ping(self):
...
-                        logger.info(f"Collected {count} rows in {database}.{table}")
-                        count = 0
-                    await asyncio.sleep(0)
+            except IndexError as exception:
+                logger.exception(
+                    f"None of rows fetched from {rows_fetched} rows in {database}.{table}. Exception: {exception}"
+                )
+                break
+            except Exception as exception:
+                logger.exception(
```
tarekziade, line 182: we don't a `warn` level here

connectors/sources/mysql.py
```diff
@@ -111,31 +117,74 @@ async def ping(self):
...
+                    f"Retry count: {retry} out of {DEFAULT_RETRY_COUNT} for rows in {database}.{table}. Exception: {exception}"
+                )
+                cursor_position = rows_fetched
+                await asyncio.sleep(2**retry)
```
tarekziade, line 186: let's make `2` a constant, or a config option

connectors/sources/mysql.py
tarekziade, line 141: if all retries have failed, you silently catch the last one and return the function. You need to keep track of the last exception and re-raise it if all retries have failed

tarekziade (commented):
looks good, one thing to fix (re-raise) and we need a test case to demo that the retries are done, and one that makes sure we still raise when all retries are exhausted

parth-crest:
Sure, we're working on the suggested changes.

parth-crest:
Hey @tarekziade We've updated PR with suggested changes.

connectors/sources/mysql.py
```diff
@@ -111,31 +120,76 @@ async def ping(self):
...
+            except IndexError as exception:
+                logger.exception(
+                    f"None of rows fetched from {rows_fetched} rows in {database}.{table}. Exception: {exception}"
+                )
+                break
+            except Exception as exception:
+                logger.warn(
+                    f"Retry count: {retry} out of {self.retry_count} for rows in {database}.{table}. Exception: {exception}"
+                )
+                if retry == self.retry_count:
```
tarekziade, line 188: adding the raise here and doing breaks make the code very hard to read. I also think that you might get connection issues when you call MySQL outside of that specific loop in other parts of the class.

I would like to suggest refactoring this a bit by creating a dedicated decorator function with the retry mechanism

something like

```
def _with_retries(self, callable, *args, **kw):
    retries = 0
    while True:
        try:
            return callable(*args, **kw)
        except CatchSomeError:
            retries += 1
            if retries >= self.retry_count:
                raise
            sleep...
```

and then in your code:

```
await self._with_retries(self._stream_rows, database, table, query)

```

parth-crest: Hey @tarekziade , We tried this approach, there are few things that we want to discuss.

- We won't be able to return the callable directly since, the callable will return the following:

1. **Generator**: the function stream_rows returns a generator. Also, we cannot return the generator directly as well since the value of the generator can be:
    - Results: In case of the successful response, the callable will return a generator with the results of the query
    - Exception: In case of a server exception, the callable will return a generator with the exception.
Hence, we'll have to iterate over the generator returned by the callable to verify if the generator contains an Exception, then retry the connection again.

2. **Direct Response(List/Tuple)**: The function execute_query directly returns the response in the form of List or Tuple. 

   
Hence we'll to tweak this function as below:

```
def _with_retries(self, callable, *args, **kw):
    retries = 0
    while True:
        try:
            if returntype(callable) == <generator>:
                for item in generator:
                    if exception:
                        retry
                    else return
            else:
                return callable(*args, **kw)
        except CatchSomeError:
            retries += 1
            if retries >= self.retry_count:
                raise
            sleep...
```

Please let us know your thoughts for the same

tarekziade: I think the easiest way to do this would be to retry only at the query level, not the generator level. 

So every method from the `cursor` object -- could be transparent to the connector's code.

One thing you could do is create a new class from `aiomysql.cursors.SSCursor` and add the retry mechanism for `fetchmany`, `scroll` and `execute`

and then use it in `async with connection.cursor(MyCustomSSCursor) as cursor:`

tarekziade: What I am not sure about is if you can actually retry a `cursor` call

tarekziade:
buildkite test this

tarekziade:
buildkite test this

connectors/sources/mysql.py
parth-crest: @tarekziade We tried several ways to manage the connection and cursor to handle retry. However, one issue with retrying inside the Cursor class is that on an exception, the current connection will be lost and if we try to create a new connection to retry, the reference b/w the cursor and connection will be lost. Hence, it doesn't seem reasonable to retry(re-create connection) inside a cursor class. We even tried assigning a new connection to the existing cursor inside the SSCursor class but the _reader fails initialization.

Since the connector is public and people might be using it, do you think it makes sense to get the basic retry functionality merged and then work on the future enhancement to make it generic?

tarekziade: sure ok, that sounds fair. The current retry is only on document retrievals, though. we would still need to have a retry for the other calls made around that loop.  I propose that you create a new bug explaining the follow up work, with a list of places that need a retry

tarekziade (approved):

praveen-elastic:
buildkite test this

praveen-elastic:
buildkite test this

parth-crest:
@tarekziade Sure. We've raised a GitHub ticket https://github.com/elastic/connectors-python/issues/187. We can have follow up on the same ticket.

parth-crest:
buildkite test this

PR #157, opened 2022-11-01
[8.5] Fix memory bloat issue on large text and dbs (#100)

tarekziade:
Backports the following commits to 8.5:
 - Fix memory bloat issue on large text and dbs (#100)


tarekziade:
waiting for https://github.com/elastic/connectors-python/pull/156 to land then I can rebase against 8.5

PR #177, opened 2022-11-09
Surface `analysis_icu` and `language_code` in config

moxarth-elastic:
## Closes https://github.com/elastic/connectors-py/issues/43

Populate `analysis_icu` and `language_code` in config.yml to further use them in framework.

#### Pre-Review Checklist


tarekziade (commented):
Looks good! thanks.
Let's add those options with non default values in https://github.com/elastic/connectors-python/blob/main/connectors/tests/config.yml 
and a small test to verify that they get picked

PR #178, opened 2022-11-10
Replace `timestamp` field with `_timestamp` in Network Drives

praveen-elastic:
Due to recent changes in byoei file we will need to use '_timestamp' in place of 'timestamp'. 

https://github.com/elastic/connectors-python/blob/4abe3f2e677d7454706cc83206f79d3fc52841d8/connectors/byoei.py#L34

#### Pre-Review Checklist


tarekziade:
Thanks

tarekziade (approved):

PR #174, opened 2022-11-08
Added `max_concurrency` for bulk requests

tarekziade:
### Closes https://github.com/elastic/enterprise-search-team/issues/3285

A few changes in the `bulk` calls:

- Added `max_concurrency` to allow concurrent bulks
- Use a pass-through for `bulk` options to reduce friction in pass-through classes
- `get_size` stays in bytes and the logger can do its thing
- document classes

#### Pre-Review Checklist
- [X] Covered the changes with automated tests
- [X] Tested the changes locally

#### Changes Requiring Extra Attention

- [ ] Security-related changes (encryption, TLS, SSRF, etc)
- [ ] New external service dependencies added.

PR #180, opened 2022-11-10
added a perf script to standardize the output of metrics

tarekziade:
- [x] Adds a script to feed a CSV with cpu/memory info during a sync
- [x] Generate a plot for RSS
- [x] Hooked the script in `make ftest`


tarekziade:
This is the first iteration for a script that can be used to collect metrics in perf tests

tarekziade:
With a small plot generated for RSS at the end

report

tarekziade:
example report on a MySQL 300k sync with high mem allowed

report

tarekziade:
example report on a MySQL 300k sync with high mem allowed (5 x 2MiB bulk and 5 MiB queue)

report

tarekziade:
example report on a MySQL 300k sync with super big mem allowed and Elasticsearch heap = 2GiB (20 x 50MiB bulk and 500 MiB queue) 

report

tarekziade:
example report on a MySQL 300k sync with high concurrency, small bulks (50 x 3MiB bulk and 500 MiB queue)

report

tarekziade:
example report on a MySQL 300k sync with low concurrency, small bulks (2 x 2MiB bulk and 3 MiB queue)
report

tarekziade:
Conclusion with sizing one node in isolation, with that series (running on a single elastic search node), we're maxing out one Elasticsearch node's abilities when it has default values. There's no specific benefit to adding more power in the service; it'll take around 1000 seconds, so 17min, to ingest all those docs.

So (5 x 2MiB bulk and 5 MiB queue) for that one connector running in isolation is a good compromise. 

To go faster, it will happen on Elasticsearch side (cluster/nodes size, maybe `indices.memory.index_buffer_size`)

tarekziade:
more tweaks here https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-indexing-speed.html

tarekziade:
and https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-disk-usage.html

tarekziade:
buildkite test this

tarekziade:
I had a bug in my script, all the times can be divided by two, it's between 400 and 600s so between 6 to 10 minutes

tarekziade:
(includes https://github.com/elastic/connectors-python/pull/174)

tarekziade:
decent defaults : 
```
bulk:
   queue_max_size: 1024
   queue_max_mem_size: 5
   display_every: 100
   chunk_size: 1000
   max_concurrency: 5
   chunk_max_mem_size: 2
```

total time 650 seconds, ~11 minutes -- peak RSS 207.8 MiB

report

tarekziade:
isolate the reporting in a dedicated project  https://github.com/tarekziade/perf8

tarekziade:
to use it 

```
pip install https://github.com/tarekziade/perf8/tarball/main
perf8 --psutil -c elastic-ingest ...
```

seanstory:
Can you move that to an Elastic repo? If it's something that our team needs to be able to use/access and is not just a personal project, it should be somewhere we can easily own/maintain in the event that you win the lottery and leave Elastic.

tarekziade:
> Can you move that to an Elastic repo? 

yeah, that is planned (like how we did for this project here) 
the repo permissions have bitten me before, so I usually start it in my space and then ask IT to move it

> If it's something that our team needs to be able to use/access and is not just a personal project, it should be somewhere we can easily own/maintain in the event that you win the lottery and leave Elastic.

I don't play the lottery, but sometimes I dream of what I'd do with a lot of money. 
I won't buy a social network, that's for sure. Not planning to leave Elastic; I would be so bored in my mansion.
I think I would try to buy a small island

PR #189, opened 2022-11-15
Remove fetch all databases

parth-crest:
## Closes https://github.com/elastic/connectors-py/issues/171

Updated code to index nothing if user has passed empty value in "List of MySQL databases" from the UI.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -26,7 +26,6 @@ DEFAULT_FETCH_SIZE = 50
 DEFAULT_RETRY_COUNT = 3
 DEFAULT_WAIT_MULTIPLIER = 2
 
-
 class MySqlDataSource(BaseDataSource):
     """Class to fetch and modify documents from MySQL server"""
 
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -24,6 +24,7 @@ QUERIES = {
 }
 DEFAULT_FETCH_SIZE = 50
 DEFAULT_RETRY_COUNT = 3
+DEFAULT_WAIT_MULTIPLIER = 2
 
 
 class MySqlDataSource(BaseDataSource):
@@ -44,6 +45,9 @@ class MySqlDataSource(BaseDataSource):
             "db": None,
             "maxsize": MAX_POOL_SIZE,
         }
+        self.retry_count = int(
+            self.configuration.get("retry_count", DEFAULT_RETRY_COUNT)
+        )
         self.connection_pool = None
 
     @classmethod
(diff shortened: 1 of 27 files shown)


parth-crest:
buildkite test this

moxarth-elastic:
buildkite test this

connectors/sources/mysql.py
```diff
@@ -331,37 +331,24 @@ async def fetch_documents(self, database, table, query):
...
-
     async def get_docs(self):
         """Executes the logic to fetch databases, tables and rows in async manner.
 
         Yields:
             dictionary: Row dictionary containing meta-data of the row.
         """
         database_config = self.configuration["database"]
-        if isinstance(database_config, str):
+        if not database_config:
```
tarekziade, line 341: let's be explicit.

`if  database_config == ''`

connectors/sources/mysql.py
```diff
@@ -331,37 +331,24 @@ async def fetch_documents(self, database, table, query):
...
+            logger.info(
+                "No database passed in the configuration. Stopping the iteration."
+            )
+            return
+        elif isinstance(database_config, str):
```
tarekziade, line 346: can it be anything else than `str` ? if it's not `str` what do we do?

tarekziade (changes requested):
I know there's another patch for field validation, but here the logic is not great because the `if` test is not explicit. Maybe treat the empty string, the string and then raise of you get something else? (until we have the other patch) (or maybe this patch can be combined with the other one)

parth-crest:
Hey @tarekziade , previously we were explicitly calling the if condition as we were getting empty list ([]) because we were using get_default_configuration() method to get config. So we were checking on `len(database)`, However our framework is evolving and with the newer update we are getting empty string('') from the UI, that change broke the logic. So we updated if condition to be implicit to cover more cases.
But from now onwards we will take care to handle if condition explicitly.

Also, as you suggested we can close this PR and accommodate the changes in the Validation Config PR.

tarekziade:
> Hey @tarekziade , previously we were explicitly calling the if condition as we were getting empty list ([]) because we were using get_default_configuration() method to get config. So we were checking on `len(database)`, However our framework is evolving and with the newer update we are getting empty string('') from the UI, that change broke the logic.

yes, and the changed logic is fine

> So we updated if condition to be implicit to cover more cases. But from now onwards we will take care to handle if condition explicitly.

yes thanks, the problem with implicit if is that your value is converted to a boolean. and that leads to unexpected results.
like, do you know what `bool(-1)` or `bool(0)` returns?  ;)

for this reason, since Python is not statically typed, it's best to be explicit and not leave room for uncertainty.


> 
> Also, as you suggested we can close this PR and accommodate the changes in the Validation Config PR.

sounds good

connectors/sources/mysql.py
parth-crest: Sure we will update that in the other Validate PR.

connectors/sources/mysql.py
parth-crest: We can have database_config from either UI or from the get_default_configuration() method.

- From the UI we will be getting string everytime.
- From the get default configuration method we will have list and we have handled that case in the else condition.

PR #197, opened 2022-11-18
arch specific requs

tarekziade:
Cleanup and tweak the requirements since we have different libs on aarch64


tarekziade:
> While change looks good, it sounds like a potentially serious problem that we use different versions of libraries per platform, should we pick libraries that are equaly well supported on both architectures?

I think it's best to support each architecture's latest version and document the gap. In an ideal world, we close the gap in the aarch64 ones. It's probable that we can rebuild them at install time from the source

Makefile
```diff
@@ -1,29 +1,30 @@
...
 	$(PYTHON) -m venv .
 	bin/pip install --upgrade pip
 
 install: bin/python
-	bin/pip install -r requirements.txt
+	bin/pip install -r requirements/$(ARCH).txt
 
 dev: install
-	bin/pip install -r test-requirements.txt
+	bin/pip install -r requirements/tests.txt
```
wangch079, line 14: I wonder what is this used for?

Makefile
```diff
@@ -1,29 +1,30 @@
...
 bin/elastic-ingest: bin/python
-	bin/pip install -r requirements.txt
+	bin/pip install -r requirements/$(ARCH).txt
 	bin/python setup.py develop
 
 bin/black: bin/python
-	bin/pip install -r requirements.txt
-	bin/pip install -r test-requirements.txt
+	bin/pip install -r requirements/$(ARCH).txt
+	bin/pip install -r test-requirements/$(ARCH).txt
```
wangch079, line 22:
```suggestion
	bin/pip install -r requirements/tests.txt
```

Makefile
```diff
@@ -1,29 +1,30 @@
 .PHONY: test lint run ftest install dev
 
 PYTHON=python3.10
+ARCH=$(shell uname -m)
```
wangch079, line 4: Should we have a fallback to set `ARCH` to `framework` if it's not one of the supported values?

Makefile
tarekziade: you mean the tests requirement? it's all the packages use to run tests, that are not installed in production to *run* the service

Makefile
tarekziade: good idea we can fallback to `x86_64` (also in setup.py)

tarekziade: actually just do `make install ARCH=x86_64`

tarekziade:
buildkite test this

j-bennet:
I wonder if you could keep the same `requirements.txt`, but use the environment markers:

https://peps.python.org/pep-0496/

tarekziade:
> I wonder if you could keep the same `requirements.txt`, but use the environment markers:
> 
> https://peps.python.org/pep-0496/

great idea, let's do it as a follow up

j-bennet (approved):
👍

PR #199, opened 2022-11-21
[MySQL] Fields and Database validation

parth-crest:
## Closes https://github.com/elastic/connectors-py/issues/185, https://github.com/elastic/connectors-py/issues/171

- Added _validate_connection_fields() methods to validate fields passed from UI are empty or not.
- Added _validate_databases() to validate passed databases are accessible by user or not.

#### Pre-Review Checklist


connectors/sources/mysql.py
```diff
@@ -106,11 +99,44 @@ async def close(self):
         await self.connection_pool.wait_closed()
         self.connection_pool = None
 
+    def _validate_connection_fields(self):
```
tarekziade, line 102: we can use it to validate any config, we could call it `_validate_configuration`

connectors/sources/mysql.py
```diff
@@ -106,11 +99,44 @@ async def close(self):
...
+            isinstance(self.configuration["port"], str)
+            and not self.configuration["port"].isnumeric()
+        ):
+            raise Exception("Configured port can't include string literal.")
+
     async def ping(self):
         """Verify the connection with MySQL server"""
+        logger.info("Validating MySQL Configuration...")
+        self._validate_connection_fields()
+        connection_string = {
```
tarekziade, line 129: since you are doing this validation just here, why not:

```
connection_string = self._build_connection_string()
```

and do the validation there?

connectors/sources/mysql.py
```diff
@@ -106,11 +99,44 @@ async def close(self):
...
+    def _validate_connection_fields(self):
+        """Validates whether user input is empty or not for connection string
+
+        Raises:
+            Exception: Configured keys can't be empty
+        """
+        connection_fields = ["host", "port", "user", "password", "database"]
+        empty_connection_fields = []
+        for field in connection_fields:
+            if not self.configuration[field]:
```
tarekziade, line 111: what is this test exactly? are you testing for `None`, for ``,  for `false` ?
always be explicit in this kind of test. Do not assume the boolean transformation will get what you want.

connectors/sources/mysql.py
```diff
@@ -106,11 +99,44 @@ async def close(self):
...
+        if empty_connection_fields:
+            raise Exception(
+                f"Configured keys: {empty_connection_fields} can't be empty."
+            )
+
+        if (
+            isinstance(self.configuration["port"], str)
+            and not self.configuration["port"].isnumeric()
+        ):
+            raise Exception("Configured port can't include string literal.")
```
tarekziade, line 123: this message is cryptic. what about `Configured port has to be an integer`

connectors/sources/tests/test_mysql.py
```diff
@@ -428,26 +420,33 @@ async def test_get_docs_with_none():
...
+    with mock.patch.object(MySqlDataSource, "_validate_databases", return_value=([])):
+        # Execute
+        async for doc, _ in source.get_docs():
+            assert doc == {"a": 1, "b": 2}
 
 
 @pytest.mark.asyncio
-async def test_get_docs():
-    """Test get docs method of Mysql"""
+async def test__validate_databases():
```
tarekziade, line 430: two `__`

tarekziade (changes requested):
Nice, there are a few things to fix, and also, this misses a test for field validation. 
We want to see a series of badly configured configurations and verify they are rejected.

connectors/sources/mysql.py
parth-crest: If we name method as _build_connection_string and validate config inside the method, it will conflict the function definition. So, it will be better if we can keep validation logic separate from building the connection string.
What do you think ?

connectors/sources/mysql.py
parth-crest: Done

connectors/sources/mysql.py
parth-crest: Done

connectors/sources/tests/test_mysql.py
parth-crest: Done

connectors/sources/mysql.py
parth-crest: Done

tarekziade:
buildkite test this

tarekziade (approved):

PR #205, opened 2022-11-24
added all licenses

tarekziade:
## Closes https://github.com/elastic/enterprise-search-team/issues/3403

Just ran `pip-licenses` against `requirements.txt`

#### Pre-Review Checklist


wangch079 (dismissed):

wangch079 (commented):
While I reviewed https://github.com/elastic/connectors-python/pull/197, I came back to this PR and realized it doesn't run against `requirements.txt`. 

From the [doc](https://pypi.org/project/pip-licenses/):
>`pip-licenses` is a CLI tool for checking the software license of installed Python packages with pip.

My understanding is that the result is not necessarily the same as `requirements.txt`, and it depends on what is installed in your local env.

Not a blocker for this PR, but I think it will be easier if we could have a make command to run against `requirements.txt`. I believe this markdown has to be updated for every release.

seanstory:
The PR description notes that it "closes" those tickets, but the tickets state a need for a `make` target. This doesn't add one. Do we need a make target? Does this need to be checked in our CI to ensure that it does not become out of date?

tarekziade:
Yes, we need to add a make target, and we also need to re-generate that file every time the requirements are changed when we introduce a new dep/change one.

The expectation is that this update has to be done manually because we need a human to review the new licenses against our open-source policy. 

I guess we can add a check in 

https://github.com/elastic/connectors-python/blob/main/docs/CONTRIBUTING.md#initial-contribution

and a note at the top of the requirements files to ensure it's done.

artem-shelkovnikov:
@tarekziade do we still want these changes merged?

tarekziade:
@artem-shelkovnikov yes we should finish and land this

artem-shelkovnikov:
So I've added some things here:

1. `make check-licenses` that checks if license list has changed - useful for CI script IMO
2. `make update-licenses` that updates current license list - can be used manually cause we still need to check licenses manually
3. They use `FAIL_ON_LICENSES` variable that is empty now, but we can add licenses that are not acceptable for us there.

Does it bring this PR closer to merging?

I can see that I can also add a CI step for checking if license list changed in this PR, but not sure if there's more to do.

wangch079:
I added a new CI step `Check licenses` to make sure `dep-licenses.md` is always up-to-date. 

https://github.com/elastic/connectors-python/blob/fe9e4997b7a1e4fb74c27730c88fda624b26ae78/.buildkite/pipeline.yml#L14-L15

But this is very system-dependent. Our CI is run in Ubuntu, and I work on an intel-based mac, so it passes. I expect I will generate a different list of dependencies once I switch to M2 Mac, and the CI will fail at that time.

We have a few options. 

1. I think we currently only release the software in docker, which, CMIIW, is running in ubuntu, so we only need to make sure the dependencies in [requirements/x86_64.txt](https://github.com/elastic/connectors-python/blob/main/requirements/x86_64.txt) comply with our policy. But it will be troublesome for someone who works on a M2 Mac if he/she wants to add a new dependency.
2. We generate `dep-licenses.md` for all dependencies in [requirements](https://github.com/elastic/connectors-python/tree/main/requirements). But I don't find an easy way to get the license of an uninstalled package.

deps-licenses.md
seanstory, line 1: I know it was Tarek's original approach, but this file should be named `NOTICE.txt`, and should be formatted like other Elastic `NOTICE.txt` files. See: https://github.com/elastic/open-source/blob/main/elastic-product-policy.md#third-party-code-in-an-elastic-repository-eg-copy-paste-vendoring-etc

wangch079: I'm following this: https://github.com/elastic/open-source/blob/main/elastic-product-policy.md#for-repositories-holding-only-code-subject-to-the-elastic-license-egml-cpp

I think the purpose of this file is to just list down all the licenses of dependencies to make sure we comply with Elastic Open Source Policy (by eyeball now, and can be automated later)

>Files containing third-party code should include: (i) the Apache 2.0 license header (when in an OSS code folder) or the Elastic license header (when in an x-pack folder) and (ii) for each third-party package from which more than ten (10) lines of code are taken, a comment containing the contents of third-party's NOTICE.txt file, license header and copyright notice (if available).

Does this only apply to copy-pasting? For all the 3rd party lib, we only install them as dependencies, and we never try to modify their code.

seanstory: My understanding from previous tickets with the github.com/elastic/open-source group is that we need our own NOTICE.txt file included in our distributables (docker images, tarballs, etc).

PR #210, opened 2022-11-25
fix MANIFEST

tarekziade:
make sure requirements files are part of the release


j-bennet (approved):
Better late than never. :)

tarekziade:
> Better late than never. :)

hehe it's only impacting tarball releases

PR #217, opened 2022-11-29
dont run pyspy on darwin

tarekziade:
tweak to not use `--pyspy` on darwin

### DIFF
--- a/connectors/tests/ftest.sh
+++ b/connectors/tests/ftest.sh
@@ -24,14 +24,15 @@ sleep 30
 
 $ROOT_DIR/bin/fake-kibana --index-name search-$NAME --service-type $NAME --debug
 
-#make load-data
+make load-data
+
 if [[ $PERF8 == "yes" ]]
 then
     if [[ $PLATFORM == "darwin" ]]
     then
-      $ROOT_DIR/bin/perf8 --memray --pyspy --psutil -c $ROOT_DIR/bin/elastic-ingest --one-sync --sync-now --debug
-    else
       $ROOT_DIR/bin/perf8 --memray --psutil -c $ROOT_DIR/bin/elastic-ingest --one-sync --sync-now --debug
+    else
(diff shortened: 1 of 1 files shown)


connectors/tests/ftest.sh
```diff
@@ -7,26 +7,39 @@ PERF8=$2
...
 cd $ROOT_DIR/connectors/sources/tests/fixtures/$NAME
 make run-stack
+
 # XXX make run-stack should be blocking until everythign is up and running by checking hbs
 sleep 30
 
 $ROOT_DIR/bin/fake-kibana --index-name search-$NAME --service-type $NAME --debug
 
-make load-data
+#make load-data
```
seanstory, line 27: Should this still be commented out?

connectors/tests/ftest.sh
```diff
@@ -7,26 +7,39 @@ PERF8=$2
...
 if [[ $PERF8 == "yes" ]]
 then
-    $ROOT_DIR/bin/perf8 --memray --pyspy --psutil -c $ROOT_DIR/bin/elastic-ingest --one-sync --sync-now --debug
-
+    if [[ $PLATFORM == "darwin" ]]
```
seanstory, line 30: The description says
> don't run pyspy on darwin

but this _only_ runs pyspy on darwin. Should this be a `-ne`?

tarekziade: ahah good catch thanks :)

connectors/tests/ftest.sh
tarekziade: debug left over, will fix

artem-shelkovnikov:
What's up with Darwin? It's not available on it?

tarekziade:
> What's up with Darwin? It's not available on it?

it requires root privileges, and even when running as a root, I had errors and problems.

artem-shelkovnikov (approved):

PR #218, opened 2022-11-30
Add pagination and only return supported connectors

wangch079:
This PR introduces 2 changes:

1. Add pagination to `get_connectors` function
2. Make `get_connectors` only return supported connectors.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -117,30 +117,40 @@ class BYOIndex(ESClient):
         )
 
     async def get_connectors(self, native_service_types=[], connectors_ids=[]):
+        if len(native_service_types) == 0 and len(connectors_ids) == 0:
+            return
+
         await self.client.indices.refresh(index=CONNECTORS_INDEX)
 
-        query = {
+        native_connectors_query = {
             "bool": {
-                "should": [
-                    {
-                        "bool": {
-                            "filter": [
-                                {"term": {"is_native": True}},
-                                {"terms": {"service_type": native_service_types}},
-                            ]
-                        }
-                    },
-                    {
-                        "bool": {
-                            "filter": [
-                                {"term": {"is_native": False}},
-                                {"terms": {"_id": connectors_ids}},
-                            ]
-                        }
-                    },
+                "filter": [
+                    {"term": {"is_native": True}},
+                    {"terms": {"service_type": native_service_types}},
                 ]
             }
         }
+        custom_connectors_query = {
+            "bool": {
+                "filter": [
+                    {"term": {"is_native": False}},
+                    {"terms": {"_id": connectors_ids}},
+                ]
+            }
+        }
+        if len(native_service_types) > 0 and len(connectors_ids) > 0:
+            query = {
+                "bool": {
+                    "should": [
+                        native_connectors_query,
+                        custom_connectors_query
+                    ]
+                }
+            }
+        elif len(native_service_types) > 0:
+            query = native_connectors_query
+        else:
+            query = custom_connectors_query
 
         async for connector in self._query_with_pagination(CONNECTORS_INDEX, query):
             yield BYOConnector(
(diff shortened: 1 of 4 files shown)


connectors/byoc.py
```diff
@@ -116,32 +116,69 @@ async def preflight(self):
             indices=[CONNECTORS_INDEX, JOBS_INDEX], pipelines=[PIPELINE]
         )
 
-    async def get_list(self):
+    async def get_connectors(self, native_service_types=[], connectors_ids=[]):
         await self.client.indices.refresh(index=CONNECTORS_INDEX)
```
artem-shelkovnikov, line 120: Will not passing `native_service_types` or `connectors_ids` result in 0 results returned in this case?

wangch079: They can be empty because they are configured in yaml. 

I added default values to allow function calls like
```python
self.connectors.get_connectors(native_service_types=native_service_types)
self.connectors.get_connectors(connectors_ids=connectors_ids)
```

But I will add this at the beginning of the function:
```python
if len(native_service_types) == 0 and len(connectors_ids) == 0:
    return
```

I should also change the query to make it work for scenarios when `native_service_types` is empty or `connectors_ids` is empty

artem-shelkovnikov (approved):

connectors/byoc.py
```diff
@@ -116,32 +116,79 @@ async def preflight(self):
...
+                    index=index,
+                    query=query,
+                    from_=offset,
+                    size=page_size,
+                    expand_wildcards="hidden",
+                )
+            except ApiError as e:
+                logger.critical(f"The server returned {e.status_code}")
+                logger.critical(e.body, exc_info=True)
+                return
```
tarekziade, line 181: `return` here will silently swallow the error. Is this intended?

connectors/runner.py
```diff
@@ -180,29 +180,20 @@ async def poll(self, one_sync=False, sync_now=True):
...
                 except Exception as e:
                     logger.critical(e, exc_info=True)
                     self.raise_if_spurious(e)
 
                 if not one_sync:
                     await self._sleeps.sleep(self.idling)
-                else:
-                    self.stop()
         finally:
+            self.stop()
```
tarekziade, line 196: this breaks `--one-sync`

tarekziade (changes requested):

tarekziade:
added @vidok as a reviewer since he's working on a branch that impacts this

connectors/byoc.py
wangch079: This is how the original code handles the `ApiError`. 

`logger.critical` will log the errors.

I expect it to return and yield nothing, identical to returning an empty list.

connectors/runner.py
wangch079: In line 188, it breaks out of the loop if it's `one_sync`, I just moved `self.stop()` to `finally`, as I think this should be run as a cleanup for all the scenarios.

wangch079 (commented):
>added @vidok as a reviewer since he's working on a branch that impacts this

Among all the tasks in 8.7, this is the most isolated work and I expect less conflicts with Dmitrii's work.

connectors/runner.py
tarekziade: Thanks, So, the `break` here does not change the `running` flag. We are going to stay in `while self.running` -- I assume this will be fixed in the other patch ?

wangch079: Updated in both https://github.com/elastic/connectors-python/pull/219/commits/15e8bd33c6f4223a5a90f81ed9d8c141db0d0d6a and https://github.com/elastic/connectors-python/pull/218/commits/9733f1754fbae723b9ce95833254379ce9f3ac0f

tarekziade (approved):
LGTM

tarekziade:
That broke `--one-sync`

tarekziade:
fixed in https://github.com/elastic/connectors-python/pull/224

PR #219, opened 2022-11-30
Job Cleanup task

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3200

This PR adds a Job cleanup task, which runs by default every 5 mins, to 
1. delete all the orphaned jobs and their content indices. An orphaned job is defined as a job whose associated connector is deleted.
2. mark stuck jobs as error. A stuck job is defined as a job which has seen no update for 1 minute. The task will only process stuck jobs of supported connectors.

#### Pre-Review Checklist


connectors/runner.py
```diff
@@ -192,17 +199,15 @@ async def poll(self, one_sync=False, sync_now=True):
...
                 except Exception as e:
                     logger.critical(e, exc_info=True)
                     self.raise_if_spurious(e)
 
                 if not one_sync:
                     await self._sleeps.sleep(self.idling)
-                else:
-                    self.stop()
         finally:
+            self.stop()
```
tarekziade, line 210: why did you move this line? this will break `--one-sync` because it will not stop anymore in the loop

tarekziade (commented):
this looks great! just one issue with `stop`

connectors/runner.py
tarekziade: oh because you added `self.job_cleanup_task.shutdown()`
that seems ok, as long as you `break` at the end of the loop if `one_sync` is True

wangch079: Even without `self.job_cleanup_task.shutdown()`,  `self.stop()` is needed when exiting the service, that's why I moved it to `finally`

tarekziade: The problem is that `stop` will toggle the flag `running` so with `--one-sync` with your change, when we `break` we will still be in the `while` loop.  I think we need to separate the stopping logic from the cleanup logic

wangch079: Oh, I see, there are 2 loops, the outer `while` loop and an inner `for` loop. I missed the 2nd one. 

IMO, we should move below check from the inner loop to the outer loop:
https://github.com/elastic/connectors-python/blob/c0c368a5906145e61786fb50bc5e5e30adc1d9bd/connectors/runner.py#L201-L202

From my understanding, `--one-sync` is to run one sync for each connector and exit, not the first connector, which is non-deterministic and not intuitive.

connectors/runner.py
```diff
@@ -94,6 +96,9 @@ def stop(self):
         self._sleeps.cancel()
         if self.connectors is not None:
             self.connectors.stop_waiting()
+        if self.job_cleanup_task is not None:
```
tarekziade, line 99: this is a problematic pattern because you delete the object without waiting for the task it holds to finish.

you probably want to make shutdown() a coroutine, and wait on it, and inside wait for the task to finish once you stop it

connectors/job_cleanup.py
```diff
@@ -0,0 +1,58 @@
...
+    ):
+        self.connectors = connectors
+        self.native_service_types = native_service_types.copy()
+        self.connector_ids = connector_ids
+        self.interval = interval
+        self.stuck_threshold = stuck_threshold
+        self.running = False
+        self.task = None
+
+    def start(self):
```
tarekziade, line 30: to make you life easer, I would remove the sync layer here, and 

- make `start` a coroutine and replace `_cleanup` 
- make `shutdown` a coroutine
- add a `stop` sync method


And users can start it with:

```
cleaner = JobCleanUp()
asyncio.create_task(cleaner.start())
```

and stop it with  
```
cleaner.stop()
```

and in the cleanups, do

```
await cleaner.shutdown()
```

connectors/services/job_cleanup.py
```diff
@@ -0,0 +1,123 @@
...
+                    connector_id = job.connector_id
+                    connector = await self.connector_index.fetch_by_id(
+                        doc_id=connector_id
+                    )
+                    if connector is None:
+                        logger.warning(
+                            f"Could not found connector by id #{connector_id}"
+                        )
+                    else:
+                        await connector._sync_done(
```
wangch079, line 107: This line will give warnings in IDE as protected methods are not supposed to be accessed.

Inside the `_sync_done` method, this line will also raise error:

https://github.com/elastic/connectors-python/blob/020a512fd61443710bc777c5e04b9df09dc48443/connectors/byoc.py#L553-L556

Because `self._start_time` is never initialized. But the `sync_done` operation is actually done, it's just the log raising the error.

I'd like to leave it as this is going to be refactored in PR https://github.com/elastic/connectors-python/pull/375, and I don't want to introduce too much un-related changes in this PR

artem-shelkovnikov:
@wangch079 the PR is still in Draft state + has red CI, should it still be reviewed?

connectors/tests/test_job_cleanup.py
```diff
@@ -0,0 +1,82 @@
...
+    fetch_by_id.return_value = connector
+    orphaned_jobs.return_value = AsyncIterator([sync_job])
+    stuck_jobs.return_value = AsyncIterator([sync_job])
+    delete_jobs.return_value = {"deleted": 1, "failures": [], "total": 1}
+
+    service = create_service()
+    asyncio.get_event_loop().call_later(0.5, service.stop)
+    await service.run()
+
+    assert delete_indices.called
```
timgrein, line 80: Do you think it makes sense to also check that the arguments are correct using `call_args_list`?

connectors/services/job_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+                await self.sync_job_index.delete_indices(indices=list(content_indices))
+            result = await self.sync_job_index.delete_jobs(job_ids=job_ids)
+            if len(result["failures"]) > 0:
+                logger.error(f"Error found when deleting jobs: {result['failures']}")
+            logger.info(
+                f"Successfully deleted {result['deleted']} out of {result['total']} orphaned jobs."
+            )
+        except Exception as e:
+            logger.critical(e, exc_info=True)
+            self.raise_if_spurious(e)
```
timgrein, line 80: I think we could extract `raise_if_spurious` to a custom decorator, which we just put above a method.
We could save the repeated explicit try/except and it's already clear when reading the method declaration.

```
@raise_if_spurious
async def _process_orphaned_jobs(self):
  logger.info("Start cleaning up orphaned jobs...")
```

I did something similar with retrying, but probably shouldn't be part of this PR.

timgrein (commented):
Just two small comments about some assertions and a small nit about `raise_if_spurious`

connectors/tests/test_job_cleanup.py
wangch079: Done in https://github.com/elastic/connectors-python/pull/219/commits/9efd2e06620b596ff592e1c81994fdbf302ed348

connectors/services/job_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+                await self.sync_job_index.close()
+        return 0
+
+    async def _process_orphaned_jobs(self):
+        try:
+            logger.info("Start cleaning up orphaned jobs...")
+            connector_ids = [
+                connector.id
+                async for connector in self.connector_index.all_connectors()
+            ]
```
artem-shelkovnikov, line 54: Do we care about connector_ids here? I think all the data should be accessible from jobs, and we can pick up all orphaned jobs, does not matter which connector ids.

wangch079: We do need the `connector_ids`, as we need to find all jobs whose connector id does not exist.

connectors/tests/test_job_cleanup.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+@pytest.mark.asyncio
+@patch("connectors.byoc.SyncJobIndex.delete_jobs")
+@patch("connectors.byoc.SyncJobIndex.delete_indices")
+@patch("connectors.byoc.SyncJobIndex.stuck_jobs")
+@patch("connectors.byoc.SyncJobIndex.orphaned_jobs")
+@patch("connectors.byoc.ConnectorIndex.fetch_by_id")
+@patch("connectors.byoc.ConnectorIndex.supported_connectors")
+@patch("connectors.byoc.ConnectorIndex.all_connectors")
```
wangch079, line 56: Not sure if there's any better way to patch so many methods.

connectors/tests/test_byoc.py
```diff
@@ -956,56 +956,105 @@ def test_nested_get(nested_dict, keys, default, expected):
...
+        self.iter = iter(seq)
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        try:
+            return next(self.iter)
+        except StopIteration:
+            raise StopAsyncIteration
```
wangch079, line 973: This is an async wrapper of iterator, to patch async generator function. Not sure whether there's any better way to do so.

This is also imported in `test_job_cleanup.py`. Is there any shared test support file for it?

cc. @tarekziade

tarekziade: we can add a `connectors/tests/support.py` to share stuff like that

connectors/tests/test_job_cleanup.py
tarekziade: It depends on what you do with `SyncJobIndex` `ConnectorIndex` 

If you don't care about thee, you could try to pass a magic mock for those

wangch079:
buildkite test

wangch079:
buildkite test this

connectors/tests/test_job_cleanup.py
wangch079: Do you mean mock `ConnectorIndex` and `SyncJobIndex` directly?

like 

```python
@patch("connectors.byoc.SyncJobIndex")
@patch("connectors.byoc.ConnectorIndex")
async def test_cleanup_jobs(
    connector_index,
    sync_job_index
):
    connector_index.all_connectors.return_value = AsyncGeneratorFake([connector])
    ...
````

tarekziade (approved):

PR #220, opened 2022-12-04
Low-memory ingest attachment streamer

tarekziade:
Implements a disk-based file streamer. The streamer will be running on its separate process because:

- files operations will add more pressure on the event loop (CPU)
- we can tweak the max RSS allowed specifically for this attachments service
- the service will be very small in memory - as long as we don't import all libs
- the service can also process files from the crawler or the ruby service

**Refactorings**

- [x] Moved the sync service to its own module (`syncer.py`)
- [x] Created a `services` subdir
- [x] Created a base `Service` class
- [x] `cli.py` is now just focusing on starting the loop with the right service, `runner.py` is gone
- [x] Refactor tests
- [ ] make sure we don't import all connectors lib when we run just the file streamer

**New features**

- [x] Defined a disk-based standard 
- [x] Added `fstreamer.py` with the Stream service and the function to drop files for that service to grab
- [x] Use `ConcurrentRunner` in `byoei.py`
- [ ] Implement all tests
- [ ] Surface all options in the YAML file to configure the disk-based service
- [x] Allow connectors to pass an async generator for the `_attachment` field to stream a file. When it's detected, it's plugged into `FileDrop.drop_file` to the file is dumped on disk
- [x] use a source doc id in the json desc. the es doc is queried and the file is sent only of we find it
- [ ] Cloud support 
  - [ ] Add a program section in supervisord
  - [ ] Surface the max disk size option in the supervisor start script in the cloud so we can tweak it depending on the VM size
  - [ ] Add memmon support

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,7 +1,8 @@
 elasticsearch:
-  host: http://localhost:9200
+  host: http://0.0.0.0:9200
   username: elastic
   password: changeme
+  #api_key: "MFBsaTdvUUItV3hiQW5jQVNCNVI6NkNlZDFKVUJTLXFYaV9SV01jYVNQUQ=="
   ssl: true
   bulk:
     queue_max_size: 1024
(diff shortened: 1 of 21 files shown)


tarekziade:
depends on https://github.com/elastic/connectors-python/pull/584

tarekziade:
Converted to draft and creating small sub patches to split the work in easy to review patches

PR #221, opened 2022-12-05
[MySQL] Add Generic retry that loops over all the server calls.

parth-crest:
## Closes https://github.com/elastic/connectors-py/issues/187

PR include following changes:

- Added _connect() method which will behave as single point of connection for all the calls made to MySQL server.
- Added retry logic in the _connect() method.

#### Pre-Review Checklist

- [x] Added retry to every call being made to MySQL server


parth-crest:
Hey @tarekziade close() is not an async method we have to use it without await, it is throwing an error "None can not be awaitable". So we have removed it.
https://github.com/aio-libs/aiomysql/blob/bb8697d827981fcbf24841e0990cccebadf49fbf/aiomysql/pool.py#L89

praveen-elastic:
buildkite test this

tarekziade:
buildkite test this

tarekziade (approved):

PR #224, opened 2022-12-06
extract the concurrent runner logic

tarekziade:
Extracted the `ConcurrentRunner` class from https://github.com/elastic/connectors-python/pull/220 to reduce patch size


tarekziade:
fly-by fix for https://github.com/elastic/connectors-python/pull/218

connectors/utils.py
```diff
@@ -300,3 +300,31 @@ async def _wait_for_room(self, item):
...
+    def _cb(self, fut):
+        self.tasks.remove(fut)
+        if self.results_cb is not None:
+            if fut.exception():
+                raise fut.exception()
+            self.results_cb(fut.result())
+
+    async def put(self, coro):
+        while len(self.tasks) >= self.max_concurrency:
+            await asyncio.sleep(0)
```
artem-shelkovnikov, line 323: Curious about that part - is it gonna destroy CPU this way?

connectors/utils.py
```diff
@@ -300,3 +300,31 @@ async def _wait_for_room(self, item):
...
+class ConcurrentRunner:
+    def __init__(self, max_concurrency=5, results_cb=None):
+        self.max_concurrency = 5
+        self.tasks = []
+        self.results_cb = results_cb
+
+    def __len__(self):
+        return len(self.tasks)
+
+    def _cb(self, fut):
```
wangch079, line 314: Does `cb` mean `callback`? What is `fut`?

Can we use full name instead of abbreviation? It will make the code more readable.

Or is it a convention of Python? I found the param was named `coro` instead of `coroutine` in the official asyncio doc. TBH, I can't link `coro` to `coroutine` if I have no context.

connectors/utils.py
```diff
@@ -300,3 +300,31 @@ async def _wait_for_room(self, item):
...
+    async def put(self, coro):
+        while len(self.tasks) >= self.max_concurrency:
+            await asyncio.sleep(0)
+        fut = asyncio.create_task(coro())
+        self.tasks.append(fut)
+        fut.add_done_callback(self._cb)
+        return fut
+
+    async def wait(self):
+        await asyncio.gather(*self.tasks)
```
wangch079, line 330: Shall we make this implicitly called at the end of `put` method?

connectors/runner.py
```diff
@@ -180,16 +180,17 @@ async def poll(self, one_sync=False, sync_now=True):
...
                     ):
                         await self._one_sync(connector, es, sync_now)
-                    if one_sync:
-                        break
                 except Exception as e:
                     logger.critical(e, exc_info=True)
                     self.raise_if_spurious(e)
-
+                finally:
+                    if one_sync:
```
wangch079, line 192: Curious why it fails. Is it because some exceptions are raised and it never reach `if one_syc`?

Moving it to `finally` will ensure the check is done in each iteration of while looo?

connectors/runner.py
```diff
@@ -180,16 +180,17 @@ async def poll(self, one_sync=False, sync_now=True):
...
+                        break
                 if not one_sync:
                     await self._sleeps.sleep(self.idling)
```
wangch079, line 195: If it's `one_sync`, it will break in the above `finally` block. It seems we don't need the `if not one_sync` check here

connectors/utils.py
tarekziade: `fut` means `future`, as this is the object type here. But you are right, I should use better names, thanks

connectors/utils.py
tarekziade: Great question!

It will return immediately to the event loop, and the assumption is that since the other coroutines are busy pushing data to Elastic or retrieving some from the backend, the CPU should be fine. But in theory you are absolutely right, if we keep on coming back here it'll destroy the CPU.

A much better pattern would be to unblock it through `asyncio.Event`. e.g. another task was finished. I will do this :)

connectors/utils.py
tarekziade: `wait` will block until all tasks are done, so we can't use it in `put`

connectors/runner.py
tarekziade: yes exactly, I had an error in my MySQL connectors, and the raise bypassed the call

connectors/runner.py
tarekziade: good point thanks

connectors/utils.py
```diff
@@ -300,3 +300,36 @@ async def _wait_for_room(self, item):
...
+    def __len__(self):
+        return len(self.tasks)
+
+    def _callback(self, task):
+        self.tasks.remove(task)
+        self._task_over.set()
+        if self.results_callback is not None:
+            if task.exception():
+                raise fut.exception()
+            self.results_callback(task.result())
```
wangch079, line 321: I remembered I submitted this comment but I don't know where it is now

I assume the purpose of `results_callback` is to return the result if a task is supposed to return something. 

If the above assumption is true, I don't think it should be passed in `__init__`, but should be a param of `put` function, with default value of `None`. Because task A can return something but task B may not.

wangch079: What I meant was, we shouldn't add `results_callback` to `__init__`, and allow `None`.

Instead, we can make it a param of `put` function:
```python
def _callback(self, task, result=None):
    ...
    if result is not None:
        if task.exception():
            raise task.exception()
        result = task.result()


async def put(self, coroutine, result=None):
    ...
    task.add_done_callback(self._callback(result))
    return task
```

tarekziade: oh, I see what you mean. I see two use cases:

1. treat all results in the same callback
2. treat each result independently

1. is the most common use case as far as I know because in general, we send the same coroutine with different arguments and the collected results are homogeneous.

For 2. the fact that `put` returns the task allows you to call await on it and gets it `result()` directly, but you will need to do it in async code.

I can add both, it does not complexity the code -- we'll see if it's useful long term

connectors/byoei.py
```diff
@@ -104,8 +106,10 @@ def _bulk_op(self, doc, operation=OP_INDEX):
 
         raise TypeError(operation)
 
-    async def _batch_bulk(self, task_num, operations):
+    async def _batch_bulk(self, operations):
         # todo treat result to retry errors like in async_streaming_bulk
+        task_num = len(self.bulk_tasks)
```
seanstory, line 111: Is this the pythonic way to do this sort of thing? I would have expected `bulk_tasks.size()` or `bulk_tasks.count()` as opposed to this more functional approach.

connectors/tests/test_utils.py
```diff
@@ -164,3 +168,72 @@ def test_get_base64_value():
     """This test verify get_base64_value method and convert encoded data into base64"""
     expected_result = get_base64_value("dummy".encode("utf-8"))
     assert expected_result == "ZHVtbXk="
+
+
+@pytest.mark.asyncio
+async def test_concurrent_runner(patch_logger):
+    results = []
+
+    def _cb(result):
```
seanstory, line 177: I find shortened function names like this really difficult to understand at a glance. And if tests aren't easy to understand, they serve much less use as developer documentation. I'd recommend naming the functions with full names like `_task_callback`. I still don't know what `coro` is supposed to be.

connectors/tests/test_utils.py
```diff
@@ -164,3 +168,72 @@ def test_get_base64_value():
...
+        if i == 5:
+            raise Exception("I FAILED")
+        return i
+
+    runner = ConcurrentTasks(results_callback=_cb)
+    for i in range(10):
+        await runner.put(functools.partial(coro, i))
+
+    with pytest.raises(Exception):
+        await runner.wait()
```
seanstory, line 210: Should this also assert the contents of `results` are 1-10, excluding 5?

connectors/tests/test_utils.py
```diff
@@ -164,3 +168,72 @@ def test_get_base64_value():
...
+    results = []
+
+    def _cb(result):
+        results.append(result)
+
+    async def coro(i):
+        await asyncio.sleep(0)
+        return i
+
+    s_results = []
```
seanstory, line 224: Another case where naming could be clearer. `second_results` or `results2` maybe?

connectors/tests/test_utils.py
```diff
@@ -164,3 +168,72 @@ def test_get_base64_value():
...
+    runner = ConcurrentTasks(results_callback=_cb)
+    for i in range(1000):
+        if i == 3:
+            cb = _second_cb
+        else:
+            cb = None
+        await runner.put(functools.partial(coro, i), result_callback=cb)
+
+    await runner.wait()
+    assert results == list(range(1000))
```
seanstory, line 238: Shouldn't this exclude 3?

connectors/utils.py
```diff
@@ -338,3 +339,57 @@ async def _wait_for_room(self, item):
...
+            # rearm
+            self._task_over.clear()
+        task = asyncio.create_task(coroutine())
+        self.tasks.append(task)
+        task.add_done_callback(
+            functools.partial(self._callback, result_callback=result_callback)
+        )
+        return task
+
+    async def wait(self):
```
seanstory, line 394: Suggestion: rename this to `wait_for_all`, so that it's really explicit the difference in blocking behavior happening here vs in the `put` function (which just waits for 1 task, IIUC)

connectors/byoei.py
tarekziade: yes, that's the Pythonic way. In Python, container classes implement `__len__` and that gives you `len(object)`
But `.size()` or `.count()` is also something you would see.

connectors/tests/test_utils.py
tarekziade: `coro` is a pretty standard abbreviation for `coroutines` in Python (You will find this notation in the standard lib) but we can use `coroutine` for sure.

connectors/utils.py
tarekziade: they are not exactly the same thing =>

`wait` will wait for all tasks to be finished and matches the `asyncio.wait` primitive notation. 

`put` will wait for a coroutine to be added for execution but not for the task created for that coroutine to be finished.
(that follows the `put` vs. `put_nowait` semantics from asyncio.Queue https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.put)

But I see what you are saying, and in the spirit of short names (see PEP 8) I think we could rename `wait` to `join` and make it more explicit

That would make the class quite similar to `asyncio.Queue` https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.join

connectors/tests/test_utils.py
tarekziade: no because the first callback is for all results, the second for one via put

seanstory (approved):

PR #225, opened 2022-12-07
add linting to CI

seanstory:
Adds `make lint` to the buildkite command

### DIFF
--- a/.buildkite/run_dockerized_test.sh
+++ b/.buildkite/run_dockerized_test.sh
@@ -3,4 +3,4 @@ set -euo pipefail
 
 docker run --rm -v $PWD:/ci -w=/ci \
     python:3.10 \
-    /bin/bash -c  'make install PYTHON=python && make test'
+    /bin/bash -c  'make install PYTHON=python && make test && make lint'


tarekziade:
we should:

- use the `--check` for `black` everywhere in `make lint` and ensure it fails when black finds issues.
- add a `make format-code` target that calls black as is it called right now in `make lint`

.buildkite/run_dockerized_test.sh
```diff
@@ -3,4 +3,4 @@ set -euo pipefail
 
 docker run --rm -v $PWD:/ci -w=/ci \
     python:3.10 \
-    /bin/bash -c  'make install PYTHON=python && make test'
```
tarekziade, line 6: I would put `lint` before `test` because it may catch syntax errors

seanstory:
Screen Shot 2022-12-07 at 11 14 18 AM
Looks like it's working

.buildkite/run_dockerized_test.sh
tarekziade: sure ok no big deal.

tarekziade (approved):

artem-shelkovnikov:
🥳🥳🥳

PR #226, opened 2022-12-07
Refactor service layout

tarekziade:
Extracted the new services layout  from https://github.com/elastic/connectors-python/pull/220 to reduce patch size

- created a base class for services
- moved `runner.py` to `services/syncer.py`
- `cli.py` now can instanciate a service given an `--action`

The https://github.com/elastic/connectors-python/pull/220 patch will introduce a new service under `--action streamer` so we can use `elastic-ingest` to run a specific service if we want

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -6,15 +6,22 @@
 """
 Command Line Interface.
 
-Parses arguments and call connectors.runner.run with them.
+Parses arguments and call run() with them.
 """
 from argparse import ArgumentParser
 import os
 import logging
+import asyncio
+import signal
+import functools
 
+from envyaml import EnvYAML
+
+from connectors.logger import logger, set_logger
+from connectors.source import get_data_sources
+from connectors.utils import get_event_loop
 from connectors import __version__
-from connectors.runner import run
-from connectors.logger import set_logger
+from connectors.services.syncer import ConnectorService
 
 
 def _parser():
@@ -72,6 +79,34 @@ def _parser():
     )
 
     return parser
+
+
+def run(args):
+    """Runner"""
+    # just display the list of connectors
+    if args.action == "list":
+        logger.info("Registered connectors:")
+        config = EnvYAML(args.config_file)
+        for source in get_data_sources(config):
+            logger.info(f"- {source.__doc__.strip()}")
+        logger.info("Bye")
+        return 0
+
+    service = ConnectorService(args)
+    coro = service.run()
+    loop = get_event_loop()
+
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(service.shutdown, sig))
+
+    try:
+        return loop.run_until_complete(coro)
+    except asyncio.CancelledError:
+        return 0
+    finally:
+        logger.info("Bye")
+
+    return -1
 
 
 def main(args=None):
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -21,7 +21,7 @@ from connectors.logger import logger, set_logger
 from connectors.source import get_data_sources
 from connectors.utils import get_event_loop
 from connectors import __version__
-from connectors.services.syncer import ConnectorService
+from connectors.services.sync import SyncService
 
 
 def _parser():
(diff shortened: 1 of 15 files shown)


tarekziade:
@wangch079 by the way, right now it runs one single service with `--action xxx` but thinking about your job cleanup service, maybe we could add here a mechanism where we can launch with several actions e.g.

```
$ elastic-ingest --action sync,cleanup
```

And in that case, use the same event loop for both actions.

For file streams we could have a default value where they all run in the same loop:

```
$ elastic-ingest --action sync,cleanup,streamer
```

and in production have the ability to run the streamer separately

```
$ elastic-ingest --action sync,cleanup
$ elastic-ingest --action streamer
```

seanstory:
This looks like we're blending some taxonomy here. "Syncer" isn't a real word. We struggled with coming up with good names in the ruby repo too, but I think what we landed on that is synonymous to what you're expressing here was `SyncJobRunner`. However in your tests, you have instances like `FakeSource` being assigned to syncer variables. Since this code is public, I think we should land on some consistent terminology, on if we're calling them sources, runners, syncers, or something else. But right now there is not a lot of clarity.

It's also unclear to me why you've extracted out a Base when there's only a single subclass. Are you planning to add more Service implementations?

tarekziade:
> This looks like we're blending some taxonomy here. "Syncer" isn't a real word. We struggled with coming up with good names in the ruby repo too, but I think what we landed on that is synonymous to what you're expressing here was `SyncJobRunner`. However in your tests, you have instances like `FakeSource` being assigned to syncer variables. Since this code is public, I think we should land on some consistent terminology, on if we're calling them sources, runners, syncers, or something else. But right now there is not a lot of clarity.

Yeah, I am 100% for sharing the same taxonomy; that is important. You are right. 

Here we have 3 things:
- a module -- following PEP8 we could use "sync.py"
- a class in that module -- we can reuse SyncJobRunner
- variables that instantiate a SyncJobRunner. It could be "runner" or "sync_runner"

> 
> It's also unclear to me why you've extracted out a Base when there's only a single subclass. Are you planning to add more Service implementations?

Yes, that is coming from https://github.com/elastic/connectors-python/pull/220  and as explained in https://github.com/elastic/connectors-python/pull/226#issuecomment-1344489587 we're planning to add other services (streamer, cleaner, etc)

tarekziade:
added @danajuratoni as a reviewer for the taxonomy part

tarekziade:
> However in your tests, you have instances like `FakeSource` being assigned to syncer variables. 

Can you show me here? I could not find it. 

The `FakeSource` class is replacing a source (see https://github.com/elastic/connectors-python/tree/main/connectors/sources ) so the test can run without having to use a real source of data. It's not related to the service class

seanstory:
> Here we have 3 things:
> 
> * a module -- following PEP8 we could use "sync.py"
> * a class in that module -- we can use SyncService (the design is different from Ruby here)
> * variables that instantiate a SyncService. It could be "service" or "sync_service"
> 
> >

👍 I like that, and that makes a lot more sense to me



> Yes, that is coming from #220 and as explained in [#226 (comment)](https://github.com/elastic/connectors-python/pull/226#issuecomment-1344489587) we're planning to add other services (streamer, cleaner, etc)

Gotcha, thanks that makes sense now 👍 



> > However in your tests, you have instances like `FakeSource` being assigned to syncer variables.
> 
> Can you show me here? I could not find it.

I was looking at lines like https://github.com/elastic/connectors-python/pull/226/files#diff-24e065d022073f09333141158003d1d87a477ce2214ba0ea2041ae09f6380e5cR23 (it's no longer `syncer` but `test_sync`, which is closer, but still conflates with what we've previously mean by "source")

tarekziade:
> (it's no longer `syncer` but `test_sync`, which is closer, but still conflates with what we've previously mean by "source")

ah ok. it used to be `test_syncer:FakeSource` -- it's the fully qualified name of the `FakeSource` class so the framework can find it and instanciates it.  That notation is because the `FakeSource` class is inside that `test_sync` module.

tarekziade:
I have moved all fake sources to a dedicate module in the tests, since they are used in many tests, not only `test_sync` -- this will be cleaner

seanstory (approved):
Thanks, I find this much more clear now. 🎉

tarekziade:
thanks for the very useful feedback and reviews

PR #235, opened 2022-12-09
Added the `--uvloop` option

tarekziade:
On Linux or macOS, `--uvloop` will use the uvloop (NodeJS) eventloop instead of the system's queue/kqueue. 

Deactivated by default.

### DIFF
--- a/requirements/framework.txt
+++ b/requirements/framework.txt
@@ -8,3 +8,4 @@ pympler==1.0.1
 guppy3==3.1.2
 cron-schedule-triggers==0.0.11
 pytz==2019.3
+uvloop==0.17.0
--- a/requirements/framework.txt
+++ b/requirements/framework.txt
@@ -8,4 +8,4 @@ pympler==1.0.1
 guppy3==3.1.2
 cron-schedule-triggers==0.0.11
 pytz==2019.3
-uvloop==0.17.0
+uvloop==0.17.0; sys_platform != 'win32'
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@
 import sys
 import os
 from setuptools import setup, find_packages
-
+from setuptools._vendor.packaging.markers import Marker
 
 try:
     ARCH = os.uname().machine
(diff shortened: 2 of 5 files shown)


connectors/runner.py
```diff
@@ -16,6 +16,13 @@
 import time
 import functools
 
+# activate uvloop if present
+try:
+    import uvloop
```
wangch079, line 21: If we decided to use `uvloop`, we don't need this check? Simply use it directly?

```python
import uvloop

asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```

wangch079 (approved):
LGTM

connectors/runner.py
tarekziade: I pushed the patch for Crest to try it out on a bench; if the results are good, we will need to make the change more cleanly because `uvloop` is not supported on `win32`

So we'd move this code in the runner and provide a new option, e.g., `--event-loop uvloop`

From my recent profiling, we spend a lot of time calling `kqueue.select` and it's probable that switching to `uvloop` will improve performances. `libuv` is the loop used by NodeJS and it usually makes Py app like our faster than Go apps :)

artem-shelkovnikov:
Curious how does it affect the runtime of python connectors?

tarekziade:
> Curious how does it affect the runtime of python connectors?

Yeah we're doing benches to compare and see if it eases CPU and event loop pressure

setup.py
```diff
@@ -33,6 +33,25 @@
...
 
 
+def extract_req(req):
+    req = req.strip().split(";")
+    if len(req) > 1:
+        env_marker = req[-1]
+        env_marker = env_marker.replace(" ", "").strip()
+        # for now we just recognize sys_platform
+        if "sys_platform==" in env_marker:
+            target = env_marker.split("!=")[-1].strip()
```
seanstory, line 43: This line has `"!=` but the `if` conditional above has `==`.  Bug?

requirements/framework.txt
```diff
@@ -8,3 +8,4 @@ pympler==1.0.1
 guppy3==3.1.2
 cron-schedule-triggers==0.0.11
 pytz==2019.3
+uvloop==0.17.0; sys_platform != 'win32'
```
seanstory, line 11: we have separate requirements texts for different architectures, why not make this a different file? Modifying the syntax of your requirements.txt seems.... nonstandard?

tarekziade: That's a standard I've created in Python 10 years ago haha https://peps.python.org/pep-0508/#environment-markers
this is fully supported by pip and al

setup.py
tarekziade: yeah that's a typo, good catch thanks!

requirements/framework.txt
tarekziade: see https://github.com/elastic/connectors-python/pull/197#issuecomment-1326799639

PR #237, opened 2022-12-09
clean up spurious output and fix mysql test

tarekziade:
This cleans up the output when we run `make test` and also fix the small coroutine error in MySQL test


praveen-elastic (approved):

connectors/sources/tests/test_mysql.py
```diff
@@ -138,15 +146,15 @@ async def test_close_with_connection_pool():
     """Test close method of MySql with connection pool"""
     # Setup
     source = create_source(MySqlDataSource)
-
-    source.connection_pool = Connection
```
parth-crest, line 142: Since we are not using the connection class now, we can remove `close()` and `wait_closed()` from the connection class.

PR #239, opened 2022-12-14
starting to refactor tarekziade/fstreamer

tarekziade:
Replaces https://github.com/elastic/connectors-python/pull/220

### DIFF
--- a/connectors/services/attachments.py
+++ b/connectors/services/attachments.py
@@ -1,0 +1,250 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+""" File streamer service.
+
+Provides:
+
+- `AttachmentService` -- watches a dir for files to upload to Elasticsearch
+- `FileDrop` -- used by the producer to drop files on disk
+
+"""
+# TODO
+#
+# - harden all requests (catch errors, retries)
+# - if a file keep on failing, put it in a failing dir
+# - add a max file size
+#
+import os
+import json
+import functools
+import re
+
+import aiohttp
+import aiofiles
+
+from connectors.logger import logger
+from connectors.utils import (
+    CancellableSleeps,
+    get_event_loop,
+    ConcurrentTasks,
+)
+from connectors.services.base import BaseService
+
+
+SUPPORTED_TEXT_FILETYPE = [".py", ".rst", ".rb", ".sh", ".md", ".txt"]
+
+# 250MB max disk size
+ONE_MEG = 1024 * 1024
+DEFAULT_MAX_DIR_SIZE = 250
+
+# Program to encode in base64 -- we could compile a SIMD-aware one
+# for an extra performance boost
+BASE64 = "base64"
+SANITIZER = re.compile("[^0-9a-zA-Z]+")
+
+
+class FileDrop:
+    def __init__(self, config):
+        self.config = config
+        aconfig = self.config["attachments"]
+        self.drop_dir = aconfig["drop_dir"]
+        self.max_disk_size = (
+            aconfig.get("max_disk_size", DEFAULT_MAX_DIR_SIZE) * ONE_MEG
+        )
+        self.elastic_config = self.config["elasticsearch"]
+
+    def can_drop(self, drop_directory):
+        current_size = 0
+        for item in os.scandir(drop_directory):
+            try:
+                if not item.is_file():
+                    continue
+                current_size += d.stat().st_size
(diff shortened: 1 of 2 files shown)


connectors/services/attachments.py
```diff
@@ -0,0 +1,250 @@
...
+        """Calls the system base64 utility to create a base64-encoded file
+
+        The base64 utility is a stream encoder, the file will not be fully
+        loaded into memory.
+
+        This is a blocking method.
+        """
+        cmd = f"{BASE64} {filename} > {filename}.b64"
+        logger.debug(f"[{os.path.basename(filename)}] Running {cmd}")
+        os.system(cmd)
```
sethmlarson, line 140: Doesn't asyncio support spawning a subprocess?

connectors/services/attachments.py
```diff
@@ -0,0 +1,250 @@
...
+
+SUPPORTED_TEXT_FILETYPE = [".py", ".rst", ".rb", ".sh", ".md", ".txt"]
+
+# 250MB max disk size
+ONE_MEG = 1024 * 1024
+DEFAULT_MAX_DIR_SIZE = 250
+
+# Program to encode in base64 -- we could compile a SIMD-aware one
+# for an extra performance boost
+BASE64 = "base64"
```
sethmlarson, line 45: Should discover the base64 binary once on startup and then use the absolute path throughout the execution of the program to avoid PATH poisoning.

connectors/services/attachments.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        b64_file = await get_event_loop().run_in_executor(None, self._to_b64, filename)
+
+        # XXX adding .txt to all known text files
+        if os.path.splitext(filename)[-1] in SUPPORTED_TEXT_FILETYPE:
+            filename += ".txt"
+
+        try:
+            async with aiofiles.open(b64_file, "rb") as f:
+                yield b'{"filename":"' + filename.encode("utf8") + b'","data":"'
```
sethmlarson, line 157: What if the filename has double quotes in it?

connectors/services/attachments.py
```diff
@@ -0,0 +1,250 @@
...
+                chunk = (await f.read(chunk_size)).strip()
```
sethmlarson, line 158: Not sure if this service supports Python 3.8+ only, but if so this is the perfect opportunity for a walrus operator:

```python
while chunk := (await f.read(chunk_size)).strip()):
    yield chunk
```

connectors/services/attachments.py
```diff
@@ -0,0 +1,250 @@
...
+                while chunk:
+                    yield chunk
+                    chunk = (await f.read(chunk_size)).strip()
+                yield b'"}'
+        finally:
+            if os.path.exists(b64_file):
```
sethmlarson, line 164: Ask for forgiveness not permission:

```python
try:
    os.remove(b64_file)
except OSError as e:
    if e.errno != errnos.ENOENT:
        raise
```

connectors/services/attachments.py
```diff
@@ -0,0 +1,250 @@
...
+                os.remove(b64_file)
+
+    def _prepare_dir(self, name):
+        default_dir = os.path.join(os.path.dirname(self.config_file), "attachments")
+        path = self._config.get(f"{name}_dir", os.path.join(default_dir, name))
+        os.makedirs(path, exist_ok=True)
+        return path
+
+    async def _send_attachment(self, desc_file, desc, session):
+        fn = desc["filename"]
```
sethmlarson, line 174: nit: I wasn't able to figure out what `fn` meant just by looking at it down below

connectors/services/attachments.py
```diff
@@ -0,0 +1,250 @@
...
+        headers = {"Content-Type": "application/json"}
+        result = os.path.join(self.results_directory, fn + ".json")
+        worked = False
+        resp = None
+
+        async with session.put(url, data=gen(), headers=headers) as resp:
+            async with aiofiles.open(result, "w") as f:
+                resp = await resp.json()
+                logger.debug(f"[{fn}] Done, results in {result}")
+                if resp.get("result") in ("updated", "created"):
```
sethmlarson, line 188: Is `noop` also possible here?

sethmlarson (commented):
Looks good, a handful of thoughts for you:

connectors/services/attachments.py
tarekziade: yes that was just for the draft POC

tarekziade: `run_in_executor` along with a `ProcessPoolExecutor` see https://pymotw.com/3/asyncio/executors.html

tarekziade:
back to refactoring in https://github.com/elastic/connectors-python/pull/220

PR #241, opened 2022-12-14
Remove query param from fetch_documents

artem-shelkovnikov:
Noticed that `query` is passed into the method but never used, instead on the first line it's already overridden:

```python
 query = QUERIES["TABLE_PRIMARY_KEY"].format(database=database, table=table)
```

It looks like the `query` argument is not needed at all and is a leftover from recent refactorings.

Small additional note:

The method first assigns to `query` here:

```python
 query = QUERIES["TABLE_PRIMARY_KEY"].format(database=database, table=table)
```

And then reassigns it with:

```python
query = QUERIES["TABLE_DATA"].format(database=database, table=table)
```

It looks like a dangerous practice, should we rename them to respectively `table_primary_key_query` and `table_data_query`?

### DIFF
--- a/connectors/sources/tests/test_mysql.py
+++ b/connectors/sources/tests/test_mysql.py
@@ -245,8 +245,6 @@ async def test_fetch_documents():
     source.connection_pool.acquire.cursor = Cursor
     source.connection_pool.acquire.cursor.is_connection_lost = False
 
-    query = "select * from table"
-
     # Execute
     with mock.patch.object(
         aiomysql, "create_pool", return_value=(await mock_mysql_response())
@@ -256,7 +254,7 @@ async def test_fetch_documents():
         mock.patch("source._connect", return_value=response)
 
     response = source.fetch_documents(
-        database="database_name", table="table_name", query=query
+        database="database_name", table="table_name"
     )
 
     # Assert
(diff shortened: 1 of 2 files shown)


tarekziade:
yeah that is not great, I would add a global function to make things a bit clearer and use it in `_connect`

```
def prepare_query(name, **kw):
    return QUERIES[name].format(**kw)

xxx
 async def _connect(self, query_name, query_args, fetch_many=False):
      query = prepare_query(query_name, **args)
      ...
xxx
```

and clean up all the code that formatting queries

parth-crest:
Hi @tarekziade @artem-shelkovnikov , We can pick up this change and refactor _connect() method as @tarekziade suggested.
We can raise the separate PR as we will not be able to commit in this branch.
Let us know if you want us to work on this change.

connectors/sources/tests/test_mysql.py
```diff
@@ -245,8 +245,6 @@ async def test_fetch_documents():
     source.connection_pool.acquire.cursor = Cursor
     source.connection_pool.acquire.cursor.is_connection_lost = False
 
-    query = "select * from table"
```
parth-crest, line 248: We are using this `query` variable  at line 252 `response = source._connect(query)` 
we can not remove this variable.

Please see the [reference](https://github.com/elastic/connectors-python/blob/e4500202993faf5c0a7f608d74d20ceb8b5cba28/connectors/sources/tests/test_mysql.py#L252)

artem-shelkovnikov: I see, I've removed it by mistake

artem-shelkovnikov:
@parth-crest go for it! I will close this PR meanwhile.

PR #242, opened 2022-12-15
Move get_docs method to ESIndex class

vidok:
### Part of the https://github.com/elastic/enterprise-search-team/issues/3252

In this PR I will move elasticsearch-related logic to from `BYOIndex`  to `ESIndex`. `ESIndex` will be also used in a new `SyncJobIndex` class. 

#### Pre-Review Checklist


connectors/utils.py
```diff
@@ -405,3 +405,77 @@ def get_event_loop():
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
     return loop
+
+
+DEFAULT_PAGE_SIZE = 100
+
+
+class ESIndex(ESClient):
```
tarekziade, line 413: `utils` is usually where we put stuff we don't know where to put else where until they grow into something more substantial and meaningful -- like a few functions and classes.

Since we starting to have a couple of classes for Elasticsearch interactions, I think it would make sense to move them to `es.py` -- not `elasticsearch.py` to avoid shadowing the elasticseatch library that already uses that namespace.

tarekziade (approved):
awesome, just one suggestion about moving es related stuff to a new module

connectors/utils.py
vidok: Yep, it makes to me 👍🏻

connectors/tests/test_es.py
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+
```
tarekziade, line 6: you can group imports when it's only two like that

connectors/tests/test_es.py
```diff
@@ -0,0 +1,33 @@
...
+    mock_responses.post(
+        "http://nowhere.com:9200/index/_search?expand_wildcards=hidden",
+        headers=headers,
+        status=200,
+        payload={"hits": {"total": {"value": 1}, "hits": [{"id": 1}]}},
+    )
+    with pytest.raises(NotImplementedError) as _:
+        async for c in index.get_all_docs():
+            c
```
tarekziade, line 33: That might make `make lint` unhappy

do this:

```
async for doc_ in index.get_all_docs():
    pass
```

tarekziade (approved):
Just small nits!

PR #243, opened 2022-12-16
merge the headers

tarekziade:
## Related https://github.com/elastic/enterprise-search-team/issues/3553

Merge the headers found in the original ent search file

- [x] verify that the case insensitivity is respected


connectors/tests/test_sync.py
```diff
@@ -488,10 +488,16 @@ async def test_connector_service_poll_trace_mem(
...
 @pytest.mark.asyncio
 async def test_connector_service_poll_with_entsearch(
-    mock_responses, patch_logger, patch_ping, set_env, catch_stdout
+    mock_responses, patch_logger, patch_ping, set_env  # , catch_stdout
 ):
     with mock.patch.dict(os.environ, {"ENT_SEARCH_CONFIG_PATH": ES_CONFIG}):
-        await set_server_responses(mock_responses)
+
+        def connectors_read(url, **kw):
+            assert kw["headers"]["x-elastic-auth"] == "SomeYeahValue"
```
carlosdelest, line 496: Is the dictionary case-insensitive for keys? As it's looking for `x-elastic-auth` instead of `X-Elastic-Auth`

tarekziade: Yeah HTTP headers are case insensitive, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

seanstory: Fun fact, they _should_ be insensitive, but are not in practice. They persist case across the network, and application code may not be designed to be insensitive, and therefore casing matters. [This bit us recently with Zendesk](https://github.com/elastic/enterprise-search-team/issues/1962#issuecomment-1176676737).

tarekziade: Gotcha. But all the python clients behave like this; I assume Elasicsearch works with case insensitive? I guess it's easy to verify

seanstory (approved):
LGTM. Can you make a follow-up issue to look into SSL options, which is another gap that we don't currently address between ent-search and connectors?

tarekziade:
the code that looks for the header is Go and uses https://pkg.go.dev/net/http@go1.19.4#CanonicalHeaderKey:~:text=HTTP%20defines%20that%20header%20names%20are%20case%2Dinsensitive.%20The

and we should have any proxy in-between that would drop the header for a casing reason.

That said, I will see if we can keep casing in the Py client

PR #245, opened 2022-12-19
Tweak perf8 options

tarekziade:
- plugged the `--asyncstats` plugin
- exposed `DATA_SIZE` so we can run various sizes of perf tests
- exposed `--refresh-rate` so we can lower the size of the reports on very long runs

### DIFF
--- a/connectors/sources/tests/fixtures/mysql/remove.py
+++ b/connectors/sources/tests/fixtures/mysql/remove.py
@@ -5,8 +5,11 @@
 #
 from mysql.connector import connect
 import random
+import os
+
 
 DATABASE_NAME = "customerinfo"
+NUM_TABLES = int(os.environ.get('NUM_TABLES', '30'))
 
 
 def main():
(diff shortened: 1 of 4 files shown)


connectors/sources/tests/fixtures/mysql/remove.py
```diff
@@ -5,16 +5,19 @@
...
+NUM_TABLES = int(os.environ.get('NUM_TABLES', '30'))
 
 
 def main():
     """Removes 10 random items per table"""
     database = connect(host="127.0.0.1", port=3306, user="root", password="changeme")
     cursor = database.cursor()
     cursor.execute(f"USE {DATABASE_NAME}")
-    for table in range(15):
+    for table in range(NUM_TABLES):
```
navarone-feekery, line 20: Are there any repercussions from increasing this default value from `15` to `30`?

navarone-feekery (approved):
LGTM with one question

connectors/sources/tests/fixtures/mysql/remove.py
tarekziade: yes!

I've changed it a bit by the way. The number of table directly changes the size of the data that's being synced:

- 5 tables is 50k docs
- 30 tables is 200k docs

I made this variable more generic with 3 values:
- small
- medium
- large

and the connector fixture can decide what those mean.

@seanstory I guess this is the first step to what we've discussed around perf test automation

PR #246, opened 2022-12-19
reorganize the namespaces

tarekziade:
Following up @vidok recent work on creating a base class for all Elasticsearch index operations, I realized we have the `index.py` file that is the code that's doing the mapping and settings work on a new index. 

This name is adding confusion at this point, and the root namespace could benefit form a sub package.

This patch is doing the following reorg:

- the `es` and `index` modules are moved as `index` and `settings` under `connectors.es`
- the `connector.es` module (init) exposes the public API used in the framework


navarone-feekery (approved):
Nice one!

PR #251, opened 2022-12-20
Add rule for flake8 to enforce imports order + make autoformat now sorts imports

artem-shelkovnikov:
Changes in the PR:

1. Add [isort](https://pycqa.github.io/isort/) package.
2. `isort` is used in `make lint` to verify that all imports are sorted in the correct order
3. `isort` is used in `make autoformat` to automatically sort imports
4. `make autoformat` is ran to sort imports correctly.

It's safe to make a github pre-commit hook to do `make autoformat` that will do the sorting for you.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,3 @@
+[settings]
+profile = black
+skip=lib
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,3 +1,3 @@
 [settings]
 profile = black
-skip=lib
+skip=lib,bin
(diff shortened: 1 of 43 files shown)


artem-shelkovnikov:
@tarekziade while it works, I'm not sure if it's better - we don't use flake8 to do `autoformat` thus I'll still to have the `isort` package installed and used for `make autoformat`.

Or should we change the `make autoformat` to also run `flake8 --fix`?

tarekziade:
> @tarekziade while it works, I'm not sure if it's better - we don't use flake8 to do `autoformat` thus I'll still to have the `isort` package installed and used for `make autoformat`.
> 
> Or should we change the `make autoformat` to also run `flake8 --fix`?

flake8 don't have this option, so you're right!

PR #255, opened 2022-12-20
more cleanups

tarekziade:
Silence!

### DIFF
--- a/connectors/sources/tests/test_gcs.py
+++ b/connectors/sources/tests/test_gcs.py
@@ -72,7 +72,7 @@ async def test_empty_configuration():
 
 
 @pytest.mark.asyncio
-async def test_ping_for_successful_connection():
+async def test_ping_for_successful_connection(catch_stdout, patch_logger):
     """Tests the ping functionality for ensuring connection to Google Cloud Storage."""
 
     # Setup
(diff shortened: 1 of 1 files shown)


connectors/sources/tests/test_gcs.py
```diff
@@ -72,7 +72,7 @@ async def test_empty_configuration():
 
 
 @pytest.mark.asyncio
-async def test_ping_for_successful_connection():
+async def test_ping_for_successful_connection(catch_stdout, patch_logger):
```
navarone-feekery, line 75: So I'm sure this is fine, but I'm curious; where do these arguments come from? Is this just part of the testing lib or something?

navarone-feekery (approved):
LGTM with a comment

connectors/sources/tests/test_gcs.py
tarekziade: yes, this is part of `pytest` fixtures.

every function decorated with `@pytest.fixture` in https://github.com/elastic/connectors-python/blob/main/connectors/conftest.py can be used in tests as arguments like this.

when they are added, the fixture is used as a wrapper :

```
def fixture():
    # do something before the test
    yield  # this calls the tests
    # do something after the test
```

tarekziade: more on fixtures https://docs.pytest.org/en/6.2.x/fixture.html

tarekziade:
The backport was not needed (GCS is 8.7 only)

PR #258, opened 2022-12-21
Implement the nightly functional test

tarekziade:
replaces https://github.com/elastic/connectors-python/pull/94


connectors/sources/tests/fixtures/mysql/Makefile
```diff
@@ -1,31 +0,0 @@
-.PHONY: run-stack stop-stack load-data
```
artem-shelkovnikov, line 1: Why was this file removed? It's not needed any more?

tarekziade: it adds a lot of friction and requires to allow overwriting
- custom python
- custom venv

I am calling everything from `ftest` now and planning to do this for all fixtures.

so the "standard" would become a few python functions

artem-shelkovnikov: Cool, makes sense!

artem-shelkovnikov (approved):

PR #259, opened 2022-12-21
Add BasicRule class matching documents against a basic rule

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3600 ##

This PR introduces the BasicRule class which can check if a document matches the constraints defined by a basic rule. The BasicRule tries to coerce the type of the rule value to the type of the document value. It also exposes a property to check whether the used policy was `include` or not (will be important for the rules engine and its execution later on during a sync).

The following comparisons are supported:
- Starts with
- Ends with
- Contains
- Regex
- Less than
- Greater than
- Equals

#### Pre-Review Checklist

### DIFF
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -3,12 +3,11 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
-import datetime
-import re
-
 from connectors.logger import logger
+import datetime
 from dateutil.parser import *
 from enum import Enum
+import re
 
 
 class BasicRule:
@@ -118,7 +117,8 @@ class BasicRule:
                 case _:
                     return str(self.value)
         except ValueError as e:
-            logger.debug(f"Failed to coerce value '{self.value}' ({type(self.value)}) based on document value '{doc_value}' ({type(doc_value)}) due to error: {type(e)}: {e}")
+            logger.debug(
+                f"Failed to coerce value '{self.value}' ({type(self.value)}) based on document value '{doc_value}' ({type(doc_value)}) due to error: {type(e)}: {e}")
             return str(self.value)
 
     @staticmethod
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -12,21 +12,69 @@ from dateutil.parser import ParserError, parser
 from connectors.logger import logger
 
 
+def to_float(value):
+    try:
+        return float(value)
+    except ValueError:
+        return value
+
+
+def to_datetime(value):
+    try:
+        date_parser = parser()
+        parsed_date_or_datetime = date_parser.parse(timestr=value)
+
+        if isinstance(parsed_date_or_datetime, datetime.datetime):
+            return parsed_date_or_datetime
+        elif isinstance(parsed_date_or_datetime, datetime.date):
+            # adds 00:00 to the date and returns datetime
+            return datetime.datetime.combine(parsed_date_or_datetime, datetime.time.min)
+        else:
+            return value
+
+    except ParserError:
+        return value
+
+
+def to_bool(value):
+    if len(value) == 0 or re.match("^(false|f|no|n|off)$", value):
+        return False
+
+    if re.match("^(true|t|yes|y|on)$", value):
+        return True
+
+    return value
+
+
+def try_coerce(value):
+    coerced_value = to_float(value)
+
+    if isinstance(coerced_value, str):
+        coerced_value = to_datetime(value)
+
+    if isinstance(coerced_value, str):
+        coerced_value = to_bool(value)
+
+    return coerced_value
+
+
+class Rule(Enum):
+    EQUALS = 1
+    STARTS_WITH = 2
+    ENDS_WITH = 3
+    CONTAINS = 4
+    REGEX = 5
+    GREATER_THAN = 6
+    LESS_THAN = 7
+
+
+class Policy(Enum):
+    INCLUDE = 1
+    EXCLUDE = 2
+
+
 class BasicRule:
     DEFAULT_RULE_ID = "DEFAULT"
-
-    class Policy(Enum):
-        INCLUDE = 1
-        EXCLUDE = 2
-
-    class Rule(Enum):
-        EQUALS = 1
-        STARTS_WITH = 2
-        ENDS_WITH = 3
-        CONTAINS = 4
-        REGEX = 5
-        GREATER_THAN = 6
-        LESS_THAN = 7
 
     def __init__(self, id_, order, policy, field, rule, value):
         self._id = id_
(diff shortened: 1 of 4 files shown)


tarekziade (commented):
This looks great; I did a first pass for making it "Pythonic"

One general remark:
- we can use functions in modules; no need to encapsulate everything in the class 
- therefore, static methods are rarely used unless you do metaprogramming
- we don't hide stuff (with `__` or property) unless there's a technical reason. if something is not "public", it's usually prefixed with a single `_`

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def matches_document(self, document):
+        if self.is_default_rule:
+            return True
+
+        if self.field not in document:
+            return False
+
+        document_value = document[self.field]
+        coerced_rule_value = self.coerce_rule_value_based_on_document_value(
```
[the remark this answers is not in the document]
timgrein: I would prefer to keep this name as this already got me caught up in Ruby. If we name it `coerce` it's not obvious that the `rule_value` is coerced living inside the object and **not** the `document_value` passed as a parameter. There's also a thread about it in Slack with Sean.

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,177 @@
...
+        return self.id_ == BasicRule.DEFAULT_RULE_ID
+
+    @property
+    def is_include(self):
+        return self.policy == BasicRule.Policy.INCLUDE
+
+    def coerce_rule_value_based_on_document_value(self, doc_value):
+        try:
+            match doc_value:
+                case str():
```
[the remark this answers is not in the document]
timgrein: Wouldn't match case be more pythonic with [PEP 634](https://peps.python.org/pep-0634/#class-patterns)?

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,177 @@
...
+        if self.field not in document:
+            return False
+
+        document_value = document[self.field]
+        coerced_rule_value = self.coerce_rule_value_based_on_document_value(
+            document_value
+        )
+
+        match self.rule:
+            case BasicRule.Rule.STARTS_WITH:
```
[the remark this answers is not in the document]
timgrein: I would prefer to keep it here as we need three different values: `document_value`, `coerced_rule_value` and `self.value`. The API would look a bit weird in `Rule` IMO.

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,177 @@
...
+                return str(document_value).startswith(self.value)
+            case BasicRule.Rule.ENDS_WITH:
+                return str(document_value).endswith(self.value)
+            case BasicRule.Rule.CONTAINS:
+                return self.value in str(document_value)
+            case BasicRule.Rule.REGEX:
+                return re.match(self.value, str(document_value)) is not None
```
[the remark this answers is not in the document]
timgrein: It's not long lived, but I'll definitely remember that for future work with regex!

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+    try:
+        date_parser = parser()
+        parsed_date_or_datetime = date_parser.parse(timestr=value)
+
+        if isinstance(parsed_date_or_datetime, datetime.datetime):
+            return parsed_date_or_datetime
+        elif isinstance(parsed_date_or_datetime, datetime.date):
+            # adds 00:00 to the date and returns datetime
+            return datetime.datetime.combine(parsed_date_or_datetime, datetime.time.min)
+        else:
```
tarekziade, line 32: nit: you can remove `else` and dedint `reurn value`

it's good practice to make sure we're treating that last as default

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+            return value
+
+    except ParserError:
+        return value
+
+
+def to_bool(value):
```
tarekziade, line 39: are you making the assumption value is a string?

if not, this function will return implicit results depending on the type.

```
>>> bool(None)
False
>>> bool(object())
True
>>> bool(list())
False
```

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+    if len(value) == 0 or re.match("^(false|f|no|n|off)$", value):
+        return False
+
+    if re.match("^(true|t|yes|y|on)$", value):
+        return True
+
+    return value
+
+
+def try_coerce(value):
```
tarekziade, line 49: why the `try` prefix?

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+    coerced_value = to_float(value)
+
+    if isinstance(coerced_value, str):
+        coerced_value = to_datetime(value)
```
tarekziade, line 53: what defines the `float then, datetime, then bool` order?

connectors/filtering/basic_rule.py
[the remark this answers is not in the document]
tarekziade: ok fair enough. Python tries to keep names as concise as possible (see pep8)

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+            case Rule.REGEX:
+                return re.match(self.value(), str(document_value)) is not None
+            case Rule.LESS_THAN:
+                return document_value < coerced_rule_value
+            case Rule.GREATER_THAN:
+                return document_value > coerced_rule_value
+            case Rule.EQUALS:
+                return document_value == coerced_rule_value
+
+    def id_(self):
```
tarekziade, line 126: For this method and all the following ones, why are you creating a method and not just exposing the attribute directly, so one may use `rule.policy` ?

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._field
+
+    def rule(self):
+        return self._rule
+
+    def value(self):
+        return self._value
+
+    def is_default_rule(self):
+        return self.id_() == BasicRule.DEFAULT_RULE_ID
```
tarekziade, line 145: in Python each indirection slows down execution, so I would remove the `id_` function and use the attribute directly here

connectors/filtering/basic_rule.py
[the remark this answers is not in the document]
tarekziade: I have never used that pattern before because each case creates an instance of an object just for comparison, and this is a very new syntax -- a bit of uncharted territory. 

So I don't know :) 

We can do this; it does look nice. In general, we try to avoid extra instance creations (not pythonic), and this is my first reaction when reviewing.

tarekziade:
looks good thanks for all the changes, I added a few more comments but it's landable

tarekziade (approved):

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+
+    except ParserError:
+        return value
+
+
+def to_bool(value):
+    if len(value) == 0 or re.match("^(false|f|no|n|off)$", value):
+        return False
+
+    if re.match("^(true|t|yes|y|on)$", value):
```
tarekziade, line 43: you should compile this one as a global:

```
IS_BOOL = re.compile("^(true|t|yes|y|on)$")
```

and then here:

`if IS_BOOL.match(value)`

connectors/filtering/basic_rule.py
```diff
@@ -0,0 +1,167 @@
...
+            return datetime.datetime.combine(parsed_date_or_datetime, datetime.time.min)
+        else:
+            return value
+
+    except ParserError:
+        return value
+
+
+def to_bool(value):
+    if len(value) == 0 or re.match("^(false|f|no|n|off)$", value):
```
tarekziade, line 40: same remark, compile your regex

tarekziade: also: if you want to be case insensitive here, use `re.I`

connectors/filtering/basic_rule.py
timgrein: Yes the value will be a string

connectors/filtering/basic_rule.py
timgrein: AFAIK this order comes from how ent-search coerces strings to types

PR #261, opened 2022-12-23
Extract pre-flight check to a function + rearrange arguments in _run function

artem-shelkovnikov:
A couple small changes:

1. Rearranged arguments in `_run` function to group them more logically (same things happen together)
2. Moved all preflight check logic to a separate method


tarekziade (approved):
Nice!

PR #262, opened 2022-12-23
Fix e2e in the CI for AWS

tarekziade:
Make sure all e2e test work in buildkite


tarekziade:
I got `Amazon` working

connectors/sources/s3.py
```diff
@@ -26,9 +26,11 @@
...
 DEFAULT_PAGE_SIZE = 100
 DEFAULT_CONTENT_EXTRACTION = True
-ENDPOINT_URL = os.getenv(
-    key="ENDPOINT_URL", default=None
-)  # For end to end tests set this environment variable with target host.
+
+if "AWS_ENDPOINT_URL" in os.environ:
+    AWS_ENDPOINT = f"{os.environ['AWS_ENDPOINT_URL']}:{os.environ['AWS_PORT']}"
+else:
+    AWS_ENDPOINT = None
```
artem-shelkovnikov, line 33: What does the `None` case mean?

artem-shelkovnikov: I guess when it's null it uses real AWS?

artem-shelkovnikov (approved):
LGTM! Left a small question

connectors/sources/s3.py
tarekziade: yes. see https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.client

PR #263, opened 2022-12-23
Separate job scheduling from execution [do not merge]

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/3639

This PR separates job scheduling from job execution.

Now the flow to run syncs is:
1. Check connector records that framework can handle
2. If connector needs to sync, create a job and immediately execute it in-place

New flow:

“Producer” - named the same as in Ruby:
1. Check connector records that framework can handle
2. If connector needs to sync, create a job
“Consumer” - named the same as in Ruby:
1. Check connector jobs for jobs that can be picked up by the framework
2. Execute the job 

Short-term goal of the change is to be able to restart suspended jobs, long-term - to be able to separately host these services.

⚠️ This PR does its best to do the separation, but some things were not done due to time constraints:

1. I did not properly and fully test PERF8 integration. I've moved it to only do benchmarks on the job executor side as job executor is the one that's gonna take a lot of resources.
2. Some tests were commented out, new were not added, some were fully removed. We'll need to return back the test coverage with newer changes
3. The consumer service - one that runs syncs - does not do preflight checks, it'll need to do it same way that the producer service does it
4. ??? - Maybe more

While some unit and integration tests are commented out, I've tested the connectors through UI plus I've ran functional tests - they are green, so connector should properly work - maybe except some small bugs might happen due to the size of the change.

#### Pre-Review Checklist


connectors/byoc.py
```diff
@@ -154,63 +153,135 @@ def _create_object(self, doc_source):
...
-            return -1
-        msec = (self.completed_at - self.created_at).microseconds
-        return round(msec / 9, 2)
+    def __init__(self, elastic_config):
+        super().__init__(index_name=JOBS_INDEX, elastic_config=elastic_config)
 
-    async def start(self, trigger_method=JobTriggerMethod.SCHEDULED):
-        self.status = JobStatus.IN_PROGRESS
+    def build_docs_query(self, connector_ids):
+        query = {"bool": {"must": [{"terms": {"status": [e2str(JobStatus.PENDING)]}}]}}
```
wangch079, line 166: `pending or suspended`

connectors/byoc.py
```diff
@@ -154,63 +153,135 @@ def _create_object(self, doc_source):
...
-        if self.completed_at is None:
-            return -1
-        msec = (self.completed_at - self.created_at).microseconds
-        return round(msec / 9, 2)
+    def __init__(self, elastic_config):
+        super().__init__(index_name=JOBS_INDEX, elastic_config=elastic_config)
 
-    async def start(self, trigger_method=JobTriggerMethod.SCHEDULED):
-        self.status = JobStatus.IN_PROGRESS
+    def build_docs_query(self, connector_ids):
```
wangch079, line 165: we should make the function name more specific and descriptive.

connectors/byoc.py
```diff
@@ -154,63 +153,135 @@ def _create_object(self, doc_source):
...
+        return self.doc_source["indexed_document_count"]
+
+    @property
+    def duration(self):
+        if self.completed_at is None:
+            return -1
+        msec = (self.completed_at - self.created_at).microseconds
+        return round(msec / 9, 2)
+
+    async def start(self):
```
wangch079, line 257:
```suggestion
    async def claim(self):
```

connectors/byoc.py
```diff
@@ -154,63 +153,135 @@ def _create_object(self, doc_source):
...
+        self.status = JobStatus.IN_PROGRESS
+        self.doc_source["status"] = e2str(self.status)
```
wangch079, line 259:
```suggestion
        self.doc_source["status"] = e2str(self.status)
        self.doc_source["started_at"] = isoutc()
```

connectors/services/consume.py
```diff
@@ -0,0 +1,152 @@
...
+from connectors.byoc import BYOIndex, DataSourceError, PipelineSettings, SyncJobIndex
+from connectors.byoei import ElasticServer
+from connectors.es import DEFAULT_LANGUAGE, defaults_for
+from connectors.logger import logger
+from connectors.services.base import BaseService
+from connectors.source import DataSourceConfiguration, get_source_klass
+from connectors.utils import CancellableSleeps, trace_mem
+
+
+class JobService(BaseService):
```
wangch079, line 26: Do you think it's better to make the class name consistent with the file name?

connectors/services/consume.py
```diff
@@ -0,0 +1,152 @@
...
+        self.jobs = SyncJobIndex(elastic_config)
+        self.connectors = BYOIndex(elastic_config)
+
+        self.language_code = self.config.get(
+            "language_code", DEFAULT_LANGUAGE
+        )  # XXX: get language code from the connector instead
+        self.bulk_options = elastic_config.get("bulk", {})
+
+    async def run(self):
+        one_sync = self.args.one_sync
```
wangch079, line 45: I wonder if it is still useful to maintain the `one_sync` as of the args, with the decoupling of producer and consumer.

We want to have feature parity between ruby and python but we don't have this functionality in ruby. 

Another removable candidate is the `sync_now` flag. I do think it will be much simple and more readable if we make the app just do one thing

connectors/services/consume.py
```diff
@@ -0,0 +1,152 @@
...
+                if one_sync and synced_anything:
+                    logger.info("Ran a round of syncs and exiting.")
+                    break
+
+                await self._sleeps.sleep(10)
+        finally:
+            self.stop()
+        return 0
+
+    async def execute(self, job):
```
wangch079, line 71: Should claim/start the job before execution.

connectors/services/sync.py
```diff
@@ -162,26 +115,80 @@ async def _run(self):
...
+            connectors_ids = [self.config.get("connector_id")]
+        else:
+            connectors_ids = []
+
+        logger.debug(f"Native support for {', '.join(native_service_types)}")
+
+        query = self.connectors.build_docs_query(native_service_types, connectors_ids)
+
+        async for connector in self.connectors.get_all_docs(query=query):
+            await connector.heartbeat()
```
wangch079, line 147: It seems the `heartbeat` simply updates `last_seen` without checking when it is last seen.

connectors/byoc.py
```diff
@@ -154,63 +153,135 @@ def _create_object(self, doc_source):
...
 
-class SyncJob:
-    def __init__(self, connector_id, elastic_client):
-        self.connector_id = connector_id
-        self.client = elastic_client
-        self.created_at = datetime.now(timezone.utc)
-        self.completed_at = None
-        self.job_id = None
-        self.status = None
+class SyncJobIndex(ESIndex):
```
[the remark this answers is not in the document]
artem-shelkovnikov: It's probably something that can be done in a separate PR - I don't really like the inheritance here and don't see why we cannot pass ESClient or as you mentioned a layer on top of it into SyncJobIndex/ConnectorsIndex initializer, for example.

I renamed BYOConnector -> Connector and BYOIndex -> ConnectorsIndex in this PR for now.

connectors/byoc.py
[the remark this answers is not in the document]
artem-shelkovnikov: Good call, fixed it as well.

connectors/byoc.py
artem-shelkovnikov: Good point, fixed

connectors/byoc.py
artem-shelkovnikov: I agree with it, however I'm following @vidok's recommendation here to inherit from ESIndex and implement the methods - I would even probably move it to a different class - something like QueryBuilder that is usable like:

```
query = JobsIndex.QueryBuilder()
    .with_native_service_types(native_service_types)
    .build()

## OR

query = JobsIndex.QueryBuilder()
    .with_native_service_types(native_service_types)
    .with_connector_ids(connector_ids)
    .build()
```

For now I'd say it's out of scope of PR and will be addressed later.

connectors/byoc.py
artem-shelkovnikov: Great catch, totally missed it!

connectors/services/consume.py
artem-shelkovnikov: For now I had to keep it the way it is - it's used within our CI to run functional tests.

You can for example run `NAME=mysql make ftest` and this command will attempt to run a one-shot sync. I don't like how it's done now - ideally I'd expose options to schedule a sync and to run a one-shot connector job against a target job as a separate CLI and then use this in our CI.

OR another option is in CI to start a server and push data into Kibana that will make the service run the sync, but it'll require a large rewrite of this as well.

connectors/services/consume.py
artem-shelkovnikov: True, added.

I found that current clients do not support optimistic concurrency and can cause race conditions - something we need to address too (can be done later).

connectors/services/sync.py
artem-shelkovnikov: I think it's okay?

Protocol says that "last_seen" should be updated at least every 30 minutes, but it's "at least" - it does not mean that it has to be updated exactly every 30 minutes, so technically we can update it every second, minute, 10 minutes, etc.

connectors/byoc.py
```diff
@@ -148,69 +146,192 @@ def build_docs_query(self, native_service_types=None, connectors_ids=None):
...
+                {
+                    "bool": {
+                        "filter": [
+                            {"terms": {"connector.id": connector_ids}},
+                        ]
+                    }
+                }
+            )
+
+        query = {"bool": {"should": subqueries}}
```
wangch079, line 216: This won't work as expected. For a `should` clause, it will match if any of the clause matches.

wangch079 (approved):

connectors/cli.py
```diff
@@ -99,15 +100,30 @@ def run(args):
...
+            service.shutdown(sig)
+            jobs_service.shutdown(sig)
+
+        loop.add_signal_handler(sig, functools.partial(shutdown_services, sig))
 
     try:
-        return loop.run_until_complete(coro)
+        codes = loop.run_until_complete(asyncio.gather(producer_coro, consumer_coro))
+
+        non_zero_codes = [e for i, e in enumerate(codes) if e != 0]
```
tarekziade, line 121: you can iterate to read teh code without `enumerate` since you don't use `i`

connectors/es/index.py
```diff
@@ -40,6 +40,18 @@ def _create_object(self, doc):
...
         raise NotImplementedError
 
+    async def fetch_by_id(self, id):
+        await self.client.indices.refresh(index=self.index_name)
+
+        try:
+            resp = await self.client.get(index=self.index_name, id=id)
+        except ApiError as e:
+            logger.critical(f"The server returned {e.status_code}")
+            logger.critical(e.body, exc_info=True)
```
tarekziade, line 50: this does not raise any error, and returns None

we need to raise since the user of this method make the expectation a connector is returned

connectors/es/index.py
```diff
@@ -50,10 +62,10 @@ async def get_all_docs(self, query=None, page_size=DEFAULT_PAGE_SIZE):
         Returns:
             Iterator
         """
-        await self.client.indices.refresh(index=self.index_name)
-
         if query is None:
-            query = {"match_all": {}}
+            return
```
tarekziade, line 70: if query has to be provided, let's make it a mandatory argument

tarekziade (changes requested):
This can't be merged at this point; it breaks some of the connector features and removes a lot of test coverage (20%) -- I will try my best to make the patch work today, but it might be better to cherry-pick changes introduced here in smaller patches.

tarekziade:
I have given it a shot, but this is getting too complex to do right in a single patch, and requires proper test coverage - let's not remove tests and lower the QA of `main` to push the PR at all costs.

There are a bunch of great changes here, that can be split into their own PR before the scheduling service is introduced:

- small refactorings and renamings
- allowing multiple services to start
- changing the heartbeat logic for a connector
- etc.

connectors/es/index.py
wangch079: Even it's mandatory, it's still possible to be `None`.

tarekziade: Yes, or -1 or any bad value because we don’t assert we get a query. But in practice if the argument is mandatory the expectation is to get a query. Which is what we do here because if we don’t get one we return and fail in practice. Query validation is another topic.

wangch079:
Hi @tarekziade , trying to break this into multiple PRs. 

Do you think it's necessary to have a `ConnectorIndex` class and a `SyncJobIndex` class to do the job of [Core:: ElasticConnectorActions](https://github.com/elastic/connectors-ruby/blob/main/lib/core/elastic_connector_actions.rb)? Or we can have a single class?

Do you think we need to make `ConnectorIndex` class and a `SyncJobIndex` class the derived classes of `EsIndex`, and then `EsClient`? Composition over inheritance?

Is it a Python convention to have multiple classes in one file?

artem-shelkovnikov:
Closing as it won't be merged any more - broken down into smaller pieces

PR #264, opened 2022-12-23
Fixes the GCS e2e nightly test

tarekziade:
add STORAGE_EMULATOR_HOST

### DIFF
--- a/connectors/sources/gcs.py
+++ b/connectors/sources/gcs.py
@@ -44,9 +44,7 @@ DEFAULT_RETRY_COUNT = 3
 DEFAULT_WAIT_MULTIPLIER = 2
 DEFAULT_CONTENT_EXTRACTION = True
 DEFAULT_FILE_SIZE_LIMIT = 10485760
-STORAGE_EMULATOR_HOST = os.getenv(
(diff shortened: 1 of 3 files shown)


connectors/sources/gcs.py
```diff
@@ -44,9 +44,7 @@
 DEFAULT_WAIT_MULTIPLIER = 2
 DEFAULT_CONTENT_EXTRACTION = True
 DEFAULT_FILE_SIZE_LIMIT = 10485760
-STORAGE_EMULATOR_HOST = os.getenv(
-    key="STORAGE_EMULATOR_HOST", default=None
-)  # For end to end tests set this environment variable with target host.
+STORAGE_EMULATOR_HOST = os.environ.get("STORAGE_EMULATOR_HOST")
```
markjhoy, line 47: I assume this will return `None` if the env variable is not found... but just want to make sure (as my python is a bit rusty ;) )

markjhoy (approved):
👍

connectors/sources/gcs.py
tarekziade: yes this is correct.

```
>>> import os
>>> print(os.environ.get("I DONT EXIST"))
None
```

PR #265, opened 2022-12-27
[Filtering] Add BasicRule parse method

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3573 ##

This PR adds a parse method to the `BasicRule` class to parse an array of dicts representing raw basic rules in the following way:

  - Map every raw basic rule in the json array to the corresponding BasicRule object
  - Filter out every basic rule, which matches the default rule
  - Sort the result in descending order according to their basic rule order (rules are executed in descending order)

Also add `from_string` factory methods for the `Policy`, `Rule` and a `from_json` factory method for `BasicRule`.

#### Pre-Review Checklist

### DIFF
--- a/connectors/tests/test_basic_rule.py
+++ b/connectors/tests/test_basic_rule.py
@@ -6,7 +6,73 @@
 
 import datetime
 
-from connectors.filtering.basic_rule import BasicRule, Policy, Rule, try_coerce
+import pytest
+
+from connectors.filtering.basic_rule import BasicRule, Policy, Rule, parse, try_coerce
+
+BASIC_RULE_ONE_ID = "1"
+BASIC_RULE_ONE_ORDER = 1
+BASIC_RULE_ONE_POLICY = "include"
+BASIC_RULE_ONE_FIELD = "field one"
+BASIC_RULE_ONE_RULE = "equals"
+BASIC_RULE_ONE_VALUE = "value one"
+
+BASIC_RULE_ONE_JSON = {
+    "id": BASIC_RULE_ONE_ID,
+    "order": BASIC_RULE_ONE_ORDER,
+    "policy": BASIC_RULE_ONE_POLICY,
+    "field": BASIC_RULE_ONE_FIELD,
+    "rule": BASIC_RULE_ONE_RULE,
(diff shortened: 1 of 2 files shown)


connectors/tests/test_basic_rule.py
```diff
@@ -33,6 +99,219 @@
...
+        basic_rule.id_ == BASIC_RULE_DEFAULT_ID
+        and basic_rule.order == BASIC_RULE_DEFAULT_ORDER
+        and basic_rule.policy == Policy.from_string(BASIC_RULE_DEFAULT_POLICY)
+        and basic_rule.field == BASIC_RULE_DEFAULT_FIELD
+        and basic_rule.rule == Rule.from_string(BASIC_RULE_DEFAULT_RULE)
+        and basic_rule.value == BASIC_RULE_DEFAULT_VALUE
+    )
+
+
+def test_include_lowercase_should_return_policy_include():
```
artem-shelkovnikov, line 171: I think this and below tests can be actually simplified - not sure what's the best way for that in Python, I found this as an example: https://docs.pytest.org/en/6.2.x/parametrize.html

So test can be parametrised, and you can pass your test cases as input and rewrite the test to something (pseudocode) like:

```python

test_case("include", Policy.INCLUDE)
test_case("INCLUDE", Policy.INCLUDE)
### etc. I'm not sure about actually correct syntax of this in Python
def test_policy_from_string_parsing_strings_correctly(parsed_string, expected_value)
    assert Policy.from_string(parsed_string) == expected_value
```

artem-shelkovnikov (approved):
LGTM. Left a comment about optimising the test suite a bit.

connectors/tests/test_basic_rule.py
timgrein: Good point, that should probably work. I also thought about this, but was a bit worried that this could introduce flaky tests, if the parameter name matches a parameter from a test fixture (which I think gets injected automatically?). But from the docs it's probably not a problem as you're explicit about the parameter names in the `parametrize` decorator. I'll try out, let's see, if it works :)

timgrein: And it works, thank you for the hint!

PR #267, opened 2022-12-28
[Filtering] Add BasicRuleEngine to execute a list of rules against a document

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3575 ##

This PR introduces the BasicRuleEngine which executes a list of rules in the order they're passed to the BasicRuleEngine. It also records which rule included or excluded how many documents (will be used for logging at the end of a sync job).

#### Pre-Review Checklist

### DIFF
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -3,6 +3,7 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+
 import datetime
 import re
 from enum import Enum
@@ -13,6 +14,38 @@ from connectors.logger import logger
 
 IS_BOOL_FALSE = re.compile("^(false|f|no|n|off)$", re.I)
 IS_BOOL_TRUE = re.compile("^(true|t|yes|y|on)$", re.I)
+
+
+def parse(basic_rules_json):
+    """
+
+    *basic_rules_json*, an array of dicts or an empty array
+
+    The parser works in the following way:
+      - Map every raw basic rule in the json array to the corresponding BasicRule object
+      - Filter out every basic rule, which returns true for is_default_rule()
+      - Sort the result in descending order according to their basic rule order (rules are executed in descending order)
+    """
+    if not basic_rules_json:
+        return []
+
+    map_to_basic_rules_list = list(
+        map(
+            lambda basic_rule_json: BasicRule.from_json(basic_rule_json),
+            basic_rules_json,
+        )
+    )
+
+    return sorted(
+        list(
+            filter(
+                lambda basic_rule: not basic_rule.is_default_rule(),
+                map_to_basic_rules_list,
+            )
+        ),
+        key=lambda basic_rule: basic_rule.order,
+        reverse=True,
+    )
 
 
 def to_float(value):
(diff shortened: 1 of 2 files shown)


connectors/filtering/basic_rule.py
```diff
@@ -94,6 +94,52 @@ def try_coerce(value):
...
+            ].increment_matched_doc_count()
+            return True
+
+        for rule in self.rules:
+            if not rule:
+                continue
+
+            if rule.matches(document):
+                rule_match_stats = self.rules_match_stats.setdefault(
+                    rule.id_, RuleMatchStats(rule.policy, 0)
```
wangch079, line 132: curious why the `id_` ends with `_`

connectors/filtering/basic_rule.py
```diff
@@ -94,6 +94,52 @@ def try_coerce(value):
     return coerced_value
 
 
+class RuleMatchStats:
+    def __init__(self, policy, matched_docs_count):
```
wangch079, line 98:
```suggestion
    def __init__(self, policy, matched_docs_count=0):
```

Probably could have a default init count

connectors/filtering/basic_rule.py
```diff
@@ -94,6 +94,52 @@ def try_coerce(value):
...
+            if rule.matches(document):
+                rule_match_stats = self.rules_match_stats.setdefault(
+                    rule.id_, RuleMatchStats(rule.policy, 0)
+                )
+                rule_match_stats.increment_matched_doc_count()
+
+                return rule.is_include()
+
+        # default behavior: ingest document, if no rule matches ("default rule")
+        self.rules_match_stats[BasicRule.DEFAULT_RULE_ID].increment_matched_doc_count()
```
wangch079, line 139: What is the purpose of `DEFAULT_RULE_ID`?

connectors/filtering/basic_rule.py
timgrein: `id` is a builtin function in module `__builtin__`. PyCharm detected that and suggested to add `_` as postfix as `_` as a prefix would indicate a private attribute (seems to be a best practice).

connectors/filtering/basic_rule.py
timgrein: To identify the default rule which matches all documents (and get's always applied if no other rule matches)

connectors/filtering/basic_rule.py
tarekziade: yes, this avoid shadowing. good practice.

connectors/filtering/basic_rule.py
```diff
@@ -94,6 +94,52 @@ def try_coerce(value):
     return coerced_value
 
 
+class RuleMatchStats:
+    def __init__(self, policy, matched_docs_count=0):
+        self.policy = policy
+        self.matched_docs_count = matched_docs_count
+
+    def increment_matched_doc_count(self, by=1):
```
tarekziade, line 102: This is a very very long name for a method in Python. what about `incr_match` or `increment_match` ?  The general guideline is to try to keep names as small as possible. If you need a very long name to make it clear, it's usually a sign there's a design issue.

You can also look at `__add__`, if you implement it for `int` you could do:

```
rule_match = RuleMatchStats(policy)
rule_match += 1
```
but that could be a stretch

connectors/filtering/basic_rule.py
```diff
@@ -94,6 +94,52 @@ def try_coerce(value):
     return coerced_value
 
 
+class RuleMatchStats:
+    def __init__(self, policy, matched_docs_count=0):
```
tarekziade, line 98: You probably don't need `docs` in this variable name. We know we're dealing with docs.

could be `matches_count`

tarekziade (commented):
Looks good. A general remark on naming in this patch: names are quite long, which adds some cognitive load that is unnecessary, and will make the code lines longer.

Naming is hard and quite language-specific. Java, for instance, grows class names with the inheritance tree, which leads to very long but exact class names. Same for method names. This is due to static typing and a specific coding philosophy.

In Python, names are as small as possible, and we try to avoid any concept duplication because its syntax tries to get closer to natural language, you can read it out loud and it should make sense. For instance, if a class has `rule` in its name or always deals with documents, its methods will usually not duplicate the word in their names. A good way to make sure names are good is in tests by asking yourself:

- can I read the line out loud, and is it close to a natural language sentence?
- does it fit in 80 columns for simple calls ? (we can go over it, but most of the time there's no need)

So for example in this test:

```
def test_rule_match_stats_increment(increments, expected_count):
    rule_match_stats = RuleMatchStats(Policy.INCLUDE, 0)

    for increment in increments:
        if increment:
            rule_match_stats.increment_matched_doc_count(by=increment)
        else:
            rule_match_stats.increment_matched_doc_count()

    assert rule_match_stats.matched_docs_count == expected_count
```

I would have written something like:

```
@pytest.mark.parametrize(
    "increments, expected_count",
    [([1, 2, 3], 6), ([1, 1, 1], 3), ([2, None], 3)],
)
def test_rule_match_stats_increment(increments, expected_count):
    stats_matcher = RuleMatchStats(Policy.INCLUDE, 0)

    for increment in increments:
            stats_matcher.incr(by=increment)

    assert stats_matcher.matches == expected_count
```

This is not a blocker, just some feedback from common practices

connectors/filtering/basic_rule.py
timgrein: Good point, implemented it 👍

timgrein: Also shortened some names, thanks for the review!

tarekziade (approved):
LGTM thanks a lot for the changes

PR #269, opened 2022-12-29
[Filtering] Add BasicRuleNoMatchAllRegexValidator

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/3583 ##

This PR adds the `BasicRuleNoMatchAllRegexValidator`. The default rule is excluded from this validation as it intentionally uses a match all regex.

#### Pre-Review Checklist

### DIFF
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -147,6 +147,50 @@ class Policy(Enum):
                 )
 
 
+class BasicRuleValidationResult:
+    def __init__(self, rule_id, is_valid, validation_message):
+        self.rule_id = rule_id
+        self.is_valid = is_valid
+        self.validation_message = validation_message
+
+    @classmethod
+    def valid_result(cls, rule_id):
+        return BasicRuleValidationResult(
+            rule_id=rule_id, is_valid=True, validation_message="Valid rule"
+        )
+
+
+class BasicRuleValidator:
+    @classmethod
+    def validate(cls, rule):
+        raise NotImplementedError
+
+
+class BasicRuleNoMatchAllRegexValidator(BasicRuleValidator):
+    MATCH_ALL_REGEXPS = [".*", "(.*)"]
+
+    @classmethod
+    def validate(cls, basic_rule_json):
+        basic_rule = BasicRule.from_json(basic_rule_json)
+
+        # default rule uses match all regex, which is intended
+        if basic_rule.is_default_rule():
+            return BasicRuleValidationResult.valid_result(rule_id=basic_rule.id_)
+
+        if basic_rule.rule == Rule.REGEX and any(
+            match_all_regex == basic_rule.value
+            for match_all_regex in BasicRuleNoMatchAllRegexValidator.MATCH_ALL_REGEXPS
+        ):
+            return BasicRuleValidationResult(
+                rule_id=basic_rule.id_,
+                is_valid=False,
+                validation_message=f"Match all regexps: '{BasicRuleNoMatchAllRegexValidator.MATCH_ALL_REGEXPS}' are "
+                f"not allowed.",
+            )
+
+        return BasicRuleValidationResult.valid_result(rule_id=basic_rule.id_)
+
+
 class BasicRule:
     DEFAULT_RULE_ID = "DEFAULT"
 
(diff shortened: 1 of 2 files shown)


connectors/filtering/basic_rule.py
```diff
@@ -147,6 +147,50 @@ def from_string(cls, string):
...
+        self.validation_message = validation_message
+
+    @classmethod
+    def valid_result(cls, rule_id):
+        return BasicRuleValidationResult(
+            rule_id=rule_id, is_valid=True, validation_message="Valid rule"
+        )
+
+
+class BasicRuleValidator:
```
timgrein, line 163: No need to review (temporary copy from https://github.com/elastic/connectors-python/pull/268)

connectors/filtering/basic_rule.py
```diff
@@ -147,6 +147,50 @@ def from_string(cls, string):
                 )
 
 
+class BasicRuleValidationResult:
```
timgrein, line 150: No need to review (temporary copy from https://github.com/elastic/connectors-python/pull/268)

connectors/tests/test_basic_rule.py
```diff
@@ -1187,3 +1194,52 @@ def test_is_not_include_for_exclude_policy():
     )
 
     assert not basic_rule.is_include()
+
+
+def basic_rule_json(merge_with=None, delete_keys=None):
```
timgrein, line 1199: No need to review (temporary copy from https://github.com/elastic/connectors-python/pull/268 until it's merged)

connectors/filtering/basic_rule.py
```diff
@@ -147,6 +147,50 @@ def from_string(cls, string):
...
+        )
+
+
+class BasicRuleValidator:
+    @classmethod
+    def validate(cls, rule):
+        raise NotImplementedError
+
+
+class BasicRuleNoMatchAllRegexValidator(BasicRuleValidator):
```
timgrein, line 169: Should be reviewed

connectors/tests/test_basic_rule.py
```diff
@@ -1187,3 +1194,52 @@ def test_is_not_include_for_exclude_policy():
...
+        (basic_rule_json(merge_with={"rule": "regex", "value": "abc"}), True),
+        # for other rule types it should be possible to use values which look like match all regexps
+        (basic_rule_json(merge_with={"rule": "contains", "value": ".*"}), True),
+        (basic_rule_json(merge_with={"rule": "contains", "value": "(.*)"}), True),
+        # invalid rules
+        (basic_rule_json(merge_with={"rule": "regex", "value": ".*"}), False),
+        (basic_rule_json(merge_with={"rule": "regex", "value": "(.*)"}), False),
+    ],
+)
+def test_basic_rule_validate_no_match_all_regex(basic_rule, should_be_valid):
```
timgrein, line 1241: Should be reviewed

connectors/filtering/basic_rule.py
```diff
@@ -147,6 +147,50 @@ def from_string(cls, string):
...
+    MATCH_ALL_REGEXPS = [".*", "(.*)"]
+
+    @classmethod
+    def validate(cls, basic_rule_json):
+        basic_rule = BasicRule.from_json(basic_rule_json)
```
wangch079, line 174: Should the `basic_rule_json` be validated before calling `.from_json`?

timgrein: No, as we'll always execute the JSON against the schema validation first. Therefore we can reuse functionality like `is_default_rule` and compare against enums without having to implement these again in this method (which would duplicate functionality we already have in `BasicRule`)

timgrein: No as `BasicRuleNoMatchAllRegexValidator` doesn't use a schema to validate against. The class performs a validation checking dependencies between fields dependant one the value of one field (this is not possible with fastjsonschema, you can only specify dependencies in the sense of: field X can only exist if field Y exists, without looking at the value). Therefore this validation should be separated from the schema validation.

connectors/filtering/basic_rule.py
```diff
@@ -147,6 +147,50 @@ def from_string(cls, string):
...
+
+        if basic_rule.rule == Rule.REGEX and any(
+            match_all_regex == basic_rule.value
+            for match_all_regex in BasicRuleNoMatchAllRegexValidator.MATCH_ALL_REGEXPS
+        ):
+            return BasicRuleValidationResult(
+                rule_id=basic_rule.id_,
+                is_valid=False,
+                validation_message=f"Match all regexps: '{BasicRuleNoMatchAllRegexValidator.MATCH_ALL_REGEXPS}' are "
+                f"not allowed.",
```
tarekziade, line 188: I am surprised the linter did not complain (`f` not required)

tarekziade (approved):
LGTM

connectors/filtering/basic_rule.py
timgrein: I think the autoformatter split it into these two lines and appended the `f`. Looks also strange to me tbh.

Issue #256, opened 2022-12-20
Implement concurrent GCS reads


tarekziade:
right now the google connector reads all documents in a sequence:

https://github.com/elastic/connectors-python/blob/main/connectors/sources/gcs.py#L345

we can push concurrent `self.get_blob_document` calls using a `ConcurrentTasks` instance https://github.com/elastic/connectors-python/blob/main/connectors/utils.py#L342

jignesh-crest:
Sure, I'll check and get back to you.

jignesh-crest:
Hey, I tried executing connector following your suggestion to include a concurrent task for the get_blob_document, and here are my findings, The caller method(prepare_docs) from byoc:419 expect the return value as a single dictionary, hence get_doc must yield a single dictionary, if I execute get_blob_document in a concurrent task, I would have to pull out and yield the resulted documents one by one, the throughput of this method would come as same as the current approach. The bottleneck is that the caller method(prepare_docs) expects a single dictionary as a return value.

Modified get_docs:
![image](https://user-images.githubusercontent.com/95845553/209102664-55cd6979-3222-4aaa-bbaa-7835936a820d.png)

Modified get_blob_document:
![image](https://user-images.githubusercontent.com/95845553/209102779-c6225df6-bca2-4d72-9e43-1ffb5a0f42d3.png)

Let us know if anything needs to be changed.

We feel that applying concurrency over get_blob_document will not make much difference as it has no I/O bound and very less CPU bound operations( just the doc mapping), the connector will be more faster if we can apply concurrency over the get_content method for content extraction. What do you think?

tarekziade:
I had in mind a different approach. You can think of it a bit differently:

1. get_docs() can create an asyncio.Queue object, and start tasks that will add docs to it. 
2. get_docs() can yield items from that queue
3. the tasks that was created to fill the queue are the one that call the Google API concurrently to get the content, using ConcurrentTasks


pseudo-code:

```python

def get_docs():

    q = asyncio.Queue()
    fetchers = ConcurrentTasks(10)


    async def grab_content(blob_document):
        q.put(blob_document, partial(self.get_content, blob_document))

    async def producer():
        async for blobs in self.fetch_blobs():
            for blob_document in self.get_blob_document(blobs=blobs)
                await fetchers.put(grab_content(blob_document))
        q.put('FINISHED')

    t = asyncio.create_task(producer())
    while True:
        item = await q.get()
        if item == 'FINISHED':
            break
        yield item

[...]

PR #271, opened 2023-01-03
[DO NOT MERGE] Add Concurrency in GCS

jignesh-crest:
This PR addresses changes suggested for concurrency in GCS as we discussed in #256.

While unit testing, a `server disconnected` error was encountered, which is an open [issue ](https://github.com/aio-libs/aiohttp/issues/4549)in the aiohttp. 
We assume this is because multiple open connections are being used at the time, as before the concurrency change we were not facing this issue.

Please let us know your thoughts on this approach.

Also please don't merge this PR directly as we have to go through one round of testing once the approach is finalized.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/gcs.py
+++ b/connectors/sources/gcs.py
@@ -17,7 +17,15 @@ from aiogoogle.auth.creds import ServiceAccountCreds
 
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import TIKA_SUPPORTED_FILETYPES, get_base64_value
+from connectors.utils import (
+    DEFAULT_MAX_CONCURRENCY,
+    DEFAULT_QUEUE_MEM_SIZE,
+    DEFAULT_QUEUE_SIZE,
+    TIKA_SUPPORTED_FILETYPES,
+    ConcurrentTasks,
+    MemQueue,
+    get_base64_value,
+)
 
 CLOUD_STORAGE_READ_ONLY_SCOPE = "https://www.googleapis.com/auth/devstorage.read_only"
 CLOUD_STORAGE_BASE_URL = "https://console.cloud.google.com/storage/browser/_details/"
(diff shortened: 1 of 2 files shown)


connectors/sources/gcs.py
```diff
@@ -327,23 +332,45 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
-                    yield blob_document, partial(self.get_content, blob_document)
+
+        async def grab_content(blob):
+            """Prepare the blob document by mapping the keys and fetching content from Google cloud storage and puts the generated document in the queue.
+
+            Args:
+                blob (dictionary): blob document fetched from Google cloud storage.
+            """
+            blob_document = await self.prepare_blob_document(blob)
+            await self.queue.put((blob_document, None))
```
tarekziade, line 351: just put the document:

```
await self.queue.put(blob_document)
```

connectors/sources/tests/test_gcs.py
```diff
@@ -470,16 +480,10 @@ async def test_get_content_when_file_size_is_large(catch_stdout, patch_logger):
...
-        doit=True,
-    )
-    assert content is None
-
-    content = await mocked_gcs_object.get_content(
+    await mocked_gcs_object.set_content(
         blob=blob_document,
     )
-    assert content is None
+    assert not bool(blob_document.get("_attachment"))
```
tarekziade, line 486: don't use `not bool` on a `None`
maybe 

```
assert '_attachment' not in blob_document
```

connectors/sources/gcs.py
```diff
@@ -17,7 +17,15 @@
 
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import TIKA_SUPPORTED_FILETYPES, get_base64_value
+from connectors.utils import (
+    DEFAULT_MAX_CONCURRENCY,
+    DEFAULT_QUEUE_MEM_SIZE,
```
tarekziade, line 22: do we really want to reuse the bulk queue default params? I think we want a specific set of default values for `gcs` here

tarekziade:
see https://github.com/elastic/connectors-python/issues/256#issuecomment-1370063322

connectors/sources/gcs.py
```diff
@@ -327,23 +332,45 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
+                    [
+                        await self.fetchers.put(partial(grab_content, blob))
+                        for blob in blobs.get("items", [])
+                    ]
+            await self.queue.put("FINISHED")
+
+        producer_task = asyncio.create_task(producer())
+
+        while True:
+            _, document = await self.queue.get()
```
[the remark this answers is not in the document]
jignesh-crest: We have used MemQueue from utils, it stores size of the document in the first index and document in the second index. thus, to discard size and to get the document only we have to unpack in this way.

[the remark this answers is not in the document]
tarekziade: oh right, sorry

PR #273, opened 2023-01-04
S3 tests are not mocking IAM calls

akanshi-crest:
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/tests/test_s3.py
+++ b/connectors/sources/tests/test_s3.py
@@ -3,6 +3,7 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+import os
 from datetime import datetime
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


connectors/sources/tests/test_s3.py
```diff
@@ -13,6 +14,12 @@
 from connectors.sources.tests.support import assert_basics, create_source
 
 
+@pytest.fixture(scope="session", autouse=True)
+def execute_before_all_tests():
+    os.environ["AWS_ACCESS_KEY_ID"] = "access_key"
```
tarekziade, line 19: You need to check first if it's there, so you don't overwrite an existing key:

```
if 'AWS_ACCESS_KEY_ID' in os.environ:
    return
```

tarekziade: you should also clean up on tear down any set keys in `os.environ`.

akanshi-crest: Hey Tarek, We already checked this thing to unset the keys, However it is not required because once the the test cases gets completed, access key and secret key is automatically removed from the os environment.
let me know if you have any doubt will share the video for the same

tarekziade (approved):

PR #274, opened 2023-01-04
[new connector] Azure Blob Storage

parth-crest:
This PR includes:

## Checklists


praveen-elastic:
buildkite test this

connectors/sources/abs.py
```diff
@@ -0,0 +1,219 @@
...
+    @classmethod
+    def get_default_configuration(cls):
+        """Get the default configuration for Azure Blob Storage
+
+        Returns:
+            dictionary: Default configuration
+        """
+        return {
+            "connection_string": {
+                "value": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;",
```
tarekziade, line 55: can we move the key in a constant at the top?

connectors/sources/abs.py
```diff
@@ -0,0 +1,219 @@
...
+            )
+            return
+
+        async with BlobClient.from_connection_string(
+            conn_str=self.connection_string,
+            container_name=blob["container"],
+            blob_name=blob_name,
+            retry_total=self.retry_count,
+        ) as blob_client:
+            data = await blob_client.download_blob()
```
tarekziade, line 167: we can probably do the same work we have done in GCS to add concurrency when downloading the blob data

connectors/sources/abs.py
parth-crest: Yes, we can replace connection string with `AccountName` and `AccountKey` but we required to form an account url with the help of `AccountName`.
Also same like GCS and S3 we need to set a variable in ftest `RUNNING_FTEST` to bifurcate the local account url for e2e and the cloud account url to run the connector. 

So, Should we procced further with `AccountName` & `AccountKey` ?

connectors/sources/abs.py
```diff
@@ -0,0 +1,219 @@
...
+            content = await data.readall()
```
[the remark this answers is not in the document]
parth-crest: Sure, we will accommodate the changes 👍

connectors/sources/abs.py
parth-crest: As discussed for GCS, Do we still need to keep concurrency for `list_blob()` as it is not taking that much time to execute?
Related comment : https://github.com/elastic/connectors-python/issues/256#issuecomment-1373338415

connectors/sources/abs.py
tarekziade: > we required to form an account url with the help of AccountName

This can be when by composing the values from the config when you need to create a connection, right?

> So, Should we procced further with AccountName & AccountKey ?

I think so, what I think could be done is replace `connection_string` with:

- account_name
- account_key
- blob_endpoint
- database_name maybe?

 and infer the value of `DefaultEndPointsProtocol`, so we do:


```
proto = extract_protocol(blob_endpoint)
connection_string = f'DefaultEndpointsProtocol={proto};AccountName={account_name};AccountKey={account_key};BlobEndpoint={blob_endpoint}/{database};'
```

connectors/sources/abs.py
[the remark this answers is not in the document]
tarekziade: Yes because you used `readall()`  and you are loading the whole file. So if it's 100MiB you are going to have a memory peak. The `data` object let you read by chunks, you can stream into a temporary file, that can get base64-ed on disk and then read back for the `_attachement`field. Of course right now we're just pushing back the problem below when we load back the base64 value, but we divide the size by 2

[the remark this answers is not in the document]
tarekziade: e.g. when the execution reaches this line:

```
get_base64_value(content=content)
```

you have content x 2 in the heap right before it returns.

tarekziade:
By the way, can you add it in https://github.com/elastic/connectors-python/blob/main/.buildkite/nightly.yml ?

tarekziade:
we can also add a line in the checklist here https://github.com/elastic/connectors-python/blob/main/docs/CONTRIBUTING.md#initial-contribution about adding the functional test to the nightly

parth-crest:
> By the way, can you add it in https://github.com/elastic/connectors-python/blob/main/.buildkite/nightly.yml ?

Sure, we will update the PR with all the suggested changes.

moxarth-elastic:
buildkite test this

connectors/sources/abs.py
tarekziade: Sounds good to not add it

parth-crest:
We have addressed suggested comments so you can re-review it.

Also, we checked that the PR [#279] gets merged to main. So do you want us to include those concurrency related changes in this PR via `tweak_bulk_options` method or wait for direct accessing the config via config class PR [#288]. 
 
Also, In this PR [#279] you missed to add `concurrent_downloads` in the config file.

tarekziade (approved):

tarekziade:
> We have addressed suggested comments so you can re-review it.
> 
> Also, we checked that the PR [#279] gets merged to main. So do you want us to include those concurrency related changes in this PR via `tweak_bulk_options` or wait for direct accessing the config via `config class` PR [#288].

let's use `tweak_bulk_options` in a follow-up change.

> 
> Also, In this PR [#279] you missed to add `concurrent_downloads` in the config file.

thanks for spotting this !

tarekziade:
Thanks! great work!

PR #276, opened 2023-01-04
added convert_to_b64

tarekziade:
- Adds a base64 helper that uses the system's `base64` encoder (fast and no memory overhead)
- Falls back to pure Python (win32)
- enforces a lower limit for tests coverage

### DIFF
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -230,7 +230,7 @@ def convert_to_b64(source, target=None, overwrite=False):
             os.remove(target)
         os.rename(temp_target, target)
 
-    return not inplace and target or source
+    return source if inplace else target
 
 
 class MemQueue(asyncio.Queue):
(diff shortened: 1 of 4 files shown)


connectors/utils.py
```diff
@@ -178,6 +183,56 @@ def get_base64_value(content):
...
+    # success, let's move the file to the right place
+    if inplace:
+        os.remove(source)
+        os.rename(temp_target, source)
+    else:
+        if os.path.exists(target):
+            os.remove(target)
+        os.rename(temp_target, target)
+
+    return not inplace and target or source
```
timgrein, line 233: This line is a bit hard to wrap my head around, what's actually happening here/returned. Maybe use a ternary operator like `source if inplace else target`. Otherwise LGTM!

timgrein (approved):
Small comment, but otherwise LGTM

connectors/utils.py
tarekziade: good idea, much better, thanks

PR #279, opened 2023-01-05
adds concurency in lazy download

tarekziade:
Adds up to 10 concurrent calls to lazy downloads

#### Pre-Review Checklist


docs/CONFIG.md
```diff
@@ -9,7 +9,8 @@ Configuration lives in [config.yml](../config.yml).
   - `api_key`: The API key for Elasticsearch connection. You can't set `api_key` when basic auth is used.
   - `ssl`: Whether SSL is used for the Elasticsearch connection.
   - `ca_certs`: Path to a CA bundle.
-  - `bulk`: Options for the Bulk API calls behavior
+  - `bulk`: Options for the Bulk API calls behavior - all options can be
+    overriden by each source class
```
wangch079, line 13: It doesn't feel right to override config by code.

If we want to override by source type, shouldn't we have something like this in config?

```
bulk:
    ...
    overrides:
        s3:
            ...
        mysql:
            ...
```

tarekziade: It offers the ability for a user to change the value in the connector configuration in Kibana, and set the default value when a connector is added.

In our source classes, `get_default_configuration` is called when we do the initial setup of a new connector to define the default configuration. If a source needs to set another default for one of these parameters, this will be the place where it's possible..

If it's done in the yaml file, we won't be able for native connectors to tweak that behavior. The use case is when a customer want to reduce or augment the number of concurrent connection to its backend

connectors/source.py
```diff
@@ -122,6 +122,15 @@ async def get_file(doit=True, timestamp=None):
         """
         raise NotImplementedError
 
+    def tweak_bulk_options(self, options):
+        """Receives the bulk options every time a sycn happens, so they can be
```
vidok, line 126:
```suggestion
        """Receives the bulk options every time a sync happens, so they can be
```

vidok (approved):
Looks good based on my Python knowledge. Just one question: if we let customers tune the `concurrent_downloads` should we define a maximum limit? I assume having 1k concurrent downloads can impact the service performance.

tarekziade:
> Looks good based on my Python knowledge. Just one question: if we let customers tune the `concurrent_downloads` should we define a maximum limit? I assume having 1k concurrent downloads can impact the service performance.

yeah, I am not sure how we can do this because I think the value is very service-specific. Maybe the connector's source class can have that limit hard coded?
For instance, the MySQL connector could do a `SHOW max_connections;` call to get that upper limit

but a value like `100` across the board seems like a good idea. how would we reject an upper value though?
Can this be part of the UI somehow (a min/max for an integer) ? cc @sphilipse

PR #280, opened 2023-01-09
[Filtering] Persist filtering in .elastic-connectors-sync-jobs

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3647 ##

With this PR active filtering within `.elastic-connectors` index will be extracted and persisted in `.elastic-connector-sync-jobs` with each new `SyncJob`.

#### Pre-Review Checklist


connectors/byoc.py
```diff
@@ -212,6 +212,38 @@ async def done(self, indexed_count=0, deleted_count=0, exception=None):
...
+        filtering = (
+            {"advanced_snippet": {}, "rules": []}
+            if (filtering is None or len(filtering) == 0)
+            else deepcopy(filtering)
+        )
+
+        # extract value for sync job
+        filtering["advanced_snippet"] = filtering.get("advanced_snippet", {}).get(
+            "value", {}
+        )
```
wangch079, line 227: If I read it correctly, the structure of filtering in `.elastic-connectors` is
```
{
  "advanced_sinppet": {
    "value": "foobar"
  }
}
```

and you want to transform it to

```
{
  "advanced_sinppet": "foobar"
}
```

May I know why? the `connector` field of `.elastic-connectors-sync-jobs` should be just a copy of the fields of `.elastic-connectors`

connectors/byoc.py
```diff
@@ -159,10 +160,11 @@ def _create_object(self, doc_source):
 
 
 class SyncJob:
-    def __init__(self, connector_id, elastic_client):
+    def __init__(self, connector_id, elastic_client, filtering):
```
wangch079, line 163: This will be refactored. the current `start` function will be split to 
1. `create` function, which index a new doc in `.elastic-connectors-sync-jobs`
2. `claim` function, which change job status from `pending` to `in_progress`.

You can proceed without waiting for the refactor. But I think the `filtering` should be a param of `start` function, not `__init__`.

connectors/byoc.py
timgrein: This was stated in such a way in the original implementation doc and it's implemented this way in Ruby (Ruby actually supports both cases; will add that to python, too) Ruby implementation: `'advanced_snippet' => snippet['value'] || snippet`. Wanted to keep both consistent to not break it for a different protocol version. I'll probably create a follow-up issue to use one (and the same) route for ruby and python.

connectors/byoc.py
timgrein: Good to know, will move it `start`, makes more sense 👍

connectors/byoc.py
wangch079: Hah, I see it in ruby

connectors/byoc.py
```diff
@@ -212,6 +217,38 @@ async def done(self, indexed_count=0, deleted_count=0, exception=None):
...
+        return filtering
+
+
+class Filtering:
+    DEFAULT_DOMAIN = "DEFAULT"
+
+    def __init__(self, filtering):
+        self.filtering = filtering if filtering else []
+
+    def get_active_filter(self, domain):
```
wangch079, line 242:
```suggestion
    def get_active_filter(self, domain=DEFAULT_DOMAIN):
```

connectors/byoc.py
timgrein: Done

wangch079 (approved):
LGTM

PR #281, opened 2023-01-09
reimplemented the put function to avoid a race condition

tarekziade:
## Closes https://github.com/elastic/connectors-py/issues/###

First draft at avoiding race condition (from blog post feedback by Quentin and Chenhui)

Quentin:

 it's indeed possible to get a time-of-check to time-of-use race condition. When calling await super.put(), it's possible for another task to enter into _wait_for_room() and it can succeed since `self.current_memsize` will only be updated in `_put()`.

#### Pre-Review Checklist

### DIFF
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -244,6 +244,9 @@ class MemQueue(asyncio.Queue):
         self._current_memsize = 0
         self.refresh_timeout = refresh_timeout
 
+    def qmemsize(self):
+        return self._current_memsize
+
     def _get(self):
         item_size, item = self._queue.popleft()
         self._current_memsize -= item_size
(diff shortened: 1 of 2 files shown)


connectors/utils.py
```diff
@@ -252,29 +255,65 @@ def _put(self, item):
...
+        #    if it's the case, we re-raise it here
+        while self.full(item_size):
+            putter = self._get_loop().create_future()
+            putter_timeout = self._get_loop().create_task(self._putter_timeout(putter))
+            self._putters.append(putter)
+            try:
+                result = await putter
+                if isinstance(result, asyncio.QueueFull):
+                    raise result
+            except:  # NOQA
```
pquentin, line 293: I realize that this is the same code as in asyncio, but I would rather use `except Exception` so that Ctrl-C continues working. (It's unfortunate that asyncio does not care about Ctrl-C.)

connectors/utils.py
```diff
@@ -252,29 +255,65 @@ def _put(self, item):
...
             await asyncio.sleep(self.refresh_interval)
-        return item_size
 
     async def put(self, item):
-        item_size = await self._wait_for_room(item)
-        return await super().put((item_size, item))
+        item_size = get_size(item)
+
+        # This block is taken from the original put() method but with two
+        # changes:
```
pquentin, line 278: The original put() is quite complex, have you considered explaining it quickly how it works and documentation that self._putters() is a deque used for its FIFO behavior?

pquentin (commented):
Assuming that this works, this looks good. It's unfortunate that this relies on implementation details of asyncio.Queue.

connectors/utils.py
tarekziade: yeah, that's quite the code, it's too bad we can extend it in a simpler way. Maybe using a custom class that uses a queue internally would simplify the approach... I'll add a note

tarekziade:
> Assuming that this works, this looks good. It's unfortunate that this relies on implementation details of asyncio.Queue.

Maybe the class can drop the `asyncio.Queue` implementation altogether. (but for a v2)

pquentin (approved):
I would have liked to write a test, maybe hypothesis-based to find the original race condition but did not take the time, sorry. After a superficial look at the code, this looks good. And thanks for the added comment.

tarekziade:
> I would have liked to write a test, maybe hypothesis-based to find the original race condition but did not take the time, sorry. After a superficial look at the code, this looks good. And thanks for the added comment.

I'll try to add it, it's interesting coverage to have

tarekziade:
Ok. "Good news" -- we don't have a race condition.

I wrote this test

```
@pytest.mark.asyncio
async def test_mem_queue_old(patch_logger):
    item = "small stuff"
    queue = MemQueue(maxmemsize=get_size(item) + 1, refresh_interval=.1,
                        refresh_timeout=30)
    max_size = 0

    async def put_one():
        nonlocal max_size
        await queue.put(item)
        qsize = queue.qmemsize()
        await asyncio.sleep(0)
        if max_size < qsize:
            max_size = qsize

    async def remove_one():
        await queue.get()

    tasks = []
    for i in range(1000):
        tasks.append(put_one())
        tasks.append(remove_one())

    random.shuffle(tasks)
    tasks = [asyncio.create_task(t) for t in tasks]

    await asyncio.gather(*tasks)
    assert max_size == get_size(item)
```


which works for the following reason: the `put` function we have will not return the execution to the event loop when 
`self._wait_for_room` returns. It will always call `super().put()` before a new coroutine get a chance to be resumed on the next `self._wait_for_room`:

```
async def put(self, item):
     item_size = await self._wait_for_room(item)
     return await super().put((item_size, item))
```

The test breaks as soon as you return to the event loop between the two calls:

```
async def put(self, item):
     item_size = await self._wait_for_room(item)
     await asyncio.sleep(0)
     return await super().put((item_size, item))
```

Thanks all for the reviews and feedback! :)

tarekziade:
There's one thing that the current implementation does not do well: on a very high load, a pending put can be stocked for a while, and see other pending puts going through.

So the `_wait_for_room` should use the same future+deque approach to have the proper sequence
Will work on this :)

pquentin:
I'm not fully convinced that there's no race condition, because `asyncio.Queue.put()` awaits on a `Future` before calling `MemQueue._put`, so you don't have a guarantee about the task that will be woke up next, right? It could well be that in practice with the current implementation of the asyncio loop and the current implementation details of `asyncio.Queue` that there's no race condition, though.

(It's for this reason that Trio chooses to do slightly more work to have a nicer definition on checkpoints: https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints. It even randomizes the next task that will run to make sure nobody relies on implementation details.)

tarekziade:
> I'm not fully convinced that there's no race condition, because asyncio.Queue.put() awaits on a Future before calling MemQueue._put, so you don't have a guarantee about the task that will be woke up next, right?

that’s what I thought too but when the event loop resumes the sequence is always running _put from the same task. I have added logs at absolutely every step to watch it. Always the case with event 10k tasks. Maybe uvloop or trio would act differently as you say. Super interesting to try.

PR #283, opened 2023-01-10
use convert_to_b64 for system level encoding

jignesh-crest:
## Closes https://github.com/elastic/connectors-py/issues/275
This PR uses on-disk base64 encoding for fetched content, using the method `convert_to_b64` from utils.

#### Pre-Review Checklist


jignesh-crest:
buildkite test this

connectors/sources/gcs.py
```diff
@@ -311,8 +313,12 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
             return
         logger.debug(f"Downloading {blob_name}")
-        content = b""
-        async with TemporaryFile("wb+") as async_buffer:
+        document = {
+            "_id": blob["id"],
+            "_timestamp": blob["_timestamp"],
+        }
+        source_file_name = ""
+        async with NamedTemporaryFile(mode="wb+", delete=False) as async_buffer:
```
tarekziade, line 321: why the `+` ?

connectors/sources/gcs.py
```diff
@@ -325,13 +331,16 @@ async def get_content(self, blob, timestamp=None, doit=None):
                 )
             )
             await async_buffer.seek(0)
```
tarekziade, line 333: why do you need to keep that seek ? I think you can remove it

connectors/sources/gcs.py
```diff
@@ -325,13 +331,16 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
-            content = await async_buffer.read()
+            source_file_name = async_buffer.name
+        await asyncio.to_thread(
```
tarekziade, line 335: can you add a logger.debug to say we are calling convert_to_b64

tarekziade (commented):
Thanks! just a few nits

connectors/sources/gcs.py
jignesh-crest: Sure, thanks.

connectors/sources/gcs.py
jignesh-crest: I forgot to remove it. Thanks for reminding me.

tarekziade:
buildkite test this

tarekziade (approved):

tarekziade:
buildkite test this

PR #285, opened 2023-01-10
Remove --one-sync and --sync-now flags

wangch079:
This PR removes `--one-sync` and `--sync-now` flags.

The CLI `elastic-ingest --action poll` will run the loop infinitely, and it has to be terminated manually. 

To sync now, it has to be manipulated via UI or the `.elastic-connectors` index directly.

#### Pre-Review Checklist


wangch079 (commented):
Hi @tarekziade , do you have any idea why `CHANGELOG.rst` and `README.rst` are added? And why `LICENSE` is changed to `Apache`?

tarekziade:
> Hi @tarekziade , do you have any idea why `CHANGELOG.rst` and `README.rst` are added? And why `LICENSE` is changed to `Apache`?

these files seem to originate from the <https://github.com/aws/base64io-python/ project. Maybe a problem when the `base64io` dependency is installed.

You should not have them in your patch in any case

connectors/tests/test_sync.py
```diff
@@ -463,6 +462,7 @@ def jobs_update(url, **kw):
     )
     await service.run()  # fails
     await service.run()  # works
```
wangch079, line 464: There are two tests (`test_connector_service_poll_clear_error` and `test_connector_settings_change`) calling `service.run()` twice. Is it expected to have the service run multiple times? I expect the 2nd call should fail with some message that the service is already running. cc. @elastic/ingestion-team

tarekziade:
@andrew-hossack has a use case for one shot syncs, and would like to keep it straightforward like what we have now.

What we could do here once you have removed the flags in the code is introduce a `--action sync` option that can be implemented in `cli.py` as a new service, a wrapper of the `SyncService`. It would:

1. toggle the `sync_now` flag for all existing connectors
2. start a coroutine that reads the connectors `last_sync_*` flags and triggers a `SyncService.shutdown` once they are done 
3. start the SyncService service

This can be used in `ftest` too.

wangch079:
> introduce a `--action sync` option that can be implemented in `cli.py` as a new service, a wrapper of the `SyncService`.

Hi @tarekziade , yes we could do this. But this will be just another service and running `ftest` against it doesn't really test the connector service.

artem-shelkovnikov:
I honestly think that `--sync-now` and `--one-sync` should be done a bit differently - instead of having checks for one-shot syncs in the service itself, the flags should change the behaviour of CLI, for example:

```python
if params.get('--one-sync') == True:
  single_sync_service = SingleSyncService(params)
  single_sync_service.run()
else:
  producer_service.start()
  consumer_service.start()
```

This way service code stays clean and predictable IMO, and `SingleSyncService` (name is arbitrary) re-uses same classes that regular services use, just does not attempt to do it in a loop.

wangch079:
Hi @artem-shelkovnikov , I think what you suggested is similar to @tarekziade . 

`single_sync_service` can simply create a job for each ready-to-sync connector, and execute the job directly, and exit.

My concern is that the `ftest` is testing `single_sync_service`, not connector service (i.e. `producer_service` and `consumer_service`), although they are using the same underlying classes.

artem-shelkovnikov:
I think in essence `ftest` tests that actual connectors work - sync data and send it to Elasticsearch. The mentioned `single_sync_service` in this case has to be really lightweight and contain no actual logic - the logic has to be delegated to the other classes (like JobsIndex, Job, Connector, etc).

Then the `single_sync_service` can be unit-tested the same way `producer_service` and `consumer_service` are unit-tested, that can provide some layer of protection against problems.

But to fully actually test the service from start to end, we'll need to have a long-running instance of connectors that works for some time - cause we still need to test things like recurring syncs and such.

Another option here is to modify `ftest` behaviour so that it started a connectors service, then somehow (maybe with CURL) scheduled a sync and then observed the connectors index to see when the sync ends to proceed the work. This way we won't need the `--sync-now` flag at all, but the script itself will become more complex.

connectors/tests/test_sync.py
```diff
@@ -486,7 +440,8 @@ def bulk_call(url, **kw):
         return CallbackResult(status=200, payload={"items": []})
 
     await set_server_responses(mock_responses, FAKE_CONFIG_TS, bulk_call=bulk_call)
-    service = create_service(CONFIG_FILE, sync_now=True, one_sync=True)
+    service = create_service(CONFIG_FILE)
+    asyncio.get_event_loop().call_later(10, service.stop)
```
artem-shelkovnikov, line 444: This test will run for 10 seconds now which will make the tests run much slower, is it possible to make it run fast?

In Ruby I was mocking `sleep` and `loop`, for example - `sleep` did nothing, `loop` ran iteration only once, do you think it can be used here?

connectors/tests/ftest.sh
```diff
@@ -44,17 +44,17 @@ if [[ $PERF8 == "yes" ]]
...
     fi
 else
-    $ELASTIC_INGEST --one-sync --sync-now --debug
+    $ELASTIC_INGEST --debug
 fi
 
 $PYTHON fixture.py --name $NAME --action remove
 
-$ELASTIC_INGEST --one-sync --sync-now --debug
+$ELASTIC_INGEST --debug
```
artem-shelkovnikov, line 57: How will `ftest` finish its run now after the deletion of flags?

connectors/tests/test_sync.py
wangch079: I don't know whether I can mock `while self.running`.

`service.stop` is used to set `self.running` to `false`.

connectors/tests/ftest.sh
wangch079: I'm making an experiment:
1. make the sync job run in the background
2. start a monitoring process and periodically check `last_synced` of the connector
3. if `last_synced` changes or it reaches time out, kill the sync job process.

connectors/tests/test_sync.py
artem-shelkovnikov: cc @tarekziade - how can we make this test fast?

connectors/tests/ftest.sh
artem-shelkovnikov: I see, the monitoring process is going to be in another PR? I don't see the logic here :(

wangch079: It's here: https://github.com/elastic/connectors-python/pull/285/commits/4f405cadea203a556d1ddaff48bc87b2982ea740

cc @tarekziade

connectors/tests/test_sync.py
tarekziade: yeah we can:

```
task = asyncio.create_task(service.run())
task.add_done_callback(service.stop)
await task
```

tarekziade: same pattern but we can switch the flag after 200ms, the sync will already be running, and the loop will exit after it's done, in the next iteration

wangch079: This way, we assume that in 200 ms, the service has done pre-flight check, and has done at least one loop. 200 ms may not be enough once we have decoupled producer and consumer so our service is more async.

Actually, this is no different from
https://github.com/elastic/connectors-python/blob/1faa0098c513f70d9b396bcad0a22eec7d53ee61/connectors/tests/test_sync.py#L449

where we can just reduce from 10 to 0.2

tarekziade: > has done at least one loop
has entered the loop, next check is when you do the loop again even if the first iteration took longer than 200ms

wangch079: Exactly. my point is it's the same as

```python
asyncio.get_event_loop().call_later(0.2, service.stop) 
```

wangch079:
@tarekziade made some comments in this commit https://github.com/elastic/connectors-python/commit/4f405cadea203a556d1ddaff48bc87b2982ea740 and I replied.

Just realized it won't show up in the PR if you comment in a commit.

tarekziade (approved):

wangch079:
@elastic/ingestion-team , do you think we want to have this PR merged in 8.7? It's not immediately needed.

artem-shelkovnikov:
It's a pretty big change so best to merge it as soon as possible, IMO. But if you find it problematic for 8.7, I think it's okay to have it for 8.8

timgrein:
As our functional tests are currently stuck with the new `JobCleanUpService` I would suggest that we merge this change asap.

tarekziade:
it fixes the issue locally for me

PR #286, opened 2023-01-10
Bump version to 8.6.1

vidok:
## Part of https://github.com/elastic/enterprise-search-team/issues/3217
Bump version to 8.6.1.0


connectors/sources/tests/fixtures/mysql/Makefile
```diff
@@ -2,7 +2,7 @@
 
 PYTHON=python3
 
-export VERSION=8.6.0-SNAPSHOT
+export VERSION=8.6.1-SNAPSHOT
```
vidok, line 5: Not sure if I need to change the version here as well

tarekziade: no let's leave it to `0` for now, this is used for pulling elasticsearch, we can use `0` for now

tarekziade (approved):

tarekziade:
buildkite test this

seanstory (approved):

PR #288, opened 2023-01-11
Add Config module

wangch079:
This PR adds a Config class, which should load YAML config once, and accessible across the project.

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -15,9 +15,8 @@ import os
 import signal
 from argparse import ArgumentParser
 
-from envyaml import EnvYAML
-
 from connectors import __version__
+from connectors.config import Config
 from connectors.logger import logger, set_logger
 from connectors.services.sync import SyncService
 from connectors.source import get_data_sources
@@ -90,11 +89,14 @@ def _parser():
 
 def run(args):
     """Runner"""
+
+    # load config
+    Config.load(args.config_file)
+
     # just display the list of connectors
     if args.action == "list":
         logger.info("Registered connectors:")
-        config = EnvYAML(args.config_file)
-        for source in get_data_sources(config):
+        for source in get_data_sources(Config.get()):
             logger.info(f"- {source.__doc__.strip()}")
         logger.info("Bye")
         return 0
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -15,7 +15,6 @@ import os
 import signal
 from argparse import ArgumentParser
 
-import connectors.config
 from connectors import __version__
 from connectors.config import load_config
 from connectors.logger import logger, set_logger
@@ -92,17 +91,17 @@ def run(args):
     """Runner"""
 
     # load config
-    load_config(args.config_file)
+    config = load_config(args.config_file)
 
     # just display the list of connectors
     if args.action == "list":
         logger.info("Registered connectors:")
-        for source in get_data_sources(connectors.config.config):
+        for source in get_data_sources(config):
             logger.info(f"- {source.__doc__.strip()}")
         logger.info("Bye")
         return 0
 
-    service = SyncService(args)
+    service = SyncService(config, args)
     coro = service.run()
     loop = get_event_loop(args.uvloop)
 
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -15,8 +15,9 @@ import os
 import signal
 from argparse import ArgumentParser
 
+import connectors.config
 from connectors import __version__
-from connectors.config import Config
+from connectors.config import load_config
 from connectors.logger import logger, set_logger
 from connectors.services.sync import SyncService
 from connectors.source import get_data_sources
@@ -91,12 +92,12 @@ def run(args):
     """Runner"""
 
     # load config
-    Config.load(args.config_file)
+    load_config(args.config_file)
 
     # just display the list of connectors
     if args.action == "list":
         logger.info("Registered connectors:")
-        for source in get_data_sources(Config.get()):
+        for source in get_data_sources(connectors.config.config):
             logger.info(f"- {source.__doc__.strip()}")
         logger.info("Bye")
         return 0
--- a/connectors/config.py
+++ b/connectors/config.py
@@ -4,29 +4,17 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 
-import copy
-
 from envyaml import EnvYAML
 
+from connectors.logger import logger
 
-class ConfigNotLoadedError(Exception):
-    pass
+config = None
 
 
-class Config:
-    _yaml = None
-
-    @classmethod
-    def load(cls, config_file):
-        cls._yaml = EnvYAML(config_file)
-
-    @classmethod
-    def get(cls, key=None, default=None):
-        if cls._yaml is None:
-            raise ConfigNotLoadedError(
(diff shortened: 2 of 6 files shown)


connectors/tests/test_sync.py
```diff
@@ -326,7 +325,8 @@ async def test_connector_service_poll(
     mock_responses, patch_logger, patch_ping, set_env
 ):
     await set_server_responses(mock_responses)
-    service = create_service(CONFIG)
+    load_config(CONFIG)
```
vidok, line 328: I know `CONFIG` isn't introduced by the current PR however, having both `Config` and `CONFIG` constants confuses me. Would it be possible to rename `CONFIG` to something like `CONFIG_FILE` as it represents a full file path?

connectors/config.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Config:
+    _yaml = None
+
+    @classmethod
+    def load(cls, config_file):
+        cls._yaml = EnvYAML(config_file)
+
+    @classmethod
+    def get(cls):
```
vidok, line 24: Are there other cases where only a specific part of the configuration is needed? Like `Config.get("elasticsearch")` or `Config.get("service")`?

connectors/tests/test_sync.py
wangch079: `CONFIG_FILE` makes sense. Will update.

connectors/config.py
wangch079: I can add this function, but it's not in use for now.

connectors/cli.py
```diff
@@ -90,11 +89,14 @@ def _parser():
 
 def run(args):
     """Runner"""
+
+    # load config
+    Config.load(args.config_file)
```
tarekziade, line 94: let's use a real instance there,

```
config = Config(args.config_file)
```

connectors/cli.py
```diff
@@ -90,11 +89,14 @@ def _parser():
...
+
     # just display the list of connectors
     if args.action == "list":
         logger.info("Registered connectors:")
-        config = EnvYAML(args.config_file)
-        for source in get_data_sources(config):
+        for source in get_data_sources(Config.get()):
```
tarekziade, line 99: I would pass `config` directly

connectors/config.py
```diff
@@ -0,0 +1,32 @@
...
+import copy
+
+from envyaml import EnvYAML
+
+
+class ConfigNotLoadedError(Exception):
+    pass
+
+
+class Config:
```
tarekziade, line 16: If you want to create a singleton, it's best to use the same pattern that logging, see https://github.com/elastic/connectors-python/blob/main/connectors/logger.py

```
_CONFIG = None

def set_config(config_file):
    global _CONFIG
    _CONFIG = Config(config_file)

def get_config():
    return _CONFIG

    
```

connectors/config.py
```diff
@@ -0,0 +1,32 @@
...
+    def load(cls, config_file):
+        cls._yaml = EnvYAML(config_file)
+
+    @classmethod
+    def get(cls, key=None, default=None):
+        if cls._yaml is None:
+            raise ConfigNotLoadedError(
+                "Config is not loaded yet, make sure to call Config.load(config_file) first."
+            )
+        yaml = copy.deepcopy(cls._yaml)
```
tarekziade, line 29: why do you deepcopy? you can return `_yaml.get(key, default)` directly

connectors/cli.py
wangch079: I want to make it a singleton and let the class variable `_yaml` hold the data, so it only loads once and is accessible globally(this doesn't mean it will, e.g. `ESClient` will still take config as a param instead of adding a dependency on `Config` class).

connectors/config.py
wangch079: The `get` function can return the whole config, which is an instance of `EnvYAML`, or anything if a key is provided.

I use deepcopy to prevent tempering of the real condig data held by `Config` class. But I can remove it as anyone can directly access `Config._yaml` even I prefix `yaml` with `_`.

tarekziade: ah I see the problem. It's not ideal in Python to have a function that sends back two different types of objects; this will lead to issues. Sending back copies of the yaml also breaks your goal of having a singleton I think, because you would end up with different config at the same time I suspect.

What is the use case to return the full `_yaml` ? if we have a singleton, why `BaseService` needs to call `Config.get()`? why not `Config` directly?

or even nothing since anyone needing a config can import `connectors.config` and use the singleton directly, right?

I am also now wondering, why do you need a `Config` class ?

is this not enough? 

```
_CONFIG = None 

def set_config(config_file): 
    global _CONFIG 
    _CONFIG = EnvYAML(config_file) 

def get_config():
    if _CONFIG is None:
        raise ConfigNotLoadedError() 
    return _CONFIG
```

I think `EnvYAML` has all the methods we need

wangch079: Yes, this is what I'm doing now. I don't even have the `get_config()` function, you can access the global variable `_CONFIG` directly.

connectors/cli.py
```diff
@@ -90,11 +90,14 @@ def _parser():
 
 def run(args):
     """Runner"""
+
+    # load config
+    load_config(args.config_file)
```
artem-shelkovnikov, line 95: How do you feel about changing it to `config = load_config(args.config_file)` so that it's not globally accessible?

artem-shelkovnikov: Ah, I see @tarekziade left the same comment. I'm not a fan of singleton config - is there a need for it now for certain task to finish?

Globally accessible config often promotes unhealthy coding styles IMO.

wangch079: Hi @artem-shelkovnikov , I implemented `config` the same way as [logger](https://github.com/elastic/connectors-python/blob/e342bf0426b00ac46a5d4a3d4f20c5827ca29769/connectors/logger.py#L16).

We had a global variable `logger` because we want to use the same logger instance to log messages across the whole project. This is the same for `config`, which we don't want to load it multiple times.

What you suggested is to introduce a `load_config` function (basically just a wrapper of `EnvYAML`), and use it every time when we need config.

artem-shelkovnikov: > What you suggested is to introduce a load_config function (basically just a wrapper of EnvYAML), and use it every time when we need config.

No, the suggestion is to keep it the same way it's done now - it's loaded once by the CLI/entry point class before the services start and then the config is passed to the services. The services then pass the config to the other classes/methods that need config.

connectors/config.py
```diff
@@ -0,0 +1,42 @@
...
+
+from envyaml import EnvYAML
+
+from connectors.logger import logger
+
+config = None
+
+
+def load_config(config_file):
+    global config
```
artem-shelkovnikov, line 17: The way it's implemented now, it seems like anyone can use `config` variable if the module is imported. It's similar to the singleton approach in potential harm, but in this case anyone who imports the module can actually modify the content of `config`.

I think it should not be globally accessible.

wangch079: Indeed. I will make it just a function to load config, and make it only called from `cli`, so all the services will take config as a param.

tarekziade (approved):
Great change, it's much better that way -- decoupling it out of the service class is perfect, thanks

PR #292, opened 2023-01-11
Rename BYOConnector to Connector

vidok:
## Part of https://github.com/elastic/enterprise-search-team/issues/3252

This PR will just rename `BYOConnector` to `Connector` which will be used as data entity for connectors documents

#### Pre-Review Checklist


vidok:
buildkite test this

connectors/byoc.py
```diff
@@ -148,7 +148,7 @@ def build_docs_query(self, native_service_types=None, connectors_ids=None):
         return query
 
     def _create_object(self, doc_source):
-        return BYOConnector(
+        return Connector(
```
seanstory, line 151: Let's rename the file while we're at it. `byoc.py` and `test_byoc.py` don't have much meaning to an external audience.

seanstory (approved):

tarekziade:
Looks good, you miss a `make autoformat` call to sort correctly imports

connectors/byoc.py
vidok: Totally agree! The reason I didn't rename is because it contains other classes/constants and I want to finish with them first.

PR #295, opened 2023-01-12
reimplemented the MemQueue class to avoid unordered inserts

tarekziade:
I closed https://github.com/elastic/connectors-python/pull/281 because I figured out that we did not have a race condition.

See https://github.com/elastic/connectors-python/pull/281#issuecomment-1379794960

But then I realized the implementation can lead to unordered inserts in the queue. So this new patch reintroduces the change which does respect the ordering.

### DIFF
--- a/connectors/tests/test_utils.py
+++ b/connectors/tests/test_utils.py
@@ -9,6 +9,7 @@ import binascii
 import contextlib
 import functools
 import os
+import random
 import tempfile
 import time
 
@@ -22,6 +23,7 @@ from connectors.utils import (
     MemQueue,
     convert_to_b64,
     get_base64_value,
+    get_size,
     next_run,
     validate_index_name,
 )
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks!

connectors/tests/test_utils.py
```diff
@@ -51,13 +53,65 @@ def test_invalid_names():
...
+    # vanilla queue
+    queue = asyncio.Queue()
+    start = time.time()
+    for i in range(1000):
+        await queue.put("x" * 100)
+    queue_duration = time.time() - start
+
+    # mem queue should be 30 times slower at the most
+    quotient, _ = divmod(mem_queue_duration, queue_duration)
+    assert quotient < 30
```
artem-shelkovnikov, line 74: Why 30? :)

connectors/tests/test_utils.py
```diff
@@ -51,13 +53,65 @@ def test_invalid_names():
...
+
+
+@pytest.mark.asyncio
+async def test_mem_queue_race(patch_logger):
+    item = "small stuff"
+    queue = MemQueue(
+        maxmemsize=get_size(item) * 2 + 1, refresh_interval=0.1, refresh_timeout=30
+    )
+    max_size = 0
```
artem-shelkovnikov, line 83: I think a comment would help to understand what's going on in the test. I vaguely understand it, but not 100% sure and not 100% sure what is the actual problem.

artem-shelkovnikov (commented):
I tried to review, but it's still pretty hard to understand what's going on. Is it possible to add more tests/comments or refactor to make the logic easier to read?

connectors/tests/test_utils.py
tarekziade: Arbitrary number. It's just set to make sure the class does not get slower over time when we change it

tarekziade:
> I tried to review, but it's still pretty hard to understand what's going on. Is it possible to add more tests/comments or refactor to make the logic easier to read?

I can add comments, but hardly refactor because it's based on the `asyncio.Queue` class logic

tarekziade:
> > I can add comments, but hardly refactor because it's based on the asyncio.Queue class logic
> 
> I see. Maybe pointing out every change to the original implementation can help?

this is what I did with the two comments (1/ 2/) in the cod, we have two changes -- the original method is here https://github.com/python/cpython/blob/3.10/Lib/asyncio/queues.py#L111

artem-shelkovnikov (approved):
LGTM

PR #297, opened 2023-01-12
Extract preflight check from SyncService

wangch079:
This PR extracts preflight check to a separate module.

The preflight check should not be run in any service (in this case, `SyncService`), but should be done (blocking) before any service can be started.

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -18,7 +18,7 @@ from argparse import ArgumentParser
 from connectors import __version__
 from connectors.config import load_config
 from connectors.logger import logger, set_logger
-from connectors.preflight_check import pre_flight_check
+from connectors.preflight_check import PreflightCheck
 from connectors.services.sync import SyncService
 from connectors.source import get_data_sources
 from connectors.utils import get_event_loop
@@ -104,8 +104,15 @@ def run(args):
         return 0
 
     loop = get_event_loop(args.uvloop)
+    preflight_check = PreflightCheck(config)
+    preflight_check_coro = preflight_check.run()
+
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(preflight_check.shutdown, sig))
+
     try:
-        loop.run_until_complete(pre_flight_check(config))
+        if not loop.run_until_complete(preflight_check_coro):
+            return -1
     except asyncio.CancelledError:
         return 0
 
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -92,8 +92,13 @@ async def _start_service(config, args, loop):
     preflight = PreflightCheck(config)
     for sig in (signal.SIGINT, signal.SIGTERM):
         loop.add_signal_handler(sig, functools.partial(preflight.shutdown, sig))
-    if not await preflight.run():
-        return -1
+    try:
+        if not await preflight.run():
+            return -1
+    finally:
+        for sig in (signal.SIGINT, signal.SIGTERM):
+            loop.remove_signal_handler(sig)
+
     service = SyncService(config, args)
     for sig in (signal.SIGINT, signal.SIGTERM):
         loop.add_signal_handler(sig, functools.partial(service.shutdown, sig))
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -88,6 +88,18 @@ def _parser():
     return parser
 
 
+async def _start_service(config, args, loop):
+    preflight = PreflightCheck(config)
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(preflight.shutdown, sig))
+    if not await preflight.run():
+        return -1
+    service = SyncService(config, args)
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(service.shutdown, sig))
+    await service.run()
+
+
 def run(args):
     """Runner"""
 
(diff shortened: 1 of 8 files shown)


connectors/cli.py
```diff
@@ -91,17 +92,22 @@ def _parser():
...
         logger.info("Registered connectors:")
-        config = EnvYAML(args.config_file)
         for source in get_data_sources(config):
             logger.info(f"- {source.__doc__.strip()}")
         logger.info("Bye")
         return 0
 
+    loop = get_event_loop(args.uvloop)
+    try:
+        loop.run_until_complete(pre_flight_check(config))
```
artem-shelkovnikov, line 105: Why not `await pre_flight_check(config)`?

wangch079: `await` only works in `async` function I think

wangch079:
Hi @tarekziade , when I run `make test` locally, this test hanged:
https://github.com/elastic/connectors-python/blob/0511b02ad754f70ab28c3f66564982556c2fb016/connectors/tests/test_cli.py#L23

I used `⌘+.` to end it, and the test showed `PASS` and proceeded with other tests.

May I know what is this test for? It seems the kill didn't work:
https://github.com/elastic/connectors-python/blob/0511b02ad754f70ab28c3f66564982556c2fb016/connectors/tests/test_cli.py#L34-L35

connectors/cli.py
```diff
@@ -91,17 +92,29 @@ def _parser():
...
 
+    loop = get_event_loop(args.uvloop)
+    preflight_check = PreflightCheck(config)
+    preflight_check_coro = preflight_check.run()
+
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(preflight_check.shutdown, sig))
+
+    try:
+        if not loop.run_until_complete(preflight_check_coro):
```
tarekziade, line 111: you cannot call `loop.run_until_complete` twice.

The proper pattern here is something like this:

```
async def start():
    preflight = PreflightCheck(config)
    checked = await preflight.run()
    if not checked:
        raise Something()
    service = SyncService(args)
        loop.add_signal_handler(sig, functools.partial(service.shutdown, sig))
    await service.run()


loop.run_until_complete(start_service())

```




```

Comment (wangch079) [reply connectors/cli.py]: I made the change here: https://github.com/elastic/connectors-python/pull/297/commits/c729a55f5047fb35fbd5fcb2ac6856b77ed6bde1, with an additional signal handle for preflight check:
https://github.com/elastic/connectors-python/blob/c729a55f5047fb35fbd5fcb2ac6856b77ed6bde1/connectors/cli.py#L93-L94

It behaves similarly when testing locally, which can exit gracefully for both preflight check and sync service.

But when running test `test_main_and_kill`, it killed the test process, not the connector service:
```
connectors/tests/test_cli.py::test_main_and_kill make: *** [Makefile:48: test] Terminated
🚨 Error: The command exited with status 2
user command error: exit status 2
```

**Comment by tarekziade** [comment]
> Hi @tarekziade , when I run `make test` locally, this test hanged:
> 
> https://github.com/elastic/connectors-python/blob/0511b02ad754f70ab28c3f66564982556c2fb016/connectors/tests/test_cli.py#L23
> 
> I used `⌘+.` to end it, and the test showed `PASS` and proceeded with other tests.
> 
> May I know what is this test for? It seems the kill didn't work:
> 
> https://github.com/elastic/connectors-python/blob/0511b02ad754f70ab28c3f66564982556c2fb016/connectors/tests/test_cli.py#L34-L35

weird, it works for me. the test makes sure that the service properly handles a `SIGTERM` signal it gets

File: connectors/cli.py [hunk 3]
```diff
@@ -87,6 +88,18 @@ def _parser():
     return parser
 
 
+async def _start_service(config, args, loop):
+    preflight = PreflightCheck(config)
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(preflight.shutdown, sig))
```
tarekziade, line 94: I think your hang issue is here; you are registering several functions against SIGTERM. You should only have one that iterates on all services:

```
def _shutdown_service(sig):
    for service in running_services:
        service.shutdown()
```

connectors/preflight_check.py
```diff
@@ -0,0 +1,71 @@
...
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+
+from connectors.byoc import CONNECTORS_INDEX, JOBS_INDEX
+from connectors.es import ESClient
+from connectors.logger import logger
+from connectors.utils import CancellableSleeps
+
+
+class PreflightCheck:
```
tarekziade, line 13: I don't think it's useful to have the preflight check implemented as a service because it's a single synchronous check you do before you start all work.

I think the pattern we are looking for in `_start_service` is:


```

def _shutdown_service(sig, services):
    for service in services:
        service.shutdown()

async _start_service():
     preflight_check()
     
     # now we can prepare services
     services = [...]

     # and prepare a shutdown hook
        loop.add_signal_handler(sig, functools.partial(_shutdown_service, sig, services))

    runners = [service.run() for service in services]
    await asyncio.gather(*services)
     
```

wangch079: As we discussed today, we want to make it a service because we want preflight check to be able to shutdown gracefully.

connectors/cli.py
```diff
@@ -87,6 +88,18 @@ def _parser():
...
+    if not await preflight.run():
```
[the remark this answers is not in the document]
wangch079: Done here:

https://github.com/elastic/connectors-python/blob/13f56dc139d2c3f1f42a97370e2cbe410c977c7b/connectors/cli.py#L95-L100

Instead of calling `shutdown`, I removed the signal handler, as `stop` will be done in the `finally` block of `PreflightCheck.run`:

https://github.com/elastic/connectors-python/blob/13f56dc139d2c3f1f42a97370e2cbe410c977c7b/connectors/preflight_check.py#L44-L47

connectors/tests/test_cli.py
```diff
@@ -32,6 +32,7 @@ def test_main_and_kill(patch_logger, mock_responses):
     )
 
     async def kill():
+        await asyncio.sleep(0.2)
```
wangch079, line 35: @tarekziade , mock patch `run` with `kill` doesn't really start the service, so I think this may make more sense? Although I don't like it either.

connectors/tests/test_sync.py
```diff
@@ -550,21 +545,6 @@ def connectors_update(url, **kw):
     raise AssertionError
 
 
-@pytest.mark.asyncio
```
wangch079, line 553: This is removed as the ping is tested in `test_preflight_check.py`

tarekziade (approved):

PR #299, opened 2023-01-12
[new connector] Generic Database Connector (PostgreSQL, Oracle)

moxarth-elastic:
This PR includes:

## Checklists

 - [x] SSL Support for Oracle.

## Licenses

- SQLAlchemy: [MIT License](https://github.com/sqlalchemy/sqlalchemy/blob/main/LICENSE)
- asyncpg: [Apache-2.0 License](https://github.com/MagicStack/asyncpg/blob/master/LICENSE)
- cx_Oralce: [BSD License](https://github.com/oracle/python-cx_Oracle/blob/main/LICENSE.txt)


tarekziade:
Awesome will review it now -- would it be possible to try it out against Azure? (I can give you an access to an Azure SQL deployment if you need)

connectors/sources/mssql.py
```diff
@@ -0,0 +1,85 @@
...
+            "ALL_TABLE": "SELECT table_name FROM information_schema.tables WHERE TABLE_SCHEMA = '{schema}'",
+            "TABLE_PRIMARY_KEY": "SELECT C.COLUMN_NAME FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS T JOIN INFORMATION_SCHEMA.CONSTRAINT_COLUMN_USAGE C ON C.CONSTRAINT_NAME=T.CONSTRAINT_NAME WHERE C.TABLE_NAME='{table}' and C.TABLE_SCHEMA='{schema}' and T.CONSTRAINT_TYPE='PRIMARY KEY' ",
+            "TABLE_DATA": 'SELECT * FROM {schema}."{table}"',
+            "TABLE_LAST_UPDATE_TIME": "SELECT last_user_update FROM sys.dm_db_index_usage_stats WHERE object_id=object_id('{schema}.{table}')",
+            "TABLE_DATA_COUNT": 'SELECT COUNT(*) FROM {schema}."{table}"',
+            "ALL_SCHEMAS": "SELECT s.name from sys.schemas s inner join sys.sysusers u on u.uid = s.principal_id",
+        }
+        self.engine = None
+
+    async def ping(self):
```
tarekziade, line 52: It feels like the ping is copy-pasted around, and could be shared in the base class.

connectors/sources/generic_database.py
```diff
@@ -0,0 +1,444 @@
...
+    def __init__(self, connector):
+        """Setup connection to the database-server configured by user
+
+        Args:
+            connector (BYOConnector): Object of the BYOConnector class
+        """
+        super().__init__(connector=connector)
+
+        # Connector configurations
+        self.retry_count = int(
```
tarekziade, line 33: with the latest changes, you can do:

```
self.retry_count = self.configuration['retry_count']
```

The base class is taking care of the default value and the type conversion now, as long as you set the right type in `get_default_configuration`

connectors/sources/mssql.py
```diff
@@ -0,0 +1,85 @@
...
+        """Setup connection to the Microsoft SQL database-server configured by user
+
+        Args:
+            connector (BYOConnector): Object of the BYOConnector class
+        """
+        super().__init__(connector=connector)
+        self.connection_string = f"mssql+pymssql://{self.user}:{quote(self.password)}@{self.host}:{self.port}/{self.database}"
+        self.queries = {
+            "PING": "SELECT 1+1",
+            "ALL_TABLE": "SELECT table_name FROM information_schema.tables WHERE TABLE_SCHEMA = '{schema}'",
```
tarekziade, line 43: Can we move the queries as global consts to make it easier to read this code?

connectors/sources/generic_database.py
```diff
@@ -0,0 +1,444 @@
...
+            is_async (bool, optional): Flag to indicates async/sync execution. Defaults to True.
+
+        Raises:
+            exception: Raise an exception after retrieving
+
+        Yields:
+            list: Column names and query response
+        """
+        query = self.queries[query_name].format(**query_kwargs)
+        size = int(self.configuration.get("fetch_size", DEFAULT_FETCH_SIZE))
```
tarekziade, line 156: same remark (it's now auto converted and defaulted)

connectors/sources/generic_database.py
```diff
@@ -0,0 +1,444 @@
...
+                executor=None, func=partial(self.connection.execute, statement=query)
+            )
+            return cursor
+        except Exception as exception:
+            logger.warning(
+                f"Something went wrong while executing query. Exception: {exception}"
+            )
+            raise
+
+    def serialize(self, doc):
```
tarekziade, line 263: this is being copy-pasted in many sources at this point.

Let's move it to `BaseDataSource` to avoid duplication (and remove it from mongo, MySQL)

connectors/sources/generic_database.py
```diff
@@ -0,0 +1,444 @@
...
+            [[row_count]] = await anext(
+                self.execute_query(
+                    query_name="TABLE_DATA_COUNT",
+                    engine=engine,
+                    is_async=is_async,
+                    schema=schema,
+                    table=table,
+                )
+            )
+            if row_count:
```
tarekziade, line 324: `row_count` is not a boolean and a rather complex structure.. (`[[]]`) can we change this for an explicit test?

connectors/sources/generic_database.py
```diff
@@ -0,0 +1,444 @@
...
+            self.execute_query(
+                query_name="ALL_TABLE",
+                engine=engine,
+                is_async=is_async,
+                user=self.user.upper(),
+                database=self.database,
+                schema=schema,
+            )
+        )
+        if list_of_tables:
```
tarekziade, line 428: explicit test, `list_of_tables` is not a bool

connectors/sources/oracle.py
```diff
@@ -0,0 +1,58 @@
...
+
+        Args:
+            connector (BYOConnector): Object of the BYOConnector class
+        """
+        super().__init__(connector=connector)
+        self.connection_string = f"oracle://{self.user}:{quote(self.password)}@{self.host}:{self.port}/{self.database}"
+        self.queries = {
+            "PING": "SELECT 1+1 FROM DUAL",
+            "ALL_TABLE": "SELECT TABLE_NAME FROM all_tables where OWNER = '{user}'",
+            "TABLE_PRIMARY_KEY": "SELECT cols.column_name FROM all_constraints cons, all_cons_columns cols WHERE cols.table_name = '{table}' AND cons.constraint_type = 'P' AND cons.constraint_name = cols.constraint_name AND cons.owner = cols.owner ORDER BY cols.table_name, cols.position",
```
tarekziade, line 29: same remark (globals)

connectors/sources/postgresql.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Args:
+            connector (BYOConnector): Object of the BYOConnector class
+        """
+        super().__init__(connector=connector)
+        self.connection_string = f"postgresql+asyncpg://{self.user}:{quote(self.password)}@{self.host}:{self.port}/{self.database}"
+        self.queries = {
+            "PING": "SELECT 1+1",
+            "ALL_TABLE": "SELECT table_name FROM information_schema.tables WHERE table_catalog = '{database}' and table_schema = '{schema}'",
+            "TABLE_PRIMARY_KEY": "SELECT c.column_name FROM information_schema.table_constraints tc JOIN information_schema.constraint_column_usage AS ccu USING (constraint_schema, constraint_name) JOIN information_schema.columns AS c ON c.table_schema = tc.constraint_schema AND tc.table_name = c.table_name AND ccu.column_name = c.column_name WHERE constraint_type = 'PRIMARY KEY' and tc.table_name = '{table}' and tc.constraint_schema = '{schema}'",
```
tarekziade, line 33: const

connectors/sources/postgresql.py
```diff
@@ -0,0 +1,81 @@
...
+        except Exception:
+            raise Exception(f"Can't connect to Postgresql Server on {self.host}")
+
+    def get_connect_args(self):
+        """Convert string to pem format and create a SSL context
+
+        Returns:
+            dictionary: Connection arguments
+        """
+        self.ssl_ca = self.ssl_ca.replace(" ", "\n")
```
tarekziade, line 61: Why do we do some specific extraction here? Shouldn't that be done in the generic connector? it looks like you are removing the `--BEGIN` etc? should be a function with an explicit name, like `parse_pem` that returns the key etc

moxarth-elastic:
> Awesome will review it now -- would it be possible to try it out against Azure? (I can give you an access to an Azure SQL deployment if you need)

Sure @tarekziade, we can try it out for Azure SQL. You can slack me the creds for the same.

connectors/sources/mssql.py
moxarth-elastic: Here we need to pass `is_async` flag to `execute_query`. So what do you think can we bifurcate it via a type of engine in the common ping method in the base class?

connectors/sources/postgresql.py
moxarth-elastic: > Why do we do some specific extraction here?

This method is specific to PostgreSQL dialect only, so we should not move it to generic connector file.

> Shouldn't that be done in the generic connector? it looks like you are removing the --BEGIN etc?

Also, we're not removing `---BEGIN--` and `---END---` from the string, we're making it like `pem` format to parse it into `cadata` - https://github.com/elastic/connectors-python/blob/c65bd1ea4b538cbc395f47d654aa2813cf2ec826/connectors/sources/postgresql.py#L65

> should be a function with an explicit name, like parse_pem that returns the key etc

The function takes care of creating a connection argument for engine that's why we named this way.

connectors/sources/generic_database.py
moxarth-elastic: Done

connectors/sources/mssql.py
moxarth-elastic: Done

connectors/sources/generic_database.py
moxarth-elastic: Done

connectors/sources/postgresql.py
tarekziade: can you create a separate 

```
pem_format = self.get_pem_format(self.ssl_ca)
```

with a unit test that passes a SSL cert and returns the pem format.
This will clarify what we are doing here.

thx

requirements/aarch64.txt
```diff
@@ -5,3 +5,7 @@ aiomysql==0.0.22
 aioboto3==9.0.0
 motor==2.5.1
 smbprotocol==1.9.0
+SQLAlchemy==1.4.44
```
tarekziade, line 8: since it's the exact same list than x86_64, let's move them to the framework.txt one

we also need for each new lib to check the license, if you can list them in the issue, so we check that we don't introduce a lib that is not in the list of allowed licenses for Elastic thx

tarekziade (commented):
just one nit or two and we can land
thanks for all the small fixes!
the most important one is licensing

connectors/sources/mssql.py
```diff
@@ -0,0 +1,87 @@
...
+    "TABLE_PRIMARY_KEY": "SELECT C.COLUMN_NAME FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS T JOIN INFORMATION_SCHEMA.CONSTRAINT_COLUMN_USAGE C ON C.CONSTRAINT_NAME=T.CONSTRAINT_NAME WHERE C.TABLE_NAME='{table}' and C.TABLE_SCHEMA='{schema}' and T.CONSTRAINT_TYPE='PRIMARY KEY' ",
+    "TABLE_DATA": 'SELECT * FROM {schema}."{table}"',
+    "TABLE_LAST_UPDATE_TIME": "SELECT last_user_update FROM sys.dm_db_index_usage_stats WHERE object_id=object_id('{schema}.{table}')",
+    "TABLE_DATA_COUNT": 'SELECT COUNT(*) FROM {schema}."{table}"',
+    "ALL_SCHEMAS": "SELECT s.name from sys.schemas s inner join sys.sysusers u on u.uid = s.principal_id",
+}
+
+
+class MSSQLDataSource(GenericBaseDataSource):
+    """Class to fetch documents from Microsoft SQL Server"""
```
tarekziade, line 41: this docstring is used in the CLI, let's tweak the name, see https://github.com/elastic/connectors-python/pull/348

connectors/sources/oracle.py
```diff
@@ -0,0 +1,60 @@
...
+    "ALL_TABLE": "SELECT TABLE_NAME FROM all_tables where OWNER = '{user}'",
+    "TABLE_PRIMARY_KEY": "SELECT cols.column_name FROM all_constraints cons, all_cons_columns cols WHERE cols.table_name = '{table}' AND cons.constraint_type = 'P' AND cons.constraint_name = cols.constraint_name AND cons.owner = cols.owner ORDER BY cols.table_name, cols.position",
+    "TABLE_DATA": "SELECT * FROM {table}",
+    "TABLE_LAST_UPDATE_TIME": "SELECT SCN_TO_TIMESTAMP(MAX(ora_rowscn)) from {table}",
+    "TABLE_DATA_COUNT": "SELECT COUNT(*) FROM {table}",
+}
+
+
+class OracleDataSource(GenericBaseDataSource):
+    """Class to fetch documents from Oracle Server"""
```
tarekziade, line 25: this docstring is used in the CLI, let's tweak the name, see https://github.com/elastic/connectors-python/pull/348

connectors/sources/postgresql.py
```diff
@@ -0,0 +1,83 @@
...
+    "TABLE_PRIMARY_KEY": "SELECT c.column_name FROM information_schema.table_constraints tc JOIN information_schema.constraint_column_usage AS ccu USING (constraint_schema, constraint_name) JOIN information_schema.columns AS c ON c.table_schema = tc.constraint_schema AND tc.table_name = c.table_name AND ccu.column_name = c.column_name WHERE constraint_type = 'PRIMARY KEY' and tc.table_name = '{table}' and tc.constraint_schema = '{schema}'",
+    "TABLE_DATA": 'SELECT * FROM {schema}."{table}"',
+    "TABLE_LAST_UPDATE_TIME": 'SELECT MAX(pg_xact_commit_timestamp(xmin)) FROM {schema}."{table}"',
+    "TABLE_DATA_COUNT": 'SELECT COUNT(*) FROM {schema}."{table}"',
+    "ALL_SCHEMAS": "SELECT schema_name FROM information_schema.schemata",
+}
+
+
+class PostgreSQLDataSource(GenericBaseDataSource):
+    """Class to fetch documents from Postgresql Server"""
```
tarekziade, line 30: this docstring is used in the CLI, let's tweak the name, see https://github.com/elastic/connectors-python/pull/348

connectors/sources/mssql.py
moxarth-elastic: Hey @tarekziade, should we move with your suggestion with the bifurcation of engine type at the common ping method?

connectors/sources/mssql.py
moxarth-elastic: Done

requirements/aarch64.txt
moxarth-elastic: Ok move all libs to framework.txt

connectors/sources/mssql.py
moxarth-elastic: We have accommodated ping in the generic file. Please check and let us know your thoughts.

connectors/sources/postgresql.py
moxarth-elastic: Done

tarekziade (changes requested):
I am sorry we have one more extra step before we can merge, sorry this did not come on my radar before:

- pymssql: [LGPL-2.1](https://github.com/pymssql/pymssql/blob/master/LICENSE)
- psycopg2: [LGPL](https://github.com/psycopg/psycopg2/blob/master/LICENSE)

These are LGPL and we need a specific process to allow heir usage. 

cc @gtback for gudiance

tarekziade:
Also relevant https://github.com/pymssql/pymssql/issues/668

@moxarth-elastic could you try

- https://github.com/mkleehammer/pyodbc (MIT) 
- https://pypi.org/project/asyncpg/ (APLv2)

moxarth-elastic:
Hey @tarekziade,

> https://pypi.org/project/asyncpg/ 

we have removed the package psycopg2 and updated our e2e implementation with the new suggested package.

> Also relevant pymssql/pymssql#668

as per their latest comment https://github.com/pymssql/pymssql/issues/668#issuecomment-1086681292 the decision has been reversed by the author and it will be continued. SQLAlchemy also allows pymssql [Reference](https://github.com/sqlalchemy/sqlalchemy/issues/8337)

Also, we did check PyODBC at the early stage of development [doc](https://docs.google.com/document/d/1Hf8ARyw9kXBqQ25FKQK36AMKEMVfRm5j7ssIPp8zvfc/edit?usp=drivesdk). However, we decided to not go through it for the following reasons:

- The prerequisite to use the connector will be an ODBC driver. The user would have to ensure the required drivers are installed and they pass the path to the driver as a connector configuration. Also, the installation of the driver may vary for the different OS as well as different versions of it.

However, if we decide to still move ahead with pyodbc, then we'll have to make the necessary changes and also retest the connector for functional, regression, as well as performance testing against the Mssql dialect. We anticipate this to take about an additional week approximately to develop, test, and update the PR. Please let us know your thoughts.

CC: @danajuratoni, @nirav-elastic

moxarth-elastic:
> Using LGPL is a blocker from a licensing Pov unfortunately. This is what I suggest
> 
> * let's remove mssql from this patch so we can land it, it has been around for a while
> * let's add mssql as a follow-up patch, and we can dig into the problem

Sure let me remove the mssql dialect so we can merge this PR.
Also, we are string implementation with pyodbc for mssql dialect and we will the follow up PR in the upcoming week.

tarekziade (approved):

tarekziade:
Thanks a lot!

one nit before we can merge: @artem-shelkovnikov added a linter that requires tests to pass under a second, we need to make sure those 2 tests do this:


```
FAILED connectors/sources/tests/test_oracle.py::test_ping_negative
FAILED connectors/sources/tests/test_postgresql.py::test_ping_negative
```

They take >6 seconds. You can look at how we sped up the azure and GCS ones for the pings -- it's just a matter of tweaking the config. I would do it myself but I can't access your branch

moxarth-elastic:
> Thanks a lot!
> 
> one nit before we can merge: @artem-shelkovnikov added a linter that requires tests to pass under a second, we need to make sure those 2 tests do this:
> 
> ```
> FAILED connectors/sources/tests/test_oracle.py::test_ping_negative
> FAILED connectors/sources/tests/test_postgresql.py::test_ping_negative
> ```
> 
> They take >6 seconds. You can look at how we sped up the azure and GCS ones for the pings -- it's just a matter of tweaking the config. I would do it myself but I can't access your branch

Hey @tarekziade,

I have updated the test cases. please take a look.

gtback:
> cc @gtback for gudiance

Sorry I missed the ping here. If you still need anything from me, let me know.

PR #305, opened 2023-01-12
[8.6] Implement the nightly functional test (#258)

tarekziade:
Backports the following commits to 8.6:
 - Implement the nightly functional test (#258)


.buildkite/nightly.yml
```diff
@@ -0,0 +1,26 @@
...
+  - label: "🔨 Amazon S3"
+    command:
+      - ".buildkite/run_nigthly.sh s3"
+    artifact_paths:
+      - "perf8-report-*/**/*"
+  - label: "🔨 Google Cloud Storage"
+    command:
+      - ".buildkite/run_nigthly.sh gcs"
+    artifact_paths:
+      - "perf8-report-*/**/*"
```
seanstory, line 26: Should this be removed if GCS isn't in 8.6?

connectors/sources/tests/fixtures/gcs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+google-auth
+google-cloud-storage
```
seanstory, line 2: I think this should be removed?

PR #310, opened 2023-01-13
Remove language_code and analysis_icu from config

artem-shelkovnikov:
Small refactoring to simplify the code.

Originally we've created connectors with CLI thus we needed to provide index settings and language code on creation of the index.

Now Kibana is responsible for this logic (cc @sphilipse), thus this code does not need to be here any more. Moreover, it causes confusion or potentially wrong behaviour - Kibana creates index with certain language code, but language code in config might be different and cause problems.


connectors/byoei.py
```diff
@@ -372,9 +372,7 @@ async def prepare_index(
             return
 
         logger.debug(f"Creating index {index}")
-        await self.client.indices.create(
-            index=index, settings=settings, mappings=mappings
-        )
+        await self.client.indices.create(index=index, mappings=mappings)
```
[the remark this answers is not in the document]
artem-shelkovnikov: Hm, that is interesting. Looking at the Ruby code, it seems like this code is potentially harmful/can lead to SDH cases.

Some thoughts:
1. This code should never run under normal conditions.
2. This code can run only if somebody/something deleted the content index while syncs are not running.
3. If this code runs, it can create the index with wrong settings UNLESS language code in config file is equal to language code that customer used to create he index via Kibana AND analysers were not modified.

Do you know by chance why this code exists in Ruby? Originally the code to create the index was added to support CLI, but now CLI is not used/supported at all.

tarekziade (commented):
Let's move the code in `kibana.py` so it can create the indices during the e2e test until we find a better way to do it

connectors/byoei.py
```diff
@@ -325,34 +329,16 @@ def __init__(self, elastic_config):
...
-    async def prepare_index(
-        self, index, *, docs=None, settings=None, mappings=None, delete_first=False
-    ):
+    async def prepare_index(self, index, *, mappings=None):
         """Creates the index, given a mapping if it does not exists."""
-        if index.startswith("."):
-            expand_wildcards = "hidden"
-        else:
-            expand_wildcards = "open"
+        expand_wildcards = ElasticServer._get_expand_wildcards(index)
```
tarekziade, line 334: why not `self._get_expand_wildcards(index)` ?

connectors/byoei.py
```diff
@@ -471,3 +486,9 @@ async def async_bulk(
             "doc_deleted": fetcher.total_docs_deleted,
             "fetch_error": fetcher.fetch_error,
         }
+
+    def _get_expand_wildcards(index_name):
```
tarekziade, line 490: if this is part of the class, you want:

```
def _get_expand_wildcards(self, index_name):
```

tarekziade: I also wonder why you used that name for this function. shouldn't this be called `index_type()` ?

connectors/byoei.py
```diff
@@ -370,11 +356,40 @@ async def prepare_index(
...
+        """
+
+        expand_wildcards = ElasticServer._get_expand_wildcards(index)
+
+        exists = await self.client.indices.exists(
+            index=index, expand_wildcards=expand_wildcards
+        )
+
+        if exists:
+            logger.debug(f"{index} exists, deleting...")
```
tarekziade, line 382: I would use `warning`

connectors/kibana.py
```diff
@@ -114,14 +116,19 @@ async def prepare(service_type, index_name, config):
         }
 
         logger.info(f"Prepare {CONNECTORS_INDEX}")
-        await es.prepare_index(CONNECTORS_INDEX, docs=[doc], delete_first=True)
+        await es._override_index(CONNECTORS_INDEX, docs=[doc])
```
tarekziade, line 119: why did you rename it? if it's used outside the class, it should be a public name. why not keep `prepare_index` ? or rename it to `override_index` ?

connectors/tests/test_byoei.py
```diff
@@ -8,11 +8,11 @@
 import pytest
 
 from connectors.byoc import PipelineSettings
-from connectors.byoei import ElasticServer
+from connectors.byoei import ElasticServer, IndexMissing
 
 
 @pytest.mark.asyncio
-async def test_prepare_index(mock_responses):
+async def test__override_index(mock_responses):
```
tarekziade, line 15: two `__`, just use one

connectors/byoei.py
artem-shelkovnikov: `_get_expand_wildcards` is a pure function that does not need access to `self`.

connectors/byoei.py
artem-shelkovnikov: Yeah, I think this name makes much more sense! It was just simple pull-to-a-method refactoring and I did not think much of the method name

connectors/byoei.py
artem-shelkovnikov: I used the logger call from the original method and `debug` might make sense - this method is supposed to be called only from `kibana.py` thus Kibana somehow expects this and does not really care much IMO. I can change to warning too, just not sure if it's needed

connectors/kibana.py
artem-shelkovnikov: Now it's a different method. They serve pretty different purpose:

1. `prepare_index` is supposed to be called on content indices to check if they have any mappings. If they do not have any mappings, a valid expected mapping is applied to it, nothing else.
2. `_override_index` (name arbitrary, didn't think too much about it) is supposed to be called only by testing code (therefore, the underscore) to simulate Enterprise Search/Kibana creating _any_ index from scratch (connectors index, jobs index, content index). This method also tries to insert some data in its original implementation and I've kept this.

These 2 methods _can_ use same underlying method, of course, but their logic barely intersects and thus I've created 2 different methods.

connectors/tests/test_byoei.py
artem-shelkovnikov: I was hesitant about this cause of `_override_index` :D

connectors/byoei.py
tarekziade: that's what you did, you made it a method. you have two options: make it a method (and you need self) or make it a pure function:

```
def _get_expand_wildcards(index):
  XXX


class ElasticServer:
    def use_it(self): 
         _get_expand_wildcards(some_index)
```


OR

```
class ElasticServer:
    def _get_expand_wildcards(self, index):
      XXX

    def use_it(self): 
         self._get_expand_wildcards(some_index)
```

connectors/kibana.py
artem-shelkovnikov: Thinking more about this, method `_override_index` can actually be named `upsert_index` and whole method can be moved to `kibana.py`. 

For `prepare_index` I'd add a validation to check that it's called only on indices that start with `search-*`.

tarekziade: ok I think I see what bother me:

since `_override_index` is to be used purely in `kibana.py`, I advise removing it from `ElasticServer` completely.

and then have a `ElasticTestServer` that inherits from `ElasticServer` and implements its own `prepare_index` that you can use in `kibana.py` ibly or elsewhere in tests

Because `_override_index` is to never be run in production, and it's still in the execution path in production even if behind a flag

connectors/tests/test_byoei.py
tarekziade: yeah gotcha. just a nit though, it's usually bad practiove to have `__` in names unless when doing name mangling

connectors/byoei.py
artem-shelkovnikov: I see! I ended up dropping this function fully after refactoring :)

connectors/kibana.py
artem-shelkovnikov: I moved the method fully to Kibana in the end - inheritance hierarchy did not make much sense now + I saw `kibana.py` just using elasticsearch client directly.

What I did not like about it also that this way of implementing `prepare_index` differently in `ElasticTestServer` is:

- Methods have different signature
- It breaks Liskov principle by drastically changing the behaviour of the service with an override.

It can be addressed later so that `kibana.py` wraps those in a class, but for now probably that's good enough.

tarekziade (approved):

PR #311, opened 2023-01-13
Add pipeline to kibana.py

artem-shelkovnikov:
Minor change - when `kibana.py` creates an entry for connector, it does not provide `pipeline` setting at all, this PR adds default pipeline using same defaults as Kibana.


timgrein (approved):
LGTM

PR #312, opened 2023-01-13
Simplify the __init__ for BaseDataSource

artem-shelkovnikov:
Small refactoring for BaseDataSource:

Before:

```python
def __init__(self, connector):
    self.connector = connector
```

After:

```python
    self.configuration = configuration
```

Idea is - neither `BaseDataSource` nor its children need access to `Connector` class instance - moreover it can be used in a harmful way. This change makes it so that DataSource classes only receive configuration as an argument in the initializer.

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -543,7 +543,7 @@ class Connector:
                 logger.info("Sync forced")
 
             try:
-                self.data_provider = self.source_klass(self)
+                self.data_provider = self.source_klass(self.configuration)
             except Exception as e:
                 logger.critical(e, exc_info=True)
                 raise DataSourceError(
@@ -585,7 +585,7 @@ class Connector:
             result = await elastic_server.async_bulk(
                 self.index_name,
                 self.prepare_docs(self.data_provider),
-                self.data_provider.connector.pipeline,
+                self.pipeline,
                 options=bulk_options,
             )
             await self._sync_done(job, result)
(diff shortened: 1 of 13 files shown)


tarekziade (approved):
Thanks, good idea!

wangch079:
Hi @tarekziade , I just had a discussion with @danajuratoni regarding this change. 

I remembered that the connector implementation didn't need to be in this repo, it will work as long as we specify the fully qualified name here:

https://github.com/elastic/connectors-python/blob/0511b02ad754f70ab28c3f66564982556c2fb016/config.yml#L40

Just want to confirm, is it true that all the connector implementations will reside in this repo in the future? I just worried this change may break a connector implementation in another repo.

tarekziade:
> Hi @tarekziade , I just had a discussion with @danajuratoni regarding this change.
> 
> I remembered that the connector implementation didn't need to be in this repo, it will work as long as we specify the fully qualified name here:
> 
> https://github.com/elastic/connectors-python/blob/0511b02ad754f70ab28c3f66564982556c2fb016/config.yml#L40

Correct.

> 
> Just want to confirm, is it true that all the connector implementations will reside in this repo in the future? 

For all the connectors we implement, this is the intent yes.

> I just worried this change may break a connector implementation in another repo.

At this point I don't think we need to worry about this; we did not make any promises yet on the class signature stability.
We did not even make any promises yet on the connectors protocols HTTP API

PR #313, opened 2023-01-13
Update the autogenerated files

artem-shelkovnikov:
This PR updates automatically generated files.

Why did I do this? I ran `make clean && make run` and as a result of this command the files in this PR were updated. I believe these files are autogenerated and are not generated unless a clean reinstall is done, thus they were not updated.

cc @tarekziade - is this logic correct? Intuitively, the logic to generate these files automatically should run with each re-run.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,27 @@
+*********
+Changelog
+*********
+
+1.0.3 -- 2018-12-10
+===================
+
+* Add support for strings on input for decoding to match functionality of :class:`base64.b64decode`.
+  `#21 <https://github.com/aws/base64io-python/issues/21>`_
+  `#23 <https://github.com/aws/base64io-python/pull/23>`_
+  `#24 <https://github.com/aws/base64io-python/pull/24>`_
+
+1.0.2 -- 2018-08-01
+===================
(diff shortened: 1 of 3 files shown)


tarekziade:
something is very odd here. those are files from the `base64io` project, not our files... it looks like the installation of `base64io` is buggy, as it creates those files in the root.

I can confirm it's a problem on their side, see https://github.com/aws/base64io-python/issues/38

artem-shelkovnikov:
Yeah I can see it's probably cause of this line:
https://github.com/aws/base64io-python/blob/master/setup.py#L41

Btw what is the use-case for the library? Should we swap to a different one or should I try to sumbit PR to the base64io-python first?

tarekziade:
> Btw what is the use-case for the library?

it a fallback solution for stream-encoding base64 when you don't have `base64` on your system (win32)

https://github.com/elastic/connectors-python/blob/main/connectors/utils.py#L220

>  should I try to sumbit PR to the base64io-python first?

Yeah I would do this and see if they are responsive

artem-shelkovnikov:
Made the PR in the package repo: https://github.com/aws/base64io-python/pull/41

PR #317, opened 2023-01-13
Add service_type to config file

seanstory:
In order to run as a connector client (non-native), you need to specify the `service_type`, but this wasn't included in the default config file.


config.yml
```diff
@@ -33,8 +33,9 @@ native_service_types:
   - gcs
   - abs
 
-# some id
-#connector_id: '1'
+# Connector client settings
```
tarekziade, line 36: maybe we can add an extra comment to say that when used on-premise, we need both (it's not explicit in Kibana's UI)

tarekziade (approved):

config.yml
wangch079: Ideally, we should have some validation. There are many places in the code where we just assume a valid config. This is the same for the ruby connector where we only have basic validation (e.g. a field is required and it must be an integer)

wangch079 (approved):

seanstory:
Thanks for merging this, Tarek. It had totally fallen off my radar.

PR #318, opened 2023-01-13
fix macOS specific issue

tarekziade:
`base64` on macOs returns `\n` instead of an empty string.  Here, we strip the data before we put it in the attachment key

A better long term fix would be to generate a real 15 bytes files in those tests, but it's still a good idea to strip

cc @praveen-elastic 

#### Pre-Review Checklist


artem-shelkovnikov (approved):
Looks good, but IMO it also asks for a comment that describes why "strip" is happening.

PR #319, opened 2023-01-13
the conversion for bool=false is broken

tarekziade:
Found a small conversion bug when I was playing with Azure+MySQL

Also clarified the label of that flag, it was unclear

- [x] added tests
- [x] tested locally 
- [x] needs 8.6 backport

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -36,7 +36,7 @@ class Field:
         elif type_ == "float":
             return float(value)
         elif type_ == "bool":
-            return value.lower() in "y", "yes", "true", "1"
+            return value.lower() in ("y", "yes", "true", "1")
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov (approved):
LGTM, but I'd add more tests to cover the problem that surfaced here

tarekziade:
> LGTM, but I'd add more tests to cover the problem that surfaced here

What do you have in mind? The failure happened when I changed `ssl_disabled` to `False` using the MySQL connector. The field got converted to 

```
False, "y", "yes", "true", "1"
```

Instead of `False` - So I added a test, reproduced the same error, and fixed the code, which covers this problem.
Happy to add more tests, but I can't think of another one that would add some value

PR #320, opened 2023-01-16
Decouple job scheduler and executor

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3639

This PR decouples the job scheduler and executor.

The job scheduler is in charge of 1) sending heartbeat for connectors, 2) updating service type and configurations for non-native connectors, and 3) scheduling jobs for connectors.

The job executor periodically polls pending jobs from the sync job index and executes them.

#### Pre-Review Checklist
- [ ] ~Considered corresponding documentation changes~
- [ ] ~Contributed any configuration settings changes to the configuration reference~


connectors/services/job_scheduler.py
```diff
@@ -0,0 +1,200 @@
...
+                    return
+                except DataSourceError as e:
+                    await connector.error(e)
+                    logger.critical(e, exc_info=True)
+                    self.raise_if_spurious(e)
+                    return
+
+            try:
+                # the heartbeat is always triggered
+                connector.start_heartbeat(self.hb)
```
artem-shelkovnikov, line 98: I think it's sufficient to call heartbeat once here - this whole function will take less than a second to execute normally. Though this can be changed later not to bloat this PR.

connectors/services/job_scheduler.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+
+    async def run(self):
+        if "PERF8" in os.environ:
+            import perf8
+
+            async with perf8.measure():
+                return await self._run()
+        else:
+            return await self._run()
```
artem-shelkovnikov, line 158: I'm not sure PERF8 is needed here - but if so, will the measurements from this service go to a different directory than the job executor? It's nice to have these metrics separated cc @tarekziade

connectors/services/job_scheduler.py
wangch079: That will introduce `no. of connectors` writes to `.elastic-connectors` index every `config.idle` seconds.

connectors/services/job_scheduler.py
wangch079: I had the similar question when I was drafting the PR: should `trace_mem` only be applied to job execution? But I see `trace_mem` was removed by another PR, so the question now is: should `PERF8` be applied at the application level, service level (i.e. job scheduler service and job executor service) or job execution level?

tarekziade: I think we can put this back into the `cli.py` top-level function because we measure these things in the context of a whole benchmark. We do want to capture interactions and problems between services. For instance the event loop lag.

connectors/byoc.py
```diff
@@ -159,6 +159,11 @@ def _create_object(self, doc_source):
         )
 
 
+class SyncJobIndex(ESIndex):
+    def _create_object(self, doc):
+        pass
```
tarekziade, line 164: This implementation is mandatory

connectors/services/job_executor.py
```diff
@@ -0,0 +1,228 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+"""
+Event loop
+
+- polls for work by calling Elasticsearch on a regular basis
```
tarekziade, line 9: can we fix this comment?

connectors/services/base.py
```diff
@@ -17,5 +20,5 @@ async def run(self):
         raise NotImplementedError()
 
     def shutdown(self, sig):
-        logger.info(f"Caught {sig.name}. Graceful shutdown.")
+        logger.info(f"Caught {sig.name}. Shutdown {self.__class__.__name__}.")
```
tarekziade, line 23: let's remove `Caught {sig.name}` this should be in the `_service_shutdown`

connectors/cli.py
```diff
@@ -101,15 +102,17 @@ def run(args):
...
 
-    service = SyncService(config, args)
-    coro = service.run()
+    services = [JobSchedulerService(config), JobExecutorService(config)]
     loop = get_event_loop(args.uvloop)
 
     for sig in (signal.SIGINT, signal.SIGTERM):
-        loop.add_signal_handler(sig, functools.partial(service.shutdown, sig))
+        for service in services:
+            loop.add_signal_handler(sig, functools.partial(service.shutdown, sig))
```
tarekziade, line 110: we can't have several signal handlers for one signal. we need to have a single function, that iterates on services shutdown

tarekziade: (but I think that's what you did in the other patch)

connectors/services/job_executor.py
```diff
@@ -0,0 +1,228 @@
...
+        self.elastic_server = elastic_server
+        self.bulk_options = bulk_options
+        self.connector_index = connector_index
+        self.sync_job_index = sync_job_index
+
+    async def execute(self):
+        await self.sync_job.claim()
+
+        start_time = time.time()
+        index_name = self.sync_job.connector["index_name"]
```
tarekziade, line 46: Performance nit

You are using `self.sync_job.connector` everywhere in this function. This adds 3 indirections in Python.

It's best to do:

```
connector = self.sync_job.connector
service_type = connector["service_type"]
language_code = connector["language"]
```

connectors/services/job_executor.py
```diff
@@ -0,0 +1,228 @@
...
+        service_type = self.sync_job.connector["service_type"]
+        language_code = self.sync_job.connector["language"]
+        pipeline = PipelineSettings(self.sync_job.connector.get("pipeline", {}))
+
+        fqn = self.sources[service_type]
+        configuration = DataSourceConfiguration(
```
tarekziade, line 52: can this fails? do we want to catch any exception here too?

connectors/services/job_executor.py
```diff
@@ -0,0 +1,228 @@
...
+            )
+            await asyncio.sleep(0)
+
+            result = await self.elastic_server.async_bulk(
+                index_name,
+                self.prepare_docs(pipeline, data_provider),
+                pipeline,
+                options=self.bulk_options,
+            )
+            for i in range(50):
```
tarekziade, line 88: That block needs to be removed. or replaced by

```
await asyncio.sleep(5)
```

when you call sleep, you have already sent back the execution to the event loop, no need to loop on smaller sleeps

connectors/services/job_executor.py
```diff
@@ -0,0 +1,228 @@
...
+class JobExecutorService(BaseService):
+    def __init__(self, config):
+        super().__init__(config)
+        self.errors = [0, time.time()]
+        self.service_config = self.config["service"]
+        self.idling = self.service_config["idling"]
+        self.connector_index = None
+        self.sync_job_index = None
+
+    def raise_if_spurious(self, exception):
```
tarekziade, line 138: this can go into the `BaseService` class along with `self.errors` -- we are likely to use it in all services.

connectors/services/job_executor.py
```diff
@@ -0,0 +1,228 @@
...
+                sources=self.config["sources"],
+                sync_job=sync_job,
+                elastic_server=self.elastic_server,
+                connector_index=self.connector_index,
+                sync_job_index=self.sync_job_index,
+                bulk_options=self.bulk_options,
+            )
+            await sync_job_runner.execute()
+
+            await asyncio.sleep(0)
```
tarekziade, line 174: we can remove that sleep (the next await is doing the switch)

connectors/services/job_scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self, config):
+        super().__init__(config)
+        self.errors = [0, time.time()]
+        self.service_config = self.config["service"]
+        self.idling = self.service_config["idling"]
+        self.hb = self.service_config["heartbeat"]
+        self.connector_index = None
+        self.sync_job_index = None
+
+    def raise_if_spurious(self, exception):
```
tarekziade, line 47: ah! that's what I was saying earlier. Move it to `BaseService`

connectors/services/job_scheduler.py
```diff
@@ -0,0 +1,206 @@
...
+                    if connector.status != Status.CONNECTED:
+                        connector.status = Status.CONNECTED
+                        await connector.sync_doc()
+                    return
+            else:
+                logger.info("Sync forced")
+
+            await self._create_job(connector)
+
+            await asyncio.sleep(0)
```
tarekziade, line 132: this can be removed

connectors/tests/test_sync.py
```diff
@@ -15,7 +15,7 @@
 from connectors.config import load_config
 from connectors.conftest import assert_re
 from connectors.filtering.validation import InvalidFilteringError
-from connectors.services.sync import SyncService
+from connectors.services.job_scheduler import JobSchedulerService
```
tarekziade, line 18: we need to create a `test_job_scheduler` and move the test there

tarekziade (commented):
looks good so far!

connectors/byoc.py
wangch079: It should be done here: https://github.com/elastic/connectors-python/pull/332

connectors/services/job_executor.py
wangch079: Will do. It's still in draft.

connectors/cli.py
wangch079: Got it. Will make the change.

connectors/services/base.py
wangch079: OK, will do.

connectors/services/job_executor.py
wangch079: Probably we could have this in `SyncJob` class:

```python
class SyncJob:

    ...

    @property
    def connector():
      self.doc_source.get('connector', {})

    @property
    def service_type():
      self.connector['service_type']

    @property
    def language_code():
      self.connector['language']
```

cc. @vidok

connectors/services/job_scheduler.py
wangch079: I have done it in this PR: https://github.com/elastic/connectors-python/pull/337

connectors/services/base.py
wangch079: I have done it in this PR: https://github.com/elastic/connectors-python/pull/337

connectors/cli.py
wangch079: I have done it in this PR: https://github.com/elastic/connectors-python/pull/337

connectors/services/job_executor.py
wangch079: I have done it in this PR: https://github.com/elastic/connectors-python/pull/337

connectors/services/job_scheduler.py
wangch079: I have done it in this PR: https://github.com/elastic/connectors-python/pull/337

vidok:
I know it's still a draft but I'd like to ask a small question: 

Would it be possible to stick to the same class/object names as in the ruby framework? `Producer/Consumer` instead of `Scheduler/Executor`? If it was intentional I'm curious about the reasons.

artem-shelkovnikov:
> Would it be possible to stick to the same class/object names as in the ruby framework? Producer/Consumer instead of Scheduler/Executor? If it was intentional I'm curious about the reasons.

Yes, absolutely! I think it's a good idea!

wangch079:
> Would it be possible to stick to the same class/object names as in the ruby framework? `Producer/Consumer` instead of `Scheduler/Executor`? If it was intentional I'm curious about the reasons.

> Yes, absolutely! I think it's a good idea!

Yes, I renamed them intentionally, as I feel `Producer/Consumer` is more technical, as in the classical [Producer-consumer problem](https://en.wikipedia.org/wiki/Producer%E2%80%93consumer_problem). `Scheduler/Executor` reflects what it really does: the producer schedules jobs (although I agree `schedule` is not 100% accurate), and the consumer executes jobs.

connectors/services/job_executor.py
tarekziade: well it's not solving the performance issue hehe, but it's cleaner, I like it!

wangch079:
Close the PR as it's deferred to later release, and it will be easier to start fresh, given the conflicts to resolve.

Issue #315, opened 2023-01-13
`test_ping_for_failed_connection` blocks on macOS


tarekziade:
### Problem Description

running `make test` - connectors/sources/tests/test_abs.py::test_ping_for_failed_connection blocks

tarekziade:
well actually it works! but it takes 88 seconds for that single test. 

```
[FMWK][09:45:16][ERROR] Error while connecting to the Azure Blob Storage.
Traceback (most recent call last):
  File "/Users/tarekziade/Dev/connectors-python/lib/python3.10/site-packages/aiohttp/connector.py", line 980, in _wrap_create_connection
    return await self._loop.create_connection(*args, **kwargs)  # type: ignore[return-value]  # noqa
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 1064, in create_connection
    raise exceptions[0]
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 1049, in create_connection
    sock = await self._connect_sock(
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/base_events.py", line 960, in _connect_sock
    await self.sock_connect(sock, address)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/selector_events.py", line 500, in sock_connect
    return await fut
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/selector_events.py", line 535, in _sock_connect_cb
    raise OSError(err, f'Connect call failed {address}')
ConnectionRefusedError: [Errno 61] Connect call failed ('127.0.0.1', 10000)
```


I suspect we miss a timeout somewhere cc @praveen-elastic

parth-crest:
@tarekziade Sure, we will take a look into it and update you. Can we change the label to enhancement as this test seems to be working fine and we need to ensure it takes lesser time to execute?

tarekziade:
Looks like @artem-shelkovnikov is looking into this in https://github.com/elastic/connectors-python/pull/322

tarekziade:
fixed by https://github.com/elastic/connectors-python/pull/322

PR #322, opened 2023-01-16
Mock some calls for abs and gcs tests

artem-shelkovnikov:
Improves test performance - some tests are taking too long to finish.

Changes in the PR:

- Remove `connectors/sources/tests/test_abs.py::test_ping_for_empty_connection_string` - judging from the implementation of `ping`, there is no actual different in our code regarding connection strings. ABS client seems to have some internal implementation difference based on connection strings, but our code does not make a difference - it just catches a wide `Exception` and handles it. Thus this test is deleted as previous test `test_ping_for_failed_connection` handles exactly same scenario.
- `connectors/sources/tests/test_abs.py::test_ping_for_failed_connection` - now mocks response from 3rd-party instead of making a real call to 3rd-party system
- `connectors/sources/tests/test_gcs.py:: test_ping_for_failed_connection` - now mocks response from 3rd-party instead of making a real call to 3rd-party system + removing retries at all (otherwise they'll take some time)

:warning: One test is still there that is slower than 10s: `connectors/tests/test_sync.py::test_connector_service_poll_large`. That seems to happen because the test just does handle reasonably large amount of data, see:

https://github.com/elastic/connectors-python/blob/main/connectors/tests/fake_sources.py#L83-L85

```python
class LargeFakeSource(FakeSource):
    """Phatey"""

    service_type = "large_fake"

    async def get_docs(self):
        for i in range(1001):
            doc_id = str(i + 1)
            yield {"_id": doc_id, "data": "big" * 1024 * 1024}, partial(
                self._dl, doc_id
            )
```

If we reduce the size of the payload, the test gets much faster, though I'm not sure about original intent of the test.

### DIFF
--- a/connectors/sources/tests/test_abs.py
+++ b/connectors/sources/tests/test_abs.py
@@ -77,7 +77,9 @@ async def test_ping_for_failed_connection():
     source = create_source(AzureBlobStorageDataSource)
 
     with patch.object(
-        BlobServiceClient, "get_account_information", side_effect=Exception("Something went wrong")
+        BlobServiceClient,
+        "get_account_information",
+        side_effect=Exception("Something went wrong"),
     ):
         # Execute
         with pytest.raises(Exception):
--- a/connectors/sources/tests/test_abs.py
+++ b/connectors/sources/tests/test_abs.py
@@ -76,22 +76,12 @@ async def test_ping_for_failed_connection():
     # Setup
     source = create_source(AzureBlobStorageDataSource)
 
-    # Execute
-    with pytest.raises(Exception):
-        await source.ping()
-
-
-@pytest.mark.asyncio
-async def test_ping_for_empty_connection_string():
-    """Test ping method of AzureBlobStorageDataSource class with empty connection string"""
-
-    # Setup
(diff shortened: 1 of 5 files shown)


timgrein (approved):
LGTM. I would also say that we either offload the test with the large FakeDataSource to the functional tests or remove it and make sure that every "real" functional test for every specific data source has a test with a large amount of data.

tarekziade:
Related https://github.com/elastic/connectors-python/issues/315

tarekziade:
> If we reduce the size of the payload, the test gets much faster, though I'm not sure about original intent of the test.

The goal of the test is to exercise batches. if you reduce the size it will never batch. To make it faster, you can reduce the size of the batch to a small number and reduce the sie of the data. You will get the same coverage

tarekziade:
buildkite test this

artem-shelkovnikov:
buildkite test this

artem-shelkovnikov:
buildkite test this

artem-shelkovnikov:
buildkite test this

artem-shelkovnikov:
buildkite test this

PR #328, opened 2023-01-16
removed guppy and tracemem

tarekziade:
we don't need `tracemem` anymore as we have `perf8` now


artem-shelkovnikov (approved):
Nice, love code being cleaned up!

PR #331, opened 2023-01-17
Add original exception message while raising it.

### LINKED ISSUE #284: Original exception body is missing
### Problem Description

When [byoc initializes](https://github.com/elastic/connectors-python/blob/4ca7026e8c3c67950703b18eaf3bd61074b3d14a/connectors/byoc.py#L448) the source class, on encountering an exception it raises a [generic exception
](https://github.com/elastic/connectors-python/blob/4ca7026e8c3c67950703b18eaf3bd61074b3d14a/connectors/byoc.py#L463)
It doesn't contain any information from the originally raised exception.

### Proposed Solution

the logger message from [this](https://github.com/elastic/connectors-python/blob/4ca7026e8c3c67950703b18eaf3bd61074b3d14a/connectors/byoc.py#L463) line can be changed to this to let user know the actual exception:
raise DataSourceError(f"Could not instantiate {fqn} for {service_type}. Exception encountered : {e}")

jignesh-crest:
## Closes #284 

Added exception body in the logged message. Which will be visible in the Kibana UI for a better user experience.

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -549,7 +549,7 @@ class Connector:
             except Exception as e:
                 logger.critical(e, exc_info=True)
                 raise DataSourceError(
-                    f"Could not instantiate {self.source_klass} for {service_type}"
+                    f"Could not instantiate {self.source_klass} for {service_type}. Exception encountered: {e}"
                 )
 
             if not await self.data_provider.changed():
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -508,7 +508,9 @@ class Connector:
                 )
         except Exception as e:
             logger.critical(e, exc_info=True)
-            raise DataSourceError(f"Could not instantiate {fqn} for {service_type}")
+            raise DataSourceError(
+                f"Could not instantiate {fqn} for {service_type}. Exception encountered: {e}"
+            )
 
         self.source_klass = source_klass
 
--- a/connectors/tests/test_sync.py
+++ b/connectors/tests/test_sync.py
@@ -578,7 +578,7 @@ async def test_connector_service_poll_buggy_service(
         doc = json.loads(kw["data"])["doc"]
         assert (
             doc["error"]
-            == "Could not instantiate <class 'fake_sources.FakeSource'> for fake"
+            == "Could not instantiate <class 'fake_sources.FakeSource'> for fake. Exception encountered: I break on init"
         )
         return CallbackResult(status=200)
 


connectors/byoc.py
```diff
@@ -508,7 +508,9 @@ async def prepare(self, config):
                 )
         except Exception as e:
             logger.critical(e, exc_info=True)
-            raise DataSourceError(f"Could not instantiate {fqn} for {service_type}")
+            raise DataSourceError(
+                f"Could not instantiate {fqn} for {service_type}. Exception encountered: {e}"
+            )
```
artem-shelkovnikov, line 513: Will [this](https://stackoverflow.com/questions/1350671/inner-exception-with-traceback-in-python) work?

```suggestion
            raise DataSourceError(f"Could not instantiate {fqn} for {service_type}") from e
```

jignesh-crest: Let me take a look, I'll update you.

jignesh-crest: I checked, But still, with this suggestion, the message will not be visible in the KibanaUI.

jignesh-crest:
Modified test to comply with the added change.

connectors/byoc.py
artem-shelkovnikov: @jignesh-crest does it have to be visible in KibanaUI? That's a purely techincal error that should be accessed via logs of the connector. It does not seem to make sense to display it to the user - maybe worth to just say something like "Check service logs for the exception" in the message.

Moreover, the user should never see this error - it's seen only when connector is in development.

praveen-elastic:
buildkite test this

tarekziade (commented):
The change looks good but I wonder if this not going to be a security issue.
e.g. if the exception we display leaks some information. 

@danajuratoni @serenachou what is your take on this?

it could potentially reveal the backend IP address etc

artem-shelkovnikov:
@tarekziade also see discussion in the PR: https://github.com/elastic/connectors-python/issues/284

danajuratoni:
Errors in Kibana UI should have a user friendly format and provide a better user experience. 
BE can define the guidelines for appropriate exception handling and if/when sanitizing exception messages makes sense.

tarekziade (approved):

connectors/byoc.py
serenachou: Where would they look at these logs if it's a native connector @artem-shelkovnikov ? I would like to confirm that once this is a native connector that we could say: Click to Discover -> Logs and surface these errors more specifically

artem-shelkovnikov: I'm actually not 100% sure - I know it'll be "same place where you look at Enterprise Search logs". So if log shipping is enabled, it should be in different deployment, otherwise in same deployment in logs as you described.

tarekziade:
We talked about this in a meeting today. The plan is to add a `validate_config` method in the base source class, that can be used to verify the config at instantiation time and raise a more precise error if the instantiation of the class cannot be done (overridable by source classes)

tarekziade (changes requested):
see https://github.com/elastic/connectors-python/pull/331#issuecomment-1430025540

jignesh-crest:
Closing this PR now as #552 would be used for delivering the validation exceptions to Kibana UI.

PR #337, opened 2023-01-18
Refactor services

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/3639

This PR made the following changes:
1. The CLI will be able to run multiple services.
2. Move `perf8` to the application level.
3. Move some common attributes to `BaseService`
4. Make one service only run once.

#### Pre-Review Checklist

### DIFF
--- a/connectors/services/sync.py
+++ b/connectors/services/sync.py
@@ -56,12 +56,6 @@ class SyncService(BaseService):
         self.preflight_idle = int(self.service_config.get("preflight_idle", 30))
         self.connectors = None
 
-    def stop(self):
-        self.running = False
-        self._sleeps.cancel()
-        if self.connectors is not None:
-            self.connectors.stop_waiting()
-
     async def _one_sync(self, connector, es, sync_now):
         if connector.native:
             logger.debug(f"Connector {connector.id} natively supported")
@@ -149,7 +143,8 @@ class SyncService(BaseService):
                         break
                 await self._sleeps.sleep(self.idling)
         finally:
-            self.stop()
-            await self.connectors.close()
+            if self.connectors is not None:
+                self.connectors.stop_waiting()
+                await self.connectors.close()
             await es.close()
         return 0
--- a/connectors/services/sync.py
+++ b/connectors/services/sync.py
@@ -107,13 +107,9 @@ class SyncService(BaseService):
         finally:
             await connector.close()
 
-    async def run(self):
+    async def _run(self):
         """Main event loop."""
-
-        await super().run()
-
         self.connectors = ConnectorIndex(self.es_config)
-        self.running = True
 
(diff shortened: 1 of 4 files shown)


connectors/tests/test_sync.py
```diff
@@ -436,37 +436,6 @@ async def test_connector_service_poll_large(mock_responses, patch_logger, set_en
     assert_re("Sync done: 1001 indexed, 0  deleted", patch_logger.logs)
 
 
-@pytest.mark.asyncio
-async def test_connector_service_poll_clear_error(
```
artem-shelkovnikov, line 440: What about the deleted tests? Was the functionality removed?

connectors/services/sync.py
```diff
@@ -132,18 +108,11 @@ async def _one_sync(self, connector, es, sync_now):
...
-            async with perf8.measure():
-                return await self._run()
-        else:
-            return await self._run()
-
-    async def _run(self):
         """Main event loop."""
 
-        self.connectors = BYOIndex(self.config["elasticsearch"])
+        await super().run()
```
artem-shelkovnikov, line 113: I'm not sure about this pattern of usage of base implementation.

What do you think instead in wrapping this function the following way:

```
class BaseService:
  def start(self):
    # do logic
    # then run the inherited class's method that does the logic
    await self.run()

  def run(self):
    raise NotImpelementedException # should be overriden in the class that inherits BaseService

```

artem-shelkovnikov: Dunno though, maybe it's a python-default way of doing so, but I find it more error-prone:

- You can forget calling `await.super().run()` in child classes
- You have to put this code in specific place in child classes for it to wor
- Once base class's `run` function grows and does more complex logic, it might want to wrap child class's `run` method from both sides - before and after execution

connectors/tests/test_sync.py
wangch079: I don't know exactly what it's trying to test, but the 2 deleted tests try to run `service.run()` twice.

connectors/services/sync.py
wangch079: made the change in https://github.com/elastic/connectors-python/pull/337/commits/b41fcb72af21f6dfab7c7b0d2f5a602e95e0d29f

artem-shelkovnikov (approved):
Looks good for the start, I can see some logic moved from `_run` to `run` - like making it `running = True`, same with `stop` - BaseService can do `running = False` and other generic stuff.

This can be done as a separate PR though.

wangch079:
I forgot to remove [these 2 fields](https://github.com/elastic/connectors-python/pull/337/commits/e93b64bd7e44481338587f4abac09ff4a331c7c1) in PR https://github.com/elastic/connectors-python/pull/297

PR #340, opened 2023-01-18
[Filtering] Integrate the BasicRuleEngine into the sync

timgrein:
### Closes https://github.com/elastic/enterprise-search-team/issues/3581

Integrate the `BasicRuleEngine` into the sync. Also add the missing `Fetcher` tests for `get_docs` as this is one of the key places for syncs to actually work.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -28,7 +28,9 @@ from collections import defaultdict
 from elasticsearch import NotFoundError as ElasticNotFoundError
 from elasticsearch.helpers import async_scan
 
+from connectors.byoc import Filtering
 from connectors.es import ESClient
+from connectors.filtering.basic_rule import BasicRuleEngine, parse
 from connectors.logger import logger
 from connectors.utils import (
     DEFAULT_CHUNK_MEM_SIZE,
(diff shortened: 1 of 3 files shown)


connectors/byoei.py
```diff
@@ -175,10 +177,13 @@ def __init__(
         queue,
         index,
         existing_ids,
+        filtering=None,
```
artem-shelkovnikov, line 180: Why allow for None here?

connectors/byoei.py
```diff
@@ -192,6 +197,10 @@ def __init__(
         self.total_docs_created = 0
         self.total_docs_deleted = 0
         self.fetch_error = None
+        self.filtering = filtering
+        self.basic_rule_engine = BasicRuleEngine(
+            parse(filtering.get_active_filter().get("rules", []))
+        )
```
artem-shelkovnikov, line 203: Intuitively feels like this syntax will be a bit better:

```suggestion
        self.basic_rule_engine = BasicRuleEngine.parse(filtering.get_active_filter().get("rules", [])))
```

timgrein: The `BasicRuleEngine`'s concern is to execute the rules and not to parse them. Actually `parse` lived inside a `BasicRuleParser` class, but Tarek meant that `parse` should be a top-level function to make it more pythonic as there's no need for an object/class wrapping it. But to me your or Tarek's way is fine tbh.

connectors/byoei.py
timgrein: Good point mixed up best practices for mutable data types and default keyword args. Will change it.

artem-shelkovnikov (approved):

PR #346, opened 2023-01-18
[Filtering] Pass valid filtering to a sync

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3571 ##

This PR makes sure that a sync is only executed if filtering is valid. Also we should not quit when filtering is invalid, but update the status in the .elastic-connectors index (will follow with another PR).

#### Pre-Review Checklist

### DIFF
--- a/connectors/filtering/validation.py
+++ b/connectors/filtering/validation.py
@@ -12,6 +12,20 @@ from connectors.filtering.basic_rule import BasicRule, Policy, Rule
 from connectors.logger import logger
 
 
+async def validate_filtering(connector):
+    validation_result = await connector.source_klass.validate_filtering(
+        connector.filtering.get_active_filter()
+    )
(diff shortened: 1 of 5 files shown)


timgrein:
Draft for now: will incorporate the update of the es-index with validation result also here as I need the changes made in this PR

timgrein:
@tarekziade Added the performance improvements into a separate issue.

connectors/filtering/validation.py
```diff
@@ -105,6 +147,12 @@ def __eq__(self, other):
 
         return self.state == other.state and self.errors == other.errors
 
+    def to_dict(self):
```
tarekziade, line 150: as an improvement for later, you could also make that class act as a dict, so it can be directly used 

see https://realpython.com/inherit-python-dict/

tarekziade (approved):

PR #347, opened 2023-01-18
Fix one flaky test relying on time

artem-shelkovnikov:
See for example pipeline https://buildkite.com/elastic/connectors-python/builds/1611#0185c5c1-f3fb-47bd-a9c3-91f50b4270fa: 

```
=================================== FAILURES ===================================
________________________________ test_next_run _________________________________

    def test_next_run():
        # can run within two minutes
        assert next_run("1 * * * * *") < 120
>       assert next_run("* * * * * *") == 0
E       AssertionError: assert 1.999814 == 0
E        +  where 1.999814 = next_run('* * * * * *')

connectors/tests/test_utils.py:35: AssertionError
```

The problem happens because test that checks when next cron trigger happens uses current time for the fixture - in some cases it seems to produce unexpected results.

This fix adds `freezegun` package and uses it to rewind the time for this specific test: https://github.com/spulec/freezegun

### DIFF
--- a/connectors/tests/test_utils.py
+++ b/connectors/tests/test_utils.py
@@ -15,6 +15,7 @@ import time
 
 import pytest
 from pympler import asizeof
+from freezegun import freeze_time
 
 from connectors import utils
 from connectors.utils import (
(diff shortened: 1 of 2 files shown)


tarekziade (approved):
looks good, 

license checked https://github.com/spulec/freezegun/blob/master/LICENSE

Issue #363, opened 2023-01-20
Display the most useful columns by default in Documents UI for S3 connectors


ppf2:
Here's a documents view of synced documents from S3:

<img width="732" alt="image" src="https://user-images.githubusercontent.com/7216393/213588913-0846d42a-22bb-4fb1-9e66-8888f28583c4.png">

You will see that by default, it displays 3 dummy columns, but it is inconsistent.  Some records show the filename in the default 3 column view, but other records do not.  

A better experience will be to always display the 3 most useful fields.  I think the issue here is that we always display the first 3 returned:

<img width="639" alt="image" src="https://user-images.githubusercontent.com/7216393/213589110-0425051c-b26d-47ef-8151-65ddc83c71aa.png">

So if the document had an extraction done against it, the filename is missed in the default view because it is now the 4th entry.  Instead of always displaying only the first 3 fields, it will be helpful to actually define 3 specific fields to display, e.g.

- bucket
- filename
- body

An even better approach will be to let the user configure their desired default fields to display.

ppf2:
Also when displaying `size`, it will be helpful to include the unit of measure.

ppf2:
We may also want to add a tooltip to `_timestamp` column in the Documents view to indicate that this is the timestamp of when the document is indexed to Elasticsearch (not to be confused with the timestamp of the document within S3).

akanshi-crest:
> Also when displaying `size`, it will be helpful to include the unit of measure.

sure we'll add this with size.

akanshi-crest:
> We may also want to add a tooltip to _timestamp column in the Documents view to indicate that this is the timestamp of when the document is indexed to Elasticsearch (not to be confused with the timestamp of the document within S3).

> Display the most useful columns by default in Documents UI for S3 connectors

> An even better approach will be to let the user configure their desired default fields to display.

These 3 points needs to taken care at the framework level or the UI level. We will update the PR with adding the measurement unit for size. 

cc: @tarekziade

khusbu-crest:
Can we close this enhancement as the PR for the same is merged? 
CC: @tarekziade  @ppf2

danajuratoni:
Measurement unit for size was added. Making improvements to the documents view is something impacting all ingestion methods, so I'll be moving this part to be together with other usability enhancements requests. cc: @serenachou

PR #364, opened 2023-01-20
Include the unit of measure for size field in Amazon S3

akanshi-crest:
Added bytes as measurement unit for the field `size` as mentioned in the issue https://github.com/elastic/connectors-python/issues/363

#### Pre-Review Checklist


tarekziade:
for s3 changes please add the `v8.6.0` label so it's backported to the `8.6` branch

connectors/sources/s3.py
```diff
@@ -185,19 +186,23 @@ async def get_docs(self):
                         doc_id = md5(
                             f"{bucket}/{obj_summary.key}".encode("utf8")
                         ).hexdigest()
+                        object_size = await obj_summary.size
 
                         doc = {
                             "_id": doc_id,
                             "filename": obj_summary.key,
-                            "size": await obj_summary.size,
+                            "size": f"{object_size} bytes",
```
tarekziade, line 194: this turns the field into a string, is this what we really want @ppf2 ?

ppf2: Certainly not :) We don't want to change the field into a string.  `size` should remain an integer/long field to support sorting and mathematical functions.  We simply need to provide an indicator of the unit of measure, not to change the actual value of the field (e.g. either update the field name to include bytes there, or add a tooltip to the UI for `size` to tell the user that it is in bytes, etc..

akanshi-crest: If we are planning to a tooltip for this so it is going to be handle from the UI level and may be @tarekziade can handle this thing but if we have to update field name to include bytes in that case can you please suggest what's the most suitable name for this.
example: **size (in bytes)**, **size (bytes)**

akanshi-crest: Please let us know if you want me to update the size key name to size (bytes) or anything else. However if we are planning to add a tooltip, It needs to be taken care at the UI level.
cc Tarek

ppf2: +@danajuratoni This feels like a PM decision on the design.

danajuratoni: Here's my delayed (sorry for the delay) comment on wether "bytes" should be in the field name (to include bytes) or a tooltip in UI. 
Atm the only feasible option is to change the field name to include bytes, there is no tooltip capability in 8.7. We'll want to add that in 8.8, but for now we should use the field name.

akanshi-crest: I have updated the PR

connectors/sources/s3.py
```diff
@@ -110,7 +110,7 @@ async def _get_content(self, doc, region, timestamp=None, doit=None):
         if os.path.splitext(filename)[-1] not in TIKA_SUPPORTED_FILETYPES:
             logger.debug(f"{filename} can't be extracted")
             return
-        if doc["size"] > DEFAULT_MAX_FILE_SIZE:
+        if doc["size (bytes)"] > DEFAULT_MAX_FILE_SIZE:
```
tarekziade, line 113: let's go for `size_in_bytes` -- I don't want to introduce a filed name that's a complex description

tarekziade (changes requested):

tarekziade (approved):

PR #368, opened 2023-01-23
Add pytest plugin to fail tests that are too slow

artem-shelkovnikov:
A couple small changes bundled:

1. Add `pytest-fail-slow` pytest plugin that fails tests that are too slow. Too slow = 1 second for now (there are a couple tests that take ~500ms). This can be changed later to something smaller, ideally ~100ms. Package page: https://pypi.org/project/pytest-fail-slow/ Package license is MIT.
2. Speed up `connectors/tests/test_utils.py::test_mem_queue_race` by decreasing intervals + set test timeout to 1 second

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 PYTHON=python3.10
 ARCH=$(shell uname -m)
 PERF8?=no
+SLOW_TEST_THRESHOLD=1 # seconds
 
 bin/python:
 	$(PYTHON) -m venv .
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov:
@timgrein it's indeed possible to set slowness criteria per test marking the test with `@pytest.mark.fail_slow(<number>)`

timgrein (approved):
Nice, very good addition IMO!

PR #369, opened 2023-01-23
[Filtering] Respect filtering feature flags (backwards compatible)

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3585 ###

This PR introduces the checks to decide whether to schedule filtering validation or not. This also decided whether the `BasicRuleEngine` and the pre-sync validation will get executed during a sync or not.

The backend checks for the old and new filtering feature flags to keep backwards compatibility.

#### Pre-Review Checklist
~~- [ ] Considered corresponding documentation changes~~
~~- [ ] Contributed any configuration settings changes to the configuration reference~~

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -595,7 +595,7 @@ class Connector:
 
     async def sync(self, elastic_server, idling, sync_now=False):
         # If anything bad happens before we create a sync job
-        # (like bad scheduling config, etc)
+        # (like bad scheduling config, etc.)
         #
         # we will raise the error in the logs here and let Kibana knows
         # by toggling the status and setting the error and status field
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -643,7 +643,9 @@ class Connector:
             bulk_options = self.bulk_options.copy()
             self.data_provider.tweak_bulk_options(bulk_options)
 
-            if self.features.sync_rules_enabled():
+            sync_rules_enabled = self.features.sync_rules_enabled()
+
+            if sync_rules_enabled:
                 await validate_filtering(self, self.index, ValidationTarget.ACTIVE)
 
             result = await elastic_server.async_bulk(
@@ -651,6 +653,7 @@ class Connector:
                 self.prepare_docs(self.data_provider),
                 self.pipeline,
                 filtering=self.filtering,
+                sync_rules_enabled=sync_rules_enabled,
                 options=bulk_options,
             )
             await self._sync_done(job, result)
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -172,13 +172,18 @@ class ConnectorIndex(ESIndex):
 
 
 class SyncJob:
-    def __init__(self, connector_id, elastic_client):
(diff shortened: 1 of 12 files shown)


connectors/utils.py
```diff
@@ -352,3 +352,16 @@ def get_event_loop(uvloop=False):
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
     return loop
+
+
+def nested_get(dictionary, keys, default=None):
```
seanstory, line 357: I'm kinda amazed that this isn't part of the language already. Nice.

seanstory (approved):

connectors/utils.py
tarekziade: that's not something I've ever need, and I suspect you don't.

we would usually do:

```
self.features["sync_rules"]["basic"]["enabled"]
```

now if you want to have defaults and have that line works even for empty dicts, you would use `collections.defaultdict` https://docs.python.org/3/library/collections.html#collections.defaultdict and implement the behavior you want for defaults.

tarekziade (commented):
@timgrein look at `defaultdict` -- this is probably what you could use instead of `nested_get`

connectors/utils.py
timgrein: But this wouldn't work for arbitrary levels right?
```
>>> from collections import defaultdict
>>> features = defaultdict(lambda: defaultdict(lambda: defaultdict(lambda: 'default')))
>>> features['a']['b']['c']['d']
TypeError: string indices must be integers
```

connectors/byoei.py
```diff
@@ -196,8 +197,10 @@ def __init__(
         self.total_docs_deleted = 0
         self.fetch_error = None
         self.filtering = filtering
-        self.basic_rule_engine = BasicRuleEngine(
-            parse(filtering.get_active_filter().get("rules", []))
+        self.basic_rule_engine = (
+            BasicRuleEngine(parse(filtering.get_active_filter().get("rules", [])))
+            if sync_rules_enabled
```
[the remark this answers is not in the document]
timgrein: That would lead to a syntax error. It also felt weird to me in the first moment when coming from Ruby.

connectors/byoc.py
```diff
@@ -285,6 +285,49 @@ def __repr__(self):
...
 
 
+class Features:
+
+    BASIC_RULES_NEW = "basic_rules_new"
+    ADVANCED_RULES_NEW = "advanced_rules_new"
+
+    # keep backwards compatibility
+    BASIC_RULES_OLD = "basic_rules_old"
+    ADVANCED_RULES_OLD = "advanced_rules_old"
```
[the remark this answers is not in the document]
timgrein: AFAIK the frontend has the frontend has the same mechanism. It's easy to remove later so I would advocate for keeping it to be aligned with the frontend.

connectors/utils.py
timgrein: As agreed with @tarekziade in a 1:1 discussion we'll move nested_get to keep it locally to the `Features` class.

PR #375, opened 2023-01-25
[DO NOT REVIEW] Refactor Connector and SyncJob

wangch079:
This PR tries to refactor `connectors-python`:
1. Make `Connector` and `SyncJob` class only data representation of a doc in `.elastic-connectors`/`.elastic-connectors-sync-jobs`. It also contains some self-mutating methods, e.g. `claim` and `suspend` methods in `SyncJob`, `heartbeat` and `prepare` methods in `Connector`.
2. Make `ConnectorIndex` and `SyncJobIndex` responsible for index-level operations only, e.g. retrieve `supported_connectors` in `ConnectorIndex`, retrieve `pending_jobs` in `SyncJobIndex`.
3. Make `SyncService` only responsible for scheduling jobs, and the job execution is managed by a new class `SyncJobRunner`.

Then we realized the scope of this PR is too large, so we decided to break it down into multiple PRs:
1. #430 
2. #431 
3. #432 
4. #449
5. #440 
6. #442 
7. #443 
8. #434
9. #460 
10. #474
11. #495
12. #498
13. #491 
14. #500 

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -243,7 +243,7 @@ class SyncJob(ESDocument):
 
     @property
     def filtering(self):
-        return Filter(self.get("connector", "filtering"))
+        return Filter(self.get("connector", "filtering", default={}))
 
     @property
     def pipeline(self):
@@ -349,11 +349,11 @@ class Filtering:
     def get_filter(self, filter_state="active", domain=DEFAULT_DOMAIN):
         return next(
             (
-                filter_[filter_state]
+                Filter(filter_[filter_state])
                 for filter_ in self.filtering
                 if filter_["domain"] == domain
             ),
-            {},
+            Filter(),
         )
 
     def to_list(self):
@@ -381,6 +381,26 @@ class Filter(dict):
 
     def has_validation_state(self, validation_state):
         return FilteringValidationState(self.validation["state"]) == validation_state
+
+
+class Filter(dict):
+    def __init__(self, filter_=None):
+        if filter_ is None:
+            filter_ = {}
+
+        super().__init__(filter_)
+
+        advanced_rules = filter_.get("advanced_snippet", {})
+
+        self.advanced_rules = advanced_rules.get("value", advanced_rules)
+        self.basic_rules = filter_.get("rules", [])
+        self.validation = filter_.get("validation", {})
+
+    def get_advanced_rules(self):
+        return self.advanced_rules
+
+    def has_advanced_rules(self):
+        return len(self.advanced_rules) > 0
 
 
 PIPELINE_DEFAULT = {
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -10,6 +10,7 @@ import socket
 import time
 from collections import UserDict
 from copy import deepcopy
+from datetime import datetime, timezone
 from enum import Enum
 
 from connectors.es import ESIndex, Mappings
@@ -26,6 +27,9 @@ CONNECTORS_INDEX = ".elastic-connectors"
 JOBS_INDEX = ".elastic-connectors-sync-jobs"
 RETRY_ON_CONFLICT = 3
 SYNC_DISABLED = -1
+
+JOB_NOT_FOUND_ERROR = "Couldn't find the job"
+UNKNOWN_ERROR = "unknown error"
 
 
 class Status(Enum):
@@ -110,26 +114,23 @@ class ConnectorIndex(ESIndex):
         super().__init__(index_name=CONNECTORS_INDEX, elastic_config=elastic_config)
 
     async def heartbeat(self, doc_id):
-        doc = {
-            "last_seen": iso_utc(),
-        }
-        await self.update(doc_id, doc)
+        await self.update(doc_id=doc_id, doc={"last_seen": iso_utc()})
 
     async def update_filtering_validation(
         self, connector, validation_result, validation_target=ValidationTarget.ACTIVE
     ):
-        doc_to_update = deepcopy(connector.doc_source)
+        filtering = connector.filtering.to_dict()
 
-        for filter_ in doc_to_update.get("filtering", []):
+        for filter_ in filtering:
             if filter_.get("domain", "") == Filtering.DEFAULT_DOMAIN:
                 filter_.get(e2str(validation_target), {"validation": {}})[
                     "validation"
                 ] = validation_result.to_dict()
 
         await self.client.update(
-            index=CONNECTORS_INDEX,
+            index=self.index_name,
             id=connector.id,
-            doc=doc_to_update,
+            doc={"filtering": filtering},
             retry_on_conflict=RETRY_ON_CONFLICT,
         )
 
@@ -174,8 +175,7 @@ class ConnectorIndex(ESIndex):
     def _create_object(self, doc_source):
         return Connector(
             self,
-            doc_source["_id"],
-            doc_source["_source"],
+            doc_source,
         )
 
     async def all_connectors(self):
@@ -218,6 +218,10 @@ class SyncJob(ESDocument):
         return str2e(self.get("status"), JobStatus)
 
     @property
+    def error(self):
+        return self.get("error")
+
+    @property
     def connector_id(self):
         return self.get("connector", "id")
 
@@ -239,11 +243,11 @@ class SyncJob(ESDocument):
 
     @property
     def filtering(self):
-        return Filtering(self.get("connector", "filtering"))
+        return Filter(self.get("connector", "filtering"))
 
     @property
     def pipeline(self):
-        return PipelineSettings(self.get("connector", "pipeline"))
+        return Pipeline(self.get("connector", "pipeline"))
 
     @property
     def terminated(self):
@@ -279,9 +283,9 @@ class SyncJob(ESDocument):
             JobStatus.COMPLETED, None, ingestion_stats, connector_metadata
         )
 
-    async def error(self, message, ingestion_stats={}, connector_metadata={}):
+    async def fail(self, message, ingestion_stats={}, connector_metadata={}):
         await self.terminate(
-            JobStatus.ERROR, message, ingestion_stats, connector_metadata
+            JobStatus.ERROR, str(message), ingestion_stats, connector_metadata
         )
 
     async def cancel(self, ingestion_stats={}, connector_metadata={}):
@@ -295,7 +299,7 @@ class SyncJob(ESDocument):
         doc = {
             "last_seen": iso_utc(),
             "completed_at": iso_utc(),
-            "status": status,
+            "status": e2str(status),
             "error": error,
         }
         doc.update(ingestion_stats)
@@ -465,7 +469,8 @@ class Connector(ESDocument):
     @property
     def last_seen(self):
         last_seen = self.get("last_seen")
-        logger.info(f"last_seen = {last_seen}, type = {last_seen.__class__}")
+        if last_seen is not None:
+            last_seen = datetime.fromisoformat(last_seen)
         return last_seen
 
     @property
@@ -485,6 +490,14 @@ class Connector(ESDocument):
         return DataSourceConfiguration(self.get("configuration"))
 
     @property
+    def index_name(self):
+        return self.get("index_name")
+
+    @property
+    def language(self):
+        return self.get("language")
+
+    @property
     def filtering(self):
         return Filtering(self.get("filtering"))
 
@@ -497,8 +510,11 @@ class Connector(ESDocument):
         return Features(self.get("features"))
 
     async def heartbeat(self, interval):
-        if self.last_seen is None or time.time() - self.last_seen > interval:
-            await self.index.heartbeat()
+        if (
+            self.last_seen is None
+            or (datetime.now(timezone.utc) - self.last_seen).total_seconds() > interval
+        ):
+            await self.index.heartbeat(doc_id=self.id)
 
     def next_sync(self):
         """Returns in seconds when the next sync should happen.
@@ -514,10 +530,7 @@ class Connector(ESDocument):
         return next_run(self.scheduling.get("interval"))
 
     async def reset_sync_now_flag(self):
-        doc = {
-            "sync_now": False,
-        }
-        await self.index.update(doc_id=self.id, doc=doc)
+        await self.index.update(doc_id=self.id, doc={"sync_now": False})
 
     async def sync_starts(self):
         doc = {
@@ -535,9 +548,9 @@ class Connector(ESDocument):
 
     async def sync_done(self, job):
         job_status = JobStatus.ERROR if job is None else job.status
-        job_error = "Couldn't find the job" if job is None else job.error
+        job_error = JOB_NOT_FOUND_ERROR if job is None else job.error
         if job_error is None and job_status == JobStatus.ERROR:
-            job_error = "unknown error"
+            job_error = UNKNOWN_ERROR
         connector_status = (
             Status.ERROR if job_status == JobStatus.ERROR else Status.CONNECTED
         )
@@ -546,7 +559,7 @@ class Connector(ESDocument):
             "last_sync_status": job_status,
             "last_synced": iso_utc(),
             "last_sync_error": job_error,
-            "status": connector_status,
+            "status": e2str(connector_status),
             "error": job_error,
         }
 
@@ -602,11 +615,18 @@ class Connector(ESDocument):
 
         try:
             await self.index.update(doc_id=self.id, doc=doc)
+            connector = await self.reload()
+            self._source = connector._source
         except Exception as e:
             logger.critical(e, exc_info=True)
             raise ConnectorUpdateError(
                 f"Could not update service type/configuration for connector {self.id}"
             )
+
+    async def document_count(self):
+        await self.index.client.indices.refresh(index=self.index_name)
+        result = await self.index.client.count(index=self.index_name)
+        return result["count"]
 
 
 STUCK_JOBS_THRESHOLD = 60  # 60 seconds
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -6,8 +6,8 @@
 """
 Implementation of BYOC protocol.
 """
-import time
 import socket
+import time
 from collections import UserDict
 from copy import deepcopy
 from enum import Enum
@@ -75,8 +75,6 @@ def supported_connectors_query(native_service_types=None, connectors_ids=None):
 
     if len(native_service_types) == 0 and len(connectors_ids) == 0:
         return
-        if len(native_service_types) == 0 and len(connector_ids) == 0:
-            return
 
     native_connectors_query = {
         "bool": {
@@ -96,9 +94,7 @@ def supported_connectors_query(native_service_types=None, connectors_ids=None):
         }
     }
     if len(native_service_types) > 0 and len(connectors_ids) > 0:
-        query = {
-            "bool": {"should": [native_connectors_query, custom_connectors_query]}
-        }
+        query = {"bool": {"should": [native_connectors_query, custom_connectors_query]}}
     elif len(native_service_types) > 0:
         query = native_connectors_query
     else:
@@ -120,7 +116,7 @@ class ConnectorIndex(ESIndex):
         await self.update(doc_id, doc)
 
     async def update_filtering_validation(
-            self, connector, validation_result, validation_target=ValidationTarget.ACTIVE
+        self, connector, validation_result, validation_target=ValidationTarget.ACTIVE
     ):
         doc_to_update = deepcopy(connector.doc_source)
 
@@ -279,15 +275,23 @@ class SyncJob(ESDocument):
         await self.index.update(doc_id=self.id, doc=doc)
 
     async def done(self, ingestion_stats={}, connector_metadata={}):
-        await self.terminate(JobStatus.COMPLETED, None, ingestion_stats, connector_metadata)
+        await self.terminate(
+            JobStatus.COMPLETED, None, ingestion_stats, connector_metadata
+        )
 
     async def error(self, message, ingestion_stats={}, connector_metadata={}):
-        await self.terminate(JobStatus.ERROR, message, ingestion_stats, connector_metadata)
+        await self.terminate(
+            JobStatus.ERROR, message, ingestion_stats, connector_metadata
+        )
 
     async def cancel(self, ingestion_stats={}, connector_metadata={}):
-        await self.terminate(JobStatus.CANCELED, None, ingestion_stats, connector_metadata)
+        await self.terminate(
+            JobStatus.CANCELED, None, ingestion_stats, connector_metadata
+        )
 
-    async def terminate(self, status, error=None, ingestion_stats={}, connector_metadata={}):
+    async def terminate(
+        self, status, error=None, ingestion_stats={}, connector_metadata={}
+    ):
         doc = {
             "last_seen": iso_utc(),
             "completed_at": iso_utc(),
@@ -533,8 +537,10 @@ class Connector(ESDocument):
         job_status = JobStatus.ERROR if job is None else job.status
         job_error = "Couldn't find the job" if job is None else job.error
         if job_error is None and job_status == JobStatus.ERROR:
-            job_error = 'unknown error'
-        connector_status = Status.ERROR if job_status == JobStatus.ERROR else Status.CONNECTED
+            job_error = "unknown error"
+        connector_status = (
+            Status.ERROR if job_status == JobStatus.ERROR else Status.CONNECTED
+        )
 
         doc = {
             "last_sync_status": job_status,
@@ -570,7 +576,9 @@ class Connector(ESDocument):
                 )
                 raise ServiceTypeNotConfiguredError("Service type is not configured.")
             doc["service_type"] = configured_service_type
-            logger.debug(f"Populated service type {configured_service_type} for connector {self.id}")
+            logger.debug(
+                f"Populated service type {configured_service_type} for connector {self.id}"
+            )
 
         if self.configuration.is_empty():
             if configured_service_type not in config["sources"]:
@@ -585,7 +593,9 @@ class Connector(ESDocument):
                 logger.debug(f"Populated configuration for connector {self.id}")
             except Exception as e:
                 logger.critical(e, exc_info=True)
-                raise DataSourceError(f"Could not instantiate {fqn} for {configured_service_type}")
+                raise DataSourceError(
+                    f"Could not instantiate {fqn} for {configured_service_type}"
+                )
 
         if len(doc) == 0:
             return
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -244,12 +244,7 @@ class SyncJob(ESDocument):
 
     @property
     def terminated(self):
-        return self.status in [
-            JobStatus.ERROR,
-            JobStatus.COMPLETED,
-            JobStatus.CANCELED,
-            JobStatus.SUSPENDED,
-        ]
+        return self.status in (JobStatus.ERROR, JobStatus.COMPLETED, JobStatus.CANCELED)
 
     @property
     def indexed_document_count(self):
@@ -301,13 +296,14 @@ class SyncJob(ESDocument):
     ):
         doc = {
             "last_seen": iso_utc(),
-            "completed_at": iso_utc(),
             "status": e2str(status),
             "error": error,
         }
-        doc.update(ingestion_stats)
+        if status in (JobStatus.ERROR, JobStatus.COMPLETED, JobStatus.CANCELED):
+            doc["completed_at"] = iso_utc()
         if status == JobStatus.CANCELED:
             doc["canceled_at"] = iso_utc()
+        doc.update(ingestion_stats)
         if len(connector_metadata) > 0:
             doc["metadata"] = connector_metadata
         await self.index.update(doc_id=self.id, doc=doc)
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -6,11 +6,10 @@
 """
 Implementation of BYOC protocol.
 """
-import asyncio
 import time
+import socket
 from collections import UserDict
 from copy import deepcopy
-from datetime import datetime, timezone
 from enum import Enum
 
 from connectors.es import ESIndex, Mappings
@@ -52,19 +51,6 @@ class JobTriggerMethod(Enum):
     SCHEDULED = 2
 
 
-CUSTOM_READ_ONLY_FIELDS = (
-    "is_native",
-    "api_key_id",
-    "pipeline",
-    "scheduling",
-)
-
-NATIVE_READ_ONLY_FIELDS = CUSTOM_READ_ONLY_FIELDS + (
-    "service_type",
-    "configuration",
-)
-
-
 class ServiceTypeNotSupportedError(Exception):
     pass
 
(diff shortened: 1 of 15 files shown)


connectors/byoc.py
```diff
@@ -164,22 +167,20 @@ def build_docs_query(self, native_service_types=None, connectors_ids=None):
...
-            self,
-            doc_source["_id"],
-            doc_source["_source"],
+            doc_source,
             bulk_options=self.bulk_options,
         )
 
 
 class SyncJob:
-    def __init__(self, elastic_index, connector_id, doc_source=None):
```
wangch079, line 175: Before it's getting too complicated, I would like to get the team's opinion on this. 

Currently each entity instance (of class `Connector`/`SyncJob`) holds an instance variable `index` (instance of `ConnectorIndex`/`SyncJobIndex`, which is the child class of `ESClient`), so all API calls to Elasticsearch are through this variable).

I want to introduce some static methods, e.g. `pending_jobs` in `SyncJob`, so I can get all pending jobs by calling `SyncJob.pending_jobs()`. But this won't work as the class doesn't hold the index instance. That's why I introduced a class variable (same in class `Connector`), and they should be set in `cli.py`.

cc @elastic/ingestion-team

connectors/cli.py
```diff
@@ -140,10 +141,16 @@ def run(args):
...
+        Connector.es_index = ConnectorIndex(config["elasticsearch"])
+        SyncJob.es_index = SyncJobIndex(config["elasticsearch"])
         return loop.run_until_complete(coro)
     except asyncio.CancelledError:
         return 0
     finally:
+        if Connector.es_index is not None:
+            loop.run_until_complete(Connector.es_index.close())
+        if SyncJob.es_index is not None:
+            loop.run_until_complete(SyncJob.es_index.close())
```
wangch079, line 153: Not really sure about this. cc @tarekziade

connectors/byoc.py
vidok: Should `pending_jobs ` be a part of `SyncJobIndex`? `SyncJobIndex` has the ability to query jobs and then create a list of `SyncJob` objects based on the elasticsearch documents (see [_create_object](https://github.com/elastic/connectors-python/blob/main/connectors/byoc.py#L702).

I think `SyncJobIndex` should be responsible for any sync job lookups and `SyncJob` should represent a single document.

wangch079: OK. That's different from my understanding, and the Ruby implementation.

So you are suggesting, the index class (`ConnectorIndex` and `SyncJobIndex`) should be responsible for any business logic related read/write. The entity class (`Connector` and `SyncJob`) only represent a single document, and is responsible for mutating itself, e.g. `job.claim()`, `job.cancel()`? 


Or do you think we could make it a simple data object, and move those mutating methods to the index class as well? This could simply it significantly as it doesn't have to hold an index instance.

connectors/byoc.py
```diff
@@ -125,7 +125,10 @@ async def update_filtering_validation(
             retry_on_conflict=RETRY_ON_CONFLICT,
         )
 
-    def build_docs_query(self, native_service_types=None, connectors_ids=None):
+    @staticmethod
```
tarekziade, line 128: why a staticmethod? if it's not using any of the class variables, make this a seperate function

connectors/byoc.py
```diff
@@ -164,22 +167,20 @@ def build_docs_query(self, native_service_types=None, connectors_ids=None):
...
-            doc_source["_id"],
-            doc_source["_source"],
+            doc_source,
             bulk_options=self.bulk_options,
         )
 
 
 class SyncJob:
-    def __init__(self, elastic_index, connector_id, doc_source=None):
+    es_index = None
```
tarekziade, line 176: don't use class variables, put this in `__init__`

connectors/byoc.py
vidok: > So you are suggesting, the index class (ConnectorIndex and SyncJobIndex) should be responsible for any business logic related read/write. 

Yes, logic like pulling documents and create object iterators, pagination. Everything we need to work with Elasticsech indices. 

an abstract code might look like: 

```
for job in sync_job_index.pending_jobs():
  job.claim()
  # execution
  job.finish()
```

> The entity class (Connector and SyncJob) only represent a single document, and is responsible for mutating itself, e.g. job.claim(), job.cancel()?

Yes!

> Or do you think we could make it a simple data object, and move those mutating methods to the index class as well? This could simply it significantly as it doesn't have to hold an index instance.

I don't think so. I think it's pretty good to access the ES client from the entity classes but we can choose another way to propagate ES connection. Now it's an es_index object which is not the ideal way but this can be easily improved because I encapsulated object creation.

tarekziade: in your refactorings, make sure you initialize stuff in `__init__` and not add class variables (as opposed to object variables) -- there are very few cases where we want this.
using class variables means you share the same object across all instances, this adds edge issues

connectors/byoc.py
```diff
@@ -232,10 +233,10 @@ async def done(self, indexed_count=0, deleted_count=0, exception=None):
 
         job_def["status"] = e2str(self.status)
 
-        return await self.client.update(index=JOBS_INDEX, id=self.job_id, doc=job_def)
+        return await self.es_index.client.update(index=JOBS_INDEX, id=self.job_id, doc=job_def)
 
-    @classmethod
-    def transform_filtering(cls, filtering):
+    @staticmethod
```
tarekziade, line 238: don't use static methods

connectors/byoc.py
```diff
@@ -707,12 +705,12 @@ def _create_object(self, doc_source):
             SyncJob
         """
         return SyncJob(
-            self,
             connector_id=doc_source["_source"]["connector"]["id"],
             doc_source=doc_source["_source"],
         )
 
-    def pending_job_query(self, connectors_ids):
+    @staticmethod
```
tarekziade, line 712: same remark

tarekziade (commented):
This looks good for a first draft, awesome. 

About `staticmethod` vs `classmethod` vs regular methods: you should always use regular methods -- when you use static or class methods you are using classes as a namespace, not OOP elements anymore. There are a few spots in filtering rules where it's ok, but here I don't see a use case

connectors/cli.py
```diff
@@ -140,10 +141,16 @@ def run(args):
     coro = _start_service(config, args, loop)
 
     try:
+        Connector.es_index = ConnectorIndex(config["elasticsearch"])
```
tarekziade, line 144: There's a small design issue. You want to create your indexes and pass them to the Services you create, and deal with their shutdown there.  The `cli` module should only deal with `Service` classes.

connectors/byoc.py
wangch079: Thanks for the comments.

Seems the team is not in favor of class variables. I will make the change accordingly.

tarekziade: yeah class variables break the class contract because it becomes a namespace and loses its OOP behaviors. It's very helpful in met-programming but in other cases, not. In Python it's completely fine also to use plain functions when a method would be static. It's cleaner

connectors/services/sync.py
```diff
@@ -64,32 +71,61 @@ async def _one_sync(self, connector, es, sync_now):
...
+                return
+
+            if connector.features.sync_rules_enabled():
+                await validate_filtering(
+                    connector, self.connectorIndex, ValidationTarget.DRAFT
+                )
+
+            if not sync_now:
+                next_sync = connector.next_sync()
+                if next_sync == SYNC_DISABLED or next_sync - self.idling > 0:
```
vidok, line 91: 2 things here: 

* Why do need the second condition against `SYNC_DISABLED`?
* I would move this logic (the entire block from L89 to L100  into a separate method

connectors/source.py
```diff
@@ -235,3 +235,11 @@ def get_data_sources(config):
     """Returns an iterator of all registered sources."""
     for name, fqn in config["sources"].items():
         yield get_source_klass(fqn)
+
+
+def get_data_source_dict(config):
```
vidok, line 240: I would not use data type in the functions` names. Does it technically return data source configuration?

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,166 @@
...
+    async def execute(self):
+        if self.running:
+            raise SyncJobRunningError(f"Sync job #{self.job_id} is already running.")
+
+        self.running = True
+        if not await self._claim_job():
+            logger.error(f"Unable to claim job #{self.job_id}")
+            return
+
+        try:
```
vidok, line 53: It's super hard to read such a long block of code. Is there a way to split it into several ones?

connectors/byoc.py
```diff
@@ -81,31 +67,53 @@ class DataSourceError(Exception):
...
-            index=CONNECTORS_INDEX,
-            id=connector.id,
-            doc=document,
-        )
+
+    async def heartbeat(self, doc_id):
+        doc = {
+            "last_seen": iso_utc(),
+        }
+        await self.update(doc_id, doc)
```
tarekziade, line 116: one line works for such a simple change:

```
await self.update(doc_id, {"last_seen": iso_utc()})
```

connectors/byoc.py
```diff
@@ -125,114 +133,144 @@ async def update_filtering_validation(
...
             doc_source = dict()
+        self.id = doc_source.get("_id")
+        self._source = doc_source.get("_source", {})
+
+    def get(self, *keys, default=None):
+        value = self._source
+        for key in keys:
+            if not isinstance(value, dict):
+                return default
+            value = value.get(key)
```
tarekziade, line 166: if you have several keys, `value` is overridden on the second round of this loop. is this intended?

tarekziade: Overriding the variable `value` also means you won't have `self._source` in it.

connectors/byoc.py
```diff
@@ -278,6 +316,8 @@ def get_filter(self, filter_state="active", domain=DEFAULT_DOMAIN):
 
 class PipelineSettings:
     def __init__(self, pipeline):
```
tarekziade, line 318: `pipeline=None` ?

connectors/byoc.py
```diff
@@ -347,131 +387,52 @@ def nested_get(dictionary, keys_, default_=None):
...
-    @status.setter
-    def status(self, value):
-        if isinstance(value, str):
-            value = Status[value.upper()]
-        if not isinstance(value, Status):
-            raise TypeError(value)
+    @property
+    def last_seen(self):
+        last_seen = self.get("last_seen")
+        logger.info(f"last_seen = {last_seen}, type = {last_seen.__class__}")
```
tarekziade, line 402: shouldn't this be `debug` ?

connectors/byoc.py
```diff
@@ -481,211 +442,168 @@ def next_sync(self):
...
-
-        self.sync_now = self.doc_source["sync_now"] = False
-        self.doc_source["last_sync_status"] = e2str(job.status)
-        self.status = Status.CONNECTED
-        await self.sync_doc()
+    async def reset_sync_now_flag(self):
+        doc = {
+            "sync_now": False,
+        }
+        await self.index.update(doc_id=self.id, doc=doc)
```
tarekziade, line 454: can be a one liner

connectors/byoc.py
```diff
@@ -481,211 +442,168 @@ def next_sync(self):
...
+            "status": e2str(Status.ERROR),
+            "error": str(error),
+        }
+        await self.index.update(doc_id=self.id, doc=doc)
+
+    async def sync_done(self, job):
+        job_status = JobStatus.ERROR if job is None else job.status
+        job_error = "Couldn't find the job" if job is None else job.error
+        if job_error is None and job_status == JobStatus.ERROR:
+            job_error = "unknown error"
```
tarekziade, line 474: let's make this a CONST at the top of the file

connectors/services/sync.py
```diff
@@ -64,32 +71,61 @@ async def _one_sync(self, connector, es, sync_now):
...
         except InvalidFilteringError as e:
             logger.error(e)
-            return
-        finally:
-            await connector.close()
 
     async def _run(self):
         """Main event loop."""
-        self.connectors = ConnectorIndex(self.es_config)
+        self.connectorIndex = ConnectorIndex(self.es_config)
```
tarekziade, line 127: use PEP 8

```
self.connector_index
self.sync_job_index
```

tarekziade (commented):
neat, lots of cool stuff

connectors/es/index.py
```diff
@@ -40,7 +40,32 @@ def _create_object(self, doc):
...
+        return self._create_object(resp)
+
+    async def index(self, doc):
+        resp = await self.client.index(index=self.index_name, document=doc)
+        return resp["_id"]
+
+    async def update(self, doc_id, doc):
+        await self.client.update(index=self.index_name, id=doc_id, doc=doc)
+
+    async def document_count(self):
```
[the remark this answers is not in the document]
wangch079: All the stats fields in `.elastic-connectors-sync-jobs` are singular, e.g. `indexed_document_count`, `total_document_count`. Just want to be consistent.

connectors/byoc.py
wangch079: To instantiate an instance of `PipelineSettings`, it's expected to provide a `pipeline`, so I am inclined to make it mandatory. But we couldn't stop someone using `pipeline = PipelineSettings(pipeline=None)`, therefore I added additional protection:
```python
if pipeline is None:
    pipeline = {}
```

connectors/byoc.py
wangch079: This log shouldn't be here. I added this for my own testing,  because I'm not sure whether the `last_seen` will be parsed as string or time in the response json. I will remove it eventually.

connectors/services/sync.py
wangch079: I think the inner if-else branch is to log different messages. I just copied this block from somewhere else. 

Will furthur refactor it.

connectors/source.py
wangch079: No, it returns a service type - source class dictionary. I just followed the pattern of the above method `get_data_sources`, and now I reallized it was named wrongly. Probably we should rename the above method to `get_source_klasses` and this method to `get_source_klass_dict`

connectors/byoc.py
wangch079: Yes, it's intended. 

> Overriding the variable `value` also means you won't have `self._source` in it.

Yes, we only need `value` pointing to `self._source` in the beginning, then we want to dig with `keys` provided, it's like the [dig method in Ruby Hash](https://ruby-doc.org/core-3.1.1/Hash.html#method-i-dig)

tarekziade: ok I see 👍

connectors/byoc.py
tarekziade: yes but this protection is just for None. This would break too:

```
pipeline = PipelineSettings(pipeline=1)
```

If pipeline is mandatory, you have to trust it's a pipeline. If there's no way to trust it, you can do:

```
if not isinstance(pipeline, dict):
    raise TypeError(pipeline)
```

but usually, we just trust it -- and you will get an error on the first attempt to use a dict method ("if it quacks like a duck, it's a duck")

connectors/source.py
vidok: What is `source class dictionary`?

connectors/es/index.py
```diff
@@ -40,7 +40,32 @@ def _create_object(self, doc):
         """
         raise NotImplementedError
 
-    async def get_all_docs(self, query=None, page_size=DEFAULT_PAGE_SIZE):
+    async def fetch_by_id(self, doc_id):
+        await self.client.indices.refresh(index=self.index_name)
```
artem-shelkovnikov, line 44: I'm curious about it - should we always call it or should we never call it?

I _think_ we do refresh stuff automatically, but also awaiting a `refresh` does not guarantee that the data is refreshed, only that the `refresh` operation was issued and accepted, is it correct?

connectors/es/index.py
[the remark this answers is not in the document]
artem-shelkovnikov: I also think it's singular - think of phrase like "word count" or "head count". I also googled a bit and saw similar examples, but haven't been able to find the actual rule:

https://english.stackexchange.com/questions/349475/singular-or-plural-noun-preceding-count
https://www.usingenglish.com/forum/threads/plural-or-singular-when-using-with-the-word-count.242655/

connectors/services/sync.py
artem-shelkovnikov: It can be a big refactoring overall (given we want to test it fully, I think there are missing tests for scheduling), so can probably be done in a separate PR

connectors/source.py
artem-shelkovnikov: I see it's just a "string" -> "string" dictionary, key is service_type, value is class FQN (see config.yml's "sources" section: https://github.com/elastic/connectors-python/blob/main/config.yml#L38)

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,166 @@
...
+            await self.connector.sync_done(self.sync_job)
+
+            logger.info(
+                f"Sync done: {indexed_count} indexed, {doc_deleted} "
+                f" deleted. ({int(time.time() - self._start_time)} seconds)"
+            )
+
+    async def _claim_job(self):
+        try:
+            await self.sync_job.claim()
```
artem-shelkovnikov, line 136: Hmmm, interesting, so with this PR our connectors will be more multi-tenant friendly?

connectors/byoc.py
wangch079: For this case, I think we do need `pipeline` in the `__init__` method. so I think we can keep it as 
```python
def __init__(self, pipeline):
```

Do you think it's unnecessary to have this?
```python
if pipeline is None:
    pipeline = {}
```

connectors/source.py
wangch079: It's not a `string` (service type) -> `string` (FQN) dictionary. It's a `string` (service type) -> `class` (child classes of `BaseDataSource`). 

It serve exactly the same purpose as the [REGISTRY](https://github.com/elastic/connectors-ruby/blob/9d5be7a79ee18c794403d14b1dcfc2c574786771/lib/connectors/registry.rb#L40) of `connectors-ruby`

connectors/es/index.py
wangch079: The `refresh` call is synchronous, so it's guaranteed that we get refreshed data.

For most of the use cases (if not all), we use `search`. When we use `fetch_by_id`, it normally means we need to refresh the document, because there will likely be an update happened. That's why we need the refresh.

Yes, ES does auto-refresh every second by default, and the official guide suggests not to use the refresh API, but [use `refresh=wait_for` when updating](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html#refresh-api-desc).

[`refresh=wait_for`](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html) will wait for a periodic refresh to happen before returning. If we apply this option to our update request, it may take, the worst case 1 second, to get the response. cc @tarekziade

connectors/sync_job_runner.py
wangch079: But we don't have concurrency control at all.

connectors/source.py
```diff
@@ -103,6 +105,9 @@ def get_fields(self):
     def is_empty(self):
         return len(self._config) == 0
 
+    def to_dict(self):
+        return deepcopy(self._raw_config)
```
tarekziade, line 109: `deepcopy` is overkill, even `copy` as long as the config is mappings of mappings -- you can use `dict`

```
>>> o = {'a': {'b': 'c'}}
>>> dict(o)
{'a': {'b': 'c'}}
```

connectors/source.py
```diff
@@ -272,7 +277,15 @@ def get_source_klass(fqn):
     return getattr(module, klass_name)
 
 
-def get_data_sources(config):
+def get_source_klasses(config):
```
tarekziade, line 280: better name!

connectors/source.py
```diff
@@ -272,7 +277,15 @@ def get_source_klass(fqn):
...
     """Returns an iterator of all registered sources."""
     for name, fqn in config["sources"].items():
         yield get_source_klass(fqn)
+
+
+def get_source_klass_dict(config):
```
tarekziade, line 286: makes me wonder if we don't want to add a `name` property to the source klass...

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,167 @@
...
+            if not await data_provider.changed():
+                logger.debug(
+                    f"No change in {self.sync_job.service_type} data provider, skipping..."
+                )
+                return
+
+            logger.debug(f"Syncing '{self.sync_job.service_type}'")
+            logger.debug(f"Pinging the {self.source_klass} backend")
+            await data_provider.ping()
+            await asyncio.sleep(0)
```
tarekziade, line 67: you can remove this one because you have another await right after.

```
await asyncio.sleep(0)
```

is useful in a block of sync calls that loops, to return execution to the event loop, and that happens on any await call

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,167 @@
...
+
+            mappings = Mappings.default_text_fields_mappings(
+                is_connectors_index=True,
+            )
+
+            logger.debug("Preparing the content index")
+            await self.elastic_server.prepare_content_index(
+                self.sync_job.index_name, mappings=mappings
+            )
+            await asyncio.sleep(0)
```
tarekziade, line 77: same remark

connectors/source.py
wangch079: We should. we actually have `name` property in Ruby.

Currently the `list` command uses `__doc__` of each source klass:

https://github.com/elastic/connectors-python/blob/c1baa3a0bdd74a173bedb545dad01df276e3faef/connectors/cli.py#L135


I can make the change in a separate PR.

But even we have `name` and `service_type` in each source klass, we still need a dict to quickly loop up the source klass

wangch079: PR is here: https://github.com/elastic/connectors-python/pull/413

connectors/byoc.py
```diff
@@ -81,180 +70,209 @@ class DataSourceError(Exception):
...
+        return self.get("total_document_count", default=0)
+
+    async def claim(self):
+        doc = {
+            "status": e2str(JobStatus.IN_PROGRESS),
+            "started_at": iso_utc(),
+            "last_seen": iso_utc(),
+            "worker_hostname": socket.gethostname(),
         }
+        await self.index.update(doc_id=self.id, doc=doc)
```
artem-shelkovnikov, line 244: No concurrency handling here yet, right?

connectors/byoc.py
```diff
@@ -297,6 +315,9 @@ def get_filter(self, filter_state="active", domain=DEFAULT_DOMAIN):
             Filter(),
         )
 
+    def to_list(self):
+        return list(self.filtering)
+
```
artem-shelkovnikov, line 320: Looks redundant, maybe better to get rid of it?

connectors/byoc.py
```diff
@@ -320,17 +341,35 @@ def has_advanced_rules(self):
...
 
     def __repr__(self):
         return (
             f"Pipeline {self.name} <binary: {self.extract_binary_content}, "
             f"whitespace {self.reduce_whitespace}, ml inference {self.run_ml_inference}>"
         )
 
+    def to_dict(self):
+        return dict(self._pipeline)
+
```
artem-shelkovnikov, line 372: I think it can be replaced with `__dict__` function and calling `dict` on instance of PipelineSettings should call it. cc @tarekziade is it correct?

connectors/byoc.py
artem-shelkovnikov: Ah, it's probably better to implement Python's Abstract Base Class that allows certain operations over filtering - maybe just `__iter__` and `__get_item__` and `__len__` will be sufficient - see https://docs.python.org/3/library/collections.abc.html

connectors/byoc.py
```diff
@@ -529,217 +501,110 @@ def next_sync(self):
         if self.sync_now:
             logger.debug("sync_now is true, syncing!")
             return 0
-        if not self.scheduling["enabled"]:
+        if not self.scheduling.get("enabled", False):
```
artem-shelkovnikov, line 504: Nice!

connectors/services/sync.py
```diff
@@ -40,63 +45,68 @@ def __init__(self, config, args):
...
-            await connector.error(e)
-            logger.critical(e, exc_info=True)
-            self.raise_if_spurious(e)
+        if not await self._prepare_connector(connector):
             return
 
         try:
             # the heartbeat is always triggered
-            connector.start_heartbeat(self.hb)
+            await connector.heartbeat(self.heartbeat_interval)
```
artem-shelkovnikov, line 66: I'd probably leave `heartbeat_interval` out of the function and do the condition here, e.g:

```python
if self.should_trigger_heartbeat():
    connector.heartbeat()
```

artem-shelkovnikov: But it's more of a nit fixable later

artem-shelkovnikov (approved):
I'm hesitant about the change purely because of its size - I intuitively understand that it extracts running sync jobs to a separate class, but also a lot of small side-changes are making it harder to understand in the sense of whether new problems are introduced or not.

I'll carefully say that if we want to merge this change, we should do it ASAP - it's been around for 2 weeks.

So unless somebody has blocking concerns (cc @tarekziade), let's actually land it asap and fix bugs on the way.

connectors/byoc.py
wangch079: No, do we have concurrency control in `connectors-python`? I think it will be handled separately?

connectors/byoc.py
wangch079: The reason why I added `to_list` to class `Filtering`, is because it was used to update the connector, where a list is required instead of an object:

https://github.com/elastic/connectors-python/blob/59d90d42ba9c01163c0455227d3637ef805d7028/connectors/byoc.py#L118-L134

Similar reason for `to_dict` of class `PipelineSettings`:

https://github.com/elastic/connectors-python/blob/59d90d42ba9c01163c0455227d3637ef805d7028/connectors/byoc.py#L649

I don't intend to make `Filtering` or `PipelineSettings` a collection.

connectors/byoc.py
artem-shelkovnikov: Let's do separately, yes. Just checking that you're not addressing it here.

connectors/byoc.py
wangch079: I don't think `__dict__` is for this purpose ([ref](https://docs.python.org/3/library/stdtypes.html#object.__dict__)):
>A dictionary or other mapping object used to store an object’s (writable) attributes.

artem-shelkovnikov: Sorry my bad. What I've meant is that if it's used as hash, then the class might actually expose dictionary functions so that it could be done outside of the object, e.g:

pipeline_dict = dict(Pipeline) or something else.

In short term this code works, but I believe we can improve on it later - in later PRs.

wangch079:
>I'm hesitant about the change purely because of its size

I agree. I'm not a fan of huge PR either. I tried to keep it small when I worked on this PR but it kept growing and seemed out of control. 

What I have done for this PR:
1. Tested locally for both custom connector and native connector.
2. Added tests and make sure we have > 92% test coverage.

>let's actually land it asap and fix bugs on the way.

I'd say yes. Having this merged will speed up the whole team, not only myself. With the above 2 points, it won't break `main`. There may still be bugs, which is more easier to spot if the whole team starts working on `main` branch after it's merged, instead of only me.

wangch079:
Hi @artem-shelkovnikov , @tarekziade , I implemented the changes of https://github.com/elastic/connectors-python/pull/376 in this commit https://github.com/elastic/connectors-python/pull/375/commits/b306deca18e514e35fe39dd21b7f23b4343df46f.

When I try to shut down the service during job execution, I can see it reached this line:

https://github.com/elastic/connectors-python/blob/b306deca18e514e35fe39dd21b7f23b4343df46f/connectors/sync_job_runner.py#L97-L98,

and it also reached the `finally` block:

https://github.com/elastic/connectors-python/blob/b306deca18e514e35fe39dd21b7f23b4343df46f/connectors/sync_job_runner.py#L103

but it didn't proceed further, so the job stays at `in_progress` status. it seems it won't wait for the `finally` block to complete.

Do you spot anything wrong in my implementation? Else I will continue tomorrow.

tarekziade (commented):
Following up on the discussion in https://elastic.slack.com/archives/C01795T48LQ/p1675759094917429

The patch is doing a lot of changes in the core that could probably happen in smaller patches, and at this point, I think it's a bit risky to land as-is for 8.7 FF.

I would advise explaining in the issue the goal of this refactoring and if possible splitting some of the work in some other PRs. I've identified several changes that can land with no risks in smaller PRs.

I would also want to see a couple of e2e test runs, with the memory and speed metrics before and after.

tarekziade:
>  let's actually land it asap and fix bugs on the way.

sorry, this is not acceptable. we don't break main and then fix it

tarekziade:
> > let's actually land it asap and fix bugs on the way.
> 
> I'd say yes. Having this merged will speed up the whole team, not only myself. With the above 2 points, it won't break `main`. There may still be bugs, which is more easier to spot if the whole team starts working on `main` branch after it's merged, instead of only me.

I don't understand. If your PR has bugs and you know about it, and if it does not "break main" it means you miss some tests.
and you can't push that on the whole team. It's the dev responsibility to provide a patch that is tested with no known bugs and tested

wangch079:
>I don't understand. If your PR has bugs and you know about it

What I meant here is that I have tested it locally and it worked, and I have added all the tests. But there are still possibly some bugs that I didn't spot.

Therefore:
>There may still be bugs, which is more easier to spot if the whole team starts working on main branch after it's merged, instead of only me.

tarekziade (changes requested):
see previous comment

tarekziade:
> > There may still be bugs, which is more easier to spot if the whole team starts working on main branch after it's merged, instead of only me.

Can you elaborate in the issue of what was the goal of this patch? there's do description. 
I am looking for a problem, how you fix it and a list of changes to do so.

The problem I see is that you added unrelated changes in the past few days that were not initially planned, growing the patch to a bigger one, so I'd like to understand what was the initial intent in this refactoring to try to minimize the risks of landing it.

tarekziade:
nightly build for this branch https://buildkite.com/elastic/connectors-python-nightly/builds/302

tarekziade:
the patch break the connectors in the e2e test and also blocks on the cleanup service.

Screenshot 2023-02-07 at 20 22 31

tarekziade:
to reproduce locally

```
make ftest NAME=mysql
```

wangch079:
> the patch break the connectors in the e2e test and also blocks on the cleanup service.

That is because the `last_seen` is incorrectly formatted:

https://github.com/elastic/connectors-python/blob/8027897416cd3b81182cb1e3cef2c70e14eedc8c/connectors/kibana.py#L155

Will introduce a separate PR to fix it. Will also try to break this PR to multiple PRs as small as possible.

tarekziade:
> Will also try to break this PR to multiple PRs as small as possible.

Thanks, happy to help identify them, once we have a description of the "MVP" here (Minimum Viable Patch :) )

tarekziade:
> That is because the `last_seen` is incorrectly formatted:

but main is not broken for this e2e test, so it has something to do with this patch I would assume

wangch079:
>but main is not broken for this e2e test, so it has something to do with this patch I would assume

Yes, here:

https://github.com/elastic/connectors-python/blob/d19c7d923438a6550936e584c8557d9d5820f3a4/connectors/byoc.py#L447-L452

connectors/filtering/validation.py
```diff
@@ -18,19 +18,19 @@ class ValidationTarget(Enum):
 
 
 async def validate_filtering(
-    connector, index, validation_target=ValidationTarget.ACTIVE
+    connector, source_klass, validation_target=ValidationTarget.ACTIVE
```
artem-shelkovnikov, line 21: This change (whole file change) can probably be in a separate PR.

connectors/services/job_cleanup.py
```diff
@@ -65,7 +65,7 @@ async def _process_orphaned_jobs(self):
                     and job.index_name not in existing_content_indices
                 ):
                     content_indices.add(job.index_name)
-                job_ids.append(job.job_id)
+                job_ids.append(job.id)
```
artem-shelkovnikov, line 68: This rename can be in a separate PR

artem-shelkovnikov: Probably whole file can be updated in a separate PR

connectors/services/sync.py
```diff
@@ -150,8 +161,50 @@ async def _run(self):
...
+            if self.connector_index is not None:
+                self.connector_index.stop_waiting()
+                await self.connector_index.close()
+            if self.sync_job_index is not None:
+                self.sync_job_index.stop_waiting()
+                await self.sync_job_index.close()
             await es.close()
         return 0
+
+    async def _prepare_connector(self, connector):
```
artem-shelkovnikov, line 173: This change and all below can probably be in a separate PR

tarekziade:
> Hi @artem-shelkovnikov , sorry I forgot to tag this PR as DO NOT REVIEW.

maybe you could convert it to draft ?

wangch079:
Close this PR as it's covered by

>1. #430 
>2. #431 
>3. #432 
>4. #449
>5. #440 
>6. #442 
>7. #443 
>8. #434
>9. #460 
>10. #474
>11. #495
>12. #498
>13. #491 
>14. #500

PR #376, opened 2023-01-25
Mark jobs that were running during service termination as Suspended

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/3567

This PR makes sure that several things happen when service is being terminated/restarted:

1. Jobs are interrupted immediately instead of waiting of it to finish
2. When jobs are interrupted, they are marked as "Suspended" so that other logic that restarts the job can kick-in

#### Pre-Review Checklist

## Related Pull Requests

- [ ] https://github.com/elastic/connectors-python/pull/382

## Release Note

When service is interrupted, it suspends the jobs immediately so that it can be restarted later when the service is back up.

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -256,6 +256,12 @@ class SyncJob:
 
         return await self.client.update(index=JOBS_INDEX, id=self.job_id, doc=job_def)
 
+    async def suspend(self):
+        self.status = JobStatus.SUSPENDED
+        job_def = {"status": e2str(self.status)}
+
+        return await self.client.update(index=JOBS_INDEX, id=self.job_id, doc=job_def)
+
     @classmethod
     def transform_filtering(cls, filtering):
         # deepcopy to not change the reference resulting in changing .elastic-connectors filtering
@@ -423,6 +429,7 @@ class Connector:
         self.bulk_options = bulk_options
         self.source_klass = None
         self.data_provider = None
+        self._sync_task = None
 
     def _update_config(self, doc_source):
         self.status = doc_source["status"]
@@ -448,6 +455,14 @@ class Connector:
     @property
     def status(self):
         return self._status
+
+    @property
+    def last_sync_status(self):
+        status = self.doc_source["last_sync_status"]
+        if status is None:
+            return None
+        value = JobStatus[status.upper()]
+        return value
 
     @status.setter
     def status(self, value):
@@ -489,6 +504,13 @@ class Connector:
         )
         self.doc_source["status"] = e2str(status)
         self._update_config(self.doc_source)
+
+    async def suspend(self):
+        if self._sync_task is not None:
+            task = self._sync_task
+            cancelled = task.cancel()
+            await task
+        await self.close()
 
     async def close(self):
         self._closed = True
@@ -552,6 +574,11 @@ class Connector:
 
     async def error(self, error):
         self.doc_source["error"] = str(error)
+        await self.sync_doc()
+
+    async def _sync_suspended(self, job):
+        await job.suspend()
+        self.doc_source["last_sync_status"] = e2str(job.status)
         await self.sync_doc()
 
     async def _sync_done(self, job, result, exception=None):
@@ -655,29 +682,25 @@ class Connector:
 
         try:
             service_type = self.service_type
-            # if we don't sync, we still want to make sure we tell kibana we are connected
-            # if the status is different from comnected
-            if self.status != Status.CONNECTED:
-                self.status = Status.CONNECTED
-                await self.sync_doc()
-
             if sync_now:
                 self.sync_now = True
                 logger.info("Sync forced")
+            elif self.last_sync_status == JobStatus.SUSPENDED:
+                logger.info("Restarting sync after suspension")
             else:
                 next_sync = self.next_sync()
-                # First we check if sync is disabled, and it terminates all other conditions
-                if next_sync == SYNC_DISABLED:
-                    logger.debug(f"Scheduling is disabled for {service_type}")
-                    return
-                # Then we check if we need to restart SUSPENDED job
-                elif self.last_sync_status == JobStatus.SUSPENDED:
-                    logger.info("Restarting sync after suspension")
-                # And only then we check if we need to run sync right now or not
-                elif next_sync - idling > 0:
-                    logger.debug(
-                        f"Next sync for {service_type} due in {int(next_sync)} seconds"
-                    )
+                if next_sync == SYNC_DISABLED or next_sync - idling > 0:
+                    if next_sync == SYNC_DISABLED:
+                        logger.debug(f"Scheduling is disabled for {service_type}")
+                    else:
+                        logger.debug(
+                            f"Next sync for {service_type} due in {int(next_sync)} seconds"
+                        )
+                    # if we don't sync, we still want to make sure we tell kibana we are connected
+                    # if the status is different from comnected
+                    if self.status != Status.CONNECTED:
+                        self.status = Status.CONNECTED
+                        await self.sync_doc()
                     return
 
             try:
@@ -699,6 +722,7 @@ class Connector:
 
         logger.debug(f"Syncing '{service_type}'")
         self._syncing = True
+        self._sync_task = asyncio.current_task()
         job = await self._sync_starts()
         try:
             logger.debug(f"Pinging the {self.data_provider} backend")
@@ -733,13 +757,15 @@ class Connector:
                 options=bulk_options,
             )
             await self._sync_done(job, result)
-
+        except asyncio.CancelledError:
+            await self._sync_suspended(job)
         except Exception as e:
             await self._sync_done(job, {}, exception=e)
             raise
         finally:
             self._syncing = False
             self._start_time = None
+            self._sync_task = None
 
 
 STUCK_JOBS_THRESHOLD = 60  # 60 seconds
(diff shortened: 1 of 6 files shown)


connectors/cli.py
```diff
@@ -102,13 +102,14 @@ async def _start_service(config, args, loop):
...
+        async def _shutdown(sig_name):
             logger.info(f"Caught {sig_name}. Graceful shutdown.")
             for service in services:
                 logger.info(f"Shutdown {service.__class__.__name__}...")
-                service.stop()
+                await service.stop()
+            return
 
-        loop.add_signal_handler(sig, functools.partial(_shutdown, sig.name))
+        loop.add_signal_handler(sig, lambda: asyncio.ensure_future(_shutdown(sig.name)))
```
tarekziade, line 112: why do you need to do this?

artem-shelkovnikov: `_shutdown` is made async in this change, thus it's not usable by `add_signal_handler` as it was before.

I made it work with the current code, but not sure if it's the best way to do so (if you revert this change you'll see tests break).

tarekziade: we can't start a future in this signal handler it needs to block (and you are not waiting for it so that's a problem)
let's sync about this

artem-shelkovnikov: I took this code from StackOverflow: https://stackoverflow.com/questions/23313720/asyncio-how-can-coroutines-be-used-in-signal-handlers

I don't have enough understanding yet about it, of course

tarekziade: Do this instead:

```
tasks = []
    def _shutdown(sig_name):
        for service in services:
            tasks.append(asyncio.create_task(service.stop()))
    loop.add_signal_handler(sig, functools.partial(_shutdown(sig.name)))

for service in services:
    tasks.append(_run_service(service))  

await asyncio.gather(*tasks)
```

connectors/byoc.py
```diff
@@ -237,6 +237,12 @@ async def done(self, indexed_count=0, deleted_count=0, exception=None):
 
         return await self.client.update(index=JOBS_INDEX, id=self.job_id, doc=job_def)
 
+    async def suspend(self):
+        self.status = JobStatus.SUSPENDED
+        job_def = {"status": e2str(self.status)}
```
wangch079, line 242: I'm surprised to see `suspended_at` is not in [the protocol](https://github.com/elastic/connectors-ruby/blob/main/docs/CONNECTOR_PROTOCOL.md). But looking at the [Monitoring project brief](https://docs.google.com/document/d/1mqPTc3Agg7FaUPVZ0YR4_R3M552HjqFwAL1UJvi57Ek/edit), it seems this is not required:

> timestamps when
> - job requested/pending
> - job started
> - job completed
> - job canceling
> - job canceled / aborted



Hi @danajuratoni , do we need the timestamp when the job is suspended?

connectors/byoc.py
```diff
@@ -529,6 +543,12 @@ async def error(self, error):
         self.doc_source["error"] = str(error)
         await self.sync_doc()
 
+    async def _sync_suspended(self, job):
+        await job.suspend()
+        self.doc_source["last_sync_status"] = e2str(job.status)
```
wangch079, line 548: Considering updating `last_synced` and `last_synced_error`

connectors/cli.py
wangch079: Will it be easier if we assign this to `self.sync_job_runner`

https://github.com/elastic/connectors-python/blob/3e5c01202539bc8a3eac79f74d1b107242f9ae24/connectors/services/sync.py#L86

And in `stop()` method of `SyncService`, we add
```python
    if self.sync_job_runner is not None:
        self.sync_job_runner.suspend()
```

artem-shelkovnikov: @tarekziade I'm not sure I understand how the approach you described works - `await asyncio.gather(*tasks)` intuitively should not wait on new tasks added to the `tasks`. I tried to make a small script to try it and it does not await for newly added tasks: https://gist.github.com/artem-shelkovnikov/e949b51657377b74a6dbce03646ca911

I've seen a diferent approach [here](https://stackoverflow.com/questions/61208300/asyncio-wait-for-each-task-in-a-dynamic-list-to-finish-or-be-cancelled) with gathering tasks as long as any are present, but it's pretty complex.

I also tested the suspension logic how it's done now and added `await asyncio.sleep(10)` in it inside `_sync_suspended` and verified that it's actually awaited. Is there even a problem to fix here?

artem-shelkovnikov: @wangch079 it can be encapsulated like that, but the `suspend` function is async and `loop. add_signal_handler` is not supposed to work with async function, if I got Tarek's concern correctly.

connectors/byoc.py
artem-shelkovnikov: Good call, thanks for noticing!

artem-shelkovnikov: Fixed in https://github.com/elastic/connectors-python/pull/376/commits/7138f9e292cba3f26de0a2f2c15726ded8322a4b

connectors/byoc.py
artem-shelkovnikov: True, we haven't added it yet! Waiting for Dana's response.

connectors/cli.py
tarekziade: synced with @artem-shelkovnikov -- we'll adapt `ConcurrentTasks` and use it

connectors/services/sync.py
```diff
@@ -45,8 +45,19 @@ def __init__(self, config, args):
             "max_concurrent_syncs", DEFAULT_MAX_CONCURRENT_SYNCS
         )
         self.connectors = None
+        self.syncs = ConcurrentTasks(max_concurrency=self.concurrent_syncs)
```
tarekziade, line 48: we can't have a single instance here because when it's joined it's thrown away - if we want to reuse the same instance of `ConcurrentTasks` you need to change the class so `join` empties `self.tasks`

tarekziade: there's also one problem here. if the same instance is used by competing tasks to put stuff, join() will take into account *all* the tasks -- I think it would be better to put this back in the loop (unless you had a requirement to make it a single instance)

tarekziade (commented):
just one nit about `self.syncs`

tarekziade (approved):

connectors/services/sync.py
```diff
@@ -134,10 +148,14 @@ async def _run(self):
...
-                    await syncs.join()
+                    await self.syncs.join()
                     if one_sync:
                         break
 
+                self.syncs = None
+                # Immediately break instead of sleeping
+                if not self.running:
+                    break
                 await self._sleeps.sleep(self.idling)
```
wangch079, line 159:
```suggestion
                if self.running:
                    await self._sleeps.sleep(self.idling)
```

connectors/services/sync.py
```diff
@@ -118,14 +130,16 @@ async def _run(self):
         es = ElasticServer(self.es_config)
         try:
             while self.running:
-                syncs = ConcurrentTasks(max_concurrency=self.concurrent_syncs)
+                # creating a pool of task for every round
+                self.syncs = ConcurrentTasks(max_concurrency=self.concurrent_syncs)
```
wangch079, line 134: Should we move this out of the `while` loop?

connectors/services/sync.py
wangch079: I think I had the similar discussion with @tarekziade , and here are my inputs:

1. I agree we should assign ConcurrentTasks to an instance variable, so we could access it when needed
2. Given point 1), the instantiation should happen outside the while loop (`while self.running`)
3. I believe we should put [connector.sync()](https://github.com/elastic/connectors-python/blob/0004f6badfae91527febfa9c837da96c2184fee5/connectors/services/sync.py#L88) inside the `ConcurrentTasks` (Later it will be [sync_job_runner.execute()](https://github.com/elastic/connectors-python/blob/014a189a947cba1e6c70b5d212feebdbf937eb81/connectors/services/sync.py#L101)), instead of `_one_sync()`
4. I don't quite understand why we need to use `await self.syncs.join()`. It will block the while loop into the next iteration until all tasks are done. The task will be scheduled to run in the background once `put`, so we don't really care when it's done.

connectors/services/sync.py
wangch079, line 159:
```suggestion
                if self.running:
                    await self._sleeps.sleep(self.idling)
```

artem-shelkovnikov: It's possible, but it'd still need a `break` somewhere.

I intuitively feel that it's more convenient to have `break` wrapped in `if`, not the main logic.

connectors/tests/test_sync.py
```diff
@@ -512,12 +511,24 @@ async def test_connector_service_poll_large(mock_responses, patch_logger, set_en
     assert_re(".*Sync done: 1001 indexed, 0  deleted", patch_logger.logs)
 
 
+@pytest.mark.asyncio
+async def test_connector_service_poll_suspended_suspends_job(
```
wangch079, line 515: Hi @artem-shelkovnikov , curious about how you simulate a `SIGTERM` signal?

artem-shelkovnikov: Hi @wangch079, for this particular test I don't need it, for manual testing I do `kill -15 <pid>`.

wangch079: Is this test case testing the service can mark a running job as `suspended` when the service shuts down?

artem-shelkovnikov: Yup, `create_and_run_service(MEM_CONFIG_FILE, 0.1)` terminates the service in 0.1 second by calling `service.stop`, and this code cancels and marks jobs as suspended!

PR #380, opened 2023-01-26
[Fix] Multiplier shouldn't be powered but multiplied

timgrein:
I think it's not the right intention to power a multiplier with retries, but to multiply the multiplier with the retry count.


timgrein:
As @artem-shelkovnikov pointed out this _could_ be intentional as this could be exponential backoff. Closing this PR for now, but advocating that we're more explicit about it through a comment or better variable names.

PR #382, opened 2023-01-26
Add logic to allow running a new job if connector was suspended before

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/3564

This change allows to continue with Suspended jobs.

When connector service checks connector and sees that "last_sync_status" was "Suspended", then service schedules an absolutely new job for this connector.

What it means in practice, also, is that if configuration/filtering of connector changed between the moment service was suspended and restarted, then new job will have new configuration/filtering as well.

#### Pre-Review Checklist

## Related Pull Requests

- [ ] https://github.com/elastic/connectors-python/pull/376

## Release Note

If connector service was restarted/terminated when connector was working on a sync, the connector will try to sync again when the service is started again.

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -443,8 +443,7 @@ class Connector:
         status = self.doc_source.get("last_sync_status")
         if status is None:
             return None
-        value = JobStatus[status.upper()]
-        return value
+        return JobStatus[status.upper()]
 
     @property
     def status(self):
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -655,25 +655,29 @@ class Connector:
 
         try:
             service_type = self.service_type
+            # if we don't sync, we still want to make sure we tell kibana we are connected
+            # if the status is different from comnected
+            if self.status != Status.CONNECTED:
+                self.status = Status.CONNECTED
+                await self.sync_doc()
+
             if sync_now:
                 self.sync_now = True
                 logger.info("Sync forced")
             else:
                 next_sync = self.next_sync()
-                if next_sync == SYNC_DISABLED or next_sync - idling > 0:
-                    if next_sync == SYNC_DISABLED:
-                        logger.debug(f"Scheduling is disabled for {service_type}")
-                    elif self.last_sync_status == JobStatus.SUSPENDED:
-                        logger.info("Restarting sync after suspension")
-                    else:
-                        logger.debug(
-                            f"Next sync for {service_type} due in {int(next_sync)} seconds"
-                        )
-                    # if we don't sync, we still want to make sure we tell kibana we are connected
-                    # if the status is different from comnected
-                    if self.status != Status.CONNECTED:
-                        self.status = Status.CONNECTED
-                        await self.sync_doc()
+                # First we check if sync is disabled, and it terminates all other conditions
+                if next_sync == SYNC_DISABLED:
+                    logger.debug(f"Scheduling is disabled for {service_type}")
+                    return
+                # Then we check if we need to restart SUSPENDED job
+                elif self.last_sync_status == JobStatus.SUSPENDED:
+                    logger.info("Restarting sync after suspension")
+                # And only then we check if we need to run sync right now or not
+                elif next_sync - idling > 0:
+                    logger.debug(
+                        f"Next sync for {service_type} due in {int(next_sync)} seconds"
+                    )
                     return
 
             try:
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -439,6 +439,14 @@ class Connector:
         self.features = Features(doc_source.get("features", {}))
 
     @property
+    def last_sync_status(self):
+        status = self.doc_source.get("last_sync_status")
+        if status is None:
+            return None
+        value = JobStatus[status.upper()]
+        return value
+
+    @property
     def status(self):
         return self._status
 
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov:
Question about documentation change - what is the place where I can put the information about connector service restarts and the logic in this PR? @elastic/ingestion-team

artem-shelkovnikov:
Question #2: I tried to find tests for the sync function and wasn't able to - are there any? If not, how can I easily test this logic?

navarone-feekery (approved):
Nice one!

navarone-feekery:
Approved before I saw your questions -- code looks good, unfortunately I don't know the answers. I'll leave the review in any case.

connectors/byoc.py
```diff
@@ -420,6 +420,14 @@ def _update_config(self, doc_source):
         self.language_code = doc_source["language"]
         self.features = Features(doc_source.get("features", {}))
 
+    @property
+    def last_sync_status(self):
+        status = self.doc_source.get("last_sync_status")
+        if status is None:
+            return None
+        value = JobStatus[status.upper()]
```
tarekziade, line 428: nit one liner

`return JobStatus[status.upper()]`

connectors/byoc.py
```diff
@@ -630,11 +638,16 @@ async def sync(self, elastic_server, idling, sync_now=False):
 
         try:
             service_type = self.service_type
-            if not sync_now:
+            if sync_now:
+                self.sync_now = True
+                logger.info("Sync forced")
```
tarekziade, line 643: should it be `debug`?

docs/DEVELOPING.md
```diff
@@ -72,7 +72,7 @@ Whether the connector should extract file content from network drive. Default va
 
 #### Content extraction
 
-The connector uses the Elastic ingest attachment processor plugin for extracting file contents. The ingest attachment processor extracts files by using the Apache text extraction library Tika. Supported file types eligible for extraction can be found [here](https://github.com/elastic/connectors-python/blob/9cf07a96288dcd542b641c51533ee2d427ef56ae/connectors/utils.py#L27).
+The connector uses the Elastic ingest attachment processor plugin for extracting file contents. The ingest attachment processor extracts files by using the Apache text extraction library Tika. Supported file types eligible for extraction can be found [here](https://github.com/elastic/connectors-python/blob/8.6/connectors/sources/network_drive.py#L19).
```
tarekziade, line 75: I'd love to get rid of line numbers because they change over time. Maybe we can give the variable name and point to the file without a line ?

connectors/byoc.py
```diff
@@ -630,11 +638,16 @@ async def sync(self, elastic_server, idling, sync_now=False):
...
+            else:
                 next_sync = self.next_sync()
                 if next_sync == SYNC_DISABLED or next_sync - idling > 0:
                     if next_sync == SYNC_DISABLED:
                         logger.debug(f"Scheduling is disabled for {service_type}")
+                    elif self.last_sync_status == JobStatus.SUSPENDED:
```
wangch079, line 649: This needs to be out of `if next_sync == SYNC_DISABLED or next_sync - idling > 0:` branch, or considering refactoring it to a new method as in:

https://github.com/elastic/connectors-python/blob/3e5c01202539bc8a3eac79f74d1b107242f9ae24/connectors/services/sync.py#L174-L188

`self.last_sync_status == JobStatus.SUSPENDED` can be added after `sync_now` check and before `next_sync = connector.next_sync()`

docs/DEVELOPING.md
artem-shelkovnikov: Oh wow, that's a bad rebase/merge, I'll get rid of this change here. +1 to proper names too.

connectors/byoc.py
artem-shelkovnikov: Good catch!

connectors/byoc.py
artem-shelkovnikov: I haven't changed this log level, but to me it makes sense to keep it in info - it's good to have indication why sync was started even in INFO IMO, what do you think?

connectors/byoc.py
```diff
@@ -630,24 +637,30 @@ async def sync(self, elastic_server, idling, sync_now=False):
...
                 self.sync_now = True
                 logger.info("Sync forced")
+            else:
+                next_sync = self.next_sync()
+                # First we check if sync is disabled, and it terminates all other conditions
+                if next_sync == SYNC_DISABLED:
+                    logger.debug(f"Scheduling is disabled for {service_type}")
+                    return
+                # Then we check if we need to restart SUSPENDED job
+                elif self.last_sync_status == JobStatus.SUSPENDED:
```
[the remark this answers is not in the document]
artem-shelkovnikov: I think it's not possible because `next_sync` returns -1 for disabled syncs, and we should not restart suspended jobs if syncs are disabled.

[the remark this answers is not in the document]
wangch079: I assumed we should always resume (re-start) a new job if it's suspended. 

The edge case:
1. a Job starts execution
2. disable sync for the connector
3. shut down the connector service
4. start the connector service

The job won't be resumed. 

I think both make sense, so it's not a blocker.

wangch079 (approved):

connectors/byoc.py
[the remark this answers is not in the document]
artem-shelkovnikov: cc @danajuratoni

PR #389, opened 2023-01-27
Create an event loop instead of getting for verify.py

artem-shelkovnikov:
Small fix for a warning when running functional tests:

> /Users/artemshelkovnikov/git_tree/connectors-py/connectors/tests/../../scripts/verify.py:80: DeprecationWarning: There is no current event loop
>  loop = asyncio.get_event_loop()


tarekziade:
A cleaner fix is to remove `loop` altogether:

```
try:
     asyncio.run(
        verify(args.service_type, args.index_name, args.size, config)
    )
    print("Bye")
except (asyncio.CancelledError, KeyboardInterrupt):
    print("Bye")
```

tarekziade (approved):

PR #393, opened 2023-01-27
Allow concurrent syncs

tarekziade:
Allows concurrent syncs to happen.

The `SyncService` class runs syncs sequentially. This patch will allow it to run several syncs at the same time (with a max concurrency value)

We can keep that value to `1` on cloud for now, because this will make the service grow in memory.

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -551,7 +551,7 @@ class Connector:
         self.doc_source["last_synced"] = iso_utc()
         await self.sync_doc()
         logger.info(
-            f"Sync done: {indexed_count} indexed, {doc_deleted} "
+            f"[{self.id}] Sync done: {indexed_count} indexed, {doc_deleted} "
             f" deleted. ({int(time.time() - self._start_time)} seconds)"
         )
 
--- a/connectors/services/sync.py
+++ b/connectors/services/sync.py
@@ -32,7 +32,7 @@ from connectors.logger import logger
 from connectors.services.base import BaseService
 from connectors.utils import ConcurrentTasks
 
-DEFAULT_MAX_CONCURRENT_SYNCS = 5
+DEFAULT_MAX_CONCURRENT_SYNCS = 1
 
 
 class SyncService(BaseService):
--- a/connectors/services/sync.py
+++ b/connectors/services/sync.py
@@ -11,6 +11,7 @@ Event loop
 - mirrors an Elasticsearch index with a collection of documents
 """
 import asyncio
+import functools
 
 from connectors.byoc import (
     ConnectorIndex,
@@ -29,6 +30,9 @@ from connectors.filtering.validation import (
 )
 from connectors.logger import logger
 from connectors.services.base import BaseService
+from connectors.utils import ConcurrentTasks
+
+DEFAULT_MAX_CONCURRENT_SYNCS = 5
 
 
 class SyncService(BaseService):
@@ -37,6 +41,9 @@ class SyncService(BaseService):
         self.args = args
         self.idling = self.service_config["idling"]
         self.hb = self.service_config["heartbeat"]
+        self.concurrent_syncs = self.service_config.get(
+            "max_concurrent_syncs", DEFAULT_MAX_CONCURRENT_SYNCS
+        )
         self.connectors = None
 
     async def _one_sync(self, connector, es, sync_now):
@@ -109,7 +116,7 @@ class SyncService(BaseService):
         )
 
         es = ElasticServer(self.es_config)
-        syncs = ConcurrentTasks()
+        syncs = ConcurrentTasks(max_concurrency=self.concurrent_syncs)
         try:
             while self.running:
                 try:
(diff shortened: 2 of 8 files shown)


connectors/tests/test_sync.py
```diff
@@ -658,3 +669,16 @@ async def _sync(*args):
...
+async def test_concurrent_syncs(mock_responses, patch_logger, set_env):
+    await set_server_responses(mock_responses, [FAKE_CONFIG, FAKE_CONFIG, FAKE_CONFIG])
+    service = create_service(CONFIG_FILE)
+    asyncio.get_event_loop().call_soon(service.stop)
+    await service.run()
+
+    # make sure we synced the three connectors
+    patch_logger.assert_present("[1] Sync done: 1 indexed, 0  deleted. (0 seconds)")
+    patch_logger.assert_present("[2] Sync done: 1 indexed, 0  deleted. (0 seconds)")
+    patch_logger.assert_present("[3] Sync done: 1 indexed, 0  deleted. (0 seconds)")
```
artem-shelkovnikov, line 684: Is it possible to do assertion on mocks/storage instead of logger?

connectors/byoc.py
```diff
@@ -528,7 +528,7 @@ async def _sync_done(self, job, result, exception=None):
         self.doc_source["last_synced"] = iso_utc()
         await self.sync_doc()
         logger.info(
-            f"Sync done: {indexed_count} indexed, {doc_deleted} "
+            f"[{self.id}] Sync done: {indexed_count} indexed, {doc_deleted} "
```
artem-shelkovnikov, line 531: We'll need to do it in a lot of places with this change, PR to track: https://github.com/elastic/enterprise-search-team/issues/3809

connectors/tests/test_utils.py
```diff
@@ -58,18 +58,28 @@ def test_invalid_names():
 @pytest.mark.asyncio
 async def test_mem_queue_speed(patch_logger):
     # with memqueue
-    queue = MemQueue(maxmemsize=1024 * 1024, refresh_interval=0.1, refresh_timeout=2)
-    start = time.time()
-    for i in range(1000):
-        await queue.put("x" * 100)
-    mem_queue_duration = time.time() - start
+    mem_queue_duration = 10000
```
artem-shelkovnikov, line 61: Should this change be in this PR actually? Seems irrelevant.

artem-shelkovnikov (commented):
Looks good, left a couple questions

connectors/tests/test_sync.py
tarekziade: Sure why not

This assertion makes sure that we reached that line and did index a document https://github.com/elastic/connectors-python/blob/main/connectors/byoc.py#L553

All calls to Elasticsearch are mocked / what do you have in mind to reach the same level of coverage+assertion?

connectors/services/sync.py
```diff
@@ -111,22 +118,26 @@ async def _run(self):
...
+                syncs = ConcurrentTasks(max_concurrency=self.concurrent_syncs)
                 try:
                     logger.debug(f"Polling every {self.idling} seconds")
                     query = self.connectors.build_docs_query(
                         native_service_types, connectors_ids
                     )
-
                     async for connector in self.connectors.get_all_docs(query=query):
-                        await self._one_sync(connector, es, sync_now)
+                        await syncs.put(
```
wangch079, line 128: should we call `put` when a real sync happens?

connectors/tests/test_sync.py
artem-shelkovnikov: For me good rule of thumb for testing is: if you delete some lines from the project and breaking the logic but the test will still pass, then test should be improved.

In essence in this case it might mean checking elasticsearch mocked requests that indicate the sync was finished, for example. Or mocking more classes and checking they were called.

connectors/tests/test_utils.py
artem-shelkovnikov: Which discussion with Tim?

timgrein: This test was flaky and sometimes broke locally and in the CI. I proposed to reduce the probability of it breaking and use `timeit` instead of `time` as `timeit` is designed for these kinds of measurements (f.e. turns of GC etc.). But this change also reduces the probability of breaking randomly a lot, so lgtm.

connectors/services/sync.py
tarekziade: not sure what you mean.  

- If there's room, `put` will push the task into the pool, and also start it.  
- if there's no run, this blocks until there's room

connectors/tests/test_utils.py
tarekziade: yeah it's similar to timeit -- I'll put it in a separate patch

connectors/tests/test_sync.py
artem-shelkovnikov: Something that we'll need to follow-up upon structurally, let's do it later, I'll prepare PRs/slides to discuss

artem-shelkovnikov (approved):

connectors/services/sync.py
wangch079: Now you `put` `_one_sync` into the `ConcurrentTasks`. Inside `_one_sync` there are still some checks, e.g. whether the connector is configured to sync, and whether the connector is scheduled/forced to sync. If all checks pass, it will sync eventually by calling `connector.sync()`(will be refactored to a separate class `SyncJobRunner`), and this should be the task to `put` into `ConcurrentTasks`

PR #398, opened 2023-01-30
[Filtering] Advanced rules validation for MySQL

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3598 ##

This PR adds MySQL advanced rules validation:
- Validates configured databases vs databases specified in advanced rules
- Validates remotely existing tables vs database/tables in advanced rules
- Validates if databases are accessible by the configured database user

The method `advanced_rules_validators` in the `BaseDataSource` is now also tied to the data source object to be able to use methods within the data source implementation (f.e. talking to a MySQL instance for validation). 

The PR also includes a new flexible `retry` decorator, which can be used to retry a coroutine with different retrying strategies (`CONSTANT`, `LINEAR_BACKOFF`, `EXPONENTIAL_BACKOFF`).

#### Pre-Review Checklist
~- [ ] Considered corresponding documentation changes~ (separate PR)

## Release Note

Validate advanced rules for the MySQL connector.

### DIFF
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -24,7 +24,7 @@ def parse(basic_rules_json):
     The parser works in the following way:
       - Map every raw basic rule in the json array to the corresponding BasicRule object
       - Filter out every basic rule, which returns true for is_default_rule()
-      - Sort the result in descending order according to their basic rule order (rules are executed in descending order)
+      - Sort the result in ascending order according to their basic rule order (rules are executed in ascending order)
     """
     if not basic_rules_json:
         return []
@@ -44,7 +44,6 @@ def parse(basic_rules_json):
             )
         ),
         key=lambda basic_rule: basic_rule.order,
-        reverse=True,
     )
 
 
--- a/connectors/tests/test_basic_rule.py
+++ b/connectors/tests/test_basic_rule.py
@@ -538,7 +538,7 @@ def test_parser_rejects_default_rule():
     assert not contains_default_rule(parsed_basic_rules)
 
 
-def test_rules_are_ordered_descending_with_respect_to_the_order_property():
+def test_rules_are_ordered_ascending_with_respect_to_the_order_property():
     raw_basic_rules = [BASIC_RULE_ONE_JSON, BASIC_RULE_THREE_JSON, BASIC_RULE_TWO_JSON]
 
     parsed_basic_rules = parse(raw_basic_rules)
@@ -547,9 +547,9 @@ def test_rules_are_ordered_descending_with_respect_to_the_order_property():
     third_rule = parsed_basic_rules[2]
 
     assert len(parsed_basic_rules) == 3
-    assert is_rule_three(first_rule)
+    assert is_rule_one(first_rule)
     assert is_rule_two(second_rule)
-    assert is_rule_one(third_rule)
+    assert is_rule_three(third_rule)
 
 
 def test_matches_default_rule():


connectors/filtering/validation.py
```diff
@@ -62,6 +64,16 @@ def valid_result(cls, rule_id):
             rule_id=rule_id, is_valid=True, validation_message="Valid rule"
         )
 
+    def __eq__(self, other):
+        if other is None:
```
tarekziade, line 68: Mmm allowing `None`, but not other types is smelly -- also what happens if I compare with an `integer` for example?

I would be more drastic here and reject all tentative to compare for equality when `other` is not a `SyncRuleValidationResult` object:

```
if not isinstance(other, SyncRuleValidationResult):
    raise TypeError(f"Can't compare SyncRuleValidationResult with {type(other)}")
```

connectors/sources/mysql.py
```diff
@@ -21,12 +27,92 @@
...
+
+        return await self._remote_validation(advanced_rules, databases_to_filter)
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def _remote_validation(self, advanced_rules, databases_to_filter):
+        await self.source.ping()
```
tarekziade, line 65: Do we want to ping again here? it's pinged by the protocol before we sync, I am just wondering if that's not going to be an extra ping on every attempt.

maybe `ping` can cache its result so we don't ping several times? and also invalidate that cache after some time

connectors/sources/mysql.py
```diff
@@ -21,12 +27,92 @@
...
+        )
+        inaccessible_databases_to_filter = inaccessible_databases.intersection(
+            databases_to_filter
+        )
+
+        if len(inaccessible_databases_to_filter) > 0:
+            return SyncRuleValidationResult(
+                SyncRuleValidationResult.ADVANCED_RULES,
+                is_valid=False,
+                validation_message=f"Inaccessible databases: {format_list(inaccessible_databases_to_filter)} for user '{self.source.configuration.get('user', 'None')}'.",
```
tarekziade, line 78: do we really want to display `None` for the user by default?

connectors/utils.py
```diff
@@ -352,3 +353,42 @@ def get_event_loop(uvloop=False):
...
+    CONSTANT = 0
+    LINEAR_BACKOFF = 1
+    EXPONENTIAL_BACKOFF = 2
+
+
+class UnknownRetryStrategyError(Exception):
+    pass
+
+
+def retryable(retries=1, interval=1, strategy=RetryStrategy.LINEAR_BACKOFF):
```
tarekziade, line 368: neat!
I would use `retries=3` by default. No incentive to user `retryable()` if it tries only once.

tarekziade, line 368: english: it's `retriable` not `retryable` 

https://en.wiktionary.org/wiki/retriable

tarekziade, line 368: if `interval` is a float, use `1.0` to make this explicit in the signature

tarekziade (commented):
nice! just a few nits and let's avoid multiple pings on the same backend if possible

connectors/utils.py
timgrein: I looked it up on some other pages and it seems like both is ok

connectors/filtering/validation.py
timgrein: Good point! Adjusted it.

connectors/sources/mysql.py
timgrein: Adjusted it to "No user configured"

connectors/sources/mysql.py
timgrein: I think we can just remove the `ping` call as we always ping before the validations. We could introduce ping caching with another PR, but I think this would be too much scope for this PR.

timgrein: I remember why I introduced `ping`. Ping actually instantiates the connection pool, if it's not instantiated. As validating can also happen outside of the context of a sync I've used to ping to make sure we can actually talk to the data source. I've noted it down and will create a tech-debt issue to separate the connection pool instantiation from `ping`.

So I would keep `ping` for now and tackle it in between documentation/test plan creation.

tarekziade: yeah ok. Even if we have a lazy instantiation of pools, this should be an internal implementation detail, and  `ping` should not be mandatory to create the pool -- even if it's a mandatory step for the SyncService.

connectors/utils.py
tarekziade: oh! TIL

tarekziade (approved):

connectors/sources/mysql.py
timgrein: Tracked by: https://github.com/elastic/enterprise-search-team/issues/3821

PR #407, opened 2023-01-31
improved the speed comparison to reduce noise

tarekziade:
run it 3 times and get the min

### DIFF
--- a/connectors/tests/test_utils.py
+++ b/connectors/tests/test_utils.py
@@ -13,7 +13,6 @@ import random
 import tempfile
 import time
 import timeit
-from textwrap import dedent
 from unittest.mock import Mock
 
 import pytest
@@ -61,19 +60,24 @@ def test_invalid_names():
 
 
 def test_mem_queue_speed(patch_logger):
+    def mem_queue():
+        import asyncio
 
-    async def run():
-        for i in range(1000):
-            await queue.put("x" * 100)
+        from connectors.utils import MemQueue
 
-    asyncio.run(run())
-    """
-    )
+        queue = MemQueue(
+            maxmemsize=1024 * 1024, refresh_interval=0.1, refresh_timeout=2
+        )
 
-    mem_queue_duration = min(timeit.repeat(memqueue_script, number=1, repeat=3))
+        async def run():
+            for i in range(1000):
+                await queue.put("x" * 100)
 
+        asyncio.run(run())
 
-     def vanilla_queue():
+    mem_queue_duration = min(timeit.repeat(mem_queue, number=1, repeat=3))
+
+    def vanilla_queue():
         import asyncio
 
         queue = asyncio.Queue()
--- a/connectors/tests/test_utils.py
+++ b/connectors/tests/test_utils.py
@@ -61,14 +61,6 @@ def test_invalid_names():
 
 
 def test_mem_queue_speed(patch_logger):
-    # with memqueue
-
-    memqueue_script = dedent(
-        """
-    import asyncio
-    from connectors.utils import MemQueue
-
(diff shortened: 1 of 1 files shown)


connectors/tests/test_utils.py
```diff
@@ -58,21 +60,42 @@ def test_invalid_names():
...
+
+    async def run():
+        for i in range(1000):
+            await queue.put("x" * 100)
+
+    asyncio.run(run())
+    """
+    )
+
+    mem_queue_duration = min(timeit.repeat(memqueue_script, number=1, repeat=3))
```
timgrein, line 81:
```suggestion

     def mem_queue():
        import asyncio
        from connectors.utils import MemQueue

        queue = MemQueue(maxmemsize=1024 * 1024, refresh_interval=0.1, refresh_timeout=2)

        async def run():
            for i in range(1000):
                await queue.put("x" * 100)

        asyncio.run(run())

    mem_queue_duration = min(timeit.repeat(mem_queue, number=1, repeat=3))
```

I think we should use nested functions as we lose auto-suggestions with `dedent` in IDEs.

connectors/tests/test_utils.py
```diff
@@ -58,21 +60,42 @@ def test_invalid_names():
...
+
+    async def run():
+        for i in range(1000):
+            await queue.put("x" * 100)
+
+    asyncio.run(run())
+    """
+    )
+
+    queue_duration = min(timeit.repeat(queue_script, number=1, repeat=3))
```
timgrein, line 98:
```suggestion

     def vanilla_queue():
        import asyncio

        queue = asyncio.Queue()

        async def run():
            for i in range(1000):
                await queue.put("x" * 100)

        asyncio.run(run())

    queue_duration = min(timeit.repeat(vanilla_queue, number=1, repeat=3))
```

Same point as above.

timgrein (commented):
Nice, that we're now using `timeit`. Added two comments, where I would suggest that we use nested functions instead of `dedent` as we're loosing auto-suggestions/highlighting etc. in IDEs.

connectors/tests/test_utils.py
```diff
@@ -58,21 +60,42 @@ def test_invalid_names():
...
-async def test_mem_queue_speed(patch_logger):
+def test_mem_queue_speed(patch_logger):
     # with memqueue
+
+    memqueue_script = dedent(
+        """
+    import asyncio
+    from connectors.utils import MemQueue
+
     queue = MemQueue(maxmemsize=1024 * 1024, refresh_interval=0.1, refresh_timeout=2)
```
timgrein, line 71:
```suggestion

```

timgrein, line 71:
```suggestion

```

connectors/tests/test_utils.py
```diff
@@ -58,21 +60,42 @@ def test_invalid_names():
...
+    )
+
+    mem_queue_duration = min(timeit.repeat(memqueue_script, number=1, repeat=3))
 
     # vanilla queue
+    queue_script = dedent(
+        """
+    import asyncio
+
     queue = asyncio.Queue()
```
timgrein, line 88:
```suggestion

```

timgrein:
Oops, my mousepad is super sensitive, accidentally pressed update haha

timgrein (approved):
LGTM, still not a fan of `dedent` vs. a nested function, but using `timeit` looks good.

tarekziade:
> LGTM, still not a fan of `dedent` vs. a nested function (from an IDE user perspective), but using `timeit` looks good. Thanks for the change.

erm, I thought I applied your proposed changes, will try again

PR #412, opened 2023-02-01
[Docs] Documentation for implementing sync rules

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3732 ##

This PR adds documentation on how to implement advanced rules, advanced rules validation and basic rules validation.

#### Pre-Review Checklist
~- [ ] Covered the changes with automated tests~
~- [ ] Tested the changes locally~

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -16,7 +16,7 @@ Each connector will have their own specific behaviors and implementations. When 
 
 #### Basic rules vs advanced rules
 
-Sync rules is another term for either basic or advanced rules. 
+Sync rules are made up of basic and advanced rules. 
 Basic rules are implemented generically on the framework-level and work out-of-the-box for every new connector. 
 Advanced rules are specific to each data source. 
 You can learn more about the use cases of sync rules [here (Sync rules documentation)](https://www.elastic.co/guide/en/enterprise-search/current/sync-rules.html). 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -62,7 +62,7 @@ Example:
 ```
 
 Note that the framework calls `get_docs` with the parameter `filtering` being of type `Filter`, which is located in [byoc.py](../connectors/sources/mysql.py).
-The `Filter` class provides convenient methods to extract advanced rules from the filter and to check, whether advanced rules are present.
+The `Filter` class provides convenient methods to extract advanced rules from the filter and to check whether advanced rules are present.
 
 #### How to validate advanced rules
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -66,9 +66,9 @@ The `Filter` class provides convenient methods to extract advanced rules from th
 
 #### How to validate advanced rules
 
-For validating advanced rules the framework takes the list of validators returned by the method `advanced_rules_validators` and calls them in the order they appear in that list.
+To validate advanced rules the framework takes the list of validators returned by the method `advanced_rules_validators` and calls them in the order they appear in that list.
 By default, this list is empty in the [BaseDataSource](../connectors/source.py) as advanced rules are always specific to the connector implementation.
-You can plug in custom validators by implementing a class containing a `validate` method, which accepts one parameter.
+Plug in custom validators by implementing a class containing a `validate` method, which accepts one parameter.
 The framework expects the custom validators to return a `SyncRuleValidationResult`, which can be found in [validation.py](../connectors/filtering/validation.py).
 
 ```python
@@ -79,7 +79,7 @@ class MyValidator(AdvancedRulesValidator):
         return SyncRuleValidationResult(...)
 ```
 
-Note that the framework will call `validate` with the parameter `advanced_rules` being of type `dict`.
+Note that the framework will call `validate` with the parameter `advanced_rules` of type `dict`.
 Now you can return a list of validator instances in `advanced_rules_validators`:
 
 ```python
@@ -96,9 +96,9 @@ The framework will handle the rest like scheduling validation, calling the custo
 We don't recommend fully overriding `basic_rule_validators`, because you'll lose the default validations.
 
 The framework already provides default validations for basic rules.
-If you want to extend the default validation you can provide custom basic rules validators.
+To extend the default validation, provide custom basic rules validators.
 There are two possible ways to validate basic rules:
-- Every rule gets validated in isolation (extend the class `BasicRuleValidator` located in [validation.py](../connectors/filtering/validation.py)):
+- **Every rule gets validated in isolation**. Extend the class `BasicRuleValidator` located in [validation.py](../connectors/filtering/validation.py):
     ```python
     class MyBasicRuleValidator(BasicRuleValidator):
         
@@ -107,7 +107,7 @@ There are two possible ways to validate basic rules:
             # custom validation logic
             return SyncRuleValidationResult(...)
     ```
-- Validate the whole set of basic rules, if you want to validate constraints on the set of rules, f.e. to detect duplicate or conflicting rules (extend the class `BasicRulesSetValidator`located in [validation.py](../connectors/filtering/validation.py)):
+- **Validate the whole set of basic rules**. If you want to validate constraints on the set of rules, for example to detect duplicate or conflicting rules. Extend the class `BasicRulesSetValidator` located in [validation.py](../connectors/filtering/validation.py):
     ```python
     class MyBasicRulesSetValidator(BasicRulesSetValidator):
         
@@ -117,7 +117,7 @@ There are two possible ways to validate basic rules:
             return SyncRuleValidationResult(...)
     ```
 
-If you want to preserve the default basic rule validations and extend it with your custom logic you can override `basic_rules_validators` like this:
+To preserve the default basic rule validations and extend these with your custom logic, override `basic_rules_validators` like this:
 ```python
 class MyDataSource(BaseDataSource):
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -61,7 +61,7 @@ Example:
 }
 ```
 
-Note that the framework calls `get_docs` with the parameter `filtering` being of type `Filter`, which is located in [byoc.py](../connectors/sources/mysql.py).
+Note that the framework calls `get_docs` with the parameter `filtering` being of type `Filter`, which is located in [byoc.py](../connectors/byoc.py).
 The `Filter` class provides convenient methods to extract advanced rules from the filter and to check whether advanced rules are present.
 
 #### How to validate advanced rules
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -30,7 +30,7 @@ Take a look at the `get_docs` method in the [MySQL connector](../connectors/sour
 
 #### How to implement advanced rules
 
-When implementing a new connector you've to follow the API of the [BaseDataSource](../connectors/source.py).
+When implementing a new connector follow the API of the [BaseDataSource](../connectors/source.py).
 The custom implementation for advanced rules is usually located inside the `get_docs` function:
 
 ```python
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -67,7 +67,7 @@ The `Filter` class provides convenient methods to extract advanced rules from th
 #### How to validate advanced rules
 
 For validating advanced rules the framework takes the list of validators returned by the method `advanced_rules_validators` and calls them in the order they appear in that list.
-On default this list is empty in the [BaseDataSource](../connectors/source.py) as advanced rules are always specific to the connector implementation.
+By default, this list is empty in the [BaseDataSource](../connectors/source.py) as advanced rules are always specific to the connector implementation.
 You can plug in custom validators by implementing a class containing a `validate` method, which accepts one parameter.
 The framework expects the custom validators to return a `SyncRuleValidationResult`, which can be found in [validation.py](../connectors/filtering/validation.py).
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -43,7 +43,7 @@ async def get_docs(self, filtering=None):
 ```
 
 For example, here you could pass custom queries to a database.
-The structure of the advanced rules is dependent on your implementation and your concrete use case.
+The structure of the advanced rules depends on your implementation and your concrete use case.
 For MySQL the advanced rules structure looks like this:
 You specify databases on the top level, which contain tables, which specify a custom query.
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more 
 
 Each connector will have their own specific behaviors and implementations. When a connector is loaded, it stays in memory, so you can come up with any strategy you want to make it more efficient. You just need to be careful not to blow memory.
 
+### Sync rules
+
+#### Basic rules vs advanced rules
+
+Sync rules is another term for either basic or advanced rules. 
+Basic rules are implemented generically on the framework-level and work out-of-the-box for every new connector. 
+Advanced rules are specific to each data source. 
+You can learn more about the use cases of sync rules [here (Sync rules documentation)](https://www.elastic.co/guide/en/enterprise-search/current/sync-rules.html). 
+
+Example:
+
+For MySQL we've implemented advanced rules to pass custom SQL queries directly to the corresponding MySQL instance.
+This has the advantage of offloading a lot of the filtering to the specific data source, which normally results in a reduction of the amount of data coming back from the data source.
+Also, data sources usually have highly optimized and specific filtering capabilities you may want to expose to the users of your connector.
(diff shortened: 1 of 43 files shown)


docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+```python
+async def get_docs(self, filtering=None):
+    if filtering and filtering.has_advanced_rules():
+        advanced_rules = filtering.get_advanced_rules()
+        # your custom advanced rules implementation
+    else:
+        # default fetch all data implementation 
+```
+
+Here you can pass f.e. custom queries to a database or something similar.
```
seanstory, line 45: is `f.e.` "front end?" 
Probably best to avoid abbreviations or acronyms where possible, for maximum clarity.

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+    "table_2": "SELECT ... FROM ...;"
+  },
+  "database_2": {
+    "table_1": "SELECT ... FROM ...;"
+  }
+}
+```
+
+Note that the framework calls `get_docs` with the parameter `filtering` being of type `Filter`, which is located in [byoc.py](../connectors/sources/mysql.py).
+The `Filter` class provides convenient methods to extract advanced rules from the filter and to check, whether advanced rules are present.
```
seanstory, line 65:
```suggestion
The `Filter` class provides convenient methods to extract advanced rules from the filter and to check whether advanced rules are present.
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+
+#### How to validate advanced rules
+
+For validating advanced rules the framework takes the list of validators returned by the method `advanced_rules_validators` and calls them in the order they appear in that list.
+On default this list is empty in the [BaseDataSource](../connectors/source.py) as advanced rules are always specific to the connector implementation.
```
seanstory, line 70:
```suggestion
By default, this list is empty in the [BaseDataSource](../connectors/source.py) as advanced rules are always specific to the connector implementation.
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+```python
+class MyDataSource(BaseDataSource):
+
+    def advanced_rules_validators(self):
+        return [MyValidator()]
+```
+
+The framework will handle the rest like scheduling validation, calling the custom validators and storing the corresponding results.
+
+#### How to provide custom basic rule validation
```
seanstory, line 94: I haven't been watching closely - is this something we've intentionally introduced? 
In 8.6 in the ruby connectors, the intent was for Advanced Rules (and basic pre-processing) to be customizable, but for (post-processing) Basic Rules to be consistent across all connectors. 

I don't think we really want to allow source-specific validations for basic rules - that makes them much less "basic".

timgrein: We've default validations and further down I state, that it's discouraged to override them. I thought if people want to intentionally override the validation (for example removing the regex validator or add another one) then we should describe how to do that. But that's just my personal opinion of documenting this explicitly, as it's possible to override it anway, also happy when we remove it 👍

docs/CONTRIBUTING.md
timgrein: "for example". Would you use "e.g." here?

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
 
 Each connector will have their own specific behaviors and implementations. When a connector is loaded, it stays in memory, so you can come up with any strategy you want to make it more efficient. You just need to be careful not to blow memory.
 
+### Sync rules
+
+#### Basic rules vs advanced rules
+
+Sync rules is another term for either basic or advanced rules.
```
leemthompo, line 19:
```suggestion
Sync rules are made up of basic and advanced rules. 
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
 
 Each connector will have their own specific behaviors and implementations. When a connector is loaded, it stays in memory, so you can come up with any strategy you want to make it more efficient. You just need to be careful not to blow memory.
 
+### Sync rules
```
leemthompo, line 15: General remark: this should be consistent with https://www.elastic.co/guide/en/enterprise-search/current/sync-rules.html. For example, I presume the introductory details are the same.

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+
+#### Basic rules vs advanced rules
+
+Sync rules is another term for either basic or advanced rules. 
+Basic rules are implemented generically on the framework-level and work out-of-the-box for every new connector. 
+Advanced rules are specific to each data source. 
+You can learn more about the use cases of sync rules [here (Sync rules documentation)](https://www.elastic.co/guide/en/enterprise-search/current/sync-rules.html).
```
leemthompo, line 22:
```suggestion
Learn more about sync rules [in the Enterprise Search documentation](https://www.elastic.co/guide/en/enterprise-search/current/sync-rules.html). 
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+
+Sync rules is another term for either basic or advanced rules. 
+Basic rules are implemented generically on the framework-level and work out-of-the-box for every new connector. 
+Advanced rules are specific to each data source. 
+You can learn more about the use cases of sync rules [here (Sync rules documentation)](https://www.elastic.co/guide/en/enterprise-search/current/sync-rules.html). 
+
+Example:
+
+For MySQL we've implemented advanced rules to pass custom SQL queries directly to the corresponding MySQL instance.
+This has the advantage of offloading a lot of the filtering to the specific data source, which normally results in a reduction of the amount of data coming back from the data source.
```
leemthompo, line 27:
```suggestion
This offloads a lot of the filtering to the data source, which helps reduce data transfer.
```

Just simplifying a long sentence.

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+Also, data sources usually have highly optimized and specific filtering capabilities you may want to expose to the users of your connector.
+You can take a look at the `get_docs` method in the [MySQL connector](../connectors/sources/mysql.py) to see an advanced rules implementation.
```
leemthompo, line 29:
```suggestion
Take a look at the `get_docs` method in the [MySQL connector](../connectors/sources/mysql.py) to see an advanced rules implementation.
```

Minor copyedit: often removing "you can", makes the sentence more direct and reduces sentence length. :)

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+
+#### How to implement advanced rules
+
+When implementing a new connector you've to follow the API of the [BaseDataSource](../connectors/source.py).
```
leemthompo, line 33:
```suggestion
When implementing a new connector follow the API of the [BaseDataSource](../connectors/source.py).
```

docs/CONTRIBUTING.md
leemthompo: I'd avoid both abbreviations, and use the full phrase "for example".

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+async def get_docs(self, filtering=None):
+    if filtering and filtering.has_advanced_rules():
+        advanced_rules = filtering.get_advanced_rules()
+        # your custom advanced rules implementation
+    else:
+        # default fetch all data implementation 
+```
+
+Here you can pass f.e. custom queries to a database or something similar.
+The structure of the advanced rules is dependent on your implementation and your concrete use case.
```
leemthompo, line 46:
```suggestion
The structure of the advanced rules depends on your implementation and your concrete use case.
```

Simplification

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+    "table_1": "SELECT ... FROM ...;",
+    "table_2": "SELECT ... FROM ...;"
+  },
+  "database_2": {
+    "table_1": "SELECT ... FROM ...;"
+  }
+}
+```
+
+Note that the framework calls `get_docs` with the parameter `filtering` being of type `Filter`, which is located in [byoc.py](../connectors/sources/mysql.py).
```
leemthompo, line 64: The link says the parameter is located in `byoc.py`, but the file linked to is called `mysql.py`.

leemthompo, line 64:
```suggestion
Note that the framework calls `get_docs` with the parameter `filtering` of type `Filter`, which is located in [byoc.py](../connectors/sources/mysql.py).
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+The `Filter` class provides convenient methods to extract advanced rules from the filter and to check whether advanced rules are present.
+
+#### How to validate advanced rules
+
+For validating advanced rules the framework takes the list of validators returned by the method `advanced_rules_validators` and calls them in the order they appear in that list.
```
leemthompo, line 69:
```suggestion
To validate advanced rules the framework takes the list of validators returned by the method `advanced_rules_validators` and calls them in the order they appear in that list.
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+By default, this list is empty in the [BaseDataSource](../connectors/source.py) as advanced rules are always specific to the connector implementation.
+You can plug in custom validators by implementing a class containing a `validate` method, which accepts one parameter.
```
leemthompo, line 71:
```suggestion
Plug in custom validators by implementing a class containing a `validate` method, which accepts one parameter.
```

Again this is a little subjective, so feel free to use "you can" whenever you want but consider eliminating it in longer sentences, for brevity. 🤓

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+
+```python
+class MyValidator(AdvancedRulesValidator):
+    
+    def validate(self, advanced_rules):
+        # custom validation logic
+        return SyncRuleValidationResult(...)
+```
+
+Note that the framework will call `validate` with the parameter `advanced_rules` being of type `dict`.
```
leemthompo, line 82:
```suggestion
Note that the framework will call `validate` with the parameter `advanced_rules` of type `dict`.
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+Now you can return a list of validator instances in `advanced_rules_validators`:
+
+```python
+class MyDataSource(BaseDataSource):
+
+    def advanced_rules_validators(self):
+        return [MyValidator()]
+```
+
+The framework will handle the rest like scheduling validation, calling the custom validators and storing the corresponding results.
```
leemthompo, line 92:
```suggestion
The framework will handle the rest; scheduling validation, calling the custom validators and storing the corresponding results.
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+
+#### How to provide custom basic rule validation
+
+The framework already provides default validations for basic rules.
+If you want to extend the default validation you can provide custom basic rules validators.
```
leemthompo, line 97:
```suggestion
To extend the default validation, provide custom basic rules validators.
```

leemthompo: Again, cutting out these introductory phrases makes the sentence much easier to read.

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+There are two possible ways to validate basic rules:
+- Every rule gets validated in isolation (extend the class `BasicRuleValidator` located in [validation.py](../connectors/filtering/validation.py)):
```
leemthompo, line 99:
```suggestion
- **Every rule gets validated in isolation**. Extend the class `BasicRuleValidator` located in [validation.py](../connectors/filtering/validation.py):
```

Bold for emphasis, and giving the action its own sentence, without parenthesis.

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+    ```python
+    class MyBasicRulesSetValidator(BasicRulesSetValidator):
+        
+        @classmethod
+        def validate(cls, set_of_rules):
+            # custom validation logic
+            return SyncRuleValidationResult(...)
+    ```
+
+If you want to preserve the default basic rule validations and extend it with your custom logic you can override `basic_rules_validators` like this:
```
leemthompo, line 118:
```suggestion
To preserve the default basic rule validations and extend it with your custom logic override `basic_rules_validators` like this:
```

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+
+    @classmethod
+    def basic_rules_validators(self):
+        return BaseDataSource.basic_rule_validators() 
+                + [MyBasicRuleValidator, MyBasicRulesSetValidator]
+```
+
+It's not recommended to fully override `basic_rule_validators` as you'll lose the default validations.
+
+Again the framework will handle the rest like scheduling validation, calling the custom validators and storing the corresponding results.
```
leemthompo, line 130:
```suggestion
Again the framework will handle the rest: scheduling validation, calling the custom validators and storing the corresponding results.
```

docs/CONTRIBUTING.md
leemthompo: > further down I state, that it's discouraged to override them.

Maybe this should be a warning at the top of this section?

docs/CONTRIBUTING.md
```diff
@@ -12,6 +12,122 @@ Take a look at the [MongoDB connector](../connectors/sources/mongo.py) for more
...
+```python
+class MyDataSource(BaseDataSource):
+
+    @classmethod
+    def basic_rules_validators(self):
+        return BaseDataSource.basic_rule_validators() 
+                + [MyBasicRuleValidator, MyBasicRulesSetValidator]
+```
+
+It's not recommended to fully override `basic_rule_validators` as you'll lose the default validations.
```
leemthompo, line 128: Maybe this should be a warning at the top of this section?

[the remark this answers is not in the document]
leemthompo: Maybe this should be a warning at the top of this section?

leemthompo (commented):
This is really nice @timgrein! 👍  I have a couple of small questions, and then a number of copyedits that I'd like to see incorporated before merging.

There is the important question of overriding `basic_rule_validators` that I'd like @seanstory to sign off on.

docs/CONTRIBUTING.md
timgrein: Sounds way better, will adjust it!

docs/CONTRIBUTING.md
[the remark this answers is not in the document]
timgrein: Good point, will move it up!

docs/CONTRIBUTING.md
timgrein: Was adjusted with a recommendation.

docs/CONTRIBUTING.md
timgrein: Moved the warning to the top. WDYT @seanstory ?

docs/CONTRIBUTING.md
timgrein: I think we want to have a more technical description here as the description in the official docs is very ui-focused IMO.

leemthompo: Definitely! But if there's anything useful here that isn't in the main docs, I think we should make a note to add it there too.

timgrein: Hmm, I'm not too sure about that. The main docs look like they're targeting users of connectors and the docs here target connector developers, which need some information, what's expected from them to implement and what not. I think we can copy this description, when we've a separate "How to build a connector" section in the main docs. 

WDYT?

leemthompo: That sounds right 👍.

leemthompo (approved):
👍

seanstory (approved):
Good stuff, Tim.

PR #413, opened 2023-02-01
Add name and service_type method to data source classes

wangch079:
This PR adds clasa variables `name` and `service_type` to each data source class.

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -120,7 +120,7 @@ def run(args):
     if args.action == "list":
         logger.info("Registered connectors:")
         for source in get_data_sources(config):
-            logger.info(f"- {source.__doc__.strip()}")
+            logger.info(f"- {source.name().strip()}")
         logger.info("Bye")
         return 0
 
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -120,7 +120,7 @@ def run(args):
     if args.action == "list":
         logger.info("Registered connectors:")
         for source in get_data_sources(config):
-            logger.info(f"- {source.name().strip()}")
+            logger.info(f"- {source.name}")
         logger.info("Bye")
         return 0
 
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -107,21 +107,16 @@ class DataSourceConfiguration:
 class BaseDataSource:
     """Base class, defines a loose contract."""
 
+    name = None
+    service_type = None
+
     def __init__(self, configuration):
         self.configuration = configuration
         assert isinstance(self.configuration, DataSourceConfiguration)
         self.configuration.set_defaults(self.get_default_configuration())
 
     def __str__(self):
-        return f"Datasource `{self.__class__.name()}`"
-
-    @staticmethod
-    def name():
-        raise NotImplementedError
-
-    @staticmethod
-    def service_type():
-        raise NotImplementedError
+        return f"Datasource `{self.__class__.name}`"
 
     @classmethod
     def get_simple_configuration(cls):
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -113,7 +113,15 @@ class BaseDataSource:
         self.configuration.set_defaults(self.get_default_configuration())
 
     def __str__(self):
-        return f"Datasource `{self.__class__.__doc__}`"
+        return f"Datasource `{self.__class__.name()}`"
+
+    @staticmethod
+    def name():
+        raise NotImplementedError
+
+    @staticmethod
+    def service_type():
+        raise NotImplementedError
 
     @classmethod
     def get_simple_configuration(cls):
--- a/connectors/sources/abs.py
+++ b/connectors/sources/abs.py
@@ -44,6 +44,14 @@ class AzureBlobStorageDataSource(BaseDataSource):
         self.retry_count = self.configuration["retry_count"]
         self.concurrent_downloads = self.configuration["concurrent_downloads"]
 
+    @staticmethod
+    def name():
+        return "Azure Blob Storage"
+
+    @staticmethod
+    def service_type():
+        return "abs"
+
     def tweak_bulk_options(self, options):
         """Tweak bulk options as per concurrent downloads support by azure blob storage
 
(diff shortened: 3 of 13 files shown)


connectors/source.py
```diff
@@ -113,7 +113,15 @@ def __init__(self, configuration):
         self.configuration.set_defaults(self.get_default_configuration())
 
     def __str__(self):
-        return f"Datasource `{self.__class__.__doc__}`"
+        return f"Datasource `{self.__class__.name()}`"
+
+    @staticmethod
```
timgrein, line 118: Should those be `@classmethod`(s)? I think `@staticmethod`(s) are similar to top-level functions in a python module.

timgrein, line 118:
```suggestion
    @classmethod
```

Should this be `@classmethod`(s)? I think `@staticmethod`(s) are similar to top-level functions in a python module.

connectors/source.py
```diff
@@ -113,7 +113,15 @@ def __init__(self, configuration):
...
+    def name():
+        raise NotImplementedError
+
+    @staticmethod
```
timgrein, line 122:
```suggestion
    @classmethod
```

connectors/sources/abs.py
```diff
@@ -44,6 +44,14 @@ def __init__(self, configuration):
         self.retry_count = self.configuration["retry_count"]
         self.concurrent_downloads = self.configuration["concurrent_downloads"]
 
+    @staticmethod
```
timgrein, line 47:
```suggestion
    @classmethod
```

connectors/sources/abs.py
```diff
@@ -44,6 +44,14 @@ def __init__(self, configuration):
...
+    def name():
+        return "Azure Blob Storage"
+
+    @staticmethod
```
timgrein, line 51:
```suggestion
    @classmethod
```

connectors/sources/directory.py
```diff
@@ -28,6 +28,14 @@ def __init__(self, configuration):
         self.pattern = self.configuration["pattern"]
         self.enable_content_extraction = self.configuration["enable_content_extraction"]
 
+    @staticmethod
```
timgrein, line 31:
```suggestion
    @classmethod
```

connectors/sources/directory.py
```diff
@@ -28,6 +28,14 @@ def __init__(self, configuration):
...
+    def name():
+        return "Directory"
+
+    @staticmethod
```
timgrein, line 35:
```suggestion
    @classmethod
```

timgrein (commented):
Some comments about `@staticmethod` vs `@classmethod`. (Didn't mark every place, also forgot to include `cls` as first parameter for my suggestions.)

wangch079:
> Some comments about `@staticmethod` vs `@classmethod`. (Didn't mark every place, also forgot to include `cls` as first parameter for my suggestions.)

Why prefer `@classmethod` to `@staticmethod`? `@classmethod` will pass `cls` as the first parameter, which is not needed here as we don't intend to modify the class state.

timgrein:
> > Some comments about `@staticmethod` vs `@classmethod`. (Didn't mark every place, also forgot to include `cls` as first parameter for my suggestions.)
> 
> Why prefer `@classmethod` to `@staticmethod`? `@classmethod` will pass `cls` as the first parameter, which is not needed here as we don't intend to modify the class state.

I think `@staticmethod` is immutable via inheritance as it's not tied to the class, which is wrong in this case right (as we want to override it for new data source classes)?

timgrein (approved):
Translated "Immutable via Inheritance" wrong in my head, sorry! LGTM

connectors/source.py
tarekziade, line 118: why a static method? should be a class variable

```
class MySource:
    name = "name"
    service_type = "service type"
```

tarekziade (commented):
Great idea, Let's make those class variables

connectors/source.py
tarekziade: or if you want to force the class to define them, use class methods+property

```
>>> class BaseDataSource:
...     @classmethod
...     @property
...     def name(cls):
...         raise NotImplementedError
...
>>> class Source(BaseDataSource):
...     pass
...
>>>
>>> s = Source()
>>> s.name
  File "<stdin>", line 5, in name
NotImplementedError
>>> class Source2(BaseDataSource):
...     @classmethod
...     @property
...     def name(cls):
...         return 'source 2'
...
>>> s = Source2()
>>> s.name
'source 2'
```

tarekziade (changes requested):
just a nit

connectors/cli.py
```diff
@@ -132,7 +132,7 @@ def run(args):
     if args.action == "list":
         logger.info("Registered connectors:")
         for source in get_data_sources(config):
-            logger.info(f"- {source.__doc__.strip()}")
```
tarekziade, line 135: `__doc__` is different from `name`

it takes the class docstring. 

```
class X:
  """This is the docstring
  """
```

so if we drop this, no need to strip. raises the question about wether we want to keep a description or not

tarekziade (changes requested):

tarekziade:
> 



> > Some comments about `@staticmethod` vs `@classmethod`. (Didn't mark every place, also forgot to include `cls` as first parameter for my suggestions.)
> 
> Why prefer `@classmethod` to `@staticmethod`? `@classmethod` will pass `cls` as the first parameter, which is not needed here as we don't intend to modify the class state.

in that case, put the name value in the module as a global variable. Why would you put it in the class?

Static methods are to be used in meta-programming, or factories. It's not intended to be used for namespacing. If a class has to hold metadata information, we use class methods

wangch079:
>in that case, put the name value in the module as a global variable. Why would you put it in the class?

Agree class variable is more appropriate, but I want to force the data source class to define them.

connectors/source.py
wangch079: From [here](https://docs.python.org/3.11/library/functions.html#classmethod), it's not supported to use @classmethod and @property together since 3.11:

>_Changed in version 3.11_: Class methods can no longer wrap other [descriptors](https://docs.python.org/3.11/glossary.html#term-descriptor) such as [property()](https://docs.python.org/3.11/library/functions.html#property).

We are on 3.10, but it's definitely not recommended. I will use class variables instead, which means it doesn't force the child class to override it. But I think that's OK because 1) ultimately it's us to review the PR 2) Neither `name` nor `service_type` here is used in any critical path.

timgrein (approved):
LGTM

tarekziade (approved):

PR #414, opened 2023-02-01
Make mongodb configuration parity with ruby

artem-shelkovnikov:
This change introduces same options for MongoDB connector as in ruby.

Now configuration also has `user`, `password` and `direct_connection` parameters that work exactly like in Ruby.
Copy for all configurable fields was updated.

### DIFF
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -17,12 +17,28 @@ class MongoDataSource(BaseDataSource):
 
     def __init__(self, configuration):
         super().__init__(configuration=configuration)
+
+        client_params = {
+        }
+
+        host = self.configuration["host"]
+        user = self.configuration["user"]
+        password = self.configuration["password"]
+
+        if self.configuration["direct_connection"]:
+            client_params["directConnection"] = True
+
+        if len(user) > 0 or len(password) > 0:
+            client_params["username"] = user
+            client_params["password"] = password
+
+
+        print(f"host: {host}, params: {client_params}")
         self.client = AsyncIOMotorClient(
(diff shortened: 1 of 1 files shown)


connectors/sources/mongo.py
```diff
@@ -17,32 +17,50 @@ class MongoDataSource(BaseDataSource):
...
+        password = self.configuration["password"]
+
+        if self.configuration["direct_connection"]:
+            client_params["directConnection"] = True
+
+        if len(user) > 0 or len(password) > 0:
+            client_params["username"] = user
+            client_params["password"] = password
+
+        print(f"host: {host}, params: {client_params}")
```
timgrein, line 34: Should this be `logger.info(f"host: {host}, params: {client_params}")`?

timgrein (approved):
Small comment about a `print` statement, otherwise LGTM!

connectors/sources/mongo.py
```diff
@@ -17,32 +17,50 @@ class MongoDataSource(BaseDataSource):
 
     def __init__(self, configuration):
         super().__init__(configuration=configuration)
-        self.client = AsyncIOMotorClient(
-            self.configuration["host"],
-            directConnection=True,
-            connectTimeoutMS=120,
-            socketTimeoutMS=120,
```
timgrein, line 24: Will the timeouts be reintroduced at another place or removed all together?

connectors/sources/mongo.py
artem-shelkovnikov: That's a leftover from debugging, will delete it, good catch!

PR #418, opened 2023-02-02
[Testing] Execute functional tests for each filtering configuration

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/3736 ##

This PR adds the capability of automatically executing a functional test for each `filtering*.json` file for the corresponding source. This enables us to test filtering and also to do performance measurements for different filtering configurations.

Also clean up `mysql/fixture.py` by improving naming and logging.

#### Pre-Review Checklist

### DIFF
--- a/connectors/kibana.py
+++ b/connectors/kibana.py
@@ -90,6 +90,12 @@ DEFAULT_PIPELINE = {
 }
 
 
+def nullable_str(string):
+    if not string:
+        return None
+    return string
+
+
 # XXX simulating Kibana click-arounds
 async def prepare(service_type, index_name, config, filtering=None):
     if filtering is None:
@@ -254,7 +260,7 @@ def _parser():
     )
     parser.add_argument(
         "--filtering",
-        type=str,
+        type=nullable_str,
         help="Path to a json file containing an array of filters",
     )
 
@@ -262,9 +268,9 @@ def _parser():
 
 
 def _load_filtering(filtering_file_path):
-    if not os.path.exists(filtering_file_path):
+    if not filtering_file_path or not os.path.exists(filtering_file_path):
         logger.warn(
-            f"filtering file at '{filtering_file_path}' does not exist. Fallback to default filtering."
+            f"filtering file {('at' + filtering_file_path) if filtering_file_path else ''} does not exist. Fallback to default filtering."
         )
         filtering = DEFAULT_FILTERING
     else:
--- a/connectors/sources/tests/fixtures/mysql/fixture.py
+++ b/connectors/sources/tests/fixtures/mysql/fixture.py
@@ -7,12 +7,28 @@ import os
 import random
 import string
 
+from faker import Faker
 from mysql.connector import connect
 
+from connectors.logger import logger
+
 DATA_SIZE = os.environ.get("DATA_SIZE", "small").lower()
+
 DATABASE_NAME = "customerinfo"
+TABLE_PREFIX = "customer_"
+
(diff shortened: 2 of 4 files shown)


connectors/kibana.py
```diff
@@ -254,17 +260,17 @@ def _parser():
...
 
     return parser
 
 
 def _load_filtering(filtering_file_path):
-    if not os.path.exists(filtering_file_path):
+    if not filtering_file_path or not os.path.exists(filtering_file_path):
         logger.warn(
-            f"filtering file at '{filtering_file_path}' does not exist. Fallback to default filtering."
+            f"filtering file {('at' + filtering_file_path) if filtering_file_path else ''} does not exist. Fallback to default filtering."
```
artem-shelkovnikov, line 273: I'd move it to variable for readability.

connectors/kibana.py
```diff
@@ -90,6 +90,12 @@
 }
 
 
+def nullable_str(string):
```
artem-shelkovnikov, line 93: Maybe just `nullable`? Cause I don't see that it does anything string-specific

connectors/sources/tests/fixtures/mysql/fixture.py
```diff
@@ -22,44 +38,88 @@ def random_text(k=1024 * 20):
...
-        f"INSERT INTO customers_{table}"
-        + "(name, age, description) VALUES (%s, %s, %s)"
+class Customer:
+    def __init__(self, id_):
+        self.id_ = id_
+        self.name = fake.name()
+        self.description = BIG_TEXT
+
+    def to_row(self):
+        return self.id_, self.name, self.description
```
artem-shelkovnikov, line 48: Honestly the class feels rudimentary - I did not look too deep into it, but it looks like you just replace it with a function similar to:

```
def customer_row(id):
    return id, fake.name(), BIG_TEXT
```

connectors/tests/ftest.sh
```diff
@@ -37,34 +37,54 @@ fi
...
+
+  $PYTHON $ROOT_DIR/scripts/verify.py --index-name search-$NAME --service-type $NAME --size $SIZE
+}
+
+# run test once without filtering
+run_test
+
+for filtering in ./filtering*.json
+do
+  run_test $filtering
```
artem-shelkovnikov, line 86: Somehow it feels that `verify` can be different per filtering setup too - different number of entries should be present or some other criteria checked. Are you planning to add it in upcoming PRs?

artem-shelkovnikov (commented):
Overall looks good - curious to hear @tarekziade's opinion too :)

connectors/tests/ftest.sh
timgrein: I've created an issue an hour ago about addressing exactly this issue (and some others). I would've probably the time before 8.8 kick-off implementing the proposal...I'll double check, if there's room this week to address those. Feel free to comment on it.

connectors/sources/tests/fixtures/mysql/fixture.py
timgrein: Fair point, thought about externalising a more complex `Customer` class, which can be reused in different tests. But for this PR it's probably better to just us a one-liner function.

artem-shelkovnikov: TBH I think that functional tests should be as simple as possible. If they get complex, it's a good question - do we really want that complexity?

timgrein:
Closing this as we agreed to start a bigger refactoring around functional tests, which will make this change obsolete.

Issue #421, opened 2023-02-03
Directory connector breaks


tarekziade:
## Bug Description

```
[FMWK][10:17:53][CRITICAL] '_timestamp'
Traceback (most recent call last):
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/services/base.py", line 40, in run
    await self._run()
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/services/sync.py", line 137, in _run
    await syncs.join()
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/utils.py", line 336, in join
    await asyncio.gather(*self.tasks)
  File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/utils.py", line 307, in _callback
    raise task.exception()
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/services/sync.py", line 88, in _one_sync
    await connector.sync(es, self.idling, sync_now)
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/byoc.py", line 727, in sync
    result = await elastic_server.async_bulk(
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/byoei.py", line 467, in async_bulk
    await asyncio.gather(fetcher_task, bulker_task)
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/byoei.py", line 217, in run
    await self.get_docs(generator)
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/byoei.py", line 307, in get_docs
    await lazy_downloads.join()
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/utils.py", line 336, in join
    await asyncio.gather(*self.tasks)
  File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/utils.py", line 307, in _callback
    raise task.exception()
  File "/azuredl-nfs/connectors-python/lib/python3.10/site-packages/connectors/byoei.py", line 220, in _deferred_index
[...]

PR #422, opened 2023-02-03
Make the directory connector a full connector

tarekziade:
## Closes  https://github.com/elastic/connectors-python/issues/421

- [x] fix the code
- [x] add tests
- [x] rename it to System Directory
- [x] add a test fixture in the e2e test so it runs in the nightly
- [x] add more metadata (info from the file system)

#### Pre-Review Checklist

### DIFF
--- a/.buildkite/nightly.yml
+++ b/.buildkite/nightly.yml
@@ -37,3 +37,8 @@ steps:
       - ".buildkite/run_nigthly.sh postgresql"
     artifact_paths:
       - "perf8-report-*/**/*"
+  - label: "🔨 System Directory"
+    command:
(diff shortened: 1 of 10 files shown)


danajuratoni:
What about the tests for supported OS in https://www.elastic.co/support/matrix? Would be awesome if we could add these and release the connector client as Beta!

tarekziade:
> What about the tests for supported OS in https://www.elastic.co/support/matrix? Would be awesome if we could add these and release the connector client as Beta!

we only test on the OS used on Cloud in the CI. Having the whole collection adds a lot of work which is probably not worth it for this connector

Issue #428, opened 2023-02-08
Add a small e2e tests for all patches


tarekziade:
### Problem Description

Some patches break the nightly e2e and we don't see it until the nightly runs

### Proposed Solution

Buildkite should run one small e2e test alongside `make test` to catch this early

tarekziade:
Questions from Diana:
> define what is a small test

A small test is a minimal end-to-end test that runs with very few documents, as fast as possible. The intent is not to benchmark the framework but to make sure the syncing still works. So it's like our current end-to-end test but with a single connector (the faster) and very few docs.

> how likely is it for this test to break for reasons non-related to the code change

What do you have in mind? The goal is to intercept a change in the code that breaks the app.

> is it going to be included in the CI for all the PRs and block a merge

yes

PR #429, opened 2023-02-08
Update PULL_REQUEST_TEMPLATE.md

tarekziade:
be explicit about runing `make ftest`

#### Pre-Review Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,7 +15,7 @@ items that may help during the review.-->
 
 #### Pre-Review Checklist
 - [ ] Covered the changes with automated tests
-- [ ] Tested the changes locally
+- [ ] Tested the changes locally by running `make test` and `make ftest NAME=mysql`
 - [ ] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)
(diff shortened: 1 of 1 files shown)


wangch079 (approved):

seanstory:
Why don't we just have CI run tests, and block the PR if they fail?

tarekziade:
the CI runs the tests, but the e2e tests are running as nightlies because they take a while. the intent is to add a small one to improve this https://github.com/elastic/connectors-python/issues/428

But the expectation is that when you change the code, you test that it really works before you ask for a review.
and running `make ftest NAME=mysql` does a full e2e test that verifies the service can still index data in Elasticsearch

seanstory:
I'm looking at it from the angle that if we have a TODO list, we can automate it and _ensure_ it is done, or we can make a manual task list and _hope_ that it gets done. If we want PRs to have run `make ftest NAME=mysql` before they get merged, we should just automate that.

tarekziade:
well `Tested the changes locally` is pretty vague. Here we're hinting at what you should do as a dev.

But I don't expect you to run `make ftest` if you know you did not change the syncing mechanism in the patch.
Maybe we can say:

`Tested the changes locally -- make sure you try an e2e test if you change the sync mechanism or a data source class`

PR #430, opened 2023-02-08
Use None instead of empty string for datetime field

wangch079:
This PR updates the ftest connector data, to use `None` instead of `""` for datetime fields 

#### Pre-Review Checklist

### DIFF
--- a/connectors/kibana.py
+++ b/connectors/kibana.py
@@ -148,15 +148,15 @@ async def prepare(service_type, index_name, config, filtering=None):
             "language": "en",
             # Last sync
             "last_sync_status": None,
-            "last_sync_error": "",
-            "last_synced": "",
+            "last_sync_error": None,
+            "last_synced": None,
             # Written by connector on each operation,
             # used by Kibana to hint to user about status of connector
-            "last_seen": "",
+            "last_seen": None,
             # Date the connector was created
-            "created_at": "",
+            "created_at": None,
             # Date the connector was updated
-            "updated_at": "",
+            "updated_at": None,
             "filtering": filtering,
             # Scheduling intervals
             "scheduling": {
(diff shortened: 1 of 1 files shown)


tarekziade:
afaik `main` works without this change, I am curious to know why the problem appeared in your branch at https://github.com/elastic/connectors-python/pull/375 and this change fixes it?

wangch079:
> afaik `main` works without this change, I am curious to know why the problem appeared in your branch at #375 and this change fixes it?

It breaks because I introduced this:
https://github.com/elastic/connectors-python/blob/d19c7d923438a6550936e584c8557d9d5820f3a4/connectors/byoc.py#L447-L452

In `main`, it starts heartbeat and sends heartbeat periodically, which doesn't take `last_seen` into consideration. If we have multiple instances running, there will be multiple heartbeats sent:

https://github.com/elastic/connectors-python/blob/66af09843512b88f28f89746a45d45ade351095b/connectors/byoc.py#L528-L534

In #375, it checks `last_seen`, and this can make sure it works even with multiple instances:

https://github.com/elastic/connectors-python/blob/d19c7d923438a6550936e584c8557d9d5820f3a4/connectors/byoc.py#L495-L500

wangch079:
buildkite test this

tarekziade:
gotcha! in that case we probably want to replace

```
last_seen = self.get("last_seen")
```

by 

```
last_seen = self["last_seen"]
```

in your patch -- because I don't think the protocol allows anything but a date field, and I don't think it's mandatory

wangch079:
> because I don't think the protocol allows anything but a date field, 

That's right. But in our ftest, we don't have mappings:

https://github.com/elastic/connectors-python/blob/66af09843512b88f28f89746a45d45ade351095b/connectors/kibana.py#L177

>and I think it's mandatory

Not mandatory, it's `NULL` initially before the first heartbeat.

That's why I will prefer `self.get("last_seen")` instead of `self["last_seen"]`, which will raise `KeyError` if it's not available.

tarekziade (approved):

PR #433, opened 2023-02-08
Fix service stop code

artem-shelkovnikov:
As discussed with @tarekziade before, `loop.add_signal_handler(sig, lambda: asyncio.ensure_future(_shutdown(sig.name)))` is not the right way to handle the cancellation - ensure_future expects a synchronous function.

This patch does the following:

- Makes BaseService.stop a non-async function like it was before
- Modifies loop.add_signal_handler to properly call shutdown without a `lambda` statement.

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -106,7 +106,7 @@ async def _start_service(config, loop):
                     service.stop()
                 except Exception as e:
                     logger.warning(
-                        f"Failed to shutdown {service.__class__.__name__} due to an error: {e}"
+                        f"Failed to correctly shutdown {service.__class__.__name__} due to an error: {e}"
                     )
 
             return
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -104,8 +104,10 @@ async def _start_service(config, loop):
                 try:
                     logger.info(f"Shutdown {service.__class__.__name__}...")
                     service.stop()
-                except Exception e:
-                    logger.warning(f"Failed to shutdown {service.__class__.__name__} due to an error: {e}")
+                except Exception as e:
+                    logger.warning(
+                        f"Failed to shutdown {service.__class__.__name__} due to an error: {e}"
+                    )
 
             return
 
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -101,8 +101,12 @@ async def _start_service(config, loop):
         def _shutdown(sig_name):
             logger.info(f"Caught {sig_name}. Graceful shutdown.")
             for service in services:
-                logger.info(f"Shutdown {service.__class__.__name__}...")
-                service.stop()
+                try:
+                    logger.info(f"Shutdown {service.__class__.__name__}...")
+                    service.stop()
+                except Exception e:
+                    logger.warning(f"Failed to shutdown {service.__class__.__name__} due to an error: {e}")
+
             return
 
         loop.add_signal_handler(sig, functools.partial(_shutdown, sig.name))
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -98,14 +98,14 @@ async def _start_service(config, loop):
     services = [SyncService(config), JobCleanUpService(config)]
     for sig in (signal.SIGINT, signal.SIGTERM):
 
-        async def _shutdown(sig_name):
+        def _shutdown(sig_name):
             logger.info(f"Caught {sig_name}. Graceful shutdown.")
             for service in services:
                 logger.info(f"Shutdown {service.__class__.__name__}...")
-                await service.stop()
+                service.stop()
             return
 
-        loop.add_signal_handler(sig, lambda: asyncio.ensure_future(_shutdown(sig.name)))
+        loop.add_signal_handler(sig, functools.partial(_shutdown, sig.name))
 
     async def _run_service(service):
         if "PERF8" in os.environ:
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -22,7 +22,7 @@ class BaseService:
         self._sleeps = CancellableSleeps()
         self.errors = [0, time.time()]
 
-    async def stop(self):
+    def stop(self):
         self.running = False
         self._sleeps.cancel()
 
@@ -42,7 +42,7 @@ class BaseService:
             logger.critical(e, exc_info=True)
             self.raise_if_spurious(e)
         finally:
-            await self.stop()
+            self.stop()
 
     def raise_if_spurious(self, exception):
         errors, first = self.errors
--- a/connectors/services/sync.py
+++ b/connectors/services/sync.py
@@ -45,8 +45,8 @@ class SyncService(BaseService):
         self.connectors = None
         self.syncs = None
 
-    async def stop(self):
-        await super().stop()
+    def stop(self):
+        super().stop()
         if self.syncs is not None:
             self.syncs.cancel()
 
--- a/connectors/tests/test_job_cleanup.py
+++ b/connectors/tests/test_job_cleanup.py
@@ -50,7 +50,7 @@ def mock_sync_job(id="1", connector_id="1", index_name="index_name"):
 async def run_service_with_stop_after(service, stop_after):
     async def _terminate():
         await asyncio.sleep(stop_after)
-        await service.stop()
+        service.stop()
 
     await asyncio.gather(service.run(), _terminate())
 
--- a/connectors/tests/test_sync.py
+++ b/connectors/tests/test_sync.py
@@ -223,7 +223,7 @@ async def run_service_with_stop_after(service, stop_after=0):
             # but if stop_after is provided we want to
             # interrupt the service after the timeout
             await asyncio.sleep(stop_after)
-            await service.stop()
+            service.stop()
 
         await asyncio.sleep(0)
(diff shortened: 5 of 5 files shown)


connectors/cli.py
```diff
@@ -89,14 +89,14 @@ async def _start_service(config, loop):
     services = [SyncService(config), JobCleanUpService(config)]
     for sig in (signal.SIGINT, signal.SIGTERM):
 
-        async def _shutdown(sig_name):
+        def _shutdown(sig_name):
             logger.info(f"Caught {sig_name}. Graceful shutdown.")
             for service in services:
                 logger.info(f"Shutdown {service.__class__.__name__}...")
-                await service.stop()
+                service.stop()
```
tarekziade, line 96: maybe you want to try catch this, so if it fails we still `stop` others

tarekziade (commented):
That seems better. just two nits:
- why did you move Service.stop() to being async initially?
- let's make sure a service raising on stop does not prevent the shutdown to complete, by catching the error and converting it to a log line

connectors/cli.py
```diff
@@ -89,14 +89,14 @@ async def _start_service(config, loop):
...
             return
 
-        loop.add_signal_handler(sig, lambda: asyncio.ensure_future(_shutdown(sig.name)))
+        loop.add_signal_handler(sig, functools.partial(_shutdown, sig.name))
```
wangch079, line 99: Can anyone help me understand, why we use `functools.partial` here? We also used it here:

https://github.com/elastic/connectors-python/blob/e6a1e9463f0633748c3cdf7a65021e5950221fb2/connectors/services/sync.py#L139-L141

From what I understand [here](https://docs.python.org/3/library/functools.html#functools.partial), it just wraps the function and freezes some arguments. But I didn't see any additional arguments are used here.

tarekziade: two changes:

- the `_shutdown` function is now sync
- lambda is [late binding](https://docs.python-guide.org/writing/gotchas/#late-binding-closures), and `functools.partial` is not 

So if we keep `lambda` we will lookup `sig.name` at execution time. I think I missed this in reviews, this probably does not work as expected. The right one to use is `functools.partial`

wangch079: I see [here](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.add_signal_handler):

>Use [functools.partial()](https://docs.python.org/3/library/functools.html#functools.partial) [to pass keyword arguments](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio-pass-keywords) to callback.

If `_shutdown` doesn't accept any argument, I assume we can do this without `functools.partial`?
```python
loop.add_signal_handler(sig, _shutdown)
``` 

BTW, I checked out your branch and shutdown doesn't work as expected:

<details>
  <summary>Stacktrace</summary>

```  
[FMWK][17:51:33][INFO] Caught SIGINT. Graceful shutdown.
[FMWK][17:51:33][INFO] Shutdown SyncService...
[FMWK][17:51:33][INFO] Shutdown JobCleanUpService...
Exception in callback ConcurrentTasks._callback(result_callback=None)(<Task cancell...s/sync.py:54>>)
handle: <Handle ConcurrentTasks._callback(result_callback=None)(<Task cancell...s/sync.py:54>>)>
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/services/sync.py", line 95, in _one_sync
    await validate_filtering(
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/filtering/validation.py", line 29, in validate_filtering
    validation_result = await connector.source_klass(
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/source.py", line 159, in validate_filtering
    return await FilteringValidator(
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/filtering/validation.py", line 206, in validate
    filtering_validation_result += await validator.validate(advanced_rules)
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/sources/mysql.py", line 57, in validate
    return await self._remote_validation(advanced_rules, databases_to_filter)
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/utils.py", line 380, in func_to_execute
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/sources/mysql.py", line 65, in _remote_validation
    await self.source.ping()
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/sources/mysql.py", line 263, in ping
    self.connection_pool = await aiomysql.create_pool(**connection_string)
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/lib/python3.10/site-packages/aiomysql/pool.py", line 29, in _create_pool
    await pool._fill_free_pool(False)
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/lib/python3.10/site-packages/aiomysql/pool.py", line 182, in _fill_free_pool
    conn = await connect(echo=self._echo, loop=self._loop,
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/lib/python3.10/site-packages/aiomysql/connection.py", line 75, in _connect
    await conn._connect()
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/lib/python3.10/site-packages/aiomysql/connection.py", line 540, in _connect
    await self._request_authentication()
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/lib/python3.10/site-packages/aiomysql/connection.py", line 833, in _request_authentication
    auth_packet = await self._read_packet()
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/lib/python3.10/site-packages/aiomysql/connection.py", line 598, in _read_packet
    packet_header = await self._read_bytes(4)
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/lib/python3.10/site-packages/aiomysql/connection.py", line 646, in _read_bytes
    data = await self._reader.readexactly(num_bytes)
  File "/usr/local/Cellar/python@3.10/3.10.7/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/streams.py", line 708, in readexactly
    await self._wait_for_data('readexactly')
  File "/usr/local/Cellar/python@3.10/3.10.7/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/streams.py", line 502, in _wait_for_data
    await self._waiter
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

  File "/usr/local/Cellar/python@3.10/3.10.7/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/utils.py", line 306, in _callback
    if task.exception():
asyncio.exceptions.CancelledError
[FMWK][17:51:33][DEBUG] Sleep canceled
[FMWK][17:51:33][INFO] Bye
```
</details>

artem-shelkovnikov: > If _shutdown doesn't accept any argument, I assume we can do this without functools.partial?

Yes, I think that's the case.


> BTW, I checked out your branch and shutdown doesn't work as expected:

What were the steps that you took?

I checked it now and it works well for me.

artem-shelkovnikov: Hmmm I've checked and it indeed happens, but it's not actually connected to the change I've made - one more piece of logic is missing "except CancelledError" before validating filtering and such, I'll address it separately.

connectors/cli.py
artem-shelkovnikov: Good one, will do!

artem-shelkovnikov:
> why did you move Service.stop() to being async initially?

I think it was because for SyncService I had to do the shutdown gracefully by awaiting on something - in my case it was that I had a variable that held the `task` that was doing the sync, and I needed to do `task.cancel(); await task`.

connectors/cli.py
```diff
@@ -89,14 +89,20 @@ async def _start_service(config, loop):
     services = [SyncService(config), JobCleanUpService(config)]
     for sig in (signal.SIGINT, signal.SIGTERM):
 
-        async def _shutdown(sig_name):
+        def _shutdown(sig_name):
             logger.info(f"Caught {sig_name}. Graceful shutdown.")
             for service in services:
-                logger.info(f"Shutdown {service.__class__.__name__}...")
-                await service.stop()
+                try:
```
tarekziade, line 95: fly-by, you can use `service.name` now

tarekziade (approved):

connectors/cli.py
artem-shelkovnikov: Hmmm, I tried but it did not work:

```
  File "/Users/artemshelkovnikov/git_tree/connectors-py/connectors/cli.py", line 109, in _shutdown
    f"Failed to correctly shutdown {service.name} due to an error: {e}"
AttributeError: 'SyncService' object has no attribute 'name'
```

I don't think our Service classes have a `name` property/attribute.

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/457)|

### Questions ?

PR #434, opened 2023-02-08
Update ConnectorIndex

wangch079:
This PR made those changes:
1. move the query-building logic to `supported_connectors` method, as it's only used here.
2. Use method `supported_connectors` directly, instead of calling `build_docs_query` and then `get_all_docs`
3. change `update_filtering_validation` to only update `filtering` instead of the whole document.
4. add some wrapping methods to `ESIndex`
5. rename all `connectors_ids` to `connector_ids`

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -13,7 +13,7 @@ from copy import deepcopy
 from datetime import datetime, timezone
 from enum import Enum
 
-from connectors.es import DEFAULT_LANGUAGE, ESIndex, Mappings
+from connectors.es import ESIndex, Mappings
 from connectors.filtering.validation import ValidationTarget, validate_filtering
 from connectors.logger import logger
 from connectors.source import DataSourceConfiguration, get_source_klass
@@ -80,11 +80,10 @@ class DataSourceError(Exception):
 class ConnectorIndex(ESIndex):
     def __init__(self, elastic_config):
         logger.debug(f"ConnectorIndex connecting to {elastic_config['host']}")
-        # initilize ESIndex instance
+        # initialize ESIndex instance
         super().__init__(index_name=CONNECTORS_INDEX, elastic_config=elastic_config)
         # grab all bulk options
         self.bulk_options = elastic_config.get("bulk", {})
-        self.language_code = elastic_config.get("language_code", DEFAULT_LANGUAGE)
 
     async def save(self, connector):
         # we never update the configuration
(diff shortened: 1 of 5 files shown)


connectors/byoc.py
```diff
@@ -107,31 +106,34 @@ async def save(self, connector):
...
+        await self.update(doc_id=doc_id, doc={"last_seen": iso_utc()})
+
     async def update_filtering_validation(
         self, connector, validation_result, validation_target=ValidationTarget.ACTIVE
     ):
-        doc_to_update = deepcopy(connector.doc_source)
+        filtering = connector.filtering.to_list()
 
-        for filter_ in doc_to_update.get("filtering", []):
+        for filter_ in filtering:
```
vidok, line 117: Just wonder if know why `_` in `filter_` is needed?

wangch079: If you rename it to `filter`, IDE will complain:
```
Shadows built-in name 'filter' 
```

vidok: That's weird because I don't see any other `filter` names it shadows.

tarekziade: `filter` is a built-in class in Python itself -- so adding `_` is the right pattern

connectors/byoc.py
```diff
@@ -303,6 +296,9 @@ def get_filter(self, filter_state="active", domain=DEFAULT_DOMAIN):
             Filter(),
         )
 
+    def to_list(self):
```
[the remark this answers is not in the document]
artem-shelkovnikov: I also feel like `to_list()` method is not adding much value here - `list(connector.filtering)` looks more or less nice to me.

connectors/byoc.py
```diff
@@ -84,11 +84,10 @@ class DataSourceError(Exception):
 class ConnectorIndex(ESIndex):
     def __init__(self, elastic_config):
         logger.debug(f"ConnectorIndex connecting to {elastic_config['host']}")
-        # initilize ESIndex instance
+        # initialize ESIndex instance
```
artem-shelkovnikov, line 87: While I think it's okay to keep it in this PR, ideally renames/typo fixes should go in a different PR, especially if there are many of them - they just distract from the actual change IMO.

connectors/byoc.py
```diff
@@ -107,31 +106,34 @@ async def save(self, connector):
...
-            index=CONNECTORS_INDEX,
+            index=self.index_name,
             id=connector.id,
-            doc=doc_to_update,
+            doc={"filtering": filtering},
             retry_on_conflict=RETRY_ON_CONFLICT,
         )
 
-    def build_docs_query(self, native_service_types=None, connectors_ids=None):
+    async def supported_connectors(self, native_service_types=None, connector_ids=None):
```
artem-shelkovnikov, line 130: cc @vidok - I know you added the `build_docs_query` originally and passed the query inside a different method. What was the goal of original design and how do you feel about this change?

I'm not against it - it looks okay, but want to understand a bit more about it

connectors/tests/test_byoc.py
```diff
@@ -337,7 +331,86 @@ async def test_heartbeat(mock_responses, patch_logger):
...
-async def test_connectors_get_list(mock_responses):
+async def test_connectors_get_list_no_connector(mock_responses):
+    config = {"host": "http://nowhere.com:9200", "user": "tarek", "password": "blah"}
+    connectors = ConnectorIndex(config)
+    conns = []
+    async for connector in connectors.supported_connectors():
+        conns.append(connector)
+
+    assert len(conns) == 0
+    await connectors.close()
```
artem-shelkovnikov, line 342:
```suggestion
    await connectors.close()
    
    assert len(conns) == 0
```

I'd just flip them around cause asserts should be last in each individual test

artem-shelkovnikov (approved):
I think change looks okay, just have a few nits for it, so it's okay to merge and refactor more IMO. If you are strongly against it, please do the "Request changes" review on it.

connectors/byoc.py
wangch079: Initially, I wanted to rename `build_docs_query` to `supported_connectors_query`, because `build_docs_query` is like a general purpose query building method, but it's not.

Then I realized even `supported_connectors_query` was only used by the method `supported_connectors`.

So In `supported_connectors`, it simply does 2 things, build the query, and then call `get_all_docs`.

connectors/byoc.py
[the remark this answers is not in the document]
wangch079: > `list(connector.filtering)` looks more or less nice to me.

It's not `list(connector.filtering)`, but `list(connector.filtering.filtering)`

PR #435, opened 2023-02-08
Fix flaky tests

wangch079:
I noticed some flaky tests in [test_sync.py](https://github.com/elastic/connectors-python/blob/main/connectors/tests/test_sync.py), due to short `stop_after` specified:

https://github.com/elastic/connectors-python/blob/66af09843512b88f28f89746a45d45ade351095b/connectors/tests/test_sync.py#L220

This PR adds default `config_file` and `stop_after`, to make sure the service doesn't stop before the job completes.

#### Pre-Review Checklist

### DIFF
--- a/connectors/tests/test_sync.py
+++ b/connectors/tests/test_sync.py
@@ -217,7 +217,7 @@ async def run_service_with_stop_after(service, stop_after):
     await asyncio.gather(service.run(), _terminate())
 
 
-async def create_and_run_service(config_file, stop_after):
+async def create_and_run_service(config_file=CONFIG_FILE, stop_after=0.2):
     service = create_service(config_file)
     await run_service_with_stop_after(service, stop_after)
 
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov:
I'm not sure if this PR makes the tests non-flaky, though it might reduce the occurrence of random problems. I tried to address it in this PR: https://github.com/elastic/connectors-python/pull/436

Conceptual problem: most of tests in `test_sync.py` rely on the fact that the service does one full loop. When doing my refactoring I had to change code a bit and instead made the service stop after some delay (like 0.2 seconds), but it's problematic - the service itself can slow down due to random CPU blips or just load on same host. I tried to address it differently - if `stop_after` parameter is not sent to the test, that means that test expects service to make on full loop. I just make `service.running = False` in this case and it guarantees that even if test is slow for any reason, it will do the full run and the tests will not fail.

wangch079:
Close this as a duplicate PR of #436

PR #436, opened 2023-02-08
Change create_and_run_service to make it more reliable

artem-shelkovnikov:
Minor problem with current `test_sync` tests is that it relies on time when interrupting the service. If the service is too slow, then the logic that the test expects to trigger might make too much time. If the timeout is too big, however, the test will take too much time to run.

This change updates `run_service_with_stop_after` to work differently:

If `stop_after` is not provided or is 0, then the test will attempt to update `service.running` with `call_soon`. In essence, it will mean that the service will attempt to terminate without cancelling jobs. (TODO: should we change a BaseService method? Like having `stop` with a `force_terminate` parameter?

If `stop_after` is provided, the logic stays how it was - service is terminated in `stop_after` seconds while cancelling running work.

### DIFF
--- a/connectors/tests/test_sync.py
+++ b/connectors/tests/test_sync.py
@@ -215,7 +215,7 @@ async def run_service_with_stop_after(service, stop_after=0):
 
     async def _terminate():
         if stop_after == 0:
-            # so we actually want the loop the service
+            # so we actually want the service
             # to run current loop without interruption
             asyncio.get_event_loop().call_soon(_stop_running_service_without_cancelling)
         else:
--- a/connectors/tests/test_sync.py
+++ b/connectors/tests/test_sync.py
@@ -209,15 +209,27 @@ def create_service(config_file):
     return service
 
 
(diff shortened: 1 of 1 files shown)


wangch079:
Possible duplicate PR: https://github.com/elastic/connectors-python/pull/435

connectors/tests/test_sync.py
```diff
@@ -209,15 +209,27 @@ def create_service(config_file):
...
-async def run_service_with_stop_after(service, stop_after):
+async def run_service_with_stop_after(service, stop_after=0):
+    def _stop_running_service_without_cancelling():
+        service.running = False
+
     async def _terminate():
-        await asyncio.sleep(stop_after)
-        await service.stop()
+        if stop_after == 0:
+            # so we actually want the loop the service
```
timgrein, line 218:
```suggestion
            # so we actually want the service
```

timgrein (approved):
LGTM. Added small comment correction

PR #437, opened 2023-02-08
Terminate all services if one exits

artem-shelkovnikov:
Problem: await asyncio.gather(*[_run_service(service) for service in services]) waits for all services exiting before proceeding and terminating the CLI. 

What it means in practice is that if one of the services exists (let's say due to an error), then another one keeps running. So if SyncService is terminated due to an exception, JobCleanupService keeps running and deployment will become unhealthy without actually indicating that it is. 

Instead, we should wait for the first promise to return and cancel all the rest - this is what this PR actually does.

Side note - might be worth to extract this logic to a separate class/module and cover with tests to make sure this logic is not broken.

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -122,7 +122,12 @@ async def _start_service(config, loop):
         else:
             return await service.run()
 
-    await asyncio.gather(*[_run_service(service) for service in services])
+    done, pending = await asyncio.wait(
+        [_run_service(service) for service in services],
+        return_when=asyncio.FIRST_COMPLETED,
+    )
+    for task in pending:
+        task.cancel()
 
 
 def run(args):
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -22,7 +22,7 @@ from connectors.preflight_check import PreflightCheck
 from connectors.services.job_cleanup import JobCleanUpService
 from connectors.services.sync import SyncService
 from connectors.source import get_source_klasses
-from connectors.utils import get_event_loop
+from connectors.utils import get_event_loop, MultiService
 
 
 def _parser():
@@ -95,39 +95,17 @@ async def _start_service(config, loop):
         for sig in (signal.SIGINT, signal.SIGTERM):
             loop.remove_signal_handler(sig)
 
-    services = [SyncService(config), JobCleanUpService(config)]
+    multiservice = MultiService([SyncService(config), JobCleanUpService(config)])
     for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, lambda: asyncio.ensure_future(multiservice.shutdown(sig.name)))
 
-        def _shutdown(sig_name):
-            logger.info(f"Caught {sig_name}. Graceful shutdown.")
-            for service in services:
-                try:
-                    logger.info(f"Shutdown {service.__class__.__name__}...")
-                    service.stop()
-                except Exception as e:
-                    logger.warning(
-                        f"Failed to correctly shutdown {service.__class__.__name__} due to an error: {e}"
-                    )
+    if "PERF8" in os.environ:
+        import perf8
 
-            return
-
-        loop.add_signal_handler(sig, functools.partial(_shutdown, sig.name))
-
-    async def _run_service(service):
-        if "PERF8" in os.environ:
-            import perf8
-
-            async with perf8.measure():
-                return await service.run()
-        else:
-            return await service.run()
-
-    done, pending = await asyncio.wait(
-        [_run_service(service) for service in services],
-        return_when=asyncio.FIRST_COMPLETED,
-    )
-    for task in pending:
-        task.cancel()
+        async with perf8.measure():
+            return await multiservice.run()
+    else:
+        return await multiservice.run()
 
 
 def run(args):
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -19,10 +19,11 @@ from connectors import __version__
 from connectors.config import load_config
 from connectors.logger import logger, set_logger
 from connectors.preflight_check import PreflightCheck
+from connectors.services.base import MultiService
 from connectors.services.job_cleanup import JobCleanUpService
 from connectors.services.sync import SyncService
 from connectors.source import get_source_klasses
-from connectors.utils import MultiService, get_event_loop
+from connectors.utils import get_event_loop
 
 
 def _parser():
@@ -97,9 +98,7 @@ async def _start_service(config, loop):
 
     multiservice = MultiService([SyncService(config), JobCleanUpService(config)])
     for sig in (signal.SIGINT, signal.SIGTERM):
-        loop.add_signal_handler(
-            sig, functools.partial(multiservice.shutdown, sig.name)
-        )
+        loop.add_signal_handler(sig, functools.partial(multiservice.shutdown, sig.name))
 
     if "PERF8" in os.environ:
         import perf8
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -96,7 +96,7 @@ async def _start_service(config, loop):
         for sig in (signal.SIGINT, signal.SIGTERM):
             loop.remove_signal_handler(sig)
 
-    multiservice = MultiService([SyncService(config), JobCleanUpService(config)])
+    multiservice = MultiService(SyncService(config), JobCleanUpService(config))
     for sig in (signal.SIGINT, signal.SIGTERM):
         loop.add_signal_handler(sig, functools.partial(multiservice.shutdown, sig.name))
 
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -22,7 +22,7 @@ from connectors.preflight_check import PreflightCheck
 from connectors.services.job_cleanup import JobCleanUpService
 from connectors.services.sync import SyncService
 from connectors.source import get_source_klasses
-from connectors.utils import get_event_loop, MultiService
+from connectors.utils import MultiService, get_event_loop
 
 
 def _parser():
@@ -97,7 +97,9 @@ async def _start_service(config, loop):
 
     multiservice = MultiService([SyncService(config), JobCleanUpService(config)])
     for sig in (signal.SIGINT, signal.SIGTERM):
-        loop.add_signal_handler(sig, lambda: asyncio.ensure_future(multiservice.shutdown(sig.name)))
+        loop.add_signal_handler(
+            sig, functools.partial(multiservice.shutdown, sig.name)
+        )
 
     if "PERF8" in os.environ:
         import perf8
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -3,6 +3,7 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+import asyncio
 import time
 
 from connectors.logger import logger
@@ -59,3 +60,25 @@ class BaseService:
 
         self.errors[0] = errors
         self.errors[1] = first
+
+
+class MultiService:
+    def __init__(self, services):
+        self._services = services
+
+    async def run(self):
+        tasks = [asyncio.create_task(service.run()) for service in self._services]
+
+        done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+
+        for task in pending:
+            task.cancel()
+            await task
+
+    def shutdown(self, sig):
+        logger.info(f"Caught {sig}. Graceful shutdown.")
+
+        for service in self._services:
+            logger.debug(f"Shutting down {service.__class__.__name__}...")
+            service.stop()
+            logger.debug(f"Done shutting down {service.__class__.__name__}...")
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -72,8 +72,8 @@ class MultiService:
         done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
 
         for task in pending:
-            task.cancel()
             try:
+                task.cancel()
                 await task
             except asyncio.CancelledError:
                 logger.error(f"Service did not handle cancellation gracefully.")
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -63,7 +63,7 @@ class BaseService:
 
 
 class MultiService:
-    def __init__(self, services):
+    def __init__(self, *services):
         self._services = services
 
     async def run(self):
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -76,7 +76,7 @@ class MultiService:
             try:
                 await task
             except asyncio.CancelledError:
-                logger.error(f"Service did not handle cancellation gracefully.")
+                logger.error("Service did not handle cancellation gracefully.")
 
     def shutdown(self, sig):
         logger.info(f"Caught {sig}. Graceful shutdown.")
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -72,8 +72,8 @@ class MultiService:
         done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
 
         for task in pending:
+            task.cancel()
             try:
-                task.cancel()
                 await task
             except asyncio.CancelledError:
                 logger.error(f"Service did not handle cancellation gracefully.")
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -73,7 +73,10 @@ class MultiService:
 
         for task in pending:
             task.cancel()
-            await task
+            try:
+                await task
+            except asyncio.CancelledError:
+                logger.error(f"Service did not handle cancellation gracefully.")
 
     def shutdown(self, sig):
         logger.info(f"Caught {sig}. Graceful shutdown.")
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -412,3 +412,25 @@ def str2e(name, enum_class):
         return None
 
     return enum_class[name]
+
+class MultiService:
+    def __init__(self, services):
+        self._services = services
+
+    async def run(self):
+        tasks = [asyncio.create_task(service.run()) for service in self._services]
+
+        done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+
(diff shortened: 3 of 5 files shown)


timgrein:
Looks good, but would also like to see a test as this is will be pretty important to work correctly.

artem-shelkovnikov:
@timgrein fair enough, I wanted to write a test but it's a little hard now. I thought of encapsulating this code in a small class and testing it instead, what do you think?

timgrein:
Makes sense, you can also just put it as a method in `utils.py`. Is probably generic enough to work with arbitrary co-routines IMO.

tarekziade:
If the bug occurs when one service raises an error, it's pretty straightforward to add a test that simulates this, by mocking `Service.run` with:

```
async run(self):
    await asyncio.sleep(.1)
    raise Exception("Muhahaha")
```

and making sure the test wraps up and exits 

happy to help

connectors/cli.py
```diff
@@ -107,7 +107,12 @@ async def _run_service(service):
         else:
             return await service.run()
 
-    await asyncio.gather(*[_run_service(service) for service in services])
+    done, pending = await asyncio.wait(
+        [_run_service(service) for service in services],
+        return_when=asyncio.FIRST_COMPLETED,
```
wangch079, line 112:
```suggestion
        return_when=asyncio.FIRST_EXCEPTION,
```

Should we use `FIRST_EXCEPTION` instead? 

When shutdown happens, it will call `_shutdown` to stop all services. when the first service stops completely, `asyncio.wait` will return, and here it will try to cancel the rest services, which are still completing.

artem-shelkovnikov:
Depends on https://github.com/elastic/connectors-python/pull/433 before proceeding

connectors/utils.py
```diff
@@ -401,3 +401,26 @@ async def func_to_execute(*args, **kwargs):
 
 def e2str(entry):
     return entry.name.lower()
+
+
+class MultiService:
```
tarekziade, line 406: I would move this class to `connectors/services/base.py`

connectors/utils.py
```diff
@@ -401,3 +401,26 @@ async def func_to_execute(*args, **kwargs):
...
+        self._services = services
+
+    async def run(self):
+        tasks = [asyncio.create_task(service.run()) for service in self._services]
+
+        done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+
+        for task in pending:
+            task.cancel()
+            await task
```
tarekziade, line 417: you want:

```
try:
    await task
    logger.error('...')
```

to make sure you don't break in this loop if the task did not handle the cancellation

connectors/utils.py
```diff
@@ -401,3 +401,26 @@ async def func_to_execute(*args, **kwargs):
...
+
+    async def shutdown(self, sig):
+        logger.info(f"Caught {sig}. Graceful shutdown.")
+        for service in self._services:
+            logger.debug(f"Shutting down {service.__class__.__name__}...")
```
tarekziade, line 422: `BaseService` has  a `name` attribute now

tarekziade (commented):
just nits

tarekziade:
oops the new test freezes in buildkite https://buildkite.com/elastic/connectors-python/builds/2327#01863667-1044-42a3-a187-c47f96426ee2

connectors/utils.py
```diff
@@ -401,3 +401,26 @@ async def func_to_execute(*args, **kwargs):
 
 def e2str(entry):
     return entry.name.lower()
+
+
+class MultiService:
+    def __init__(self, services):
```
tarekziade, line 407: nit:

```
    def __init__(self, *services):
```

so we can pass:

```
m = MultiService(s1, s2, s3)
```

connectors/utils.py
```diff
@@ -401,3 +401,26 @@ async def func_to_execute(*args, **kwargs):
...
+        self._services = services
+
+    async def run(self):
```
tarekziade, line 410: do you want the function to return a list of done, pending or some kind of way to know that one died

this will be useful to return the correct exit code or re-raise an error

connectors/cli.py
artem-shelkovnikov: cc @tarekziade what do you think about it?

I've added tests to test both cases - when service exits with an exception or exits normally, and FIRST_COMPLETED works well, but I'm not 100% sure about it here.

connectors/utils.py
artem-shelkovnikov: Hmmm I cannot find it, can you point to the code that defines it?

connectors/utils.py
artem-shelkovnikov: I can return either stats or an exit code or raise an exception if any exited due to an exception, what would be the best in your opinion?

I feel like raising an exception is the simplest.

tarekziade: > I feel like raising an exception is the simplest.
sounds good!

connectors/cli.py
tarekziade: the documentation says:

- `FIRST_COMPLETED` - The function will return when any future finishes or is cancelled
- `FIRST_EXCEPTION` - The function will return when any future finishes by raising an exception. If no future raises an exception then it is equivalent to ALL_COMPLETED.

So yeah `COMPLETED` is valid.

here a full working example:

```
import asyncio

async def one():
    await asyncio.sleep(2)


async def two():
    await asyncio.sleep(1)
    raise Exception("two failed")

async def run():
    t1 = asyncio.create_task(one())
    t2 = asyncio.create_task(two())
    done, pending = await asyncio.wait([t1, t2], return_when=asyncio.FIRST_COMPLETED)

    for t in pending:
        t.cancel()
        try:
            await t
            pass

    for t in done:
        exc = t.exception()
        # raise first one
        if exc is not None:
            raise exc

    return done, pending

print(asyncio.run(run()))
```

wangch079: All the services in this repo are long-running services, so there are only 2 ways of existing:
1. service shutdown (expected)
2. some error happens (unexpected)

If it's `FIRST_EXCEPTION`:
For scenario 1), it's equivalent to `ALL_COMPLETED`, and `pending` will be empty, so it's no-op.
For scenario 2), when any task raise error, `asyncio.wait` will return and all the `pending` taks will be canceled.

If it's `FIRST_COMPLETED`:

For scenario 1), all tasks will be canceled (by setting `running` to `false`), and tasks will complete at different times. The first completed task will cause the rest to be `canceled`, I don't think this is desired.
For scenario 2), I assume raising an exception is not considered `completed`, so it will leave the rest running?

That's why I feel `FIRST_EXCEPTION` is what we needed here. But all these analyses are based on the `asyncio.wait` doc, and I have never tested it.

connectors/utils.py
tarekziade: oh crap, my mistake, I was thinking about the DataSource classes :)

wangch079: @tarekziade I guess you confuse `BaseService ` with `BaseDataSource`

connectors/cli.py
tarekziade: `FIRST_COMPLETED` is triggered on exception, where an example with your two scenarios:


```
import asyncio

async def one():
    await asyncio.sleep(2)


async def two():
    await asyncio.sleep(1)
    raise Exception("two failed")

async def run(services):
    tasks = [asyncio.create_task(s) for s in services]
    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)

    for t in pending:
        t.cancel()
        try:
            await t
            pass

    for t in done:
        exc = t.exception()
        # raise first one
        if exc is not None:
            raise exc

    return done, pending


# scenario 1
scenario_1 = one(), two()
try:
    print(asyncio.run(run(scenario_1)))
except Exception as e:
    print(e)

# scenario 2
scenario_2 = one(), one()
try:
    print(asyncio.run(run(scenario_2)))
except Exception:
    print(e)
```

connectors/services/base.py
```diff
@@ -59,3 +60,28 @@ def raise_if_spurious(self, exception):
...
+
+class MultiService:
+    def __init__(self, services):
+        self._services = services
+
+    async def run(self):
+        tasks = [asyncio.create_task(service.run()) for service in self._services]
+
+        done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
+
```
wangch079, line 73: It will be helpful to add some logs, what tasks complete, and what are pending

connectors/utils.py
tarekziade: yes! (we could add `name` to the service too I suppose)

connectors/services/base.py
```diff
@@ -59,3 +60,28 @@ def raise_if_spurious(self, exception):
...
+        for task in pending:
+            try:
+                task.cancel()
```
tarekziade, line 76: that line should be before the `try`

connectors/cli.py
wangch079: If you run this, you will see under a normal shutdown, only the first service completes successfully and the rest are canceled:
```python
import asyncio

class BaseService:
    def __init__(self):
        self.running = False

    async def run(self):
        print(f"Start {self.__class__.__name__}")
        self.running = True
        while self.running:
            await self._run()
        print(f"Complete {self.__class__.__name__}")

    def stop(self):
        print(f"Stop {self.__class__.__name__}")
        self.running = False

class ServiceOne(BaseService):
    async def _run(self):
        await asyncio.sleep(2)

class ServiceTwo(BaseService):
    async def _run(self):
        await asyncio.sleep(6)


async def run(services):
    tasks = [asyncio.create_task(s.run()) for s in services]
    # simulate SIGTERM to stop all services
    def _stop_services():
        for s in services:
            s.stop()
    asyncio.get_event_loop().call_later(1, _stop_services)
    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)

    for t in pending:
        t.cancel()
        try:
            await t
            pass

    for t in done:
        exc = t.exception()
        # raise first one
        if exc is not None:
            raise exc

    return done, pending


services = [ServiceOne(), ServiceTwo()]
try:
    print(asyncio.run(run(services)))
except Exception as e:
    print(e)

```

You will see this:
```shell
chenhui.wang@SG-Chenhui Downloads % python3 asyncio_test.py
Start ServiceOne
Start ServiceTwo
Stop ServiceOne
Stop ServiceTwo
Complete ServiceOne
({<Task finished name='Task-2' coro=<BaseService.run() done, defined at /Users/chenhui.wang/Downloads/asyncio_test.py:7> result=None>}, {<Task cancelled name='Task-3' coro=<BaseService.run() done, defined at /Users/chenhui.wang/Downloads/asyncio_test.py:7>>})
```

tarekziade: - you did not fully simulate SIGTERM in your example
- `ServiceTwo` gets properly stopped, but we just ignore the `CancelledError`in the example

If you want to have a `Complete` called on cancellation, you need to add it.

Example:

```
import asyncio
import os
import signal


class BaseService:
    def __init__(self):
        self.running = False

    async def run(self):
        print(f"Start {self.__class__.__name__}")
        try:
            self.running = True
            while self.running:
                await self._run()
        finally:
            print(f"Complete {self.__class__.__name__}")

    def stop(self):
        print(f"Stop {self.__class__.__name__}")
        self.running = False

class ServiceOne(BaseService):
    async def _run(self):
        await asyncio.sleep(2)

class ServiceTwo(BaseService):
    async def _run(self):
        await asyncio.sleep(6)


async def run(services):

    def _shutdown():
        for s in services:
            s.stop()

    asyncio.get_event_loop().add_signal_handler(signal.SIGTERM, _shutdown)

    tasks = [asyncio.create_task(s.run()) for s in services]

    # simulate SIGTERM to stop all services
    def _stop_services():
        os.kill(os.getpid(), signal.SIGTERM)

    asyncio.get_event_loop().call_later(1, _stop_services)
    done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)

    for t in pending:
        t.cancel()
        try:
            await t
            pass

    for t in done:
        exc = t.exception()
        # raise first one
        if exc is not None:
            raise exc

    return done, pending


services = [ServiceOne(), ServiceTwo()]
try:
    print(asyncio.run(run(services)))
except Exception as e:
    print(e)

```

One pattern you can have is to add a `tear_down` method `BaseService` for cleanups if you want

tarekziade: example of integrating cancellation at the base class level:

```
class BaseService:
    def __init__(self):
        self.running = False

    async def run(self):
        print(f"Start {self.__class__.__name__}")
        try:
            self.running = True
            while self.running:
                await self._run()

            print(f"Complete {self.__class__.__name__}")
            self.cancelled()

    def _canceled(self):
        pass.  # can be overridden by concrete implementations

    def cancelled(self):
        print(f"Cancelled {self.__class__.__name__}")
        self._cancelled()

    def stop(self):
        print(f"Stop {self.__class__.__name__}")
        self.running = False
```

connectors/services/base.py
artem-shelkovnikov: I'm actually not sure if it's possible - we will get some ids of tasks, but it's hard to correlate task id to a specific action that the task was doing.

We can try to do it in a later PR

connectors/cli.py
```diff
@@ -95,34 +96,17 @@ async def _start_service(config, loop):
         for sig in (signal.SIGINT, signal.SIGTERM):
             loop.remove_signal_handler(sig)
 
-    services = [SyncService(config), JobCleanUpService(config)]
+    multiservice = MultiService([SyncService(config), JobCleanUpService(config)])
     for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(multiservice.shutdown, sig.name))
```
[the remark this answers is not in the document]
artem-shelkovnikov: Should it be though? It requires access to `loop` that is running and `cli.py` already has it, so I'd just leave the code as is.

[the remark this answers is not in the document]
tarekziade: The loop is available everywhere once your app starts, with `asyncio.get_running_loop` https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop

connectors/utils.py
artem-shelkovnikov: Hmmm it's actually not that simple looks like. It's really hard to distinguish the situation of "service died" vs "service stopped cause of Cancellation and gracefully handled it". It probably will require extending service classes to store some metadata about their run - properties like "failed", "cancelled" and so so that we could ask the services whether they ran successfully or not.

Shall we discuss it more in a different PR?

tarekziade: >  It's really hard to distinguish the situation of "service died" vs "service stopped cause of Cancellation and gracefully handled it"

unless we specifically ask the Service class to handle `CancelledError` like in https://github.com/elastic/connectors-python/pull/437#discussion_r1103153507 and have a `canceled` flag

connectors/cli.py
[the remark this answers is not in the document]
artem-shelkovnikov: I know, but it does not feel like adding signal handlers should leave in multiple places of the project.

connectors/utils.py
artem-shelkovnikov: Indeed, that's what I meant but it feels like it's better to handle it in a separate PR, what do you think?

connectors/cli.py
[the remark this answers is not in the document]
tarekziade: Indeed, this is why I was thinking about this.

if we look at this code here: https://github.com/elastic/connectors-python/blob/41a54228b0181f4a538707a58572e4f16beec5cd/connectors/cli.py#L89-L101

There's a pattern where we give a service the handling of signals, and that code is duplicated. If the Service class internalizes it, the block could become:

```python
    preflight = PreflightCheck(config)
    with preflight.handle_shutdown():
        if not await preflight.run():
            return -1

    multiservice = MultiService(SyncService(config), JobCleanUpService(config))
    with multiservice.handle_shutdown():
        return await multiservice.run()
```

and simplified with:

```python

    with PreflightCheck(config) as service:
        if not await service.run():
            return -1

    with MultiService(SyncService(config), JobCleanUpService(config)) as service:
        return await service.run()
```

but that's just food for thoughts for another PR, not required!

connectors/utils.py
tarekziade: absolutely! small PRs FTW

connectors/cli.py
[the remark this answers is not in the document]
artem-shelkovnikov: My main concern is that I'm just not 100% sure what happens, if you add multiple signal handlers at the same time to different services.

Imagine, for example, by mistake the following logic is introduced:

```
multiservice = MultiService(SyncService(config), JobCleanUpService(config))
with multiservice.handle_shutdown():
     return await multiservice.run()

# Somewhere deep inside the framework
     loop.add_signal_handler(sig, functools.partial(logger.debug, "something")) 

```

Will in this case both be handled or only the last handler called?

tarekziade (approved):

connectors/cli.py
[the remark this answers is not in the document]
tarekziade: no you should not have mutiple services handling the same signals. This is why implementing it at the base class level solves it -- the base class can deal with this unicity and make sure the handler cannot be registered twice. The current code is doing this manually.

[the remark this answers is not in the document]
tarekziade: > PreflightCheck is not a service though

it's implemented as a service right now (same APIs)

> To me it becomes more confusing - now it's all in one file, with the change it'll be in different namespaces.

I think it boils down to:

-  the fact that the code that registers and unregisters signals is currently duplicated because we have a two-phase startup. 
- defining the responsibility of each class. If the `Service` class is the one responsible for a shutdown, it in its scope 

Maybe the confusion is around what `MultiService` is. Maybe it's not a service and should have its own implementation where it takes care of the preflight check and starting services

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/463)|

### Questions ?

connectors/cli.py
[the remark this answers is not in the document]
artem-shelkovnikov: > Maybe the confusion is around what MultiService is. Maybe it's not a service and should have its own implementation where it takes care of the preflight check and starting services

Hmmm that makes sense, I'll see how things can move around so that there's only one registration in the whole code!

PR #439, opened 2023-02-08
Clean up config field names for all sources

sphilipse:
This cleans up the labels of configuration fields in various sources, making them more easily understandable and consistent.

### DIFF
--- a/connectors/sources/abs.py
+++ b/connectors/sources/abs.py
@@ -85,24 +85,19 @@ class AzureBlobStorageDataSource(BaseDataSource):
                 "label": "Azure Blob Storage blob endpoint",
                 "type": "str",
             },
-            "connector_name": {
-                "value": "Azure Blob Storage Connector",
-                "label": "Friendly name for the connector",
-                "type": "str",
-            },
             "enable_content_extraction": {
                 "value": DEFAULT_CONTENT_EXTRACTION,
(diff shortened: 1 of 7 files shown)


artem-shelkovnikov (approved):

connectors/sources/abs.py
```diff
@@ -85,24 +85,19 @@ def get_default_configuration(cls):
...
             },
-            "connector_name": {
-                "value": "Azure Blob Storage Connector",
-                "label": "Friendly name for the connector",
-                "type": "str",
-            },
             "enable_content_extraction": {
                 "value": DEFAULT_CONTENT_EXTRACTION,
-                "label": "Flag to check if content extraction is enabled or not",
+                "label": "Enable content extraction",
```
danajuratoni, line 90: This is entered as boolean. Can we add `(true/false)` to the name as we did for the other booleans?

connectors/sources/directory.py
```diff
@@ -46,7 +46,7 @@ def get_default_configuration(cls):
             },
             "enable_content_extraction": {
                 "value": DEFAULT_CONTENT_EXTRACTION,
-                "label": "Flag to check if content extraction is enabled or not",
+                "label": "Enable content extraction",
```
danajuratoni, line 49: + `(true/false)`

danajuratoni:
Looks good!

PR #440, opened 2023-02-08
Implement Pipeline as a derived class of UserDict

wangch079:
This PR implements the class `Pipeline` as a derived class of `UserDict`, which provides default values for:
1. name
2. extract_binary_content
3. reduce_whitespace
4. run_ml_inference

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -322,16 +322,34 @@ class Filter(dict):
 
 class PipelineSettings:
     def __init__(self, pipeline):
-        self.name = pipeline.get("name", "ent-search-generic-ingestion")
-        self.extract_binary_content = pipeline.get("extract_binary_content", True)
-        self.reduce_whitespace = pipeline.get("reduce_whitespace", True)
-        self.run_ml_inference = pipeline.get("run_ml_inference", True)
+        self._pipeline = pipeline
+        if self._pipeline is None:
+            self._pipeline = {}
+
+    @property
+    def name(self):
+        return self._pipeline.get("name", "ent-search-generic-ingestion")
+
+    @property
+    def extract_binary_content(self):
+        return self._pipeline.get("extract_binary_content", True)
+
+    @property
+    def reduce_whitespace(self):
+        return self._pipeline.get("reduce_whitespace", True)
+
+    @property
+    def run_ml_inference(self):
+        return self._pipeline.get("run_ml_inference", True)
 
     def __repr__(self):
         return (
             f"Pipeline {self.name} <binary: {self.extract_binary_content}, "
             f"whitespace {self.reduce_whitespace}, ml inference {self.run_ml_inference}>"
         )
+
+    def to_dict(self):
+        return dict(self._pipeline)
 
 
 class Features:
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -8,6 +8,7 @@ Implementation of BYOC protocol.
 """
 import asyncio
 import time
+from collections import UserDict
 from copy import deepcopy
 from datetime import datetime, timezone
 from enum import Enum
@@ -20,7 +21,6 @@ from connectors.utils import e2str, iso_utc, next_run
 
 CONNECTORS_INDEX = ".elastic-connectors"
 JOBS_INDEX = ".elastic-connectors-sync-jobs"
-PIPELINE = "ent-search-generic-ingestion"
 RETRY_ON_CONFLICT = 3
 SYNC_DISABLED = -1
 
@@ -320,36 +320,19 @@ class Filter(dict):
         return len(self.advanced_rules) > 0
 
 
-class PipelineSettings:
-    def __init__(self, pipeline):
-        self._pipeline = pipeline
-        if self._pipeline is None:
-            self._pipeline = {}
+PIPELINE_DEFAULT = {
+    "name": "ent-search-generic-ingestion",
+    "extract_binary_content": True,
+    "reduce_whitespace": True,
+    "run_ml_inference": True,
+}
 
-    @property
-    def name(self):
-        return self._pipeline.get("name", "ent-search-generic-ingestion")
 
-    @property
-    def extract_binary_content(self):
-        return self._pipeline.get("extract_binary_content", True)
-
-    @property
-    def reduce_whitespace(self):
-        return self._pipeline.get("reduce_whitespace", True)
-
-    @property
-    def run_ml_inference(self):
-        return self._pipeline.get("run_ml_inference", True)
-
-    def __repr__(self):
-        return (
-            f"Pipeline {self.name} <binary: {self.extract_binary_content}, "
-            f"whitespace {self.reduce_whitespace}, ml inference {self.run_ml_inference}>"
-        )
-
-    def to_dict(self):
-        return dict(self._pipeline)
+class Pipeline(UserDict):
+    def __init__(self, data):
+        default = PIPELINE_DEFAULT.copy()
+        default.update(data)
+        super().__init__(default)
 
 
 class Features:
@@ -453,7 +436,7 @@ class Connector:
         self.index_name = doc_source["index_name"]
         self._configuration = DataSourceConfiguration(doc_source["configuration"])
         self.scheduling = doc_source["scheduling"]
-        self.pipeline = PipelineSettings(doc_source.get("pipeline", {}))
+        self.pipeline = Pipeline(doc_source.get("pipeline", {}))
         self._dirty = True
         self._filtering = Filtering(doc_source.get("filtering", []))
         self.language_code = doc_source["language"]
@@ -621,9 +604,9 @@ class Connector:
 
         async for doc, lazy_download in data_provider.get_docs(filtering=filtering):
             # adapt doc for pipeline settings
-            doc["_extract_binary_content"] = self.pipeline.extract_binary_content
-            doc["_reduce_whitespace"] = self.pipeline.reduce_whitespace
-            doc["_run_ml_inference"] = self.pipeline.run_ml_inference
+            doc["_extract_binary_content"] = self.pipeline["extract_binary_content"]
+            doc["_reduce_whitespace"] = self.pipeline["reduce_whitespace"]
+            doc["_run_ml_inference"] = self.pipeline["run_ml_inference"]
             yield doc, lazy_download
 
     async def prepare(self, config):
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -69,7 +69,7 @@ class Bulker:
     - `max_concurrency` -- a maximum number of concurrent bulk requests
 
     Extra options:
-    - `pipeline_settings` -- ingest pipeline settings to pass to the bulk API
+    - `pipeline` -- ingest pipeline settings to pass to the bulk API
     """
 
     def __init__(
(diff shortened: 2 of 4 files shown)


tarekziade (commented):
What is the goal of using properties instead of instance variables ?

For adding `to_dict`, who will use it? I don't see changes in the patch that uses this function (or tests)

wangch079:
> What is the goal of using properties instead of instance variables ?

I don't have any preference. I want to keep the original `pipeline`, which is to be used in `to_dict`.

If we use instance variables, e.g.

```python
self.extract_binary_content = pipeline.get("extract_binary_content", True)
```

and we define `to_dict` like this:
```python
def to_dict(self)
def to_dict(self):
    return {
        "extract_binary_content": self.extract_binary_content
    }
```

we won't know whether this value is from the original `pipeline` or from the default value.

> For adding `to_dict`, who will use it? I don't see changes in the patch that uses this function (or tests)

It will be used when creating a new job, where the connector pipeline setting will be copied to the job document. This will be implemented in a separate PR after this is merged.

I will add some tests

wangch079:
Hi @tarekziade , I have added test here: https://github.com/elastic/connectors-python/pull/440/commits/c12ff983dd4d82f9925fc951163fd1773814a9fb

tarekziade:
> > What is the goal of using properties instead of instance variables ?
> 
> I don't have any preference. I want to keep the original `pipeline`, which is to be used in `to_dict`.

so if there's no goal let's keep it simple and not add properties. properties are useful only when you need to do some dynamic stuff when the value is read. It seems overkill here.


> > For adding `to_dict`, who will use it? I don't see changes in the patch that uses this function (or tests)
> 
> It will be used when creating a new job, where the connector pipeline setting will be copied to the job document. This will be implemented in a separate PR after this is merged.

Ok, is the goal is to convert it to something you want to serialize in JSON? In that case I would rename it `to_json`

something that is also not clear here is : do you want to keep the data from the `pipeline` argument of the initializer when you serialize ? or just the keys you curate in the init function?

If it's the former then I think you just want to get rid of `PipelineSettings` altogether because I am not sure to see the value of a class that just hols a dict, to them return it in to_dict

and your test that tests `to_dict` seems to imply that `PipelineSettings` does nothing

wangch079:
> Ok, is the goal is to convert it to something you want to serialize in JSON? In that case I would rename it `to_json`

No, I don't need JSON. When constructing the payload to ES API call, I still need `dict`.

> do you want to keep the data from the `pipeline` argument of the initializer when you serialize ? or just the keys you curate in the init function?

I want to keep the original `pipeline`. 

> If it's the former then I think you just want to get rid of `PipelineSettings` altogether because I am not sure to see the value of a class that just hols a dict, to them return it in to_dict

The main purpose of `PipelineSettings` is to have a central place to hold the default value. Pipeline is used by both `Connector` and `SyncJob`. If we use `dict` directly, we need to handle `pipeline.get("extract_binary_content", True)` in different places.

What do you think if we implementing `PipelineSettings` as a derived class of `dict`? similar to `Filter`:

https://github.com/elastic/connectors-python/blob/e6a1e9463f0633748c3cdf7a65021e5950221fb2/connectors/byoc.py#L307-L324

tarekziade:
ok so if you just want a `dict` with some default values, given an initial one, why not:

```
def pipeline_settings(config):
    settings = config.copy()

    for key, default in (
        ("name", "ent-search-generic-ingestion"),
        ("extract_binary_content", True),
        ("reduce_whitespace", True),
        ("run_ml_inference", True),
    ):
        if key not in config:
            config[key] = default

    return config
```

tarekziade:
or even simpler:

```
DEFAULT = {
    "name": "ent-search-generic-ingestion",
    "extract_binary_content": True,
    "reduce_whitespace": True,
    "run_ml_inference": True,
}


def pipeline_settings(config):
    settings = DEFAULT.copy()
    setting.update(config)
    return settings
```

tarekziade:
Now if you want to keep `__repr__` as it is right now, you can use `UserDict` see https://docs.python.org/3/library/collections.html#collections.UserDict

Our Filter class should use `UserDict` by the way, and not use `dict` cc @timgrein

tarekziade:
example: 

```
from collections import UserDict

DEFAULT = {
    "name": "ent-search-generic-ingestion",
    "extract_binary_content": True,
    "reduce_whitespace": True,
    "run_ml_inference": True,
}


class PipelineSettings(UserDict):
    def __repr__(self):
        return (
            f"Pipeline {self['name']} "
            f"<binary: {self['extract_binary_content']} "
            f"whitespace {self['reduce_whitespace']} "
            f"ml inference {self['run_ml_inference']}>"
        )


def pipeline_settings(config=None):
    settings = PipelineSettings(DEFAULT)
    if config is not None:
        settings.update(config)
    return settings


print(pipeline_settings({"extra": 1}))
```

running it:

```
Dev/connectors-python % bin/python pipe.py
Pipeline ent-search-generic-ingestion <binary: True whitespace True ml inference True>
```

tarekziade:
Simplified to match your current API signature:

```
from collections import UserDict

DEFAULT = {
    "name": "ent-search-generic-ingestion",
    "extract_binary_content": True,
    "reduce_whitespace": True,
    "run_ml_inference": True,
}


class PipelineSettings(UserDict):
    def __init__(self, data):
        default = DEFAULT.copy()
        default.update(data)
        super().__init__(default)

    def __repr__(self):
        return (
            f"Pipeline {self['name']} "
            f"<binary: {self['extract_binary_content']} "
            f"whitespace {self['reduce_whitespace']} "
            f"ml inference {self['run_ml_inference']}>"
        )


print(PipelineSettings({"extra": 1}))
```

wangch079:
I think I didn't explained clearly, here is what I need:

In `Connector` class:

```python
Class Connector:
    ...

    @property
    def pipeline(self):
        return PipelineSettings(self._source["pipeline"])

    ...
```

so I can use `connector.pipeline. extract_binary_content` directly

The same apples to `SyncJob` class

```python
Class SyncJob:
    ...

    @property
    def pipeline(self):
        return PipelineSettings(self._source["connector"]["pipeline"])

    ...
```

I need the `to_dict` method, because when creating the sync job:
```python
async def create_job(self, connector):
    doc = {
        "connector": {
            "pipeline": connector.pipeline.to_dict()
            ...
        },
        ...
    }

    await self.index(doc=doc)
```

Although [this](https://github.com/elastic/connectors-python/pull/440#issuecomment-1423613249) will work it may result in data difference between the `connector` and `sync_job` Elasticsearch document.

tarekziade:
> Although [this](https://github.com/elastic/connectors-python/pull/440#issuecomment-1423613249) will work it may result in data difference between the `connector` and `sync_job` Elasticsearch document.

what do you mean exactly? if you create two instances of `PipelineSettings`, yes, you have two instances.
If you need a single `PipelineSettings` instance, the problem is not the class itself, it's how you pass it around

tarekziade (changes requested):
For this patch to land, we need to describe precisely in the description of the issue the behavior we expect from `PipelineSettings` -- it's unclear. We need to formalize its contract and role with other classes -- so we can do the required changes

- does it hold read-only values initialized once?
- can we have several instances, once in `SyncJob` and one in `Connector`? if so, why
- etc.

wangch079:
>what do you mean exactly?

In my previous example, let's say the connector document in ES is like:
```json
{
  "pipeline": {
    "extract_binary_content": false
  }
}
```

When it's copied to `SyncJob`, the sync job document in ES will be like:
```json
{
  "connector": {
    "pipeline": {
      "name": "ent-search-generic-ingestion",
      "extract_binary_content": false,
      "reduce_whitespace": true,
      "run_ml_inference": true
    }
  }
}
```
That is what I meant by data difference.

> * does it hold read-only values initialized once?

Yes, it hold read-only value. No, `PipelineSettings` is just a normal class and can be instantiated mulitple times.

> * can we have several instances, once in `SyncJob` and one in `Connector`? if so, why

Yes, we can. When `SyncJob` is persisted into ES, it copies all the necessary info (e.g. `configuration`, `index_name`, `filtering`, etc..) from connector to execute a job, that's why we need to keep 2 copies of info in `Connector` and `SyncJob`. If anything is updated in `Connector`, it won't affect already scheduled jobs.

tarekziade:
> > what do you mean exactly?
> 
> In my previous example, let's say the connector document in ES is like:
> 
> ```json
> {
>   "pipeline": {
>     "extract_binary_content": false
>   }
> }
> ```
> 
> When it's copied to `SyncJob`, the sync job document in ES will be like:
> 
> ```json
> {
>   "connector": {
>     "pipeline": {
>       "name": "ent-search-generic-ingestion",
>       "extract_binary_content": false,
>       "reduce_whitespace": true,
>       "run_ml_inference": true
>     }
>   }
> }
> ```
> 
> That is what I meant by data difference.

Is this what you want as a behavior? or that's something to avoid?

Can you provide a small snippet to show how we load the connector pipeline from ES, and how we copy it over the job document? I think this is key for me to understand the desired behavior and properties.

wangch079:
>Is this what you want as a behavior? or that's something to avoid?

something I want to avoid.

Let's say below is the ES document for `connector`:
```json
{
  "pipeline": {
    "extract_binary_content": false
  }
}
```

It will be loaded to class `Connector`:
```python
connector = Connector(source=source)
print(connector.pipeline.name) # expecting ent-search-generic-ingestion
print(connector.pipeline.extract_binary_content) # expecting false
print(connector.pipeline.reduce_whitespace) # expecting true
print(connector.pipeline.run_ml_inference) # expecting true
```

A new job is created by copying connector info, you can refer to [this PR](https://github.com/elastic/connectors-python/pull/443/files#diff-b9a5d063adca223525457d0a3c06df668bd814fe46069dda6678b694c277edc9R829)

After the job is created, I expect exactly the same content in ES as in the connector document:
```json
{
  "connector": {
    "pipeline": {
      "extract_binary_content": false
    }
  }
}
```

And when I load the document to class `SyncJob`, I expect the same behavior:
```python
sync_job = SyncJob(source=source)
print(sync_job.pipeline.name) # expecting ent-search-generic-ingestion
print(sync_job.pipeline.extract_binary_content) # expecting false
print(sync_job.pipeline.reduce_whitespace) # expecting true
print(sync_job.pipeline.run_ml_inference) # expecting true
```

tarekziade:
Ok gotcha, so you are loading a pipeline, adding defaults in it, and when you copy it back to Elasticsearch, you want to get rid of the extra defaults you have added.

Why having those default values set and written back in Elasticsearch is a problem when they are not set initially?
All those fields are part of the [protocol](https://github.com/elastic/connectors-ruby/blob/main/docs/CONNECTOR_PROTOCOL.md#elasticsearch-mappings-for-elastic-connectors)

wangch079:
>Why having those default values set and written back in Elasticsearch is a problem when they are not set initially?

I don't see a huge issue here. One edge case is When a job is scheduled with some default values (e.g. extract_binary_content), then the service shuts down and upgrades, and we change the default value for `extract_binary_content` from `True` to `False`.

This is a nearly impossible case. TBH, it doesn't really bother me.

What I really want is to keep the `filtering` between `Connector` and `SyncJob` document in ES identical, and I think this is a good practice.

tarekziade:
> What I really want is to keep the filtering between Connector and SyncJob document in ES identical,

But in practice they are because every time you use in the app the settings, you add the default values on the fly, so you will never ever have a case where you need to remove the default values you've added. This is why I think [a simple dict like this works](https://github.com/elastic/connectors-python/pull/440#issuecomment-1423613249).

artem-shelkovnikov (approved):

tarekziade (approved):

PR #442, opened 2023-02-09
Use active filter in async_bulk

wangch079:
### Changes made:

Pass the active filtering (an instance of `Filter`) instead of filtering (a list of filtering including `active` and `draft` filtering, an instance of `Filtering`).

### Why the change:

Moving forward, we will use `filtering` data in `syncJob`, where only active filtering data is available.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -177,7 +177,7 @@ class Fetcher:
         queue,
         index,
         existing_ids,
-        filtering=Filter(),
+        filter=Filter(),
         sync_rules_enabled=False,
         queue_size=DEFAULT_QUEUE_SIZE,
         display_every=DEFAULT_DISPLAY_EVERY,
@@ -196,9 +196,9 @@ class Fetcher:
         self.total_docs_created = 0
         self.total_docs_deleted = 0
         self.fetch_error = None
-        self.filtering = filtering
+        self.filter = filter
         self.basic_rule_engine = (
-            BasicRuleEngine(parse(filtering.get("rules", [])))
+            BasicRuleEngine(parse(filter.get("rules", [])))
             if sync_rules_enabled
             else None
         )
@@ -414,7 +414,7 @@ class ElasticServer(ESClient):
         index,
         generator,
         pipeline,
-        filtering=Filter(),
+        filter=Filter(),
         sync_rules_enabled=False,
         options=None,
     ):
@@ -445,7 +445,7 @@ class ElasticServer(ESClient):
             stream,
             index,
             existing_ids,
-            filtering=filtering,
+            filter=filter,
             sync_rules_enabled=sync_rules_enabled,
             queue_size=queue_size,
             display_every=display_every,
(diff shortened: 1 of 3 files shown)


connectors/byoei.py
```diff
@@ -177,7 +177,7 @@ def __init__(
         queue,
         index,
         existing_ids,
-        filtering=Filtering(),
+        filtering=Filter(),
```
tarekziade, line 180: if you use `Filter` instead of `Filtering`, shouldn't we rename the `filtering` argument as well?

wangch079: I know it's a bit confusing, we actually use `filtering` for everything interchangeably. 

This is the structure of the connector filtering:
```
"filtering": [
  {
    "domain": "DEFAULT",
    "draft": {"advanced_snippet": {}, "rules": [], "validation": {}},
    "active": {"advanced_snippet": {}, "rules": [], "validation": {}}
  }
]
```

The `Filtering` instance represents the outer array, but there should only be one element, which we filter by `domain`, and the `Filter` instance represents either the `draft` or `active` filtering. 

I think we should have some proper naming. cc @timgrein `Filterings` and `Filtering`? I don't think it's a good idea to name a class plural.

timgrein: Maybe we should go for `FilterCollection`, which contains multiple `Filter` instances. The assumption that we always only have one `Filter` instance inside `filtering` will probably not be true anymore in 8.8 as we may look into using separate `Filter` instances for different tables (as an example).

++ on class names shouldn't be plural, would make it harder to distinguish between `Filters` and `Filter`

https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098 mentions some further reasons under "Class names"

wangch079:
This PR includes some changes in PR #440 , which is being blocked.

Will only merge this after #440 is merged.

connectors/byoei.py
wangch079: I kind of think should rename `Filter` to `Filtering`, as this is the first and only place I see it's named like this.

In the Project brief, its official name is `Filtering Rules`. In UI, it's Sync Rules.

WDYT if we rename `Filtering` to `FilteringSetting`, and rename `Filter` to `Filtering`.

Speaking of plural class name, we do have class `PipelineSettings`.

timgrein: My personal preference would be to have a class name, which makes it obvious that we're having multiple instances of a specific type at hand. A `FilteringSetting` could also be _one_ filtering setting for _one_ domain. From the workplace search docs I see that they're using the term _filters_. Maybe @danajuratoni can help, how we'll call a single filter/filtering/filtering setting (or whatever) in the future and base our decision on that?

But for me it's fine as long as it's consistent and obvious what's the container and what's the single instance. I don't want to block here 👍

wangch079: Agree, I don't think this is blocking. Not customer-facing. I will just rename this as `Filter`.

artem-shelkovnikov (approved):
I think it's okay, but I'm a little away from filter/filtering design, @timgrein how do you feel?

PR #445, opened 2023-02-09
[Refactoring] Move e2str to utils to avoid cyclic imports

timgrein:
`e2str` is a pretty generic method and should probably live in `utils.py`, so we don't run into possible cyclic imports, when we import it.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
We'll also eventually stop using this logic, hopefully, and make our enums strings (there seem to be no point to make them integers)

timgrein:
@artem-shelkovnikov agree, would be much cleaner. Probably worth to create a tech-debt issue to have it on our radar?

artem-shelkovnikov:
> Probably worth to create a tech-debt issue to have it on our radar?

Yes, absolutely!

PR #446, opened 2023-02-09
Add --log-level setting for ingest service instead of --debug flag

artem-shelkovnikov:
Minor improvement for the CLI:

Instead of `--debug` flag setting log level to DEBUG and having INFO as a default, we can now specify log level in 2 ways:

1. By providing it in config in `service.log_level: <LOG_LEVEL>` setting
2. By running the service with --log-level=<LOG_LEVEL>

Precedence of the operation is:
1. Option from CLI arguments
2. Option from configuration file
3. If none provided, fall back to INFO

#### Pre-Review Checklist

## Release Note

Now it's possible to choose the log level of the service by either using a "--log-level=<LOG_LEVEL>" argument when running the service over CLI or setting a `service.log_level: <LOG_LEVEL>` option in configuration file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,13 +57,13 @@ ftest: bin/pytest bin/elastic-ingest
 	connectors/tests/ftest.sh $(NAME) $(PERF8)
 
 run: install
-	bin/elastic-ingest --debug
+	bin/elastic-ingest --log-level=DEBUG
 
 docker-build:
 	docker build -t docker.elastic.co/enterprise-search/elastic-connectors:$(VERSION)-SNAPSHOT .
 
 docker-run:
-	docker run -v $(PWD):/config docker.elastic.co/enterprise-search/elastic-connectors:$(VERSION)-SNAPSHOT /app/bin/elastic-ingest -c /config/config.yml --debug
+	docker run -v $(PWD):/config docker.elastic.co/enterprise-search/elastic-connectors:$(VERSION)-SNAPSHOT /app/bin/elastic-ingest -c /config/config.yml --log-level=DEBUG
 
 docker-push:
 	docker push docker.elastic.co/enterprise-search/elastic-connectors:$(VERSION)-SNAPSHOT
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ ftest: bin/pytest bin/elastic-ingest
 	connectors/tests/ftest.sh $(NAME) $(PERF8)
 
 run: install
-	bin/elastic-ingest --log-level=DEBUG
+	bin/elastic-ingest
 
 docker-build:
 	docker build -t docker.elastic.co/enterprise-search/elastic-connectors:$(VERSION)-SNAPSHOT .
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -117,7 +117,10 @@ def run(args):
     # load config
     config = load_config(args.config_file)
     # Presedence: CLI args >> Config Setting >> INFO
-    set_logger(args.log_level or config['service']['log_level'] or logging.INFO, filebeat=args.filebeat)
+    set_logger(
+        args.log_level or config["service"]["log_level"] or logging.INFO,
+        filebeat=args.filebeat,
+    )
 
     # just display the list of connectors
     if args.action == "list":
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -44,12 +44,20 @@ def _parser():
         default=os.path.join(os.path.dirname(__file__), "..", "config.yml"),
     )
 
-    parser.add_argument(
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
         "--log-level",
         type=str,
         choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
         default=None,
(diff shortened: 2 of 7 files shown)


connectors/cli.py
```diff
@@ -45,9 +45,10 @@ def _parser():
...
 
     parser.add_argument(
-        "--debug",
-        action="store_true",
-        default=False,
+        "--log-level",
+        type=str,
+        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
+        default=None,
         help="Run the event loop in debug mode.",
```
wangch079, line 52:
```suggestion
        help="The log level of the application.",
```

connectors/cli.py
```diff
@@ -115,6 +116,11 @@ def run(args):
...
     # load config
     config = load_config(args.config_file)
+    # Precedence: CLI args >> Config Setting >> INFO
+    set_logger(
+        args.log_level or config["service"]["log_level"] or logging.INFO,
+        filebeat=args.filebeat,
+    )
 
     # just display the list of connectors
     if args.action == "list":
```
wangch079, line 126: if log level is set to `ERROR`, `bin/elastic-ingest --action list` won't work.

I think it worth changing to `print()` instead of `logger.info()`:

```
Registered connectors:
- MongoDB
- Amazon S3
- System Directory
- MySQL
- Network Drive
- Google Cloud Storage
- Azure Blob Storage
- PostgreSQL
- Oracle Database
Bye
```

This looks much better than 
```
[FMWK][22:24:42][INFO] Registered connectors:
[FMWK][22:24:42][INFO] - MongoDB
[FMWK][22:24:42][INFO] - Amazon S3
[FMWK][22:24:42][INFO] - System Directory
[FMWK][22:24:42][INFO] - MySQL
[FMWK][22:24:42][INFO] - Network Drive
[FMWK][22:24:42][INFO] - Google Cloud Storage
[FMWK][22:24:42][INFO] - Azure Blob Storage
[FMWK][22:24:42][INFO] - PostgreSQL
[FMWK][22:24:42][INFO] - Oracle Database
[FMWK][22:24:42][INFO] Bye
```

Makefile
```diff
@@ -57,13 +57,13 @@ ftest: bin/pytest bin/elastic-ingest
 	connectors/tests/ftest.sh $(NAME) $(PERF8)
 
 run: install
-	bin/elastic-ingest --debug
+	bin/elastic-ingest --log-level=DEBUG
```
wangch079, line 60: Both `--log-level=DEBUG` and `--log-level DEBUG` will work, but I think we should follow the same convention in the application to use space instead of `=`

connectors/cli.py
artem-shelkovnikov: I agree, can change in a separate PR though. @tarekziade how do you feel?

connectors/cli.py
artem-shelkovnikov: Yup, noticed it as well :) Thanks for attentive review!

Makefile
artem-shelkovnikov: Yup agreed, but I also removed --log-level from `run` - it will use the one defined in config file, IMO it's better.

artem-shelkovnikov: We can have `make run-debug` as a shorthand for this command instead, idk if we need tho.

Makefile
```diff
@@ -57,13 +57,13 @@ ftest: bin/pytest bin/elastic-ingest
 	connectors/tests/ftest.sh $(NAME) $(PERF8)
 
 run: install
-	bin/elastic-ingest --debug
+	bin/elastic-ingest
```
tarekziade, line 60: do we want to put back `--debug`

connectors/cli.py
```diff
@@ -44,11 +44,20 @@ def _parser():
         default=os.path.join(os.path.dirname(__file__), "..", "config.yml"),
     )
 
-    parser.add_argument(
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
+        "--log-level",
+        type=str,
+        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
+        default=None,
```
tarekziade, line 52: shouldn't we have a default that one of the `choices` items?

connectors/cli.py
tarekziade: yeah `list` is very specific, it should not pick the log level, good catch @wangch079

tarekziade (approved):
just a nit

connectors/cli.py
wangch079: I think Artem want to use the log level in config if it's not specified in args.

artem-shelkovnikov: Yeah if we default to something here, the setting from config file is never picked up

Makefile
artem-shelkovnikov: I'm not sure - I think it's okay to use the setting from configuration file now - if you want debug locally, you put debug locally, otherwise it's just INFO. 

No hard opinion on that though.

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/456)|

### Questions ?

Issue #399, opened 2023-01-31
System dep for the Oracle lib


tarekziade:
## Bug Description

See https://buildkite.com/elastic/connectors-python-nightly/builds/276

solution:
https://gist.github.com/kimus/10012910

todo:

- [ ] fix the e2e test in `connectors-python`
- [ ] backport the system fix to `ent-search` cloud image build process

tarekziade:
maybe we should use https://oracle.github.io/python-oracledb/ ?

tarekziade:
also https://python-oracledb.readthedocs.io/en/latest/user_guide/appendix_c.html#upgrading83

parth-crest:
Hey @tarekziade, we are planning to upgrade SQLAlchemy to version 2.0.0 which supports python-oracledb instead of cx_Oracle, where the need for installing oracle client would be eliminated.
We have created a [ticket](https://github.com/elastic/connectors-python/issues/400) on Github to track this change separately. Hence this issue will be taken care once that change is implemented.

parth-crest:
Hey @tarekziade, After upgrading SQLAlchemy v2.0 This dependency is removed. We can close this bug once you verify that the build is passing successfully. 
We have verified and attached the screenshot.
![image (16)](https://user-images.githubusercontent.com/109953470/220064209-c29f8e83-99d3-42f0-b777-eb6c97a89a06.png)

danajuratoni:
@tarekziade can this issue be closed?

khusbu-crest:
@tarekziade Looking forward to your confirmation in order to close this issue.

khusbu-crest:
@tarekziade Could you please look into this, we are waiting for your confirmation in order to close this issue?

PR #447, opened 2023-02-09
Upgrade SQLAlchemy v2.0 to use python-oracledb and improve performance

moxarth-elastic:
## Closes https://github.com/elastic/connectors-py/issues/400, https://github.com/elastic/connectors-python/issues/399

This PR includes:

- SQLAlchemy v2.0 upgraded.
- Upgraded cx_Oracle to python-oracledb to remove users's dependency to install oracle instant client
- Improved performance for PostgreSQL & Oracle.

#### Pre-Review Checklist
- [x] Tested the changes locally by running `make test` and `make ftest NAME=oracle`

- [x] Security-related changes (encryption, TLS, SSRF, etc). User needs to use Thick mode to use SSL.
- [x] New external service dependencies added. Considering we have removed cx_oracle and installed python-oracledb which comes under [Apache License, Version 2.0](https://github.com/oracle/python-oracledb/blob/main/LICENSE.txt).


connectors/sources/tests/fixtures/oracle/fixture.py
```diff
@@ -48,33 +51,42 @@ def inject_lines(table, cursor, start, lines):
...
+            cursor = connection.cursor()
+            for table in range(NUM_TABLES):
+                print(f"Adding data in {table}...")
+                sql_query = f"CREATE TABLE customers_{table} (id int, name VARCHAR(255), age int, description long, PRIMARY KEY (id))"
+                cursor.execute(sql_query)
+                for i in range(10):
+                    inject_lines(table, cursor, i * 1000, 1000)
+            connection.commit()
+            break
+        except Exception as exception:
```
timgrein, line 77: There's a util method for retrying. See `retryable` decorator in https://github.com/elastic/connectors-python/blob/main/connectors/utils.py.

connectors/sources/tests/test_oracle.py
```diff
@@ -36,3 +36,14 @@ async def test_ping_negative(patch_logger):
...
+
+@pytest.mark.asyncio
+@patch("connectors.sources.oracle.create_engine")
+async def test_create_engine(mock_fun):
+    """Test create_engine method of OracleDataSource class"""
+    # Setup
+    source = create_source(OracleDataSource)
+    source.oracle_home = "/home/devuser"
+    mock_fun.return_value = MagicMock()
+    source._create_engine()
```
timgrein, line 49: What does this test check? That no exception is thrown?

I can see that the `_create_engine` method has two cases:
- oracle_home is already present -> change the environment variable and create the engine in thick mode
- oracle_home is not present -> just create it from the connection string

I would advocate for having explicit assertions/tests, that the engine is created correctly for both cases. You could check that the passed args for `create_engine` are the expected ones.

connectors/sources/tests/fixtures/oracle/fixture.py
```diff
@@ -48,33 +51,42 @@ def inject_lines(table, cursor, start, lines):
...
+            connection = oracledb.connect(
+                user="system", password="Password_123", dsn=DSN, encoding="UTF-8"
+            )
+            cursor = connection.cursor()
+            cursor.execute("CREATE USER admin IDENTIFIED by Password_123")
+            cursor.execute("GRANT CONNECT, RESOURCE, DBA TO admin")
+            connection.commit()
+
+            connection = oracledb.connect(
+                user="admin", password="Password_123", dsn=DSN, encoding="UTF-8"
```
timgrein, line 66: `user`, `password` and `encoding` seem to be constant and are used at multiple places. You could create constants and reuse them.

connectors/sources/tests/test_generic_database.py
```diff
@@ -338,7 +342,9 @@ async def test_fetch_documents_negative(patch_logger):
         "execute_query",
         side_effect=InternalClientError("Something went wrong"),
     ):
-        source.engine = create_engine("oracle://admin:changme@127.0.0.1:1521/testdb")
+        source.engine = create_engine(
+            "oracle+oracledb://admin:changme@127.0.0.1:1521/testdb"
```
timgrein, line 346: Also seems to be a constant used in multiple places. You could create a `CONNECTION_STRING` constant and reuse it.

tarekziade:
> Improved performance for PostgreSQL & Oracle

Is this measurable?  Could you run `make ftest` on them and grab perf8 report before/after ?

connectors/sources/generic_database.py
```diff
@@ -103,6 +105,16 @@ def get_default_configuration(cls):
                 "label": "Oracle connection protocol",
                 "type": "str",
             },
+            "oracle_home": {
```
tarekziade, line 108: why are the oracle options in the generic database class? shouldn't they be in the oracle class?

connectors/sources/generic_database.py
```diff
@@ -44,6 +44,8 @@ def __init__(self, configuration):
         self.queries = None
         self.ssl_disabled = self.configuration["ssl_disabled"]
         self.ssl_ca = self.configuration["ssl_ca"]
+        self.oracle_home = self.configuration["oracle_home"]
```
tarekziade, line 47: should `self.oracle_home` and `self.wallet_config` be in the oracle class?

connectors/sources/tests/fixtures/oracle/requirements.txt
```diff
@@ -1 +1 @@
-cx-Oracle==7.3.0
+oracledb
```
timgrein, line 1:
```suggestion
oracledb==1.2.2
```

E2E-test should use the same version as the framework.

moxarth-elastic:
Sure we are working on the suggested changes. we will let you know once we update the PR.

connectors/sources/tests/fixtures/oracle/fixture.py
moxarth-elastic: Done, replace with `retryable`.

connectors/sources/tests/test_oracle.py
moxarth-elastic: Added two separate test cases for `_create_engine` and added assert with `assert_called_with`.

connectors/sources/generic_database.py
moxarth-elastic: Sure, moved in `oracle.py` file. Also applied a similar change in PostgreSQL

moxarth-elastic:
Hey @tarekziade, @timgrein Updated PR with the changes. Please take a look.

connectors/sources/oracle.py
```diff
@@ -34,16 +39,58 @@ def __init__(self, configuration):
...
         self.dialect = "Oracle"
 
+    @classmethod
+    def get_default_configuration(cls):
+        """Get the default configuration for database-server configured by user
+
+        Returns:
+            dictionary: Default configuration
+        """
+        oracle_configuration = copy.deepcopy(super().get_default_configuration())
```
tarekziade, line 59: nit: since we don't change the items, you can probably just use `super().get_default_configuration().copy()` (shallow)

connectors/sources/tests/fixtures/oracle/fixture.py
```diff
@@ -5,16 +5,23 @@
...
-import cx_Oracle
+import oracledb
+
+from connectors.utils import RetryStrategy, retryable
 
 DATA_SIZE = os.environ.get("DATA_SIZE", "small").lower()
 _SIZES = {"small": 5, "medium": 10, "large": 30}
 NUM_TABLES = _SIZES[DATA_SIZE]
 
+# If the Oracle Service takes too much time to respond, modify the following retry constants accordingly.
```
tarekziade, line 21: should these be config options?

tarekziade (approved):
just a few nits on my side

connectors/sources/tests/fixtures/oracle/fixture.py
moxarth-elastic: As mentioned E2E tests are performed by developers So we kept these attributes constant and users can update whenever they want as it depends on the system configuration.

connectors/sources/oracle.py
moxarth-elastic: done

moxarth-elastic:
Hey @tarekziade, Updated PR with the changes. Please take a look.

tarekziade:
buildkite test this

timgrein (approved):
LGTM

PR #448, opened 2023-02-09
[Filtering] Add method for checking a filter validation state

timgrein:
Add a convencience method for checking, that a `Filter` instance is in a specific validation state. This will be used by @wangch079 in his refactoring.
 
#### Pre-Review Checklist


connectors/filtering/validation.py
```diff
@@ -103,11 +103,17 @@ def __str__(self):
...
+class InvalidFilteringValidationStateError(ValueError):
+    pass
+
+
 class FilteringValidationState(Enum):
     VALID = 1
     INVALID = 2
     EDITED = 3
 
+    STATES = [VALID, INVALID, EDITED]
```
tarekziade, line 115: nit, make it immutable:

```
STATES = VALID, INVALID, EDITED
```

connectors/filtering/validation.py
```diff
@@ -118,6 +124,20 @@ def to_s(cls, value):
             case FilteringValidationState.EDITED:
                 return "edited"
 
+    @classmethod
+    def from_string(cls, string):
+        match string.casefold():
+            case "valid":
+                return FilteringValidationState.VALID
```
tarekziade, line 131: would it help to move

```
class FilteringValidationState(Enum):
    VALID = 1
    INVALID = 2
    EDITED = 3
```

to 

```
class FilteringValidationState(Enum):
    VALID = "valid"
    INVALID = "invalid"
    EDITED = "edited"
```

connectors/filtering/validation.py
tarekziade: you can also use:

```
STATES = tupe(item.value for item in FilteringValidationState)
```

(I am not entirely sure why you need `STATES` though)

connectors/filtering/validation.py
```diff
@@ -118,6 +124,20 @@ def to_s(cls, value):
...
+            case "invalid":
+                return FilteringValidationState.INVALID
+            case "edited":
+                return FilteringValidationState.EDITED
+            case _:
+                raise InvalidFilteringValidationStateError(
+                    f"'{string}' is an unknown value for the enum FilteringValidationState. Allowed states: {FilteringValidationState.STATES}.)"
```
tarekziade, line 138: I would remove `STATE` and just call `[e.value for e in FilteringValidationState]` here

wangch079: Let's use the `str2e` function in #449 . The function assumes `None` or a valid enum value, and ignore invalid values, which I think it's good enough for now.

Let's keep the processing of Enum in a single place and we can enhance later.

connectors/tests/test_byoc.py
```diff
@@ -1056,6 +1071,21 @@ def test_advanced_rules_present(filtering, should_advanced_rules_be_present):
...
 
+@pytest.mark.parametrize(
+    "filtering, validation_state, has_expected_validation_state",
+    [
+        (FILTER_VALIDATION_STATE_VALID, FilteringValidationState.EDITED, False),
+        (FILTER_VALIDATION_STATE_INVALID, FilteringValidationState.EDITED, False),
+        (FILTER_VALIDATION_STATE_EDITED, FilteringValidationState.EDITED, True),
+    ],
+)
+def test_is_in_state_edited(filtering, validation_state, has_expected_validation_state):
```
wangch079, line 1082: probably should rename it to `test_has_validation_state`

connectors/filtering/validation.py
timgrein: Why did we use numbers in the first place for other enums? Your suggestion seems to be way more convenient. TIL

connectors/filtering/validation.py
timgrein: Removed it by taking your suggestion further down into consideration 👍

connectors/tests/test_byoc.py
timgrein: Yes, good catch! Forgot to rename it

connectors/filtering/validation.py
tarekziade: > Why did we use numbers in the first place for other enums? 

because Enum are often used as a collection of integers, but that was a mistake to do so.

> Your suggestion seems to be way more convenient. 

Yeah and you can do stuff like this :

```
state = FilteringValidationState('valid')
```

timgrein: Cool, adjusted it. Thanks for creating the tech-debt issue

tarekziade (approved):

PR #449, opened 2023-02-09
Add str2e function

wangch079:
This PR adds a `str2e` function in module `connectors.utils`, to convert a string to an Enum value

#### Pre-Review Checklist

### DIFF
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -401,3 +401,14 @@ def retryable(retries=3, interval=1.0, strategy=RetryStrategy.LINEAR_BACKOFF):
 
 def e2str(entry):
     return entry.name.lower()
+
+
+def str2e(name, enum_class):
+    if name is None:
+        return None
+
+    name = str(name).upper()
+    if name not in enum_class.__members__:
+        return None
+
+    return enum_class[name]
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -16,7 +16,7 @@ from connectors.es import DEFAULT_LANGUAGE, ESIndex, Mappings
 from connectors.filtering.validation import ValidationTarget, validate_filtering
 from connectors.logger import logger
 from connectors.source import DataSourceConfiguration, get_source_klass
-from connectors.utils import e2str, iso_utc, next_run
+from connectors.utils import e2str, iso_utc, next_run, str2e
 
 CONNECTORS_INDEX = ".elastic-connectors"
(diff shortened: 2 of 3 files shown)


tarekziade:
related: @vidok proposed to move Enums to strings, 

e.g.

```
class Status(Enum):
    CREATED = 1
    NEEDS_CONFIGURATION = 2
    CONFIGURED = 3
    CONNECTED = 4
    ERROR = 5
```

becomes

```
class Status(Enum):
    CREATED = 'created'
    NEEDS_CONFIGURATION = 'needs_configuration'
    CONFIGURED = 'configured'
    CONNECTED = 'connected'
    ERROR = 'error'
```

ISTM that this would make our lives easier and avoid extra conversions

connectors/utils.py
```diff
@@ -401,3 +401,14 @@ async def func_to_execute(*args, **kwargs):
 
 def e2str(entry):
     return entry.name.lower()
+
+
+def str2e(name, enum_class):
+    if name is None:
```
tarekziade, line 407: it seems that this function will return `None` when the name is not know. Shouldn't we raise an error instead? I can think of problems where None will be used silently and can't think of scenarios where we ask for a name that is not supposed to exist in the Enum

wangch079:
>related: @vidok proposed to move Enums to strings,

I think a `str2e` function is still needed?

We can use `Status.CREATED.value` instead of `e2str` but I don't think this is much difference.

tarekziade:
> 



> > related: @vidok proposed to move Enums to strings,
> 
> I think a `str2e` function is still needed?

no because you can then do:

```
Status('created')
```

> 
> We can use `Status.CREATED.value` instead of `e2str` but I don't think this is much difference.

connectors/utils.py
tarekziade: Gotcha. I guess if we use strings in the enums we can have an explicit set:

```
enum_value = None if value is None else JobStatus(value)
```

tarekziade: or an unset value:

```
>>> class MyEnum(Enum):
...     one = 'one'
...     two = 'two'
...     unset = None
...
>>> MyEnum(None)
<MyEnum.unset: None>
>>> MyEnum('one')
<MyEnum.one: 'one'>
```

tarekziade (approved):
please consider all my feedback as possible improvements, but the new function is good enough if that helps you to unblock other PRS

PR #453, opened 2023-02-10
[Fix] Fix flaky test "test_sync_job"

timgrein:
Mock the elasticsearch client directly and assert the passed kwargs.
Also remove explicit sleep for some speed up.
Use a local variable instead of a "constant", which can get mutated dependent on the test case order.

#### Pre-Review Checklist

### DIFF
--- a/connectors/tests/test_byoc.py
+++ b/connectors/tests/test_byoc.py
@@ -253,7 +253,7 @@ async def test_sync_job():
     await job.done(12, 34)
 
     assert job.status == JobStatus.COMPLETED
-    assert job.duration >= -1
+    assert job.duration > -1
 
     updated_job_def = update_mock.call_args.kwargs["doc"]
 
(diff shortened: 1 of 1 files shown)


connectors/tests/test_byoc.py
```diff
@@ -213,70 +212,54 @@ def test_utc():
...
-
-    # verify what was sent
-    assert len(sent_docs) == 2
-    doc, update = sent_docs
-    assert doc["status"] == "in_progress"
-    assert doc["connector"]["filtering"] == expected_filtering
-    assert update["doc"]["status"] == "completed"
-    assert update["doc"]["indexed_document_count"] == 12
-    assert update["doc"]["deleted_document_count"] == 34
+    assert job.duration > -1
```
artem-shelkovnikov, line 256: Hmmm, but can it actually be in the range (-1; 0]. If so, how come?

timgrein: Oh wow of course you're right, was so focused on tracking down, why we use microseconds and if it has implications somewhere else. Adjusted it!

artem-shelkovnikov (approved):

PR #458, opened 2023-02-10
bump elasticsearch client

tarekziade:
bump to `8.6.1`


wangch079 (commented):
I think we only bump for the major or minor version? Or we should always try to use the latest version?

tarekziade:
> we should always try to use the latest version?

We should always try to use the latest `PATCH` release because it's bugfixes, security fixes, and sometimes minor backward compatible features -- but we don't care for the latter 

Our `8.MINOR` branch should follow the latest `MINOR` of `elasticsearch` client

wangch079 (approved):

PR #460, opened 2023-02-12
Refactor SyncJob class

wangch079:
This PR refactors `SyncJob` class. The changes are:

1. Add a new class `ESDocument`, to be a parent class of `SyncJob` and `Connector` class.
2. Make `SyncJob` data representation of a document in `.elastic-connectors-sync-jobs`, and include self-mutating methods.
3. Move `transform_filtering` method from `SyncJob` to `Filter`.
4. Add `validate_filtering` method to `SyncJob`.
5. Make the system work with the refactored `SyncJob`.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -16,8 +16,8 @@ from enum import Enum
 from connectors.es import ESIndex, Mappings
 from connectors.filtering.validation import (
     FilteringValidationState,
+    InvalidFilteringError,
     ValidationTarget,
-    validate_filtering,
 )
 from connectors.logger import logger
 from connectors.source import DataSourceConfiguration, get_source_klass
@@ -259,6 +259,18 @@ class SyncJob(ESDocument):
     @property
     def total_document_count(self):
         return self.get("total_document_count", default=0)
+
+    async def validate_filtering(self, validator):
+        validation_result = await validator.validate_filtering(self.filtering)
+
+        if validation_result.state != FilteringValidationState.VALID:
+            raise InvalidFilteringError(
+                f"Filtering in state {validation_result.state}. Expected: {FilteringValidationState.VALID}."
+            )
+        if len(validation_result.errors):
+            raise InvalidFilteringError(
+                f"Filtering validation errors present: {validation_result.errors}."
+            )
 
     async def claim(self):
         doc = {
@@ -794,7 +806,7 @@ class Connector:
             sync_rules_enabled = self.features.sync_rules_enabled()
 
             if sync_rules_enabled:
-                await validate_filtering(self, self.index, ValidationTarget.ACTIVE)
+                await job.validate_filtering(validator=self.data_provider)
 
             result = await elastic_server.async_bulk(
                 self.index_name,
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -265,11 +265,7 @@ class SyncJob(ESDocument):
 
         if validation_result.state != FilteringValidationState.VALID:
             raise InvalidFilteringError(
-                f"Filtering in state {validation_result.state}. Expected: {FilteringValidationState.VALID}."
-            )
-        if len(validation_result.errors):
-            raise InvalidFilteringError(
-                f"Filtering validation errors present: {validation_result.errors}."
+                f"Filtering in state {validation_result.state}, errors: {validation_result.errors}."
             )
 
     async def claim(self):
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -255,29 +255,33 @@ class SyncJob(ESDocument):
         }
         await self.index.update(doc_id=self.id, doc=doc)
 
-    async def done(self, ingestion_stats={}, connector_metadata={}):
+    async def done(self, ingestion_stats=None, connector_metadata=None):
         await self._terminate(
             JobStatus.COMPLETED, None, ingestion_stats, connector_metadata
         )
 
-    async def fail(self, message, ingestion_stats={}, connector_metadata={}):
+    async def fail(self, message, ingestion_stats=None, connector_metadata=None):
         await self._terminate(
             JobStatus.ERROR, str(message), ingestion_stats, connector_metadata
         )
 
-    async def cancel(self, ingestion_stats={}, connector_metadata={}):
+    async def cancel(self, ingestion_stats=None, connector_metadata=None):
         await self._terminate(
             JobStatus.CANCELED, None, ingestion_stats, connector_metadata
         )
 
-    async def suspend(self, ingestion_stats={}, connector_metadata={}):
+    async def suspend(self, ingestion_stats=None, connector_metadata=None):
         await self._terminate(
             JobStatus.SUSPENDED, None, ingestion_stats, connector_metadata
         )
 
     async def _terminate(
-        self, status, error=None, ingestion_stats={}, connector_metadata={}
+        self, status, error=None, ingestion_stats=None, connector_metadata=None
     ):
+        if ingestion_stats is None:
+            ingestion_stats = {}
+        if connector_metadata is None:
+            connector_metadata = {}
         doc = {
             "last_seen": iso_utc(),
             "status": status.value,
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -7,10 +7,10 @@
 Implementation of BYOC protocol.
 """
 import asyncio
+import socket
 import time
 from collections import UserDict
 from copy import deepcopy
-from datetime import datetime, timezone
 from enum import Enum
 
 from connectors.es import ESIndex, Mappings
@@ -181,81 +181,130 @@ class ConnectorIndex(ESIndex):
             yield connector
 
 
-class SyncJob:
-    def __init__(self, elastic_index, connector_id, doc_source=None):
-        self.connector_id = connector_id
-        self.elastic_index = elastic_index
-        self.created_at = datetime.now(timezone.utc)
-        self.completed_at = None
-        self.filtering = Filter()
-        self.client = elastic_index.client
+class ESDocument:
+    def __init__(self, elastic_index, doc_source=None):
+        self.index = elastic_index
         if doc_source is None:
             doc_source = dict()
+        self.id = doc_source.get("_id")
+        self._source = doc_source.get("_source", {})
 
-        self.doc_source = doc_source
-        self.job_id = self.doc_source.get("_id")
-        self.status = JobStatus(self.doc_source.get("_source", {}).get("status"))
+    def get(self, *keys, default=None):
+        value = self._source
+        for key in keys:
+            if not isinstance(value, dict):
+                return default
+            value = value.get(key)
+        if value is None:
+            return default
+        return value
+
+    async def reload(self):
+        return await self.index.fetch_by_id(self.id)
+
+
+class SyncJob(ESDocument):
+    @property
+    def status(self):
+        return JobStatus(self.get("status"))
 
     @property
-    def duration(self):
-        if self.completed_at is None:
-            return -1
-        msec = (self.completed_at - self.created_at).microseconds
-        return round(msec / 9, 2)
+    def error(self):
+        return self.get("error")
+
+    @property
+    def connector_id(self):
+        return self.get("connector", "id")
 
     @property
     def index_name(self):
-        return self.doc_source["_source"]["connector"]["index_name"]
+        return self.get("connector", "index_name")
 
-    async def start(self, trigger_method=JobTriggerMethod.SCHEDULED, filtering=None):
-        if filtering is None:
-            filtering = Filter()
+    @property
+    def language(self):
+        return self.get("connector", "language")
 
-        self.status = JobStatus.IN_PROGRESS
-        self.filtering = filtering
+    @property
+    def service_type(self):
+        return self.get("connector", "service_type")
 
-        job_def = {
-            "connector": {
-                "id": self.connector_id,
-                "filtering": SyncJob.transform_filtering(filtering),
-            },
-            "trigger_method": trigger_method.value,
-            "status": self.status.value,
-            "error": None,
-            "deleted_document_count": 0,
-            "indexed_document_count": 0,
-            "created_at": iso_utc(self.created_at),
-            "completed_at": None,
+    @property
+    def configuration(self):
+        return DataSourceConfiguration(self.get("connector", "configuration"))
+
+    @property
+    def filtering(self):
+        return Filter(self.get("connector", "filtering", default={}))
+
+    @property
+    def pipeline(self):
+        return Pipeline(self.get("connector", "pipeline"))
+
+    @property
+    def terminated(self):
+        return self.status in (JobStatus.ERROR, JobStatus.COMPLETED, JobStatus.CANCELED)
+
+    @property
+    def indexed_document_count(self):
+        return self.get("indexed_document_count", default=0)
+
+    @property
+    def indexed_document_volume(self):
+        return self.get("indexed_document_volume", default=0)
+
+    @property
+    def deleted_document_count(self):
+        return self.get("deleted_document_count", default=0)
+
+    @property
+    def total_document_count(self):
+        return self.get("total_document_count", default=0)
+
+    async def claim(self):
+        doc = {
+            "status": JobStatus.IN_PROGRESS.value,
+            "started_at": iso_utc(),
+            "last_seen": iso_utc(),
+            "worker_hostname": socket.gethostname(),
         }
-        resp = await self.client.index(index=JOBS_INDEX, document=job_def)
-        self.job_id = resp["_id"]
-        return self.job_id
+        await self.index.update(doc_id=self.id, doc=doc)
 
-    async def done(self, indexed_count=0, deleted_count=0, exception=None):
-        self.completed_at = datetime.now(timezone.utc)
+    async def done(self, ingestion_stats={}, connector_metadata={}):
+        await self._terminate(
+            JobStatus.COMPLETED, None, ingestion_stats, connector_metadata
+        )
 
-        job_def = {
-            "deleted_document_count": deleted_count,
-            "indexed_document_count": indexed_count,
-            "completed_at": iso_utc(self.completed_at),
+    async def fail(self, message, ingestion_stats={}, connector_metadata={}):
+        await self._terminate(
+            JobStatus.ERROR, str(message), ingestion_stats, connector_metadata
+        )
+
+    async def cancel(self, ingestion_stats={}, connector_metadata={}):
+        await self._terminate(
+            JobStatus.CANCELED, None, ingestion_stats, connector_metadata
+        )
+
+    async def suspend(self, ingestion_stats={}, connector_metadata={}):
+        await self._terminate(
+            JobStatus.SUSPENDED, None, ingestion_stats, connector_metadata
+        )
+
+    async def _terminate(
+        self, status, error=None, ingestion_stats={}, connector_metadata={}
+    ):
+        doc = {
+            "last_seen": iso_utc(),
+            "status": status.value,
+            "error": error,
         }
-
-        if exception is None:
-            self.status = JobStatus.COMPLETED
-            job_def["error"] = None
-        else:
-            self.status = JobStatus.ERROR
-            job_def["error"] = str(exception)
-
-        job_def["status"] = self.status.value
-
-        return await self.client.update(index=JOBS_INDEX, id=self.job_id, doc=job_def)
-
-    async def suspend(self):
-        self.status = JobStatus.SUSPENDED
-        job_def = {"status": self.status.value}
-
-        await self.client.update(index=JOBS_INDEX, id=self.job_id, doc=job_def)
+        if status in (JobStatus.ERROR, JobStatus.COMPLETED, JobStatus.CANCELED):
+            doc["completed_at"] = iso_utc()
+        if status == JobStatus.CANCELED:
+            doc["canceled_at"] = iso_utc()
+        doc.update(ingestion_stats)
+        if len(connector_metadata) > 0:
+            doc["metadata"] = connector_metadata
+        await self.index.update(doc_id=self.id, doc=doc)
 
     @classmethod
     def transform_filtering(cls, filtering):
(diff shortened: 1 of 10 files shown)


connectors/byoc.py
```diff
@@ -178,96 +178,142 @@ async def all_connectors(self):
...
-            if (filtering is None or len(filtering) == 0)
-            else deepcopy(filtering)
+    async def validate_filtering(self, validator):
+        validation_result = await validator.validate_filtering(self.filtering)
+
+        if validation_result.state != FilteringValidationState.VALID:
+            raise InvalidFilteringError(
+                f"Filtering in state {validation_result.state}. Expected: {FilteringValidationState.VALID}."
+            )
+        if len(validation_result.errors):
```
wangch079, line 267: Hi @timgrein , is there any scenario where the state is `VALID` but `errors` is not empty?

connectors/byoc.py
```diff
@@ -321,6 +355,19 @@ def has_advanced_rules(self):
...
+        """
+        Transform the filtering in .elastic-connectors to filtering ready-to-use in .elastic-connectors-sync-jobs
+        """
+        # deepcopy to not change the reference resulting in changing .elastic-connectors filtering
+        filtering = (
+            {"advanced_snippet": {}, "rules": []} if len(self) == 0 else deepcopy(self)
+        )
+
+        # extract value for sync job
+        filtering["advanced_snippet"] = self.advanced_rules
```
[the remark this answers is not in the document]
timgrein: If you check the protocol you'll see that `advanced_snippet` nests everything under `value`, when it stores it alongside rules in the `.elastic-connectors`. As a sync job does not nest the actual content under `value` this is expected.

So the original filtering will never be: 
```
{
  "advanced_snippet": {
     "query": {}
     }, 
   "rules": []
}
```
but would be 
```
{
  "advanced_snippet": {
     "value": {
       "query": {}
      }, 
     "rules": []
}
```.

So yes this is expected. Did you test it locally?

File: connectors/byoc.py (hunk 1 above)
Comment (timgrein) [reply connectors/byoc.py:267]: No scenario from our side, but I want to be as robust as possible against connector implementations outside of our control. We could use an invariant (similar to "Design by Contract") and add an `assert` here, which simply signals a connector developer, that the provided filtering implementation is semantically wrong (or we are not as strict and log a warning or an error).

File: connectors/byoc.py [hunk 3]
```diff
@@ -178,96 +178,142 @@ async def all_connectors(self):
...
-        self.job_id = resp["_id"]
-        return self.job_id
+    @property
+    def configuration(self):
+        return DataSourceConfiguration(self.get("connector", "configuration"))
 
-    async def done(self, indexed_count=0, deleted_count=0, exception=None):
-        self.completed_at = datetime.now(timezone.utc)
+    @property
+    def filtering(self):
```
timgrein, line 233: Shouldn't `filtering` be of type array in a sync job? (see: [Protocol](https://github.com/elastic/connectors-ruby/blob/main/docs/CONNECTOR_PROTOCOL.md)) Btw I think we need to update these docs and maybe add them also to connectors-python.

connectors/byoc.py
```diff
@@ -178,96 +178,142 @@ async def all_connectors(self):
...
+
+    async def done(self, ingestion_stats={}, connector_metadata={}):
+        await self._terminate(
+            JobStatus.COMPLETED, None, ingestion_stats, connector_metadata
         )
 
-        # extract value for sync job
-        filtering["advanced_snippet"] = filtering.get("advanced_snippet", {}).get(
-            "value", {}
+    async def fail(self, message, ingestion_stats={}, connector_metadata={}):
```
timgrein, line 286: Should we call this `error` as the names of `cancel` and `suspend` also adhere to the corresponding `JobStatus`?

connectors/byoc.py
```diff
@@ -178,96 +178,142 @@ async def all_connectors(self):
...
+    async def claim(self):
+        doc = {
+            "status": e2str(JobStatus.IN_PROGRESS),
+            "started_at": iso_utc(),
+            "last_seen": iso_utc(),
+            "worker_hostname": socket.gethostname(),
+        }
+        await self.index.update(doc_id=self.id, doc=doc)
+
+    async def done(self, ingestion_stats={}, connector_metadata={}):
```
timgrein, line 281: Should we call this `complete` as the names of `cancel` and `suspend` also adhere to the corresponding `JobStatus`?

connectors/byoc.py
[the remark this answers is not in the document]
wangch079: Yes, I did test it locally. I actually also added a test case in this PR:

https://github.com/elastic/connectors-python/blob/d7053db73522da993bd4d32e074af4012030c13f/connectors/tests/test_byoc.py#L912-L915

The change I have made in this line actually changed the behavior.

I the above example, it returns the same `filtering`, while it returns `{"advanced_snippet": {}, "rules": []}` before. That's why I want to confirm.

connectors/byoc.py
wangch079: I kind of think this is part of our protocol, although it's not explicitly stated, it's just a convention.

In another PR, the draft rules will overwrite active rules, if the state is `VALID` after validation. I don't think we need to make it complicated.

connectors/byoc.py
wangch079: `error` is used by another method:

https://github.com/elastic/connectors-python/blob/d7053db73522da993bd4d32e074af4012030c13f/connectors/byoc.py#L208-L210

It's not required to use the same method name as the job status.

connectors/byoc.py
timgrein: Yes, that would be a problem. Also the new sync rule object model will probably use multiple filters for different tables. Therefore it definitely should be an array.

connectors/byoc.py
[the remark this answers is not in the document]
timgrein: Just double-checked as we're now passing an instance of `Filter` to the sync job, which already does the `value` extraction it should be fine to remove the explicit extraction within `SyncJob`.

connectors/byoc.py
timgrein: Just thought it would be a bit cleaner/easier to read. But totally fine to keep it like that 👍

connectors/byoc.py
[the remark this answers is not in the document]
wangch079: I'm not sure why the transformation is necessary:

In `connector` doc:
```json
{
  "advanced_snippet": {
     "value": {
       "query": {}
      }
   }, 
   "rules": []
}
```

which will be transformed to
```json
{
  "advanced_snippet": {
    "query": {}
   }, 
   "rules": []
}
```

Can we make it consistent before GA? cc @tarekziade

connectors/byoc.py
```diff
@@ -178,96 +178,142 @@ async def all_connectors(self):
...
+        for key in keys:
+            if not isinstance(value, dict):
+                return default
+            value = value.get(key)
+        if value is None:
+            return default
+        return value
 
+    async def reload(self):
+        return await self.index.fetch_by_id(self.id)
```
tarekziade, line 200: what happens if `self.id` is `None` (it's possible with the current constructor)

scenario:

```python
doc = ESDocument('index')
await doc.reload()  
```

we might want to add a safeguard

connectors/byoc.py
```diff
@@ -178,96 +178,138 @@ async def all_connectors(self):
...
+        return value
+
+    async def reload(self):
+        return await self.index.fetch_by_id(self.id)
+
+
+class SyncJob(ESDocument):
+    @property
+    def status(self):
+        return str2e(self.get("status"), JobStatus)
```
tarekziade, line 206: The default value needs to be an instance of `JobStatus(1)` or `JobStatus.PENDING` etc..

```python
>>> from connectors.byoc import JobStatus
>>> JobStatus
<enum 'JobStatus'>
>>> JobStatus(1)
<JobStatus.PENDING: 1>
```

connectors/byoc.py
```diff
@@ -178,96 +178,138 @@ async def all_connectors(self):
...
 
-class SyncJob:
-    def __init__(self, elastic_index, connector_id, doc_source=None):
-        self.connector_id = connector_id
-        self.elastic_index = elastic_index
-        self.created_at = datetime.now(timezone.utc)
-        self.completed_at = None
-        self.filtering = Filter()
-        self.client = elastic_index.client
+class ESDocument:
```
tarekziade, line 181: Should we move that class to `connectors/es/document.py` ?

connectors/byoc.py
tarekziade: oh sorry, I just realized how the  `str2e` works by grabbing the `Enum` class.

Forget about this comment. On a side note, this is very hard to understand and not semantically correct. We should make this simpler later.

I think we'll want this to work eventually:

```
@property
def status(self):
    return JobStatus(self.get("status"))
```

tarekziade:
related https://github.com/elastic/connectors-python/pull/464

connectors/byoc.py
wangch079: What do you think we add to `__init__` of `ESDocument`?

```python
class InvalidDocumentSourceError(Exception):
    pass


class ESDocument:
    def __init__(self, elastic_index, doc_source=None):
        self.index = elastic_index
        if not isinstance(doc_source, dict):
            raise InvalidDocumentSourceError(f"Invalid doc_source type, expected: {dict.__name__}, got: {doc_source.__class__.__name__}")
        self.id = doc_source.get("_id")
        if not isinstance(self.id, str):
            raise InvalidDocumentSourceError(
                f"Invalid id type, expected: {str.__name__}, got: {self.id.__class__.__name__}")
        self._source = doc_source.get("_source", {})
        if not isinstance(self._source, dict):
            raise InvalidDocumentSourceError(
                f"Invalid source type, expected: {dict.__name__}, got: {self._source.__class__.__name__}")
```

connectors/byoc.py
wangch079: Yes, we discussed this before, this will work:
```python
class JobStatus(Enum):
    PENDING = "pending"
    IN_PROGRESS = "in_progress"
    CANCELING = "canceling"
    CANCELED = "canceled"
    SUSPENDED = "suspended"
    COMPLETED = "completed"
    ERROR = "error"

# enum to str
JobStatus.PENDING.value

# str to enum
JobStatus("pending")
```

But we also need to consider the edge case when it's `None`, probably as you suggested before, we could have a special `unset` value for each enum.

connectors/byoc.py
wangch079: Eventually I didn't use `isinstance(doc_source, dict)` as the response from the ES call is an `ObjectApiResponse` object, not a `dict`. Here is the change: https://github.com/elastic/connectors-python/pull/460/commits/5d03016809dcc71b19b10fb38df21dbfd79efa46

connectors/byoc.py
wangch079: Done in https://github.com/elastic/connectors-python/commit/5d03016809dcc71b19b10fb38df21dbfd79efa46

connectors/byoc.py
wangch079: Consolidated the proposal in the comment: https://github.com/elastic/connectors-python/pull/464#issuecomment-1429286586

connectors/byoc.py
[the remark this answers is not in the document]
wangch079: Tech-tebt issue created: https://github.com/elastic/enterprise-search-team/issues/3950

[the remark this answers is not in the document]
timgrein: Thank you for creating the issue 👍

connectors/es/document.py
```diff
@@ -0,0 +1,45 @@
...
+#
+
+from connectors.logger import logger
+
+
+class InvalidDocumentSourceError(Exception):
+    pass
+
+
+class ESDocument:
```
tarekziade, line 14: This block can be greatly simplified like this:

```python
class ESDocument:
    def __init__(self, elastic_index, doc_source):
        self.index = elastic_index
        self.id = doc_source.get("_id")
        if not isinstance(self.id, str):
            raise InvalidDocumentSourceError(f"Invalid format found for id: {self.id}")

        self._source = doc_source.get("_source", {})
        if not isinstance(self._source, dict):
            raise InvalidDocumentSourceError(
                f"Invalid format found for source: {self._source}"
            )
```

wangch079: The reason I wrap it with `try... except`, is to handle edge case where some funny value is passed in for `doc_source`, e.g. `None`, or a string like `"foobar"`

connectors/tests/test_es_document.py
```diff
@@ -0,0 +1,54 @@
...
+        ({}, InvalidDocumentSourceError),
+        ({"_id": {}}, InvalidDocumentSourceError),
+        ({"_id": "1", "_source": "hahaha"}, InvalidDocumentSourceError),
+        ({"_id": "1", "_source": {}}, None),
+    ],
+)
+def test_es_document(doc_source, exception, patch_logger):
+    try:
+        ESDocument(elastic_index=None, doc_source=doc_source)
+    except Exception as e:
```
tarekziade, line 25: look at `pytest.raises` https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

connectors/byoc.py
```diff
@@ -181,96 +181,116 @@ async def all_connectors(self):
...
+    async def claim(self):
+        doc = {
+            "status": JobStatus.IN_PROGRESS.value,
+            "started_at": iso_utc(),
+            "last_seen": iso_utc(),
+            "worker_hostname": socket.gethostname(),
+        }
+        await self.index.update(doc_id=self.id, doc=doc)
+
+    async def done(self, ingestion_stats={}, connector_metadata={}):
```
tarekziade, line 258: You can't have this signature:
```
async def done(self, ingestion_stats={}, connector_metadata={}):
```

do:

```
async def done(self, ingestion_stats=None, connector_metadata=None):
    if ingestion_stats is None:
        ingestion_stats = {}
```

reason: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

connectors/es/document.py
wangch079: This is exactly what I proposed initially: https://github.com/elastic/connectors-python/pull/460#discussion_r1105347125, and why I didn't go that way: https://github.com/elastic/connectors-python/pull/460#discussion_r1105420783

I'm also not trying to check `isinstance(doc_source, ObjectApiResponse)`, as I think if a `dict` is provided as the `doc_source`, it should work as well.

connectors/tests/test_es_document.py
wangch079: Yes, I planned to use `pytest.raises` initially, but there are test cases which don't raise errors, and `pytest.raises` won't work.

Then I say this: https://miguendes.me/how-to-check-if-an-exception-is-raised-or-not-with-pytest#heading-how-to-assert-that-no-exception-is-raised, and I think this pattern works well for this particular test.

connectors/byoc.py
wangch079: TIL

wangch079: Done in https://github.com/elastic/connectors-python/pull/460/commits/25427141df45140874ae859b947218f2c950d726

connectors/es/document.py
tarekziade: you should still try to be explicit because having different types allowed for  `doc_source` makes the instanciation of this class a bit fuzzy.

If you want to support several types as inputs, what is usually done is the following:

```
class ESDocument:
    def __init__(self, elastic_index, doc_source):
        self.index = elastic_index

        if not isinstance(doc_source, dict):            
            raise InvalidDocumentSourceError(f"Invalid format found for doc_source: {str(doc_source)}")
 
        self.id = doc_source.get("_id")
        if not isinstance(self.id, str):
            raise InvalidDocumentSourceError(f"Invalid format found for id: {self.id}")

        self._source = doc_source.get("_source", {})
        if not isinstance(self._source, dict):
            raise InvalidDocumentSourceError(
                f"Invalid format found for source: {self._source}"
            )

    @classmethod 
    def from_api_response(cls, api_response):
        return cls(api_response.body)    
```

connectors/tests/test_es_document.py
tarekziade: > there are test cases which don't raise errors, and pytest.raises won't work.

if you don't want to assert that an error was not raised, don't use `pytest.raises` for sure, the problem is your attempt to parameterized even when it's not raising..

here's your test fixed (untested):

```
@pytest.mark.parametrize(
    "doc_source, exception",
    [None,  "hahaha", {}, {"_id": {}}, {"_id": "1", "_source": "hahaha"}] 
)
def test_es_document_raise(doc_source, patch_logger):
    with pytest.raises(InvalidDocumentSourceError):
        ESDocument(elastic_index=None, doc_source=doc_source)

def test_es_document_ok():
    doc = ESDocument(elastic_index=None, doc_source={"_id": "1", "_source": {}})
    assert something on doc
```

connectors/es/document.py
wangch079: `ObjectApiResponse` is defined in `elastic-transport`, and has method `__getitem__` defined:
https://github.com/elastic/elastic-transport-python/blob/6d6e64ae26a4088801f6df1f3bdb40fea88b9b75/elastic_transport/_response.py#L177-L188

The reason why I don't want to use `isinstance(doc_source, ObjectApiResponse)`:
1. It should work for `dict` as well (actually any object with method `get`)
2. I don't want to add dependency on `elastic_transport` here

Or we could simply do this: https://github.com/elastic/connectors-python/pull/460#discussion_r1106263366, which will raise any possible exception, instead of `InvalidDocumentSourceError`

connectors/tests/test_es_document.py
wangch079: Done here: https://github.com/elastic/connectors-python/pull/460/commits/b2b1060798ce2f2e401951b609d49870c09fd5e2

connectors/es/document.py
tarekziade: > It should work for dict as well (actually any object with method get)

Can you elaborate on this?  Do you mean it should support the mapping interface? Will we really use this class for any kind of object ? 

> I don't want to add dependency on elastic_transport here

In that case you can check for `get` presence:

```python
if not hasattr(data_source, 'get'):
    raise InvalidDocumentSourceError("Expected a class with get")
```

But it would be best to be explicit about what the class expects IMHO. 
You could also make it compatible only with dict and ask the caller to take care of passing a dict, so you separate concerns. THe code that manipulates `ObjectApiResponse` could simply do:

```
ESDocument(index, my_response.body)
```

connectors/tests/test_sync.py
```diff
@@ -700,5 +730,3 @@ async def test_concurrent_syncs(mock_responses, patch_logger, set_env):
 
     # make sure we synced the three connectors
     patch_logger.assert_present("[1] Sync done: 1 indexed, 0  deleted. (0 seconds)")
-    patch_logger.assert_present("[2] Sync done: 1 indexed, 0  deleted. (0 seconds)")
```
tarekziade, line 703: why do you remove this coverage? we want to make sure we sync the 3  connectors

wangch079: Because I will soon rewrite this test when I refactor `SyncService`, where I will mock `ConcurrentTasks`, and 
```python
calls = [
    call(functools.partial(self._one_sync, connector1, es)),
    call(functools.partial(self._one_sync, connector2, es)),
    functools.partial(self._one_sync, connector3, es),
]
syncs.put.assert_has_awaits(calls)
```

I will fix the test in this PR.

connectors/es/document.py
wangch079: >Can you elaborate on this?

If a caller wants to instantiate an `ESDocument` object, he/she can pass in an `ObjectApiResponse` object (as what we do in `connectors-python`:
```python
resp = a_client.get(index=".elastic-connectors`, id="1") # an ObjectApiResponse object
doc = ESDocument(an_index, resp)
```
he/she can also pass in a dict:
```python
connector_doc = {
    "_id": "1",
    "_source": {}
}
doc = ESDocument(an_index, connector_doc)
```

Or any object with `get` method.

But again, `YAGNI`.

The reason why I don't want to handle `ObjectApiResponse` explicitly, it's that I'm hesitate to have `ObjectApiResponse` imported as a dependency:
```python
# _response module starts with _ implies it's not supposed to be used directly?
from elastic_transport._response import ObjectApiResponse

class ESDocument:
    def __init__(self, elastic_index, doc_source):
        self.index = elastic_index
        if not isinstance(doc_source, ObjectApiResponse):
            raise InvalidDocumentSourceError(
                f"Invalid doc_source type found: {doc_source.__class__.__name__}, expected: {ObjectApiResponse.__class__.__name__}"
            )
        ...
```

tarekziade: But what is the intent of `ESDocument` -- can you add a two-sentence description in its docstring?

Why a `dict` or an `ObjectApiResponse`?
Will it ever be instantiated with something else than a document from an Elasticsearch client? 
It seems to me that we will never use that class with different type of data.

In any case, if you want both, you don't need to import `elastic_transport`, what prevents you to do: 

```
resp = a_client.get(index=".elastic-connectors`, id="1") # an ObjectApiResponse object
doc = ESDocument(an_index, resp.body)
```
Because your class just needs the dict from the body and that's what `ObjectApiResponse.body` returns

connectors/tests/test_sync.py
tarekziade: ah ok, great thanks! let's make sure we don't lose that coverage

connectors/es/document.py
wangch079: >Will it ever be instantiated with something else than a document from an Elasticsearch client?

No, not for now.

> In any case, if you want both, you don't need to import `elastic_transport`, what prevents you to do

Will update.

wangch079: Done here: https://github.com/elastic/connectors-python/pull/460/commits/f7bb6fd9bb919caaad753682cb8335172a60b55e

connectors/tests/test_sync.py
wangch079: Done here: https://github.com/elastic/connectors-python/pull/460/commits/73454007a23342b0b04346d5f2230da9351b6942

tarekziade (approved):
Thanks for all the interesting discussions and bearing with all my comments

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/504)|

### Questions ?

PR #462, opened 2023-02-13
[Fix] delete_indices used a default mutable argument

timgrein:
`delete_indices` used a default mutable argument, which could've lead to weird bugs in the future.

#### Pre-Review Checklist
- [ ] Tested the changes locally by running `make test` and `make ftest NAME=mysql`

### DIFF
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -125,5 +125,7 @@ class ESClient:
             except NotFoundError:
                 raise PreflightCheckError(f"Could not find pipeline {pipeline}")
 
-    async def delete_indices(self, indices=[]):
+    async def delete_indices(self, indices=None):
+        if indices is None:
+            indices = []
         await self.client.indices.delete(index=indices, ignore_unavailable=True)


artem-shelkovnikov (approved):

connectors/es/client.py
```diff
@@ -125,5 +125,7 @@ async def check_exists(self, indices=None, pipelines=None):
             except NotFoundError:
                 raise PreflightCheckError(f"Could not find pipeline {pipeline}")
 
-    async def delete_indices(self, indices=[]):
+    async def delete_indices(self, indices=None):
+        if indices is None:
+            indices = []
```
wangch079, line 130: Should we return directly?

timgrein: Yep we could, sorry already merged it. You can adjust if you want.

wangch079: No, it shouldn't. Will have a follow up to 1) make it mandatory, 2) not call ES if it's empty.

PR #464, opened 2023-02-13
remove e2str and str2e

tarekziade:
Get rid of those functions and use string-based Enums

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -21,7 +21,7 @@ from connectors.filtering.validation import (
 )
 from connectors.logger import logger
 from connectors.source import DataSourceConfiguration, get_source_klass
-from connectors.utils import e2str, iso_utc, next_run, str2e
+from connectors.utils import iso_utc, next_run
 
 CONNECTORS_INDEX = ".elastic-connectors"
 JOBS_INDEX = ".elastic-connectors-sync-jobs"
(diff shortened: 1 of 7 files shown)


wangch079:
Consolidate comments from multiple PRs, here is the suggested implmentation from @tarekziade :

```python
>>> class JobStatus(Enum):
...     PENDING = "pending"
...     IN_PROGRESS = "in_progress"
...     CANCELING = "canceling"
...     CANCELED = "canceled"
...     SUSPENDED = "suspended"
...     COMPLETED = "completed"
...     ERROR = "error"
...     UNSET = None
... 
>>> JobStatus(None)
<JobStatus.UNSET: None>
>>> JobStatus.UNSET.value
>>> JobStatus("pending")
<JobStatus.PENDING: 'pending'>
>>> JobStatus.PENDING.value
'pending'
```

timgrein (commented):
We probably want to have `UNSET` also in `FilteringValidationState`.

wangch079:
> We probably want to have `UNSET` also in `FilteringValidationState`.

`UNSET` is likely to be present in every enum.

wangch079:
🤦‍♂️  I didn't realized this is a PR. I thought it was an issue in https://github.com/elastic/enterprise-search-team, and that's why I put a comment here: https://github.com/elastic/connectors-python/pull/464#issuecomment-1429286586

wangch079 (commented):
LGTM, probably can also add `UNSET` for enum in `connectors/filtering/validation` as suggested by @timgrein

artem-shelkovnikov (approved):

artem-shelkovnikov:
Just a quick question - why not raise error instead of having UNSET?

tarekziade:
> LGTM, probably can also add `UNSET` for enum in `connectors/filtering/validation` as suggested by @timgrein

it's already the case

PR #474, opened 2023-02-14
Update Connector class to only include connector attributes and self-mutating methods

wangch079:
This PR refactors `Connector` class. The changes are:
1. Make `Connector` data representation of a document in `.elastic-connectors`, i.e. once a `Connector` instance is created, you can easily access its member attributes like `connector.status`, `connector.service_type`, etc.
2. Add self-mutating methods to class `Connector`, e.g. `heartbeat` method to send heartbeat, `prepare` method to update `service_type` and `configuration` for a `Build a connector` connector, etc. 
3. Make the system work with the refactored `Connector`.

Note: Some methods like `sync`, `prepare_docs`, `_sync_starts`, and `_sync_done` are still kept in this PR. This is to keep the scope of this PR as small as possible. The logic will eventually be moved to `SyncService` and new `SyncJobRunner` class, and those methods will be removed by then.

Why the changes are needed:

The main purpose of this change is to separate concerns. Currently, there are many actions initiated by `connector`, e.g. to start a sync, it calls `connector.sync()`, which 1) creates and claims the job, 2) updates `last_synced_*` of the connector. But `connector` class should only concern connector actions, and it shouldn't be responsible for creating/claiming/completing a job.

This is especially important when working on job cancelation and reporting job metadata, because I need to handle edge cases where 1) job is deleted but connector exists 2) job exists but connector is deleted 3) both are deleted. It's made much easier to have separate concerns for each class. 

#### Pre-Review Checklist


connectors/tests/test_byoc.py
```diff
@@ -641,31 +712,127 @@ class Client:
...
+    connector = Connector(
+        elastic_index=index, doc_source=DOC_SOURCE_WITH_EDITED_FILTERING
+    )
+    index.fetch_by_id = AsyncMock(return_value=connector)
+    await connector.validate_filtering(validator=validator)
+
+    validator.validate_filtering.assert_awaited()
+    index.update.assert_awaited_with(
+        doc_id=CONNECTOR_ID, doc=expected_validation_update_doc
+    )
```
wangch079, line 835: Weird test failure.

For these two tests, if I run them separately, they all pass. if I run
```
bin/pytest -sv connectors/tests/test_byoc.py
```
or 
```
make test
```

It randomly fails the second one:
```shell
_________________________________________________________________________________________________________________ test_connector_validate_filtering_invalid __________________________________________________________________________________________________________________

patch_logger = <connectors.conftest.Logger object at 0x106828e80>

    @pytest.mark.asyncio
    async def test_connector_validate_filtering_invalid(patch_logger):
        index = Mock()
        index.update = AsyncMock()
        validation_result = FilteringValidationResult(
            state=FilteringValidationState.INVALID,
            errors=[FilterValidationError(ids=[1], messages="something wrong")],
        )
        validator = Mock()
        validator.validate_filtering = AsyncMock(return_value=validation_result)
        expected_validation_update_doc = {
            "filtering": [
                {
                    "domain": DEFAULT_DOMAIN,
                    "draft": DRAFT_FILTERING_DEFAULT_DOMAIN
                    | {"validation": validation_result.to_dict()},
                    "active": ACTIVE_FILTERING_DEFAULT_DOMAIN,
                },
                FILTERING_OTHER_DOMAIN,
            ]
        }
    
        connector = Connector(
            elastic_index=index, doc_source=DOC_SOURCE_WITH_EDITED_FILTERING
        )
        await connector.validate_filtering(validator=validator)
    
>       validator.validate_filtering.assert_awaited()

connectors/tests/test_byoc.py:800: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <AsyncMock name='mock.validate_filtering' id='4407170128'>

    def assert_awaited(self):
        """
        Assert that the mock was awaited at least once.
        """
        if self.await_count == 0:
            msg = f"Expected {self._mock_name or 'mock'} to have been awaited."
>           raise AssertionError(msg)
E           AssertionError: Expected validate_filtering to have been awaited.

/usr/local/Cellar/python@3.10/3.10.7/Frameworks/Python.framework/Versions/3.10/lib/python3.10/unittest/mock.py:2240: AssertionError
```

I added some logs, and confirm that `validator` and `validator.validate_filtering` in two tests are different objects.

I'm still digging, but in case you encountered the same scenario before, cc @tarekziade

connectors/byoc.py
```diff
@@ -578,39 +511,42 @@ def next_sync(self):
...
+
+    async def sync_starts(self):
+        doc = {
+            "last_sync_status": JobStatus.IN_PROGRESS.value,
+            "last_sync_error": None,
+            "status": Status.CONNECTED.value,
+        }
+        await self.index.update(doc_id=self.id, doc=doc)
 
     async def _sync_starts(self, sync_job_index):
```
wangch079, line 530: `_sync_starts` will eventually be removed once job execution is moved out of `Connector` class in a separate PR.

connectors/byoc.py
```diff
@@ -578,39 +511,42 @@ def next_sync(self):
...
-    async def _sync_suspended(self, job):
-        await job.suspend()
-        self.doc_source["last_sync_status"] = job.status.value
-        self.doc_source["last_sync_error"] = None
-        self.doc_source["error"] = None
-        self.doc_source["last_synced"] = iso_utc()
-        await self.sync_doc()
-        logger.info(f"Sync suspended, Job id: {job.job_id}")
-
-    async def _sync_done(self, job, status, result, exception=None):
```
wangch079, line 613: `_sync_done` will eventually be removed once job execution is moved out of `Connector` class in a separate PR.

connectors/byoc.py
```diff
@@ -630,26 +566,38 @@ async def _sync_done(self, job, status, result, exception=None):
...
+            "error": job_error,
+        }
+
+        if job is not None and job.terminated:
+            doc["last_indexed_document_count"] = job.indexed_document_count
+            doc["last_deleted_document_count"] = job.deleted_document_count
+
+        await self.index.update(doc_id=self.id, doc=doc)
 
     async def prepare_docs(self, data_provider, pipeline=None, filtering=None):
```
wangch079, line 602: `prepare_docs` will eventually be removed once job execution is moved out of `Connector` class in a separate PR.

connectors/byoc.py
```diff
@@ -668,71 +616,99 @@ async def prepare_docs(self, data_provider, pipeline=None, filtering=None):
...
+    async def document_count(self):
+        await self.index.client.indices.refresh(
+            index=self.index_name, ignore_unavailable=True
+        )
+        result = await self.index.client.count(
+            index=self.index_name, ignore_unavailable=True
+        )
+        return result["count"]
+
+    async def sync(
```
wangch079, line 699: `sync` will eventually be removed once job execution is moved out of `Connector` class in a separate PR.

Part of this function (before job is claimed) will be moved to `SyncService`, and the rest (after job is created) will be moved to a new class `SyncJobRunner`.

connectors/services/sync.py
```diff
@@ -92,19 +91,27 @@ async def _one_sync(self, connector, es):
...
+                source_klass = self.source_klass_dict[connector.service_type]
                 if connector.features.sync_rules_enabled():
-                    await validate_filtering(
-                        connector, self.connectors, ValidationTarget.DRAFT
+                    await connector.validate_filtering(
+                        validator=source_klass(connector.configuration)
                     )
 
-                await connector.sync(self.sync_job_index, es, self.idling)
+                await connector.sync(
```
wangch079, line 104: Will be refactored in a separate PR.

connectors/tests/test_byoc.py
wangch079: Finally made it work. Here is the root cause:

Both tests use the same doc source `DOC_SOURCE_WITH_EDITED_FILTERING` to instantiate the connector:
https://github.com/elastic/connectors-python/blob/008f48ffa9f4e86481f75bc0d9f952f1462d405c/connectors/tests/test_byoc.py#L795-L797

https://github.com/elastic/connectors-python/blob/008f48ffa9f4e86481f75bc0d9f952f1462d405c/connectors/tests/test_byoc.py#L826-L828

when the first test runs, the method `validate_filtering` will try to update the data source here:
https://github.com/elastic/connectors-python/blob/54096a4afcedf4ffd923d4f1d499e78d983bc3e1/connectors/byoc.py#L687

which also changes `DOC_SOURCE_WITH_EDITED_FILTERING` state to either `valid` or `invalid`. When the 2nd test runs, it's not edited, thus it's not validated.

The fix is to add `deepcopy` in each test:
https://github.com/elastic/connectors-python/blob/54096a4afcedf4ffd923d4f1d499e78d983bc3e1/connectors/tests/test_byoc.py#L795-L797

https://github.com/elastic/connectors-python/blob/54096a4afcedf4ffd923d4f1d499e78d983bc3e1/connectors/tests/test_byoc.py#L826-L828

connectors/byoc.py
```diff
@@ -630,26 +566,38 @@ async def _sync_done(self, job, status, result, exception=None):
...
-            logger.info(f"Sync suspended, Job id: {job.id}")
         else:
             await job.done(ingestion_stats=ingestion_stats)
 
         job = await job.reload()
         await self.sync_done(job)
         logger.info(
             f"[{job_id}] Sync done: {doc_updated + doc_created} indexed, {doc_deleted} "
-            f" deleted. ({int(time.time() - self._start_time)} seconds)"
+            f" deleted. ({int(time.time() - start_time)} seconds)"
```
timgrein, line 576:
```suggestion
            f" deleted. (Took {int(time.time() - start_time)} seconds)"
```

Maybe `{int(time.time() - start_time)}` could also be a variable called `sync_duration` or something similar

connectors/byoc.py
```diff
@@ -668,71 +616,99 @@ async def prepare_docs(self, data_provider, pipeline=None, filtering=None):
...
+                f"Populated service type {configured_service_type} for connector {self.id}"
+            )
 
-        fqn = config["sources"][service_type]
-        try:
-            source_klass = get_source_klass(fqn)
+        if self.configuration.is_empty():
+            if configured_service_type not in config["sources"]:
+                raise ServiceTypeNotSupportedError(configured_service_type)
+            fqn = config["sources"][configured_service_type]
```
timgrein, line 645:
```suggestion
            source_name = config["sources"][configured_service_type]
```

I think `fqn` is a bit cryptic, if you've never heard it before.

connectors/byoc.py
```diff
@@ -668,71 +616,99 @@ async def prepare_docs(self, data_provider, pipeline=None, filtering=None):
...
+                f"Could not update service type/configuration for connector {self.id}"
+            )
+
+    async def validate_filtering(self, validator):
+        draft_filter = self.filtering.get_draft_filter()
+        if not draft_filter.has_validation_state(FilteringValidationState.EDITED):
+            return
 
-        self.source_klass = source_klass
+        logger.info(f"Filtering of connector {self.id} is edited, validating...)")
```
timgrein, line 674:
```suggestion
        logger.info(f"Filtering of connector {self.id} is {FilteringValidationState.EDITED.value}, validating...)")
```

Just to be more robust against possible changes.

connectors/tests/test_sync.py
```diff
@@ -278,6 +278,17 @@ async def set_server_responses(
     mock_responses.post(
         f"{host}/.elastic-connectors-sync-jobs/_refresh", headers=headers, repeat=True
     )
+    mock_responses.post(
```
timgrein, line 281: I think that we should get rid of mock_responses at some point and mock the elasticsearch client directly and not the underlying aiohttp client.

timgrein (commented):
Only nits around some variable names/log messages.

connectors/tests/test_byoc.py
timgrein: I also ran into similar errors as I tend to forget that upper case only indicates something being constant without actually making it constant. Should we use the type hint `Final` (https://docs.python.org/3/library/typing.html#typing.Final) for constants in tests? We could add `mypy` to `make test` and fail when a constant is mutated. cc @tarekziade

connectors/byoc.py
wangch079: One example is `connectors.sources.mysql:MySqlDataSource`, so `fqn` is more accurate than `source_name`. 

I think it's just a convention to use abbreviation in Python and you need to get used to it (I was not initially as well)

connectors/tests/test_sync.py
wangch079: Indeed, but that will happen in the `Refactor SyncService class` PR. In this PR, I just want to make it green.

connectors/byoc.py
wangch079: There are many tests (mainly in `test_sync.py`), which asserts some log presence, e.g:
https://github.com/elastic/connectors-python/blob/e75e7b30f5124596151451d296fca7022bb728af/connectors/tests/test_sync.py#L425

I'm not in favor of this style, but I don't intend to change them in this PR.

The proposed change can be done when I refactor `SyncService`.

connectors/byoc.py
```diff
@@ -630,26 +566,38 @@ async def _sync_done(self, job, status, result, exception=None):
...
         )
 
     async def sync_done(self, job):
-        self.doc_source["last_sync_status"] = job.status.value
-        self.doc_source["last_sync_error"] = job.error
-        self.doc_source["error"] = job.error
-        self.status = (
-            Status.ERROR if job.status == JobStatus.ERROR else Status.CONNECTED
+        job_status = JobStatus.ERROR if job is None else job.status
+        job_error = JOB_NOT_FOUND_ERROR if job is None else job.error
```
artem-shelkovnikov, line 581: I'm not sure this should even happen - raise an error in console but we should not actually save anything IMO.

connectors/byoc.py
```diff
@@ -335,7 +318,9 @@ def __init__(self, filter_=None):
 
         self.advanced_rules = filter_.get("advanced_snippet", {})
         self.basic_rules = filter_.get("rules", [])
-        self.validation = filter_.get("validation", {"state": "", "errors": []})
+        self.validation = filter_.get(
+            "validation", {"state": FilteringValidationState.VALID.value, "errors": []}
+        )
```
artem-shelkovnikov, line 323: Is this change relevant to the PR? Seems like should be done separately.

connectors/byoc.py
timgrein: Makes sense, would love to see removing some of the logging assertions 👍

connectors/byoc.py
```diff
@@ -498,77 +430,78 @@ def status(self, value):
...
+    def last_sync_status(self):
+        return JobStatus(self.get("last_sync_status"))
+
+    async def heartbeat(self, interval):
+        if (
+            self.last_seen is None
+            or (datetime.now(timezone.utc) - self.last_seen).total_seconds() > interval
+        ):
+            logger.info(f"*** Connector {self.id} HEARTBEAT")
+            await self.index.heartbeat(doc_id=self.id)
```
artem-shelkovnikov, line 504: I think else with `logger.debug` is good here.

But even better to move the `if` statement to the caller of the heartbeat. IMO this logic should reside outside of the class.

connectors/byoc.py
```diff
@@ -578,39 +511,42 @@ def next_sync(self):
...
+        if not self.scheduling.get("enabled", False):
             logger.debug("scheduler is disabled")
             return SYNC_DISABLED
-        return next_run(self.scheduling["interval"])
+        return next_run(self.scheduling.get("interval"))
+
+    async def reset_sync_now_flag(self):
+        await self.index.update(doc_id=self.id, doc={"sync_now": False})
+
+    async def sync_starts(self):
```
artem-shelkovnikov, line 522: Why have both `sync_starts` and `_sync_starts`?

connectors/byoc.py
artem-shelkovnikov: Why is `_sync_suspended` deleted?

connectors/byoc.py
```diff
@@ -668,71 +616,101 @@ async def prepare_docs(self, data_provider, pipeline=None, filtering=None):
...
-
-        Returns the source class.
         """
         configured_connector_id = config.get("connector_id", "")
-        configured_service_type = config.get("service_type", "")
+        if self.id != configured_connector_id:
+            return
+
+        if self.service_type is not None and not self.configuration.is_empty():
+            return
```
artem-shelkovnikov, line 627: Both ifs should have some debug logging to tell that something went wrong. Maybe first one should actually be a warning.

artem-shelkovnikov (commented):
Added some comments, but worried again that PR is pretty big and I don't quite understand why the refactoring is done. Can you update title and description to explain why this refactoring is needed to move forward? What will it allow us in future?

connectors/byoc.py
wangch079: It fails one of my tests. But I can make it a separate PR.

connectors/byoc.py
wangch079: > I think else with `logger.debug` is good here.

This was copied from here:
https://github.com/elastic/connectors-python/blob/e75e7b30f5124596151451d296fca7022bb728af/connectors/byoc.py#L565

I keep this because some tests asset presence of this log.


> But even better to move the `if` statement to the caller of the heartbeat. IMO this logic should reside outside of the class.

I can do that.

connectors/byoc.py
wangch079: I want to separate concerns. When you start a sync, you should:
1. claim the job, via `sync_job.claim()`
2. update `last_synced_*` of the connector, via `connector.sync_starts()`.

Currently, there are many logics initiated by the connector, e.g. `connector._sync_starts` will do job creation and claim, and update `last_synced_*`. That's the main purpose of this refactor.

Why have we kept both `sync_starts` and `_sync_starts` now? Because I don't want to make this PR too complicated. Once I refactor `SyncService`, it will 1) create a job via `sync_job_index.create()`, 2) claim the job via `sync_job.claim()`, 3) update `last_synced_*` of the connector via `connector.sync_starts()` (task 2 and 3 will be further refactored to `JobExecutionService` in the future).  

Til then, `_sync_starts` will no longer used and can be removed. The same applies to methods like `sync`, `prepare_docs` of class `Connector`.

connectors/byoc.py
wangch079: `_sync_suspended` is replaced by calling
https://github.com/elastic/connectors-python/blob/544cfea7eeb6bd874592155afd73405a6d917b8d/connectors/byoc.py#L781

connectors/byoc.py
wangch079: It can happen, when a job was deleted during execution (e.g. via DevTools). In such a case, we should update `last_synced_*` of the connector.

connectors/byoc.py
wangch079: This `prepare` method is for `Build a connector` connector, and we expect it always returns here after its first execution.

During the first execution, this method will update the `service_type` (from config) and the `configuration` (from connector implementation) for the `Build a connector `connector.

IMO, we should only log the first execution, and silently return for the rest.

wangch079:
>worried again that PR is pretty big and I don't quite understand why the refactoring is done. Can you update title and description to explain why this refactoring is needed to move forward? What will it allow us in future?

Hi @artem-shelkovnikov , updated the description. Let me know if it's still not clear.

connectors/byoc.py
wangch079: Made the change in a spearate PR: #495

connectors/byoc.py
wangch079: Think twice, I think it also makes sense to let `Connector` handle when to send heartbeat. If we move the `if` statement to the caller(i.e. `SyncService`), it will be `SyncService` which controls when to send heartbeat. And `Connector.heartbeat()` method will simply update `last_seen`.

There is a very similar case in method `prepare` of class `Connector`. The `prepare` method will update `service_type` and `configuration` for `Build a connector` connector for the first time. The `prepare` method controls when to update, and when to return early. The caller (`SyncService`) simply calls `connector.prepare(config)`.

For all the native connectors, and all the subsequent calls (ideally) of `Build a connector` connectors, it will return directly from here:

https://github.com/elastic/connectors-python/blob/544cfea7eeb6bd874592155afd73405a6d917b8d/connectors/byoc.py#L626-L627

connectors/tests/test_byoc.py
tarekziade: That opens a can of worms. 

- if we introduce type annotations we're going to dramatically change how we work. But I think that's worth a discussion
- `mypy` is a heavy dependency, if we can avoid it... maybe Google's `pytype` would work for checking this. Or Microsoft's `PyRight` -- to be tested

connectors/byoc.py
tarekziade: Can you add a `TODO` ?

connectors/byoc.py
tarekziade: can you add a `TODO`

connectors/byoc.py
tarekziade: can you add these notes as `TODO` in the patch

connectors/byoc.py
tarekziade: It's not very hard to add as a test fixture (`synced_data` for example) with a few attributes and assertions methods, since the data is present in the classes.

connectors/byoc.py
tarekziade: yeah it's "Fully Qualified Name" -- maybe we could use `class_path` to make it easier to understand

tarekziade:
I have identified that moving `validate_filtering` can be a distinct PR from the rest of this body of work

connectors/byoc.py
wangch079: Done in https://github.com/elastic/connectors-python/pull/474/commits/1224df3fe0836512deffca22102671570b8ca9a7

connectors/byoc.py
wangch079: Done in https://github.com/elastic/connectors-python/pull/474/commits/1224df3fe0836512deffca22102671570b8ca9a7

connectors/byoc.py
wangch079: Done in https://github.com/elastic/connectors-python/pull/474/commits/1224df3fe0836512deffca22102671570b8ca9a7

wangch079:
> I have identified that moving `validate_filtering` can be a distinct PR from the rest of this body of work

I have reverted the changes in this commit: https://github.com/elastic/connectors-python/pull/474/commits/6f7d1c392f53122d561c01c74fb7f0033a9c5b4b, and introduced a new PR #498 , which includes changes in this PR. So #498 is easier to review after merge of this PR.

artem-shelkovnikov:
> Hi @artem-shelkovnikov , updated the description. Let me know if it's still not clear.

Thanks, it gives better hint about what to check! What do you think about changing title to "Move non-connector related logic from Connector to other classes" or similar?

wangch079:
>What do you think about changing title to "Move non-connector related logic from Connector to other classes" or similar?

Hi @artem-shelkovnikov , think again, I don't think this title is appropriate as I still maintain some non-connector related logic in `Connector`. I'd like to keep the title brief and have detailed description in PR description (just updated).

artem-shelkovnikov:
@wangch079  Main point for me is that PR title is something that you see when doing `git blame` on a file or checking blame in the Github interface and "Refactor connector class" might cause confusion. It's not a super big deal, but I think we need to make our PR titles more scannable for this use-case.

connectors/byoc.py
```diff
@@ -55,19 +59,6 @@ class JobTriggerMethod(Enum):
     UNSET = None
 
 
-CUSTOM_READ_ONLY_FIELDS = (
```
artem-shelkovnikov, line 58: It was removed because this logic is now a part of Connector class and properties/methods that it provides to mutate the object state.

connectors/byoc.py
```diff
@@ -89,25 +80,6 @@ def __init__(self, elastic_config):
         logger.debug(f"ConnectorIndex connecting to {elastic_config['host']}")
         # initialize ESIndex instance
         super().__init__(index_name=CONNECTORS_INDEX, elastic_config=elastic_config)
-        # grab all bulk options
-        self.bulk_options = elastic_config.get("bulk", {})
-
-    async def save(self, connector):
```
artem-shelkovnikov, line 95: Same - `save` is now a part of `Connector`

connectors/byoc.py
```diff
@@ -580,39 +498,44 @@ def next_sync(self):
...
             return 0
-        if not self.scheduling["enabled"]:
+        if not self.scheduling.get("enabled", False):
             logger.debug("scheduler is disabled")
             return SYNC_DISABLED
-        return next_run(self.scheduling["interval"])
+        return next_run(self.scheduling.get("interval"))
+
+    async def reset_sync_now_flag(self):
+        await self.index.update(doc_id=self.id, doc={"sync_now": False})
```
artem-shelkovnikov, line 507: Note from discussion: either update _source too, or somehow indicate that you need to reload the object so that it's up-to-date.

connectors/services/sync.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, config):
         self.concurrent_syncs = self.service_config.get(
             "max_concurrent_syncs", DEFAULT_MAX_CONCURRENT_SYNCS
         )
+        self.bulk_options = self.es_config.get("bulk", {})
```
artem-shelkovnikov, line 47: Technical change - before it was passed into Connector

connectors/services/sync.py
```diff
@@ -85,7 +86,7 @@ async def _one_sync(self, connector, es):
 
         try:
             # the heartbeat is always triggered
-            connector.start_heartbeat(self.hb)
+            await connector.heartbeat(self.hb)
```
artem-shelkovnikov, line 89: For now if job takes too long to run, connectors will not heartbeat until jobs are complete

artem-shelkovnikov: Should we address it @wangch079?

connectors/tests/test_byoc.py
```diff
@@ -369,48 +348,215 @@ async def test_all_connectors(mock_responses):
...
+    "sync_now, scheduling_enabled, expected_next_sync",
+    [
+        (True, False, 0),
+        (False, False, SYNC_DISABLED),
+        (False, True, 10),
+    ],
+)
+@patch("connectors.byoc.next_run")
+async def test_connector_next_sync(
+    next_run, sync_now, scheduling_enabled, expected_next_sync, patch_logger
```
artem-shelkovnikov, line 547:
```suggestion
    next_run, sync_now, scheduling_enabled, expected_next_sync
```

artem-shelkovnikov (dismissed):
Looks good, but change about heartbeat should be addressed IMO

connectors/services/sync.py
wangch079: Hi @artem-shelkovnikov , I looked deeper, and I'm inclined not to revert the change, because:
1. the previous `start_heartbeat` methods starts a task to periodically send heartbeat
2. there must be a close method for the connector so that the task can be canceled.

https://github.com/elastic/connectors-python/blob/02597f5d8c449ff02839e37b44de47742379a8dd/connectors/byoc.py#L544-L551

Instead, we can have a workaround here:

https://github.com/elastic/connectors-python/blob/52abdd686a07f41c81b160d2693da2937621f084/connectors/sync_job_runner.py#L93-L94

The changes are:

```python
    self.connector.heartbeat(interval)
    await asyncio.sleep(JOB_CHECK_INTERVAL) 
```

It's much easier and clean. It can also be removed easily once we have the decoupling ready. WDYT?

artem-shelkovnikov: > The changes are:
> 
> while not self.elastic_server.done():
>     self.connector.heartbeat(interval)
>     await asyncio.sleep(JOB_CHECK_INTERVAL) 
> 
> It's much easier and clean. It can also be removed easily once we have the decoupling ready. WDYT?

As long as it works I'm good with it :) 

I'm not sure about the code that you mentioned - will it not block if syncs take too much to execute?

wangch079: >We need to close our resources when we're finishing the sync.

Good call, this is moved to `SyncJobRunner` class. I will add close in the `finally` block

>will it not block if syncs take too much to execute?

That is inside the job execution itself, so it doesn't matter how long the job takes.

PR #477, opened 2023-02-15
[Fix] Store filtering in its original form in .elastic-connectors-sync-job

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3950 ##

The frontend expects the advanced snippet payload to be stored under "value" in .elastic-connectors-sync-jobs to correctly show it in the sync-flyout. As ES doesn't have a dedicated array type it'll store a single filtering/filter instance as object directly. This will therefore be fixed in the FE with https://github.com/elastic/kibana/pull/151175.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -21,7 +21,7 @@ from connectors.filtering.validation import (
 )
 from connectors.logger import logger
 from connectors.source import DataSourceConfiguration, get_source_klass
-from connectors.utils import e2str, iso_utc, next_run, str2e
+from connectors.utils import iso_utc, next_run
 
 CONNECTORS_INDEX = ".elastic-connectors"
 JOBS_INDEX = ".elastic-connectors-sync-jobs"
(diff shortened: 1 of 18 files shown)


connectors/byoc.py
```diff
@@ -309,14 +305,12 @@ def __init__(self, filter_=None):
...
-        advanced_rules = filter_.get("advanced_snippet", {})
-
-        self.advanced_rules = advanced_rules.get("value", advanced_rules)
+        self.advanced_rules = filter_.get("advanced_snippet", {})
         self.basic_rules = filter_.get("rules", [])
         self.validation = filter_.get("validation", {"state": "", "errors": []})
 
     def get_advanced_rules(self):
-        return self.advanced_rules
+        return self.advanced_rules.get("value", self.advanced_rules)
```
artem-shelkovnikov, line 313: So we return content of "advanced_snippet.value" if there's anything inside, otherwise content of "advanced_snippet"?

timgrein: It's probably better to return `{}` as default. Shouldn't happen anyway, but it's more obvious, that "value" doesn't exist.

artem-shelkovnikov: Yeah I saw original defaulting logic used `{}`, probably indeed better default to it

timgrein: Adjusted it 👍

artem-shelkovnikov (approved):

wangch079:
Hi @timgrein , are you going to backport this to 8.7? If so, I think you should add the label `v8.7.0` instead of `v8.7.0.0`

seanstory:
The backportrc utilizes the first two digits after the `v`, so either should work. But I bet this PR's source branch wasn't rebased onto `main` after the backportrc got changed to map `8.7` to a branch vs main, which is why the backport failed. I'd recommend manually running the backport command, Tim, and just linking the PR.

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/540)|

### Questions ?

wangch079:
Manually generated the backport PR: #540 . Turns out some changes are already in 8.7 branch.

wangch079:
Manually generated the backport PR: #540 . Turns out some changes are already in 8.7 branch.

PR #478, opened 2023-02-15
Remove assert from source.py and replace it with manual validation

artem-shelkovnikov:
`assert` statements are supposed to be used only in tests. We have an `assert` statement in our production code in `BaseDataSource` class. This assert attempts to validate the type of `configuration` variable passed into the class.

This PR removes `assert` and replaces it with a regular `if` statement that checks type of the variable and raises a `TypeError` in case the type is unexpected. Choice of exception type was made based on https://docs.python.org/3/library/exceptions.html#TypeError

#### Pre-Review Checklist

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -115,8 +115,12 @@ class BaseDataSource:
     service_type = None
 
     def __init__(self, configuration):
+        if not isinstance(configuration, DataSourceConfiguration):
+            raise TypeError(
+                f"Configuration expected type is DataSourceConfiguration, actual: {type(configuration).__name__}."
+            )
+
         self.configuration = configuration
-        assert isinstance(self.configuration, DataSourceConfiguration)
         self.configuration.set_defaults(self.get_default_configuration())
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


navarone-feekery (approved):

connectors/source.py
```diff
@@ -115,8 +115,12 @@ class BaseDataSource:
     service_type = None
 
     def __init__(self, configuration):
+        if not isinstance(configuration, DataSourceConfiguration):
+            raise TypeError(
+                f"Configuration expected type is DataSourceConfiguration, actual: {type(configuration).__name__}."
```
timgrein, line 120:
```suggestion
                f"Configuration expected type is {type(self).__name__}, actual: {type(configuration).__name__}."
```

Would make it a bit more robust to a change of the class name.

timgrein (approved):
Small nit about making a log message more dynamic, otherwise LGTM.

connectors/source.py
artem-shelkovnikov: Good call!

tarekziade (approved):

PR #479, opened 2023-02-15
[Docs] Move connector protocol to python

timgrein:
I think it makes more sense to have the connector protocol in the python repository as we're moving onwards with python.

TODOs:
- [ ] Change links in ent-search docs
- [ ] Change links in connectors-ruby
- [ ] Change links in connectors-python
- [ ] (Create a redirect in connectors-ruby?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,6 @@ If you are looking for the implementation in Ruby, See [connectors-ruby](https:/
 - [Developer guide](docs/DEVELOPING.md)
 - [Security Policy](docs/SECURITY.md)
 - [Elastic-internal guide](docs/INTERNAL.md)
-- [Connector Protocol](https://github.com/elastic/connectors-python/docs/CONNECTOR_PROTOCOL.md)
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov:
Wanted to double check - have you changed the Ruby protocol file to point here instead?

tarekziade:
@timgrein can you create a patch in ruby where you keep the old file and add a link to the new one -- we have links out there on that location, thanks

timgrein:
Will do that, thanks for pointing out. Will also check with docs team where we point from the official docs to the protocol.

leemthompo:
A quick ripgrep of **enterprise-search-pubs** for `https://github.com/elastic/connectors-ruby/blob/main/docs/CONNECTOR_PROTOCOL.md` shows we don't link to this doc currently in the Enterprise Search docs

tarekziade (approved):

leemthompo:
Just thinking out loud but do we have links in blogs to this file that should be redirected?

tarekziade:
> Just thinking out loud but do we have links in blogs to this file that should be redirected?

I thought about https://www.elastic.co/blog/bring-your-own-enterprise-search-connector but nope

PR #480, opened 2023-02-15
Revert "move log from method _sync_done to finally block"

artem-shelkovnikov:
Reverts elastic/connectors-python#431

Reason of revert: https://github.com/elastic/enterprise-search-team/issues/3935

This code does not work because `result` in `if` condition have not been assigned before - assignment happens inside of `try` block, but this code is happening in `finally` code:

```python
            if result is None:
                result = {}
```

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -594,6 +594,10 @@ class Connector:
 
         self.doc_source["last_synced"] = iso_utc()
         await self.sync_doc()
+        logger.info(
+            f"[{self.id}] Sync done: {indexed_count} indexed, {doc_deleted} "
+            f" deleted. ({int(time.time() - self._start_time)} seconds)"
+        )
 
     async def prepare_docs(self, data_provider, filtering=None):
         if filtering is None:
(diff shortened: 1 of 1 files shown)


tarekziade:
> This code does not work

Is there a way to add a test here that shows the problem?

timgrein:
I can confirm that we can run into a "non defined" error with the current code.

tarekziade:
Yeah, creating empty results on the fly there does not make sense.

wangch079:
Hi @artem-shelkovnikov , I think the bug is already fixed with the merge of #460:

https://github.com/elastic/connectors-python/blob/43923d7a271768f4776079875dba9c18236d6b64/connectors/byoc.py#L796-L814

I have also tested it locally and it worked for me. So I think https://github.com/elastic/enterprise-search-team/issues/3935 can be closed.

Meanwhile, I see that the service shutdown is not gracefully handled. Maybe you want to look into this separately. 

```shell
[FMWK][16:52:57][INFO] Caught SIGINT. Graceful shutdown.
[FMWK][16:52:57][WARNING] Retry count: 1 out of 3. Exception: 'NoneType' object has no attribute 'readexactly'
[FMWK][16:52:57][INFO] Sync suspended, Job id: lvttWYYBteCpHA24GvF-
[FMWK][16:52:57][INFO] [lvttWYYBteCpHA24GvF-] Sync done: 0 indexed, 0  deleted. (9 seconds)
[FMWK][16:52:57][INFO] Sync suspended, Job id: lvttWYYBteCpHA24GvF-
[FMWK][16:52:57][ERROR] Service did not handle cancellation gracefully.
[FMWK][16:52:57][INFO] Bye
```

artem-shelkovnikov:
Indeed @wangch079 it's fixed with your PR.

As for the error you see, it does not look like a breaking bug and I'm planning to address it for 8.8 release

artem-shelkovnikov:
Closing as fixed by a different PR

PR #485, opened 2023-02-15
Fix problem with index_name being unavailable for a job

artem-shelkovnikov:
Fixes a problem reported in Slack:

```
[FMWK][15:05:17][CRITICAL] 'index_name'
  File "/Users/artemshelkovnikov/git_tree/connectors-py/connectors/services/job_cleanup.py", line 64, in _process_orphaned_jobs
    job.index_name is not None
  File "/Users/artemshelkovnikov/git_tree/connectors-py/connectors/byoc.py", line 208, in index_name
    return self.doc_source["_source"]["connector"]["index_name"]
KeyError: 'index_name'
```

Problem happens because when job is created the property "source.index_name" is never populated, while logic that cleans up jobs uses this key.

This fix is suboptimal - better refactor the whole structure of classes, but I know @wangch079 is addressing some of it in his PRs, thus I'm proposing this temporary fix for the problem until refactoring lands.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -207,10 +207,14 @@ class SyncJob:
     def index_name(self):
         return self.doc_source["_source"]["connector"]["index_name"]
 
-    async def start(self, connector, trigger_method=JobTriggerMethod.SCHEDULED, filtering=None):
-        # Okay that's temporary here. When we separate scheduling from execution , this code will go away
+    async def start(
+        self, connector, trigger_method=JobTriggerMethod.SCHEDULED, filtering=None
+    ):
+        # TODO: that's temporary here. When we separate scheduling from execution, this code will go away
         if connector.id != self.connector_id:
-            raise Exception("Job.connector_id is not equal to connector.id that is passed into the method")
+            raise Exception(
+                "Job.connector_id is not equal to connector.id that is passed into the method"
+            )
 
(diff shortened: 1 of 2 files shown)


connectors/byoc.py
```diff
@@ -207,7 +207,11 @@ def duration(self):
     def index_name(self):
         return self.doc_source["_source"]["connector"]["index_name"]
 
-    async def start(self, trigger_method=JobTriggerMethod.SCHEDULED, filtering=None):
+    async def start(self, connector, trigger_method=JobTriggerMethod.SCHEDULED, filtering=None):
+        # Okay that's temporary here. When we separate scheduling from execution , this code will go away
```
tarekziade, line 211: I would add `XXX` or `TODO` as a prefix to this comment

tarekziade (approved):

wangch079 (approved):
This will be eventually fixed after #460 and #474 are merged. But let's have a working `main` first.

connectors/byoc.py
timgrein: ++ for "# TODO:", which is caught by nearly every IDE and shows up in the corresponding TODOs tab.

artem-shelkovnikov:
Closing as not needed any more

PR #489, opened 2023-02-16
Make param indices of delete_indices method mandatory

wangch079:
## Follow-up on PR https://github.com/elastic/connectors-python/pull/462

This PR
1. makes the param `indices` of method `ESClient.delete_indices` mandatory, 
2. returns early without making ES request if `indices` is empty.

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -125,7 +125,8 @@ class ESClient:
             except NotFoundError:
                 raise PreflightCheckError(f"Could not find pipeline {pipeline}")
 
-    async def delete_indices(self, indices=None):
-        if indices is None:
-            indices = []
+    async def delete_indices(self, indices):
+        if indices is None or len(indices) == 0:
+            logger.debug("indices are not provided, skipping...")
+            return
         await self.client.indices.delete(index=indices, ignore_unavailable=True)
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -126,7 +126,4 @@ class ESClient:
                 raise PreflightCheckError(f"Could not find pipeline {pipeline}")
 
     async def delete_indices(self, indices):
-        if indices is None or len(indices) == 0:
-            logger.debug("indices are not provided, skipping...")
-            return
         await self.client.indices.delete(index=indices, ignore_unavailable=True)
--- a/connectors/tests/test_es_client.py
+++ b/connectors/tests/test_es_client.py
@@ -5,6 +5,7 @@
 #
 import base64
 from unittest import mock
+from unittest.mock import AsyncMock, Mock
 
 import pytest
 from elasticsearch import ConnectionError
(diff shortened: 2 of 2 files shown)


connectors/es/client.py
```diff
@@ -125,7 +125,8 @@ async def check_exists(self, indices=None, pipelines=None):
             except NotFoundError:
                 raise PreflightCheckError(f"Could not find pipeline {pipeline}")
 
-    async def delete_indices(self, indices=None):
-        if indices is None:
-            indices = []
+    async def delete_indices(self, indices):
+        if indices is None or len(indices) == 0:
```
tarekziade, line 129: You are still allowing `indices` to be `None` -- do we really want to support a call to `delete_indices` to be a no-op ?

tarekziade (approved):

tarekziade (changes requested):
the only usage of that method is here: 

https://github.com/elastic/connectors-python/blob/main/connectors/services/job_cleanup.py#L75-L76

where you check that the indices are not empty, so I think `delete_indices` can directly pass the call to `self.client`

connectors/es/client.py
wangch079: Yes, the caller can still call `an_index.delete_indices(indices=None)`

wangch079:
> where you check that the indices are not empty, so I think `delete_indices` can directly pass the call to `self.client`

Indeed, but I don't think we should define the method based on it's caller's implementation.

connectors/es/client.py
tarekziade: > > where you check that the indices are not empty, so I think `delete_indices` can directly pass the call to `self.client`
> 
> Indeed, but I don't think we should define the method based on it's caller's implementation.

But we're not implementing an API for hypothetical ways of calling it -- it's added for actual use cases, and right now the one single use case is to delete indices. Why would you call it if you don't intend to delete some indices? what would be the actual use case of passing `None` to it?  I don't see that requirement/need, and I think allowing `None` and silently ignoring it by returning is a recipe for problems.

tarekziade:
> > where you check that the indices are not empty, so I think `delete_indices` can directly pass the call to `self.client`
> 
> Indeed, but I don't think we should define the method based on it's caller's implementation.

But we're not implementing an API for hypothetical ways of calling it -- it's added for actual use cases, and right now the one single use case is to delete indices. Why would you call it if you don't intend to delete some indices? what would be the actual use case of passing None to it? I don't see that requirement/need, and I think allowing None and silently ignoring it by returning is a recipe for problems.

wangch079:
Done here: https://github.com/elastic/connectors-python/pull/489/commits/0e802356e276461378db8ffe8e5a3406f6086fa9

tarekziade (approved):

PR #491, opened 2023-02-16
Add SyncJobRunner class

wangch079:
### Changes made:

This PR adds a `SyncJobRunner` class, which takes a pending job and executes it.

The current job execution is implemented in `Connector.sync` method:

https://github.com/elastic/connectors-python/blob/029039390758ea2194dfdce86babf453eb425f46/connectors/byoc.py#L678

which combines job scheduling and execution.

In `Connector.sync` method, all the logic before this line (including job creation defined in `_sync_starts` method) should be moved to `SyncService` (Done in #500)

https://github.com/elastic/connectors-python/blob/029039390758ea2194dfdce86babf453eb425f46/connectors/byoc.py#L722

And all the logic after this line (including job claim defined in `_sync_starts` method) should be moved to `SyncJobRunner.execute` (Done in #500)

### Why the change?

1. We should make `Connector` clean and it's not responsible for job scheduling and execution
2. We should separate job scheduling (should be handled by `SyncService`) and job execution (currently handled by `SyncService` via `SyncJobRunner`, but will be handled by `JobExecutionService` in 8.8)

### Where is `SyncJobRunner` used

The `SyncJobRunner` class takes a `sync_job`, and manages its execution from `claim` to `complete`.

In 8.7, `SyncService` will initiate the job execution:

```python
job_id = await self.sync_job_index.create(connector)
if connector.sync_now:
    await connector.reset_sync_now_flag()
sync_job = await self.sync_job_index.fetch_by_id(job_id)
sync_job_runner = SyncJobRunner(
    source_klass=source_klass,
    sync_job=sync_job,
    connector=connector,
    elastic_server=es,
    bulk_options=self.bulk_options,
)
await self.syncs.put(sync_job_runner.execute)
```

In 8.8, we will have the decoupling of job scheduling and execution. So the job of `SyncService` is done after the first 3 lines of the above code

The `JobExecutionService` will take responsibility of the rest:
```python
async for sync_job in self.sync_job_index.pending_jobs():
    connector = self.connector_index.fetch_by_id(sync_job.connector_id)
    sync_job_runner = SyncJobRunner(
        source_klass=source_klass,
        sync_job=sync_job,
        connector=connector,
        elastic_server=es,
        bulk_options=self.bulk_options,
    )
    await self.syncs.put(sync_job_runner.execute)
```

#### Pre-Review Checklist


connectors/sync_job_runner.py
```diff
@@ -0,0 +1,151 @@
...
+        if self.running:
+            raise SyncJobRunningError(f"Sync job #{self.job_id} is already running.")
+
+        self.running = True
+        if not await self._claim_job():
+            logger.error(f"Unable to claim job #{self.job_id}")
+            return
+
+        try:
+            data_provider = self.source_klass(self.sync_job.configuration)
```
timgrein, line 49: Should we call this `data_source` for consistency? The class is `BaseDataSource` and not `BaseDataProvider`.

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,151 @@
...
+            if not await data_provider.changed():
```
timgrein, line 50: Do we actually implement this method in a meaningful way somewhere? It's always `true` in the base class and `directory.py` implements it also always returning `true`.

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,151 @@
...
+                logger.debug(
+                    f"No change in {self.sync_job.service_type} data provider, skipping..."
+                )
+                await self._sync_done(sync_status=JobStatus.COMPLETED, result={})
+                return
+
+            logger.debug(f"Syncing '{self.sync_job.service_type}'")
```
timgrein, line 57:
```suggestion
            logger.debug(f"Syncing '{self.sync_job.service_type}' for connector '{self.connector_id}'")
```

As the framework supports running multiple connectors per connector server/instance I think it's useful to distinguish on connector id level and not only `service_type`.

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,151 @@
...
+        self.running = False
+
+    async def execute(self):
+        if self.running:
+            raise SyncJobRunningError(f"Sync job #{self.job_id} is already running.")
+
+        self.running = True
+        if not await self._claim_job():
+            logger.error(f"Unable to claim job #{self.job_id}")
+            return
```
timgrein, line 46: Should we also raise something here to the caller?

connectors/tests/test_sync_job_runner.py
```diff
@@ -0,0 +1,249 @@
...
+        source_klass=source_klass,
+        sync_job=sync_job,
+        connector=connector,
+        elastic_server=elastic_server,
+        bulk_options=bulk_options,
+    )
+
+
+@pytest.mark.asyncio
+async def test_job_claim_fail(patch_logger):
```
timgrein, line 70: Same as in the other PR. Did we agree on silencing the logger during unit tests? Could make debugging harder in CIs.

timgrein (commented):
Left a comment around a log message and some smaller nits.

artem-shelkovnikov (commented):
The PR looks good, but since it's not used in real code it's hard to tell if it breaks anything or not.

Probably it can be merged and then a different PR to use it introduced, and `ftest` will be a good indication that everything works?

wangch079:
>Probably it can be merged and then a different PR to use it introduced, and ftest will be a good indication that everything works?

It will be used in #500

connectors/sync_job_runner.py
wangch079: Makes sense.

connectors/sync_job_runner.py
wangch079: Copied from previous code. Let's change it in a separate PR if needed.

connectors/sync_job_runner.py
wangch079: Done here: https://github.com/elastic/connectors-python/pull/491/commits/ab938242d7e559fb9870bcc7fe2d173600c07f31

connectors/sync_job_runner.py
wangch079: Done here: https://github.com/elastic/connectors-python/pull/491/commits/4d7c3a9375d5079a2cfa41749edb9826aba92213

connectors/sync_job_runner.py
tarekziade: the initial intent was to bypass a sync if we know nothing has changed. I used it in the first version of the Mongo connector. In S3/GCP it can be implemented by looking a the bucket last changed timestamp

tarekziade:
Can you elaborate on why this class is added and where/how it'll be used? It's adding without removing code, it's hard to review in isolation.

maybe the summary can give a list of classes and how they will interact? and the class docstring extended

tarekziade:
If we are migrating code, we need this patch to include the rest where you delete the code you are moving in this class

wangch079:
> Can you elaborate on why this class is added and where/how it'll be used? 
> 
> maybe the summary can give a list of classes and how they will interact? 

Added in the PR description

>and the class docstring extended

Will do

>If we are migrating code, we need this patch to include the rest where you delete the code you are moving in this class

This will be done in #500. I just want to make the scope of the PR as small as possible, and easy to review.

tarekziade:
> This will be done in https://github.com/elastic/connectors-python/pull/500. I just want to make the scope of the PR as small as possible, and easy to review.

I would split the patches a bit differently by having here the deletion of the code you are moving to this class because it's harder to review if we don't know exactly where things are being removed.

tarekziade:
Thanks for the added explanation. If the goal is to split part of the `Connector` class into another class, it's going to be easier to review this as a single PR. I would expect this PR to do almost zero change to the code itself but move some methods out of `Connector` and into `SyncJobRunner`

wangch079:
>I would split the patches a bit differently by having here the deletion of the code you are moving to this class

That also means I need to make the change to wherever uses the deleted code. 

> I would expect this PR to do almost zero change to the code itself but move some methods out of `Connector` and into `SyncJobRunner`

Not exactly. Part of `Connector.sync` is moved to `SyncService` (in #500), and the rest are moved here.

Actually, if you want to see the deleted code and change of usage, you can change the base branch of #500 to `main`.

But there are conflicts in #500, I will resolve it now.

tarekziade:
> Not exactly. Part of `Connector.sync` is moved to `SyncService` (in #500), and the rest are moved here.

The intent is to have each patch move the project in a working state. If the refactoring of `Connector.sync` leads to build two new classes - can you do this separately? e.g. can you first extract `SyncJobRunner` and things work as expected, and then do `SyncService`? 

If not, I am afraid we need to have both patches done in a single patch, and make it as small as possible. 
I've already identified some stuff you do in #500 that can be a different PR.

For example, you are adding `_prepare_connector` -- this can be done independently from the rest. It has zero impact, and zero code change as far as I can see. It just moves some code to a new function. This is the little things that make those PR bigger and harder to review

wangch079:
Hi @tarekziade , I just change the base branch of #500 to `main`, and it's not that bad. It only adds the two files in this PR. Another bigger change I made in #500 is to rewrite the `test_sync.py`, see [here](https://github.com/elastic/connectors-python/pull/500#issuecomment-1438565195) for the explanation.

image

wangch079:
Close this PR as it will be included in #500

PR #493, opened 2023-02-16
[GCS]-Parse configurations from Kibana UI

jignesh-crest:
## Closes https://github.com/elastic/connectors-py/issues/466

This PR addresses the following changes:
- Parse the private-key present in the configurations passed from UI.
- Moves the configuration validation to a separate method.

#### Pre-Review Checklist
- [x] Tested the changes locally by running `make test` and `make ftest NAME=google_cloud_storage`
- [ ] ~~Considered corresponding documentation changes~~
- [ ] ~~Contributed any configuration settings changes to the configuration reference~~

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -20,6 +20,16 @@ from connectors.filtering.validation import (
 """
 
 
+class DataSourceConfigurationError(Exception):
+    """Raise when configurations are invalid.
+
+    Args:
+        Exception (DataSourceConfigurationError): Invalid configurations exception.
+    """
+
+    pass
+
+
 class Field:
     def __init__(self, name, label=None, value="", type="str"):
         if label is None:
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -20,16 +20,6 @@ from connectors.filtering.validation import (
 )
 
 
-class DataSourceConfigurationError(Exception):
-    """Raise when configurations are invalid.
-
-    Args:
-        Exception (DataSourceConfigurationError): Invalid configurations exception.
-    """
-
-    pass
-
-
 class Field:
     def __init__(self, name, label=None, value="", type="str"):
         if label is None:
--- a/connectors/sources/google_cloud_storage.py
+++ b/connectors/sources/google_cloud_storage.py
@@ -10,7 +10,6 @@ import json
 import os
 import urllib.parse
 from functools import partial
-from json.decoder import JSONDecodeError
 
 import aiofiles
 from aiofiles.os import remove
@@ -19,7 +18,7 @@ from aiogoogle import Aiogoogle
 from aiogoogle.auth.creds import ServiceAccountCreds
 
 from connectors.logger import logger
-from connectors.source import BaseDataSource, DataSourceConfigurationError
+from connectors.source import BaseDataSource
 from connectors.utils import TIKA_SUPPORTED_FILETYPES, convert_to_b64
 
 CLOUD_STORAGE_READ_ONLY_SCOPE = "https://www.googleapis.com/auth/devstorage.read_only"
@@ -73,13 +72,11 @@ class GoogleCloudStorageDataSource(BaseDataSource):
             configuration (DataSourceConfiguration): Object of DataSourceConfiguration class.
         """
         super().__init__(configuration=configuration)
-
         self.retry_count = self.configuration["retry_count"]
         self.enable_content_extraction = self.configuration["enable_content_extraction"]
 
         self.service_account_credentials = None
         self.user_project_id = None
-        self.json_creds = None
 
     @classmethod
     def get_default_configuration(cls):
@@ -115,6 +112,23 @@ class GoogleCloudStorageDataSource(BaseDataSource):
                 "type": "bool",
             },
         }
+
+    async def validate_config(self):
+        """Validates whether user inputs are valid or not for configuration field.
+
+        Raises:
+            Exception: The service account json is empty.
+            Exception: The format of service account json is invalid.
+        """
+        if (
+            self.configuration["service_account_credentials"] == ""
+            or self.configuration["service_account_credentials"] is None
+        ):
+            raise Exception("Google Cloud service account json can't be empty.")
+        try:
+            _ = json.loads(self.configuration["service_account_credentials"])
+        except ValueError:
+            raise Exception("Google Cloud service account is not a valid JSON.")
 
     async def _api_call(
         self,
@@ -187,24 +201,6 @@ class GoogleCloudStorageDataSource(BaseDataSource):
                 )
                 await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry_counter)
 
-    def _validate_configurations(self):
-        """Validates whether user inputs are valid or not for configuration field.
-
-        Raises:
-            DataSourceConfigurationError: The service account json is empty.
-            DataSourceConfigurationError: The service account json is in invalid format.
-        """
-        if self.configuration["service_account_credentials"] == "":
-            raise DataSourceConfigurationError("Service account json can't be empty.")
-        try:
-            self.json_creds = json.loads(
-                self.configuration["service_account_credentials"]
-            )
-        except JSONDecodeError:
-            raise DataSourceConfigurationError(
-                "Service account json is not in valid format."
-            )
-
     def get_pem_format(self, private_key, max_split=-1):
         """Convert key into PEM format.
 
@@ -222,28 +218,33 @@ class GoogleCloudStorageDataSource(BaseDataSource):
 
     def _initialize_configurations(self):
         """Initialize the ServiceAccountCreds"""
+        if self.service_account_credentials is not None:
+            return
+
+        json_credentials = json.loads(self.configuration["service_account_credentials"])
+
         if (
-            self.json_creds.get("private_key")
-            and "\n" not in self.json_creds["private_key"]
+            json_credentials.get("private_key")
+            and "\n" not in json_credentials["private_key"]
         ):
-            self.json_creds["private_key"] = self.get_pem_format(
-                private_key=self.json_creds["private_key"].strip(), max_split=2
+            json_credentials["private_key"] = self.get_pem_format(
+                private_key=json_credentials["private_key"].strip(),
+                max_split=2,
             )
 
         self.service_account_credentials = ServiceAccountCreds(
             scopes=[CLOUD_STORAGE_READ_ONLY_SCOPE],
-            **self.json_creds,
+            **json_credentials,
         )
         self.user_project_id = self.service_account_credentials.project_id
 
     async def ping(self):
-        """Verify the configurations and connection with Google Cloud Storage"""
-
-        self._validate_configurations()
-        self._initialize_configurations()
+        """Verify the connection with Google Cloud Storage"""
 
         if RUNNING_FTEST:
             return
+
+        self._initialize_configurations()
         try:
             await anext(
                 self._api_call(
@@ -385,6 +386,7 @@ class GoogleCloudStorageDataSource(BaseDataSource):
         Yields:
             dictionary: Documents from Google Cloud Storage.
         """
+        self._initialize_configurations()
         async for buckets in self.fetch_buckets():
             if not buckets.get("items"):
                 continue
--- a/connectors/sources/google_cloud_storage.py
+++ b/connectors/sources/google_cloud_storage.py
@@ -273,8 +273,8 @@ class GoogleCloudStorageDataSource(BaseDataSource):
             resource="buckets",
             method="list",
             full_response=True,
-            project=self.google_storage_client.user_project_id,
-            userProject=self.google_storage_client.user_project_id,
+            project=self.get_storage_client().user_project_id,
+            userProject=self.get_storage_client().user_project_id,
         ):
             yield bucket
 
@@ -293,7 +293,7 @@ class GoogleCloudStorageDataSource(BaseDataSource):
                 method="list",
                 full_response=True,
                 bucket=bucket["id"],
-                userProject=self.google_storage_client.user_project_id,
+                userProject=self.get_storage_client().user_project_id,
             ):
                 yield blob
 
@@ -366,7 +366,7 @@ class GoogleCloudStorageDataSource(BaseDataSource):
                     bucket=blob["bucket_name"],
                     object=blob_name,
                     alt="media",
-                    userProject=self.google_storage_client.user_project_id,
+                    userProject=self.get_storage_client().user_project_id,
                     pipe_to=async_buffer,
                 )
             )
--- a/connectors/sources/google_cloud_storage.py
+++ b/connectors/sources/google_cloud_storage.py
@@ -19,7 +19,7 @@ from aiogoogle.auth.creds import ServiceAccountCreds
 
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import TIKA_SUPPORTED_FILETYPES, convert_to_b64
+from connectors.utils import TIKA_SUPPORTED_FILETYPES, convert_to_b64, get_pem_format
 
 CLOUD_STORAGE_READ_ONLY_SCOPE = "https://www.googleapis.com/auth/devstorage.read_only"
 CLOUD_STORAGE_BASE_URL = "https://console.cloud.google.com/storage/browser/_details/"
@@ -201,23 +201,12 @@ class GoogleCloudStorageDataSource(BaseDataSource):
                 )
                 await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry_counter)
 
-    def get_pem_format(self, private_key, max_split=-1):
-        """Convert key into PEM format.
-
-        Args:
-            private_key (str): Private_key in raw format.
-            max_split (int): Specifies how many splits to do. Defaults to -1.
+    def get_service_account_credentials(self):
+        """Initialize and return the ServiceAccountCreds
 
         Returns:
-            string: PEM format
+            ServiceAccountCreds: An instance of the ServiceAccountCreds.
         """
-        private_key = private_key.replace(" ", "\n")
-        private_key = " ".join(private_key.split("\n", max_split))
-        private_key = " ".join(private_key.rsplit("\n", max_split))
-        return private_key
-
-    def _initialize_configurations(self):
-        """Initialize the ServiceAccountCreds"""
         if self.service_account_credentials is not None:
             return
 
@@ -227,8 +216,8 @@ class GoogleCloudStorageDataSource(BaseDataSource):
             json_credentials.get("private_key")
             and "\n" not in json_credentials["private_key"]
         ):
-            json_credentials["private_key"] = self.get_pem_format(
-                private_key=json_credentials["private_key"].strip(),
+            json_credentials["private_key"] = get_pem_format(
+                key=json_credentials["private_key"].strip(),
                 max_split=2,
             )
 
@@ -237,6 +226,7 @@ class GoogleCloudStorageDataSource(BaseDataSource):
             **json_credentials,
         )
         self.user_project_id = self.service_account_credentials.project_id
+        return self.service_account_credentials
 
     async def ping(self):
         """Verify the connection with Google Cloud Storage"""
@@ -244,7 +234,7 @@ class GoogleCloudStorageDataSource(BaseDataSource):
         if RUNNING_FTEST:
             return
 
-        self._initialize_configurations()
+        self.get_service_account_credentials()
         try:
             await anext(
                 self._api_call(
@@ -386,7 +376,7 @@ class GoogleCloudStorageDataSource(BaseDataSource):
         Yields:
             dictionary: Documents from Google Cloud Storage.
         """
-        self._initialize_configurations()
+        self.get_service_account_credentials()
         async for buckets in self.fetch_buckets():
             if not buckets.get("items"):
                 continue
--- a/connectors/sources/google_cloud_storage.py
+++ b/connectors/sources/google_cloud_storage.py
@@ -59,6 +59,96 @@ DEFAULT_PEM_FILE = os.path.join(
 )
 
 
+class GoogleCloudStorageClient:
+    """A google client to handle api calls made to Google Cloud Storage."""
+
+    def __init__(self, retry_count, json_credentials):
+        """Initialize the ServiceAccountCreds class using which api calls will be made.
+
+        Args:
+            retry_count (int): Maximum retries for the failed requests.
+            json_credentials (dict): Service account credentials json.
+        """
+        self.retry_count = retry_count
+        self.service_account_credentials = ServiceAccountCreds(
+            scopes=[CLOUD_STORAGE_READ_ONLY_SCOPE],
+            **json_credentials,
+        )
+        self.user_project_id = self.service_account_credentials.project_id
+
+    async def _api_call(
+        self,
+        resource,
+        method,
+        sub_method=None,
+        full_response=False,
+        **kwargs,
+    ):
+        """Make a GET call for Google Cloud Storage with retry for the failed API calls.
+
+        Args:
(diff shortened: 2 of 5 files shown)


praveen-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
```diff
@@ -199,8 +185,54 @@ async def _api_call(
                 )
                 await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry_counter)
 
+    def _validate_configurations(self):
+        """Validates whether user input is empty or not for configuration field.
+
+        Raises:
+            Exception: Configured keys can't be empty.
+        """
+        if self.configuration["service_account_credentials"] == "":
```
timgrein, line 194: Should we also check for 'None' here?

connectors/sources/google_cloud_storage.py
```diff
@@ -199,8 +185,54 @@ async def _api_call(
...
+        private_key = private_key.replace(" ", "\n")
+        private_key = " ".join(private_key.split("\n", max_split))
+        private_key = " ".join(private_key.rsplit("\n", max_split))
+        return private_key
+
+    def _initialize_configurations(self):
+        """Initialize the ServiceAccountCreds"""
+        json_creds = json.loads(self.configuration["service_account_credentials"])
+        if (
+            not RUNNING_FTEST
```
timgrein, line 216: It feels a bit odd to have code specific to testing inside a connector. Can we solve this differently?

connectors/sources/google_cloud_storage.py
```diff
@@ -199,8 +185,54 @@ async def _api_call(
...
+        self.user_project_id = self.service_account_credentials.project_id
+
     async def ping(self):
-        """Verify the connection with Google Cloud Storage"""
+        """Verify the configurations and connection with Google Cloud Storage"""
+
+        self._validate_configurations()
+        self._initialize_configurations()
+
         if RUNNING_FTEST:
```
timgrein, line 236: Same as above. Do we really need a check for `RUNNING_FTEST` here?

connectors/sources/google_cloud_storage.py
```diff
@@ -199,8 +185,54 @@ async def _api_call(
...
                 await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry_counter)
 
+    def _validate_configurations(self):
+        """Validates whether user input is empty or not for configuration field.
+
+        Raises:
+            Exception: Configured keys can't be empty.
+        """
+        if self.configuration["service_account_credentials"] == "":
+            raise Exception("service_account_credentials can't be empty.")
```
timgrein, line 195: Could we have a more specific exception like `DataSourceConfigurationError` here, which we'll use in all `_validate_configuration` implementations?

timgrein (commented):
Some comments around test specific code in connector implementations and a suggestion for narrowing an exception type.

connectors/sources/google_cloud_storage.py
jignesh-crest: I checked setting empty field from Kibana and we are getting `""` empty string.
I can't think of a scenario when we might get `None` as a value for `service_account_credentials` for now.

connectors/sources/google_cloud_storage.py
jignesh-crest: It was done to prevent method calls explicitly in e2e tests only. I've removed the shield, either way for e2e, the if-block will not be executed.

connectors/sources/google_cloud_storage.py
jignesh-crest: Added `DataSourceConfigurationError` in `source.py`, and raised the same with custom messages for the validations.
[bcc60b1](https://github.com/elastic/connectors-python/pull/493/commits/bcc60b1b9b30a0cedc14da979193d02437690281)

connectors/sources/google_cloud_storage.py
jignesh-crest: The change was added in this PR #264 by @tarekziade. The GCS E2E takes care of the ping also, but I suppose it was done to reduce the total execution time.

jignesh-crest:
Hey, I've rebased the branch on the latest main. on a side note, This PR needs to be backported to 8.7 as these changes will be required to make the Google Cloud Connector work with Kibana UI. But I'm not able to add or remove any labels, so can someone do that, would be a great help.

cc: @timgrein @tarekziade

artem-shelkovnikov (commented):
I'm not sure about approach taken here.

Ideally, when connector is initialised, it already has to have all fields initialised and ready to work. In this PR, however, values are initialised later with special private method - doesn't seem like a good pattern to me.

I want to discuss it with @tarekziade - ideally you validate config in initialiser and raise errors there, and framework will show errors in UI correctly.

jignesh-crest:
@artem-shelkovnikov I've been validating the configurations at the time of initialization only, I've moved it to ping keeping [this ](https://github.com/elastic/connectors-python/issues/284#issuecomment-1385297780) comment in mind.

artem-shelkovnikov:
I understand, however it feels like this validation should happen in the initialiser OR we should actually introduce a method in the framework `validate_configuration` that each connector has to implement. Both of these changes require framework change that we need to discuss with @tarekziade.

Problem that I see with current approach is that connector becomes more fragile. If it happens so that we don't call `ping` before calling `get_docs`, then connector will break - configuration is populated only in `ping`, not in `get_docs`

connectors/source.py
```diff
@@ -20,6 +20,16 @@
 """
 
 
+class DataSourceConfigurationError(Exception):
+    """Raise when configurations are invalid.
+
+    Args:
+        Exception (DataSourceConfigurationError): Invalid configurations exception.
```
wangch079, line 27: This `Args` comment seems irrelevant.

connectors/sources/google_cloud_storage.py
wangch079: TBH I don't think we should rely on Kibana implementation and should be more defensive here.

wangch079 (commented):
I agree with @artem-shelkovnikov that `_validate_configurations()` and `_initialize_configurations()` shouldn't happens in `ping()`. Any particular reason to move them out of `__init__`?

It also requires to be called in sequence, e.g. it will fail if you call `_initialize_configurations()` without calling `_validate_configurations()`.

connectors/sources/google_cloud_storage.py
jignesh-crest: Sure, I'll add the check for the `None` also.

moxarth-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
```diff
@@ -199,10 +201,50 @@ async def _api_call(
...
+    def _initialize_configurations(self):
+        """Initialize the ServiceAccountCreds"""
+        if self.service_account_credentials is not None:
+            return
+
+        json_credentials = json.loads(self.configuration["service_account_credentials"])
+
+        if (
+            json_credentials.get("private_key")
+            and "\n" not in json_credentials["private_key"]
```
wangch079, line 228: Just curious what's the purpose of `\n` check.

What if a user accidentally adds a linebreak in Kibana?

connectors/sources/google_cloud_storage.py
```diff
@@ -128,6 +113,23 @@ def get_default_configuration(cls):
...
+            Exception: The service account json is empty.
+            Exception: The format of service account json is invalid.
+        """
+        if (
+            self.configuration["service_account_credentials"] == ""
+            or self.configuration["service_account_credentials"] is None
+        ):
+            raise Exception("Google Cloud service account json can't be empty.")
+        try:
+            _ = json.loads(self.configuration["service_account_credentials"])
```
wangch079, line 129:
```suggestion
            json.loads(self.configuration["service_account_credentials"])
```

If you don't need the parsed result, just don't assing it.

connectors/sources/google_cloud_storage.py
```diff
@@ -128,6 +113,23 @@ def get_default_configuration(cls):
...
+        except ValueError:
```
wangch079, line 130: From the [doc](https://docs.python.org/3/library/json.html#json.loads), it raises `JSONDecodeError`:

>If the data being deserialized is not a valid JSON document, a [JSONDecodeError](https://docs.python.org/3/library/json.html#json.JSONDecodeError) will be raised.

connectors/sources/google_cloud_storage.py
jignesh-crest: > What if a user accidentally adds a linebreak in Kibana?

The service account credential generated from the Google Cloud includes `\n` for line breaks [[ref](https://cloud.google.com/iam/docs/keys-create-delete#creating)], which is expected by the package. but, when there will be `" "`(line breaks-incase if it's modified by the user) then only we will have to replace it with '\n'.

connectors/sources/google_cloud_storage.py
jignesh-crest: [JSONDecodeError ](https://docs.python.org/3/library/json.html#json.JSONDecodeError)is the Subclass of [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

Since we are raising the custom error message & to avoid an extra import, we used ValueError to catch decode errors. 

Still Let me know if it needs to be changed to `JSONDecodeError`, I will update it.

connectors/sources/google_cloud_storage.py
wangch079: In `get_pem_format`, it seems you try to first replace `" "` with `"\n"`, then replace `"\n"` with `" "` again?

What's the purpose of `get_pem_format`? If the credential is modified by the user, should we just raise an error?

I see it just tried to escape unicode originally:
https://github.com/elastic/connectors-python/blob/4ec45095b65e1ba2c2dd64496a67ad040d029e03/connectors/sources/google_cloud_storage.py#L78-L83

jignesh-crest: `get_pem_format` adds new line inplace of `" "`. for reference I've added a [test](https://github.com/elastic/connectors-python/pull/493/files#diff-ebc6238e4238c85b506e32488d7fceccd540da7ffbf05065475f0379721545b6R539) to illustrate the same.

> What's the purpose of `get_pem_format`? If the credential is modified by the user, should we just raise an error?

The package itself will raise an SSL error if the key is malformed. But, I added support for the `get_pem_format` keeping this [Google cloud storage add public key](https://cloud.google.com/iam/docs/keys-upload#uploading) in mind, which enables users to add personal Public-Private Key pair and use those to authenticate with the Google Cloud. In this case, user might add Private-Key with `" "` in it instead of `\n`.

praveen-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
```diff
@@ -199,10 +201,50 @@ async def _api_call(
                 )
                 await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry_counter)
 
+    def get_pem_format(self, private_key, max_split=-1):
```
timgrein, line 204: This function looks familiar, can we deduplicate it by moving it to `utils.py`?

connectors/sources/google_cloud_storage.py
```diff
@@ -199,10 +201,50 @@ async def _api_call(
...
+        )
+        self.user_project_id = self.service_account_credentials.project_id
+
     async def ping(self):
         """Verify the connection with Google Cloud Storage"""
+
         if RUNNING_FTEST:
             return
+
+        self._initialize_configurations()
```
timgrein, line 247: Shouldn't we initialize configurations in the constructor and not in `ping`?

connectors/sources/google_cloud_storage.py
```diff
@@ -344,6 +386,7 @@ async def get_docs(self, filtering=None):
         Yields:
             dictionary: Documents from Google Cloud Storage.
         """
+        self._initialize_configurations()
```
timgrein, line 389: Why do we initialize configurations at two places? Shouldn't we do it at one place, i.e. the constructor?

connectors/sources/tests/test_google_cloud_storage.py
```diff
@@ -62,10 +63,30 @@ async def test_empty_configuration():
...
-        _ = GoogleCloudStorageDataSource(configuration=configuration)
+    with pytest.raises(
+        Exception, match="Google Cloud service account json can't be empty."
+    ):
+        await gcs_object.validate_config()
+
+
+@pytest.mark.asyncio
+async def test_invalid_configuration():
+    """Tests that the service account credential is in invalid json format"""
```
timgrein, line 77:
```suggestion
async def test_raise_on_invalid_configuration():
```

timgrein (commented):
Could we make one call for `initialize_configurations` instead of two?

connectors/sources/google_cloud_storage.py
jignesh-crest: To initialize `ServiceAccountCreds` we need to validate the Json first which will be done in validate_config. Thus, Updated the `self._initialize_configurations()` as a getter method, with a check to return if the `ServiceAccountCreds` is already initialized.

moxarth-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
```diff
@@ -199,10 +201,40 @@ async def _api_call(
...
                 await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry_counter)
 
+    def get_service_account_credentials(self):
+        """Initialize and return the ServiceAccountCreds
+
+        Returns:
+            ServiceAccountCreds: An instance of the ServiceAccountCreds.
+        """
+        if self.service_account_credentials is not None:
+            return
```
timgrein, line 211: I think you want to return `self.service_account_credentials` here, right?

connectors/sources/google_cloud_storage.py
```diff
@@ -199,10 +201,40 @@ async def _api_call(
...
+        self.user_project_id = self.service_account_credentials.project_id
+        return self.service_account_credentials
+
     async def ping(self):
         """Verify the connection with Google Cloud Storage"""
+
         if RUNNING_FTEST:
             return
+
+        self.get_service_account_credentials()
```
timgrein, line 237: I had a chat with @praveen-elastic about that. I think we should encapsulate instantiating a session/client in a separate class, which provides methods for talking with the corresponding API. Because here we call a getter and ignore the result and rely on the internal side effect, that `self.service_account_credentials` is now instantiated, which is not obvious.

My suggestion would be something like the following:
(This is only a suggestion, you can choose another name, if you think it's more appropriate)
```
class GoogleCloudStorageClient:
    def __init__()
        self.service_account_credentials = ...
        # you can also store a client instance here        

    def api_call(self, resource):
        # execute api call

class GoogleCloudStorageDataSource(BaseDataSource):
     def get_client(self):
         if self.client is not None:
             return self.client
         self.client = GoogleCloudStorageClient(...)
         
         return self.client 
```

timgrein, line 237: I had a chat with @praveen-elastic about that. I think we should encapsulate instantiating a session/client in a separate class, which provides methods for talking with the corresponding API. Because here we call a getter and ignore the result and rely on the internal side effect, that `self.service_account_credentials` is now instantiated, which is not obvious.

My suggestion would be something like the following:
(This is only a suggestion, you can choose another name, if you think it's more appropriate)
```
class GoogleCloudStorageClient:
    def __init__()
        self.service_account_credentials = ...
        # you can also store a client instance here        

    def api_call(self, resource):
        # execute api call

class GoogleCloudStorageDataSource(BaseDataSource):
     def get_client(self):
         if self.client is not None:
             return self.client
         self.client = GoogleCloudStorageClient(...)
         
         return self.client 
```

timgrein (commented):
Added a suggestion how we can encapsulate a client/session.

connectors/sources/google_cloud_storage.py
jignesh-crest: Updated the modified approach in ca65deb.

connectors/sources/google_cloud_storage.py
```diff
@@ -199,17 +148,114 @@ async def _api_call(
...
+
         try:
             await anext(
-                self._api_call(
+                self.get_storage_client()._api_call(
                     resource="projects",
                     method="serviceAccount",
                     sub_method="get",
-                    projectId=self.user_project_id,
+                    projectId=self.google_storage_client.user_project_id,
```
timgrein, line 258:
```suggestion
                    projectId=self.get_storage_client().user_project_id,
```

Let's remove the assumption of the initialization side effect explicitly and always use the getter

connectors/sources/google_cloud_storage.py
```diff
@@ -223,12 +269,12 @@ async def fetch_buckets(self):
...
             Dictionary: Contains the list of fetched buckets from Google Cloud Storage.
         """
-        async for bucket in self._api_call(
+        async for bucket in self.get_storage_client()._api_call(
             resource="buckets",
             method="list",
             full_response=True,
-            project=self.user_project_id,
-            userProject=self.user_project_id,
+            project=self.google_storage_client.user_project_id,
```
timgrein, line 276: The `user_project_id` seems to be already set inside the client, right? Do we need to pass it explicitly here or can we use the field inside `_api_call`?

connectors/sources/google_cloud_storage.py
```diff
@@ -242,12 +288,12 @@ async def fetch_blobs(self, buckets):
             Dictionary: Contains the list of fetched blobs from Google Cloud Storage.
         """
         for bucket in buckets.get("items", []):
-            async for blob in self._api_call(
+            async for blob in self.google_storage_client._api_call(
```
timgrein, line 291:
```suggestion
            async for blob in self.get_storage_client()._api_call(
```

connectors/sources/google_cloud_storage.py
```diff
@@ -242,12 +288,12 @@ async def fetch_blobs(self, buckets):
...
                 resource="objects",
                 method="list",
                 full_response=True,
                 bucket=bucket["id"],
-                userProject=self.user_project_id,
+                userProject=self.google_storage_client.user_project_id,
```
timgrein, line 296: Same as above, we probably don't need that and use the field inside the client class?

connectors/sources/google_cloud_storage.py
```diff
@@ -266,7 +312,7 @@ def prepare_blob_document(self, blob):
         blob_name = urllib.parse.quote(blob_document["name"], safe="'")
         blob_document[
             "url"
-        ] = f"{CLOUD_STORAGE_BASE_URL}{blob_document['bucket_name']}/{blob_name};tab=live_object?project={self.user_project_id}"
+        ] = f"{CLOUD_STORAGE_BASE_URL}{blob_document['bucket_name']}/{blob_name};tab=live_object?project={self.google_storage_client.user_project_id}"
```
timgrein, line 315:
```suggestion
        ] = f"{CLOUD_STORAGE_BASE_URL}{blob_document['bucket_name']}/{blob_name};tab=live_object?project={self.get_storage_client().user_project_id}"
```

connectors/sources/google_cloud_storage.py
```diff
@@ -314,13 +360,13 @@ async def get_content(self, blob, timestamp=None, doit=None):
         source_file_name = ""
         async with NamedTemporaryFile(mode="wb", delete=False) as async_buffer:
             await anext(
-                self._api_call(
+                self.google_storage_client._api_call(
```
timgrein, line 363:
```suggestion
                self.get_storage_client()._api_call(
```

connectors/sources/tests/test_google_cloud_storage.py
```diff
@@ -32,6 +32,7 @@ def get_mocked_source_object():
         {"service_account_credentials": SERVICE_ACCOUNT_CREDENTIALS, "retry_count": 0}
     )
     mocked_gcs_object = GoogleCloudStorageDataSource(configuration=configuration)
+    mocked_gcs_object.get_storage_client()
```
timgrein, line 35: Do we need that here? This is probably here for the initialization side effect, right?
We may want a test for `get_storage_client`, that we get the same instance back on subsequent calls on the same data source instance?

timgrein (commented):
Left some comments, if `user_project_id` needs to be a parameter or can live inside the client class directly.

moxarth-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
jignesh-crest: Right we can have `user_project_id` in the client class itself, but it is passed as value to many different parameters depending on the API we are trying to hit. To accommodate this in the client class, I would have to have checks on the API and then assign the `user_project_id` to the respective parameter.
To avoid multiple if-else ladder, I considered using it as a parameter to `_api_call()`.
WDYT? Should I keep it the same way?

moxarth-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
jignesh-crest: Hey @timgrein , suggested changes are not scope of this PR. This was raised to parse JSON from the Kibana UI.
Feel free to create a new Github issue where we can continue the discussion on the new architecture.
We should not block this PR to merge for the enhancement suggested. WDYT?

connectors/sources/tests/test_google_cloud_storage.py
```diff
@@ -491,11 +509,20 @@ async def test_api_call_for_attribute_error(catch_stdout, patch_logger):
...
-            userProject=mocked_gcs_object.user_project_id,
+            project=mocked_gcs_object.get_storage_client().user_project_id,
+            userProject=mocked_gcs_object.get_storage_client().user_project_id,
         ):
             print("Method called successfully....")
+
+
+def test_get_storage_client():
+    """Test that the instance returned is always the same for the same datasource class."""
+    mocked_gcs_object = get_mocked_source_object()
```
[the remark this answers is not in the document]
jignesh-crest: Updated the method name.

moxarth-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
timgrein: I see, could you still update `api_call` removing the leading underscore, happy to approve afterwards 👍

timgrein (dismissed):
LGTM

moxarth-elastic:
buildkite test this

connectors/sources/google_cloud_storage.py
```diff
@@ -199,17 +148,114 @@ async def _api_call(
...
+    def __init__(self, configuration):
+        """Set up the connection to the Google Cloud Storage Client.
+
+        Args:
+            configuration (DataSourceConfiguration): Object of DataSourceConfiguration class.
+        """
+        super().__init__(configuration=configuration)
+        self.enable_content_extraction = self.configuration["enable_content_extraction"]
+
+        self.google_storage_client = None
```
artem-shelkovnikov, line 167:
```suggestion
        self._google_storage_client = None
```

I'd make it private to indicate that it should not be used.

connectors/sources/google_cloud_storage.py
```diff
@@ -199,17 +148,114 @@ async def _api_call(
...
+            self.configuration["service_account_credentials"] == ""
+            or self.configuration["service_account_credentials"] is None
+        ):
+            raise Exception("Google Cloud service account json can't be empty.")
+        try:
+            json.loads(self.configuration["service_account_credentials"])
+        except ValueError:
+            raise Exception("Google Cloud service account is not a valid JSON.")
+
+    def get_storage_client(self):
```
artem-shelkovnikov, line 221: Since it's not an async client, you can make it a cached property instead and it will be much cleaner: https://docs.python.org/3/library/functools.html#functools.cached_property

artem-shelkovnikov (approved):

praveen-elastic:
buildkite test this

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/589)|

### Questions ?

PR #495, opened 2023-02-16
Make filtering validation VALID by default

wangch079:
`""` is used as the default validation state if it's not availabe, but `""` is not a valid `FilteringValidationState` value.

This PR uses `FilteringValidationState.VALID` as the default value.

#### Pre-Review Checklist

### DIFF
--- a/.buildkite/run_dockerized_test.sh
+++ b/.buildkite/run_dockerized_test.sh
@@ -1,6 +1,47 @@
 #!/bin/bash
-set -euo pipefail
 
-docker run --rm -v $PWD:/ci -w=/ci \
-    python:3.10 \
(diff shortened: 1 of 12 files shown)


timgrein (approved):
Nit: we could have an explicit test to avoid an unintentional regression, but it's a super small change, so probably fine to have it like this.

tarekziade (approved):
LGTM-- any chance to change/add a test to reflect this default?

wangch079:
Added the test: https://github.com/elastic/connectors-python/pull/495/commits/9bc809ce559964915904eaa8b064cac8575ef1ec

tarekziade (approved):

PR #498, opened 2023-02-16
Validate filtering only when it's in EDITED state

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3901

This PR
1. adds `validate_filtering` method to class `Connector`, which only validates when draft filtering is in `EDITED` state.
2. removes `validate_filtering` function from `connectors.filtering.validation` module.
3. makes `SyncService` use `validate_filtering` method from class `Connector`.

#### Pre-Review Checklist


connectors/byoc.py
```diff
@@ -656,6 +637,29 @@ async def prepare(self, config):
                 f"Could not update service type/configuration for connector {self.id}"
             )
 
+    async def validate_filtering(self, validator):
+        draft_filter = self.filtering.get_draft_filter()
+        if not draft_filter.has_validation_state(FilteringValidationState.EDITED):
```
timgrein, line 642: In Ruby we've a debug log , when we skip validation. Should we add it here, too? _Could_ make it easier to debug.

`logger.debug(f"Filtering of connector {self.id} is in state {draft_filter["validation"]["state"]}, skipping validation.")`

connectors/byoc.py
```diff
@@ -656,6 +637,29 @@ async def prepare(self, config):
...
+            return
+
+        logger.info(
+            f"Filtering of connector {self.id} is in state {FilteringValidationState.EDITED.value}, validating...)"
+        )
+        validation_result = await validator.validate_filtering(draft_filter)
+        logger.info(f"The filtering is validated to be {validation_result.state.value}")
```
timgrein, line 649:
```suggestion
        logger.info(f"Filtering validation result for connector '{self.id}': {validation_result.state.value}")
```

connectors/tests/test_byoc.py
```diff
@@ -787,6 +756,85 @@ async def fetch_by_id(self, doc_id):
...
+    await connector.validate_filtering(validator=validator)
+
+    validator.validate_filtering.assert_awaited()
+    index.update.assert_awaited_with(
+        doc_id=CONNECTOR_ID, doc=expected_validation_update_doc
+    )
+
+
+@pytest.mark.asyncio
+async def test_connector_validate_filtering_valid(patch_logger):
```
timgrein, line 807: I remember we had the discussion around keeping/removing `patch_logger`. Should we keep it for log messages in tests here? AFAIU it was mainly used to silent log messages during test execution.

connectors/tests/test_byoc.py
```diff
@@ -787,6 +756,85 @@ async def fetch_by_id(self, doc_id):
...
+
+    connector = Connector(elastic_index=index, doc_source=DOC_SOURCE)
+    await connector.validate_filtering(validator=validator)
+
+    validator.validate_filtering.assert_not_awaited()
+    index.update.assert_not_awaited()
+
+
+@pytest.mark.asyncio
+async def test_connector_validate_filtering_invalid(patch_logger):
```
timgrein, line 774: Same question as below.

timgrein (approved):
Some small comments/nits, but otherwise LGTM

connectors/tests/test_byoc.py
wangch079: Yes, had a discussion with @artem-shelkovnikov , will have a separate PR to remove all the unnecessary `patch_logger` in tests.

artem-shelkovnikov (approved):

PR #500, opened 2023-02-16
Refactor SyncService class

wangch079:
### Changes made:

1. moves methods `sync`, `_sync_starts`, `_sync_done` and `prepare_docs` to a new class `SyncJobRunner`, which takes a `sync_job` object and execute the job.
2. updates `SyncService` to execute jobs via `SyncJobRunner` instead of calling `connector.sync`.
3. rewrites all the tests for `SyncService`. 

### Why the changes:

1. Currently the sync job is executed via `connector.sync`. With the planned decoupling of job scheduling and execution, a new job execution should be initiated from a `SyncJob` object. That's why I made the change to extract all job execution logic from `Connector` class to a new class `SyncJobRunner`.
2. The scope of `test_sync.py` is too big and the tests go too deep level. I think the unit tests of `SyncService` is to test whether a job is scheduled/executed for various scenarios. It's the responsibility of `test_sync_job_runner.py` to test the job execution.

#### Pre-Review Checklist


connectors/services/sync.py
```diff
@@ -35,13 +37,13 @@ class SyncService(BaseService):
     def __init__(self, config):
         super().__init__(config)
         self.idling = self.service_config["idling"]
-        self.hb = self.service_config["heartbeat"]
+        self.heartbeat_interval = self.service_config["heartbeat"]
```
timgrein, line 40: We could think about renaming the option in the yaml file also. Would be more consistent and also obvious what's the consequence of setting heartbeat to something

connectors/services/sync.py
```diff
@@ -35,13 +37,13 @@ class SyncService(BaseService):
...
         self.concurrent_syncs = self.service_config.get(
             "max_concurrent_syncs", DEFAULT_MAX_CONCURRENT_SYNCS
         )
         self.bulk_options = self.es_config.get("bulk", {})
         self.source_klass_dict = get_source_klass_dict(config)
-        self.connectors = None
+        self.connector_index = None
         self.sync_job_index = None
         self.syncs = None
```
timgrein, line 48: Where is this initialised? I see a `self.syncs.put` call, but no explicit assignment in this file.

timgrein (commented):
One comment about possible future work and a question around initialising `syncs`

connectors/services/sync.py
artem-shelkovnikov: Yes, but probably in a separate PR given that we're backporting this change to 8.7.

Can address in 8.8

connectors/services/sync.py
artem-shelkovnikov: Indeed looks like a problem!

artem-shelkovnikov: It's assigned outside of changes in this PR:

https://github.com/elastic/connectors-python/blob/b6b23580b202e8ee41f0e0fd5d43f345697d47a0/connectors/services/sync.py#L130

artem-shelkovnikov (commented):
I see a lot of tests removed, are we losing some test coverage with this change?

wangch079:
>I see a lot of tests removed, are we losing some test coverage with this change?

Hi @artem-shelkovnikov, No. If you see the PR description, point 5) is `rewrites all the tests for SyncService`. So don't try to compare `before` and `after` for `test_sync.py`.

IMO, the previous tests went too deep(it goes all the way to EsClient, and only mock requests to ES). It tests whether a sync is done successfully or suspended...

But I think that's the scope of `SyncJobRunner` class, and should be unit tested in `test_sync_job_runner.py`. 

To unit test `SyncService`, what I care about is:
1. When a connector is forced to sync, it should `put` `sync_job_runner.execute` to the concurrent tasks pool.
2. When a connector is not configured yet, there shouldn't be any task put into `the concurrent tasks pool.
3. etc...

I used Mocks in `test_sync.py`, e.g. I mocked `Connector`, `ConnectorIndex`, `SyncJobIndex`,...

wangch079:
Hi @elastic/ingestion-team , I had another PR #491 which only contains the new class `SyncJobRunner` without usage. @tarekziade mentioned it's hard to review without the changes to the existing code, so I included changes in #491 in this PR.  Here are some reviewing advice, if it helps:

1. Changes in `byoc.py` are just to remove the methods moved to the `SyncJobRunner` class. Same to `test_byoc.py`.
2. All removed methods in `byoc.py` formed `sync_job_runner.py`, and the test file `test_sync_job_runner.py`.
3. The main change to `sync.py` is to create a job, and call `sync_job_runner.execute` instead of calling `connector.sync`.
4. A whole re-write of `test_sync.py`. Don't try to compare with the old version.

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,175 @@
...
+    The class to run a sync job.
+
+    It takes a sync job, and tries to claim it, and executes it. It also makes sure the sync job is updated
+    appropriately when it errors out, is canceled, completes successfully or the service shuts down.
+
+    :param source_klass: The source class of the connector
+    :param sync_job: The sync job to run
+    :param connector: The connector of the sync job
+    :param elastic_server: The sync orchestrator used to fetch data from 3rd-party source and ingest into Elasticsearch
+    :param bulk_options: The bulk options used for the ingestion
```
timgrein, line 33: I think we're using the following structure at other places for the doc string:

```
Arguments:
    - source_klass: ...
    - sync_job: ...
```

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,175 @@
...
+        self._start_time = None
+        self.running = False
+
+    async def execute(self):
+        if self.running:
+            raise SyncJobRunningError(f"Sync job #{self.job_id} is already running.")
+
+        self.running = True
+        if not await self._claim_job():
+            logger.error(f"Unable to claim job #{self.job_id}")
```
timgrein, line 62: I think we should add the `connector_id` to the log message thinking of a possible future, where we run multiple connectors concurrently. Would make it easier to debug IMO.

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,175 @@
...
+        try:
+            self.data_provider = self.source_klass(self.sync_job.configuration)
+            if not await self.data_provider.changed():
+                logger.debug(
+                    f"No change in {self.sync_job.service_type} data provider, skipping..."
+                )
+                await self._sync_done(sync_status=JobStatus.COMPLETED, result={})
+                return
+
+            logger.debug(f"Validating configuration for {self.data_provider}")
```
timgrein, line 74: I would also add the `connector_id` to that log message for better traceability.

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,175 @@
...
+            await self.data_provider.validate_config()
+
+            logger.debug(
+                f"Syncing '{self.sync_job.service_type}' for connector '{self.connector_id}'"
+            )
+            logger.debug(f"Pinging the {self.source_klass} backend")
```
timgrein, line 80: Same as above

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,175 @@
...
+
+            result = await self.elastic_server.async_bulk(
+                self.sync_job.index_name,
+                self.prepare_docs(),
+                self.sync_job.pipeline,
+                filter_=self.sync_job.filtering,
+                sync_rules_enabled=sync_rules_enabled,
+                options=bulk_options,
+            )
+            sync_error = result.get("fetch_error")
```
timgrein, line 108: The method body is pretty long we may want to split this up into 3-4 private methods for readability. But that's just my personal preference. Feel free to keep it like that!

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,175 @@
...
+            sync_status = JobStatus.COMPLETED if sync_error is None else JobStatus.ERROR
+            await self._sync_done(
+                sync_status=sync_status, result=result, sync_error=sync_error
+            )
+        except asyncio.CancelledError:
+            await self._sync_done(sync_status=JobStatus.SUSPENDED, result={})
+        except Exception as e:
+            await self._sync_done(sync_status=JobStatus.ERROR, result={}, sync_error=e)
+        finally:
+            self._start_time = None
```
timgrein, line 118: Do we need to reset the `_start_time` here? (do we reuse a sync job runner instance)? Just curious.

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,175 @@
...
+        if result is None:
+            result = {}
+        doc_updated = result.get("doc_updated", 0)
+        doc_created = result.get("doc_created", 0)
+        doc_deleted = result.get("doc_deleted", 0)
+        indexed_count = doc_updated + doc_created
+
+        ingestion_stats = {
+            "indexed_document_count": indexed_count,
+            "indexed_document_volume": 0,
```
timgrein, line 132: Should this be zero or do we need to calculate something here?

timgrein (commented):
Some comments around log messages, `_start_time` and `indexed_document_volume`. I can approve it afterwards. Still a pretty big change, but I can totally understand, that it's probably hard to split it further up. Looks definitely way better than before and cool to see monitoring!

connectors/sync_job_runner.py
wangch079: Mmm... I don't think I can split it up further. Basically what it does is
```
# 1. check whether the remote source changed
# 2. validate config
# 3. ping remote source to see whether it's healthy
# 4. validate filtering
# 5. prepare content index
# 6. start ingestion
```

I think it looks long because of the logs. But I don't think it's reasonable to group any of the above to a separate method.

connectors/sync_job_runner.py
wangch079: Not really, I copied from `connector.sync`. I will remove this.

connectors/sync_job_runner.py
wangch079: This will be addressed later: https://github.com/elastic/connectors-python/pull/515

connectors/sync_job_runner.py
wangch079: For this line I copied from here: https://github.com/elastic/connectors-python/pull/552#discussion_r1117323803

I think we can leave it.

connectors/sync_job_runner.py
wangch079: Same as the above reason.

timgrein:
I'll approve as soon as you've your changes done, so we don't need the reapproval.

wangch079:
Hi @timgrein , done the changes in https://github.com/elastic/connectors-python/pull/500/commits/eb7c767b827d0634f634cf53fd2462cba3d870d7.

I leave https://github.com/elastic/connectors-python/pull/500#discussion_r1121800930 and https://github.com/elastic/connectors-python/pull/500#discussion_r1121801325 unchanged.

timgrein (approved):
LGTM, thank you for addressing the suggestions!

connectors/sync_job_runner.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+    async def _claim_job(self):
+        try:
+            await self.sync_job.claim()
+            await self.connector.sync_starts()
+            self._start_time = time.time()
+            return True
+        except Exception as e:
+            logger.critical(e, exc_info=True)
```
tarekziade, line 161: Nit: since we recover from this error here, do we want to issue a `warning` instead?

tarekziade (approved):

connectors/sync_job_runner.py
wangch079: Actually, we don't recover from here. If an exception is caught, `_claim_job` will return `False`, and the caller will raise `JobClaimError`, so `sync_job_runner.execute` is done.

The current implementation does not take race conditions into account. 

Once we decouple the job execution to a new `JobExecutionService`, there can be multiple instances trying to claim a single job, and it's expected that claim can fail. In that case, we will just log a message, and exit without any exception.

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/689)|

### Questions ?

PR #509, opened 2023-02-17
[Generic Database Connector] Added Microsoft SQL Connector 

moxarth-elastic:
This PR includes:

## Checklists

## Licenses

- PyODBC: [MIT License](https://github.com/mkleehammer/pyodbc/blob/master/LICENSE.txt)

### DIFF
--- a/connectors/sources/generic_database.py
+++ b/connectors/sources/generic_database.py
@@ -14,9 +14,27 @@ from connectors.logger import logger
 from connectors.source import BaseDataSource
 from connectors.utils import iso_utc
 
+ALL_TABLES = "*"
+
 DEFAULT_FETCH_SIZE = 50
 DEFAULT_RETRY_COUNT = 3
 DEFAULT_WAIT_MULTIPLIER = 2
+
+
+def configured_tables(tables):
+    def table_filter(table):
+        return table is not None and len(table) > 0
+
+    return (
+        list(
+            filter(
+                lambda table: table_filter(table),
+                map(lambda table: table.strip(), tables.split(",")),
+            )
+        )
+        if isinstance(tables, str)
+        else list(filter(lambda table: table_filter(table), tables))
+    )
 
 
 class GenericBaseDataSource(BaseDataSource):
@@ -39,6 +57,7 @@ class GenericBaseDataSource(BaseDataSource):
         self.host = self.configuration["host"]
         self.port = self.configuration["port"]
         self.database = self.configuration["database"]
+        self.tables = self.configuration["tables"]
         self.is_async = False
         self.engine = None
         self.dialect = ""
@@ -75,8 +94,13 @@ class GenericBaseDataSource(BaseDataSource):
             },
             "database": {
                 "value": "xe",
-                "label": "Databases",
+                "label": "Database",
                 "type": "str",
+            },
+            "tables": {
+                "value": ALL_TABLES,
+                "label": "Comma-separated list of tables",
+                "type": "list",
             },
             "fetch_size": {
                 "value": DEFAULT_FETCH_SIZE,
@@ -96,13 +120,7 @@ class GenericBaseDataSource(BaseDataSource):
         Raises:
             Exception: Configured keys can't be empty
         """
-        connection_fields = [
-            "host",
-            "port",
-            "user",
-            "password",
-            "database",
-        ]
+        connection_fields = ["host", "port", "user", "password", "database", "tables"]
 
         if empty_connection_fields := [
             field for field in connection_fields if self.configuration[field] == ""
@@ -363,23 +381,15 @@ class GenericBaseDataSource(BaseDataSource):
         Yields:
             Dict: Row document to index
         """
-        # Query to get all table names from a database
-        list_of_tables = await anext(
-            self.execute_query(
-                query_name="ALL_TABLE",
-                user=self.user.upper(),
-                database=self.database,
+        tables_to_fetch = await self.get_tables_to_fetch(schema)
+
+        for table in tables_to_fetch:
+            logger.debug(f"Found table: {table} in database: {self.database}.")
+            async for row in self.fetch_documents(
+                table=table,
                 schema=schema,
-            )
-        )
-        if len(list_of_tables) > 0:
-            for [table_name] in list_of_tables:
-                logger.debug(f"Found table: {table_name} in database: {self.database}.")
-                async for row in self.fetch_documents(
-                    table=table_name,
-                    schema=schema,
-                ):
-                    yield row
+            ):
+                yield row
         else:
             if schema:
                 logger.warning(
@@ -387,3 +397,27 @@ class GenericBaseDataSource(BaseDataSource):
                 )
             else:
                 logger.warning(f"Fetched 0 tables for the database: {self.database}")
+
+    async def get_tables_to_fetch(self, schema):
+        tables = configured_tables(self.tables)
+        fetch_all_tables = tables == ALL_TABLES or (
+            isinstance(tables, list) and tables == [ALL_TABLES]
+        )
+
+        tables_to_fetch = (
+            map(
+                lambda table: table[0],
+                await anext(
+                    self.execute_query(
+                        query_name="ALL_TABLE",
+                        user=self.user.upper(),
+                        database=self.database,
+                        schema=schema,
+                    )
+                ),
+            )
+            if fetch_all_tables
+            else tables
+        )
+
+        return tables_to_fetch
--- a/connectors/sources/generic_database.py
+++ b/connectors/sources/generic_database.py
@@ -15,7 +15,7 @@ from connectors.logger import logger
 from connectors.source import BaseDataSource
 from connectors.utils import iso_utc
 
-ALL_TABLES = "*"
+WILDCARD = "*"
 
 DEFAULT_FETCH_SIZE = 50
 DEFAULT_RETRY_COUNT = 3
@@ -23,6 +23,15 @@ DEFAULT_WAIT_MULTIPLIER = 2
 
 
 def configured_tables(tables):
+    """Split a string containing a comma-seperated list of tables by comma and strip the table names.
+
+    Filter out `None` and zero-length values from the tables.
+    If `tables` is a list return the list also without `None` and zero-length values.
+
+    Arguments:
+    - `tables`: string containing a comma-seperated list of tables or a list of tables
+    """
+
     def table_filter(table):
         return table is not None and len(table) > 0
 
@@ -36,6 +45,10 @@ def configured_tables(tables):
         if isinstance(tables, str)
         else list(filter(lambda table: table_filter(table), tables))
     )
+
+
+def is_wildcard(tables):
+    return tables in (WILDCARD, [WILDCARD])
 
 
 class Queries(ABC):
@@ -92,7 +105,7 @@ class GenericBaseDataSource(BaseDataSource):
         self.retry_count = self.configuration["retry_count"]
 
         # Connection related configurations
-        self.user = self.configuration["user"]
+        self.user = self.configuration["username"]
         self.password = self.configuration["password"]
         self.host = self.configuration["host"]
         self.port = self.configuration["port"]
@@ -122,7 +135,7 @@ class GenericBaseDataSource(BaseDataSource):
                 "label": "Port",
                 "type": "int",
             },
-            "user": {
+            "username": {
                 "value": "admin",
                 "label": "Username",
                 "type": "str",
@@ -138,7 +151,7 @@ class GenericBaseDataSource(BaseDataSource):
                 "type": "str",
             },
             "tables": {
-                "value": ALL_TABLES,
+                "value": WILDCARD,
                 "label": "Comma-separated list of tables",
                 "type": "list",
             },
@@ -160,7 +173,14 @@ class GenericBaseDataSource(BaseDataSource):
         Raises:
             Exception: Configured keys can't be empty
         """
-        connection_fields = ["host", "port", "user", "password", "database", "tables"]
+        connection_fields = [
+            "host",
+            "port",
+            "username",
+            "password",
+            "database",
+            "tables",
+        ]
 
         if empty_connection_fields := [
             field for field in connection_fields if self.configuration[field] == ""
@@ -257,7 +277,9 @@ class GenericBaseDataSource(BaseDataSource):
             cursor: Asynchronous cursor
         """
         try:
-            async with self.engine.connect() as connection:
+            if self.engine is None:
+                self._create_engine()
+            async with self.engine.connect() as connection:  # pyright: ignore
                 cursor = await connection.execute(text(query))
                 return cursor
         except Exception as exception:
@@ -282,10 +304,12 @@ class GenericBaseDataSource(BaseDataSource):
             cursor: Synchronous cursor
         """
         try:
+            if self.engine is None:
+                self._create_engine()
             loop = asyncio.get_running_loop()
             if self.connection is None:
                 self.connection = await loop.run_in_executor(
-                    executor=None, func=self.engine.connect
+                    executor=None, func=self.engine.connect  # pyright: ignore
                 )
             cursor = await loop.run_in_executor(
                 executor=None,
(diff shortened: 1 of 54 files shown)


moxarth-elastic:
Hey @tarekziade As we have a dependency of ODBC driver with pyodbc, in E2E we added steps for installing ODBC driver.
So, please run this branch in nightly to check E2E of mssql connector before merging.

connectors/sources/mssql.py
```diff
@@ -0,0 +1,126 @@
...
+        """Setup connection to the Microsoft SQL database-server configured by user
+
+        Args:
+            configuration (DataSourceConfiguration): Object of DataSourceConfiguration class.
+        """
+        super().__init__(configuration=configuration)
+        self.mssql_driver = self.configuration["mssql_driver"]
+        self.secured_connection = self.configuration["secured_connection"]
+        self.connection_string = ""
+        self.queries = QUERIES
```
timgrein, line 52: Why do we store a constant in `self.queries`? Do we reassign `self.queries` anywhere or allow it via the API of this data source?

connectors/sources/tests/test_generic_database.py
```diff
@@ -25,6 +26,18 @@
 ORACLE_CONNECTION_STRING = "oracle+oracledb://admin:changme@127.0.0.1:1521/testdb"
 
 
+class mock_engine:
+    """This Class create mock engine for mssql dialect"""
```
timgrein, line 30: Why does this class live in `test_generic_database.py`, if it's specific for MSSQL?

connectors/sources/tests/test_generic_database.py
```diff
@@ -278,6 +291,42 @@ async def test_get_docs_oracle(patch_logger):
...
+            "_timestamp": 10,
+            "Database": "xe",
+            "Table": 10,
+            "schema": 10,
+        },
+    ]
+
+    # Execute
+    async for i in source.get_docs():
+        actual_response.append(i[0])
```
timgrein, line 324:
```suggestion
    async for doc in source.get_docs():
        actual_response.append(doc[0])
```

connectors/sources/tests/test_mssql.py
```diff
@@ -0,0 +1,39 @@
...
+@pytest.mark.asyncio
+async def test_ping(patch_logger):
+    """Test ping method of MSSQLDataSource class"""
+    # Setup
+    source = create_source(MSSQLDataSource)
+    source._create_engine = Mock()
+    source.execute_query = Mock(return_value=AsyncIterator(["table1", "table2"]))
+
+    # Execute
+    await source.ping()
```
timgrein, line 25: As we're inheriting this from `generic_database.py` and we're mocking the real database calls it would make sense to lift these test cases to `test_generic_database.py` without reimplementing them for every data source inheriting from `generic_database.py`.

timgrein (commented):
Some comments around where we put certain things: Should it live inside a test/class specific to the datasource or can it live inside the generic test/class?

connectors/sources/mssql.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class MSSQLDataSource(GenericBaseDataSource):
+    """Microsoft SQL Server"""
+
+    def __init__(self, configuration):
+        """Setup connection to the Microsoft SQL database-server configured by user
+
+        Args:
+            configuration (DataSourceConfiguration): Object of DataSourceConfiguration class.
```
artem-shelkovnikov, line 46:
```suggestion
            configuration (DataSourceConfiguration): Instance of DataSourceConfiguration class.
```

connectors/sources/mssql.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        super().__init__(configuration=configuration)
+        self.mssql_driver = self.configuration["mssql_driver"]
+        self.secured_connection = self.configuration["secured_connection"]
+        self.connection_string = ""
```
artem-shelkovnikov, line 51: Why make it empty and initialise later?

Ideally, this variable should be initialized immediately or used as a local variable later.

connectors/sources/mssql.py
```diff
@@ -0,0 +1,126 @@
...
+            self.connection_string = URL.create(
+                "mssql+pyodbc",
+                username=self.user,
+                password=self.password,
+                host=self.host,
+                port=self.port,
+                database=self.database,
+                query={"driver": self.mssql_driver, "TrustServerCertificate": "yes"},
+            )
+        self.engine = create_engine(self.connection_string)
```
artem-shelkovnikov, line 105: Why not create an engine in the initializer?

connectors/sources/tests/fixtures/mssql/fixture.py
```diff
@@ -0,0 +1,154 @@
...
+
+def append_to_path(current_path):
+    """set the environment variable"""
+    if DRIVER_PATH not in current_path:
+        os.environ["PATH"] += os.pathsep + DRIVER_PATH
+
+
+# Microsoft ODBC Driver: https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
+def setup():
+    """Install ODBC Driver"""
```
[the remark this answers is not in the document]
artem-shelkovnikov: cc @tarekziade

connectors/sources/mssql.py
moxarth-elastic: `QUERIES` constant is different for each dialect, so when we run the connector data source file will be called first. that's why we have initialized the `self.queries` in each data source file and later used it in the generic_database file.

connectors/sources/tests/fixtures/mssql/fixture.py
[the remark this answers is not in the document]
moxarth-elastic: We already discussed this dependency with @tarekziade as we need to install odbc driver for MS-SQL dialect [refer here](https://github.com/elastic/connectors-python/pull/299#issuecomment-1400585608). We can surely add detailed instructions inside the documentation to aware user about it.

connectors/sources/mssql.py
moxarth-elastic: As we need to pass the connection string to create the engine, we need to call a separate method inside `__init__()` and just to follow best practice not to call any complex method inside `__init__()` we created this method. However, we can move it to `__init__()` if you want. (for this and other dialects as well)

connectors/sources/mssql.py
```diff
@@ -0,0 +1,126 @@
...
+            dictionary: Default configuration
+        """
+        mssql_configuration = super().get_default_configuration().copy()
+        mssql_configuration.update(
+            {
+                "mssql_driver": {
+                    "value": "ODBC Driver 18 for SQL Server",
+                    "label": "Microsoft SQL Driver",
+                    "type": "str",
+                },
```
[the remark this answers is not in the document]
moxarth-elastic: Sure we can update it as  `Microsoft SQL Driver Name` and if we want to add an example it would become something like `Microsoft SQL Driver Name (ODBC Driver 18 for SQL Server)`.

Let us know your thoughts on what should we keep.

connectors/sources/mssql.py
timgrein: If `QUERIES` always has the same structure/keys you could create a `Queries` class providing a clear interface, what types of queries `generic_database.py` expects. By using an abstract class and its capabilities enforcing a specific API it's more obvious what queries you need to provide in an concrete implementation:

`
class Queries(ABC):

  @abstractmethod
  def ping(self):
    pass
    
  @abstractmethod
  def all_tables(self, **kwargs):
    pass
    
  # other details omitted
`
(lives inside `generic_database.py`)

`
class MSSQLQueries(Queries):
  def ping(self):
    return "SELECT 1+1"
    
  def all_tables(self, **kwargs):
     return f"SELECT table_name FROM information_schema.tables WHERE TABLE_SCHEMA = '{schema}'"
     
   # other details omitted
`

(lives inside `mssql.py`)

connectors/sources/mssql.py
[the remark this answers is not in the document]
artem-shelkovnikov: I'm still not sure what you mean. 

What are the possible values for this field? What does it affect? How, as a user, I can understand what to input into this field?

connectors/sources/mssql.py
moxarth-elastic: Done

moxarth-elastic:
hey @timgrein, did you get a chance to review the updated changes?

connectors/sources/mssql.py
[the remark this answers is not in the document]
artem-shelkovnikov: Hey @moxarth-elastic, any updates on this?

connectors/sources/mssql.py
artem-shelkovnikov: That is true - in case you don't want it in `__init__()`, you can make it a lazy property of the class so that it's initialised once it's used.

connectors/sources/generic_database.py
```diff
@@ -36,6 +38,45 @@ def table_filter(table):
     )
 
 
+class Queries(ABC):
```
timgrein, line 41: Nice! Looks way better IMO. Could we move this change and the corresponding changes in the specific data sources to a separate PR? Would make it easier to review and it's more explicit in the git history.

connectors/sources/tests/test_generic_database.py
```diff
@@ -155,8 +90,32 @@ def fetchmany(self, size):
         return []
 
     def fetchall(self):
```
timgrein, line 92:
```suggestion
    def fetch_all(self):
```

connectors/sources/tests/test_generic_database.py
```diff
@@ -155,8 +90,32 @@ def fetchmany(self, size):
...
-        """This method returns object of Return class"""
-        return [(10,)]
+        """This method returns results of query"""
+        self.query = str(self.query)
+        query_object_mssql = MSSQLQueries()
```
timgrein, line 95: should probably not live inside in `test_generic_database.py`. This is specific for the corresponding data source.

connectors/sources/tests/test_generic_database.py
```diff
@@ -155,8 +90,32 @@ def fetchmany(self, size):
...
+        query_object_oracle = OracleQueries()
```
timgrein, line 96: Same as above

connectors/sources/tests/test_generic_database.py
```diff
@@ -384,16 +324,21 @@ def test_configured_tables(tables, expected_tables):
...
+        source.queries.all_tables(schema="schema")
+        == source.execute_query.call_kwargs[0]["query"]
+    )
 
 
 @pytest.mark.asyncio
 async def test_get_tables_to_fetch_configured_tables():
     source = create_source(GenericBaseDataSource)
     tables = ["table_1", "table_2"]
+    source.queries = MSSQLQueries()
```
timgrein, line 341: Maybe don't use `MSSQLQueries`, but a mock or a fake?

timgrein (commented):
I like the new queries class, some minor comments, where to put some test cases.

connectors/sources/tests/test_generic_database.py
moxarth-elastic: You mean `CursorSync` class? If yes, It is common for `test_oracle` and `test_mssql`. So, we put it in test_generic_database.py file.

moxarth-elastic: To reuse the queries I created an object of `MSSQLQueries` class. Otherwise, we need to use hardcoded queries value.

connectors/sources/tests/test_generic_database.py
timgrein: I see. Do we need a full class here or could we use mocking?

connectors/sources/tests/test_generic_database.py
timgrein: I think you should separate the faking of the query results for every datasource and push the corresponding classes down to the concrete tests (`test_mssql.py` and `test_oracle.py`). It's probably also better to just mock the methods inside the concrete datasource tests without creating a dedicated class.

connectors/sources/mssql.py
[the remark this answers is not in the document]
moxarth-elastic: Currently, we have labeled it as `Microsoft SQL Driver Name (ODBC Driver 18 for SQL Server)`. It means the user needs to pass the driver's name. By executing the command `odbcinst -j` user will able to find the path of the `odbcinst.ini` file. By opening `odbcinst.ini` file the user finds the installed driver name in their system. So, the user will find the name of the driver from `odbcinst.ini` file.   

Reference links: 
 - https://stackoverflow.com/questions/53575855/how-to-find-list-of-odbc-drivers-installed-on-linux
 - https://docs.sqlalchemy.org/en/14/dialects/mssql.html#hostname-connections

connectors/sources/tests/test_generic_database.py
moxarth-elastic: As we have multiple functions in `Cursor` class as `keys`, `fetchmany` and `fetchall`. It's good to have full class.

connectors/sources/generic_database.py
moxarth-elastic: Hey @timgrein, I have separated the logic of Queries class in PR #565

connectors/sources/mssql.py
moxarth-elastic: I have added the change in #565.

connectors/sources/mssql.py
```diff
@@ -0,0 +1,147 @@
...
+        else:
+            connection_string = URL.create(
+                "mssql+pyodbc",
+                username=self.user,
+                password=self.password,
+                host=self.host,
+                port=self.port,
+                database=self.database,
+                query={"driver": self.mssql_driver, "TrustServerCertificate": "yes"},
+            )
```
wangch079, line 125:
```suggestion
        if self.secured_connection:
            query={
                "driver": self.mssql_driver,
                "TrustServerCertificate": "no",
                "Encrypt": "Yes",
            }
        else:
            query={"driver": self.mssql_driver, "TrustServerCertificate": "yes"}

        connection_string = URL.create(
            "mssql+pyodbc",
            username=self.user,
            password=self.password,
            host=self.host,
            port=self.port,
            database=self.database,
            query=query,
        )
```

connectors/sources/tests/test_generic_database.py
wangch079: I see too many `if...else...` in the dummy class. Agree to Tim that we should move them to specific test files and remove the `if...else...`, or even better, we can use mock instead of defining dummy class.

wangch079 (commented):
Moving tests from `test_generaic_database.py` to `test_oracle.py` and `test_postgresql.py` can be a separate PR

connectors/sources/mssql.py
moxarth-elastic: Done

connectors/sources/tests/test_generic_database.py
moxarth-elastic: As we have multiple functions it's better to use a dummy class.

connectors/sources/tests/test_mssql.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Returns:
+            connection: Instance of ConnectionSync
+        """
+        return ConnectionSync(MSSQLQueries())
+
+
+@patch("connectors.sources.mssql.create_engine")
+@patch("connectors.sources.mssql.URL.create")
+def test_create_engine(mock_url, mock_fun):
```
timgrein, line 32:
```suggestion
def test_create_engine(mock_create_url, mock_create_engine):
```

I would be a bit more explicit here about what's mocked.

connectors/sources/tests/test_mssql.py
```diff
@@ -0,0 +1,87 @@
...
+    mock_fun.return_value = "Mock engine"
+    mock_url.return_value = MSSQL_CONNECTION_STRING
+
+    # Execute
+    source._create_engine()
+
+    # Assert
+    mock_fun.assert_called_with(MSSQL_CONNECTION_STRING)
+
+    # Test with secured_connection = True
```
timgrein, line 44:
```suggestion
```

connectors/sources/tests/test_mssql.py
moxarth-elastic: Done

connectors/sources/tests/test_mssql.py
moxarth-elastic: Done

timgrein (approved):

timgrein:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/627)|

### Questions ?

PR #510, opened 2023-02-20
Use TODOs consistently throughout the codebase

timgrein:
### Closes https://github.com/elastic/enterprise-search-team/issues/3985

We've agreed on using "TODO" instead of "XXX" throughout the codebase to document todos.

### DIFF
--- a/connectors/sources/tests/fixtures/fixture.py
+++ b/connectors/sources/tests/fixtures/fixture.py
@@ -100,7 +100,7 @@ def main(args=None):
         os.chdir(os.path.join(os.path.dirname(__file__), args.name))
         if args.action == "start_stack":
             os.system("docker compose up -d")
-            # XXX do better
+            # TODO: do better
             time.sleep(30)
         else:
             os.system("docker compose down --volumes")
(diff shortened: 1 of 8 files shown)


connectors/sources/tests/fixtures/fixture.py
```diff
@@ -100,7 +100,7 @@ def main(args=None):
         os.chdir(os.path.join(os.path.dirname(__file__), args.name))
         if args.action == "start_stack":
             os.system("docker compose up -d")
-            # XXX do better
+            # TODO: do better
```
navarone-feekery, line 103: I want to change `do better` but I'm not even sure what it's referencing. I assume the sleep below it but I don't know.

navarone-feekery (approved):
LGTM with a comment

connectors/sources/tests/fixtures/fixture.py
timgrein: Makes sense, this will be addressed in https://github.com/elastic/enterprise-search-team/issues/3853.

PR #513, opened 2023-02-20
Make async_bulk a task

wangch079:
This PR made the following changes to class `connectors.byoei.ElasticServer`:
1. `async_bulk` will create a fetcher task and bulker task, and return, instead of blocking until the job completes.
2. Add a `done` method to check whether the sync completes.
3. Add a `cancel` method to cancel the sync.
4. Add an `ingestion_stats` method to retrieve ingestion stats during job execution.

#### Pre-Review Checklist


connectors/byoei.py
```diff
@@ -91,6 +96,9 @@ def __init__(
         self.chunk_mem_size = chunk_mem_size * 1024 * 1024
         self.max_concurrent_bulks = max_concurrency
         self.bulk_tasks = ConcurrentTasks(max_concurrency=max_concurrency)
+        self.indexed_document_count = 0
+        self.indexed_document_volume = 0
+        self.deleted_document_count = 0
```
timgrein, line 101: Should we also have `deleted_document_volume`?

connectors/byoei.py
```diff
@@ -100,7 +108,7 @@ def _bulk_op(self, doc, operation=OP_INDEX):
             return [{operation: {"_index": index, "_id": doc_id}}, doc["doc"]]
         if operation == OP_UPSERT:
             return [
-                {"update": {"_index": index, "_id": doc_id}},
+                {operation: {"_index": index, "_id": doc_id}},
```
timgrein, line 111: I think we should encapsulate an operation in a class `Operation`. Would make it more readable and would be more consistent with other parts of the codebase, where we've started to encapsulate plain dicts into classes extending `UserDict`.

connectors/byoei.py
artem-shelkovnikov: We don't have it in ruby either, so if we want this then it's a different PR

connectors/byoei.py
artem-shelkovnikov: Agreed, but not in the scope of this PR IMO

connectors/byoei.py
```diff
@@ -127,12 +135,29 @@ async def _batch_bulk(self, operations):
...
         finally:
             self.bulk_time += time.time() - start
 
         return res
 
     async def run(self):
+        try:
+            await self._run()
+        except asyncio.CancelledError:
+            logger.info("Task is canceled, stop Bulker...")
```
artem-shelkovnikov, line 158: Do we need to have a `finally` block here that disposes of used resources/tasks?

connectors/byoei.py
```diff
@@ -214,7 +238,10 @@ def __str__(self):
         )
 
     async def run(self, generator):
-        await self.get_docs(generator)
+        try:
+            await self.get_docs(generator)
+        except asyncio.CancelledError:
+            logger.info("Task is canceled, stop Fetcher...")
```
artem-shelkovnikov, line 244: Do we need to have a `finally` block here that disposes of used resources/tasks?

artem-shelkovnikov (commented):
Looking for feedback from @tarekziade - when we were talking some time ago, there was an idea of passing a callback to the ESServer that could be used to update the stats, similar to this:

```
def update_stats(stats):
    job.update_stats(stats)

EsServer.sync(..., on_bulk_result=update_stats)
```

connectors/byoei.py
```diff
@@ -57,6 +57,11 @@ def get_mb_size(ob):
     return round(get_size(ob) / (1024 * 1024), 2)
 
 
+def get_byte_size(ob):
+    """Returns the size of ob in bytes"""
```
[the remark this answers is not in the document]
wangch079: Yes, copied from above function `get_mb_size`.

wangch079:
>Do we need to have a finally block here that disposes of used resources/tasks?

@artem-shelkovnikov , I don't see any object-level resources to be released in `Fetcher`, there's a `ConcurrentTasks` instance assigned to a local variable `lazy_downloads`. There's a `ConcurrentTasks` instance assigned to `Bulker` instance variable `bulk_tasks`. For both, I don't say any finally in the original code to call `cancel`.

>there was an idea of passing a callback to the ESServer that could be used to update the stats

Had the same discussion with @tarekziade too. My opinion is that we should keep all the `sync_job` operations inside class `SyncJob`. Although I agree that it's more efficient to batch the call. But I think the cost is acceptable as we only try to update job metadata every X seconds (In ruby it's hardcoded to 10 seconds now, it will be eventually configurable).

tarekziade:
I don't think we need to do this. Can you explain why you need to do it?

connectors/byoei.py
```diff
@@ -127,12 +135,29 @@ async def _batch_bulk(self, operations):
...
                             logger.error(f"operation {op} failed, {data['error']}")
                             raise Exception(data["error"]["reason"])
+
+            for op in operations:
+                if OP_INDEX in op or OP_UPSERT in op:
+                    self.indexed_document_count += 1
+                elif OP_DELETE in op:
+                    self.deleted_document_count += 1
+                elif "doc" in op:  # update payload
+                    self.indexed_document_volume += get_byte_size(op["doc"])
```
tarekziade, line 145: let's make this new thing another PR

connectors/byoei.py
```diff
@@ -177,14 +202,13 @@ def __init__(
         queue,
         index,
         existing_ids,
-        filter=None,
+        filter_=None,
```
tarekziade, line 205: another PR for the renaming

connectors/byoei.py
```diff
@@ -409,17 +444,61 @@ async def get_existing_ids(self, index):
             ts = doc["_source"].get(TIMESTAMP_FIELD)
             yield doc_id, ts
 
+    def done(self):
+        if self._fetcher_task is not None and not self._fetcher_task.done():
```
tarekziade, line 448: you can reunite everything in a single op:

```python
return (
  (self._fetcher_task is None and self._bulker_task is None) or    # no tasks
  (self._fetcher_task.done() and self._bulker_task.done())         
) 
```

connectors/byoei.py
```diff
@@ -409,17 +444,61 @@ async def get_existing_ids(self, index):
...
+            return False
+        if self._bulker_task is not None and not self._bulker_task.done():
+            return False
+        return True
+
+    async def cancel(self):
+        if self._fetcher_task is not None:
+            self._fetcher_task.cancel()
+            await self._fetcher_task
```
tarekziade, line 457: need to try catch `CancelledError`

connectors/byoei.py
```diff
@@ -409,17 +444,61 @@ async def get_existing_ids(self, index):
...
+        if self._bulker_task is not None:
+            self._bulker_task.cancel()
+            await self._bulker_task
```
tarekziade, line 460: need to try catch `CancelledError`

connectors/byoei.py
```diff
@@ -409,17 +444,61 @@ async def get_existing_ids(self, index):
...
+    def fetch_error(self):
+        return None if self._fetcher is None else self._fetcher.fetch_error
+
     async def async_bulk(
         self,
         index,
         generator,
         pipeline,
-        filter=None,
+        filter_=None,
```
tarekziade, line 494: do this in another PR

tarekziade:
You can reach the same with fewer changes because the `async_bulk` async function is already a composition of tasks that are running concurrently. You don't need to do a loop with a sleep call.

If the `ingestion_stats` method is started there, along with the `Fetcher` and `Bulker` -- you will have the ability to stop them by using the `cancel` method you are adding `ElasticServer`.

wangch079:
> You can reach the same with fewer changes because the `async_bulk` async function is already a composition of tasks that are running concurrently. 

I think this PR is trying to make the `fetcher` and `bulker` run in the background, instead of blocking/`awaiting` the coroutine `async_bulk`. It also adds the functionality to be able to cancel the `fetcher` and `bulker`.

>You don't need to do a loop with a sleep call.

If we don't make the change to `async_bulk`, I still need to do a loop with a sleep call, but in the backgroud.

> If the `ingestion_stats` method is started there, along with the `Fetcher` and `Bulker` -- you will have the ability to stop them by using the `cancel` method you are adding `ElasticServer`.

`ingestion_stats` is not coroutine or task, it just return the real-time stats of `Fetcher` and `Bulker`, see implementation:

https://github.com/elastic/connectors-python/blob/52abdd686a07f41c81b160d2693da2937621f084/connectors/byoei.py#L462-L484

tarekziade:
> > You can reach the same with fewer changes because the `async_bulk` async function is already a composition of tasks that are running concurrently.
> 
> I think this PR is trying to make the `fetcher` and `bulker` run in the background, instead of blocking/`awaiting` the coroutine `async_bulk`. It also adds the functionality to be able to cancel the `fetcher` and `bulker`.

What prevents you to just do this to add your stats work?

```python
async def async_bulk():

        bulker_task = asyncio.create_task(bulker.run())
        fetcher_task = asyncio.create_task(fetcher.run())
        stats_task =  asyncio.create_task(stats_task.run())    # new concurrent task
        await asyncio.gather(fetcher_task, bulker_task, stats_task)
```

connectors/byoei.py
wangch079: If one of `_fetcher_task` and `_bulker_task` is None, this will fail.

connectors/byoei.py
wangch079: Could you elaborate why?

The `asyncio.CancelledError` is caught in `SyncJobRunner.execute`, where it will try to call `elastic_server.cancel()`.

tarekziade: if you don't catch it here you might not cancel the `bulker` task because it with raise and get off before.

wangch079:
>What prevents you to just do this to add your stats work?

Yes, this will work, as long as stats_task has access to `bulker` (to get the ingestion stats), and the `sync_job` (to update job metadata). IMO, it will make it more complicated. I would like to promote separate concerns. `async_bulk` should only be responsible for fetching data from 3rd-party source and indexing them into the content index. it shouldn't care about the sync job.

We also need to periodically check whether a cancelation is requested in Kibana (i.e. set job status to `canceling`), so that we can cancel the `async_bulk`, and mark the job as `canceled`

connectors/byoei.py
wangch079: Done here: https://github.com/elastic/connectors-python/pull/525

tarekziade:
That also begs the question of concurrent syncs for different backends. We might need to make the class evolve so it holds a list of tasks per index for example:

```python
def cancel(self, index):
    # XXX

async def async_bulk(index):
    self.tasks[index].append(asyncio.create_task(bulker.run()))
    self.tasks[index].append(asyncio.create_task(fetcher.run()))
    self.tasks[index].append(asyncio.create_task(stats_task.run())    # new concurrent task
    await asyncio.gather(**self.tasks[index])
```

connectors/byoei.py
wangch079: Done here: https://github.com/elastic/connectors-python/pull/528

wangch079:
> In that context, what is exactly the `sync job`?

You need to update the sync_job metadata, i.e. calling `sync_job.update_metadata(ingestion_stats)`. I understand you don't want to make this call a separate ES API call, but a recond in bulk API call. In this case, you need to add this into the batch:

```python
[
    {"update": {"_index": ".elastic-connectors-sync_jobs", "_id": sync_job_id}},
    {"doc": "some ingestion data", "doc_as_upsert": True},
]
```

IMO, `async_bulk` shouldn't care about `sync_job_id`.



> if the `ElasticServer` surfaces a `cancel` operation, and the instance of `ElasticServer` is available elsewhere in the framework I don't see this as being an issue.

Is it like this?
```python
async def check_job(sync_job, connector, elastic_server):
    while True:
        sync_job = sync_job.reload()
        if sync_job.status == JobStatus.CANCELING:
            await elastic_server.cancel()
            sync_job.cancel()
            sync_job = sync_job.reload()
            connector.sync_done(sync_job)


def execute():
    ...

    check_job_task = asyncio.create_task(functools.partial(check_job, sync_job, connector, es))
    await es.async_bulk()
    check_job_task.cancel()
    ...
```

As the `SyncJobRunner.execute` is to execute the job, I do think it's more natural to do any job processing in the forground, and let the ingestion happen in the background.

>That also begs the question of concurrent syncs for different backends

It is an issue now as the `ElasticServer` is instantiated in `SyncService` and shared by all the jobs. In this PR, I also move the instantiation into `SyncJobRunner`, so that each job execution has a separate instance of `ElasticServer`.

connectors/byoei.py
wangch079: Do you mean
```python
if self._fetcher_task is not None:
    try:
        self._fetcher_task.cancel()
        await self._fetcher_task
    catch asyncio.CancelledError:
        pass
```

tarekziade:
> > In that context, what is exactly the `sync job`?
> 
>  I understand you don't want to make this call a separate ES API call, but a recond in bulk API call. In this case, you need to add this into the batch:
> 
> ```python
> [
>     {"update": {"_index": ".elastic-connectors-sync_jobs", "_id": sync_job_id}},
>     {"doc": "some ingestion data", "doc_as_upsert": True},
> ]
> ```
> 
> IMO, `async_bulk` shouldn't care about `sync_job_id`.

The separation of concerns is about what is doing what.

Currently, the `ElasticServer` class has a single identified role: collect in a queue some documents, and send them sent to Elasticsearch via batched bulk requests -- it can stay agnostic about the documents it sends.

It is important to do this in a single place in our project because we need to make sure we control in memory in a single place the stream of batch requests we do against Elasticsearch. That is also how we can throttle things and deal with back pressure.

Documents are provided by a single `Fetcher` class instance and converted into bulk operation, and the rest is taken care of by the `Bulker` class. the `Fetcher` class is index specific.

You want to add extra documents that go to another index. It's a new stream of documents, and it's part of a sync.

In practice, it means that instead of having one `Fetcher` instance, you could have two, because fetcher works on a provided generator (`get_docs`)

> 
> > if the `ElasticServer` surfaces a `cancel` operation, and the instance of `ElasticServer` is available elsewhere in the framework I don't see this as being an issue.
> 
> Is it like this?
> 
> ```python
> async def check_job(sync_job, connector, elastic_server):
>     while True:
>         sync_job = sync_job.reload()
>         if sync_job.status == JobStatus.CANCELING:
>             await elastic_server.cancel()
>             sync_job.cancel()
>             sync_job = sync_job.reload()
>             connector.sync_done(sync_job)
> 
> 
> def execute():
>     ...
> 
>     check_job_task = asyncio.create_task(functools.partial(check_job, sync_job, connector, es))
>     await es.async_bulk()
>     check_job_task.cancel()
>     ...
> ```
> 
> As the `SyncJobRunner.execute` is to execute the job, I do think it's more natural to do any job processing in the foreground, and let the ingestion happen in the background.

There's no background or foreground with async functions, they are all tasks that can run in parallel, and we add sync points. I think the gist of the discussion is about how we group the tasks that are running when we do a sync.

Right now the orchestrator is `async_bulk` because it's iterating on the connectors docs using an iterator, and that is how we deal with back pressure and control how we call the backend. If we pile up too many calls to Elasticsearch we can slow down and stop calling the backend until it's flushed.

It cal be organized into multiple classes for sure. What I am interested into is making sure we keep a single pipeline of data so we can control its memory and network behaviors

wangch079:
> It is important to do this in a single place in our project because we need to make sure we control in memory in a single place the stream of batch requests we do against Elasticsearch. That is also how we can throttle things and deal with back pressure.

This makes a lot of sense, and I think that's why we have a single `ElasticServer` instance instantiated in `SyncService`, and used by all the sync job executions. I tried to instantiate `ElasticServer` in each `SyncJobRunner` instance, which I think it's wrong, as the `MemQueue` only bounds the memory used of one particular sync job, but it can still pile up if we have multiple jobs executing (can be limited by `ConcurrentTasks`, but I think there are better approach).

The current implementation in `main` is the same. Each `async_bulk` call will instantiate a new `MemQueue` instance.

> Documents are provided by a single `Fetcher` class instance and converted into bulk operation, and the rest is taken care of by the `Bulker` class. the `Fetcher` class is index specific.
> 
> You want to add extra documents that go to another index. It's a new stream of documents, and it's part of a sync.
> 
> In practice, it means that instead of having one `Fetcher` instance, you could have two, because fetcher works on a provided generator (`get_docs`)

Very insightful. What I'm thinking now is, we shouldn't bundle `Fetcher`, `Bulker` and `MemQueue` together inside `ElasticServer`. What do you think of this proposal:
1. There should be only one `MemQueue` instance, one `Bulker` instance and 0 - N `Fetcher` instances per application
2. There should be a new long-living `BulkerService` which infinitely polls data from a `MemQueue` instance, and bulk to ES.
3. The `MemQueue` instance is instantiated when the application starts, and passes to `BulkerService.__init__`, and `JobExecutionService.__init__` (a new service will poll pending jobs from `.elastic-connectors-sync-jobs` and execute them)
4. When `SyncJobRunner` executes a job, it creates a `Fetch` instance, which fetches data from the 3rd-party source and pushes it to the `MemQueue`. We consider the job to be done once `Fetch` is done (There's no need to send `END_DOCS` or `FETCH_ERROR`).

Pros:
1. There's the real decoupling of Fetcher**S** and Bulker, and we can control the memory usage at the application level.
2. It can improve efficiency as the Bulker is index-agnostic, and it can bulk docs for multiple indices.

Cons:
1. It will be difficult to get real-time ingestion stats, if not impossible. I tried to update the ingestion stats (e.g. `indexed_document_count`) to the sync_job from the `bulker`, so that `indexed_document_count = 100` means 100 docs are indexed into ES, not 100 are fetched from the 3rd-party. With `Bulker` running as a separate service and it bulks docs for multiple indices, I don't think this is easy ( or in an elegant way) to get the real-time data for a particular index
2. It keeps the `Bulk` out of the scope of each job execution. What if some operations fail?

Even with this proposal, I'm hesitant to bulk the ES API call to update sync job metadata. Besides the reasons I mentioned previously, 
1. Any ES API calls to system indices (i.e. `.elastic-connectors` and `.elastic-connectors-sync-jobs`, not Elasticsearch system indices) should be real-time, we can't afford the request waiting for other docs to be bulked.
2. Any ES API calls to system indices should have concurrency control (in 8.8), which `Bulker` neither have nor need.


I would like to hear your opinion. I think this also deserves a topic in architecture bi-weekly, but I see it last happened nearly 4 months ago.

tarekziade:
> This makes a lot of sense, and I think that's why we have a single `ElasticServer` instance instantiated in `SyncService`, and used by all the sync job executions. I tried to instantiate `ElasticServer` in each `SyncJobRunner` instance, which I think it's wrong, as the `MemQueue` only bounds the memory used of one particular sync job, but it can still pile up if we have multiple jobs executing (can be limited by `ConcurrentTasks`, but I think there are better approach).
> 
> The current implementation in `main` is the same. Each `async_bulk` call will instantiate a new `MemQueue` instance.

This current model is fine -- we can have a dedicated memory space per sync. When doing the sizing work, we can use the "concurrent syncs" parameters to determine how many concurrent syncs we allow. Right now it's `1` in cloud.

I don't expect this value to be big. We can probably bump it to 4 or 5 on-prem, and 2 or 3 on the bigger cloud instances.
I don't anticipate we will have a lot of concurrent syncs. 


> Very insightful. What I'm thinking now is, we shouldn't bundle `Fetcher`, `Bulker` and `MemQueue` together inside `ElasticServer`. What do you think of this proposal:
> 
>     1. There should be only one `MemQueue` instance, one `Bulker` instance and 0 - N `Fetcher` instances per application
> 
>     2. There should be a new long-living `BulkerService` which infinitely polls data from a `MemQueue` instance, and bulk to ES.
> 
>     3. The `MemQueue` instance is instantiated when the application starts, and passes to `BulkerService.__init__`, and `JobExecutionService.__init__` (a new service will poll pending jobs from `.elastic-connectors-sync-jobs` and execute them)
> 
>     4. When `SyncJobRunner` executes a job, it creates a `Fetch` instance, which fetches data from the 3rd-party source and pushes it to the `MemQueue`. We consider the job to be done once `Fetch` is done (There's no need to send `END_DOCS` or `FETCH_ERROR`).


why not, yeah. a good first step would be to allow several Fetchers per sync.

> 
> Pros:
> 
>     1. There's the real decoupling of Fetcher**S** and Bulker, and we can control the memory usage at the application level.
> 
>     2. It can improve efficiency as the Bulker is index-agnostic, and it can bulk docs for multiple indices.
> 
> 
> Cons:
> 
>     1. It will be difficult to get real-time ingestion stats, if not impossible. I tried to update the ingestion stats (e.g. `indexed_document_count`) to the sync_job from the `bulker`, so that `indexed_document_count = 100` means 100 docs are indexed into ES, not 100 are fetched from the 3rd-party. With `Bulker` running as a separate service and it bulks docs for multiple indices, I don't think this is easy ( or in an elegant way) to get the real-time data for a particular index

a way to implement this is to have the `Bulker` class implement events per operation, consumers can register to. 
So you would have a task that says "please ping me on each X operation the bulker is doing"

> 
>     2. It keeps the `Bulk` out of the scope of each job execution. What if some operations fail?

right now it's optimistic fails. it gets reported and "until next sync"

> 
> 
> Even with this proposal, I'm hesitant to bulk the ES API call to update sync job metadata. Besides the reasons I mentioned previously,
> 
>     1. Any ES API calls to system indices (i.e. `.elastic-connectors` and `.elastic-connectors-sync-jobs`, not Elasticsearch system indices) should be real-time, we can't afford the request waiting for other docs to be bulked.
> 
>     2. Any ES API calls to system indices should have concurrency control (in 8.8), which `Bulker` neither have nor need.
> 
> 
> I would like to hear your opinion. I think this also deserves a topic in architecture bi-weekly, but I see it last happened nearly 4 months ago.

sure we should, it's very important to make sure we align there. in any case, we should always try not to refactor unless it's mandatory for a change we want to do. 

For this patch here, back to the initial requirement, I think we should aim at the smallest change possible with what exists and not aim at a refactoring that changes the streaming concept we have in place, even if it's unperfect and can be improved

wangch079:
> a way to implement this is to have the `Bulker` class implement events per operation, consumers can register to.
> So you would have a task that says "please ping me on each X operation the bulker is doing"

Besides controlling memory usage, another reason that we introduced a queue, is to decouple `Fetcher` and `Bulker`. If I understand your suggestion correctly, it means when an `operation` is fetched from the queue by the `Bulker`, there should be a callback inside the `operation`, so the `Bulker` can update when the real bulk API call happens. Is this callback trying to update the `Fetcher`? What if when the bulk API call happens, the fetch is completed and the `fetcher` object is reclaimed?

With the decoupling of `Fetcher` and `Bulker`, I think a job should only cares about the `Fetcher`, it's considered done when the fetch is done, so I think it's fine that we populate the ingestions stats from `Fetcher`. If you agree to this, I will make the change to #528, and have those stats implemented in `Fetcher`.

>right now it's optimistic fails. it gets reported and "until next sync"

Currently `Bulker` will raise an exception if any operation fails:
https://github.com/elastic/connectors-python/blob/05c4b7a02f168abe281abcdfe02aeada98bfcbed/connectors/byoei.py#L124C13-L129

I think we should make it more error tolerable, especially when it runs separately and bulk for multiple indices.

>I think we should aim at the smallest change possible with what exists and not aim at a refactoring that changes the streaming concept we have in place

I will first try to get #500 merged (not backport to 8.7), then we can discuss how to do job cancelation and reporting metadata.

As you suggested, I have drafted 2 PRs which are extracted from #500:
1. #535 
2. #536

tarekziade:
> > a way to implement this is to have the `Bulker` class implement events per operation, consumers can register to.
> > So you would have a task that says "please ping me on each X operation the bulker is doing"
> 
> Besides controlling memory usage, another reason that we introduced a queue, is to decouple `Fetcher` and `Bulker`. If I understand your suggestion correctly, it means when an `operation` is fetched from the queue by the `Bulker`, there should be a callback inside the `operation`, so the `Bulker` can update when the real bulk API call happens. Is this callback trying to update the `Fetcher`? What if when the bulk API call happens, the fetch is completed and the `fetcher` object is reclaimed?
> 
> With the decoupling of `Fetcher` and `Bulker`, I think a job should only cares about the `Fetcher`, it's considered done when the fetch is done, so I think it's fine that we populate the ingestions stats from `Fetcher`. If you agree to this, I will make the change to #528, and have those stats implemented in `Fetcher`.

I am not sure to follow what you mean here. The stats you want to report are about the number of documents that are really sent to Elasticsearch IIRC. So the only place where we know what operations went through and where some did not, is in the Bulker class, which can report on that (actively, or passively)

I think we need to pause this refactoring until we agree on how to implement the change. The key to do this progressively is to ask ourselves: with the current code, if you were to add the `ingestion_stats` somewhere, where would it be? (and nothing else).

tarekziade:
This is where the `bulk` response gets back from Elasticsearch

https://github.com/elastic/connectors-python/blob/main/connectors/byoei.py#L124

The response has an `items` array: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-response-body -- where you know what was created, deleted, indexed, updated in which index.

The `Bulker` can keep some counters, per indices of each operation, and you can ask for it on a regular basis. 
For adding this, I don't think we need any refactoring

wangch079:
>So the only place where we know what operations went through and where some did not, is in the Bulker class, which can report on that (actively, or passively)

100% agree. But we can also get the stats from `Fetcher`, if it's acceptable that the number is not 100% accurate. e.g. 100 docs fetched but not all of them indexed, or some index operations failed.

Another reason to get stats from `Fetcher` is, imagine we have fully decoupled Fetcher and Bulker, i.e. there's a single Bulker running infinitely, when a sync job runs, the job should know nothing about the bulker. If Fetcher reports that all docs are fetched, then we mark the job as `completed`.

> The `Bulker` can keep some counters, per indices of each operation, and you can ask for it on a regular basis.

That's what I did in this PR, except for the `per indices` part, as now we run a new `async_bulk` for each job.

I didn't intend to change this flow:
https://github.com/elastic/connectors-python/blob/8fa1bec3106dd354a01efba62390fdf6eb3ec2c8/connectors/byoei.py#L16-L19

Currently, this is how we use `async_bulk`:
https://github.com/elastic/connectors-python/blob/8fa1bec3106dd354a01efba62390fdf6eb3ec2c8/connectors/byoc.py#L721

the `sync_bulk` creates a `fetcher` and a `bulker` object, runs them, and only returns after the ingestion is completed. Before it returns, the above code (line 721) blocks.

What I tried to do is make `async_bulk` create `fetcher` and `bulker` object, assign to instance variables, and wrap the coroutines in tasks (via `create_task`), so that it can return without waiting for the ingestion to complete. Then we can do something else in the `SyncJobRunner`, e.g. periodically report back metadata, and check connector/job validity.

tarekziade:
> > So the only place where we know what operations went through and where some did not, is in the Bulker class, which can report on that (actively, or passively)
> 
> 100% agree. But we can also get the stats from `Fetcher`, if it's acceptable that the number is not 100% accurate. e.g. 100 docs fetched but not all of them indexed, or some index operations failed.

Why would we send back inaccurate values?


> What I tried to do is make `async_bulk` create `fetcher` and `bulker` object, assign to instance variables, and wrap the coroutines in tasks (via `create_task`), so that it can return without waiting for the ingestion to complete. Then we can do something else in the `SyncJobRunner`, e.g. periodically report back metadata, and check connector/job validity.

This is the part that I don't understand. Why do you have to return from `elastic_server.async_bulk` to a periodical task?
Let's consider this very simple change:


```python
async def on_bulk_request(result):
    # do you periodical work here, specific to this sync
         
result = await elastic_server.async_bulk(
                job.index_name,
                self.prepare_docs(data_provider, job.pipeline, job.filtering),
                job.pipeline,
                filter_=job.filtering,
                sync_rules_enabled=sync_rules_enabled,
                options=bulk_options,
                on_bulk_request=on_bulk_request
            )
```

wangch079:
>Let's consider this very simple change

This can work for [Python connector reports back metadata during job execution](https://github.com/elastic/enterprise-search-team/issues/3194). 

> Why do you have to return from `elastic_server.async_bulk` to a periodical task?

For another issue [Python connector cancels a job during job execution](https://github.com/elastic/enterprise-search-team/issues/3160), I still need to periodically check whether the job is canceled (i.e. in status `canceling`), along with other edge cases, e.g. job/connector is deleted. For all those scenarios, I need to 1) mark the job and connector accordingly, e.g. `canceled` or `error` 2) stop the `async_bulk`.

That's why I think this works the best and simplest for both issues:
```python
await self.elastic_server.async_bulk(...

    self.sync_job.update_metadata(self.elastic_server.ingestion_stats())
    if self.sync_job.status = JobStatus.CANCELING:
        await self.elastic_server.cancel()
        self.sync_job.cancel()
        self.connector.sync_done(self.sync_job)
    asyncio.sleep(JOB_CHECK_INTERVAL)
```

tarekziade:
We met yesterday with Chenhui and aligned on the next steps to finish this PR. He will update this PR as soon as he gets back to it.

connectors/byoei.py
```diff
@@ -499,6 +569,8 @@ async def async_bulk(
         - sync_rules_enabled: if enabled, applies rules -- default: `False`
         - options: dict of options (from `elasticsearch.bulk` in the config file)
         """
+        if self._fetcher_task is not None or self._bulker_task is not None:
+            raise AsyncBulkRunningError("Async bulk task has already started.")
```
artem-shelkovnikov, line 573: It feels a bit weird - is it possible to move initialisation of `fetcher` and `bulker` to the initialiser?

The pattern of "This class's method X can be called only once but its signature shows that it can be used for different purposes" is super confusing to me.

E.g. `async_bulk` takes index name as a parameter, meaning that single ElasticServer instance can serve ingestion into multiple indices. However, if you try to do so, the error will be raised - it's counter-intuitive and really confusing to read IMO.

connectors/sync_job_runner.py
```diff
@@ -99,39 +103,42 @@ async def execute(self):
...
                 sync_rules_enabled=sync_rules_enabled,
                 options=bulk_options,
             )
-            sync_error = result.get("fetch_error")
-            sync_status = JobStatus.COMPLETED if sync_error is None else JobStatus.ERROR
-            await self._sync_done(
-                sync_status=sync_status, result=result, sync_error=sync_error
+
+            while not self.elastic_server.done():
+                await asyncio.sleep(JOB_CHECK_INTERVAL)
```
artem-shelkovnikov, line 116: Do we want to update job stats here?

connectors/sync_job_runner.py
```diff
@@ -99,39 +103,42 @@ async def execute(self):
             bulk_options = self.bulk_options.copy()
             self.data_provider.tweak_bulk_options(bulk_options)
 
-            result = await self.elastic_server.async_bulk(
+            await self.elastic_server.async_bulk(
```
artem-shelkovnikov, line 106: I'd probably rename method to `start_async_bulk` to indicate that it starts the process but does not finish.

artem-shelkovnikov:
> We met yesterday with Chenhui and aligned on the next steps to finish this PR. He will update this PR as soon as he gets back to it.

What was the conclusion of the meeting? It's hard to review the PR without knowing what was discussed and what conclusion was reached.

connectors/sync_job_runner.py
wangch079: It will be done in PR #515

connectors/byoei.py
wangch079: >E.g. `async_bulk` takes index name as a parameter, meaning that single ElasticServer instance can serve ingestion into multiple indices.

I believe it was designed like this in the first place, and that's why there's only one instance of `ElasticServer` instantiated in `JobSchedulingService` and shared by all sync jobs:
https://github.com/elastic/connectors-python/blob/3cfabc3f0b366da7ab900042c752525f8b3ad65f/connectors/services/job_scheduling.py#L141

But actually when you call `async_bulk`, it creates a new `Fetcher` object, a new `Bulker` object and a new `MemQueue`:
https://github.com/elastic/connectors-python/blob/3cfabc3f0b366da7ab900042c752525f8b3ad65f/connectors/byoei.py#L473
https://github.com/elastic/connectors-python/blob/3cfabc3f0b366da7ab900042c752525f8b3ad65f/connectors/byoei.py#L485
https://github.com/elastic/connectors-python/blob/3cfabc3f0b366da7ab900042c752525f8b3ad65f/connectors/byoei.py#L464 

The purpose of this PR is to move this flow to the background
https://github.com/elastic/connectors-python/blob/3cfabc3f0b366da7ab900042c752525f8b3ad65f/connectors/byoei.py#L16-L19

So that `async_bulk` will return immediately after `bulker` and `fetcher` tasks are created, without waiting for it to complete. 

Therefore I want to keep the change as small as possible. I moved the instantiation of `ElasticServer` to each `SyncJobRunner` because I need the object to hold `bulker_task` and `fetcher_task`, so it won't be able to be shared by multiple sync jobs.

Ideally, I would like to make it work like what I described here: https://github.com/elastic/connectors-python/pull/528#discussion_r1143601281.

artem-shelkovnikov (approved):

DianaJourdan:
@wangch079 @tarekziade anything more needed on this PR or can it be merged?

wangch079:
Hi @tarekziade , as we discussed before, I run `perf8` test on both `main` branch and `chenhui/make-async-bulk-task` branch. And here is the report:

Report on `main` branch: https://upload.elastic.co/d/f02b2d5b518c9d08f6d8a9abc948642635fae2f69ac0b97a523809b931e16b94
Token: 70186ef00cc1363d

Report on this branch: https://upload.elastic.co/d/7fe5beeeaef4371c84c89eb2d6fd541e63f4d993a71d76b8c09a7bc708f833cb
Token: 6eb4225428cc5673

I don't see much difference, but I need your expert advice. It seems I couldn't open `Memory Flamegraph` tab.

When I test on this branch, I see these logs when I hit `command + .`:
```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x11e445390>
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x11e3c30a0>, 5085039.421409315)]']
connector: <aiohttp.connector.TCPConnector object at 0x11e445360>
```

Will investigate tomorrow.

wangch079:
> When I test on this branch, I see these logs when I hit `command + .`:
> 
> ```
> Unclosed client session
> client_session: <aiohttp.client.ClientSession object at 0x11e445390>
> Unclosed connector
> connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x11e3c30a0>, 5085039.421409315)]']
> connector: <aiohttp.connector.TCPConnector object at 0x11e445360>
> ```
> 
> Will investigate tomorrow.

Forgot to close `elastic_server` in the `finally` block: https://github.com/elastic/connectors-python/pull/513/commits/7ebbed66dcd57d1c6157a89f21c7a9ff7ffe6b9d

artem-shelkovnikov:
@wangch079 how do you make the changes so that my approval is not invalidated? :D

artem-shelkovnikov (dismissed):

artem-shelkovnikov (dismissed):

artem-shelkovnikov (approved):

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/703)|

### Questions ?

PR #514, opened 2023-02-20
Temporary solve the problem with cancellation

artem-shelkovnikov:
## Attempting to quickly fix a problem reported in https://elastic.slack.com/archives/C01795T48LQ/p1676865361259049

This solution might be temporary as I ran out of ideas how to gracefully fix the issue.

Problem:

Imagine, that there is a problem that makes one of services terminate with an exception OR just a mistake that makes one of services return.

The way we run multiple services before this PR is:

```python
done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
```
Then we collect all tasks that are `pending` and cancel them.

What it means in practice, that if we CTRL+C or terminate the task, then we try to stop all our running services. We have 2 now - SyncSerivice and JobCleanupService. Once one of them stops gracefully, `asyncio.wait` above will return the following:

* `done` < ---- service that stopped first
* `pending` < ---- service that haven't stopped yet.

SyncService always takes more time than JobCleanupService to stop, meaning that it will always be in `pending` and will be attempted to be cancelled.

The fun part is, that SyncService is already attempting to stop and handling a Cancellation signal. It _seems_ to me, that when it's being cancelled second time, graceful cancellation just breaks and if there is any Sync job running, then connector is left in a broken state - job is marked as Suspended, but connector is still running and is not marked as if it has a suspended job.

The way this fix works is - if something goes wrong and one of services raises unhandled exception, then it's still possible that this problem happens, but chances of this are really really low. Thus, I removed error handling logic from BaseService (it also does not seem to make sense) and changed `asyncio.wait` to return on first error - this way graceful shutdown when CTRL+C or SIGTERM will not cause issues.

We will still need to address this problem in future, however.

#### Pre-Review Checklist


connectors/services/base.py
```diff
@@ -39,9 +39,6 @@ async def run(self):
         self.running = True
         try:
             await self._run()
-        except Exception as e:
```
wangch079, line 42: I think this still makes sense?

As a derived class of `BaseService`, you can implement `_run` without any `try...except...`.

wangch079 (approved):
LGTM except for one small comment.

connectors/services/base.py
artem-shelkovnikov: I think it's okay?

I don't see why BaseService should not re-raise the error if it happens in `self._run`

wangch079: Got it. Thanks for the explanation

artem-shelkovnikov:
I've added a test to check this scenario - that we do not attempt to cancel a stopping service once services gracefully stop.

artem-shelkovnikov:
⚠️ I also had to remove `test_spurious_continue` and `test_spurious` tests as they don't seem to make much sense - the way errors are handled currently, sync service just exists the execution. So if we swallow the error with `raise_if_spurious` then service just exists with no error code while in reality it should continue running.

wangch079:
> ⚠️ I also had to remove `test_spurious_continue` and `test_spurious` tests as they don't seem to make much sense - the way errors are handled currently, sync service just exists the execution.

I think we should make it consistent on what kind of errors we should raise. 

E.g. Errors from job execution should be marked as job errors, but some (e.g. Job claim error) should be raised. Those should be out of the scope of `SyncSerivce` because they will be decoupled to `JobExecutionSerivce`.

Specific to `SyncService`, Should we raise
1. if `service_type` is not configured for a custom connector?
2. if `service_type` of a customer connector is not configured in `source`
3. if we couldn't find the class from `fqn`

Point 3) currently raises errors while the rest don't:
https://github.com/elastic/connectors-python/blob/bf3760a1ebb46ef3e5030b327d70cc8d31f7882d/connectors/services/sync.py#L78-L82

I don't think it should use `raise_if_spurious` here, as it's used outside.

Same here, we raise errors when the source class is not configured for a native connector:

https://github.com/elastic/connectors-python/blob/bf3760a1ebb46ef3e5030b327d70cc8d31f7882d/connectors/services/sync.py#L95-L98

Should we make it consistent to raise errors for all configuration errors? Or should we log and skip it? As there are other properly running connectors.

artem-shelkovnikov:
@wangch079 I agree, but I'm not planning to address these changes in this PR - I know that you have large refactoring PRs ongoing, so I think this PR can stay the way it is and after the refactoring is done we address the problem with inconsistent error handling later.

What do you think?

wangch079:
Agree the above concern should be addressed separately.

The changes look good to me. I'm ok with removing the tests as I'm also rewriting `test_sync.py` in #500 . But @tarekziade may have different opinions. So I leave it to @tarekziade

tarekziade:
> Agree the above concern should be addressed separately.
> 
> The changes look good to me. I'm ok with removing the tests as I'm also rewriting `test_sync.py` in #500 . But @tarekziade may have different opinions. So I leave it to @tarekziade

well, it's all about not losing coverage for some exits scenarios

simple test: add a raise inside `get_docs` in a connector and make sure it quits trying to sync after a few attempts.
If you don't have that expected behavior, then I would expect that test to be added.
If it behaves as expected, I would expect that we have a similar path into our tests

artem-shelkovnikov:
> well, it's all about not losing coverage for some exits scenarios
> 
> simple test: add a raise inside get_docs in a connector and make sure it quits trying to sync after a few attempts.
> If you don't have that expected behavior, then I would expect that test to be added.
> If it behaves as expected, I would expect that we have a similar path into our tests

In the refactoring we've lost ability to restart the service if something went wrong. As far as I could understand we've _had_ this logic some time ago - we attempted to start service several times if errors happen, and then fail if too many errors happen in short period of time. It's not the case any more - if something happens within `_run` function and error is raised, service just silently quits.

In practice it means that if SyncService has exception, it will crash, but JobCleanupService will keep on running and the overall connector service will be degraded. It means that on cloud it will restart only if both services crash.

This patch attempts to fix it non-invasively before doing a better fix - after this patch if one service crashes, then both services stop and service will attempt to restart. It's far from perfect, but better than without patch - it can cause customers' deployments become unhealthy without any indication of it.

The tests were removed because they don't test anything _now_ - before they attempted to test the logic described above (attempt to re-run service for several times before exiting).

Once we provide a solid solution, the tests will go back.

tarekziade (approved):

tarekziade (approved):

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/655)|

### Questions ?

PR #515, opened 2023-02-20
Report ingestion stats during job execution

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3194

This PR makes the change to report the following ingestion stats during job execution:
1. indexed_document_count
2. indexed_document_volume
3. deleted_document_count

#### Pre-Review Checklist


connectors/sync_job_runner.py
```diff
@@ -90,7 +92,9 @@ async def execute(self):
                 options=bulk_options,
             )
 
+            self._reporting_cycle_start_time = time.time()
             while not self.elastic_server.done():
```
artem-shelkovnikov, line 96: I'm not 100% sure how it'll work - I think previous call to elastic_server is blocking?

wangch079: PR #513 made changes to `async_bulk`, although it's `awaited` and blocking, it will return once it has scheduled (i.e. `create_task`) the fetcher task and bulker task.

`self.elastic_server.done()` will return `True` if both the fetcher task and bulker task are done.

connectors/sync_job_runner.py
```diff
@@ -182,3 +186,22 @@ async def prepare_docs(self):
             doc["_reduce_whitespace"] = self.sync_job.pipeline["reduce_whitespace"]
             doc["_run_ml_inference"] = self.sync_job.pipeline["run_ml_inference"]
             yield doc, lazy_download
+
+    async def check_job(self):
```
artem-shelkovnikov, line 190: Method name/signature are confusing to me. `check_X` implies that it should either return something or raise an exception, but instead it has a side-effect of writing ingestion stats to DB.

Would breaking it into two methods like  `should_update_ingestion_stats` and `update_ingestion_stats` be better?

connectors/sync_job_runner.py
```diff
@@ -112,7 +114,9 @@ async def execute(self):
                 options=bulk_options,
             )
 
+            self._reporting_cycle_start_time = time.time()
```
artem-shelkovnikov, line 117: Id' probably move it to `check_job`. Actually it might be even nicer to encapsulate this logic into a class that can be reusable in other classes. We have similar logic (do something every now and then) in other places - heartbeat, job cleanup, this.

connectors/sync_job_runner.py
artem-shelkovnikov: Also this function does not seem to be tested, can you add tests to it?

wangch079: Will move the periodic function to a separate class, and rename the method to `update_ingestion_stats`.

>Also this function does not seem to be tested, can you add tests to it?

Reporting meta is tested in `test_sync_job_runner.py`. Do you mean the `periodically` part? No, it's not tested. Will test it in the new class.

wangch079: Add a periodical checker in https://github.com/elastic/connectors-python/pull/515/commits/f1e21232c3154ccd5afb0a889038328dab49a3a8 and https://github.com/elastic/connectors-python/pull/515/commits/f9fb26a08ac6f6d5df1e0318ee798b1c60f338d2

connectors/byoc.py
```diff
@@ -204,6 +204,20 @@ async def claim(self):
         }
         await self.index.update(doc_id=self.id, doc=doc)
 
+    async def update_metadata(self, ingestion_stats=None, connector_metadata=None):
```
artem-shelkovnikov, line 207: Name of the method is confusing - it's updating `last_seen`, `metadata` + arbitrary set of fields passed in `ingestion_stats`.

Also what's stored inside `metadata`?

connectors/utils.py
```diff
@@ -472,3 +472,19 @@ def get_pem_format(key, max_split=-1):
...
+
+    def should_check(self):
+        if self._start is None:
+            self._start = time.time()
+
+        if time.time() - self._start < self._interval:
+            return False
+        else:
+            self._start = time.time()
+            return True
```
artem-shelkovnikov, line 490: Just looking at the implementation, seems like using `asyncio` Tasks can be a cleaner way to achieve this, for example like here https://stackoverflow.com/questions/37512182/how-can-i-periodically-execute-a-function-with-asyncio

What do you think?

connectors/byoc.py
wangch079: `ingestion_stats` are common data like `indexed_document_count`.
`connector_metadata` is connector-specific data, e.g. for MySQL, it can be how many tables have been queried.

For now, it's not been implemented by any connector yet.

artem-shelkovnikov: Shall we drop metadata for now then?

connectors/utils.py
wangch079: From the implementation, we have 2 intervals: 
1. JOB_REPORTING_INTERVAL - how often to report ingestion stats and send job heartbeat. This is hardcoded to 10 seconds in Ruby.
2. JOB_CHECK_INTERVAL - how often do we check whether the sync is done, it's supposed to run more frequently.
https://github.com/elastic/connectors-python/blob/f9fb26a08ac6f6d5df1e0318ee798b1c60f338d2/connectors/sync_job_runner.py#L118-L122

Not sure how the `asyncio.Task` solve this problem.

artem-shelkovnikov: What I mean is that it's possible to make an abstraction over `asyncio.Task` that repeats certain logic every X seconds. In the end current abstraction works and can be left as is (though `should_check` name is confusing too - should it be named differently, for example `period_elapsed` or `is_triggered`)

connectors/byoc.py
wangch079: Hi @danajuratoni do you think we will still implement the connector-specific metadata? If so, I think we can still leave it unchanged.

artem-shelkovnikov: I mean, if it's not something that we're gonna do in 8.7/8.8, then why have it now? :)

connectors/byoc.py
```diff
@@ -204,6 +204,20 @@ async def claim(self):
...
+        if ingestion_stats is None:
+            ingestion_stats = {}
+        if connector_metadata is None:
+            connector_metadata = {}
+
+        doc = {
+            "last_seen": iso_utc(),
+        }
+        doc.update(ingestion_stats)
```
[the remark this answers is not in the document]
wangch079: Done it in a separate PR: https://github.com/elastic/connectors-python/pull/654

connectors/byoc.py
wangch079: What I meant here is, we already have the `metadata` field in `.elastic-connectors-sync-jobs`, and this method accepts the optional parameter `connector_metadata`, so it can be easily ignored.

Unless @danajuratoni confirms that we are not going to implement connector-specific metadata any more, I think we can keep it for now.

connectors/utils.py
wangch079: I don't feel it's necessary to use `asyncio.Task`. In the [example](https://stackoverflow.com/questions/37512182/how-can-i-periodically-execute-a-function-with-asyncio) you posted, it was pushed to `create_task`, but it still needs to run `run_until_complete` to wait for it to complete. 

Same situation here that we need to wait for `self.elastic_server.done()` to proceed to `self._sync_done`. So I'd prefer to keep it in the foreground.

For `periodical_checker.should_check():`, I want to simulate [periodically](https://github.com/elastic/connectors-ruby/blob/21b8b6697007ae30f5d0555bd6271119e2155776/lib/core/sync_job_runner.rb#L244-L249) in ruby, but it seems there's no similar syntax of `yield`.

Regarding the name, the class is named `PeriodicalChecker`, which means it periodically (by configured interval) checks something, therefore the method name `should_check`. I think the method name should match the class name (`check` vs `checker`)

wangch079: I implemented it as a task: https://github.com/elastic/connectors-python/pull/515/commits/8f8854c9439b939342af7a9cb42984bc4a2ca2e1. But I haven't test it locally as I have issue starting the stack now.

artem-shelkovnikov (approved):

wangch079:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/706)|

### Questions ?

PR #516, opened 2023-02-20
Make MySQL connector terminate faster when connector is retrying

artem-shelkovnikov:
Small fix for a sleep line in MySQL connector:

```
RETRY_INTERVAL = 2

await asyncio.sleep(RETRY_INTERVAL**retry)
```

Retry count is user's input. What it means is that if something goes wrong and connector has to retry, let's say, 5 times, last sleep will be 2^5 = 32 seconds. If service is interrupted during this time, it'll have to wait until these 32 seconds pass.

This change makes use of CancellableSleep so that if service is stopped, user won't have to wait for remaining sleep time until service is interrupted.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -4,7 +4,6 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 """MySQL source module responsible to fetch documents from MySQL"""
-import asyncio
 import ssl
 from collections import OrderedDict
 
@@ -16,7 +15,7 @@ from connectors.filtering.validation import (
 )
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import RetryStrategy, retryable
+from connectors.utils import CancellableSleeps, RetryStrategy, retryable
 
 MAX_POOL_SIZE = 10
 QUERIES = {
@@ -126,6 +125,7 @@ class MySqlDataSource(BaseDataSource):
             configuration (DataSourceConfiguration): Object of DataSourceConfiguration class.
         """
         super().__init__(configuration=configuration)
+        self._sleeps = CancellableSleeps()
         self.retry_count = self.configuration["retry_count"]
         self.connection_pool = None
         self.ssl_disabled = self.configuration["ssl_disabled"]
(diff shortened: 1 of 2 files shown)


timgrein (approved):
LGTM, we could think about incorporating this change into `retryable` under `utils.py` and an option that the sleep between retries can be cancellable.

artem-shelkovnikov:
Created issue: https://github.com/elastic/enterprise-search-team/issues/4016

artem-shelkovnikov:
@wangch079 I see you dismissed the review, should I fix something?

timgrein (approved):

timgrein:
Reapproved, I guess @wangch079 you've dismissed it, because it was stale? Or should something else be addressed in this PR?

wangch079 (dismissed):
I don't remember that I dismissed the approval. 🤔 

It's a good practice to use `CancellableSleeps`, and it also makes sense to use it for all native connectors.

But it's still the implementation details of each connector. What we are doing in framework-level, is to `await data_provider.close()` in the `finally` block. And this can still happen for a custom-built connector.

Anyway, I might be overthinking, the developer should be responsible for the behaviour of his/her own connector. We just need to make sure whatever goes to `main` follow the best practice.

LGTM

wangch079 (approved):

artem-shelkovnikov:
I think we need to add `CancellableSleeps` to BaseConnector later and encourage usage of it to the developers of connectors - we can review current connectors for it later as well.

wangch079:
Approved again.

Is it really a good idea to always dismiss the previous approval even for a forced commit? cc @elastic/ingestion-team 

This is particularly troublesome considering we are located across multiple time zones. When you need something to be merged so that you can proceed, you may find no one is online.

wangch079:
>Have you faced similar problems by chance?

When I say `you may find no one is online`, that is normally me. But I don't face this problem now as recently I almost worked half in the CET time zone and half in PT time zone😅

artem-shelkovnikov:
Okay looks like it's not easily backportable, but it's not a big problem that it's not in 8.7 either, so I'm giving up backporting for sake of simplicity.

PR #517, opened 2023-02-20
Connector service cancels a job during execution

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3160

This PR made the changes to `SyncJobRunner` class:
1. to fail the job if 1) connector is deleted, or 2) sync job is deleted 3) sync job is not running (i.e. not in `in_progress` status)
2. to cancel the job if the job status is `canceling`.

#### Pre-Review Checklist


artem-shelkovnikov:
Should this PR be based on top of one of other PRs @wangch079 ?

wangch079:
>Should this PR be based on top of one of other PRs @wangch079 ?

Yes, I'm working on it now.

connectors/sync_job_runner.py
```diff
@@ -143,20 +167,28 @@ async def _sync_done(self, sync_status, sync_error=None):
...
-        if sync_status == JobStatus.ERROR:
-            await self.sync_job.fail(sync_error, ingestion_stats=ingestion_stats)
-        elif sync_status == JobStatus.SUSPENDED:
-            await self.sync_job.suspend(ingestion_stats=ingestion_stats)
-        elif sync_status == JobStatus.CANCELED:
-            await self.sync_job.cancel(ingestion_stats=ingestion_stats)
-        else:
-            await self.sync_job.done(ingestion_stats=ingestion_stats)
+        if await self.reload_sync_job():
+            if await self.reload_connector():
```
artem-shelkovnikov, line 173: What will be "else" in this case? Do we want to log something in "else"?

connectors/sync_job_runner.py
```diff
@@ -143,20 +167,28 @@ async def _sync_done(self, sync_status, sync_error=None):
...
+                await self.sync_job.suspend(ingestion_stats=ingestion_stats)
+            elif sync_status == JobStatus.CANCELED:
+                await self.sync_job.cancel(ingestion_stats=ingestion_stats)
+            else:
+                await self.sync_job.done(ingestion_stats=ingestion_stats)
+
+            await self.reload_sync_job()
+
+        if await self.reload_connector():
+            await self.connector.sync_done(self.sync_job)
```
artem-shelkovnikov, line 190: Same, what about `else`?

artem-shelkovnikov (approved):
Looks good, just left a couple questions

connectors/sync_job_runner.py
wangch079: Make sense. Will add a log.

connectors/sync_job_runner.py
wangch079: `else` means the connector is somehow deleted, so `self.connector.sync_done` will not be called. Will add a log for `else`.

connectors/sync_job_runner.py
artem-shelkovnikov: Are you planning to add a log line here?

connectors/sync_job_runner.py
artem-shelkovnikov: Same remark, I think it's worth to add a log line here

artem-shelkovnikov (approved):

connectors/sync_job_runner.py
wangch079: I added the log to `reload_connector` method:

https://github.com/elastic/connectors-python/blob/06325cc0e563317fa296dcc05d9f2c7efa887353/connectors/sync_job_runner.py#L264-L272

But I just realized it won't always log when it returns `False`. I will add a log soon.

artem-shelkovnikov (approved):

PR #519, opened 2023-02-21
[MySQL] Only allow one database to be configured

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3994##

Allow only one database to be configured for a MySQL connector. This will remove the database validation concern during filtering validation. I've considered moving the whole MySQL implementation to use `generic_database.py`, but this would need some more investigation and also prioritisation. Can be discussed later, when all SQL-like connectors follow the same topology.

A big chunk of the diff is due to removal of methods like `validate_databases`, `configured_databases` and the corresponding tests.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -6,7 +6,6 @@
 """MySQL source module responsible to fetch documents from MySQL"""
 import asyncio
 import ssl
-from collections import OrderedDict
 
 import aiomysql
 
@@ -20,7 +19,6 @@ from connectors.utils import RetryStrategy, retryable
 
 MAX_POOL_SIZE = 10
 QUERIES = {
-    "ALL_DATABASE": "SHOW DATABASES",
     "ALL_TABLE": "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = '{database}'",
     "TABLE_DATA": "SELECT * FROM {database}.{table}",
     "TABLE_PRIMARY_KEY": "SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = '{database}' AND TABLE_NAME = '{table}' AND COLUMN_KEY = 'PRI'",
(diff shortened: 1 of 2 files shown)


connectors/sources/mysql.py
```diff
@@ -123,13 +86,14 @@ def __init__(self, configuration):
         """Set up the connection to the MySQL server.
 
         Args:
-            configuration (DataSourceConfiguration): Object of DataSourceConfiguration class.
+            configuration (DataSourceConfiguration): Object  of DataSourceConfiguration class.
```
timgrein, line 89: Interesting, where this space comes from...I'll check, if that was our autoformat.

artem-shelkovnikov (commented):
Change looks good, curious - do we want to mention it somewhere to our users that we're making non-backwards-compatible change?

tarekziade:
> Change looks good, curious - do we want to mention it somewhere to our users that we're making non-backwards-compatible change?

We're not GA yet, tech preview, so I think it's ok not to

connectors/sources/mysql.py
```diff
@@ -435,80 +387,42 @@ async def fetch_documents(self, database, table, query=QUERIES["TABLE_DATA"]):
                     keys_value += f"{row.get(key)}_" if row.get(key) else ""
                 row.update(
                     {
-                        "_id": f"{database}_{table}_{keys_value}",
+                        "_id": f"{self.database}_{table}_{keys_value}",
```
parth-crest, line 390: If we are planning to move forward with only one database we can remove database for preparing `_id` and same for preparing keys [`keys.append(f"{self.database}_{table}_{column_name[0]}")`](https://github.com/elastic/connectors-python/pull/519/files#diff-996ac0843d7656c71035ae294db22a975939f9fd73685747f5faa8e3b6d64d4eR369)

timgrein: This will be part of future work as the `generic_database_source.py` does the same thing for one database. An upcoming PR will make it consistently across all SQL-like connectors

artem-shelkovnikov (approved):

PR #524, opened 2023-02-21
Set logger before parsing config

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/3900

Quoting:

> Certain logging like connectors yml parsing errors have no logger or log level specification so it is easy to be missed by admins/support.

We want Cloud (and other) deployments to properly report errors about parsing config files. To do so, I'm wrapping parsing of config file in try/except and in case error happens use provided `--filebeat` option + set log level to INFO (can actually do DEBUG, doesn't matter as long as it reports errors).

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -113,7 +113,17 @@ def run(args):
     """Runner"""
 
     # load config
-    config = load_config(args.config_file)
+    config = {}
+    try:
+        config = load_config(args.config_file)
+    except Exception as e:
+        # If something goes wrong while parsing config file, we still want
+        # to set up the logger so that Cloud deployments report errors to
+        # logs properly
+        set_logger(logging.INFO, filebeat=args.filebeat)
+        logger.exception(f"Could not parse {args.config_file}:\n{e}")
+        raise
+
     # Precedence: CLI args >> Config Setting >> INFO
     set_logger(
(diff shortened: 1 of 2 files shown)


tarekziade (approved):

wangch079:
Sorry late to the review.

I agree we should wrap `load_config` with `try...except...` (or we should wrap it inside `load_config`), but I don't think it's necessary to `set_logger` inside `except` block.

There is a `set_logger` call inside `logger` module:
https://github.com/elastic/connectors-python/blob/05c4b7a02f168abe281abcdfe02aeada98bfcbed/connectors/logger.py#L75

And the logger is also used inside `load_config`:
https://github.com/elastic/connectors-python/blob/05c4b7a02f168abe281abcdfe02aeada98bfcbed/connectors/config.py#L14-L15

The problem is that the default `set_logger` won't respect `filebeat` args.

artem-shelkovnikov:
> The problem is that the default set_logger won't respect filebeat args.

That's indeed the problem and it seems like there's no way around it except for calling `set_logger` in the `cli.py` module.

wangch079:
What do you think if we check the presence of EnvVar `ENT_SEARCH_CONFIG_PATH` if no `filebeat` param is provided? cc @tarekziade 

This will work until we decide to deploy connector service separately

Issue #275, opened 2023-01-04
Avoid double-load of files in GCS


tarekziade:
In this section of the code:

```
        logger.debug(f"Downloading {blob_name}")
        content = b""
        async with TemporaryFile("wb+") as async_buffer:
            await anext(
                self._api_call(
                    resource="objects",
                    method="get",
                    bucket=blob["bucket_name"],
                    object=blob_name,
                    alt="media",
                    userProject=self.user_project_id,
                    pipe_to=async_buffer,
                )
            )
            await async_buffer.seek(0)
            content = await async_buffer.read()
        logger.debug(f"Downloaded {blob_name} for {blob_size} bytes ")
        return {
            "_id": blob["id"],
            "_timestamp": blob["_timestamp"],
            "_attachment": get_base64_value(content=content),
        }
```

We are writing the file on disk, to load it back into a variable, to then base64 it. 
This means we get twice the file size in memory right before it's returned.

We can change this slightly by doing on-disk base64 encoding, and it will be faster and will reduce memory usage.

steps:
- create in utils a function that uses the system's `base64` CLI, see my patch in https://github.com/elastic/connectors-python/pull/239/files
- once the file is on disk, call the utility via `run_in_executor` to produce a b64 file
- load it back in `_attachment`

in the future, we will change that part so we have a decoupled service that uploads files

tarekziade:
see https://github.com/elastic/connectors-python/pull/276

PR #527, opened 2023-02-21
[New Connector] Jira connector - On Cloud and Server

moxarth-elastic:
## PR Content

Jira connector with the support for both on-cloud and on-prem platforms.
- Ping functionality with unit test cases.
- Indexing functionality with configured connection string.
- E2E test cases using Flask.

## Checklists

 - [x] Declare your dependencies in [requirements](https://github.com/elastic/connectors-python/tree/main/requirements). Make sure you pin these dependencies

## New Dependency

- [x] New external service dependency added i.e. Flask to run E2E tests.


connectors/source.py
```diff
@@ -236,6 +237,22 @@ def tweak_bulk_options(self, options):
...
+
+        Returns:
+            ssl_context: SSL context with certificate
+        """
+        certificate = certificate.replace(" ", "\n")
+        certificate = " ".join(certificate.split("\n", 1))
+        certificate = " ".join(certificate.rsplit("\n", 1))
+        ctx = ssl.create_default_context()
+        ctx.load_verify_locations(cadata=certificate)
+        return ctx
```
artem-shelkovnikov, line 254: This function should go to `utilities` module.

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+            document["_attachment"] = (await async_buffer.read()).strip()
+        await remove(temp_filename)
+        return document
+
+    async def ping(self):
+        """Verify the connection with Jira"""
+        self._validate_configuration()
+        self.session = self._generate_session()
+        if not self.ssl_disabled:
+            self.ssl_ctx = self._ssl_context(certificate=self.certificate)
```
artem-shelkovnikov, line 336: This should not happen in `ping` - it's not used in this function and `get_docs` depends on it. You should initialise it in such a way, that both `ping` and `get_docs` could work if called individually, not sequentially.

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+    @classmethod
+    def get_default_configuration(cls):
+        """Get the default configuration for Jira
+
+        Returns:
+            dictionary: Default configuration.
+        """
+        return {
+            "is_cloud": {
+                "value": IS_CLOUD,
```
artem-shelkovnikov, line 96: This default value should be inlined

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                "label": "Jira Cloud API token",
+                "type": "str",
+            },
+            "host_url": {
+                "value": "http://127.0.0.1:8080",
+                "label": "Jira host url",
+                "type": "str",
+            },
+            "ssl_disabled": {
+                "value": DEFAULT_SSL_DISABLED,
```
artem-shelkovnikov, line 126: This default value should be inlined

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                "label": "Disable SSL verification. True if SSL is disabled else False",
+                "type": "bool",
+            },
+            "ssl_ca": {
+                "value": "",
+                "label": "SSL certificate",
+                "type": "str",
+            },
+            "enable_content_extraction": {
+                "value": DEFAULT_CONTENT_EXTRACTION,
```
artem-shelkovnikov, line 136: This default value should be inlined

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                "label": "Enable content extraction (true/false)",
+                "type": "bool",
+            },
+            "retry_count": {
+                "value": RETRIES,
```
artem-shelkovnikov, line 141: This default value should be inlined

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                "type": "int",
+            },
+            "concurrent_downloads": {
+                "value": 50,
+                "label": "Number of concurrent downloads for fetching attachment content",
+                "type": "int",
+            },
+        }
+
+    def tweak_bulk_options(self, options):
```
artem-shelkovnikov, line 152: Isn't this function a validation function + assignment? I believe it could be partially moved into `_validate_configuration` (validation part), other part can go to initialiser.

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+            raise Exception(
+                f"Configured keys: {empty_connection_fields} can't be empty."
+            )
+
+        if not self.ssl_disabled and self.certificate == "":
+            raise Exception("When ssl_disabled is False then ssl_ca can't be empty.")
+
+    def _generate_session(self):
+        """Generates a new session"""
+        logger.debug("Creating a session...")
```
artem-shelkovnikov, line 192: More information could help - what's the session about? What is it responsible for? What can go wrong with session and be helpful to see in logs to trace problems?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+        )
+
+    async def close(self):
+        """Closes unclosed client session"""
+        if self.session is None:
+            return
+        await self.session.close()
+
+    async def _api_call(self, url_name, is_pagination=False, **url_kwargs):
+        """Method for adding retries whenever exception raised during an api calls
```
artem-shelkovnikov, line 224: This docstring seems to be irrelevant to the function, can you fix it?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+
+        Args:
+            url_name (str): Jira query name to be executed.
```
artem-shelkovnikov, line 227: What's `url_name`? Description says it's a "query name", name is "url_name", what name is this actually?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+            is_pagination (bool): Flag to check if the pagination is required
```
artem-shelkovnikov, line 228: Why can it be not required?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+        Returns:
+            dictionary: Content document with _id, _timestamp and attachment content
+        """
+        attachment_size = int(attachment["size"])
+        if not (self.enable_content_extraction and doit and attachment_size > 0):
+            return
+
+        attachment_name = attachment["filename"]
+        if os.path.splitext(attachment_name)[-1] not in TIKA_SUPPORTED_FILETYPES:
+            logger.debug(f"{attachment_name} can't be extracted")
```
artem-shelkovnikov, line 296:
```suggestion
            logger.debug(f"{attachment_name} is not supported by TIKA, skipping")
```

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                await asyncio.sleep(RETRY_INTERVAL**retry)
+
+    async def get_content(self, issue_key, attachment, timestamp=None, doit=False):
+        """Extracts the content for allowed file types.
+
+        Args:
+            issue_key (str): Issue key to generate `_id` for attachment document
+            attachment (dictionary): Formatted attachment document.
+            timestamp (timestamp, optional): Timestamp of attachment last modified. Defaults to None.
+            doit (boolean, optional): Boolean value for whether to get content or not. Defaults to False.
```
artem-shelkovnikov, line 285: Why have this flag?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+            return
+
+        attachment_name = attachment["filename"]
+        if os.path.splitext(attachment_name)[-1] not in TIKA_SUPPORTED_FILETYPES:
+            logger.debug(f"{attachment_name} can't be extracted")
+            return
+
+        if attachment_size > FILE_SIZE_LIMIT:
+            logger.warning(
+                f"File size {attachment_size} of file {attachment_name} is larger than {FILE_SIZE_LIMIT} bytes. Discarding the file content"
```
artem-shelkovnikov, line 301:
```suggestion
                f"File size {attachment_size} of file {attachment_name} is larger than {FILE_SIZE_LIMIT} bytes. Discarding file content"
```

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                    await async_buffer.write(data)
+                temp_filename = async_buffer.name
+
+        logger.debug(f"Calling convert_to_b64 for file : {attachment_name}")
+        await asyncio.to_thread(convert_to_b64, source=temp_filename)
+        async with aiofiles.open(file=temp_filename, mode="r") as async_buffer:
+            # base64 on macOS will add a EOL, so we strip() here
+            document["_attachment"] = (await async_buffer.read()).strip()
+        await remove(temp_filename)
+        return document
```
artem-shelkovnikov, line 329: Why do this logic with temporary files? I see that you first download the content into file, then on line 325 you read this file into memory fully, so why have a temporary file in the first place?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+
+    async def ping(self):
+        """Verify the connection with Jira"""
+        self._validate_configuration()
+        self.session = self._generate_session()
```
artem-shelkovnikov, line 334: Same as below - if it's used in `get_docs`, it should not get initialised in `ping`

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                        content,
+                        partial(
+                            self.get_content,
+                            issue_key=issue_key,
+                            attachment=copy(attachment),
+                        ),
+                    )
+                )
+
+    async def get_docs(self, filtering=None):
```
artem-shelkovnikov, line 427: Something feels weird in this function - framework is able to handle getting attachments from 3rd-party system, all you need to do in your `get_docs` is call `yield document, partial(self.get_content, document)` and framework will call the partial function you provided to download the document and update the document content with extracted content.

Why did you do it differently here?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+        attachment_task = asyncio.create_task(_document_task())
+        completed_tasks = 0
+        while True:
+            _, item = await self.queue.get()
+            if item == "FINISHED":
+                completed_tasks += 1
+                if completed_tasks == 2:
+                    break
+            else:
+                yield item
```
artem-shelkovnikov, line 464: I'm confused with this function, why is it done this way? What is it trying to do?

connectors/sources/tests/test_jira.py
```diff
@@ -0,0 +1,575 @@
...
+    )
+    source._get_attachments = Mock(return_value=AsyncIter(issue_data))
+    source._get_content = Mock(return_value={"id": "123"})
+
+    # Execute
+    await source._grab_content("TP-1", issue_data[0])
+
+
+@pytest.mark.asyncio
+async def test_get_docs():
```
artem-shelkovnikov, line 486: Ideally you should test multiple cases with different data types returned here, it's your main function that you implement and it has only single test for a very basic scenario, can you make sure you test the method properly?

connectors/sources/jira.py
moxarth-elastic: Understood your concern here. But, the connector is designed in such a way that `ping` will always gets executed first and then the `get_docs` method. So, there would not be a chance of the `get_docs` method to be called individually since the framework itself is not allowing it. However, once we have a conclusion on this [comment](https://github.com/elastic/connectors-python/pull/493#issuecomment-1438452169), we'll update the code accordingly.

artem-shelkovnikov: It's not designed this way, it works this way now because it was the most convenient way to do. It might and will change in future, so we want to build connectors that are resilient to changes.

In short, you should treat `ping` and `get_docs()` as methods that can be called in any sequence OR even on their own, all these 3 cases are valid:

1. `ping(); get_docs()`
2. `ping(); ping()`
3. `get_docs()`

connectors/sources/jira.py
moxarth-elastic: Oh, good catch! It's the url_name, not query name. I'll update the description.

connectors/sources/jira.py
moxarth-elastic: Since, Atlassian APIs does not provide pagination for fetching projects and returns all projects at a time. Also, pagination is not needed for ping since it's using `\myself` endpoint. So, it is only required for fetching the Jira issues.

artem-shelkovnikov: In this case I advise to separate this method into two - one with pagination, one without. They can both use same private methods to simplify logic.

Flags that control the flow of method that drastically is normally an anti-pattern

connectors/sources/jira.py
moxarth-elastic: Oh! got it. However this is related to this [comment](https://github.com/elastic/connectors-python/pull/493#issuecomment-1438452169) so do you want me to change this before getting the conclusion on it, or should wait for it? Till the time, i'll address other comments.

artem-shelkovnikov: I'll try to address the topic in the comment in the mean time, I'll try to get back to you asap

connectors/sources/jira.py
moxarth-elastic: This is a way that was suggested by Tarek to achieve the concurrency at the connector level. You can refer this Tarek's comment - https://github.com/elastic/connectors-python/issues/256#issuecomment-1367643403

connectors/sources/jira.py
moxarth-elastic: This is a part of Tarek's suggestion - https://github.com/elastic/connectors-python/issues/256#issuecomment-1367643403. This code is working as a consumer i.e. yielding documents to BYOC. In `get_docs` method, we've made two tasks for fetching projects and issues(including the attachments) and once both tasks are completed, consumer stops.

connectors/source.py
moxarth-elastic: Alright, will move it to utils module.

connectors/sources/jira.py
moxarth-elastic: This flag is used to manage the already indexed attachments. If the attachment has been updated then the content will get extracted (`doit` is True at that time) since the _timestamp has also updated for that attachment else `doit` remains False and content will not get extract.

connectors/sources/jira.py
moxarth-elastic: This is used to avoid the double-load of files. You can refer this issue - https://github.com/elastic/connectors-python/issues/275

tarekziade: Discussed with Artem in a meeting. This is the right approach. It can also be generalized at the framework level I added https://github.com/elastic/connectors-python/issues/534

connectors/sources/jira.py
moxarth-elastic: We are following a best practice to create a constant for the default values since that's not gonna change and pass it to configurations. So, do you want us to eliminate the constant and directly pass the default value i.e. True for `is_cloud`?

artem-shelkovnikov: My thought process here:

1. These default-value constants are used in a single place - in `get_default_configuration` method, so they are not re-usable
2. If your IDE does not support tooltips (I use vim, so I'm biased here), then to figure out the value of constant while looking at the configuration, you'll need to scroll to the beginning of file to see the value - that breaks the flow of thoughts/reading slightly.

All in all I just don't see a benefit of moving it to the beginning of the file in a constant instead of using default value. Think of it from this perspective - why labels of configurable fields are not constants too? Why field names are not constants?

To me out of these 3 only field names are worth being constants (as you actually use them in code in multiple places), and all others can just be regular values instead of constants.

connectors/sources/jira.py
artem-shelkovnikov: I see. One small thing -  I think it's nice to test this logic if possible and document what this loop means - it takes quite a while to understand what it does tbh.

connectors/sources/jira.py
moxarth-elastic: I'll update the docstring and logger to add more info about it.

connectors/sources/jira.py
moxarth-elastic: Cool, Got it. I'll make the configurations inline.

connectors/sources/jira.py
moxarth-elastic: Sure, please. Do let me know your opinion on this.

connectors/sources/jira.py
moxarth-elastic: Sure, I'll document it by adding a comment above this code hoping that's fine for you and add a test case for this logic.

connectors/sources/tests/test_jira.py
moxarth-elastic: It would be helpful if you provide some more idea which kind of test you want for this method? Since this test is covering multiple data types.

connectors/sources/jira.py
moxarth-elastic: I see, thanks for the suggestion. I'll update the code accordingly.

connectors/sources/jira.py
moxarth-elastic: This logic is already covered in the `test_get_docs` test case.

moxarth-elastic:
@tarekziade did you get a chance to review this PR? If you've any feedbacks, I will address them with the Artem's one.

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+
+        await self.fetchers.join()
+        await asyncio.gather(project_task, attachment_task)
```
wangch079, line 467: Is this line needed? If you exit the infinite loop above, does it mean both `project_task` and `attachment_task` are done?

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                self.configuration["service_account_id"],
+                self.configuration["api_token"],
+            )
+        else:
+            auth = self.configuration["username"], self.configuration["password"]
+
+        basic_auth = aiohttp.BasicAuth(
+            login=auth[0],
+            password=auth[1],
+        )
```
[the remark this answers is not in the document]
moxarth-elastic: Yes, this will work for both cloud and on-prem.

connectors/sources/jira.py
moxarth-elastic: Yes, the consumer stops when it gets `FINISHED` twice from the queue and that is going to happen at the time when both tasks are done.

connectors/sources/jira.py
```diff
@@ -0,0 +1,467 @@
...
+                "value": DEFAULT_CONTENT_EXTRACTION,
+                "label": "Enable content extraction (true/false)",
+                "type": "bool",
+            },
+            "retry_count": {
+                "value": RETRIES,
+                "label": "Maximum retries for failed requests",
+                "type": "int",
+            },
+            "concurrent_downloads": {
```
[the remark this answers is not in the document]
moxarth-elastic: Since, the framework by default support 10 concurrent downloads. So, if the product itself allows more concurrent downloads, we can increase it using this config param.

moxarth-elastic:
@wangch079 I've incorporated your comments, can you please give it a moment to review? Also, @tarekziade did you get a chance to review the PR?

connectors/sources/jira.py
```diff
@@ -0,0 +1,493 @@
...
+    ssl_context,
+)
+
+RETRY_INTERVAL = 2
+FILE_SIZE_LIMIT = 10485760
+
+FETCH_SIZE = 100
+CHUNK_SIZE = 1024
+QUEUE_SIZE = 1024
+QUEUE_MEM_SIZE = 5
```
vidok, line 40: I see you are multiplying this value by 1024*1024 then it is worth mentioning that the value is in megabytes. Another option is to move the multiplication to this line from line 86.

connectors/sources/jira.py
moxarth-elastic: Keeping this to both the places i.e. `ping` and `get_docs`

connectors/sources/jira.py
moxarth-elastic: I'll mention that the value is in megabytes through an inline comment.

moxarth-elastic:
buildkite test this

connectors/sources/jira.py
```diff
@@ -0,0 +1,493 @@
...
+
+        if self.ssl_enabled and self.certificate == "":
+            raise Exception("SSL certificate must be configured.")
+
+        if self.concurrent_downloads > MAX_CONCURRENT_DOWNLOADS:
+            raise Exception(
+                f"Configured concurrent downloads can't be set more than {MAX_CONCURRENT_DOWNLOADS}."
+            )
+
+    def _generate_session(self):
```
timgrein, line 195: I would really like to see the same pattern we've used for the Confluence connector (see: https://github.com/elastic/connectors-python/pull/568/files), i.e. creating a `JiraClient` class, which encapsulates the interaction with the server/cloud instance. Makes the connectors way more consistent.

timgrein (commented):
I think we should use the same pattern as in the confluence connector encapsulating and memoizing the client (instance).

connectors/sources/jira.py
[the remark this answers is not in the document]
wangch079: I just want to confirm, if `self.is_cloud is True`, you are trying to use `service_account_id` and `api_token` to build a basic auth.

connectors/sources/jira.py
moxarth-elastic: Sure, however the testing cycle has been completed and it would take quite some efforts from the QA end to re-test it with new changes, so I strongly recommend we raise a GitHub issue and cover it as an enhancement PR. Let me know your thoughts regarding the same.

tarekziade:
> I had concerns about how the `get_docs` is implemented.

I'll do a fill review now

> I also think it's overkill to use a queue as well. Having dedicated coroutine for project and issues, not a big deal, but I don't feel it's necessary. cc @tarekziade

`ConcurrentTasks` is just a way to group async tasks together, I don't think it's a problem to do it this way if this source class can boost how fast it's able to browse content from the 3rd party that way and stream them in a queue in a controlled throttled way. Here it's a nice way to browse different kinds of objects in parallel, that all end up being docs (projects and issues).

This technique speeds up the time taken by the connector to grab content.

connectors/sources/jira.py
```diff
@@ -0,0 +1,493 @@
...
+                async for data in response.content.iter_chunked(CHUNK_SIZE):
+                    await async_buffer.write(data)
+                temp_filename = str(async_buffer.name)
+
+        logger.debug(f"Calling convert_to_b64 for file : {attachment_name}")
+        await asyncio.to_thread(convert_to_b64, source=temp_filename)
+        async with aiofiles.open(file=temp_filename, mode="r") as async_buffer:
+            # base64 on macOS will add a EOL, so we strip() here
+            document["_attachment"] = (await async_buffer.read()).strip()
+        await remove(temp_filename)
```
tarekziade, line 342: let's try except this with just a warning if remove failed

tarekziade (dismissed):
just a few nits. LGTM otherwise

connectors/sources/jira.py
```diff
@@ -0,0 +1,493 @@
...
+
+RETRY_INTERVAL = 2
+FILE_SIZE_LIMIT = 10485760
+
+FETCH_SIZE = 100
+CHUNK_SIZE = 1024
+QUEUE_SIZE = 1024
+QUEUE_MEM_SIZE = 5  # Size in Megabytes
+MAX_CONCURRENCY = 5
+MAX_CONCURRENT_DOWNLOADS = 50  # Max concurrent download supported by jira
```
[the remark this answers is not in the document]
moxarth-elastic: I've already defined a config parameter for this value - https://github.com/elastic/connectors-python/blob/1c1d6a747221581ba23a713470540a5759cd2527/connectors/sources/jira.py#L148.

connectors/sources/jira.py
```diff
@@ -0,0 +1,493 @@
...
+PING = "ping"
+PROJECT = "project"
+ISSUES = "all_issues"
+ISSUE_DATA = "issue_data"
+ATTACHMENT_CLOUD = "attachment_cloud"
+ATTACHMENT_SERVER = "attachment_server"
+URLS = {
+    PING: "/rest/api/2/myself",
+    PROJECT: "/rest/api/2/project?expand=description,lead,url",
+    ISSUES: "/rest/api/2/search?maxResults={maxResults}&startAt={startAt}",
```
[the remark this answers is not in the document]
moxarth-elastic: Got it

connectors/sources/jira.py
```diff
@@ -0,0 +1,493 @@
...
+        Returns:
+            dictionary: Content document with _id, _timestamp and attachment content
+        """
+        attachment_size = int(attachment["size"])
+        if not (self.enable_content_extraction and doit and attachment_size > 0):
+            return
+
+        attachment_name = attachment["filename"]
+        if os.path.splitext(attachment_name)[-1] not in TIKA_SUPPORTED_FILETYPES:
+            logger.debug(f"{attachment_name} is not supported by TIKA, skipping")
```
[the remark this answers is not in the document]
moxarth-elastic: Yes, we can put a warning logger instead.

connectors/sources/jira.py
```diff
@@ -0,0 +1,493 @@
...
+
+        Args:
+            attachments (list): List of attachments for an issue
+            issue_key (str): Issue key for generating `_id` field
+        """
+        async for content, attachment in self._get_attachments(
+            attachments=attachments, issue_key=issue_key
+        ):
+            await self.queue.put(
+                (  # pyright: ignore
```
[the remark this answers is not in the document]
moxarth-elastic: Error we're getting here - 

image

tarekziade:
> > `ConcurrentTasks` is just a way to group async tasks together, I don't think it's a problem to do it this way if this source class can boost how fast it's able to browse content from the 3rd party that way and stream them in a queue in a controlled throttled way. Here it's a nice way to browse different kinds of objects in parallel, that all end up being docs (projects and issues).
> 
> FYI @tarekziade , after adding concurrency using `ConcurrentTasks`, we faced `QueueFull` error several times and so the connector would get terminated. So, can we add a `debug` or `warning` logger to show the `QueueFull` message instead of raising it?

The way it works: 
- you will get debug messages on that event (https://github.com/elastic/connectors-python/blob/main/connectors/utils.py#LL229C6-L229C6)
- if the queue stays full for more than 60 seconds, it gets raised (https://github.com/elastic/connectors-python/blob/main/connectors/utils.py#L226)

So if you are raising it means you are filing the queue super fast and the rest of the pipeline can't follow. 

Try this:

```
self.queue = MemQueue(maxsize=0, maxmemsize=10*1024*1024, refresh_timeout=120)
```


o raise it to `120`

connectors/sources/jira.py
```diff
@@ -0,0 +1,496 @@
...
+        """
+        start_at = 0
+
+        should_paginate = True
+        while should_paginate:
+            async for response in self._api_call(
+                url_name=url_name, start_at=start_at, max_results=FETCH_SIZE
+            ):
+                response_json = await response.json()
+                total = response_json.get("total")
```
tarekziade, line 283: if `total` does not exists you get `None` and the next test will fail. 

if you always get `total` and it's an int, change this to:

```
total = response_json["total"]
```

if not, make that code more robust:

```
total = int(response_json.get("total", 0))
```

connectors/sources/jira.py
```diff
@@ -0,0 +1,496 @@
...
+                attachment_name=attachment["filename"],
+            ):
+                async for data in response.content.iter_chunked(CHUNK_SIZE):
+                    await async_buffer.write(data)
+                temp_filename = str(async_buffer.name)
+
+        logger.debug(f"Calling convert_to_b64 for file : {attachment_name}")
+        await asyncio.to_thread(convert_to_b64, source=temp_filename)
+        async with aiofiles.open(file=temp_filename, mode="r") as async_buffer:
+            # base64 on macOS will add a EOL, so we strip() here
```
tarekziade, line 338: please add an issue about this so we have the same behavior for `convert_to_b64` on all platform

connectors/sources/jira.py
```diff
@@ -0,0 +1,496 @@
...
+            timezone = await response.json()
+            self.timezone = timezone["timeZone"]
+
+        async for response in self._api_call(url_name=PROJECT):
+            response = await response.json()
+            for project in response:
+                yield {
+                    "_id": f"{project['name']}-{project['id']}",
+                    "_timestamp": iso_utc(
+                        when=datetime.now(pytz.timezone(self.timezone))
```
tarekziade, line 378: why do you feed `when` ? it looks like the default value as it's done in `iso_utc`

connectors/sources/jira.py
```diff
@@ -0,0 +1,496 @@
...
+        async for content, attachment in self._get_attachments(
+            attachments=attachments, issue_key=issue_key
+        ):
+            await self.queue.put(
+                (  # pyright: ignore
+                    content,
+                    partial(
+                        self.get_content,
+                        issue_key=issue_key,
+                        attachment=copy(attachment),
```
tarekziade, line 443: why `copy` here?

connectors/utils.py
```diff
@@ -412,9 +412,7 @@ def ssl_context(certificate):
     Returns:
         ssl_context: SSL context with certificate
     """
-    certificate = certificate.replace(" ", "\n")
-    certificate = " ".join(certificate.split("\n", 1))
-    certificate = " ".join(certificate.rsplit("\n", 1))
+    certificate = get_pem_format(certificate, max_split=1)
```
tarekziade, line 415: nice!

tarekziade (commented):
LGTM, just a few nits to address

tarekziade (approved):
Approved, pending one more approval since it's a big patch, +my nits to address

connectors/sources/jira.py
moxarth-elastic: Since, we're indexing project every time with _timestamp with `iso_utc` and other jira documents are using the timestamp getting from the response. So, if Jira server is hosted in different timezone other than UTC, user will face timestamp mismatch for projects and other objects in term of timezone. So, keep it consitent, we're feeding `when` to the `iso_utc` of the timezone of Jira server.

connectors/sources/jira.py
moxarth-elastic: Here, attachment is a dictionary so any changes in the `partial` should not be reflected in the below code. Hence, we passed the `copy` to avoid changes done by the partial in the passed param.

timgrein (approved):
LGTM

PR #529, opened 2023-02-21
Documented  the `cli` module

tarekziade:

Adds doc for `cli`

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -7,14 +7,8 @@
 Command Line Interface.
 
 This is the main entry point of the framework. When the project is installed as
-a Python package, an `elastic-ingest` executable is added in the PATH.
-
-
-`main` is doing the following:
-
-- parses the arguments
-- reads the config file
-- starts the event loop against a series of services
+a Python package, an `elastic-ingest` executable is added in the PATH and
+executes the `main` function of this module, which starts the service.
 """
 import asyncio
 import functools
@@ -127,10 +121,11 @@ async def _start_service(config, loop):
 
 
 def run(args):
-    """Loads the config file, sets the logger and starts the event loop.
+    """Loads the config file, sets the logger and executes an action.
 
-    When `--action list` is used, the list of connectors is displayed and the
-    function exits.
+    Actions:
+    - list: prints out a list of all connectors and exits
+    - poll: starts the event loop and run forever (default)
     """
 
     # load config
@@ -173,10 +168,10 @@ def run(args):
 
 
 def main(args=None):
-    """Main entry point.
+    """Entry point to the service, responsible for all operations.
 
     Parses the arguments and calls `run` with them.
-    If `--version` is used, it displays the version and exits.
+    If `--version` is used, displays the version and exits.
     """
     parser = _parser()
     args = parser.parse_args(args=args)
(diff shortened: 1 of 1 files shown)


connectors/cli.py
```diff
@@ -86,6 +97,12 @@ def _parser():
 
 
 async def _start_service(config, loop):
+    """Starts the service.
+
+    Steps:
+    - performs a preflight checks using `PreflightCheck`
+    - instanciates a `MultiService` instance and run its `run` async function
```
artem-shelkovnikov, line 104:
```suggestion
    - instantiates a `MultiService` instance and runs its `run` async function
```

connectors/cli.py
```diff
@@ -86,6 +97,12 @@ def _parser():
 
 
 async def _start_service(config, loop):
+    """Starts the service.
+
+    Steps:
+    - performs a preflight checks using `PreflightCheck`
```
navarone-feekery, line 103:
```suggestion
    - performs a preflight check using `PreflightCheck`
```

connectors/cli.py
artem-shelkovnikov: I'm not sure about the description here:
- should it explicitly explain what services it runs?
- should it mention the signal handling?

connectors/cli.py
```diff
@@ -6,7 +6,15 @@
...
-Parses arguments and call run() with them.
+This is the main entry point of the framework. When the project is installed as
+a Python package, an `elastic-ingest` executable is added in the PATH.
+
+
+`main` is doing the following:
+
+- parses the arguments
+- reads the config file
+- starts the event loop against a series of services
```
artem-shelkovnikov, line 17: Is this needed since the `main` function is already documented?

connectors/cli.py
navarone-feekery: I'm assuming it's one preflight check. If there are multiple then of course `performs preflight checks` would be better.

connectors/cli.py
```diff
@@ -152,6 +173,11 @@ def run(args):
 
 
 def main(args=None):
+    """Main entry point.
+
+    Parses the arguments and calls `run` with them.
+    If `--version` is used, it displays the version and exits.
+    """
```
artem-shelkovnikov, line 180: should this description focus more on usage of `main` function? 

Something along the lines with

"This function is the entry point to the Connectors Service, it's responsible for all operations that the service can do:

- Listing all connectors (when args.action == 'list')
- Running connector service with provided args and configuration file stored in config.yaml
"

connectors/cli.py
wangch079:
```suggestion
    - starts a series of services via a `MultiService` instance
```

connectors/cli.py
tarekziade: we are doing several steps but I would assume that `preflight check` is fine -- devs can go and look at `PreflightCheck` I assume

connectors/cli.py
tarekziade: The intent of the module docstring is to offer a short explanation of what this module does without having to look around inside the module. Ideally, it should not be duplicating too much content -- so yeah we could reduce this for sure.

connectors/cli.py
tarekziade: sounds good!

artem-shelkovnikov (approved):

PR #530, opened 2023-02-21
Document the `byoei` module

tarekziade:

- added docstrings for `byoei.py`
- cleanup dead code

### DIFF
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -136,9 +136,10 @@ class Bulker:
         return res
 
     async def run(self):
-        """Creates batches of bulk calls given a queue of documents.
+        """Creates batches of bulk calls given a queue of items.
 
-        Exits when the document is "END_DOCS" or "FETCH_ERROR".
+        An item is a (size, object) tuple. Exits when the
+        item is the `END_DOCS` or `FETCH_ERROR` string.
 
         Bulk calls are executed concurrently with a maximum number of concurrent
         requests.
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -250,7 +250,7 @@ class Fetcher:
         """Iterate on a generator of documents to fill a queue of bulk operations for the `Bulker` to consume.
 
         A document might be discarded if its timestamp has not changed.
-        For documents that contain files to extract, the extraction happens in a separate task.
+        Extraction happens in a separate task, when a document contains files.
         """
         logger.info("Starting doc lookups")
         self.sync_runs = True
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -183,7 +183,7 @@ class Fetcher:
     - queue: an `asyncio.Queue` to put docs in
     - index: the target Elasticsearch index
     - existing_ids: a list of existing Elasticsearch document ids found in the index
-    - filter: a `Filter` instance -- default: `None`
+    - filter_: an instance of `Filter` to apply on the fetched document -- default: `None`
     - sync_rules_enabled: if `True`, we apply rules -- default: `False`
     - display_every -- display a log every `display_every` doc -- default: `DEFAULT_DISPLAY_EVERY`
     - concurrent_downloads: -- concurrency level for downloads -- default: `DEFAULT_CONCURRENT_DOWNLOADS`
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -68,7 +68,7 @@ class Bulker:
     Arguments:
 
     - `client` -- an instance of `connectors.es.ESClient`
-    - `queue` -- an `asyncio,Queue` instance to pull docs from
+    - `queue` -- an instance of `asyncio.Queue` to pull docs from
     - `chunk_size` -- a maximum number of operations to send per request
     - `pipeline` -- ingest pipeline settings to pass to the bulk API
     - `chunk_mem_size` -- a maximum size in MiB for each bulk request
--- a/connectors/byoei.py
+++ b/connectors/byoei.py
@@ -46,6 +46,8 @@ from connectors.utils import (
     iso_utc,
 )
 
+__all__ = ["ElasticServer"]
+
 OP_INDEX = "index"
 OP_UPSERT = "update"
 OP_DELETE = "delete"
@@ -63,13 +65,14 @@ class Bulker:
     This class runs a coroutine that gets operations out of a `queue` and collects them to
     build and send bulk requests using a `client`
 
-    The bulk requests are controlled in several ways:
+    Arguments:
+
+    - `client` -- an `connectors.es.ESClient` instance
+    - `queue` -- an `asyncio,Queue` instance to pull docs from
     - `chunk_size` -- a maximum number of operations to send per request
+    - `pipeline` -- ingest pipeline settings to pass to the bulk API
     - `chunk_mem_size` -- a maximum size in MiB for each bulk request
     - `max_concurrency` -- a maximum number of concurrent bulk requests
-
-    Extra options:
-    - `pipeline` -- ingest pipeline settings to pass to the bulk API
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


connectors/byoei.py
```diff
@@ -63,13 +65,14 @@ class Bulker:
     This class runs a coroutine that gets operations out of a `queue` and collects them to
     build and send bulk requests using a `client`
 
-    The bulk requests are controlled in several ways:
+    Arguments:
+
+    - `client` -- an `connectors.es.ESClient` instance
```
navarone-feekery, line 70:
```suggestion
    - `client` -- a `connectors.es.ESClient` instance
```

connectors/byoei.py
```diff
@@ -167,25 +177,30 @@ async def run(self):
...
     """Grabs data and adds them in the queue for the bulker.
 
-    This class runs a coroutine that puts docs in `queue`, given
-    a document generator.
+    This class runs a coroutine that puts docs in `queue`, given a document generator.
+
+    Arguments:
+    - queue: an `asyncio.Queue` to put docs in
+    - index: the target Elasticsearch index
+    - existing_ids: a list of existing ELasticsearch documents ids found in the index
```
navarone-feekery, line 185:
```suggestion
    - existing_ids: a list of existing Elasticsearch document ids found in the index
```

navarone-feekery (commented):
Added comments on some typos

connectors/byoei.py
artem-shelkovnikov: What about "an instance of `TYPE`" convention for such cases?

navarone-feekery: I like that more ☝🏻

navarone-feekery: Updated my comment to reflect that

connectors/byoei.py
```diff
@@ -63,13 +65,14 @@ class Bulker:
     This class runs a coroutine that gets operations out of a `queue` and collects them to
     build and send bulk requests using a `client`
 
-    The bulk requests are controlled in several ways:
+    Arguments:
+
+    - `client` -- an `connectors.es.ESClient` instance
+    - `queue` -- an `asyncio,Queue` instance to pull docs from
```
navarone-feekery, line 71: Similar to the other comment;
```suggestion
    - `queue` -- an instance of `asyncio,Queue` to pull docs from
```

connectors/byoei.py
```diff
@@ -167,25 +177,30 @@ async def run(self):
...
 
-    This class runs a coroutine that puts docs in `queue`, given
-    a document generator.
+    This class runs a coroutine that puts docs in `queue`, given a document generator.
+
+    Arguments:
+    - queue: an `asyncio.Queue` to put docs in
+    - index: the target Elasticsearch index
+    - existing_ids: a list of existing ELasticsearch documents ids found in the index
+    - filter: a `Filter` instance -- default: `None`
```
navarone-feekery, line 186: Similar to other comment;
```suggestion
    - filter: an instance of `Filter` -- default: `None`
```

connectors/byoei.py
```diff
@@ -418,6 +435,16 @@ async def async_bulk(
         sync_rules_enabled=False,
         options=None,
     ):
+        """Performs a batch of `_bulk` calls, given a generator of documents
+
+        Arguments:
+        - index: target index
+        - generator: documents generator
+        - pipeline: ingest pipeline settings to pass to the bulk API
+        - filter: a `Filter` instances -- default: `None`
```
navarone-feekery, line 444:
```suggestion
        - filter: an instance of `Filter` -- default: `None`
```

connectors/byoei.py
```diff
@@ -133,6 +136,13 @@ async def _batch_bulk(self, operations):
         return res
 
     async def run(self):
+        """Creates batches of bulk calls given a queue of documents.
+
+        Exits when the document is "END_DOCS" or "FETCH_ERROR".
```
timgrein, line 141:
```suggestion
        Exits when the operation is either "END_DOCS" or "FETCH_ERROR".
```

Shouldn't that be "operation"?

connectors/byoei.py
```diff
@@ -133,6 +136,13 @@ async def _batch_bulk(self, operations):
...
+
+        Bulk calls are done concurrently with a maximum number of concurrent
```
timgrein, line 143:
```suggestion
        Bulk calls are executed concurrently with a maximum number of concurrent
```

connectors/byoei.py
```diff
@@ -234,7 +246,12 @@ async def _deferred_index(self, lazy_download, doc_id, doc, operation):
             }
         )
 
-    async def get_docs(self, generator):
+    async def run(self, generator):
+        """Iterate on a generator of documents to fill a queue of bulk operations for the `Bulker` to consume.
+
+        A document might be discarded if its timestamp has not changed.
+        For documents that contain files to extract, the extraction happens in a separate task.
```
timgrein, line 253:
```suggestion
        Extraction happens in a separate task, when a document contains files.
```

timgrein (commented):
Some small comments and one question.

connectors/byoei.py
```diff
@@ -63,13 +65,14 @@ class Bulker:
...
 
-    The bulk requests are controlled in several ways:
+    Arguments:
+
+    - `client` -- an `connectors.es.ESClient` instance
+    - `queue` -- an `asyncio,Queue` instance to pull docs from
     - `chunk_size` -- a maximum number of operations to send per request
+    - `pipeline` -- ingest pipeline settings to pass to the bulk API
     - `chunk_mem_size` -- a maximum size in MiB for each bulk request
     - `max_concurrency` -- a maximum number of concurrent bulk requests
```
wangch079, line 75: Might not relate to this PR. But I think we should have default values for fields like `chunk_size`, `chunk_mem_size` and `max_concurrency`.

connectors/byoei.py
wangch079:
```suggestion
    - filter_: an instance of `Filter` to apply on the fetched document -- default: `None`
```

connectors/byoei.py
```diff
@@ -167,25 +177,30 @@ async def run(self):
...
-    This class runs a coroutine that puts docs in `queue`, given
-    a document generator.
+    This class runs a coroutine that puts docs in `queue`, given a document generator.
+
+    Arguments:
+    - queue: an `asyncio.Queue` to put docs in
+    - index: the target Elasticsearch index
+    - existing_ids: a list of existing ELasticsearch documents ids found in the index
+    - filter: a `Filter` instance -- default: `None`
+    - sync_rules_enabled: if `True`, we apply rules -- default: `False`
```
wangch079, line 187: Why is this param is added in the first place? I think it's redundant to `Fetcher`, which should only care about `filter_`.

connectors/byoei.py
wangch079:
```suggestion
        - filter_: an instance of `Filter` to apply on the fetched document  -- default: `None`
```

connectors/byoei.py
tarekziade: well, technically the queue holds docs, except for that very specific item that is used as a signal.
Maybe we could use the word `item` instead, and say that items in the queue can be documents or on of those constants

connectors/byoei.py
tarekziade: we do have defaults, they are set in the function that creates the instance, see https://github.com/elastic/connectors-python/blob/main/connectors/byoei.py#L425-L433

but maybe they could be moved at the signature level yeah. good follow-up

connectors/byoei.py
tarekziade: it's used to instantiate `BasicRuleEngine`, which will run for each document with `self.basic_rule_engine.should_ingest(doc)` to decide if we drop it or not

artem-shelkovnikov (approved):

artem-shelkovnikov (approved):

PR #532, opened 2023-02-22
[Generic Database Connector] Make tables configurable

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/3997

This PR adds a new configuration to the generic database connector. This will be inherited by the oracle and PostgreSQL connector. 
The default value is set to fetch all tables, which is aligned with the previous behavior.

The PR contains a needed modification to `AsyncIterator`, which influenced one test in `test_byoc.py` therefore the change in that file.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/generic_database.py
+++ b/connectors/sources/generic_database.py
@@ -13,9 +13,27 @@ from connectors.logger import logger
 from connectors.source import BaseDataSource
 from connectors.utils import iso_utc
 
+ALL_TABLES = "*"
+
 DEFAULT_FETCH_SIZE = 50
 DEFAULT_RETRY_COUNT = 3
 DEFAULT_WAIT_MULTIPLIER = 2
+
+
+def configured_tables(tables):
+    def table_filter(table):
+        return table is not None and len(table) > 0
+
+    return (
+        list(
+            filter(
+                lambda table: table_filter(table),
+                map(lambda table: table.strip(), tables.split(",")),
+            )
+        )
+        if isinstance(tables, str)
+        else list(filter(lambda table: table_filter(table), tables))
+    )
 
 
 class GenericBaseDataSource(BaseDataSource):
@@ -38,6 +56,7 @@ class GenericBaseDataSource(BaseDataSource):
         self.host = self.configuration["host"]
         self.port = self.configuration["port"]
         self.database = self.configuration["database"]
+        self.tables = self.configuration["tables"]
         self.is_async = False
         self.engine = None
         self.dialect = ""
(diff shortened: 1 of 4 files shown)


connectors/sources/generic_database.py
```diff
@@ -77,6 +96,7 @@ def get_default_configuration(cls):
                 "label": "Database",
                 "type": "str",
             },
+            "tables": {"value": ALL_TABLES, "label": "Tables", "type": "list"},
```
artem-shelkovnikov, line 99:
```suggestion
            "tables": {"value": ALL_TABLES, "label": "Comma-separated list of tables", "type": "list"},
```

Should we mention this for now since we support only strings?

connectors/sources/generic_database.py
```diff
@@ -362,27 +376,43 @@ async def fetch_rows(self, schema=None):
...
+                        query_name="ALL_TABLE",
+                        user=self.user.upper(),
+                        database=self.database,
+                        schema=schema,
+                    )
+                ),
+            )
+            if fetch_all_tables
+            else tables
+        )
```
artem-shelkovnikov, line 416: I think it still makes sense to do some sort of validation of tables that user entered before trying to fetch data - what if they entered invalid table names, or ones that are not accessible by the user?

timgrein: I'll create a tech-debt issue for that and check, if we want to address/prioritise this on our side or simply let the datasource errors bubble up.

connectors/sources/generic_database.py
timgrein: Will the label be rendered as the text directly above the input field or will the key be capitalized and rendered above the input field?

connectors/sources/generic_database.py
artem-shelkovnikov: I'm not quite sure about it because it changes logic a lot and should be delivered in same release this change is delivered at least.

Before this problem would not exist - if user specify correct database, tables cannot be incorrect, thus it's not possible to have database inconsistently synced (some tables yes, some tables not).

So to sum up, I think it's okay to handle it in separate PR, but I think it should happen in this release as well.

timgrein: Valid points, then we should also do the same validation for the database. Does it exist? is it accessible?

Created the issue for 8.8: https://github.com/elastic/enterprise-search-team/issues/4030

artem-shelkovnikov: For database it's a bit more lax cause if DB does not exist/can't be accessed, you'll get no data. For tables you'll get partial data which is worse than no data at all IMO

artem-shelkovnikov (approved):

PR #533, opened 2023-02-22
[Docs] Document the filtering module

timgrein:

Add documentation for the most important classes and methods in the filtering module. Skipped on the documentation, where it doesn't add any information, which can not be derived from the class name.

### DIFF
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -328,7 +328,7 @@ class BasicRule:
         return self.policy == Policy.INCLUDE
 
     def coerce_rule_value_based_on_document_value(self, doc_value):
-        """Method coercing the value inside the basic rule.
+        """Coerce the value inside the basic rule.
 
         This method tries to coerce the value inside the basic rule to the type used in the document.
 
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -96,6 +96,7 @@ def try_coerce(value):
 
 class RuleMatchStats:
     """Class to record how many documents a basic rule matched and which policy it used."""
+
     def __init__(self, policy, matches_count):
         self.policy = policy
         self.matches_count = matches_count
@@ -117,6 +118,7 @@ class RuleMatchStats:
 
 class BasicRuleEngine:
     """Class executing a list of basic rules in order against a document."""
+
     def __init__(self, rules):
         self.rules = rules
         self.rules_match_stats = {
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -288,7 +288,7 @@ class BasicRule:
             - GREATER_THAN: Is the document's field value greater than the basic rule's value?
             - EQUALS: Is the document's field value equal to the basic rule's value?
 
-        If the basic rule is the default rule it's always a match.
+        If the basic rule is the default rule it's always a match (the default rule matches every document).
         If the field is not in the document it's always a no match.
 
         Arguments:
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -132,7 +132,7 @@ class BasicRuleEngine:
     def should_ingest(self, document):
         """Check, whether a document should be ingested or not.
 
-        By default the document will be ingested, if it doesn't match any rule.
+        By default, the document will be ingested, if it doesn't match any rule.
 
         Arguments:
         - `document`: document matched against the basic rules
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -95,7 +95,7 @@ def try_coerce(value):
 
 
 class RuleMatchStats:
-    """Class to record how many documents a basic rule matched and which policy it used."""
+    """RuleMatchStats records how many documents a basic rule matched and which policy it used."""
 
     def __init__(self, policy, matches_count):
         self.policy = policy
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -117,7 +117,11 @@ class RuleMatchStats:
 
 
 class BasicRuleEngine:
-    """Class executing a list of basic rules in order against a document."""
+    """BasicRuleEngine matches a document against a list of basic rules in order.
+
+    The main concern of the engine is to decide, whether a document should be ingested during a sync or not.
+    It also records stats, which basic rule matched how many documents with a certain policy.
+    """
 
     def __init__(self, rules):
         self.rules = rules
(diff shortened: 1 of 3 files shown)


connectors/filtering/basic_rule.py
```diff
@@ -94,6 +95,8 @@ def try_coerce(value):
 
 
 class RuleMatchStats:
+    """Class to record how many documents a basic rule matched and which policy it used."""
```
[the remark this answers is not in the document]
artem-shelkovnikov: Also looking at the class, why is it made immutable? I see it creates a new instance of RuleMatchStats every time something is added to the stats?

Why a simple hash would not make this job?

connectors/filtering/basic_rule.py
```diff
@@ -114,13 +117,22 @@ def __eq__(self, other):
 
 
 class BasicRuleEngine:
+    """Class executing a list of basic rules in order against a document."""
```
artem-shelkovnikov, line 120: Same as above - comment can explain better:

1. What does it actually do - current description is so abstract it does not explain what the class is supposed to be used for
2. Provide an example of usage

connectors/filtering/basic_rule.py
```diff
@@ -114,13 +117,22 @@ def __eq__(self, other):
...
     def __init__(self, rules):
         self.rules = rules
         self.rules_match_stats = {
             BasicRule.DEFAULT_RULE_ID: RuleMatchStats(Policy.INCLUDE, 0)
         }
 
     def should_ingest(self, document):
+        """Method to check, whether a document should be ingested or not.
+
+        On default the document will be ingested, if it doesn't match any rule.
```
artem-shelkovnikov, line 131:
```suggestion
        By default the document will be ingested, if it doesn't match any rule.
```

What about "not by default" case?

connectors/filtering/basic_rule.py
```diff
@@ -220,6 +232,8 @@ def from_string(cls, string):
 
 
 class BasicRule:
+    """Class representing a basic rule, which is part of the connector's filtering."""
```
artem-shelkovnikov, line 235: Same as above - "Class" s redundant, what is this class for, why is it a basic rule and what's not a basic rule?

connectors/filtering/basic_rule.py
```diff
@@ -253,6 +267,23 @@ def default_rule(cls):
...
+        A basic rule matches or doesn't match a document based on the following comparisons:
+            - STARTS_WITH: Does the document's field value start with the basic rule's value?
+            - ENDS_WITH: Does the document's field value end with the basic rule's value?
+            - CONTAINS: Does the document's field value contain the basic rule's value?
+            - REGEX: Does the document's field value match the basic rule's regex?
+            - LESS_THAN: Is the document's field value less than the basic rule's value?
+            - GREATER_THAN: Is the document's field value greater than the basic rule's value?
+            - EQUALS: Is the document's field value equal to the basic rule's value?
+
+        If the basic rule is the default rule it's always a match.
```
artem-shelkovnikov, line 281: What's a "default rule"?

connectors/filtering/basic_rule.py
```diff
@@ -253,6 +267,23 @@ def default_rule(cls):
...
+        If the field is not in the document it's always a no match.
+
+        Arguments:
+        - `document`: document to check, if it matches
```
artem-shelkovnikov, line 285:
```suggestion
        - `document`: document to check for matches
```

connectors/filtering/basic_rule.py
```diff
@@ -287,6 +318,14 @@ def is_include(self):
         return self.policy == Policy.INCLUDE
 
     def coerce_rule_value_based_on_document_value(self, doc_value):
+        """Method coercing the value inside the basic rule.
+
+        This method tries to coerce the value inside the basic rule to the type used in the document.
+
+        Arguments:
+        - `doc_value`: one value of a document
```
artem-shelkovnikov, line 326:
```suggestion
        - `doc_value`: value of the field in the document to coerce
```

connectors/filtering/validation.py
```diff
@@ -196,6 +196,12 @@ def validate(cls, rules):
 
 
 class BasicRulesSetSemanticValidator(BasicRulesSetValidator):
+    """Class validating, that a set of rules does not contain semantic duplicates.
```
artem-shelkovnikov, line 199:
```suggestion
    """BasicRulesSetSemanticValidator can be used to validate that a set of filtering rules does not contain semantic duplicates.
```

connectors/filtering/validation.py
```diff
@@ -196,6 +196,12 @@ def validate(cls, rules):
...
+
+    A semantic duplicate is defined as two basic rules having the same values for `field`, `rule` and `value`.
+    Therefore, two basic rules are also seen as semantic duplicates, if their `policy` values differ.
```
artem-shelkovnikov, line 202: What happens in this case? Is the rule set declared as invalid?

connectors/filtering/validation.py
```diff
@@ -255,6 +261,8 @@ def validate(cls, rule):
 
 
 class BasicRuleNoMatchAllRegexValidator(BasicRuleValidator):
+    """Class validating that a basic rule does not use a match all regex."""
```
artem-shelkovnikov, line 264:
```suggestion
    """BasicRuleNoMatchAllRegexValidator can be used to check that a basic rule does not use a match all regex."""
```

What's the deal with match all regex here?

connectors/filtering/validation.py
```diff
@@ -279,6 +287,8 @@ def validate(cls, basic_rule_json):
 
 
 class BasicRuleAgainstSchemaValidator(BasicRuleValidator):
+    """Class validating if a basic rule follows the specified json schema."""
```
artem-shelkovnikov, line 290:
```suggestion
    """BasicRuleAgainstSchemaValidator can be used to check if basic rule follows specified json schema."""
```

connectors/filtering/validation.py
timgrein: I actually can not recall, what was the exact rationale to prohibit match all regular expressions. Should be somewhere in the original implementation doc.

connectors/filtering/basic_rule.py
timgrein: A document has one match per rule or no match. "for matches" would imply that the doc can be matched multiple for the same rule IMO

connectors/filtering/basic_rule.py
timgrein: Added an explanation

connectors/filtering/validation.py
timgrein: Added an explanation.

tarekziade (approved):

PR #535, opened 2023-02-23
Refactor the prepare connector to a separate method

wangch079:
This PR did a small refactoring, to extract the prepare connector logic into a separate method.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -586,8 +586,7 @@ class Connector(ESDocument):
             yield doc, lazy_download
 
     async def prepare(self, config):
-        """
-        Prepares the custom connector
+        """Prepares the custom connector
 
         When a custom connector is configured in the config.yml, this method updates service type and the default
         configuration when they are absent.
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -586,9 +586,11 @@ class Connector(ESDocument):
             yield doc, lazy_download
 
     async def prepare(self, config):
-        """Prepares the connector, given a configuration
-        If the connector id and the service type is in the config, we want to
-        populate the service type and then sets the default configuration.
+        """
+        Prepares the custom connector
+
+        When a custom connector is configured in the config.yml, this method updates service type and the default
+        configuration when they are absent.
         """
         configured_connector_id = config.get("connector_id", "")
         if self.id != configured_connector_id:
--- a/connectors/services/sync.py
+++ b/connectors/services/sync.py
@@ -60,23 +60,7 @@ class SyncService(BaseService):
         if connector.native:
             logger.debug(f"Connector {connector.id} natively supported")
 
-        try:
-            await connector.prepare(self.config)
-        except ServiceTypeNotConfiguredError:
-            logger.error(
-                f"Service type is not configured for connector {self.config['connector_id']}"
-            )
-            return
-        except ConnectorUpdateError as e:
-            logger.error(e)
-            return
-        except ServiceTypeNotSupportedError:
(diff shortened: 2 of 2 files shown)


connectors/services/sync.py
```diff
@@ -164,3 +148,27 @@ async def _run(self):
                 await self.sync_job_index.close()
             await es.close()
         return 0
+
+    async def _prepare_connector(self, connector):
```
artem-shelkovnikov, line 152: What do you think of adding a docstring for this method @wangch079?

artem-shelkovnikov (approved):
LGTM, added a suggestion about docstring

connectors/services/sync.py
wangch079: This method `_prepare_connector` just calls `connector.prepare()`, catches all the possible exceptions, and then returns `True` or `False` indicating whether this operation is successful.

I'm not sure whether I should add `docstring` to a **private** method, so I updated the `docstring` of `Connector.prepare()` method.

connectors/byoc.py
```diff
@@ -586,9 +586,11 @@ async def prepare_docs(self, data_provider, pipeline=None, filtering=None):
             yield doc, lazy_download
 
     async def prepare(self, config):
-        """Prepares the connector, given a configuration
-        If the connector id and the service type is in the config, we want to
-        populate the service type and then sets the default configuration.
+        """
```
tarekziade, line 589: pep8, don't leave an empty line when you start docstrings

tarekziade (commented):
I am wondering what is the goal of this refactoring.

wangch079:
>I am wondering what is the goal of this refactoring.

To make the `_one_sync` function more readable. Actually what the method `_prepare_connector` does is to call `connector.prepare`, and catch all the exceptions, log the error, and return `True` or `False`. 

I think we don't even need `_prepare_connector` method. We can just log the errors inside `connector.prepare`. WDYT?

tarekziade:
> > I am wondering what is the goal of this refactoring.
> 
> To make the `_one_sync` function more readable. Actually what the method `_prepare_connector` does is to call `connector.prepare`, and catch all the exceptions, log the error, and return `True` or `False`.
> 
> I think we don't even need `_prepare_connector` method. 

I don't find it more readable because you are adding more lines, with some `return True` and `return False` and creating a new method that is only used by `_one_sync` - and introducing a change where you catch all exceptions.

The original code is smaller and easy to read in one place. 

> We can just log the errors inside `connector.prepare`. WDYT?

That is a different change, `connector.prepare` was built by raising errors and it's already logging them.

If your goal is to reduce the size of `_one_sync`, maybe the trick is to have only one try..catch:

```python
        try:
            await connector.prepare(self.config)
        except ConnectorPrepareError as e:
            if e.critical:
                await connector.error(e)
                self.raise_if_spurious(e)
            else:
                logger.error(e.msg)
            return
```

and create a `ConnectorPrepareError` with a hierarchy of errors, and a `critical` flag

wangch079:
>The original code is smaller and easy to read in one place.

👌 

> If your goal is to reduce the size of `_one_sync`, maybe the trick is to have only one try..catch:
> and create a `ConnectorPrepareError` with a hierarchy of errors, and a `critical` flag

I think it's better to log different messages for different errors. 

Will close this PR.

tarekziade:
> I think it's better to log different messages for different errors.

this is what a hierarchy of errors gives you:

```
class ConnectorPrepareError(Exception):
    pass

class ServiceTypeNotConfiguredError(ConnectorPrepareError):
    msg = 'Service not configured'
    critical = False

```

PR #536, opened 2023-02-23
Extract ready_to_sync method from Connector to SyncService

wangch079:
This PR extracts this part of `Connector.sync` method to `SyncService`:

https://github.com/elastic/connectors-python/blob/a5976d20cd8277ae46511f7176662afc889e56ec/connectors/byoc.py#L686-L719

The logic checks whether the connector is ready to sync (on-demand, scheduled or restart a suspended job), and whether there's any change in the remote source. The logic should not be part of `sync` method.

#### Pre-Review Checklist

### DIFF
--- a/connectors/services/sync.py
+++ b/connectors/services/sync.py
@@ -98,24 +98,17 @@ class SyncService(BaseService):
             )
 
         source_klass = self.source_klass_dict[connector.service_type]
-
-        try:
-            data_provider = source_klass(connector.configuration)
-        except Exception as e:
-            logger.critical(e, exc_info=True)
-            raise DataSourceError(
-                f"Could not instantiate {source_klass} for {connector.service_type}"
+        if connector.features.sync_rules_enabled():
+            await connector.validate_filtering(
+                validator=source_klass(connector.configuration)
             )
 
-        if connector.features.sync_rules_enabled():
-            await connector.validate_filtering(validator=data_provider)
-
-        if not await self._ready_to_sync(connector, data_provider):
+        if not await self._ready_to_sync(connector):
             return
 
         await connector.sync(
             self.sync_job_index,
-            data_provider,
+            source_klass,
             es,
             self.bulk_options,
         )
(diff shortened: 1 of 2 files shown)


connectors/services/sync.py
```diff
@@ -164,3 +177,32 @@ async def _run(self):
...
+                logger.debug("Restarting sync after suspension")
+                return True
+            # And only then we check if we need to run sync right now or not
+            if next_sync - self.idling > 0:
+                logger.debug(
+                    f"Next sync for connector {connector.id} due in {int(next_sync)} seconds"
+                )
+                return False
+            # Lastly we skip job scheduling if there's no change in the remote source
+            if not await data_provider.changed():
```
wangch079, line 199: Hi @tarekziade , in the current implementation, when customers click `Sync` in Kibana or scheduled a sync, but there's no change in the remote source, there won't be any job scheduled. Especially for the forced sync, it will definitely confuse customers and they may think this is a bug. cc @danajuratoni 

Another approach is that we skip this check and schedule a job, and directly mark the job as `completed` during job execution:
https://github.com/elastic/connectors-python/blob/c71b575df81846e0ac4976e5b1278cc53155fdd9/connectors/sync_job_runner.py#L66-L71

WDYT?

tarekziade: Makes a lot of sense

connectors/services/sync.py
```diff
@@ -164,3 +177,32 @@ async def _run(self):
                 await self.sync_job_index.close()
             await es.close()
         return 0
+
+    async def _ready_to_sync(self, connector, data_provider):
```
tarekziade, line 181: I am wondering if we should call this `_should_sync`

tarekziade (dismissed):
LGTM. The behavior is slightly changed in the ordering I think though, I am wondering how come we did not need to change some tests or add one

tarekziade (approved):

PR #537, opened 2023-02-23
[new connector] SharePoint - Online & Servers

akanshi-crest:
## PR Content
SharePoint connector with the support for both on-cloud and on-prem platforms.

Ping functionality with unit test cases.
Indexing functionality.
E2E test cases using Flask.

#### Pre-Review Checklist
- [x] Declare your dependencies in [requirements](https://github.com/elastic/connectors-python/tree/main/requirements). Make sure you pin these dependencies
- [ ] The test backend needs to return more than 10k documents due to 10k being a default size limit for Elasticsearch pagination. Having more than 10k documents returned from the test backend will help the testing connector more deeply.

#### New Dependency

- [x] New external service dependency added i.e. Flask to run E2E tests.


connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+        self.site_collections = self.configuration["site_collections"]
+        self.ssl_disabled = self.configuration["ssl_disabled"]
+        self.host_url = self.configuration["host_url"]
+        self.certificate = self.configuration["ssl_ca"]
+        self.enable_content_extraction = self.configuration["enable_content_extraction"]
+        self.retry_count = self.configuration["retry_count"]
+        self.ssl_ctx = False
+        self.session = None
+        self.access_token = None
+        self.token_expires_in = None
```
artem-shelkovnikov, line 157: As mentioned in slack, these properties should be handled differently - they are initialised in `ping` and other functions which can cause breakage if `get_docs` is going to be called by different thread/host.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+
+    @classmethod
+    def get_default_configuration(cls):
+        """Get the default configuration for Sharepoint
+
+        Returns:
+            dictionary: Default configuration.
+        """
+        return {
```
artem-shelkovnikov, line 166: What's our stance on default field values @tarekziade @danajuratoni?

We have them here for FTests, but after @timgrein does his changes we can remove them or provide defaults only for the fields that we need.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+                ):
+                    await self.close()
+                    self.session = await self._generate_session()
+                retry += 1
+                if retry > self.retry_count:
+                    raise exception
+                logger.warning(
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                )
+                await asyncio.sleep(RETRY_INTERVAL**retry)
```
artem-shelkovnikov, line 374: Instead of doing asyncio.sleep, please use CancellableSleeps, see pattern here: https://github.com/elastic/connectors-python/blob/main/connectors/sources/mysql.py#L91

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+
+    async def ping(self):
+        """Verify the connection with Sharepoint"""
+
+        self._validate_configuration()
+
+        if not self.ssl_disabled:
+            self.ssl_ctx = self._ssl_context(certificate=self.certificate)
+
+        self.session = await self._generate_session()
```
artem-shelkovnikov, line 438: As mentioned, `ping` should not populate `self.<PROPERTY>` at all. You cannot rely on it being called before `get_docs`.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+        try:
+            await anext(
+                self._api_call(
+                    url_name=PING,
+                    site_collections=self.site_collections[0],
+                    host_url=self.host_url,
+                )
+            )
+            logger.info("Successfully connected to the Sharepoint")
```
artem-shelkovnikov, line 447:
```suggestion
            logger.debug("Successfully connected to the Sharepoint")
```

artem-shelkovnikov: Also you can add host_url so that it's easier to figure out where it was connecting

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+            file_relative_url (str): Relative url of file
+            site_url (str): Site path of sharepoint
+            timestamp (timestamp, optional): Timestamp of item last modified. Defaults to None.
+            doit (boolean, optional): Boolean value for whether to get content or not. Defaults to None.
+
+        Returns:
+            dictionary: Content document with id, timestamp & text.
+        """
+
+        if not (self.enable_content_extraction and doit and document["size"]):
```
artem-shelkovnikov, line 468: cc @tarekziade - is `self.enable_content_extraction` a valid use-case?

I know that we have deployment-wide settings in pipeline to enable/disable binary content extraction. Why have additional configurable field for it?

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+            source_file_name = async_buffer.name
+
+        await asyncio.to_thread(
+            convert_to_b64,
+            source=source_file_name,
+        )
+        async with aiofiles.open(file=source_file_name, mode="r") as target_file:
+            # base64 on macOS will add a EOL, so we strip() here
+            attachment_content = (await target_file.read()).strip()
+        await remove(source_file_name)
```
artem-shelkovnikov, line 498: Is it possible to make it a utility function that you re-use in all connectors?

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+        """Invokes a GET call to the Sharepoint Server/Online.
+
+        Args:
+            site_url(string): site url to the sharepoint farm.
+            param_name(string): parameter name whether it is SITES, LISTS, DRIVE_ITEM, LIST_ITEM.
+            list_id(string, optional): Id of list item or drive item. Defaults to None.
+        Yields:
+            Response of the GET call.
+        """
+        skip, top = 0, 5000
```
artem-shelkovnikov, line 515: 5000 can be a constant

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+                        skip=skip,
+                        top=top,
+                        host_url=self.host_url,
+                    )
+                )
+                response_result = response.get("value", [])
+                yield response_result
+
+                skip += 5000
+                if len(response_result) < 5000:
```
artem-shelkovnikov, line 532: Do they guarantee to always return 5000 documents if there are documents left? Just asking cause for Salesforce their page sizes are dynamic depending on size of payload, so wanted to double check here too.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+        """Executes the logic to fetch lists and items (list-item, drive-item) in async manner.
+
+        Args:
+            site(string): Path of site.
+        Yields:
+            dictionary: dictionary containing meta-data of the list, list item and drive item.
+        """
+        async for list_data in self.invoke_get_call(site_url=site, param_name=LISTS):
+            for result in list_data:
+                if result.get("BaseType") == 1:
```
artem-shelkovnikov, line 677: This needs a comment. What is `1`?

artem-shelkovnikov: Or it can be a constant.

connectors/sources/tests/fixtures/sharepoint/fixture.py
```diff
@@ -0,0 +1,248 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+"""Module to handle api calls received from connector."""
```
artem-shelkovnikov, line 6: So we're faking the data in Sharepoint cause there's no way to host a Sharepoint Instance locally?

cc @tarekziade

tarekziade: yes, this is a compromise to have an e2e test without requiring an external service. Deploying Sharepoint requires a final human intervention on a web page. An improvement in the future will be to script the browser with Cypress or equivalent.

connectors/sources/sharepoint.py
tarekziade: why is it a problem to have a default value for each field? it allows the connector to work out of the box for the e2e test

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+
+def get_expires_at(expires_in):
+    """Adds seconds to the current utc time.
+
+    Args:
+        expires_in (int): Seconds after which the token will expire.
+    """
+    expires_at = datetime.utcnow() + timedelta(seconds=expires_in)
+    # account for clock skew
+    expires_at -= timedelta(seconds=20)
```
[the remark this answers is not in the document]
akanshi-elastic: I just refer this stack overflow answer for dealing with time variation at the time of creating token: https://stackoverflow.com/questions/47153080/clock-skew-and-tokens

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+            "client_secret": self.configuration["secret_id"],
+        }
+        headers = {"Content-Type": "application/x-www-form-urlencoded"}
+
+        async with aiohttp.request(
+            method="POST", url=url, data=data, headers=headers
+        ) as response:
+            json_data = await response.json()
+            self.access_token = json_data["access_token"]
+            self.token_expires_in = get_expires_at(int(json_data["expires_in"]))
```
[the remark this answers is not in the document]
akanshi-elastic: We are setting the value of `token_expires_in` here because every token has a expiry time which needs to replaced every time token expires with the newer expiry time.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+    async def _set_access_token(self):
+        """Set access token using configuration fields"""
+        if not is_expired(expires_at=self.token_expires_in):
+            return
+        tenant_id = self.configuration["tenant_id"]
+        logger.info("Generating access token...")
+        url = f"https://accounts.accesscontrol.windows.net/{tenant_id}/tokens/OAuth/2"
+        data = {
+            "grant_type": "client_credentials",
+            "resource": f"00000003-0000-0ff1-ce00-000000000000/{self.configuration['tenant']}.sharepoint.com@{tenant_id}",
```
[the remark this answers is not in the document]
akanshi-elastic: No it is required to generate the token for sharepoint online. attaching the screenshot where this thing is mentioned.
Screenshot (57)

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
akanshi-elastic: In this case we will be end up calling retry for try block also, which we don't need. I agree that we can convert existing except block contains if-else to multiple exception but again we have to duplicate same code of retry in multiple except block as well as for other exception which we didn't handle.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+                    self.session = await self._generate_session()
+                retry += 1
+                if retry > self.retry_count:
+                    raise exception
+                logger.warning(
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                )
+                await asyncio.sleep(RETRY_INTERVAL**retry)
+
+    def format_document(
```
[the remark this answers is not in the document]
akanshi-elastic: Can you please elaborate more on this?

connectors/sources/sharepoint.py
artem-shelkovnikov: It's displayed in UI and most of the time it does not help - users will just delete them. Some of these values make sense as reasonable defaults, but some don't IMO - such as `password`, `client_id`, `client_secret`, `tenant`, `site_collections`. We also set `ssl_disabled=True` for tests but for real users we probably should default it to False instead.

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
artem-shelkovnikov: I've read the post in StackOverflow, but didn't understand how it's relevant for us. Can you elaborate in details?

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
artem-shelkovnikov: Can you add a comment about it to the connector?

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
artem-shelkovnikov: This function has multiple `if` statements inside - it's important to verify that you test the output of each `if` and `else` branch meaningfully - e.g. on test for type = list_item, one for drive_item, one for document_library, etc.

connectors/sources/sharepoint.py
akanshi-elastic: No it is not possible to make it a utility function because we are calling `_api_call` method on line 48, which can create issue(unclosed session).

connectors/sources/sharepoint.py
akanshi-elastic: yes it is mentioned here:  https://alextofan.com/2019/08/22/how-to-get-more-than-5000-item-from-sharepoint-online-in-flow/

connectors/sources/sharepoint.py
artem-shelkovnikov: I think you can delegate it to the function by providing the function that returns chunks as one of parameters for this function, something like this (name is purely for example):

Does it make sense? cc @tarekziade 

```python
async def download_and_stream(async_batch_generator):
    async with NamedTemporaryFile(mode="wb", delete=False) as async_buffer:
            async for data in async_batch_generator():
                await async_buffer.write(data)
            source_file_name = async_buffer.name

        await asyncio.to_thread(
            convert_to_b64,
            source=source_file_name,
        )
        async with aiofiles.open(file=source_file_name, mode="r") as target_file:
            # base64 on macOS will add a EOL, so we strip() here
            attachment_content = (await target_file.read()).strip()
        await remove(source_file_name)
```

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
akanshi-elastic: we can add skew when SharePoint service span at multiple servers and at different locations. But we can also remove this change as it will not make much difference. I added this just to remove the skew and handle the failed request normally.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+            field for field in connection_fields if self.configuration[field] == ""
+        ]:
+            raise Exception(
+                f"Configured keys: {empty_connection_fields} can't be empty."
+            )
+        if not self.ssl_disabled and self.certificate == "":
+            raise Exception("Configured key ssl_ca can't be empty when SSL is enabled")
+
+    @retryable(
+        retries=RETRIES,
```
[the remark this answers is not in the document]
akanshi-elastic: Do we need to manage it via global variable or class variable?

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
akanshi-elastic: Since we have already wrote test cases for drive item, list item and sites and also all the lines for this method is already covered. Do you want us to write specific test case for lists and doument library?

connectors/sources/sharepoint.py
danajuratoni: Defaults should not be required for fields. I agree it would be good to set reasonable defaults especially for fields such as `retry_count`, but requiring users to delete `password` before being able to enter their own, is not a good user experience.

connectors/sources/tests/fixtures/sharepoint/fixture.py
wangch079: If we fake a sharepoint deployment, it's equivalent to unit with mocked api calls. 

Have you considered using a long-running sharepoint deployment (we have one deployment for Workplace Search test), and grab credentials from vaults?

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+                if isinstance(
+                    exception, aiohttp.client_exceptions.ClientResponseError
+                ) and "token has expired" in exception.headers.get(
+                    "x-ms-diagnostics", ""
+                ):
+                    await self._set_access_token()
+                elif isinstance(
+                    exception, aiohttp.client_exceptions.ServerDisconnectedError
+                ):
+                    await self.close()
```
wangch079, line 366: `self.close()` should be called by the framework.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+            raise_for_status=True,
+        )
+
+    async def close(self):
+        """Closes unclosed client session"""
+        if self.session is None:
+            return
+        await self.session.close()
+
+    async def _api_call(self, url_name, url="", **url_kwargs):
```
wangch079, line 325: Can `@retryable` decorator be used here?

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+                    "x-ms-diagnostics", ""
+                ):
+                    await self._set_access_token()
+                elif isinstance(
+                    exception, aiohttp.client_exceptions.ServerDisconnectedError
+                ):
+                    await self.close()
+                    self.session = await self._generate_session()
+                retry += 1
+                if retry > self.retry_count:
```
wangch079, line 369:
```suggestion
                if retry >= self.retry_count:
```

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+            document_type(string): Type of document(i.e. list_item and drive_item).
+            item_type(string, optional): Type of item i.e. File or Folder. Defaults to None.
+            file_name(string, optional): Name of file. Defaults to None.
+
+        Returns:
+            dictionary: Modified document with the help of adapter schema.
+        """
+        document = {"type": document_type}
+
+        if document_type in [LISTS, DOCUMENT_LIBRARY]:
```
wangch079, line 396: Consider using the match and case statement: https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,715 @@
...
+        self, document, file_relative_url, site_url, timestamp=None, doit=None
+    ):
+        """Get content of list items and drive items
+
+        Args:
+            document (dictionary): Modified document.
+            file_relative_url (str): Relative url of file
+            site_url (str): Site path of sharepoint
+            timestamp (timestamp, optional): Timestamp of item last modified. Defaults to None.
+            doit (boolean, optional): Boolean value for whether to get content or not. Defaults to None.
```
wangch079, line 462: Why not default to `False`

connectors/sources/sharepoint.py
akanshi-elastic: Here to handle ServerDisconnectedError error we are closing the session.

connectors/sources/sharepoint.py
akanshi-elastic: We need to handle different exceptions here, that's why we can't use the decorator here.

connectors/sources/sharepoint.py
akanshi-elastic: We are starting `retry` counter from 0 so if we add `if retry >= self.retry_count:` this condition here , it will only look for 2 retries and then raise an exception.

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
akanshi-elastic: To make it work with configuration `retry_count` , there will be some tweaks in `retryable` decorator. The change will bound the users of `retryable` decorator to have  `self.retry_count` as their instance variable.
```python
def retryable(interval=1.0, strategy=RetryStrategy.LINEAR_BACKOFF):
    def wrapper(func):
        @functools.wraps(func)
        async def func_to_execute(self, *args, **kwargs):
            retry = 1
            while retry <= self.retry_count:
                try:
                except Exception as e:
                    if retry >= self.retry_count:
```
To make `retryable`  more generic we can have check for `retries` and if it is `None` then we can use class variable i.e. `self.retry_count`
@timgrein what's your opinion on this?

connectors/sources/sharepoint.py
akanshi-elastic: if we will go with match case then we can't use constant. We will have to use hard coded values
```python
match document_type:
    case "lists" | "document_library":
        document["url"] = urljoin(
            self.host_url, item["RootFolder"]["ServerRelativeUrl"]
        )
    case "drive_item":
        document.update(
            {
                "_id": item["GUID"],
                "size": item.get("File", {}).get("Length"),
                "url": urljoin(self.host_url, item[item_type]["ServerRelativeUrl"]),
                "type": item_type,
            }
        )
    case "list_item":
        document.update(
            {
                "_id": item["_id"] if "_id" in item.keys() else item["GUID"],
                "file_name": file_name,
                "size": item.get("size"),
                "url": item["url"],
            }
                )
```

connectors/sources/sharepoint.py
wangch079: Can we just do this instead?
```suggestion
                    await self.session.close()
```

It's possible that there are other clean up logic added to `close()`.

connectors/sources/sharepoint.py
wangch079: That depends on how you define `retry_count`. Is it the maximum number of requests you will make or the number of **RE**try you will make?

If `retry_count=3`, and the remote server is not available, how many requests do you expect it to make? 3 or 4?

If it's 3, `if retry >= self.retry_count:` will work.

If it's 4, we should use `>`, but the final log will be 
>Retry count: 4 out of 3. Exception: blah blah

akanshi-elastic:
@tarekziade @wangch079 I have updated the PR comments

connectors/sources/sharepoint.py
vidok: @danajuratoni did I get it right that we need to ask @akanshi-crest to remove some of the default fields' values?

connectors/utils.py
```diff
@@ -399,3 +401,60 @@ async def func_to_execute(*args, **kwargs):
...
+
+def get_expires_at(expires_in):
+    """Adds seconds to the current utc time.
+
+    Args:
+        expires_in (int): Seconds after which the token will expire.
+    """
+    expires_at = datetime.utcnow() + timedelta(seconds=expires_in)
+    # account for clock skew
+    expires_at -= timedelta(seconds=20)
```
vidok, line 447: How did you come up with 20 seconds? Is this something that SharePoint recommends? Have you ever encountered a clock skew issue while working with Sharepoint?

connectors/utils.py
```diff
@@ -399,3 +401,60 @@ async def func_to_execute(*args, **kwargs):
...
+        original_string(string): String containing special characters
+
+    Returns:
+        encoded_string(string): Parsed string without single quotes
+    """
+    encoded_string = urllib.parse.quote(original_string, safe="'")
+    return encoded_string.replace("'", "''")
+
+
+def get_expires_at(expires_in):
```
vidok, line 439: I am concerned that this method should be in utils. Is it gonna be used somewhere else? 

a few other points: 

* The method name is misleading. What the method does is calculate a time delta. 
* You refer to `tokens` in doc strings but the method's signature knows nothing about tokens

connectors/utils.py
```diff
@@ -399,3 +401,60 @@ async def func_to_execute(*args, **kwargs):
...
+
+
+def is_expired(expires_at):
+    """Compares the given time with present time
+
+    Args:
+        expires_at (datetime): Time of token expires.
+    """
+    # Recreate in case there's no expires_at present
+    if expires_at is None:
```
vidok, line 458: This might lead to unexpected issues (I'm assuming all engineers can use this method because it's in utils) because when I pass `None` I would expect an exception rather than a positive outcome. 

It seems like the method signature is tight to the Sharepoint connector (see comment on line 457) so it's not a good idea to keep the connector's specific logic in utils.

connectors/sources/sharepoint.py
vidok: @tarekziade ping

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,704 @@
...
+            exception: An instance of an exception class.
+
+        Yields:
+            data: API response.
+        """
+        retry = 0
+        # If pagination happens for list and drive items then next pagination url comes in response which will be passed in url field.
+        if url == "":
+            url = URLS[url_name].format(**url_kwargs)
+        headers = None
```
vidok, line 312: I think an empty line is missing here.

connectors/sources/sharepoint.py
```diff
@@ -0,0 +1,704 @@
...
+        Raises:
+            exception: An instance of an exception class.
+
+        Yields:
+            data: API response.
+        """
+        retry = 0
+        # If pagination happens for list and drive items then next pagination url comes in response which will be passed in url field.
+        if url == "":
+            url = URLS[url_name].format(**url_kwargs)
```
vidok, line 311: an empty line would improve the code readability

connectors/sources/sharepoint.py
danajuratoni: Yes, let's remove defaults that are not useful. We'll get a better user experience via rich configurable fields coming in 8.8 where we will handle adding placeholders, tooltips, etc.

danajuratoni: By "useful defaults" I mean those for fields used to fine tune a connector such as `retry_count` or `fetch_size` or disabled SSL for example.

akanshi-elastic: I have removed the non useful defaults from configurable fields such as `client_id`, `secret_id`, `tenant` and `tenant_id` but `username` and `password` values are required and we can't remove it as it is used to run e2e tests.

connectors/utils.py
```diff
@@ -399,3 +401,60 @@ async def func_to_execute(*args, **kwargs):
...
+    replaces single quote with two single quote since quote
+    is treated as an escape character
+
+    Args:
+        original_string(string): String containing special characters
+
+    Returns:
+        encoded_string(string): Parsed string without single quotes
+    """
+    encoded_string = urllib.parse.quote(original_string, safe="'")
```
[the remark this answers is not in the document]
akanshi-elastic: Good catch,  I think we can update this.

connectors/utils.py
akanshi-elastic: > I am concerned that this method should be in utils. Is it gonna be used somewhere else?

I got a comment to add it in utils.

connectors/utils.py
akanshi-elastic: I got a comment to add it in utils.
 https://github.com/elastic/connectors-python/pull/537#discussion_r1121183957

connectors/utils.py
akanshi-elastic: > How did you come up with 20 seconds?

Assuming that the servers are in sync with NIST time(https://www.nist.gov/pml/time-and-frequency-division/time-distribution/internet-time-service-its), then the difference should be negligible at around a couple of seconds only. keeping this in mind I chosed 20s

> Is this something that SharePoint recommends?Have you ever encountered a clock skew issue while working with Sharepoint?
   
No it's not receommend by sharepoint but we did this just for precautionary purpose. Suppose we have two auth servers.How do you know that they will have the same time? What if they are in fact off by a few minutes. One server will return that the token has expired and the other won't.
we don't need it it to be required here but if you have two auth servers then we can face clock skew issue.

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
akanshi-elastic: we are planning to add a client call which will take care of token generation as well as the API call just like we did in confluence. but since all the testing are carried on this architecture, it will be better if we can merge this PR meanwhile we have already started that enhancement and we would be raising and incremental the PR once this gets merged.

timgrein (approved):
LGTM, I'll create follow-up issues to encapsulate the API interaction in a client class/memoizing the instance.

PR #552, opened 2023-02-24
Introduce "validate_configuration" method to BaseSource and use it

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/4023

We want to be able to validate connector configuration meaningfully on the level of the framework - not as a part of other methods like `ping` or `get_docs`.

This PR introduces new method `validate_configuration` to BaseSource that does nothing when called. Individual sources can override it to validate configuration and raise errors if configuration was invalid. If configuration was invalid and exception was raised, framework will just handle this error as any other errors - it will update connector status to `ERROR` and display error message coming from `validate_configuration` in UI.

It would have been a large change to implement for all connectors, so this PR only implements the change for MySQL connector. Later I'm planning to follow-up with this change for all other connectors and make `validate_configuration` raise an error if not implemented.

Thinking out loud - should we provide an error class `ValidationError` and ask raising it in `validate_configuration`? I think it's a good idea, since any other errors can happen in the code - typos, bad error handling - and we want to separate them from Validation errors.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -698,7 +698,6 @@ class Connector(ESDocument):
 
             logger.debug(f"Validating configuration for {data_provider}")
             await data_provider.validate_configuration()
-            await asyncio.sleep(0)
 
             logger.debug(f"Pinging the {data_provider} backend")
             await data_provider.ping()
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -697,7 +697,7 @@ class Connector(ESDocument):
                 return
 
             logger.debug(f"Validating configuration for {data_provider}")
-            await data_provider.validate_configuration()
+            await data_provider.validate_config()
 
             logger.debug(f"Pinging the {data_provider} backend")
             await data_provider.ping()
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -700,6 +700,10 @@ class Connector(ESDocument):
             await data_provider.ping()
             await asyncio.sleep(0)
 
+            logger.debug(f"Validating configuration for {data_provider}")
+            await data_provider.validate_configuration()
+            await asyncio.sleep(0)
+
             mappings = Mappings.default_text_fields_mappings(
                 is_connectors_index=True,
             )
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -696,12 +696,12 @@ class Connector(ESDocument):
                 await self._sync_done(job, JobStatus.COMPLETED, {}, start_time)
                 return
 
-            logger.debug(f"Pinging the {data_provider} backend")
-            await data_provider.ping()
-            await asyncio.sleep(0)
-
             logger.debug(f"Validating configuration for {data_provider}")
             await data_provider.validate_configuration()
+            await asyncio.sleep(0)
+
+            logger.debug(f"Pinging the {data_provider} backend")
+            await data_provider.ping()
             await asyncio.sleep(0)
 
             mappings = Mappings.default_text_fields_mappings(
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -186,13 +186,13 @@ class BaseDataSource:
         """
         return True
 
-    async def validate_configuration(self):
+    async def validate_config(self):
         """When called, validates configuration of the connector that is contained in self.configuration
 
         If connector configuration is invalid, this method will raise an exception
         with human-friendly and actionable description
         """
-        # TODO: when validate_configuration is implemented everywhere, we should make this method raise NotImplementedError
+        # TODO: when validate_config is implemented everywhere, we should make this method raise NotImplementedError
         pass
 
     async def ping(self):
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -186,6 +186,15 @@ class BaseDataSource:
         """
         return True
 
+    async def validate_configuration(self):
+        """When called, validates configuration of the connector that is contained in self.configuration
+
+        If connector configuration is invalid, this method will raise an exception
+        with human-friendly and actionable description
+        """
+        # TODO: when validate_configuration is implemented everywhere, we should make this method raise NotImplementedError
+        pass
+
     async def ping(self):
         """When called, pings the backend
 
--- a/connectors/tests/fake_sources.py
+++ b/connectors/tests/fake_sources.py
@@ -26,6 +26,7 @@ class FakeSource:
         if configuration.has_field("raise"):
             raise Exception("I break on init")
         self.fail = configuration.has_field("fail")
+        self.configuration_invalid = configuration.has_field("configuration_invalid")
 
     async def changed(self):
         return True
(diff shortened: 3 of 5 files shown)


artem-shelkovnikov:
Additionally found problems:

Services don't handle errors happening in the loop very well. If something goes wrong (like validation error happens), sync service crashes. Instead, it should just swallow these exceptions as Ruby does - we expect random errors happen every now and then in the connectors, for example sources being not available, fields being invalid, etc.

We'll need to follow up on this in a different PR.

connectors/byoc.py
```diff
@@ -700,6 +700,10 @@ async def sync(self, sync_job_index, source_klass, elastic_server, bulk_options)
             await data_provider.ping()
             await asyncio.sleep(0)
 
+            logger.debug(f"Validating configuration for {data_provider}")
```
timgrein, line 703: What will the output of `data_provider` look like in that log message? Do we've access to the connector id and the service type maybe? Would probably make debugging easier later on as we can distinguish, which configuration was validated.

timgrein (dismissed):
nit around a log message, otherwise LGTM

timgrein:
And I like the idea of having `ValidationError`, but we may want to name it `ConfigurationValidationError` (maybe even `DataSourceConfiguration(Validation)Error`?) to clearly distinguish from `FilterValidationError`.

connectors/byoc.py
```diff
@@ -700,6 +700,10 @@ async def sync(self, sync_job_index, source_klass, elastic_server, bulk_options)
...
+            await data_provider.validate_configuration()
+            await asyncio.sleep(0)
```
tarekziade, line 705: you can remove this `await asyncio.sleep(0)`

connectors/tests/fake_sources.py
```diff
@@ -57,6 +58,11 @@ async def validate_filtering(cls, filtering):
             state=FilteringValidationState.VALID, errors=[]
         )
 
+    async def validate_configuration(self):
+        if self.configuration_invalid:
+            raise ValueError("I fail when validating configuration")
+        pass
```
tarekziade, line 64: remove this `pass` (only to be used when a function is empty)

tarekziade:
oh just one more thing, what about `validate_config` in pursue of shorter names

connectors/byoc.py
```diff
@@ -700,6 +700,10 @@ async def sync(self, sync_job_index, source_klass, elastic_server, bulk_options)
             await data_provider.ping()
             await asyncio.sleep(0)
 
+            logger.debug(f"Validating configuration for {data_provider}")
+            await data_provider.validate_configuration()
```
praveen-elastic, line 704: The patch looks good,. I just have one doubt. Should we validate the configurations before this [line](https://github.com/elastic/connectors-python/pull/552/files#diff-b9a5d063adca223525457d0a3c06df668bd814fe46069dda6678b694c277edc9R700), i.e. before the ping ? 
The source class should be sure of the validity of the configurations before invoking the ping IMO.

connectors/source.py
```diff
@@ -186,6 +186,15 @@ async def changed(self):
...
         return True
 
+    async def validate_configuration(self):
+        """When called, validates configuration of the connector that is contained in self.configuration
+
+        If connector configuration is invalid, this method will raise an exception
+        with human-friendly and actionable description
+        """
+        # TODO: when validate_configuration is implemented everywhere, we should make this method raise NotImplementedError
+        pass
```
wangch079, line 196: Should this method do some basic validation, e.g. presence of mandatory fields, type match.

wangch079: This can also be moved here:
https://github.com/elastic/connectors-python/blob/fbceb616d446c314eb28a48130a3a836662bce2f/connectors/source.py#L118-L121

wangch079 (commented):
Should `validate_configuration` be called in `__init__` of `BaseDataSource`?

connectors/byoc.py
artem-shelkovnikov: Example log line:

```
[FMWK][11:50:15][DEBUG] Validating configuration for Datasource `MySQL`
```

As for the connector id and service type - I believe it's something that needs to be addressed in a logging epic. Log lines should work depending on context - if log happens inside connector, we need to output connector id/type, if inside job - also job id.

connectors/byoc.py
artem-shelkovnikov: Good call, swapped in https://github.com/elastic/connectors-python/pull/552/commits/3eb88f4c21ff9e59fecd63040316706018e64449

connectors/byoc.py
artem-shelkovnikov: Removed in https://github.com/elastic/connectors-python/pull/552/commits/8b3ad71119e47b86569aa95aa03568f981c0d4eb

connectors/source.py
artem-shelkovnikov: I'm not sure about it - inheritance model here makes it easy to actually break the implementation. I'll need to do refactoring for this, which will make this PR bigger than it should.

What I mean - if validation logic is inside `BaseSource.validate_configuration`, then to do it connector class should call the base class implementation with `super.validate_configuration()`. It's quite easy not to do this and lose validation of types and such. So we'd have to probably break validation code into 2 methods:

```
class BaseSource
    async def validate_configutarion(self):
        # do generic validation for configuration 
        self.validate_source_specific_configuration() # then do validation of source-specific stuff.
```

connectors/tests/fake_sources.py
artem-shelkovnikov: Ahhh, missed it - removed at https://github.com/elastic/connectors-python/pull/552/commits/dba6430863e5abd7de106cc113e6b9ba75c40d46

artem-shelkovnikov:
> Should validate_configuration be called in __init__ of BaseDataSource?

I'm not sure about it - I honestly don't see a lot of value in `__init__` of BaseDataSource and it's easy to make a mistake to not do `super().__init__()` in source classes.

Also some configuration validation might require network calls, making constructor have side-effects and calls to 3rd-party services which is IMO not a good thing.


> oh just one more thing, what about validate_config in pursue of shorter names

Done!

> And I like the idea of having ValidationError, but we may want to name it ConfigurationValidationError (maybe even DataSourceConfiguration(Validation)Error?) to clearly distinguish from FilterValidationError.

Makes a lot of sense to me!

I think this PR should be merged as is - it unblocks work for Crest team - and we can improve on validation logic in subsequent PRs.

tarekziade:
related: https://github.com/elastic/connectors-python/pull/560#pullrequestreview-1316428827

wangch079:
> related: [#560 (review)](https://github.com/elastic/connectors-python/pull/560#pullrequestreview-1316428827)

Agree with this. This is the basic validation I was talking about. In the configuration, we should specify `mandatory` (similar to `empty_allowed`) and `type` (already in place), and validate in `BaseDataSource` (Same validation for all sources and does not require remote connection). 

But we can leave it to a separate issue.

wangch079 (approved):
LGTM except a small comment, which can be addressed separately

artem-shelkovnikov:
> LGTM except a small comment, which can be addressed separately

@wangch079 which one specifically? About more extensive basic validation?

wangch079:
> @wangch079 which one specifically? About more extensive basic validation?

This: https://github.com/elastic/connectors-python/pull/552#issuecomment-1447653864

I think we can leave it first, as I think it will be overlapping with the Rich Configurable Fields project. I also see @timgrein closed his PR.

PR #555, opened 2023-02-27
added pyright and fixed a few issues

tarekziade:
`pyright` is Microsoft's static type checker (MIT)

It detected two major bugs in our codebase (yay):

- a serializing error in the Mongo connector
- a coroutine that was never awaited in `byoc.py`
 
This patch fixes them and ignore a few warnings. It's now plugged into `make lint`

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -412,7 +412,7 @@ class Connector(ESDocument):
     def last_seen(self):
         last_seen = self.get("last_seen")
         if last_seen is not None:
-            last_seen = datetime.fromisoformat(last_seen)
+            last_seen = datetime.fromisoformat(last_seen)  # pyright: ignore
         return last_seen
 
     @property
@@ -693,7 +693,7 @@ class Connector(ESDocument):
                 logger.debug(
                     f"No change in {self.service_type} data provider, skipping..."
                 )
-                self._sync_done(job, JobStatus.COMPLETED, {}, start_time)
+                await self._sync_done(job, JobStatus.COMPLETED, {}, start_time)
                 return
 
             logger.debug(f"Pinging the {data_provider} backend")
--- a/connectors/sources/generic_database.py
+++ b/connectors/sources/generic_database.py
@@ -152,6 +152,9 @@ class GenericBaseDataSource(BaseDataSource):
         Yields:
             list: Column names and query response
         """
+        if self.queries is None:
+            raise NotImplementedError()
+
         query = self.queries[query_name].format(**query_kwargs)
         size = self.configuration["fetch_size"]
 
@@ -216,6 +219,8 @@ class GenericBaseDataSource(BaseDataSource):
         Returns:
             cursor: Asynchronous cursor
         """
+        if self.engine is None:
+            raise NotImplementedError()
         try:
             async with self.engine.connect() as connection:
                 cursor = await connection.execute(text(query))
@@ -241,6 +246,8 @@ class GenericBaseDataSource(BaseDataSource):
         Returns:
             cursor: Synchronous cursor
         """
+        if self.engine is None:
+            raise NotImplementedError()
         try:
             loop = asyncio.get_running_loop()
             self.connection = await loop.run_in_executor(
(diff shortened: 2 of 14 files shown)


connectors/sources/s3.py
```diff
@@ -183,7 +185,7 @@ async def get_docs(self, filtering=None):
             dictionary: Document from Amazon S3.
         """
         bucket_list = self.buckets if self.buckets != ["*"] else self.get_bucket_list()
-        page_size = int(self.configuration.get("page_size", DEFAULT_PAGE_SIZE))
+        page_size = self.configuration["page_size"]
```
wangch079, line 188: Not quite sure about this change, it seems it remove the usage of `DEFAULT_PAGE_SIZE`

connectors/sources/mysql.py
```diff
@@ -243,8 +243,7 @@ async def _connect(self, query, fetch_many=False, **query_kwargs):
 
         query_kwargs["database"] = self.database
         formatted_query = query.format(**query_kwargs)
-        size = int(self.configuration.get("fetch_size", DEFAULT_FETCH_SIZE))
-
+        size = self.configuration["fetch_size"]
```
wangch079, line 246: It removes `DEFAULT_FETCH_SIZE`. What if there's no key `fetch_size` in `self.configuration`

connectors/sources/generic_database.py
```diff
@@ -152,6 +152,9 @@ async def execute_query(self, query_name, fetch_many=False, **query_kwargs):
         Yields:
             list: Column names and query response
         """
+        if self.queries is None:
+            raise NotImplementedError()
```
wangch079, line 156: It makes sense to raise error if `self.queries` is None, but `NotImplementedError`?

connectors/byoc.py
```diff
@@ -693,7 +693,7 @@ async def sync(self, sync_job_index, source_klass, elastic_server, bulk_options)
                 logger.debug(
                     f"No change in {self.service_type} data provider, skipping..."
                 )
-                self._sync_done(job, JobStatus.COMPLETED, {}, start_time)
+                await self._sync_done(job, JobStatus.COMPLETED, {}, start_time)
```
wangch079, line 696: Good catch

wangch079 (commented):
> - a coroutine that was never awaited in byoc.py

This is extremely useful.

connectors/sources/mongo.py
```diff
@@ -73,7 +73,7 @@ def serialize(self, doc):
         def _serialize(value):
             if isinstance(value, ObjectId):
                 value = str(value)
-            if isinstance(value, (list, tuple)):
+            elif isinstance(value, (list, tuple)):
```
artem-shelkovnikov, line 76: Oh wow

artem-shelkovnikov (commented):
I've seen several "# pyright: ignore" added - what about them? Is the rule okay but bad for these lines? Are we planning to address these later?

connectors/sources/s3.py
tarekziade: no, it's still used but done directly in the base class now on init, this code is now useless.
https://github.com/elastic/connectors-python/blob/main/connectors/source.py#L124

connectors/sources/generic_database.py
tarekziade: `self.queries` is set to None in this base class. If the concrete class does not set it and this function is called, it won't work. So in that case it means the concrete class is incomplete. And for incomplete concrete class, you use `NotImplementedError`

tarekziade:
> I've seen several "# pyright: ignore" added - what about them? Is the rule okay but bad for these lines? Are we planning to address these later?

It happens when we use things not as intended. When we derive `asyncio.Queue` we use `_members` that are marked as being private to that class in the typing definitions in the stdlib.  And it's a bit of a grey area: should we rewrite the whole class? or use composition? To address this, we would need to rewrite that class in particular

artem-shelkovnikov (approved):

wangch079 (approved):

PR #559, opened 2023-02-27
[PostgreSQL] Allow only one schema to be configured

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4056

As a schema in PostgreSQL acts like a logical grouping for database objects (including tables) we need to enforce the configuration of one schema to keep it conistent with other SQL-like connectors.

#### Pre-Review Checklist
~- [ ] Considered corresponding documentation changes~ (separate documentation PR)
~- [ ] Contributed any configuration settings changes to the configuration reference~ (separate PR)


connectors/sources/postgresql.py
```diff
@@ -11,8 +11,6 @@
 
 from connectors.sources.generic_database import GenericBaseDataSource
 
-# Below schemas are system schemas and the tables of the systems schema's will not get indexed
-SYSTEM_SCHEMA = ["pg_toast", "pg_catalog", "information_schema"]
```
artem-shelkovnikov, line 15: Probably good to disallow these schemas from extraction still, what do you think?

artem-shelkovnikov (dismissed):
Looks good, but left one question

connectors/sources/postgresql.py
timgrein: Maybe a customer wants to ingest those for reasons we (or at least I) don't know. Therefore I've just removed this assumption that it's _never_ ingested from the code as this needs to be explicitly configured anyway. I think we can easily put it in again, if there's a need. WDYT?

artem-shelkovnikov: Yeah I don't have a really strong stance. I can see export of metrics from `information_schema` can be useful, but not in our use-case - they could work well if we appended them instead of overwriting data in the index to have time-series data over performance.

So I doubt anyone will sync those, but if we don't validate this it should not be a problem.

tarekziade: It's good defaults though -- I don't think anyone would ever want to index these, can't think of a use case, it's internal PG data. Maybe this connector could have an option with a list of tables to ignore, with these 3 by default.

connectors/sources/postgresql.py
```diff
@@ -57,6 +55,7 @@ def get_default_configuration(cls):
         postgresql_configuration = super().get_default_configuration().copy()
         postgresql_configuration.update(
             {
+                "schema": {"value": "public", "label": "Schema", "type": "str"},
```
moxarth-elastic, line 58: you might need to add a `schema` parameter in validation_configuration as well.

seanstory:
doing some pr scanning - does this need to backport to 8.7?

timgrein:
This PR is currently on hold as this functionality is now considered a stretch for 8.8.

timgrein:
Need to update the FTest accordingly to include the schema config, will do it today or tomorrow.

connectors/sources/postgresql.py
```diff
@@ -436,20 +439,20 @@ async def get_docs(self, filtering=None):
...
-                        schema=schema,
-                    ):
-                        yield row, None
-                if table_count < 1:
-                    self._logger.warning(
-                        f"Fetched 0 tables for schema: {schema} and database: {self.database}"
-                    )
+        table_count = 0
+
+        async for table in self.postgresql_client.get_tables_to_fetch(self.schema):
```
wangch079, line 444: Probably we should consider passing `schema` (or simpler, `configuration`) to `PostgreSQLClient` initializer.

wangch079 (commented):
You should also remove `get_all_schemas` from `PostgreSQLClient`, and `all_schemas` from `Queries` class and its sub-classes.

connectors/sources/postgresql.py
navarone-feekery: I've removed `get_all_schemas` but I wasn't sure on the impact of removing `all_schemas` from `Queries` as other connectors seem dependent on it, so I've just removed the content of the func from the `PostgreSQLQueries` class.

navarone-feekery: Also passed `schema` on to `PostgreSQLClient` so it can be initialized there.

vidok (approved):

PR #560, opened 2023-02-27
[Generic Database Connector] Make validate_configuration a top-level function and use it for composition

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/4067

**\~ 80% of LOC is due to parametrized testing**

- Created a new function `validate_non_empty_config_fields`, which can be used by arbitrary sources to validate, that mandatory configuration fields are non-empty
- Created a new function `validate_db_connection_fields`, which specifies validation for common database connection fields (`host`, `port`, `user`, `password`)
- Remove the datasource specific checks from the `generic_database` module and implement them at the right places in `oracle.py` and `postgresql.py`
- Mark two places with TODOs, which are currently depending on open PRs (will be removed by me as soon as the PRs are merged)
- Add error classes specific for the validation of a configuration:
  - `ConfigurationFieldMissingError`
  - `ConfigurationFieldEmptyError`
  - `ConfigurationFieldTypeError`
  - `ConfigurationValueError`

#### Pre-Review Checklist

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -82,12 +82,7 @@ class Field:
         elif type_ == "bool":
             return value.lower() in ("y", "yes", "true", "1")
         elif type_ == "list":
-            return list(
-                filter(
-                    lambda item: len(item) > 0,
-                    [item.strip() for item in value.split(",")],
-                )
-            )
+            return [item.strip() for item in value.split(",") if len(item.strip()) > 0]
         return value
 
 
@@ -133,7 +128,7 @@ class DataSourceConfiguration:
         if any(not self.has_field(name) for name in names):
             raise ConfigurationFieldMissingError
 
-        empty_fields = list(filter(lambda name: self._config[name].is_empty(), names))
+        empty_fields = [name for name in names if self._config[name].is_empty()]
 
         return len(empty_fields) > 0, empty_fields
 
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -20,6 +20,31 @@ from connectors.filtering.validation import (
 )
 
 
+class ConfigurationFieldMissingError(ValueError):
+    pass
+
+
+class ConfigurationFieldEmptyError(ValueError):
+    pass
+
+
+class ConfigurationFieldTypeError(ValueError):
+    pass
+
+
+class ConfigurationValueError(ValueError):
+    pass
+
+
+def validate_non_empty_config_fields(expected_fields, configuration):
+    any_field_is_empty, empty_fields = configuration.any_field_is_empty(expected_fields)
+
+    if any_field_is_empty:
+        raise ConfigurationFieldEmptyError(
+            f"Configurable fields '{empty_fields}' can't be empty."
+        )
+
+
 class Field:
     def __init__(self, name, label=None, value="", type="str"):
         if label is None:
@@ -38,6 +63,13 @@ class Field:
         self._type = value
         self.value = self._convert(self.value, self._type)
 
+    def is_empty(self):
+        return (
+            self.value is None
+            or (isinstance(self.value, list) and len(self.value) == 0)
+            or len(str(self.value)) == 0
+        )
+
     def _convert(self, value, type_):
         if not isinstance(value, str):
             # we won't convert the value if it's not a str
@@ -50,7 +82,12 @@ class Field:
         elif type_ == "bool":
             return value.lower() in ("y", "yes", "true", "1")
         elif type_ == "list":
-            return [item.strip() for item in value.split(",")]
+            return list(
+                filter(
+                    lambda item: len(item) > 0,
+                    [item.strip() for item in value.split(",")],
+                )
+            )
         return value
 
 
@@ -91,6 +128,14 @@ class DataSourceConfiguration:
 
     def has_field(self, name):
         return name in self._config
+
+    def any_field_is_empty(self, names):
+        if any(not self.has_field(name) for name in names):
+            raise ConfigurationFieldMissingError
+
+        empty_fields = list(filter(lambda name: self._config[name].is_empty(), names))
+
+        return len(empty_fields) > 0, empty_fields
 
     def set_field(self, name, label=None, value="", type="str"):
         self._config[name] = Field(name, label, value, type)
--- a/connectors/sources/generic_database.py
+++ b/connectors/sources/generic_database.py
@@ -10,7 +10,11 @@ from asyncpg.exceptions._base import InternalClientError
 from sqlalchemy import text
 
 from connectors.logger import logger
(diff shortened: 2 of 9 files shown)


seanstory (approved):
great tests, as always. 👏

connectors/source.py
```diff
@@ -20,6 +20,31 @@
...
+
+class ConfigurationFieldTypeError(ValueError):
+    pass
+
+
+class ConfigurationValueError(ValueError):
+    pass
+
+
+def validate_non_empty_config_fields(expected_fields, configuration):
```
tarekziade, line 39: that's a very long name

tarekziade (commented):
Looks good but I feel like we're doing a lot of code where we could be more declarative and generic.

One idea: the configuration is defined by this function here 

https://github.com/elastic/connectors-python/blob/main/connectors/sources/generic_database.py#L67-L114

Maybe could add a new flag there to facilitate validation.

so host for instance, becomes:

```python
"host": {
  "value": "127.0.0.1",
  "label": "Host",
  "type": "str",
  "empty_allowed:": False
},
```

if you implement it this we, we can avoid adding new code for each specific validation sets (databases, etc)

connectors/source.py
```diff
@@ -50,7 +82,12 @@ def _convert(self, value, type_):
         elif type_ == "bool":
             return value.lower() in ("y", "yes", "true", "1")
         elif type_ == "list":
-            return [item.strip() for item in value.split(",")]
+            return list(
```
tarekziade, line 85: you can simplify this 

```
return [item.strip() for item in value.split(",") if len(item.strip()) > 0]
```

timgrein:
@tarekziade Thank you for the list comprehension suggestions, added them!

> One idea: the configuration is defined by this function here
> https://github.com/elastic/connectors-python/blob/main/connectors/sources/generic_database.py#L67-L114
> Maybe could add a new flag there to facilitate validation.

Makes sense, we could translate the configuration to a schema, which is understood by `fastjsonschema`. I would prefer to merge this now as is and check first, what's the scope and the anticipated changes of rich-configurable fields for the python backend. I'll create an issue for it!

connectors/sources/generic_database.py
```diff
@@ -36,6 +40,33 @@ def table_filter(table):
...
+    db_server_connection_fields = ["host", "port", "user", "password"]
+
+    validate_non_empty_config_fields(db_server_connection_fields, configuration)
+
+    port = configuration["port"]
+
+    if isinstance(port, str) and not port.isnumeric():
+        raise ConfigurationFieldTypeError(
+            f"Configured port {port} has to be an integer."
+        )
```
wangch079, line 67: I feel all these can be moved to `basic_validation`.

artem-shelkovnikov:
> Looks good but I feel like we're doing a lot of code where we could be more declarative and generic.
> 
> One idea: the configuration is defined by this function here
> 
> https://github.com/elastic/connectors-python/blob/main/connectors/sources/generic_database.py#L67-L114
> 
> Maybe could add a new flag there to facilitate validation.

That's IMO tightly coupled with Rich Configuration fields work - some of these validations that can be declarative will have to be properly reflected in the frontend - such as "non-empty" for instance. It'll be weird to validate in on a backend only IMO.

connectors/source.py
```diff
@@ -92,6 +124,14 @@ def get(self, key, default=None):
...
         return name in self._config
 
+    def any_field_is_empty(self, names):
+        if any(not self.has_field(name) for name in names):
+            raise ConfigurationFieldMissingError
+
+        empty_fields = [name for name in names if self._config[name].is_empty()]
+
+        return len(empty_fields) > 0, empty_fields
+
```
artem-shelkovnikov, line 134: This can probably be a separate PR with separate test

connectors/source.py
```diff
@@ -92,6 +124,14 @@ def get(self, key, default=None):
     def has_field(self, name):
         return name in self._config
 
+    def any_field_is_empty(self, names):
+        if any(not self.has_field(name) for name in names):
+            raise ConfigurationFieldMissingError
```
artem-shelkovnikov, line 129: Should this error mention which field is missing? Ideally it should include all fields that are missing IMO

connectors/source.py
```diff
@@ -20,6 +20,31 @@
 )
 
 
+class ConfigurationFieldMissingError(ValueError):
```
artem-shelkovnikov, line 23: Why different exception types?

I don't see any code catching these types, so it might be an overkill?

connectors/source.py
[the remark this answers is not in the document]
timgrein: I'm already checking that in the corresponding project brief, if that's also in the scope.

connectors/sources/generic_database.py
timgrein: This is specific to databases, so I don't think it's generic enough to have it in a `basic_validation` method inside `DataSourceConfiguration`.

connectors/source.py
timgrein: I thought it's useful to keep them distinct from each other, especially if we provide specific constructors, e.g. as you've pointed out we can pass the missing fields to `ConfigurationFieldMissingError` and for `ConfigurationFieldTypeError ` we could pass the expected type, actual type, field and value.

connectors/source.py
timgrein: It's tested in `test_any_field_is_empty` and two other tests in `test_source.py`. I can create a separate PR for the new error classes and this functionality 👍 Just don't want to overload our open PRs. We already have ~ 15 PRs open since one week.

artem-shelkovnikov: Yeah makes sense, just a note for the future :) I love small PRs and take much less time approving them :)

connectors/source.py
artem-shelkovnikov: Yeah makes sense, I think we can extend these error classes later with specific constructors, awesome!

connectors/source.py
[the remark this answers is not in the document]
navarone-feekery: Mandatory fields are indeed in the Rich Configurable Fields scope. Empty validation will rely on mandatory fields and also take defaults (if any) into consideration.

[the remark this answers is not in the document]
navarone-feekery: My general thoughts are; if you need this validation to fulfill some other development requirements, it's okay if you add it. I will just build out from what you've added or refactor it if it doesn't fit the implementation goals of Rich Configurable Fields.
Unfortunately I haven't written an implementation doc yet so I can't say whether the approach would be the same as yours here.

timgrein:
Closed as it has too much overlap with rich-configurable fields.

PR #562, opened 2023-02-27
[MySQL connector] Fix connection pool initialisation to be a context method - with_connection_pool

artem-shelkovnikov:
## In attempt to close https://github.com/elastic/enterprise-search-team/issues/3821

Problem: `ping` function initializes `self.connection_pool` variable that is needed to make queries to the SQL database. Later `get_docs` function uses this variable and it's suboptimal - we don't provide guarantees that `ping` is called before `get_docs`. Now it really is like that, but we might change it or add any other functions that are called before `ping` - like `validate_config`, for example, and if `validate_config` needs access to `self.connection.pool`, code will break.

Approach taken here is to provide `with_connection_pool` context that each method can use for its logic, for example:

```
def ping(self):
    async with self.with_connection_pool() as connection_pool:
         connection = connection_pool.acquire()
         connection.query("SELECT 1;")
```

What I like about this approach - this pattern ensures that connection_pool is properly disposed of after the function finishes its run. It also keeps granularity of pool/connection as original code makes this difference, I did not want to mess with it.

#### Pre-Review Checklist


tarekziade:
I think an async property sounds good, see projects like https://github.com/ryananguiano/async_property

but this is just cosmetics. You could also simply have an async ` get_pool` and have everyone use it to interact with the pool, and do the lazy init in it.

tarekziade:
btw I think you can hide all of the complexity because the pattern we want is to have a connection, so what about implementing a context manager?

```
@asynccontextmanager
async def get_connection(self):
    # lazy loading of the pool goes here
    async with self._connection_pool.acquire() as conn:
        yield conn
```

and then all the code does:

```
async with self.get_connection() as conn:
    xxx
```

connectors/sources/mysql.py
```diff
@@ -218,11 +216,21 @@ async def ping(self):
             if not self.ssl_disabled
             else None,
         }
+
+        if self._connection_pool is None:
+            self._connection_pool = await aiomysql.create_pool(**connection_string)
+
+        return self._connection_pool
```
parth-crest, line 223: As we are storing in instance variable `_connection_pool` then we can avoid returning and can be directly used `self.__connection_pool`

wangch079 (commented):
Or we can use the pattern similar to ruby:

```python
@property
async def connection_pool(self):
    if self._connection_pool is None:
        self._connection_pool = await self._get_connection_pool()
    return self._connection_pool
```

So you can use `self.connection_pool` directly in the code.

tarekziade:
> Or we can use the pattern similar to ruby:
> 
> ```python
> @property
> async def connection_pool(self):
>     if self._connection_pool is None:
>         self._connection_pool = await self._get_connection_pool()
>     return self._connection_pool
> ```
> 
> So you can use `self.connection_pool` directly in the code.

yes it's kind of similar but you can't have async properties unless you do a bit of work to wait for the task in there

connectors/sources/tests/test_mysql.py
```diff
@@ -303,15 +293,15 @@ async def test_fetch_rows_from_tables():
...
+        source, "with_connection_pool", return_value=ConnectionPool()
     ):
         response = source._connect(query)
 
         mock.patch("source._connect", return_value=response)
 
-    # Assert
-    async for row in source.fetch_rows_from_tables("table"):
-        assert "_id" in row
+        # Assert
```
timgrein, line 302:
```suggestion
```

I'm not the biggest fan of the `# Setup`, `# Execute` and `# Assert` comments tbh. Feel free to keep it for consistency, but I think we can remove them when we see them.

timgrein (approved):
LGTM, just two nits and a question

connectors/sources/tests/test_mysql.py
```diff
@@ -173,6 +173,23 @@ async def __aexit__(self, exception_type, exception_value, exception_traceback):
         pass
 
 
+class ConnectionPool:
```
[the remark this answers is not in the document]
artem-shelkovnikov: Oh it returned back cause it was there, I think you deleted it later.

I think it was Crest team's way of mocking things in a structured way. I'm not a fan of it, but to remove it I'd have to do several steps and I did not bother, maybe I should :D

connectors/sources/tests/test_mysql.py
```diff
@@ -188,50 +205,25 @@ async def mock_mysql_response():
...
-    source.connection_pool = None
-
-    await source.close()
-
-    assert source.connection_pool is None
-
-
-@pytest.mark.asyncio
-async def test_close_with_connection_pool():
+async def test_close():
```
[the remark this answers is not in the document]
artem-shelkovnikov: Great idea!

connectors/sources/tests/test_mysql.py
artem-shelkovnikov: Agreed!

connectors/sources/tests/test_mysql.py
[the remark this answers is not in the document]
wangch079: I think this is the proper way in python to test a function doesn't raise any exception

connectors/sources/tests/test_mysql.py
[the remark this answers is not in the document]
wangch079: I also think you can just patch `aiomysql.create_pool`

wangch079 (commented):
Nice work. It's much cleaner now.

Just minor comments on the tests

connectors/sources/mysql.py
```diff
@@ -224,16 +218,27 @@ async def ping(self):
             if self.ssl_enabled
             else None,
         }
-        logger.info("Pinging MySQL...")
-        if self.connection_pool is None:
-            self.connection_pool = await aiomysql.create_pool(**connection_string)
+
         try:
```
tarekziade, line 222: A more correct pattern is:

```
try:
    connection_pool = await aiomysql.create_pool(**connection_string)
except Exception as e:
    # deal with errors creating it and exit
    XXX

try:
    yield connection_pool
finally:
    connection_pool.close()
    await connection_pool.wait_closed()
```

so you don't mix issues with pool creation and pool usage
    
```

File: connectors/sources/mysql.py [hunk 6]
```diff
@@ -224,16 +218,27 @@ async def ping(self):
...
+                connection_pool.close()
+                await connection_pool.wait_closed()
+
+    async def ping(self):
+        """Verify the connection with MySQL server"""
+
+        logger.info("Pinging MySQL...")
+        async with self.with_connection_pool() as connection_pool:
+            try:
+                async with connection_pool.acquire() as connection:
```
tarekziade, line 236: It would be nice to include `acquire` in the context manager so we don't need this extra step here

connectors/sources/mysql.py
artem-shelkovnikov: Totally missed that, thanks!

connectors/sources/mysql.py
artem-shelkovnikov: As discussed, I did not want to make a functional change here - it was implemented by Crest this way (1 connection pool -> X connections). Intuitively in positive cases it would not change anything, but in case retry policy kicks in there might be some impact. Probably minor, but at this point I just don't feel like taking the risk, this change can be done in a separate PR later.

artem-shelkovnikov:
I decided not to rename/touch other tests than the ones affected by my change to not further increase the size of the PR.

timgrein (approved):
LGTM

PR #571, opened 2023-03-06
Fix typo in `oracle.py`

leemthompo:
~`DEFAUTL_WALLET_CONFIGURATION_PATH`~ -> `DEFAULT_WALLET_CONFIGURATION_PATH`

Not sure about **backports**, _iff_ this typo needs correcting.

### DIFF
--- a/connectors/sources/oracle.py
+++ b/connectors/sources/oracle.py
@@ -21,7 +21,7 @@ QUERIES = {
 }
 DEFAULT_PROTOCOL = "TCP"
 DEFAULT_ORACLE_HOME = ""
-DEFAUTL_WALLET_CONFIGURATION_PATH = ""
+DEFAULT_WALLET_CONFIGURATION_PATH = ""
 
 
 class OracleDataSource(GenericBaseDataSource):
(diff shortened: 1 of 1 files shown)


connectors/sources/oracle.py
```diff
@@ -69,7 +69,7 @@ def get_default_configuration(cls):
                     "type": "str",
                 },
                 "wallet_configuration_path": {
-                    "value": DEFAUTL_WALLET_CONFIGURATION_PATH,
+                    "value": DEFAULT_WALLET_CONFIGURATION_PATH,
```
timgrein, line 72:
```suggestion
                    "value": "",
```

connectors/sources/oracle.py
```diff
@@ -21,7 +21,7 @@
 }
 DEFAULT_PROTOCOL = "TCP"
 DEFAULT_ORACLE_HOME = ""
-DEFAUTL_WALLET_CONFIGURATION_PATH = ""
+DEFAULT_WALLET_CONFIGURATION_PATH = ""
```
timgrein, line 24:
```suggestion
```

timgrein (commented):
If it's only used once, we can simply inline it. As it's also just an empty string there's probably no point to extract it as the constant could be misleading, as at least I would expect that such a long constant actually contains a "real" value and not an empty string.

timgrein (approved):

PR #582, opened 2023-03-17
Update _validate_configuration to validate_config of BaseDataSource Class

moxarth-elastic:
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/tests/test_generic_database.py
+++ b/connectors/sources/tests/test_generic_database.py
@@ -130,37 +130,40 @@ def test_get_configuration(patch_logger):
     assert config["host"] == "127.0.0.1"
 
 
-def test_validate_configuration_missing_fields(patch_logger):
-    """Test _validate_configuration method check missing fields"""
+@pytest.mark.asyncio
+async def test_validate_config_missing_fields(patch_logger):
+    """Test validate_config method check missing fields"""
     # Setup
     source = create_source(GenericBaseDataSource)
     with pytest.raises(Exception):
         source.configuration.set_field(name="host", value="")
 
         # Execute
-        source._validate_configuration()
+        await source.validate_config()
 
 
-def test_validate_configuration_port(patch_logger):
-    """Test _validate_configuration method check port"""
+@pytest.mark.asyncio
+async def test_validate_config_port(patch_logger):
+    """Test validate_config method check port"""
     # Setup
     source = create_source(GenericBaseDataSource)
     with pytest.raises(Exception):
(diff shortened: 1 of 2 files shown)


connectors/sources/tests/test_generic_database.py
```diff
@@ -130,37 +130,40 @@ def test_get_configuration(patch_logger):
     assert config["host"] == "127.0.0.1"
 
 
-def test_validate_configuration_missing_fields(patch_logger):
-    """Test _validate_configuration method check missing fields"""
+@pytest.mark.asyncio
+async def test_validate_config_missing_fields(patch_logger):
```
timgrein, line 134: IMO this a very good example, where we can use `@pytest.parametrize` to test all fields, which are not allowed to be empty.

connectors/sources/tests/test_generic_database.py
```diff
@@ -130,37 +130,40 @@ def test_get_configuration(patch_logger):
...
+async def test_validate_config_ssl(patch_logger):
+    """Test validate_config method check ssl"""
     # Setup
     source = create_source(PostgreSQLDataSource)
     source.configuration.set_field(name="ssl_disabled", value=False)
 
     with pytest.raises(Exception):
         # Execute
-        source._validate_configuration()
+        await source.validate_config()
```
timgrein, line 166: Is there also a positive test case somewhere, i.e. the configuration is valid and no exception is raised?

moxarth-elastic: Done

timgrein (commented):
One small comment, otherwise LGTM. Can approve afterwards

connectors/sources/tests/test_generic_database.py
```diff
@@ -130,37 +130,55 @@ def test_get_configuration(patch_logger):
...
-    """Test _validate_configuration method check missing fields"""
+@pytest.mark.asyncio
+async def test_validate_config_valid_fields(patch_logger):
+    # Setup
+    source = create_source(GenericBaseDataSource)
+
+    # Execute
+    try:
+        await source.validate_config()
+        assert True
```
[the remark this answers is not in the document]
moxarth-elastic: we added this considering `explicit is better than implicit`

timgrein (approved):
LGTM

connectors/sources/tests/test_generic_database.py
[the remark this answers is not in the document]
moxarth-elastic: remove `assert True`

PR #583, opened 2023-03-17
Add concurrency control functionality

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/4157

This PR:
1. adds new instance variable `_seq_no` and `_primary_field` to `ESDocument` class
2. makes the `reload` method of `ESDocument` class update the instance variables directly
3. removes `@setter` methods, and makes `prepare` and `validate_filtering` methods of `Connector` class to `reload`.
4. makes the `get_all_docs` method of `ESIndex` to retrieve `_seq_no` and `_primary_field`
5. makes the `update` method of `ESIndex` to accept `if_seq_no` and `if_primary_field` paremeters

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -389,22 +389,9 @@ class Connector(ESDocument):
     def status(self):
         return Status(self.get("status"))
 
-    @status.setter
-    def status(self, value):
-        if isinstance(value, str):
-            value = Status(value)
-        if not isinstance(value, Status):
-            raise TypeError(value)
-
-        self._source["status"] = value
-
     @property
     def service_type(self):
         return self.get("service_type")
-
-    @service_type.setter
-    def service_type(self, value):
-        self._source["service_type"] = value
 
     @property
     def last_seen(self):
@@ -429,10 +416,6 @@ class Connector(ESDocument):
     def configuration(self):
         return DataSourceConfiguration(self.get("configuration"))
 
-    @configuration.setter
-    def configuration(self, value):
-        self._source["configuration"] = value
-
     @property
     def index_name(self):
         return self.get("index_name")
@@ -444,10 +427,6 @@ class Connector(ESDocument):
     @property
     def filtering(self):
         return Filtering(self.get("filtering"))
-
-    @filtering.setter
-    def filtering(self, value):
-        self._source["filtering"] = value
 
     @property
     def pipeline(self):
@@ -544,7 +523,7 @@ class Connector(ESDocument):
                     f"Service type is not configured for connector {configured_connector_id}"
                 )
                 raise ServiceTypeNotConfiguredError("Service type is not configured.")
-            self.service_type = doc["service_type"] = configured_service_type
+            doc["service_type"] = configured_service_type
             logger.debug(
                 f"Populated service type {configured_service_type} for connector {self.id}"
             )
@@ -557,10 +536,8 @@ class Connector(ESDocument):
                 source_klass = get_source_klass(fqn)
 
                 # sets the defaults and the flag to NEEDS_CONFIGURATION
-                self.configuration = doc[
-                    "configuration"
-                ] = source_klass.get_simple_configuration()
-                self.status = doc["status"] = Status.NEEDS_CONFIGURATION.value
+                doc["configuration"] = source_klass.get_simple_configuration()
+                doc["status"] = Status.NEEDS_CONFIGURATION.value
                 logger.debug(f"Populated configuration for connector {self.id}")
             except Exception as e:
                 logger.critical(e, exc_info=True)
@@ -570,6 +547,7 @@ class Connector(ESDocument):
 
         try:
             await self.index.update(doc_id=self.id, doc=doc)
+            await self.reload()
         except Exception as e:
             logger.critical(e, exc_info=True)
             raise ConnectorUpdateError(
@@ -601,8 +579,8 @@ class Connector(ESDocument):
                 if validation_result.state == FilteringValidationState.VALID:
                     filter_["active"] = filter_.get("draft")
 
-        self.filtering = filtering
         await self.index.update(doc_id=self.id, doc={"filtering": filtering})
+        await self.reload()
 
     async def document_count(self):
         await self.index.client.indices.refresh(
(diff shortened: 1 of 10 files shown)


connectors/sync_job_runner.py
```diff
@@ -144,7 +144,11 @@ async def _sync_done(self, sync_status, result=None, sync_error=None):
         else:
             await self.sync_job.done(ingestion_stats=ingestion_stats)
 
-        self.sync_job = await self.sync_job.reload()
+        try:
+            await self.sync_job.reload()
+        except Exception as e:
+            logger.error(f"Failed to reload sync job {self.job_id}. Error: {e}")
+            self.sync_job = None
```
tarekziade, line 151: putting `None` here seems pretty drastic, what happens next on the sync_doc call?

tarekziade (commented):
just one question about what happens when `reload` fails

connectors/sync_job_runner.py
wangch079: ~For now, this is the only place `reload` is used.~

You just reminds me I need to add `reload` in `prepare` and `validate_filtering` methods: https://github.com/elastic/connectors-python/pull/583/commits/e269d684601249972e37979c2d74cf9ce2259f66

>what happens next on the sync_doc call?

It will updates `last_sync_status` to `error`:
https://github.com/elastic/connectors-python/blob/e269d684601249972e37979c2d74cf9ce2259f66/connectors/byoc.py#L483

This is one of the edge cases we need to handle in https://github.com/elastic/connectors-python/pull/517 during job execution.

wangch079: ~For now, this is the only place `reload` is used.~

You just reminds me I need to add `reload` in `prepare` and `validate_filtering` methods: https://github.com/elastic/connectors-python/pull/583/commits/e269d684601249972e37979c2d74cf9ce2259f66

>what happens next on the sync_doc call?

It will updates `last_sync_status` to `error`:
https://github.com/elastic/connectors-python/blob/e269d684601249972e37979c2d74cf9ce2259f66/connectors/byoc.py#L483

This is one of the edge cases we need to handle in https://github.com/elastic/connectors-python/pull/517 during job execution.

artem-shelkovnikov:
@wangch079 I'm looking at the PR and understand individual pieces of it, but don't understand the high-level approach taken. Is there a document/code snippets/diagrams that I could check to understand how it will work when multiple parties try to update the same record?

wangch079:
>@wangch079 I'm looking at the PR and understand individual pieces of it, but don't understand the high-level approach taken. Is there a document/code snippets/diagrams that I could check to understand how it will work when multiple parties try to update the same record?

Hi @artem-shelkovnikov , the implementation doc is [here](https://docs.google.com/document/d/1D3CeQQcDZkHsuGYOpfoVoXVFphlW27FpT_Rifw52wD0/edit#). 

This PR is just to add the concurrency control to the `update` method (make it accept optional params `if_seq_no` and `if_primary_term`), but the real concurrency control is not used until https://github.com/elastic/enterprise-search-team/issues/4159

connectors/byoc.py
```diff
@@ -389,23 +389,10 @@ class Connector(ESDocument):
...
 
-    @status.setter
-    def status(self, value):
-        if isinstance(value, str):
-            value = Status(value)
-        if not isinstance(value, Status):
-            raise TypeError(value)
-
-        self._source["status"] = value
-
```
artem-shelkovnikov, line 400: So to double check, setters were removed because they are not used?

connectors/byoc.py
```diff
@@ -575,6 +552,7 @@ async def prepare(self, config):
             raise ConnectorUpdateError(
                 f"Could not update service type/configuration for connector {self.id}"
             )
+        await self.reload()
```
artem-shelkovnikov, line 555: What's the reason of adding the reload here?

connectors/byoc.py
```diff
@@ -601,8 +579,8 @@ async def validate_filtering(self, validator):
                 if validation_result.state == FilteringValidationState.VALID:
                     filter_["active"] = filter_.get("draft")
 
-        self.filtering = filtering
         await self.index.update(doc_id=self.id, doc={"filtering": filtering})
+        await self.reload()
```
artem-shelkovnikov, line 583: Same - do we actually want to reload everywhere then?

connectors/es/index.py
```diff
@@ -49,17 +57,25 @@ async def fetch_by_id(self, doc_id):
...
-        return self._create_object(resp.body)
+        return resp.body
 
     async def index(self, doc):
         resp = await self.client.index(index=self.index_name, document=doc)
         return resp["_id"]
 
-    async def update(self, doc_id, doc):
-        await self.client.update(index=self.index_name, id=doc_id, doc=doc)
+    async def update(self, doc_id, doc, if_seq_no=None, if_primary_term=None):
```
artem-shelkovnikov, line 71: So for now `if_seq_no` is not used, should it be added in a different PR, or should logic that uses it be added here?

artem-shelkovnikov (commented):
I look at the changeset and don't quite understand - nothing seem to functionally change regarding concurrency control in this PR. I see `if_seq_no` added, but not used - should this change be broken into multiple ones instead? 
So changes about `DocumentNotFoundError` can be a different PR. Changes about `reload` - are they relevant to `if_seq_no` change?

I'm pretty confused tbh.

connectors/byoc.py
wangch079: So the `connector` object has the latest `service_type` and `configuration`.

connectors/byoc.py
wangch079: Not everywhere. We try not to do so if possible. But this place is critical. Imagine a customer edits filtering in Kibana, BE validates it, and updates the active filtering. If we don't reload here, the connector will still hold the out-dated filtering, which will be used to schedule a new sync job.

connectors/es/index.py
wangch079: I'm trying to break the changes as small as possible. The purpose of this PR is, as the title says, to `Add concurrency control functionality`, i.e. make it capable of using optimistic concurrency control, but not necessarily use it.

wangch079:
Close this PR as it has been further broken down into
1. #607
2. #608 
3. #609 
4. #610 
5. #611

PR #584, opened 2023-03-19
service factory metaclass

tarekziade:
This patch prepares https://github.com/elastic/connectors-python/pull/220/files where I'll add a new service.

It introduces a simple metaclass to simplify the registry of services in the CLI (one service == one action)

It allows running the CLI like this:

```
elastic-ingest --action sync cleanup upload
```

And automatically create a `MultiService` instance with the two services -- removing the need to define services in CLI.
To add a new service, you just create the class and name the action

### DIFF
--- a/connectors/services/base.py
+++ b/connectors/services/base.py
@@ -14,7 +14,29 @@ class ServiceAlreadyRunningError(Exception):
     pass
 
 
-class BaseService:
+_SERVICES = {}
+
+
+def get_services(names, config):
+    return MultiService(*[get_service(name, config) for name in names])
+
+
+def get_service(name, config):
+    return _SERVICES[name](config)
+
+
+class _Registry(type):
+    def __new__(cls, name, bases, dct):
+        service_name = dct.get("name")
+        class_instance = super().__new__(cls, name, bases, dct)
+        if service_name is not None:
+            _SERVICES[service_name] = class_instance
+        return class_instance
+
+
+class BaseService(metaclass=_Registry):
+    name = None
+
     def __init__(self, config):
         self.config = config
         self.service_config = self.config["service"]
--- a/connectors/tests/test_cli.py
+++ b/connectors/tests/test_cli.py
@@ -51,7 +51,7 @@ def test_run(mock_responses, patch_logger, set_env):
     args = mock.MagicMock()
     args.log_level = "DEBUG"
     args.config_file = CONFIG
-    args.action = "list"
+    args.action = ["list"]
     with patch("sys.stdout", new=StringIO()) as patched_stdout:
         assert run(args) == 0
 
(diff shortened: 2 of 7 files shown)


connectors/services/base.py
```diff
@@ -14,7 +14,29 @@ class ServiceAlreadyRunningError(Exception):
...
+class _Registry(type):
+    def __new__(cls, name, bases, dct):
+        service_name = dct.get("name")
+        class_instance = super().__new__(cls, name, bases, dct)
+        if service_name is not None:
+            _SERVICES[service_name] = class_instance
+        return class_instance
+
+
+class BaseService(metaclass=_Registry):
```
wangch079, line 37: TIL

wangch079 (approved):
Cool, this will make it possible to run job scheduler or executor only

connectors/tests/test_cli.py
```diff
@@ -63,6 +63,17 @@ def test_run(mock_responses, patch_logger, set_env):
         assert "Bye" in output
 
 
+def test_run_snowflake(mock_responses, patch_logger, set_env):
```
artem-shelkovnikov, line 66: Why snowflake? :D

connectors/services/base.py
```diff
@@ -14,7 +14,29 @@ class ServiceAlreadyRunningError(Exception):
...
+
+def get_services(names, config):
+    return MultiService(*[get_service(name, config) for name in names])
+
+
+def get_service(name, config):
+    return _SERVICES[name](config)
+
+
+class _Registry(type):
```
artem-shelkovnikov, line 28: Why make it private? I think this class can be pretty public :)

connectors/services/base.py
```diff
@@ -14,7 +14,29 @@ class ServiceAlreadyRunningError(Exception):
...
+    def __new__(cls, name, bases, dct):
+        service_name = dct.get("name")
+        class_instance = super().__new__(cls, name, bases, dct)
+        if service_name is not None:
+            _SERVICES[service_name] = class_instance
+        return class_instance
+
+
+class BaseService(metaclass=_Registry):
+    name = None
```
artem-shelkovnikov, line 38: I think this deserves a comment why it should be defined.

Also maybe it should be `action`?

artem-shelkovnikov (approved):

connectors/services/base.py
tarekziade: this metaclass is an internal thing for the base class, I don't see any reason to have it public for now, unless we reuse it in another class

connectors/services/base.py
tarekziade: Great point! I'll add a comment.

> Also maybe it should be action?

`action` is part of the `argparse` / CLI verbiage. Maybe we could go the other way around and use `--service` in the cli?

artem-shelkovnikov (approved):

PR #588, opened 2023-03-20
[MySQL] Rename a test

timgrein:
It was not obvious what the test actually tested. Using a more explicit name now.


artem-shelkovnikov (approved):
I'd probably rename it to follow the convention:

`test_<METHOD>_with_<CONDITION>_<ACTION>`. I don't think we strictly follow this convention, we can discuss it, but IMO it's quite readable and simple.

So here it'd be `test_get_docs_with_empty_db_fields_raises_error`

artem-shelkovnikov (approved):

PR #593, opened 2023-03-21
Decouple job execution service

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4155

This PR decouples the job execution to a separate service.

1. The `JobSchedulingService` will only be responsible for scheduling jobs. It will create a job for a connector when the connector is scheduled to sync, but won't execute the job.
2. The `JobExecutionService` will only be responsible for executing jobs. It will periodically query jobs with `pending`/`suspended` status, and try to execute them.

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -35,8 +35,8 @@ def _parser():
     parser.add_argument(
         "--action",
         type=str,
-        default=["poll", "cleanup"],
-        choices=["poll", "list", "cleanup"],
+        default=["schedule", "execute", "cleanup"],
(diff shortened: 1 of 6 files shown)


artem-shelkovnikov (commented):
What does functionally change in the service with this change?

It feels like `schedule` service now does both execution and scheduling at the same time, while `execute` does only execution, is it correct?

It's not possible to run them both at the same time in same process, is it correct as well?

artem-shelkovnikov (approved):
Thanks for explanation, lgtm!

artem-shelkovnikov (approved):

PR #604, opened 2023-03-22
[Confluence] Add indexing functionality

praveen-elastic:
#### Pre-Review Checklist
- [x] ~~Considered corresponding documentation changes~~
- [x] ~~Contributed any configuration settings changes to the configuration reference~~

### DIFF
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -5,19 +5,52 @@
 #
 """Confluence source module responsible to fetch documents from Confluence Cloud/Server.
 """
+import asyncio
 import os
+from copy import copy
+from functools import partial
 
+import aiofiles
 import aiohttp
+from aiofiles.os import remove
+from aiofiles.tempfile import NamedTemporaryFile
 from aiohttp.client_exceptions import ServerDisconnectedError
 
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import CancellableSleeps, iso_utc, ssl_context
+from connectors.utils import (
+    TIKA_SUPPORTED_FILETYPES,
+    CancellableSleeps,
+    ConcurrentTasks,
+    MemQueue,
+    convert_to_b64,
+    iso_utc,
+    ssl_context,
+)
 
 FILE_SIZE_LIMIT = 10485760
 RETRY_INTERVAL = 2
+SPACE = "space"
+ATTACHMENT = "attachment"
+CONTENT = "content"
+DOWNLOAD = "download"
+SPACE_QUERY = "limit=100"
+ATTACHMENT_QUERY = "limit=100&expand=version"
+CONTENT_QUERY = (
+    "limit=100&expand=children.attachment,history.lastUpdated,body.storage,space"
+)
+
+URLS = {
+    SPACE: "rest/api/space?{api_query}",
+    CONTENT: "rest/api/content?{api_query}",
+    ATTACHMENT: "rest/api/content/{id}/child/attachment?{api_query}",
+}
 PING_URL = "rest/api/space?limit=1"
 MAX_CONCURRENT_DOWNLOADS = 50  # Max concurrent download supported by confluence
+CHUNK_SIZE = 1024
+MAX_CONCURRENCY = 50
+QUEUE_SIZE = 1024
+QUEUE_MEM_SIZE = 25 * 1024 * 1024  # Size in Megabytes
 
 
 class ConfluenceClient:
@@ -112,6 +145,36 @@ class ConfluenceClient:
                 )
                 await self._sleeps.sleep(RETRY_INTERVAL**retry_counter)
 
+    async def paginated_api_call(self, url_name, **url_kwargs):
+        """Make a paginated API call for Confluence objects using the passed url_name.
+        Args:
+            url_name (str): URL Name to identify the API endpoint to hit
+        Yields:
+            response: JSON response.
+        """
+        should_paginate = True
+        url = os.path.join(self.host_url, URLS[url_name].format(**url_kwargs))
+        while should_paginate:
+            try:
+                async for response in self.api_call(
+                    url=url,
+                ):
+                    json_response = await response.json()
+                    links = json_response.get("_links")
+                    yield json_response
+                    if links.get("next") is None:
+                        should_paginate = False
+                        break
+                    url = os.path.join(
+                        self.host_url,
+                        links.get("next")[1:],
+                    )
+            except Exception as exception:
+                logger.warning(
+                    f"Skipping data for type {url_name} from {url}. Exception: {exception}."
+                )
+                break
+
 
 class ConfluenceDataSource(BaseDataSource):
     """Confluence"""
@@ -129,6 +192,10 @@ class ConfluenceDataSource(BaseDataSource):
         self.enable_content_extraction = self.configuration["enable_content_extraction"]
         self.concurrent_downloads = self.configuration["concurrent_downloads"]
         self.confluence_client = ConfluenceClient(configuration)
+
+        self.queue = MemQueue(maxsize=QUEUE_SIZE, maxmemsize=QUEUE_MEM_SIZE)
+        self.fetchers = ConcurrentTasks(max_concurrency=MAX_CONCURRENCY)
+        self.fetcher_count = 0
 
     @classmethod
     def get_default_configuration(cls):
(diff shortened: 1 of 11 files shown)


connectors/sources/confluence.py
```diff
@@ -256,6 +323,214 @@ async def ping(self):
             logger.exception("Error while connecting to Confluence")
             raise
 
+    async def get_spaces(self):
```
timgrein, line 326: I think that should be part of the `ConfluenceClient`

connectors/sources/confluence.py
```diff
@@ -256,6 +323,214 @@ async def ping(self):
...
-        # yield dummy document to make the chunked PR work, subsequent PR will replace it with actual implementation
-        yield {"_id": "123", "timestamp": iso_utc()}, None
+        """Executes the logic to fetch Confluence content in async manner.
+
+        Args:
+            filtering (Filtering): Object of Class Filtering
+        Yields:
+            dictionary: dictionary containing meta-data of the content.
+        """
+        await self.fetchers.put(self._space_coro)
```
timgrein, line 527: A similar mechanism is already present in the framework. Is there a specific reason why we don't yield lazy download co-routines to the framework and let the framework handle it?

connectors/sources/tests/test_confluence.py
```diff
@@ -5,16 +5,188 @@
...
+    async def __aexit__(self, exc_type, exc, tb):
+        """Closes an async with block"""
+        pass
+
+    async def __aenter__(self):
+        """Enters an async with block"""
+        return self
+
+
+class AsyncIter:
```
timgrein, line 176: There's already an `AsyncIterator` class present, which can be used for testing. Is there a specific reason to create a new class?

connectors/sources/confluence.py
```diff
@@ -256,6 +323,214 @@ async def ping(self):
...
+            for attachment in attachments:
+                yield {
+                    "type": attachment["type"],
+                    "title": attachment["title"],
+                    "_id": attachment["id"],
+                    "space": parent_space,
+                    parent_type: parent_name,
+                    "_timestamp": attachment["version"]["when"],
+                    "size": attachment["extensions"]["fileSize"],
+                    "url": os.path.join(
```
timgrein, line 409: This could be a variable to indicate what the value contains.

connectors/sources/confluence.py
```diff
@@ -256,6 +323,214 @@ async def ping(self):
...
+        if os.path.splitext(attachment_name)[-1] not in TIKA_SUPPORTED_FILETYPES:
+            logger.warning(f"{attachment_name} can't be extracted")
+            return
+
+        if attachment_size > FILE_SIZE_LIMIT:
+            logger.warning(
+                f"File size {attachment_size} of file {attachment_name} is larger than {FILE_SIZE_LIMIT} bytes. Discarding file content"
+            )
+            return
+        logger.debug(f"Downloading {attachment_name}")
```
timgrein, line 440: Would be great if we can have the size information here and also that it's downloaded from confluence to improve traceability.

timgrein (commented):
Added some comments. Main question for me: Why do we recreate the fetcher functionality?

connectors/sources/confluence.py
praveen-elastic: `get_spaces` fetches the all the spaces from the confluence. It is similar to `fetch_documents` (fetches pages, blogs) and `fetch_attachments` (files, images, etc.) in terms of functionality, so I believe it should reside in the DataSource class.

I think the naming is a bit misleading. I should've named it as `fetch_spaces` instead. I'll update it.

connectors/sources/tests/test_confluence.py
praveen-elastic: I agree, I tried to use the framework one, but it only yields one item, I tried to make it generic to yield more than one item in this [PR](https://github.com/elastic/connectors-python/pull/539) but other framework test were failing, so I added it here for Confluence.

timgrein: Did you wrap the items in a list?

connectors/sources/confluence.py
praveen-elastic: `lazy_download` used in the framework is responsible for downloading content of objects that were yielded. In Confluence, content is only present in the `attachment` object. 
But, the fetching of  `attachments` needs to happen after `pages` and `blogs`  are fetched to use their ids as reference for api call. This dependency slows down the connector if we don't allow the connector to start grabbing attachments for the already fetched pages, blogs. 

Using the ConcurrentTasks directly here improves the performance of the connector, as it allows fetching all objects in parallel. 
Downloading the content of `attachment` is still happening via `lazy_download` by the framework.

connectors/sources/tests/test_confluence.py
praveen-elastic: Understood, this worked, updated using framework `AsyncIterator` from framework 🎉 [7e417b1](https://github.com/elastic/connectors-python/pull/604/commits/7e417b16a27e7f021fe5048cdb274cbceb7f11b8)

connectors/sources/confluence.py
```diff
@@ -112,6 +145,36 @@ async def api_call(self, url):
...
+    async def paginated_api_call(self, url_name, **url_kwargs):
+        """Make a paginated API call for Confluence objects using the passed url_name.
+        Args:
+            url_name (str): URL Name to identify the API endpoint to hit
+        Yields:
+            response: JSON response.
+        """
+        should_paginate = True
+        url = os.path.join(self.host_url, URLS[url_name].format(**url_kwargs))
+        while should_paginate:
```
timgrein, line 157: Do we even need `should_paginate` or could we just use `while True` and simply break, if we encounter a condition to break?

connectors/sources/confluence.py
```diff
@@ -112,6 +145,36 @@ async def api_call(self, url):
                 )
                 await self._sleeps.sleep(RETRY_INTERVAL**retry_counter)
 
+    async def paginated_api_call(self, url_name, **url_kwargs):
```
timgrein, line 148: From a method called `paginate_...` I would expect to be able to set a page number and a page size to fetch. What does paginate mean in the context of this method?

connectors/sources/tests/test_confluence.py
```diff
@@ -5,16 +5,173 @@
...
+    "_timestamp": "2023-01-03T09:24:50.633Z",
+    "title": "demo.py",
+    "size": 230,
+    "space": "DEMO",
+    "blog": "demo-blog",
+    "url": f"{HOST_URL}/pages/viewpageattachments.action?pageId=1113&preview=demo.py",
+}
+
+
+class MockJsonResponse:
```
timgrein, line 140: Could we use `AsyncMock` instead?

connectors/sources/tests/test_confluence.py
```diff
@@ -5,16 +5,173 @@
...
+    async def __aexit__(self, exc_type, exc, tb):
+        """Closes an async with block"""
+        pass
+
+    async def __aenter__(self):
+        """Enters an async with block"""
+        return self
+
+
+class MockObjectResponse:
```
timgrein, line 161: Could we use `AsyncMock` instead?

connectors/sources/tests/test_confluence.py
```diff
@@ -161,7 +318,168 @@ def test_validate_configuration_for_ssl_enabled(patch_logger):
...
+            response = await source.download_attachment(
+                url="download/attachments/1113/demo.py?version=1&modificationDate=1672737890633&cacheVersion=1&api=v2",
+                attachment=EXPECTED_ATTACHMENT,
+                doit=True,
+            )
+            assert response == EXPECTED_CONTENT
+
+
+@pytest.mark.asyncio
+async def test_download_attachment_when_filesize_is_large():
```
timgrein, line 386:
```suggestion
async def test_download_attachment_when_filesize_is_large_then_download_fails():
```

Do we expect a failure here or just an empty response?

connectors/sources/tests/test_confluence.py
```diff
@@ -161,7 +318,168 @@ def test_validate_configuration_for_ssl_enabled(patch_logger):
...
+            response = await source.download_attachment(
+                url="download/attachments/1113/demo.py?version=1&modificationDate=1672737890633&cacheVersion=1&api=v2",
+                attachment=attachment,
+                doit=True,
+            )
+            assert response is None
+
+
+@pytest.mark.asyncio
+async def test_download_attachment_for_unsupported_filetype():
```
timgrein, line 412:
```suggestion
async def test_download_attachment_when_unsupported_filetype_used_then_fail_download_fails():
```

Same as above: What's happening? Does the download file or is it simply skipped?

timgrein (commented):
Some nits and questions in some tests. Can approve afterwards 👍 Thank you for the changes you've already made.

connectors/sources/tests/test_confluence.py
praveen-elastic: We expect the content of attachment as empty, updating the name to  `test_download_attachment_when_filesize_is_large_then_download_skips`

connectors/sources/tests/test_confluence.py
praveen-elastic: In this case too the content download is skipped, renaming test to `test_download_attachment_when_unsupported_filetype_used_then_fail_download_skips`

connectors/sources/confluence.py
praveen-elastic: Yes, we can use directly `while True` then `break` inplace of a variable like `should_paginate`. I kept it this way as it maintains the code readability otherwise the structure becomes like below:

```
while True:
   ...
   if links.get("next") is None:
      break    # to come out of async for loop
      break    # to come out of pagination (the while loop)
```

My only concern is two breaks one after the other is a bit confusing to understand.

connectors/sources/confluence.py
praveen-elastic: This method performs api calls to the client, by taking care of next page, if the next page url is returned by the previous call. 
It is not intended to be called multiple times for a given object. Instead once called, it internally handles the next pages calls.

connectors/sources/tests/test_confluence.py
praveen-elastic: We need json attribute and AsyncMock will not have a .json() attribute

connectors/sources/tests/test_confluence.py
praveen-elastic: Similarly here we need a StreamReader object to be returned named `content`, so we had to create a custom Class

connectors/sources/confluence.py
timgrein: Couldn't you just return?

connectors/sources/tests/test_confluence.py
praveen-elastic: Updated, thanks [7880cf0](https://github.com/elastic/connectors-python/pull/604/commits/7880cf0cd4d5c1b14509a6b7659b7af65bc54f50)

connectors/sources/confluence.py
praveen-elastic: Correct, thanks, Updated with commit id: [7880cf0](https://github.com/elastic/connectors-python/pull/604/commits/7880cf0cd4d5c1b14509a6b7659b7af65bc54f50)

timgrein (approved):
LGTM, thank you for the changes

PR #611, opened 2023-03-23
Remove setters

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4157

This PR removes `setters` of the `Connector` class, and explicitly `reload` the connector after `prepare` and `validate_filtering`.

### Why the changes:

The `setters` was initially added to save another ES call to retrieve the latest document, by setting the fields directly, and these are used in methods `prepare` and `validate_filtering`. With the requirement of concurrency control, the document needs to hold the latest `_seq_no` and `_primary_term`, which means it has to make the ES call. And this makes those `setters` redundant and that's they are removed in this PR.

#### Pre-Review Checklist


artem-shelkovnikov:
Change looks good, but I don't understand the reasoning for it. Can you add the description of why this change is beneficial to this PR @wangch079?

artem-shelkovnikov (approved):

PR #612, opened 2023-03-23
[Confluence] Add E2E test for the connector

praveen-elastic:
#### Pre-Review Checklist
- [x] ~~Covered the changes with automated tests~~
- [x] ~~Considered corresponding documentation changes~~
- [x] Contributed any configuration settings changes to the configuration reference: Added Confluence to `nightly.yml`

- [x] New external dependency added for flask


connectors/sources/tests/fixtures/confluence/fixture.py
```diff
@@ -0,0 +1,161 @@
...
+    args = request.args
+    content = {
+        "results": [],
+        "start": 0,
+        "limit": 100,
+        "size": 100,
+        "_links": {"next": None},
+    }
+    document_type = args.get("type", "page")
+    for content_count in range(100):
```
timgrein, line 91: I would suggest, that we use a variable with a useful name for `100`.

connectors/sources/tests/fixtures/confluence/fixture.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    attachments = {
+        "results": [],
+        "start": 0,
+        "limit": 100,
+        "size": 5,
+        "_links": {"next": None},
+    }
+
+    for attachment_count in range(1, 6):
```
timgrein, line 127: I would also suggest here to use variables for `1` and `6`.

timgrein (commented):
I'm not 100% sure, if that's the whole test right? That's more like setting up a fake API?

connectors/sources/tests/fixtures/confluence/fixture.py
```diff
@@ -0,0 +1,161 @@
...
+        spaces["results"].append(
+            {
+                "id": "space 0",
+                "name": "Demo Space 0",
+                "_links": {
+                    "webui": "/spaces/space0",
+                },
+            }
+        )
+    else:
```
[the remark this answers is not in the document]
praveen-elastic: We are using this for ping api call, by keeping limit as 1.

timgrein:
@praveen-elastic Makes sense. But still this PR does not include the actual test right or am I missing something?

praveen-elastic:
> @praveen-elastic Makes sense. But still this PR does not include the actual test right or am I missing something?

This is the complete set of changes for E2E to run, we haven't added `load`, `remove` methods but instead mocking the APIs and passing the response. Morover, we still generate data for attachments via `LARGE_DATA = "".join([random.choice(string.ascii_letters) for _ in range(FILE_SIZE)])`

timgrein (approved):
LGTM

PR #619, opened 2023-03-23
Illustrative change to newer test conventions

artem-shelkovnikov:
For discussion - changing `connectors/tests/test_base_service.py` to follow the new test convention:

1. Tests for specific class of the module are wrapped in test class with name `Test_<CLASS_NAME>`.
2. Test cases have format of `test_<method_name>_when_<condition>_then_<outcome>`

Some additional notes:

1. We can actually keep names of tests short and instead provide description of the test in the docstring, for example:

```
@pytest.mark.asyncio
    async def test_run_stops_on_exception(self):
        """
        When one service raises exception then all services are stopped
        """
        # ...
```
It's possible to change pytest in such a way, that it would output these docstrings instead of method names when running pytest in verbose mode, and I kinda like it.

2. Not reflected in the PR, but I think that we can just move all tests in <ROOT>/tests instead of having different directories. Later we can put different codeowners rules for connectors and framework if needed, but for now I don't see a major problem having tests fully in a separate directory. When it's done, the structure of `tests` directory will fully repeat the structure of `connectors` directory, so test for `connectors/sources/mysql.py` will be located in `tests/sources/test_mysql.py`.

Let's use this PR to finalise the approach, and I'll create an issue afterwards to move all existing tests under this structure.

### DIFF
--- a/connectors/conftest.py
+++ b/connectors/conftest.py
@@ -145,3 +145,20 @@ def assert_re(expr, items):
             return
 
     raise AssertionError(f"{expr} not found in {items}")
+
+
+def pytest_itemcollected(item):
+    def _collect_recursively(n):
+        from _pytest.python import Module
+
+        if isinstance(n, Module):
+            return n.obj.__doc__.strip() if n.obj.__doc__ else n.obj.__class__.__name__
+
+        parent = n.parent
+
+        parent_description = _collect_recursively(parent) if parent else ""
+        node_description = n.obj.__doc__.strip() if n.obj.__doc__ else n.obj.__name__
+
+        return ".".join((parent_description, node_description))
+
+    item._nodeid = _collect_recursively(item)
--- a/connectors/conftest.py
+++ b/connectors/conftest.py
@@ -148,6 +148,22 @@ def assert_re(expr, items):
 
 
 def pytest_itemcollected(item):
+    """
+    This code is here to format the output of verbose run of pytest.
+
+    What it does is - it receives a object representing a test that will be ran in {item} argument.
+    Test can be located either just in the module, or inside of the class,
+    or inside of the class inside of the class inside of the class, etc.
+
+    So the idea of the code inside is to collect the name/docstring of current node
+    and then repeat it recursively until an instance of _pytest.python.Module is met.
+
+    It gives the following results:
+    <module>.<class>.<test> or <module>.<test> or <module>.<class>.<class>.<test>
+
+    It tries to use node's docstring if there's one, otherwise its name (or class name for modules)
+    """
+
     def _collect_recursively(n):
         from _pytest.python import Module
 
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov:
> Would it require much work?

Not really, it's doable really quick if it's not already implemented in pytest.

tarekziade:
this is a great improvement.
One thing we could also think about is whether we want to split pure unit tests from functional tests.

Projects use in general an `ftests` dir for functional tests and a `tests` for unit

artem-shelkovnikov:
> One thing we could also think about is whether we want to split pure unit tests from functional tests.
> 
> Projects use in general an ftests dir for functional tests and a tests for unit

I think it's a great proposal, having our functional tests in `<ROOT_DIR>/ftest` makes a lot of sense to me.

wangch079:
>but I think that we can just move all tests in /tests instead of having different directories.

I'd really like to keep the source test together with the source implementation. Looking at the code now, all source tests are in one tests folder: https://github.com/elastic/connectors-python/tree/main/connectors/sources/tests. I think it's much cleaner to put the source implementation and source test in one folder so it's easily isolated from other sources.

artem-shelkovnikov:
> I think it's much cleaner to put the source implementation and source test in one folder so it's easily isolated from other sources.

What about unit/functional tests? Should they also be in the connector folder?

So our `sources` directory will have a different structure? Can you give an example of how you see it structured?

wangch079:
> What about unit/functional tests? Should they also be in the connector folder?
> 
> So our `sources` directory will have a different structure? Can you give an example of how you see it structured?

I'm thinking this structure:

```
├── sources
│   ├── mysql
│   │   ├── mysql.py
│   │   ├── tests
│   │    │   ├── test_mysql.py
│   ├── ...
```

But I didn't realize there are function tests as well, which can also fit into this structure (adding a new fixtures folder). My main point is that each connector is very well isolated.

tarekziade:
> > What about unit/functional tests? Should they also be in the connector folder?
> > So our `sources` directory will have a different structure? Can you give an example of how you see it structured?
> 
> I'm thinking this structure:
> 
> ```
> ├── sources
> │   ├── mysql
> │   │   ├── mysql.py
> │   │   ├── tests
> │   │    │   ├── test_mysql.py
> │   ├── ...
> ```
> 
> But I didn't realize there are function tests as well, which can also fit into this structure (adding a new fixtures folder). My main point is that each connector is very well isolated.

that's a lot of nesting for each connector imho

seanstory:
I like the proposed organization, FWIW. I'd also support a structure like:

```
.
├── lib
│   └── sources
│       ├── mongo
│       │   └── mongo.py
│       └── mysql
│           └── mysql.py
└── test
    └── sources
        ├── mongo
        │   └── test_mongo.py
        └── mysql
            └── test_mysql.py
```

artem-shelkovnikov:
Closing as it's actually now implemented

PR #624, opened 2023-03-23
switched to pytds

tarekziade:
`pyodbc` requires the installation of `Microsoft ODBC` and `` or `iODBC` + `FreeTDS` which are (L)GPL based. 

These licenses are problematic to deploy our connectors in the cloud. Switching to `pytds` (MIT) resolves this


tarekziade:
https://buildkite.com/elastic/connectors-python-nightly/builds/435#0187112e-0b99-4419-b35a-29872f0bda71

timgrein (approved):
Just curious: What's the rationale? Is it faster/more reliable?

timgrein:
Oops, missed that it's in draft.

tarekziade:
> Just curious: What's the rationale? Is it faster/more reliable?

no OSS licensing issue. (L)GPL is viral we can't use it

moxarth-elastic:
@tarekziade we need to remove the config parameter `mssql_driver`. Also currently in E2E we use `pyodbc` for data generation so do you think we need to change that as well?

moxarth-elastic:
> > 
> 
> ideally, yea, so we don't have any remaining (L)GPL lib in the project. Do you mind having a look @moxarth-elastic ?

Sure, I will take a look to change dependency in E2E.

connectors/sources/mssql.py
```diff
@@ -100,15 +93,14 @@ def _create_engine(self):
         """Create sync engine for mssql"""
         if self.secured_connection:
             query = {
-                "driver": self.mssql_driver,
                 "TrustServerCertificate": "no",
```
moxarth-elastic, line 96: Regarding `TrustServerCertificate` and `Encrypt` it's specific to odbc driver. so I think we need to change this configuration.

docs/DEVELOPING.md
```diff
@@ -640,7 +640,6 @@ Microsoft SQL Server versions 2017, 2019, and Azure SQL, Amazon RDS for Microsof
 
 **Prerequisites**
 
-- The Microsoft SQL connector requires an ODBC driver to execute. Connector supports `Microsoft ODBC Driver 18`. To install ODBC driver please refer following [documentation](https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-driver-manager).
```
seanstory, line 643: do we also need changes to the configuration docs? See line 710: `##### mssql_driver` (too far below the diff for github to let me comment on it)

seanstory:
Added the 8.7.0 label to make sure it backports, as this isn't included in https://github.com/elastic/connectors-python/pull/630

requirements/framework.txt
```diff
@@ -16,4 +16,5 @@ azure-storage-blob==12.13.0
 SQLAlchemy==2.0.1
 oracledb==1.2.2
 asyncpg==0.27.0
-pyodbc==4.0.34 # Latest version is v4.0.35. However it is not compatible with MAC M1 chip so pining this version
```
seanstory, line 19: Should also remove this from the mssql tests requirements file

PR #636, opened 2023-03-24
Replace print with logger.info for directory connector

artem-shelkovnikov:
Minor change - it's the only place in code where `print` is used instead of `logger.info`, thus this PR contains only 1 line of change :)


timgrein (approved):
LGTM

Issue #601, opened 2023-03-22
Add a an error class / collection of error classes for validation-related errors


artem-shelkovnikov:
### Problem Description

Currently connectors that implement `validate_config` are expected to raise an exception when configuration is invalid. For now there is not exception class that was specifically created for this purpose, so all implementations just raise `Exception`.

This is problematic potentially as we want to handle validation errors differently vs other errors - now our `except Exception` catches every possible exception, thus we can swallow `ValueError`s,  `TypeError`s and other.

### Proposed Solution
Add a class (set of classes) that are supposed to be raised from the `validate_config` - for example `ConfigurationInvalidError` for now should be sufficient. In future, if we want more granular error handling, we can add other classes, such as `ConfigurableFieldRequiredError` and similar.

### Alternatives
`validate_config` can instead of relying on exceptions return a ValidationResult. For example:

```
def validate_config(self):
    if not self.configuration["user"]:
        return ValidationResult.invalid("user is required")
```

PR #637, opened 2023-03-24
Rename _validate_confuration -> validate_config and remove internal usage of it from connector

artem-shelkovnikov:
## Part of https://github.com/elastic/connectors-python/issues/601

As a part of standardisation effort, we're moving all validation of connector configurations into `validate_config` method of the connector.

There was already a `_validate_configuration` method in the connector, so I've renamed it to `validate_config` and removed usage of `_validate_configuration` from `ping` method.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/s3.py
+++ b/connectors/sources/s3.py
@@ -72,7 +72,7 @@ class S3DataSource(BaseDataSource):
         ) as s3:
             yield s3
 
-    def _validate_configuration(self):
+    def validate_config(self):
         """Validates whether user input is empty or not for configuration fields
 
         Raises:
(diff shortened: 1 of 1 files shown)


timgrein (approved):
LGTM, but a bit concerned, that there are no corresponding changes in `test_s3.py`. Probably the test simply doesn't exist?

artem-shelkovnikov:
@timgrein you are right, there was one test that did not break cause it was catching general `Exception` class. I'm planning to introduce a separate class for validation errors so that it would not happen, but meanwhile I reviewed and fixed this test.

PR #639, opened 2023-03-27
Introduce and use ConfigurableFieldValueError

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors-python/issues/601

This PR adds a new error type - `connectors.utils.ConfigurableFieldValueError` that is supposed to be used in `valdiate_config` of individual connector source files.

For now I haven't changed the way the errors are treated in the framework - framework still catches `Exception` and fails the sync, but in future I'm planning to add separate handling to also properly distinguish validation errors vs other errors (type errors, value errors, problems accessing 3rd-party system, etc).

What I also don't address - incompatible types for values are still handled in a messy way. E.g. if connector expects an integer in "port", but you give it a string, you'll get an unreadable exception message.

#### Pre-Review Checklist

### DIFF
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -472,7 +472,3 @@ def get_pem_format(key, max_split=-1):
     key = " ".join(key.split("\n", max_split))
     key = " ".join(key.rsplit("\n", max_split))
     return key
-
-
-class ConfigurableFieldValueError(Exception):
-    pass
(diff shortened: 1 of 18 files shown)


connectors/utils.py
```diff
@@ -472,3 +472,7 @@ def get_pem_format(key, max_split=-1):
     key = " ".join(key.split("\n", max_split))
     key = " ".join(key.rsplit("\n", max_split))
     return key
+
+
+class ConfigurableFieldValueError(Exception):
```
timgrein, line 477: Should we move this to `source.py` as `DataSourceConfiguration` also lives there and it feels like both belong together?

timgrein (approved):
LGTM, just a comment, where to put the error class

connectors/utils.py
artem-shelkovnikov: Yeah makes sense, will do that

PR #645, opened 2023-03-28
Remove enable_content_extraction

navarone-feekery:
## Part of 
- https://github.com/elastic/enterprise-search-team/issues/4194
- https://github.com/elastic/enterprise-search-team/issues/4108

## Summary

Remove the field `enable_content_extraction` from every connector.
This field is not required because content extraction can be handled by the index's mapping metadata.

#### Pre-Review Checklist


artem-shelkovnikov:
I think the behaviour it introduces is different though - just wanted to confirm.

I think this flag was responsible for "do not even attempt to download files from 3rd-party source", while deployment-wise flag is "yeah download files and send them to Elasticsearch, but don't run the pipeline on them" which are quire different.

I had a talk with @danajuratoni about it and I remember that we had some confusion talking about these features. So do we need a follow-up to change deployment-wide setting to also disable file downloads for the connectors?

navarone-feekery:
@artem-shelkovnikov good question, I didn't realise they were doing different things. I'm going to confirm this and get back to you.

navarone-feekery:
@artem-shelkovnikov so I've confirmed with @danajuratoni that `extract_binary_content` should drive all extraction logic, so we need to also add a check against that in the if statements where I removed the `enable_content_extraction` check.
I'll include it in this PR and re-request a review when it's ready.

artem-shelkovnikov:
@navarone-feekery is it worth to pull this logic into framework code instead, so that the authors of the connectors would not have to copy it over to their code?

connectors/sources/azure_blob_storage.py
```diff
@@ -37,17 +36,17 @@ class AzureBlobStorageDataSource(BaseDataSource):
...
 
         Args:
             configuration (DataSourceConfiguration): Object of DataSourceConfiguration class.
         """
         super().__init__(configuration=configuration)
         self.connection_string = None
-        self.enable_content_extraction = self.configuration["enable_content_extraction"]
         self.retry_count = self.configuration["retry_count"]
         self.concurrent_downloads = self.configuration["concurrent_downloads"]
+        self.content_extraction_enabled = self.pipeline["extract_binary_content"]
```
navarone-feekery, line 49: @artem-shelkovnikov using this draft to test ideas; is it okay to pass the pipeline over like this so it can access `extract_binary_content` or is there a better way to approach this?

connectors/tests/test_byoei.py
```diff
@@ -543,6 +570,20 @@ async def setup_fetcher(basic_rule_engine, existing_docs, queue, sync_rules_enab
             deleted(0),
             total_downloads(0),
         ),
+        (
+            # content_extraction_enabled is false,
+            # indexing should still work, but nothing should be downloaded
+            [],
+            [(DOC_ONE, None)],
```
artem-shelkovnikov, line 577: I think here you should pass the function for download so that you had a setup where there is a download available, but it's turned off.

artem-shelkovnikov (commented):
Looks good overall, left a small comment about the test - I think the change is not tested fully

connectors/tests/test_byoei.py
artem-shelkovnikov: Yup, that was the one, thanks!

PR #648, opened 2023-03-28
Update jira configurable fields to new standard

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/4194

Updates Jira configurable fields to have the extra field properties required for 8.8 Kibana UI.

- Remove `enable_content_extraction` which is now handled by pipeline flags ([PR](https://github.com/elastic/connectors-python/pull/645))
- Change `is_cloud` of bool type to `data_source` of str type
  - The UI will display this as a dropdown
- Add misc new field properties

#### Pre-Review Checklist


connectors/sources/jira.py
```diff
@@ -93,59 +92,75 @@ def get_default_configuration(cls):
         """
         return {
```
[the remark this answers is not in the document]
moxarth-elastic: Yes, I've recently raised an enhancement PR for the same. - https://github.com/elastic/connectors-python/pull/652

connectors/sources/jira.py
```diff
@@ -93,59 +92,75 @@ def get_default_configuration(cls):
...
-                "value": "me@example.com",
                 "label": "Jira Cloud service account id",
+                "order": 4,
                 "type": "str",
+                "value": "me@example.com",
             },
             "api_token": {
-                "value": "abc#123",
                 "label": "Jira Cloud API token",
+                "order": 5,
```
praveen-elastic, line 122: Shouldn't we keep "sensitive": True, for api_token since api_tokens are directly used as passwords by the connector in case of Cloud deployment ?

connectors/sources/jira.py
```diff
@@ -93,59 +92,75 @@ def get_default_configuration(cls):
...
             "ssl_enabled": {
-                "value": False,
-                "label": "Enable SSL verification (true/false)",
+                "display": "toggle",
+                "label": "Enable SSL verification",
+                "order": 7,
                 "type": "bool",
+                "value": False,
             },
             "ssl_ca": {
```
moxarth-elastic, line 139: IMO, we should add `required=False` for this field since it is only required when `ssl_enabled` is set to `True`.

navarone-feekery:
@moxarth-elastic @praveen-elastic thanks for the comments folks. However this is still a draft, so please don't think what you see here is a final product. I will mark it as ready for review when the PR is done.

navarone-feekery:
@praveen-elastic @moxarth-elastic this is ready for review now, to keep it simple I removed SharePoint changes. I'll do those in another PR.

connectors/sources/jira.py
navarone-feekery: This is handled by its dependency. If `ssl_enabled: True` then it should be required. If `False` then it won't be required/validated.

connectors/sources/jira.py
```diff
@@ -216,65 +220,92 @@ def get_default_configuration(cls):
             dictionary: Default configuration.
         """
         return {
-            "is_cloud": {
-                "value": True,
-                "label": "True if Jira Cloud, False if Jira Server",
-                "type": "bool",
+            "data_source": {
```
praveen-elastic, line 223: I remember @danajuratoni had a nice suggestion regarding this in design document: is it possible to control which fields to show based on the value of `data_source` ?
Since Server uses `username` and `password`, when `data_source` is Server, then `api_token` and `service_account_id` are not visible or disabled and vice-versa for Cloud.

connectors/sources/jira.py
```diff
@@ -216,65 +220,92 @@ def get_default_configuration(cls):
...
             },
             "concurrent_downloads": {
-                "value": 50,
+                "default_value": 50,
+                "display": "numeric",
                 "label": "Number of concurrent downloads for fetching attachment content",
+                "order": 10,
+                "required": False,
                 "type": "int",
+                "ui_restrictions": ["advanced"],
```
moxarth-elastic, line 307: Just curious, what does `ui_restrictions` control and what other options are there?

connectors/sources/jira.py
navarone-feekery: Yes that's what the `depends_on` field is for. If the `data_source` value is `jira_online` then only Jira Cloud inputs will appear below it. Vice versa exists for `jira_server`.

connectors/sources/jira.py
moxarth-elastic: Oh Gotcha, Thanks!

praveen-elastic (approved):

moxarth-elastic (approved):
LGTM

wangch079 (approved):

PR #649, opened 2023-03-28
[MongoDB] Advanced rules "aggregate"

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4232

This PR adds the capability for executing aggregation pipelines to the MongoDB connector. As pipeline is a non-optional parameters it's extracted and passed separately to the `aggregate` method of `Collection`. Using `deepcopy` ensures, that the original `Filter` object won't be changed.

Also add `pymongo[srv]`, which is needed to connect to a MongoDB instance using a host with the following format "mongodb+srv://...".

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -3,6 +3,7 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+from copy import deepcopy
 from datetime import datetime
 
 import fastjsonschema
@@ -188,6 +189,15 @@ class MongoDataSource(BaseDataSource):
                 find_kwargs = advanced_rules.get("find", {})
 
                 async for doc in self.collection.find(**find_kwargs):
+                    yield self.serialize(doc), None
+
+            elif "aggregate" in advanced_rules:
+                aggregate_kwargs = deepcopy(advanced_rules.get("aggregate", {}))
+                pipeline = aggregate_kwargs.pop("pipeline", [])
+
+                async for doc in self.collection.aggregate(
+                    pipeline=pipeline, **aggregate_kwargs
+                ):
                     yield self.serialize(doc), None
         else:
             async for doc in self.collection.find():
--- a/connectors/sources/tests/test_mongo.py
+++ b/connectors/sources/tests/test_mongo.py
@@ -191,6 +191,40 @@ async def test_mongo_data_source_get_docs_when_advanced_rules_find_present():
     assert find_call_kwargs[0] == filtering.get_advanced_rules().get("find")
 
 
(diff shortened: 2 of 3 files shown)


requirements/aarch64.txt
```diff
@@ -5,3 +5,4 @@ aiomysql==0.0.22
 aioboto3==9.0.0
 motor==2.5.1
 smbprotocol==1.9.0
+pymongo[srv]==3.13.0
```
timgrein, line 8: I'll create a follow-up issue here to investigate, whether we can use the same versions for `motor`.

connectors/sources/mongo.py
```diff
@@ -189,6 +190,15 @@ async def get_docs(self, filtering=None):
 
                 async for doc in self.collection.find(**find_kwargs):
                     yield self.serialize(doc), None
+
+            elif "aggregate" in advanced_rules:
+                aggregate_kwargs = deepcopy(advanced_rules.get("aggregate", {}))
```
artem-shelkovnikov, line 195: Why `deepcopy` here? Does the below code somehow modify the `aggregate_kwargs`?

requirements/aarch64.txt
artem-shelkovnikov: Hmmm how did it appear here? Should it be in a separate PR?

connectors/sources/mongo.py
timgrein: Yes see the line afterwards, also added the reason in the PR description.

requirements/aarch64.txt
timgrein: Reason is in the PR description. Needed for testing it correctly

artem-shelkovnikov: I've tested connection to our cloud MongoDB instance without this change and it has been working all along. I will test again now, but what was the error that you've encountered during testing?

timgrein: `pymongo.errors.ConfigurationError: The "dnspython" module must be installed to use mongodb+srv:// URIs. To fix this error install pymongo with the srv extra:`. Maybe we're having issues in `make clean install` and something didn't install correctly? 🤔

timgrein: Maybe the version in `aarch64.txt` is too old to include this change: https://jira.mongodb.org/browse/MOTOR-284.

timgrein: Yep, running on M1.

artem-shelkovnikov: I see. I'd separate this change to a different PR and test it on all archs. Happy to help with x86_64 testing!

timgrein: Removed it, let's if the CI runs without an error with the `3.0.0` motor version

PR #661, opened 2023-03-29
[SharePoint] Add separate client class

akanshi-elastic:
## PR Content
Includes the changes suggested by @timgrein in Confluence PR i.e. moving all low level stuffs such as generating client session, API calls to a new class called SharepointClient.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint.py
+++ b/connectors/sources/sharepoint.py
@@ -91,6 +91,213 @@ SCHEMA = {
 }
 
 
+class SharepointClient:
+    """Sharepoint client to handle API calls made to Sharepoint"""
+
+    def __init__(self, configuration):
+        self._sleeps = CancellableSleeps()
+        self.configuration = configuration
+        self.is_cloud = self.configuration["is_cloud"]
+        self.host_url = self.configuration["host_url"]
+        self.certificate = self.configuration["ssl_ca"]
+        self.ssl_enabled = self.configuration["ssl_enabled"]
+        self.retry_count = self.configuration["retry_count"]
+        self.site_collections = self.configuration["site_collections"]
+
+        self.session = None
+        self.access_token = None
+        self.token_expires_at = None
+
+        if self.ssl_enabled and self.certificate:
+            self.ssl_ctx = ssl_context(certificate=self.certificate)
+        else:
+            self.ssl_ctx = False
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def _set_access_token(self):
+        """Set access token using configuration fields"""
+        expires_at = self.token_expires_at
+        if self.token_expires_at and (not isinstance(expires_at, datetime)):
+            expires_at = datetime.fromisoformat(expires_at)  # pyright: ignore
+        if not is_expired(expires_at=expires_at):
+            return
+        tenant_id = self.configuration["tenant_id"]
+        logger.debug("Generating access token")
+        url = f"https://accounts.accesscontrol.windows.net/{tenant_id}/tokens/OAuth/2"
+        # GUID in resource is always a constant used to create access token
+        data = {
+            "grant_type": "client_credentials",
+            "resource": f"00000003-0000-0ff1-ce00-000000000000/{self.configuration['tenant']}.sharepoint.com@{tenant_id}",
+            "client_id": f"{self.configuration['client_id']}@{tenant_id}",
+            "client_secret": self.configuration["secret_id"],
+        }
+        headers = {"Content-Type": "application/x-www-form-urlencoded"}
+
+        async with aiohttp.request(
+            method="POST", url=url, data=data, headers=headers
+        ) as response:
+            json_data = await response.json()
+            self.access_token = json_data["access_token"]
+            self.token_expires_at = evaluate_timedelta(
+                seconds=int(json_data["expires_in"]), time_skew=20
+            )
+
+    def _get_session(self):
+        """Generate base client session using configuration fields
+
+        Returns:
+            ClientSession: Base client session.
+        """
+        if self.session:
+            return self.session
+        logger.info("Generating aiohttp Client Session...")
+        request_headers = {
+            "accept": "application/json",
+            "content-type": "application/json",
+        }
+        timeout = aiohttp.ClientTimeout(total=None)  # pyright: ignore
+
+        if self.is_cloud:
+            basic_auth = None
+        else:
+            basic_auth = aiohttp.BasicAuth(
+                login=self.configuration["username"],
+                password=self.configuration["password"],
+            )
+        self.session = aiohttp.ClientSession(
+            auth=basic_auth,
+            headers=request_headers,
+            timeout=timeout,
+            raise_for_status=True,
+        )
+        return self.session
+
+    async def close_session(self):
+        """Closes unclosed client session"""
+        self._sleeps.cancel()
+        if self.session is None:
+            return
+        await self.session.close()
+        self.session = None
+
+    async def api_call(self, url_name, url="", **url_kwargs):
+        """Make an API call to the Sharepoint Server/Online
+
+        Args:
+            url_name (str): Sharepoint url name to be executed.
+            url(str, optional): Paginated url for drive and list items. Defaults to "".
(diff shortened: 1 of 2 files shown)


connectors/sources/sharepoint.py
```diff
@@ -339,8 +230,7 @@ async def _api_call(self, url_name, url="", **url_kwargs):
                     exception,
                     ServerDisconnectedError,
                 ):
-                    await self.session.close()
-                    await self._generate_session()
+                    await self.close_session()
```
seanstory, line 233: Why close the session with every API call? Seems like you'd want to keep the session open if you're going to be making more api calls.

connectors/sources/sharepoint.py
```diff
@@ -406,29 +488,28 @@ def format_document(
...
-        if self.session is None:
-            await self._generate_session()
-
-        if self.ssl_enabled and (not self.ssl_ctx):
-            self.ssl_ctx = ssl_context(certificate=self.certificate)
-
         try:
             await anext(
-                self._api_call(
+                self.sharepoint_client.api_call(
```
seanstory, line 493: Does this not need to verify the results from the `PING` call at all? I can't see that this checks for a 200 or anything.

connectors/sources/sharepoint.py
```diff
@@ -569,12 +597,12 @@ async def get_list_items(self, list_id, site_url, server_relative_url):
...
         """
         file_relative_url = None
-        async for list_items_data in self.invoke_get_call(
+        async for list_items_data in self.sharepoint_client.invoke_get_call(
             site_url=site_url, param_name=LIST_ITEM, list_id=list_id
         ):
             for result in list_items_data:
                 if not result.get("Attachments"):
-                    url = f"{self.host_url}{server_relative_url}/DispForm.aspx?ID={result['Id']}&Source={self.host_url}{server_relative_url}/AllItems.aspx&ContentTypeId={result['ContentTypeId']}"
+                    url = f"{self.sharepoint_client.host_url}{server_relative_url}/DispForm.aspx?ID={result['Id']}&Source={self.sharepoint_client.host_url}{server_relative_url}/AllItems.aspx&ContentTypeId={result['ContentTypeId']}"
```
seanstory, line 605: These kinds of templates to build API path strings feel like they belong in the client. Any reason why this whole function shouldn't be in the client too?

akanshi-elastic:
> 

As all these methods mentioned above are not directly interacted with SharePoint. So that's why we are not including them in the  client class.

connectors/sources/sharepoint.py
akanshi-elastic: This will only call when `ServerDisconnectedError` error occurs.

connectors/sources/sharepoint.py
akanshi-elastic: We are using `raise_for_status` parameter while creating session which throws exception if response status is other than `200`.
that's why we are not verifying the result here

connectors/sources/sharepoint.py
akanshi-elastic: Here it is not interacting with SharePoint. It just manipulating the URL.

connectors/sources/sharepoint.py
```diff
@@ -350,6 +240,195 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+            site_url(string): site url to the sharepoint farm.
+            param_name(string): parameter name whether it is SITES, LISTS, DRIVE_ITEM, LIST_ITEM.
+            list_id(string, optional): Id of list item or drive item. Defaults to None.
+        Yields:
+            Response of the GET call.
+        """
+        skip = 0
+        next_url = ""
+        while True:
+            if param_name in [SITES, LISTS]:
```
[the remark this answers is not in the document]
akanshi-elastic: This will create the code duplicity and might affect the best practice. Let me know your thoughts on this?

connectors/sources/sharepoint.py
seanstory: 🤦 totally misread this. Thanks for the correction.

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
seanstory: It is important to consider code reuse. But there is a balance between readable code and completely DRY code. In client code I've worked on before, I've seen patterns that define helpers to make requests and to handle responses, and then have separate functions for each operation that reuse those helpers. This helps to strike that balance.

seanstory:
> As all these methods mentioned above are not directly interacted with SharePoint. So that's why we are not including them in the client class.

I don't think that's a good metric for if something belongs in a client class. If it were, there wouldn't need to be a SharepointClient class at all, because the only code _really_ communicating with Sharepoint is the underlying http transport client.

A better metric is to ask, "If someone wanted to write a library to interact with Sharepoint and perform common functions against their API, what functions would they need?"

I think they'd need:
- get_content
- get_sites
- get_list_items
- get_drive_items
- get_lists_and_items

I get that the line is blurry here between a SharepointClient and a SharepointDataSource. I'd think that anything that has business logic (filtering, validations, forming up of output documents, concurrency) should be part of the DataSource, but anything that's more generic and providing an interface for Sharepoint's APIs should be in the Client.

artem-shelkovnikov:
@akanshi-elastic happy to jump on a call to talk about this PR if needed

akanshi-elastic:
I have updated the PR comments

connectors/sources/sharepoint.py
```diff
@@ -432,151 +304,134 @@ def format_document(
...
-            "_attachment": attachment_content,
-        }
+                logger.warning(
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                )
+                await self._sleeps.sleep(RETRY_INTERVAL**retry)
 
-    async def invoke_get_call(self, site_url, param_name, list_id=None):
-        """Invokes a GET call to the SharePoint Server/Online.
+    async def invoke_list_and_drive_call(self, site_url, list_id, param_name):
```
artem-shelkovnikov, line 363: Why unite two calls into one method? Why not have `get_lists` and `get_drives`?

connectors/sources/sharepoint.py
```diff
@@ -704,33 +562,183 @@ async def get_lists_and_items(self, site):
...
+            self.sharepoint_client.ssl_enabled
+            and self.sharepoint_client.certificate == ""
+        ):
+            raise ConfigurableFieldValueError("SSL certificate must be configured.")
+
+    async def ping(self):
+        """Verify the connection with Sharepoint"""
+        try:
+            await anext(
+                self.sharepoint_client.api_call(
```
artem-shelkovnikov, line 701: Why not instead make `self.sharepoint_client.ping` method?

artem-shelkovnikov (commented):
I look at the changes and I'm not sure they're properly addressing the problem. Do you mind if we schedule a call about this change?

connectors/sources/sharepoint.py
```diff
@@ -666,71 +484,357 @@ async def get_drive_items(self, list_id, site_url, server_relative_url):
...
+        connection_fields = (
+            [
+                "host_url",
+                "site_collections",
+                "client_id",
+                "client_secret",
+                "tenant",
+                "tenant_id",
+            ]
+            if self.sharepoint_client.is_cloud
```
wangch079, line 640: `validate_config` should validate anything from `self.sharepoint_client`

akanshi-elastic: we are validating all the required fields on the basis of `is_cloud` True/False.

connectors/sources/sharepoint.py
```diff
@@ -302,7 +166,6 @@ async def _generate_session(self):
 
         if self.is_cloud:
             basic_auth = None
-            await self._set_access_token()
```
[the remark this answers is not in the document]
akanshi-elastic: We have kept these two things different because we are going to check `_get_session` for every api call and if we include access token here then this will also get check every time.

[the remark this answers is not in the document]
akanshi-elastic: I have kept  `_set_access_token` and  `get_session`  separate because for every `_api_call` we have to check `session` and `access_token`

artem-shelkovnikov (approved):

artem-shelkovnikov (approved):

PR #662, opened 2023-03-29
Allow override of connector record in the index when running functional tests

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/4247

Right now running `make ftest` attempts to populate configurable field values for the connector from the default connector configuration. It's not perfect - meaningful default values for the connector configuration are different from the values that we need to run ftest. Functional tests should not affect the design of the connector at all.

This PR allows to specify `--connector-definition` when running `kibana.py` to create a connector. It will load the connector definition from this file (for example, see `mysql/connector.json`) and insert this record into the connector index instead of creating a default one. This allows us to actually populate configurable fields in such files (connector.json) and provide meaningful defaults for these fields.

As a side effect, I deleted `--filtering` flag from `kibana.py` as it's supported by `--connector-definition` flag. I've inlined the only `filtering.json` file that existed for MySQL connector into this connector's `connector.json`.

#### Pre-Review Checklist


artem-shelkovnikov:
Tested it by running `NAME=mysql make ftest`, `NAME=mongodb make ftest` and `NAME=dir make ftest`

artem-shelkovnikov:
@navarone-feekery if there are no major objections and we merge this PR, we'll be able to change the default field values for the connectors to something sensible.

connectors/kibana.py
```diff
@@ -136,7 +133,7 @@ async def prepare(service_type, index_name, config, filtering=None):
 
     try:
         # https:#github.com/elastic/enterprise-search-team/discussions/2153#discussioncomment-2999765
-        doc = {
+        doc = connector_definition or {
```
wangch079, line 136: do you think we should do something like:
```python
connector_definition["pipeline"] = {
    "extract_binary_content": True,
    "name": "ent-search-generic-ingestion",
    "reduce_whitespace": True,
    "run_ml_inference": True,
}
```

Or we need to `put_pipeline` for the pipeline from `connector_definition`?

connectors/kibana.py
```diff
@@ -259,49 +261,31 @@ def _parser():
...
     if not os.path.exists(config_file):
         raise IOError(f"config file at '{config_file}' does not exist")
 
-    logger.info(f"Loading filters from {filtering_file}")
-    filtering = _load_filtering(filtering_file)
-
     set_logger(args.debug and logging.DEBUG or logging.INFO)
     config = EnvYAML(config_file)
+    connector_definition = None
+    if connector_definition_file and os.path.exists(connector_definition_file):
```
wangch079, line 281:
```suggestion
    if connector_definition_file and os.path.exists(connector_definition_file) and os.path.isfile(connector_definition_file):
```

wangch079 (commented):
I think we can consider making the `connector.json` a mandatory file in ftest of each connector.

connectors/kibana.py
artem-shelkovnikov: Sorry, I don't understand it - for now pipeline is both expected in `connector.json` and is present in the default definition, why do we need to override the pipeline?

artem-shelkovnikov: Hmmm maybe, but not now - I don't think we also allow changing the pipeline anywhere in UI, do we?

I thought of doing it later.

timgrein (commented):
Two nits and a question + I think something needs to be on a different level in `connector.json`

connectors/kibana.py
```diff
@@ -162,7 +159,7 @@ async def prepare(service_type, index_name, config, filtering=None):
             "created_at": None,
             # Date the connector was updated
```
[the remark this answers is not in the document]
artem-shelkovnikov: I don't think we have an official term - connector "configuration" embodies configurable fields, scheduling and filtering in my head, definition = record in the index. In reality, we don't have a glossary about it.

I'll probably avoid removing unrelated things for now, focusing mainly on functional change.

connectors/kibana.py
```diff
@@ -162,7 +159,7 @@ async def prepare(service_type, index_name, config, filtering=None):
...
             "updated_at": None,
-            "filtering": filtering,
+            "filtering": None,
```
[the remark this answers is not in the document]
artem-shelkovnikov: It's actually defaulting to a whole Filtering record, I'll adjust the PR accordingly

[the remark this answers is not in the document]
timgrein: Ah right, it probably includes the default basic rule also.

connectors/kibana.py
```diff
@@ -259,49 +261,35 @@ def _parser():
...
     config = EnvYAML(config_file)
+    connector_definition = None
+    if (
+        connector_definition_file
+        and os.path.exists(connector_definition_file)
+        and os.path.isfile(connector_definition_file)
+    ):
+        with open(connector_definition_file) as f:
+            connector_definition = json.load(f)
+
```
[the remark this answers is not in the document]
artem-shelkovnikov: Good call, done!

connectors/kibana.py
artem-shelkovnikov: > Do you have the plan to enable connectors to specify different pipelines?

Not for now at least, we haven't discussed this at all. I think as soon as we need this capability, we can add the logic! For now if someone specifies invalid pipeline it will probably fail the test (need to test it actually)

navarone-feekery (approved):
LGTM

PR #666, opened 2023-03-30
[MySQL] Parse tables string to list of tables

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4239

This PR adds a function to parse a MySQL tables string to a list of tables. It splits the string by comma. The regular expression also takes into account that a table name can contain a comma, if it's wrapped in backticks (possible in MySQL).

This change is needed for the new advanced rules structure coming in 8.8.

#### Pre-Review Checklist


seanstory (dismissed):
Gross. Does this need to be documented anywhere?

connectors/sources/mysql.py
```diff
@@ -18,6 +19,8 @@
 from connectors.sources.generic_database import WILDCARD, configured_tables, is_wildcard
 from connectors.utils import CancellableSleeps, RetryStrategy, retryable
 
+SPLIT_BY_COMMA_OUTSIDE_BACKTICKS_PATTERN = r"`(?:[^`]|``)+`|\w+"
```
tarekziade, line 22: not super important, but it's good practive to compile it:

```
SPLIT_TABLES = re.compile(r"`(?:[^`]|``)+`|\w+")
```

and then:

```
SPLIT_TABLES.findall(tables_string)
```

timgrein: Right, thank you for the hint! Adjusted it.

timgrein:
> Gross. Does this need to be documented anywhere?

@seanstory We can add it to the MySQL connector docs, that we support wrapping table names in backticks 👍

seanstory (approved):

PR #669, opened 2023-03-30
[SharePoint] - Handle throttling via Aiolimiter

akanshi-elastic:
## PR Content

- Add Aiolimiter to handle too many request error
- Add logic of `retry-after` for handling request error

#### Pre-Review Checklist


tarekziade:
Maybe a good candidate for a generic limiter at the base class level

akanshi-elastic:
This is a follow-up PR and will be open for review once this [PR](https://github.com/elastic/connectors-python/pull/661) merge.

connectors/sources/sharepoint.py
```diff
@@ -219,17 +224,18 @@ async def get_content(
...
+        async for response in self._api_call(
+            url_name=ATTACHMENT,
+            host_url=self.host_url,
+            value=site_url,
+            file_relative_url=file_relative_url,
+        ):
+            async with self.limiter:
+                async with NamedTemporaryFile(mode="wb", delete=False) as async_buffer:
+                    async for data in response.content.iter_chunked(CHUNK_SIZE):
+                        await async_buffer.write(data)
```
artem-shelkovnikov, line 236: Should `self.limiter` be here? Does `iter_chunked` do request per chunk?

akanshi-elastic: Yeah, `iter_chunked` do request per chunk.

akanshi-elastic:
> I don't think this lib looks at the sharepoint headers as described there:
> 
> https://learn.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#how-to-handle-throttling
> 
> We need to look at `Retry-After` and `RateLimit` to drive the throttling

Retry-After and RateLimit  parameters can only be retrieve when 429 too many request error occurs and for now included backoff with retry after parameter when we face too many request

connectors/sources/sharepoint.py
```diff
@@ -111,6 +112,10 @@ def __init__(self, configuration):
         self.access_token = None
         self.token_expires_at = None
 
+        self.limiter = AsyncLimiter(
```
timgrein, line 115: Should these values be "advanced" configurable fields? We could fall back to these specific values as default, if none are configured.

akanshi-elastic: We can add this in configurable fields but what if user entered value more than rate limit value?

artem-shelkovnikov:
Indeed I believe you need to update this PR as per @tarekziade 's suggestions and look at Sharepoint's headers to understand how much time needs to pass until you can make another request.

parth-elastic:
> Indeed I believe you need to update this PR as per @tarekziade 's suggestions and look at Sharepoint's headers to understand how much time needs to pass until you can make another request.

We did go through this approach already when we encountered the "TOO MANY REQUEST" error.
There are a few things that we want to highlight. 
As per our findings, 

- We agree that we should handle throttling via `rate-limit` and `retry-after` headers, out of which we have already used `retry-after` header in our code.
- The reason why we didn't use the `rate-limit` header is because we get those headers once we encounter  "TOO MANY REQUEST" error. We don't those headers if we don't encounter the "TOO MANY REQUEST" error.
- If we go with that approach we would not have the rate limit value when we initializing the limiter. We would have to wait for the exception to come then after fetching the `rate-limit` header from that we need to initialize limiter.
- So with that reason we decided to move further limiting the API calls at starting of the connector only, so we can avoid "TOO MANY REQUEST" Error.

Let us know your thoughts for the same.

artem-shelkovnikov:
"TOO MANY REQUESTS" error is there to throttle it for us, and since they send meta-information about how to de-throttle, it's valuable to use it.

Main point to me is that it eliminates our need to fine-tune throttling metrics ourselves - we just sync until we hit "TOO MANY REQUESTS", then wait using `retry-after` and `rate-limit` headers. This way if user, for example, changes deployment size and throttling policies the connector is able to automatically adapt to the limits without manual work.

Otherwise we'll either hit TOO MANY REQUESTS anyway OR we'll severely underutilize resources that we have - we can have 1000RPS, but use 500, as an example, because of the throttling settings.

connectors/sources/sharepoint.py
timgrein: I think it's possible to catch that with the `constraint` field via declarative configuration, right? So you basically specify an upper bound constraint in `get_default_configuration`

tarekziade:
@parth-elastic I wonder if this is the same behavior for Graph APIs though, as we need to move to them, see https://github.com/elastic/connectors-python/issues/760

artem-shelkovnikov:
Closing this PR as Aiolimiter does not support our needs

PR #673, opened 2023-03-31
[Sync Job] Check length of "_id" of a yielded doc

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4256

As the generation of the "_id" field is outside of the control of the framework we need to make sure, that we're not trying to ingest documents, which have an "_id" with a length > 512. This constraint is now validated and an error is logged on a violation.

Future work will add support for hashing an id, which is too long. As the hash function will probably also be outside of our control this error log will still be needed.

#### Pre-Review Checklist


artem-shelkovnikov:
> Future work will add support for hashing an id, which is too long. 

I'm not a fan of this, IMO it should be handled on the connector side to let the author of the connector have traceability of which document goes where.

artem-shelkovnikov (approved):

timgrein:
> > Future work will add support for hashing an id, which is too long.
> 
> I'm not a fan of this, IMO it should be handled on the connector side to let the author of the connector have traceability of which document goes where.

What I had in mind is to design a solution, where a connector developer has the control to decide, whether ids of docs with long ids should be hashed or not. If not it's simply rejected and an error is logged. Could be a good topic for an architecture meeting.

PR #674, opened 2023-03-31
Update SharePoint connector configurable fields

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/4194

Updates SharePoint configurable fields to have the extra field properties required for 8.8 Kibana UI.

- Remove `enable_content_extraction` which is now handled by pipeline flags ([PR](https://github.com/elastic/connectors-python/pull/645))
- Change `is_cloud` of bool type to `data_source` of str type
  - The UI will display this as a dropdown
- Add misc new field properties

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint.py
+++ b/connectors/sources/sharepoint.py
@@ -3,7 +3,7 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
-"""Sharepoint source module responsible to fetch documents from Sharepoint Server/Online.
+"""SharePoint source module responsible to fetch documents from SharePoint Server/Online.
 """
 import asyncio
 import os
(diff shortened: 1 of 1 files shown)


connectors/sources/sharepoint.py
```diff
@@ -118,71 +123,102 @@ def __init__(self, configuration):
...
         return {
-            "is_cloud": {
-                "value": False,
-                "label": "True if Sharepoint Online, False if Sharepoint Server",
-                "type": "bool",
+            "data_source": {
+                "display": "dropdown",
+                "label": "SharePoint data source",
+                "options": [
+                    {"label": SHAREPOINT_ONLINE_LABEL, "value": SHAREPOINT_ONLINE},
```
timgrein, line 136: nit: If we use them only one time I think we can inline `SHAREPOINT_ONLINE_LABEL` and `SHAREPOINT_SERVER_LABEL`. But feel free to keep it like that, I try to only use constants, if the strings appear more than once or the value is too cryptic to understand it directly (like numbers, regular expressions etc.).

timgrein (approved):
One small nit, but nothing blocking it from a merge. LGTM!

timgrein:
I always forget to check, whether it's a draft or not. Happy to re-approve, if it's "open" :)

navarone-feekery:
@timgrein thanks for the review! It is indeed a draft, I was waiting to see what feedback came in for the Jira updates before opening this one.
I'll ping you again when it's ready as there might be changes 😄

navarone-feekery:
@timgrein ready for review now

if you don't mind could you look at the Jira changes too? https://github.com/elastic/connectors-python/pull/648
The content is basically identical but it needs an ingestion approval.

wangch079 (approved):

PR #678, opened 2023-03-31
[Microsoft SQL] Update MS-SQL Connector as per Pytds library 

moxarth-rathod:
## PR Content
Update MS-SQL Connector as per Pytds library 

## Performance Stats
**before:**
 - Perf8 Report: https://drive.google.com/file/d/1HS8NOtOjVq_BHGymeQtDqzrHdXUZCBrN/view?usp=share_link

**after:**
 - Perf8 Report: https://drive.google.com/file/d/1Gj5fMl768ykzDzk8OjSQQAfd1eR4N2uo/view?usp=share_link

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -4,7 +4,6 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 """MySQL source module responsible to fetch documents from MySQL"""
-import asyncio
 import re
 
 import aiomysql
@@ -147,9 +146,13 @@ class MySQLClient:
         self.connection_pool = await aiomysql.create_pool(**connection_string)
         self.connection = await self.connection_pool.acquire()
 
+        self._sleeps = CancellableSleeps()
+
         return self
 
     async def __aexit__(self, exception_type, exception_value, exception_traceback):
+        self._sleeps.cancel()
+
         self.connection_pool.release(self.connection)
         self.connection_pool.close()
         await self.connection_pool.wait_closed()
@@ -185,7 +188,12 @@ class MySQLClient:
         async with self.connection.cursor(aiomysql.cursors.SSCursor) as cursor:
             await cursor.execute(query)
 
-            return [f"{table}_{column[0]}" for column in cursor.description]
+            return [f"{table}_{column[0]}" for column in await cursor.fetchall()]
+
+    async def get_primary_key_column_names(self, table):
+        return await self.get_column_names(
+            table, query=self.queries.table_primary_key(table)
+        )
 
     @retryable(
         retries=RETRIES,
@@ -212,6 +220,15 @@ class MySQLClient:
         async for row in self._fetchmany_in_batches(self.queries.table_data(table)):
             yield row
 
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def yield_rows_for_query(self, query):
+        async for row in self._fetchmany_in_batches(query):
+            yield row
+
     async def _fetchmany_in_batches(self, query):
         async with self.connection.cursor(aiomysql.cursors.SSCursor) as cursor:
             await cursor.execute(query)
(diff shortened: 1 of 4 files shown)


connectors/sources/generic_database.py
```diff
@@ -316,6 +320,15 @@ async def _async_connect(self, query):
...
         """Close the connection to the database server."""
+        if self.dialect == "Microsoft SQL" and os.path.exists(
+            "ssl_certificate_mssql.pem"
+        ):
+            try:
+                os.remove("ssl_certificate_mssql.pem")
+            except Exception as exception:
+                logger.warning(
+                    f"Something went wrong while removing temporary certificate file. Exception: {exception}"
+                )
```
artem-shelkovnikov, line 331: Should this logic be in `mssql.py`?

connectors/sources/generic_database.py
```diff
@@ -17,6 +18,9 @@
 
 WILDCARD = "*"
 
+# Connector will skip the below tables if it gets from the input
+TABLES_TO_SKIP = ["sysutility_ucp_configuration_internal"]
```
artem-shelkovnikov, line 22: What's the table? Is it specific to a particular DB?

connectors/sources/mssql.py
```diff
@@ -79,44 +79,53 @@ def get_default_configuration(cls):
...
+            connect_args = {
+                "cafile": "ssl_certificate_mssql.pem",
+                "validate_host": self.validate_host,
+            }
+        self.engine = create_engine(connection_string, connect_args=connect_args)
+
+    def create_pem_file(self):
+        """Create pem file for SSL Verification"""
+        pem_format = get_pem_format(key=self.ssl_ca, max_split=1)
+        with open("ssl_certificate_mssql.pem", "w") as cert:
```
artem-shelkovnikov, line 127: Should you create somewhere in `/tmp`?

Also what if you have several MSSQL connectors running together?

connectors/sources/generic_database.py
moxarth-elastic: While indexing system table `sysutility_ucp_configuration_internal` of database `msdb`. We are getting [this](https://github.com/elastic/connectors-python/issues/742) Error. So, temporarily we skip this table. If you want we can create one dictionary with the key as the database name and values as a list of tables. So we skip this tables for msdb only.

connectors/sources/mssql.py
moxarth-elastic: Sure, we will move this file to the`/tmp` folder. Also to avoid conflict we append the current time in the file name.

connectors/sources/generic_database.py
moxarth-elastic: I think we use this `TABLES_TO_SKIP` for other dialects as well. But if you want we can move this to `mssql` data source.

artem-shelkovnikov: Ideally generic sql connector should not have DB-specific logic, only facilitate it.

For example, you can have base method `tables_to_skip` and override it in each individual connector, for example.

connectors/sources/generic_database.py
moxarth-elastic: We checked that we need to keep the file existing while our connection is open with mssql. that's why we move this code to the `close` method.

artem-shelkovnikov: I understand, but you can also have `close` method in `mssql.py`

connectors/sources/generic_database.py
moxarth-elastic: Since we are planning to skip tables for MS-SQL only, we can move `TABLES_TO_SKIP = {"msdb": ["sysutility_ucp_configuration_internal"]}` to mssql.py and use it with self.tables_to_skip for better code readability. WDYT?

artem-shelkovnikov: Sure, sounds good to me!

connectors/sources/generic_database.py
moxarth-elastic: sure we will override the close method.

connectors/sources/generic_database.py
moxarth-elastic: Done

connectors/sources/mssql.py
moxarth-elastic: done, moved to `/tmp`

moxarth-elastic:
> Can you add Perf8 metrics before/after to the PR?

Sure, Currently we are generating a perf8 report. Meanwhile, you can review the changes.

moxarth-elastic:
> Looks good, what about the Perf8 reports?

Added perf8 reports.

artem-shelkovnikov (dismissed):

moxarth-elastic:
> @moxarth-elastic should this PR be backported to 8.7?

There are some changes that are not present in 8.7. So I'll raise a custom backport PR.

PR #679, opened 2023-03-31
Remove `pyodbc` library from requirements.txt

artem-shelkovnikov:
Our CI is currently blocked because `pyodbc` is not installed in Cloud docker image.

We're getting rid of this library and migrating to `pytds` here: https://github.com/elastic/connectors-python/pull/624#discussion_r1150295132

The migration is not finished yet, but soon will be. Meanwhile we need to unblock our CI by removing the dependency. On-prem SQL connectors using generic sql connector code will be broken with this change and repaired with the mentioned PR.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -36,7 +36,7 @@ native_service_types:
   - postgresql
   - oracle
   - dir
-  - mssql
+  # - mssql # temporary not working, migrating it to a different library
 
 
 # Connector client settings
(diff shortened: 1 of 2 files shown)


timgrein (approved):
LGTM, thank you for taking care of it!

tarekziade (approved):
thanks!

PR #710, opened 2023-04-03
Add configurable field validations to connector framework

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/4196

This PR adds validations to the Field class based on its `validations` field.
These are currently not called anywhere, but will be added to the `validate_config` method for each connector if this PR is approved.

#### Pre-Review Checklist


connectors/source.py
```diff
@@ -107,6 +117,111 @@ def is_empty(self):
...
+
+        if len(validation_errors) > 0:
+            raise ConfigurableFieldValueError(
+                f"Field validations: a{'; '.join(validation_errors)}"
+            )
+
+        return True
+
+    def dependencies_satisfied(self, field):
+        """Used to check if a Field has its dependencies validated.
```
timgrein, line 148:
```suggestion
        """Used to check if a Field has its dependencies satisfied.
```

connectors/source.py
```diff
@@ -107,6 +117,111 @@ def is_empty(self):
...
+
+        Returns True if all dependencies are satisfied, or no dependencies exist.
+        Retruns False if one or more dependencies are not satisfied.
```
timgrein, line 151:
```suggestion
        Returns False if one or more dependencies are not satisfied.
```

connectors/tests/test_source.py
```diff
@@ -106,6 +107,245 @@ def test_get_source_klass_dict():
...
+                        },
+                    ],
+                }
+            }
+        ),
+    ],
+)
+async def test_is_valid_when_validations_succeed_returns_true(config):
+    c = DataSourceConfiguration(config)
+    assert c.is_valid() == True
```
timgrein: 👏 nice coverage

connectors/source.py
```diff
@@ -107,6 +117,111 @@ def is_empty(self):
...
+        """
+        if len(field.depends_on) <= 0:
+            return True
+
+        satisfied = True
+
+        for dependency in field.depends_on:
+            if dependency["field"] not in self._config:
+                # cannot check dependency if field does not exist
+                logger.warn(
```
timgrein, line 161: I think that could be an `error`, right? As this only occurs, when someone provides a wrong implementation.

timgrein (commented):
Left some comments/questions, but looks good overall! Really, excited for this!

connectors/source.py
navarone-feekery: True, if we're throwing errors for everything else we should be consistently strict here.

connectors/source.py
```diff
@@ -107,6 +117,111 @@ def is_empty(self):
...
+        for _, field in self._config.items():
+            if not self.dependencies_satisfied(field):
+                # we don't validate a field if its dependencies are not met
+                next
+
+            validation_errors = validation_errors + self.validate_field(field)
+
+        if len(validation_errors) > 0:
+            raise ConfigurableFieldValueError(
+                f"Field validations: a{'; '.join(validation_errors)}"
```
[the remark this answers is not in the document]
navarone-feekery: I just noticed this before your PR, looks like you were too fast for me 😄

connectors/source.py
```diff
@@ -107,6 +117,111 @@ def is_empty(self):
...
+        label = field.label
+
+        validation_errors = []
+
+        for validation in field.validations:
+            validation_type = validation["type"]
+            constraint = validation["constraint"]
+
+            match validation_type:
+                case "less_than":
```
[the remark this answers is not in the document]
navarone-feekery: Happy to make them enums, but where do I do that in Python? In Ruby I'd make them consts at the top of the file within the class.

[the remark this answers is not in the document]
timgrein:
```
class Status(Enum):
    CREATED = "created"
    NEEDS_CONFIGURATION = "needs_configuration"
    CONFIGURED = "configured"
    CONNECTED = "connected"
    ERROR = "error"
    UNSET = None
```

That's a sample from `byoc.py`. You can just create the enum top-level inside `source.py`. We use `UNSET` to indicate, if an enum instance doesn't have a value.

connectors/source.py
timgrein: I've thought to change the log-level to `error`, but it's probably also valid to just raise an error.

navarone-feekery: Added, also I added more tests to check that this is handled appropriately.

connectors/tests/test_source.py
```diff
@@ -106,6 +109,346 @@ def test_get_source_klass_dict():
...
+        (
+            # regex
+            {
+                "email": {
+                    "type": "str",
+                    "value": "real@email.com",
+                    "validations": [
+                        {
+                            "type": ValidationTypes.REGEX.value,
+                            "constraint": "([A-Za-z0-9]+[.-_])*[A-Za-z0-9]+@[A-Za-z0-9-]+(\\.[A-Z|a-z]{2,})+",
```
timgrein, line 231: nit: The email regex could be a constant I guess as it's reused at multiple places.

timgrein (approved):
LGTM, one small nit, but nothing blocking from my side. Great stuff!

connectors/tests/test_source.py
navarone-feekery: Yeah this will be especially needed as we expand it to other connectors. I'll include it in the next PR that adds these validations to connectors.

connectors/source.py
```diff
@@ -107,6 +132,114 @@ def is_empty(self):
...
+
+        Raises ConfigurableFieldValueError if any validation errors are found,
+        otherwise returns True.
+        """
+        validation_errors = []
+
+        for _, field in self._config.items():
+            if not self.dependencies_satisfied(field):
+                # we don't validate a field if its dependencies are not met
+                logger.info(
```
tarekziade, line 146: shouldn't this be `debug` or `warning` ?

connectors/source.py
```diff
@@ -107,6 +132,114 @@ def is_empty(self):
...
+                    f"Field {field.label} was not validated because its dependencies were not met."
+                )
+                continue
+
+            validation_errors = validation_errors + self.validate_field(field)
```
tarekziade, line 151: `validation_errors.extend(self.validate_field(field))`

connectors/source.py
```diff
@@ -107,6 +132,114 @@ def is_empty(self):
     def to_dict(self):
         return dict(self._raw_config)
 
+    def is_valid(self):
+        """Validates every Field against its `validations`.
+
+        Raises ConfigurableFieldValueError if any validation errors are found,
```
tarekziade, line 138: signature nit: if the function raises on validation error, then we don't need it to return `True` as we probably won't read that value when it does not raise.

connectors/source.py
```diff
@@ -107,6 +132,114 @@ def is_empty(self):
...
+
+        for dependency in field.depends_on:
+            if dependency["field"] not in self._config:
+                # cannot check dependency if field does not exist
+                raise ConfigurableFieldDependencyError(
+                    f'Configuration `{field.label}` depends on configuration `{dependency["field"]}`, but it does not exist.'
+                )
+
+            if self._config[dependency["field"]].value != dependency["value"]:
+                satisfied = False
```
tarekziade, line 179: You can probably get ridd of the `satisfied` variable and return `False` here since the first occurrence is enough to return

connectors/source.py
tarekziade: also, if we just raise, I'd rename it to `check_valid` as `is_valid` implies it returns `True` or `False`

tarekziade:
@navarone-feekery do you mind addressing my comments in a follow-up patch?

connectors/source.py
navarone-feekery: It's not an issue if there are no dependencies, so I'll change it to `debug`.

navarone-feekery:
@tarekziade sorry I am very trigger happy with automerge, here's a PR addressing your change requests: https://github.com/elastic/connectors-python/pull/715

PR #713, opened 2023-04-03
[MySQL] Introduce basic MySQLClient class

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/4266

Why do we need this change?

Currently too many concerns are implemented in `_connect` inside `MySQLDataSource`:
- Fetching metadata of a table (column names)
- Fetching all rows from a table
- Fetching all rows from a table using pagination

This makes it very hard to adapt the code to the new advanced rules structure as it's not too easy to understand the interaction with the database server in the first place.

We also started to introduce dedicated client classes in other data sources ([Jira Client](https://github.com/elastic/connectors-python/blob/main/connectors/sources/jira.py#L64), [Confluence Client](https://github.com/elastic/connectors-python/blob/main/connectors/sources/confluence.py#L57), etc.) for similar reasons. To further align different connectors and increase the maintainability a dedicated client class for the MySQL data source is useful.

This will also loosen the coupling between the advanced rules validator and the data source as the advanced rules validator can then use the client.

This PR only moves `with_connection_pool` for now, more high-level methods breaking up `_connect` will follow in separate PRs to keep this one small.

`ssl_context` needed to be moved outside of `MySQLDataSource` as it's needed in the new client + it doesn't need to be a method inside `MySQLDataSource` (there are no dependencies on the internal state of the data source object).

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -5,7 +5,6 @@
 #
 """MySQL source module responsible to fetch documents from MySQL"""
 import re
-import ssl
 from contextlib import asynccontextmanager
 from functools import cached_property
 
@@ -23,7 +22,7 @@ from connectors.sources.generic_database import (
     configured_tables,
     is_wildcard,
 )
-from connectors.utils import CancellableSleeps, RetryStrategy, retryable
+from connectors.utils import CancellableSleeps, RetryStrategy, retryable, ssl_context
 
 SPLIT_BY_COMMA_OUTSIDE_BACKTICKS_PATTERN = re.compile(r"`(?:[^`]|``)+`|\w+")
 
@@ -139,7 +138,7 @@ class MySQLClient:
             "password": self.password,
             "db": self.db,
             "maxsize": self.max_pool_size,
-            "ssl": _ssl_context(certificate=self.ssl_certificate)
+            "ssl": ssl_context(certificate=self.ssl_certificate)
             if self.ssl_enabled
             else None,
         }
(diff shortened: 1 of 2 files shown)


connectors/sources/mysql.py
```diff
@@ -252,7 +329,7 @@ async def validate_config(self):
         strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
     )
     async def _remote_validation(self):
-        async with self.with_connection_pool() as connection_pool:
+        async with self._mysql_client.with_connection_pool() as connection_pool:
```
timgrein, line 332: This will be replaced in the future with a more high-level method. `_with_connection_pool` will probably be an internal method of `MySQLClient`. Just wanted to keep the PR small as possible for now.

connectors/sources/mysql.py
```diff
@@ -108,6 +109,71 @@ async def _remote_validation(self, advanced_rules):
...
+            yield connection_pool
+        except Exception:
+            logger.error("Failed to create connection pool for MySQL.")
+            raise
+        finally:
+            connection_pool.close()
+            await connection_pool.wait_closed()
+
+
+def _ssl_context(certificate):
```
tarekziade, line 160: why not reusign `connectors.utils.ssl_context`

timgrein: Wasn't aware that `_ssl_context` was duplicated in the first place. Will fix that

timgrein: Adjusted it, thank you for the hint 👍

connectors/sources/mysql.py
tarekziade: yeah, I don't think we need to surface two layers -- the code just needs a pool, the rest is internal implementation detail. I guess in the future that might change if one connector talks to 2 different SQL servers

PR #716, opened 2023-04-03
[MySQL] Move "ping" method to MySQLClient

timgrein:
## Follow up of https://github.com/elastic/connectors-python/pull/713

This PR moved the `ping` method to the `MySQLClient`. This changed revealed some wrong mocking in three unit tests, which are also fixed with this PR as the CI would break otherwise.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -155,6 +155,16 @@ class MySQLClient:
             connection_pool.close()
             await connection_pool.wait_closed()
 
+    async def ping(self):
+        async with self.with_connection_pool() as connection_pool:
+            try:
+                async with connection_pool.acquire() as connection:
+                    await connection.ping()
+                    logger.info("Successfully connected to the MySQL Server.")
+            except Exception:
+                logger.exception("Error while connecting to the MySQL Server.")
+                raise
+
 
 class MySqlDataSource(BaseDataSource):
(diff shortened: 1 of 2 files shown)


wangch079:
I see this pattern is used in many places:
```python
async with self.with_connection_pool() as connection_pool:
    try:
        async with connection_pool.acquire() as connection:
            ...
    except Exception:
        logger.exception("Error while connecting to the MySQL Server.")
        raise
```

Possible to extract it to `MySQLClient`?

timgrein:
> Hmmm actually I'm not super sure about the abstraction chosen - what's the benefit of having MySQL client here? It looks now like it's just a wrapper around `with_connection_pool`

@artem-shelkovnikov It will provide more high-level methods, which are easier to use later on with the new advanced rules for MySQL. The current structure makes it very hard to do the changes needed as too many concerns are mixed IMO. It should also act as an example for other SQL-like connectors. Other data sources like `Confluence` and `Jira` showed during reviews, that it's way easier to understand the code, if the interaction with the remote API is encapsulated in a client class.

timgrein:
@wangch079 Yes, that's the plan, see: https://github.com/elastic/connectors-python/pull/713#discussion_r1156018617

timgrein:
buildkite test this

PR #717, opened 2023-04-03
Remove default values for mongodb default configurable fields

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/4022

MongoDB has no default configurable field values that make sense at this point (at least, there was none in Ruby). This PR adjusts the Python implementation with Ruby implementation, removing defaults.

#### Pre-Review Checklist


connectors/sources/mongo.py
```diff
@@ -118,39 +118,29 @@ def get_default_configuration(cls):
...
+                "value": None,
             },
+            "database": {"label": "Database", "order": 2, "type": "str", "value": None},
             "direct_connection": {
                 "display": "toggle",
                 "label": "Direct connection?",
                 "order": 3,
                 "type": "bool",
-                "value": True,
+                "value": None,
```
wangch079, line 129: How is `toggle` displayed if this is `None`? There must be a value for `bool`?  cc @navarone-feekery

navarone-feekery: This is true, `bool` types can't be `None`. They need to be either `True` or `False`. Thanks for spotting that!

wangch079 (approved):

Issue #684, opened 2023-03-31
Lazy loading of connectors sources


tarekziade:
The framework loads all sources defined in the config file. That's 40MiB now. 
We should lazy load each *used* source for ripping off some RAM usage

<img width="1178" alt="Screenshot 2023-03-31 at 09 53 16" src="https://user-images.githubusercontent.com/250019/229207723-ba43cee3-cbfa-48b1-9206-958c390de82b.png">

PR #718, opened 2023-04-03
Memory saver: lazy loading of source classes

tarekziade:
## Closes  https://github.com/elastic/connectors-python/issues/684

Removes `get_source_klass_dict,` which loads all sources and imports all libs, even if they are never used.

This saves up to 40MiB of heap and counting


tarekziade:
Impact on memory

- with the patch - https://buildkite.com/elastic/connectors-python-nightly/builds/472
- without the patch - https://buildkite.com/elastic/connectors-python-nightly/builds/469

connectors/services/job_scheduling.py
```diff
@@ -24,7 +24,7 @@
 from connectors.es.index import DocumentNotFoundError
 from connectors.logger import logger
 from connectors.services.base import BaseService
-from connectors.source import get_source_klass_dict
+from connectors.source import get_source_klass, get_source_list
```
wangch079, line 27: Don't feel it's necessary to introduce these 2 methods, which just return dict keys and a dict.

artem-shelkovnikov: Sorry I did not get your comment @wangch079, what do you mean?

tarekziade: oh I see ! you're right :)

connectors/services/job_scheduling.py
```diff
@@ -23,7 +23,9 @@
 )
 from connectors.logger import logger
 from connectors.services.base import BaseService
-from connectors.source import get_source_klass_dict
+from connectors.source import get_source_klass
+
+DEFAULT_MAX_CONCURRENT_SYNCS = 1
```
wangch079, line 28: This is not needed here

tarekziade: probably a merge flake, thanks

wangch079 (approved):

PR #719, opened 2023-04-04
[POC] use a local tika service

tarekziade:
Calls a local dockerized tika service

### DIFF
--- a/connectors/sources/sharepoint.py
+++ b/connectors/sources/sharepoint.py
@@ -6,6 +6,7 @@
 """SharePoint source module responsible to fetch documents from SharePoint Server/Online.
 """
 import os
+import time
 from datetime import datetime
 from functools import partial
 from urllib.parse import urljoin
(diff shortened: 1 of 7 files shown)


connectors/sources/sharepoint.py
```diff
@@ -23,17 +22,17 @@
...
     retryable,
+    send_to_tika,
     ssl_context,
     url_encode,
 )
 
 RETRY_INTERVAL = 2
 RETRIES = 3
-FILE_SIZE_LIMIT = 10485760
+FILE_SIZE_LIMIT = -1
```
jimczi, line 35: boom

seanstory:
Before this lands, is there going to be any team discussion or RFC?
We were very intentional about NOT using local tika for Crawler, even though its available and we had been using local tika for Workplace Search for ages. And we made the same intentional pivot in connectors - wanting to use Tika NOT in the same process as our connectors. I concerned that we haven't properly revisited why we made those decisions and concluded that they no longer apply.

tarekziade:
> Before this lands, is there going to be any team discussion or RFC? We were very intentional about NOT using local tika for Crawler, even though its available and we had been using local tika for Workplace Search for ages. And we made the same intentional pivot in connectors - wanting to use Tika NOT in the same process as our connectors. I'm concerned that we haven't properly revisited why we made those decisions and concluded that they no longer apply.

This does not land. It's a POC that helps us enable some demos. There's a document about binary files where we can discuss what we should do, and your feedback there with your experience in WS will be invaluable.

PR #722, opened 2023-04-04
Add update_by_script method

wangch079:

This PR adds a `update_by_script` method to `ESIndex` class.

#### Pre-Review Checklist

### DIFF
--- a/connectors/tests/test_es_index.py
+++ b/connectors/tests/test_es_index.py
@@ -3,6 +3,8 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+from unittest.mock import AsyncMock, Mock
+
 import pytest
 from elasticsearch import ApiError, ConflictError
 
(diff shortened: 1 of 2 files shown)


timgrein:
Curious: For what do we need that?

wangch079:
>Curious: For what do we need that?

Check [here](https://docs.google.com/document/d/1D3CeQQcDZkHsuGYOpfoVoXVFphlW27FpT_Rifw52wD0/edit#heading=h.db8oh0h6nx3p). It will be used for concurrency control.

connectors/tests/test_es_index.py
```diff
@@ -182,6 +184,20 @@ async def test_update_with_concurrency_control(mock_responses):
...
+    script = {"source": ""}
+    index = ESIndex(index_name, config)
+    index.client = Mock()
+    index.client.update = AsyncMock()
+
+    await index.update_by_script(doc_id, script)
+    index.client.update.assert_awaited_once_with(
+        index=index_name, id=doc_id, script=script
+    )
+
```
navarone-feekery, line 199: (Possibly ignorant) question: I see that this test is only checking that the `self.client.update` was called, so you've set the arg `script = {"source": ""}`.
I'm guessing that it isn't possible to test if an actual script would work, right?

wangch079: >I'm guessing that it isn't possible to test if an actual script would work, right?

Could you elaborate? The test is just testing the correct params are passed to `update` function.

`if an actual script would work` sounds like it requires a real ES instance, which I think it's out of scope of unit test.

navarone-feekery (approved):
LGTM

PR #725, opened 2023-04-04
Use required validation and update ABS to new validations

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/4196

- Adds a check in validations for empty values in required fields.
  - If a field is required but the value is empty, return a validation error.
  - If a field is not required and the value is empty, skip this form of validation
- Adds usage of new validations to ABS.

#### Pre-Review Checklist


timgrein (approved):
Left a question and some small nits, otherwise LGTM

connectors/source.py
```diff
@@ -23,6 +23,24 @@
 from connectors.logger import logger
 
 
+class Defaults(Enum):
```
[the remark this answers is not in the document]
navarone-feekery: I don't think we plan to extend it, so I moved it to be a constant.

connectors/source.py
```diff
@@ -168,14 +208,26 @@ def dependencies_satisfied(self, field):
...
-                    f'Configuration `{field.label}` depends on configuration `{dependency["field"]}`, but it does not exist.'
+                    f'`{field.label}` depends on configuration `{dependency["field"]}`, but it does not exist.'
                 )
 
             if self._config[dependency["field"]].value != dependency["value"]:
                 return False
 
         return True
 
+    def is_value_empty(self, field):
```
wangch079, line 219: This method can be moved to `Field` class. `dependencies_satisfied` as well.

navarone-feekery: > This method can be moved to Field class

👍🏻 
The same is true for `validate_field`. I'll move both over.

> dependencies_satisfied as well.

`dependencies_satisfied` needs access to other configuration fields to make comparisons, so I think it needs to be here. Unless there is a way to access them from within Field?

wangch079 (approved):

connectors/source.py
navarone-feekery: I've added this to my to-do list on the epic: https://github.com/elastic/enterprise-search-team/issues/3065 (under Connectors)

PR #726, opened 2023-04-04
[MySQL] Add "get_all_table_names" to "MySQLClient"

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/4266

Note: changes with regards to the connection_pool will be made, if all new high-level methods are implemented, so this change can be achieved in a separate PR in a consistent way.

This method will replace the `fetch_all_tables` method in a follow-up PR.

#### Pre-Review Checklist
- [x]  Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -117,17 +117,18 @@ class MySQLClient:
         password,
         ssl_enabled,
         ssl_certificate,
-        db=None,
+        database=None,
         max_pool_size=MAX_POOL_SIZE,
     ):
         self.host = host
         self.port = port
         self.user = user
         self.password = password
-        self.db = db
+        self.database = database
         self.max_pool_size = max_pool_size
         self.ssl_enabled = ssl_enabled
         self.ssl_certificate = ssl_certificate
+        self.queries = MySQLQueries(self.database)
 
     @asynccontextmanager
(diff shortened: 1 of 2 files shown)


wangch079 (approved):

connectors/sources/mysql.py
```diff
@@ -159,6 +159,18 @@ async def with_connection_pool(self):
...
             await connection_pool.wait_closed()
 
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def get_all_table_names(self):
+        async with self.with_connection_pool() as connection_pool:
+            async with connection_pool.acquire() as connection:
```
artem-shelkovnikov, line 169: I just find it a bit weird that we're unable to re-use connection pool between methods. I think it _might_ get problematic. Do you think we can make sure that the client has one connection pool throughout all calls?

timgrein: Yes, I planned to add it as a field and reuse it throughout the instance methods.

artem-shelkovnikov (approved):

PR #727, opened 2023-04-04
Increase timeout for test_mem_queue

artem-shelkovnikov:
There's a problem that `test_mem_queue` fails often with a message:

```
E                   asyncio.queues.QueueFull: MemQueue has been full for 0.1005s. while timeout is 0.1s.

connectors/utils.py:266: QueueFull
```

I'm increasing timeout for now to make CI fail less often, but we really need to look into it.

### DIFF
--- a/connectors/tests/test_utils.py
+++ b/connectors/tests/test_utils.py
@@ -135,7 +135,7 @@ async def test_mem_queue_race():
 
 @pytest.mark.asyncio
 async def test_mem_queue():
-    queue = MemQueue(maxmemsize=1024, refresh_interval=0, refresh_timeout=0.1)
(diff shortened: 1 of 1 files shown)


timgrein (approved):
LGTM. I agree, that we need a better long term solution here.

artem-shelkovnikov:
Actually I thought about it and this change is permanent - problem happens if something lags slightly.

In the test `remove_data` function sleeps for 0.1 seconds and queue timeout is 0.1 seconds. So if `await asyncio.sleep(0.1)` takes _slightly_ more than 0.1second to wake up (like 0.1005 in the error message), then test fails. This change makes the problem go away UNLESS loop lags for 0.05 seconds more - but in this case it can indicate a real problem.

I can also increase timeout to 1 second but it will make the test slower cause test setup waits for the timeout.

PR #731, opened 2023-04-05
Make next_run return datetime

wangch079:

This PR made the following changes:
1. makes `next_run` function return a `datetime` instead of the number of seconds due.
2. separates job scheduling for `on-demand` sync and `scheduled` sync, which means if a job is scheduled to sync soon, and the `sync_now` flag is set as well, one `scheduled` sync and one `on-demand` sync will be created.

#### Pre-Review Checklist


connectors/services/job_scheduling.py
```diff
@@ -94,19 +95,8 @@ async def _schedule(self, connector):
...
-            if connector.sync_now
-            else JobTriggerMethod.SCHEDULED
-        )
-        await self.sync_job_index.create(
-            connector=connector, trigger_method=trigger_method
-        )
-        if connector.sync_now:
-            await connector.reset_sync_now_flag()
+        await self._on_demand_sync(connector)
+        await self._scheduled_sync(connector)
```
timgrein, line 99: Should this also be called `_on_scheduled_sync` for consistency?

timgrein (approved):
LGTM, only one small nit

connectors/services/job_scheduling.py
artem-shelkovnikov: I think `on_demand` and `scheduled` here make sense - on is not an indicator of "event", but just a part of "on demand" word combination

wangch079: `on_demand` and `scheduled` are the two values of `JobTriggerMethod` Enum.
https://github.com/elastic/connectors-python/blob/947b8734fc48f3bc760e648487ff1944ba0c6501/connectors/byoc.py#L60-L63

PR #736, opened 2023-04-05
Add update_last_sync_scheduled_at method

wangch079:

This PR adds the method `update_last_sync_scheduled_at` to the `Connector` class and this method will work in multi-instances environment, and will make sure only one instance can update `last_sync_scheduled_at` successfully.

The new method will be used in the subsequent PRs.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -503,6 +503,20 @@ class Connector(ESDocument):
         resp = await self.index.update_by_script(doc_id=self.id, script=script)
         return resp["result"] == "updated"
 
+    async def update_last_sync_scheduled_at(self, new_ts):
+        """Update last_sync_scheduled_at with update_by_script, and return True only if last_sync_scheduled_at is
+        updated."""
+        script = {
+            "lang": "painless",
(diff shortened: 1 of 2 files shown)


connectors/byoc.py
```diff
@@ -503,6 +503,20 @@ async def reset_sync_now_flag(self):
...
+        """Update last_sync_scheduled_at with update_by_script, and return True only if last_sync_scheduled_at is
+        updated."""
+        script = {
+            "lang": "painless",
+            "source": "if (ctx._source.last_sync_scheduled_at == params['old_ts']) { ctx._source.last_sync_scheduled_at = params['new_ts'] } else { ctx.op = 'noop' }",
+            "params": {
+                "old_ts": self.last_sync_scheduled_at,
+                "new_ts": new_ts,
+            },
+        }
```
navarone-feekery, line 516: Question: if `ctx._source.last_sync_scheduled_at` and `old_ts` are the same, it means another connector updated this document and the update will return `noop`, right?

If so, could you add a small comment explaining that functionality? It wasn't immediately apparent to me.

artem-shelkovnikov (commented):
I'm curious - why choose a script instead of using optimistic concurrency with version?

artem-shelkovnikov (approved):

PR #738, opened 2023-04-05
[MySQL] Implement "hash_id" for the MySQLDataSource

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4258

This PR adds an implementation for "hash_id" for MySQL as future changes will immensely increase the likelihood of violating the length constraint of elasticsearch ids. 

#### Pre-Review Checklist


connectors/sources/mysql.py
```diff
@@ -317,6 +318,11 @@ def _mysql_client(self):
     def advanced_rules_validators(self):
         return [MySQLAdvancedRulesValidator(self)]
 
+    @classmethod
```
tarekziade, line 321: why not a simple function in utils? or a regular method?

tarekziade:
This new function is not used, just added. Any reason not to use it within this patch?

connectors/sources/mysql.py
wangch079: I also think this is not necessary a connector-specific implementation: https://github.com/elastic/connectors-python/pull/721#discussion_r1158086080

timgrein: @wangch079 @tarekziade I've created a now a `hash_id` function in `utils.py`, which will also be used as default implementation inside `BaseDataSource` so you don't have to worry about it as someone building a new connector.

timgrein:
> This new function is not used, just added. Any reason not to use it within this patch?

It's used here: https://github.com/elastic/connectors-python/blob/main/connectors/sync_job_runner.py#L241

connectors/sources/mysql.py
wangch079: I meant, you don't have to implement a `hash_id` method in `BaseDataSource` class

artem-shelkovnikov (dismissed):

artem-shelkovnikov (approved):

PR #745, opened 2023-04-06
[Utils] Retryable for async generators

timgrein:
This change is needed to support retry semantics for async generators. The new `MySQLClient` will use this, when fetching rows.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -188,6 +188,22 @@ class MySQLClient:
 
             return [f"{table}_{column[0]}" for column in cursor.description]
 
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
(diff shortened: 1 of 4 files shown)


artem-shelkovnikov (approved):
One thing to note - there is a problem that `await asyncio.sleep()` is not cancellable. We use CancellableSleeps in other places for this purpose.

Nothing has to be done in this PR, just something that will need to be changed in future

timgrein:
> One thing to note - there is a problem that `await asyncio.sleep()` is not cancellable. We use CancellableSleeps in other places for this purpose.
> 
> Nothing has to be done in this PR, just something that will need to be changed in future

Adding it to my list, I think we also need some more support for being able to log/execute something on retry, specifying exceptions, which should stop retrying immediately etc.

timgrein:
buildkite test this

PR #752, opened 2023-04-11
[Confluence] Add spaces as a configurable field

praveen-elastic:
## Closes https://github.com/elastic/connectors-py/issues/659

This PR contains an enhancement suggested by Dana in a Technical Documentation. It introduces a new configuration parameter named `spaces`.

Following Changes have been added:
- Space in configuration parameters
- Default value of * to index all spaces and its content
- method to verify configured spaces are available
- corresponding tests for the method to verify spaces

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -37,12 +37,12 @@ DOWNLOAD = "download"
 SPACE_QUERY = "limit=100"
 ATTACHMENT_QUERY = "limit=100&expand=version"
 CONTENT_QUERY = (
-    "limit=100&expand=children.attachment,history.lastUpdated,body.storage,space"
+    "limit=50&expand=children.attachment,history.lastUpdated,body.storage,space"
 )
 
 URLS = {
     SPACE: "rest/api/space?{api_query}",
-    CONTENT: "rest/api/content?{api_query}",
+    CONTENT: "rest/api/content/search?{api_query}",
     ATTACHMENT: "rest/api/content/{id}/child/attachment?{api_query}",
 }
 PING_URL = "rest/api/space?limit=1"
@@ -65,6 +65,7 @@ class ConfluenceClient:
         self.configuration = configuration
         self.is_cloud = self.configuration["data_source"] == CONFLUENCE_CLOUD
         self.host_url = self.configuration["confluence_url"]
+        self.spaces = self.configuration["spaces"]
         self.ssl_enabled = self.configuration["ssl_enabled"]
         self.certificate = self.configuration["ssl_ca"]
         self.retry_count = self.configuration["retry_count"]
@@ -177,6 +178,25 @@ class ConfluenceClient:
                 )
                 break
 
+    async def verify_spaces(self):
+        """Checks if user configured spaces are available in confluence
+
+        Raises:
+            Exception: Configured unavailable spaces: <unavailable_space_keys>
+        """
+        if self.spaces == ["*"]:
+            return
+        space_keys = []
+        async for response in self.paginated_api_call(
+            url_name=SPACE, api_query=SPACE_QUERY
+        ):
+            spaces = response.get("results", [])
+            space_keys.extend([space["key"] for space in spaces])
+        if unavailable_spaces := set(self.spaces) - set(space_keys):
+            raise Exception(
+                f"Configured unavailable spaces: {', '.join(unavailable_spaces)}"
+            )
+
 
 class ConfluenceDataSource(BaseDataSource):
     """Confluence"""
@@ -253,17 +273,24 @@ class ConfluenceDataSource(BaseDataSource):
                 "type": "str",
                 "value": "http://127.0.0.1:5000",
             },
+            "spaces": {
+                "display": "textarea",
+                "label": "Confluence Space Keys",
+                "order": 7,
+                "type": "list",
+                "value": "*",
+            },
             "ssl_enabled": {
                 "display": "toggle",
                 "label": "Enable SSL verification",
-                "order": 7,
+                "order": 8,
                 "type": "bool",
                 "value": False,
             },
             "ssl_ca": {
                 "depends_on": [{"field": "ssl_enabled", "value": True}],
                 "label": "SSL certificate",
-                "order": 8,
+                "order": 9,
                 "type": "str",
                 "value": "",
             },
@@ -271,7 +298,7 @@ class ConfluenceDataSource(BaseDataSource):
                 "default_value": 3,
                 "display": "numeric",
                 "label": "Maximum retries per request",
-                "order": 9,
+                "order": 10,
                 "required": False,
                 "type": "int",
                 "ui_restrictions": ["advanced"],
@@ -281,7 +308,7 @@ class ConfluenceDataSource(BaseDataSource):
                 "default_value": 50,
                 "display": "numeric",
                 "label": "Maximum concurrent downloads",
-                "order": 10,
+                "order": 11,
                 "required": False,
                 "type": "int",
                 "ui_restrictions": ["advanced"],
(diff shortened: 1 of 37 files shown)


connectors/sources/confluence.py
```diff
@@ -37,12 +37,12 @@
 SPACE_QUERY = "limit=100"
 ATTACHMENT_QUERY = "limit=100&expand=version"
 CONTENT_QUERY = (
-    "limit=100&expand=children.attachment,history.lastUpdated,body.storage,space"
+    "limit=50&expand=children.attachment,history.lastUpdated,body.storage,space"
```
artem-shelkovnikov, line 40: Why change the limit to 50?

connectors/sources/confluence.py
```diff
@@ -37,12 +37,12 @@
...
 )
 
 URLS = {
     SPACE: "rest/api/space?{api_query}",
-    CONTENT: "rest/api/content?{api_query}",
+    CONTENT: "rest/api/content/search?{api_query}",
```
artem-shelkovnikov, line 45: What about this change too?

praveen-elastic: The search endpoint allows us to filter the contents based on spaces.

connectors/sources/confluence.py
praveen-elastic: We are now using search endpoint and the max allowed results for  search endpoint is 50.

connectors/sources/confluence.py
```diff
@@ -177,6 +178,25 @@ async def paginated_api_call(self, url_name, **url_kwargs):
                 )
                 break
 
+    async def verify_spaces(self):
```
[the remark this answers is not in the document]
praveen-elastic: Did you mean I need to override a method from framework for custom validation and move this logic to that method ? I don't see currently any method in the framework.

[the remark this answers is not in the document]
artem-shelkovnikov: Indeed, that's the way - the method is called `validate_config`, you can take it from any other connector as an example.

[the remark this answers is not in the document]
praveen-elastic: I know, but the method `validate_config` is executed before ping. This method `verify_spaces` does an API call to the Confluence instance, so logically this verification should happen after ping.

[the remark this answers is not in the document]
praveen-elastic: Hey, apologies for the misunderstanding. I agree all 3 ping, validate and get_docs should be independent.
I only wanted to convey that `verify_spaces` method is actually a part of get_docs itself as it is fetching all the spaces which is a sub task of get_docs. The method get_docs fetches spaces, pages/blogs and attachments.

connectors/sources/confluence.py
```diff
@@ -177,6 +178,25 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+            return
+        space_keys = []
+        async for response in self.paginated_api_call(
+            url_name=SPACE, api_query=SPACE_QUERY
+        ):
+            spaces = response.get("results", [])
+            space_keys.extend([space["key"] for space in spaces])
+        if unavailable_spaces := set(self.spaces) - set(space_keys):
+            raise Exception(
+                f"Configured unavailable spaces: {', '.join(unavailable_spaces)}"
```
artem-shelkovnikov, line 197: I would also change the error to something like.

Spaces "{', '.join(unavailable_spaces)}" are not available. Available spaces are: {', '.join(space_keys)}

connectors/sources/confluence.py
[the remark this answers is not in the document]
artem-shelkovnikov: Oh no worries. Both `verify_spaces` and `get_docs` can use same methods to fetch spaces, just that `verify_spaces` should provide user with actionable error. `get_docs` though is not supposed to raise errors if user entered wrong configuration.

connectors/sources/tests/test_confluence.py
```diff
@@ -298,6 +315,38 @@ async def test_close_without_client_session():
     assert source.confluence_client.session is None
 
 
+@pytest.mark.asyncio
+async def test_verify_spaces_when_space_keys_are_valid():
+    """Test verify_spaces method with valid space keys"""
```
timgrein, line 320: nit: 
```suggestion
```

IMO we should remove those comments, if they're basically having the same information as the test function name

connectors/sources/tests/test_confluence.py
```diff
@@ -298,6 +315,38 @@ async def test_close_without_client_session():
...
+        return_value=JSONAsyncMock(RESPONSE_SPACE_KEYS)
+    )
+
+    with mock.patch("aiohttp.ClientSession.get", return_value=async_response):
+        await source.confluence_client.verify_spaces()
+
+
+@pytest.mark.asyncio
+async def test_verify_spaces_when_space_keys_are_unavailable_then_raise_exception():
+    """Test verify_spaces method with unavailable space keys"""
```
timgrein, line 334:
```suggestion
```

Same as above

connectors/sources/confluence.py
```diff
@@ -524,11 +554,19 @@ async def get_docs(self, filtering=None):
         Yields:
             dictionary: dictionary containing meta-data of the content.
         """
+        await self.confluence_client.verify_spaces()
```
timgrein, line 557: As @artem-shelkovnikov already pointed I think we should remove verification logic from `get_docs`. At this point all verification/validations should be done already IMO. You could simply move that logic to `validate_config`, right? Then you can execute the declarative validation first and the "remote validation" afterwards (see [MySQL remote validation](https://github.com/elastic/connectors-python/blob/main/connectors/sources/mysql.py#L393) as an example)

connectors/sources/confluence.py
```diff
@@ -524,11 +554,19 @@ async def get_docs(self, filtering=None):
...
+        if self.confluence_client.spaces == ["*"]:
+            cql = "cql=type="
+        else:
+            quoted_spaces = "','".join(self.confluence_client.spaces)
+            cql = f"cql=space in ('{quoted_spaces}') AND type="
```
timgrein, line 562: nit: I assume `cql` stands for Confluence Query Language, but what's the actual query here? Is it used to fetch content from configured spaces? I would use a more detailed variable name to make it easier to read.

timgrein (commented):
Some nits around tests and variable names and one suggestion around validation.

praveen-elastic:
buildkite test this

connectors/sources/confluence.py
```diff
@@ -304,6 +312,34 @@ def tweak_bulk_options(self, options):
...
+        """Validates whether user input is empty or not for configuration fields.
+        Also validate, if user configured spaces are available in Confluence.
+
+        Raises:
+            Exception: Configured keys can't be empty
+        """
+        self.configuration.check_valid()
+        await self._remote_validation()
+
+    async def _remote_validation(self):
```
timgrein, line 325: I think we should add `@retryable` to the remove validation to have consistent behaviour across the different connectors.

connectors/sources/confluence.py
```diff
@@ -304,6 +312,34 @@ def tweak_bulk_options(self, options):
...
+            async for response in self.confluence_client.paginated_api_call(
+                url_name=SPACE, api_query=SPACE_QUERY
+            ):
+                spaces = response.get("results", [])
+                space_keys.extend([space["key"] for space in spaces])
+            if unavailable_spaces := set(self.spaces) - set(space_keys):
+                raise ConfigurableFieldValueError(
+                    f"Spaces '{', '.join(unavailable_spaces)}' are not available. Available spaces are: '{', '.join(space_keys)}'"
+                )
+        except Exception:
```
timgrein, line 339: I don't think we need that `except` block here as every exception from `validate_config` are caught on framework level and logged already (see: [Sync Job Runner](https://github.com/elastic/connectors-python/blob/main/connectors/sync_job_runner.py#L161))

connectors/sources/confluence.py
timgrein: Makes sense, missed that `api_call` already uses `retryable`.

timgrein (approved):
LGTM, thank you for addressing the suggestions

PR #759, opened 2023-04-11
Use default_value for value when applicable

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/4198

This change allows `Field` to return a `default_value` when its `value` is requested.
This will only happen if `required: False` and `value` is empty.

If this is approved, I will make a followup PR to set the `value` for all non-required configuration fields to `None` and add an appropriate `default_value`.

#### Pre-Review Checklist

### DIFF
--- a/connectors/tests/test_source.py
+++ b/connectors/tests/test_source.py
@@ -130,7 +130,7 @@ def test_default():
         ("bool", False, True, None, True),
     ],
 )
-def test_value_when_str_returns_correct_value(
+def test_value_returns_correct_value(
     type, required, default_value, value, expected_value
 ):
     assert (
--- a/connectors/tests/test_source.py
+++ b/connectors/tests/test_source.py
@@ -28,19 +28,37 @@ from connectors.source import (
 
 CONFIG = {
     "host": {
-        "value": "mongodb://127.0.0.1:27021",
         "label": "MongoDB Host",
         "type": "str",
+        "value": "mongodb://127.0.0.1:27021",
     },
     "database": {
-        "value": "sample_airbnb",
         "label": "MongoDB Database",
         "type": "str",
+        "value": "sample_airbnb",
     },
     "collection": {
-        "value": "listingsAndReviews",
         "label": "MongoDB Collection",
         "type": "str",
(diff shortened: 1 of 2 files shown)


connectors/tests/test_source.py
```diff
@@ -77,6 +95,56 @@ def test_default():
...
+        # lists
+        ("list", True, ["1", "2"], ["3", "4"], ["3", "4"]),
+        ("list", True, ["1", "2"], [], []),
+        ("list", True, ["1", "2"], [""], [""]),
+        ("list", True, ["1", "2"], [None], [None]),
+        ("list", True, ["1", "2"], None, None),
+        ("list", False, ["1", "2"], ["3", "4"], ["3", "4"]),
+        ("list", False, ["1", "2"], [], ["1", "2"]),
+        ("list", False, ["1", "2"], [""], ["1", "2"]),
+        ("list", False, ["1", "2"], [None], ["1", "2"]),
```
seanstory, line 124: These surprise me. In most languages, while `[]` might be considered "empty" or "falsey", `[""]` and `[null]` would not, as the list is literally not empty (has a size of 1). I'm not sure that it matters, since IDK how you could possibly set a list like this from Kibana, but if someone was using an API or programmatic access with the protocol, this could be valid, and isn't the behavior I'd expect.

seanstory (approved):
One small comment, but overall I think this is great. Fantastic test case.

artem-shelkovnikov (approved):
Awesome PR!

connectors/tests/test_source.py
navarone-feekery: I don't expect this to ever happen either, and Python doesn't consider these empty by default. I specifically wrote checks against these "empty-ish" list cases [here](https://github.com/elastic/connectors-python/blob/1267c8aebc59c2417728066e74dca80bda997cac/connectors/source.py#L109-L114) because someone could somehow add these by accident. Empty-ish lists would break things if they aren't validated properly.

That being said, this `.value` function doesn't care about validations, so it just returns whatever invalid value the user has added anyway.

Issue #606, opened 2023-03-22
Remove not useful defaults for configurable fields


danajuratoni:
## Bug Description
Not useful defaults should not be provided for configurable fields. e.g. devstoreaccount1 as account name or an account_key in Azure Blob storage. We will keep defaults where it simplifies the user experience, for fields such as `fetch_size` or `retry_count`.

Useful examples for config parameters such as the example for blob_endpoint should be provided in the documentation and once we have rich configurable fields, as tooltips.

**Preliminary list with configurable fields that don't need defaults**
ABS: account_name, account_key, blob_endpoint
Google Cloud Storage: service_account_credentials
Generic DB: host, port, username, password, database
Oracle: -
Postgres: -
MSSQL: mssql_driver
Network drive: username, password, server_ip, server_port, drive_path
etc.

## AC
- [ ] Provide only useful default values all connector clients. See preliminary list above for examples.
- [ ] Update documentation to reflect changes.
- [ ] If Rich configurable fields available, add tooltip information.

danajuratoni:
cc: @navarone-feekery

artem-shelkovnikov:
I think it's not possible until the job to make our functional tests have custom connector configuration is done.

navarone-feekery:
@danajuratoni I can look at this alongside https://github.com/elastic/enterprise-search-team/issues/4198

@artem-shelkovnikov could you point me to what functional tests need to be updated for this to work? And is someone already working on it?

danajuratoni:
@navarone-feekery just making sure you're aware of this issue, but probably it's already on your radar

navarone-feekery:
Because of time constraints I will look at this after working on Kibana, so closer to/just after FF.
I have a draft PR that is half-finished, but it's impacting ftests too much so is taking more time than I expected to complete: https://github.com/elastic/connectors-python/pull/761

danajuratoni:
@navarone-feekery Will this make it to 8.8?

navarone-feekery:
@danajuratoni because I am on SDH this week I don't think this will make it to 8.8.

If someone wants to take it over, I have a draft PR that addresses the core issue: https://github.com/elastic/connectors-python/pull/761, but only for a few connectors.
All that's required for the remaining connectors is:

- remove the `value` field from every configurable field entirely
- add an appropriate `connector.json` to the ftest directory for that connector
- ensure none of the connector ftests were broken by this change

DianaJourdan:
[...]

PR #761, opened 2023-04-12
Remove value field from connector default RCFs (batch 1)

navarone-feekery:

Remove `value` field from connector default RCFs, allowing them to be initialised as a type-safe empty value.

- mongo
- mssql
- postgresql
- oracle
- mysql

#### Pre-Review Checklist


connectors/sources/azure_blob_storage.py
```diff
@@ -105,7 +106,7 @@ def get_default_configuration(cls):
                 "validations": [
                     {"type": "less_than", "constraint": MAX_CONCURRENT_DOWNLOADS + 1}
                 ],
-                "value": MAX_CONCURRENT_DOWNLOADS,
+                "value": None,
```
tarekziade, line 109: if we always have to set `value` to `None`, what if we just remove the `value` key in that case for the default config -- seems to just be noise at that point or am I missing something ?

navarone-feekery: That's what I'll be doing! This PR is only complete for the DB ones.
(Also this is a draft please don't review it yet haha)

artem-shelkovnikov:
@navarone-feekery should we close this PR? Should we still do the changes in it?

navarone-feekery:
@artem-shelkovnikov I still want to do these changes but I didn't have time this release to finish it, can we hang onto the PR for one more release?
If I haven't finished it by then we can assume it isn't happening.

navarone-feekery:
Memo: I need to merge [a prerequisite PR](https://github.com/elastic/connectors-python/pull/1323) first before I can continue on this current PR.

artem-shelkovnikov:
> can we hang onto the PR for one more release?

Of course! Just checking if it's relevant work, since it's relevant we can keep it for now

seanstory (approved):
LGTM.

Does this need a corresponding kibana PR to remove all the `value` values for native connectors?

navarone-feekery:
@seanstory I will take a look at the native connectors at the end of this, but at a glance they initialise with type-safe empty values already. Having the `value` field is okay, just trying to remove junk data.

PR #762, opened 2023-04-12
[8.7] Backport latest changes of MS-SQL connector to 8.7

moxarth-elastic:
## Backport

Backport Pytds Changes of MS-SQL Connector into 8.7

### DIFF
--- a/connectors/sources/generic_database.py
+++ b/connectors/sources/generic_database.py
@@ -116,6 +116,8 @@ class GenericBaseDataSource(BaseDataSource):
         self.dialect = ""
         self.connection = None
         self.queries = None
+        self.certfile = None
+        self.tables_to_skip = {}
 
     @classmethod
     def get_default_configuration(cls):
@@ -196,7 +198,7 @@ class GenericBaseDataSource(BaseDataSource):
             raise Exception("Configured port has to be an integer.")
 
         if (
-            self.dialect == "Postgresql"
+            self.dialect != "Oracle"
             and self.configuration["ssl_enabled"]
             and (
                 self.configuration["ssl_ca"] == ""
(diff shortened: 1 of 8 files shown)


connectors/sources/mssql.py
```diff
@@ -73,42 +81,67 @@ def get_default_configuration(cls):
...
+                os.remove(f"/tmp/{self.certfile}")
+            except Exception as exception:
+                logger.warning(
+                    f"Something went wrong while removing temporary certificate file. Exception: {exception}"
+                )
+        if self.connection is None:
+            return
+        self.connection.close()
+
+    def create_pem_file(self):
```
tarekziade, line 140: use a named temp file, and not an explicit `/tmp/` path, and have that function return the file path, you can store for deletion

tarekziade (commented):
just a nit on tmp file

connectors/sources/mssql.py
tarekziade: added for a follow up fix https://github.com/elastic/connectors-python/issues/764

moxarth-elastic: Sure, We'll raise follow-up PR for #764. I think we can land this backport PR for now.

connectors/sources/generic_database.py
```diff
@@ -196,7 +198,7 @@ async def validate_config(self):
             raise Exception("Configured port has to be an integer.")
 
         if (
-            self.dialect == "Postgresql"
+            self.dialect != "Oracle"
```
wangch079, line 201: Where is this change coming from?

moxarth-elastic: because in current main branch we have removed `validate_config` method so it's not present in main but we need to keep this in 8.7. I have added this change in original PR but after taking pull from main branch it was removed from the original PR.

wangch079 (approved):

PR #767, opened 2023-04-13
Make user and password optional for MongoDb

artem-shelkovnikov:
Small fix for a small bug - all configurable fields are required by default, but MongoDB connector can operate on MongoDB instance without auth, so this PR is making `user` and `password` fields optional

### DIFF
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -134,13 +134,20 @@ class MongoDataSource(BaseDataSource):
                 "type": "str",
                 "value": "",
             },
-            "user": {"label": "Username", "order": 5, "type": "str", "value": ""},
+            "user": {
+                "label": "Username",
+                "order": 5,
+                "type": "str",
+                "value": "",
+                "required": False,
+            },
             "password": {
                 "label": "Password",
                 "order": 6,
                 "sensitive": True,
                 "type": "str",
                 "value": "",
+                "required": False,
             },
         }
 


connectors/sources/mongo.py
```diff
@@ -134,13 +134,20 @@ def get_default_configuration(cls):
...
                 "value": "",
             },
-            "user": {"label": "Username", "order": 5, "type": "str", "value": ""},
+            "user": {
+                "label": "Username",
+                "order": 5,
+                "type": "str",
+                "value": "",
+                "required": False,
+            },
```
navarone-feekery, line 143: If username and password are empty, MongoDB will just run without them, correct?
We could hide this behind a `depends_on` on a boolean field like `use_authentication` or something. Not required, but if you want that functionality we can have it.

navarone-feekery (approved):
LGTM! I have a question about MongoDB as well.

connectors/sources/mongo.py
seanstory: I think this is a good idea - with defaulting `use_authentication` to `true`.

artem-shelkovnikov: I hesitated to do so to allow seamless migration from Ruby to Python MongoDB connector, idk if it's a big deal.

@danajuratoni what do you think?

danajuratoni: I see username and password are not mandatory, so the functionality is already there. Since I don't expect prod env with no authentication, we're good to keep the UI changes at a minimum for now and avoid adding a new field.

timgrein (approved):
LGTM

timgrein (approved):
LGTM...again! 😄

PR #768, opened 2023-04-13
Add concurrency control to job execution

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4159

This PR adds concurrency control to job execution:
1. Only one sync job can be claimed and executed per connector.
2. A sync job can only be claimed and executed by one connector instance.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -9,6 +9,7 @@ import time
 from connectors.byoc import JobStatus
 from connectors.byoei import ElasticServer
 from connectors.es import Mappings
+from connectors.es.client import with_concurrency_control
 from connectors.es.index import DocumentNotFoundError
 from connectors.logger import logger
 
@@ -20,10 +21,6 @@ ES_ID_SIZE_LIMIT = 512
 
 
 class SyncJobRunningError(Exception):
-    pass
-
-
-class JobClaimError(Exception):
     pass
 
 
@@ -85,13 +82,16 @@ class SyncJobRunner:
     async def execute(self):
         if self.running:
             raise SyncJobRunningError(f"Sync job {self.job_id} is already running.")
-
         self.running = True
-        if not await self._claim_job():
-            logger.error(
-                f"Unable to claim job {self.job_id} for connector {self.connector_id}"
+
+        if not await self.sync_starts():
+            logger.debug(
+                f"Failed to start sync for connector {self.connector_id}, the sync is executed by another connector service."
             )
-            raise JobClaimError
+            return
+
+        await self.sync_job.claim()
+        self._start_time = time.time()
 
         try:
             self.data_provider = self.source_klass(self.sync_job.configuration)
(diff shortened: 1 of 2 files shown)


connectors/sync_job_runner.py
```diff
@@ -85,13 +82,16 @@ def __init__(
     async def execute(self):
         if self.running:
             raise SyncJobRunningError(f"Sync job {self.job_id} is already running.")
-
         self.running = True
-        if not await self._claim_job():
-            logger.error(
-                f"Unable to claim job {self.job_id} for connector {self.connector_id}"
+
+        if not await self.sync_starts():
```
artem-shelkovnikov, line 87: What does `sync_starts` mean? I read the code and can't understand it - method name is confusing.

Also the flow of the function that returns True/False confuses me too, maybe better build it on exceptions or inline the function?

wangch079: >What does sync_starts mean?

It's from the `connector.sync_starts` method. In `Connector` class, there is a `sync_starts` method (update `last_sync_status` to `in_progress`) and a `sync_done` method (update `last_sync_status` to one of the terminal statuses).

>Also the flow of the function that returns True/False confuses me too, maybe better build it on exceptions or inline the function?

The returned value indicates whether it starts the sync successfully. If we make it raise an exception, do you think we should 1) catch it in `execute` or 2) propagate it and exit `execute`?

If 1), I don't feel it's necessary to raise can catch it here. If 2), do you think this is an exception? I feel this is one of the expected results. Especially for this scenario: When there is a running job and a pending job for a connector, in each iteration, the pending job will be picked, but returned here immediately because `if self.connector.last_sync_status == JobStatus.IN_PROGRESS` is evaluated to `true`.

artem-shelkovnikov: Maybe it's worth to rename it to `start_sync` and `mark_as_done`?

As to catching - depends on which class needs to handle it. It can be an exception with text "Failed to start sync for connector {self.connector_id}, the sync is executed by another connector service." that is caught by the service executor and I think it'll be fine.

To me exceptions are the way to specify exceptional scenarios and this is one of them - you attempt to start a sync but fail cause someone else did. It's more of a way to express code/scenarios, I feel that method that returns whether it did something in expected manner or not with True/False is not expressive enough, but it might be purely taste thing to me

artem-shelkovnikov (approved):

connectors/sync_job_runner.py
artem-shelkovnikov: I think it's a great idea, thanks for the elaborate explanation too!

I'll approve this PR and it can be implemented either in the scope of this one or in the scope of a different PR

wangch079: Hi @artem-shelkovnikov , I will merge the PR first, and make the change in a subsequent PR.

wangch079: PR created: https://github.com/elastic/connectors-python/pull/788

PR #774, opened 2023-04-14
[8.7] Remove types from "native" which aren't native in this version

seanstory:
Reported by @ppf2 in [this slack thread](https://elastic.slack.com/archives/C01795T48LQ/p1681503868161219?thread_ts=1681155985.696539&cid=C01795T48LQ).

In 8.7, the only native connectors are MongoDB and MySQL. See: https://www.elastic.co/guide/en/enterprise-search/8.7/native-connectors.html

#### Pre-Review Checklist

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -29,15 +29,6 @@ service:
 native_service_types:
   - mongodb
   - mysql
-  - network_drive
-  - s3
-  - google_cloud_storage
-  - azure_blob_storage
-  - postgresql
-  - oracle
-  - dir
-  - mssql
-
 
 # Connector client settings
 #connector_id: 'changeme'


timgrein (approved):
LGTM

artem-shelkovnikov (changes requested):
It's not possible to merge this change as our functional test framework depends on connectors that are tested to be in `native_service_types`.

I tried running `NAME=dir make ftest` to run a functional test for our directory connector and it never started, so this change is breaking.

We need to do some follow-ups before we can remove connectors from this list. :(

seanstory:
> It's not possible to merge this change as our functional test framework depends on connectors that are tested to be in native_service_types.

Thanks for taking a look into this. This seems like poor design on our part. Our functional tests should not rely on the default production YAML file. If I get some free cycles, I'll see if I can decouple them. In the mean time I'll switch this to a draft pr.

wangch079:
>This seems like poor design on our part. Our functional tests should not rely on the default production YAML file.

Agree. We should provide a default `config.yml` in [connectors/sources/tests](https://github.com/elastic/connectors-python/tree/main/connectors/sources/tests), and maybe allow each connector to override it.

seanstory:
supplanted by https://github.com/elastic/connectors-python/pull/1038

PR #777, opened 2023-04-17
Add concurrency control to connector validate_filtering method

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/4161

This PR adds concurrency control to the `validate_filtering` method of `Connector` class

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -13,6 +13,7 @@ from datetime import datetime, timezone
 from enum import Enum
 
 from connectors.es import ESDocument, ESIndex
+from connectors.es.client import with_concurrency_control
 from connectors.filtering.validation import (
     FilteringValidationState,
     InvalidFilteringError,
@@ -602,7 +603,9 @@ class Connector(ESDocument):
             )
         await self.reload()
 
+    @with_concurrency_control()
     async def validate_filtering(self, validator):
+        await self.reload()
         draft_filter = self.filtering.get_draft_filter()
         if not draft_filter.has_validation_state(FilteringValidationState.EDITED):
             logger.debug(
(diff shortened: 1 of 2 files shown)


connectors/byoc.py
```diff
@@ -627,7 +630,12 @@ async def validate_filtering(self, validator):
                 if validation_result.state == FilteringValidationState.VALID:
                     filter_["active"] = filter_.get("draft")
 
-        await self.index.update(doc_id=self.id, doc={"filtering": filtering})
+        await self.index.update(
```
timgrein, line 633: I'm wondering if we could provide a method inside the client class itself handling sequence numbers. To me personally this feels more as a concern of the client and not the connector, but just thinking out loud. So we're having something like `update_with_concurrency_control` and the API looks the same as `update`, but uses concurrency control under the hood.

timgrein (approved):
Only one question, if we should put a method somewhere else, but not blocking. LGTM!

connectors/byoc.py
wangch079: >I'm wondering if we could provide a method inside the client class itself handling sequence numbers.

Did you mean the [ESIndex](https://github.com/elastic/connectors-python/blob/db83cac4a66e0687e3741da80475e6d0692ca738/connectors/es/index.py#L18) class?

> So we're having something like `update_with_concurrency_control` and the API looks the same as `update`, but uses concurrency control under the hood.

Could you elaborate?

Do you mean we should add this method to [ESIndex](https://github.com/elastic/connectors-python/blob/db83cac4a66e0687e3741da80475e6d0692ca738/connectors/es/index.py#L18) class?

```python
async def update_with_concurrency_control(self, es_document, doc):
    return await self.client.update(
        index=self.index_name,
        id=es_document.id,
        doc=doc,
        if_seq_no=es_document._seq_no,
        if_primary_term=es_document._primary_term,
    )
```

wangch079: Hi @timgrein , I will merge this PR first, we can make a separate PR if needed.

PR #780, opened 2023-04-17
[MySQL] Validate new advanced rules structure

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4238

This PR adapts the `MySQLAdvancedRulesValidator` to validate the new advanced rules structure. Similar to MongoDB it uses `fastjsonschema`. Also introduced a check, if advanced rules are empty (array or object), which is always valid.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -7,6 +7,8 @@
 import re
 
 import aiomysql
+import fastjsonschema
+from fastjsonschema import JsonSchemaValueException
 
 from connectors.filtering.validation import (
     AdvancedRulesValidator,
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov (commented):
Why have tables if `query` already does everything?

timgrein:
> Why have tables if `query` already does everything?

That's something we've agreed on with product, that we don't parse query strings as the cost is simply too high from an implementation perspective. To stay consistent in the id generation/tracing capabilities we need to include the tables information, therefore it's explicitly present in the new structure.

artem-shelkovnikov:
But you can still have join/UNION query that include multiple tables with a single table? How will it be treated?

timgrein:
> But you can still have join/UNION query that include multiple tables with a single table? How will it be treated?

```
{
  "tables": ["table_1"],
  "query": "SELECT * FROM table_1 INNER JOIN table_2 ON table_1.id = table_2.id"
}
```
You mean something like this or something different?

That'll be documented explicitly that you need to make sure, that the tables array need to match the used tables in the query. We may introduce parsing table names later, when we have more SQL connectors with a similar implementation AFAIK.

artem-shelkovnikov:
> You mean something like this or something different?

Indeed, that's the case.

What will happen with the setup that you provided?

Will all records from the query have some field marking that they come from table_1? What if you specify too tables?

artem-shelkovnikov (approved):

PR #786, opened 2023-04-17
Align aioboto3 version with x86

vidok:
Fixes this test on M1: 

```
self = s3.ServiceResource()

    def close(self):
>       return self.meta.client.close()
ERROR       AttributeError: 'S3' object has no attribute 'close'

lib/python3.10/site-packages/aioboto3/resources/base.py:17: AttributeError
ERROR    asyncio:base_events.py:1758 Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x12af3a170>
```


navarone-feekery (approved):
Nice!

PR #787, opened 2023-04-18
[SharePoint]Add Support for ingesting Site Pages

akanshi-elastic:
#### This PR contains ####

- Support for ingesting the content for site pages(includes '.aspx' filetype also)

#### Pre-Review Checklist


vidok (changes requested):

connectors/sources/sharepoint.py
```diff
@@ -241,7 +244,83 @@ async def get_content(
...
+    async def get_site_pages_content(
+        self, document, site_url, response_data, timestamp=None, doit=False
+    ):
+        """Get content of site pages for SharePoint
+        Args:
+            document (dictionary): Modified document.
+            site_url (str): Site path of sharepoint
+            response_data (dict): Dictionary of item response
+            timestamp (timestamp, optional): Timestamp of item last modified. Defaults to None.
+            doit (boolean, optional): Boolean value for whether to get content or not. Defaults to False.
```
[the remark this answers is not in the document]
akanshi-elastic: doit is handled from the framework side.

connectors/sources/sharepoint.py
```diff
@@ -44,15 +44,18 @@
...
+LIST_BY_TITLE = " list_by_title"
+TIKA_SUPPORTED_FILETYPES.append(".aspx")
 
 URLS = {
     PING: "{host_url}/sites/{site_collections}/_api/web/webs",
     SITES: "{host_url}{parent_site_url}/_api/web/webs?$skip={skip}&$top={top}",
     LISTS: "{host_url}{parent_site_url}/_api/web/lists?$skip={skip}&$top={top}&$expand=RootFolder&$filter=(Hidden eq false)",
     ATTACHMENT: "{host_url}{value}/_api/web/GetFileByServerRelativeUrl('{file_relative_url}')/$value",
-    DRIVE_ITEM: "{host_url}{parent_site_url}/_api/web/lists(guid'{list_id}')/items?$select=Modified,Id,GUID,File,Folder&$expand=File,Folder&$top={top}",
+    DRIVE_ITEM: "{host_url}{parent_site_url}/_api/web/lists(guid'{list_id}')/items?$select=*&$expand=File,Folder&$top={top}",
```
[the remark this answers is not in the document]
akanshi-elastic: For fetching the site pages content for SharePoint sever we are getting all the data in a field called as `WikiField` but this field is missing for SharePoint online and if I added `WikiField` here for server then it will throw error for online. That's why I replace it with * .Otherwise we can also manage it with two different URL where the selecting list of fields will be different for server and online

connectors/sources/sharepoint.py
```diff
@@ -238,14 +241,105 @@ async def get_content(
...
+        if response_data is None:
+            return
+
+        async with NamedTemporaryFile(mode="wb", delete=False) as async_buffer:
+            await async_buffer.write(bytes(response_data, "utf-8"))
+
+            source_file_name = async_buffer.name
+
+        await asyncio.to_thread(
+            convert_to_b64,
```
vidok, line 336: Why do you convert the result into base64? `response_data ` is a raw HTML of the rendered Site Page right?

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
akanshi-elastic: @vidok I am attaching the response for using * as well as without * , the only difference is that now we are getting these extar field in the response: `FileSystemObjectType`,`ServerRedirectedEmbedUri`,`ServerRedirectedEmbedUrl`,`ContentTypeId`,`Created`,`AuthorId`,`EditorId`,`OData__CopySource`,`CheckoutUserId`,`OData__ColorTag`,`OData__UIVersionString`,`ComplianceAssetId`,`Title`

But If we need to go with limited fields then we need to execute same API with different list of selected fields  for server and online.. As we need WikiField to be present for server to extract the aspx content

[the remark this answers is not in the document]
akanshi-elastic: [response.txt](https://github.com/elastic/connectors-python/files/11285742/response.txt)

connectors/sources/sharepoint.py
akanshi-elastic: when I try to pass the content directly to attachment field , It showing error on elastic saying: [ "Processor 'attachment' in pipeline 'ent-search-generic-ingestion' failed with message 'Illegal base64 character 3c'" ] and when I try to pass it in body directly then complete html content is indexing on elastic with the tags(attached the screenshot).  Initially with base64 conversion we were getting the data without tags attached.
Screenshot (87)
Screenshot (88)

vidok: Do you mean that passing `base64` encoded raw content into `_attachemnt` will result to a content without HTML tags in `body`?

connectors/sources/sharepoint.py
```diff
@@ -238,14 +241,105 @@ async def get_content(
...
+        async with aiofiles.open(file=source_file_name) as target_file:
             # base64 on macOS will add a EOL, so we strip() here
             attachment_content = (await target_file.read()).strip()
-        await remove(source_file_name)  # pyright: ignore
+        try:
+            await remove(source_file_name)  # pyright: ignore
+        except Exception as exception:
+            logger.warning(
+                f"Could not remove file from: {source_file_name}. Error: {exception}"
+            )
```
navarone-feekery, line 285: Nit: I don't think `from` is necessary here, unless we are talking about a directory.

```suggestion
            logger.warning(
                f"Could not remove file: {source_file_name}. Error: {exception}"
            )
```

connectors/sources/sharepoint.py
```diff
@@ -238,14 +241,105 @@ async def get_content(
...
+                url_name=LIST_BY_TITLE,
+                host_url=self.host_url,
+                parent_site_url=site_url,
+                filename=filename,
+            )
+        )
+        return response["value"][0].get("CanvasContent1")  # pyright: ignore
+
+    async def convert_file_to_b64(self, source_file_name):
+        """This method convert th file content into b64
```
navarone-feekery, line 271:
```suggestion
        """This method converts the file content into b64
```

connectors/sources/sharepoint.py
```diff
@@ -238,14 +241,105 @@ async def get_content(
...
+
+        Returns:
+            dictionary: Content document with id, timestamp & text.
+        """
+        document_size = int(document["size"])
+        filename = (
+            document["title"] if document["type"] == "File" else document["file_name"]
+        )
+        if not (doit and document_size):
+            return
```
navarone-feekery, line 313: `filename` isn't needed before checking `if not (doit and document_size)` so we should move it below the check.

```suggestion
        document_size = int(document["size"])
        if not (doit and document_size):
            return

        filename = (
            document["title"] if document["type"] == "File" else document["file_name"]
        )
```

navarone-feekery (approved):
LGTM

connectors/sources/sharepoint.py
```diff
@@ -44,15 +44,18 @@
 LIST_ITEM = "list_item"
 ATTACHMENT_DATA = "attachment_data"
 DOCUMENT_LIBRARY = "document_library"
+LIST_BY_TITLE = "list_by_title"
+TIKA_SUPPORTED_FILETYPES.append(".aspx")
```
vidok, line 48: Why do we want to send the `aspx` file to Apache Tika?

connectors/sources/sharepoint.py
```diff
@@ -774,8 +877,14 @@ async def get_docs(self, filtering=None):
         for site_url in server_relative_url:
             async for list_data in self.sharepoint_client.get_lists(site_url=site_url):
                 for result in list_data:
+                    is_site_page = False
+                    selected_field = "Modified,Id,GUID,File,Folder"
```
vidok, line 881: This might be just 2 constants right? 

```
CLOUD_SELECTED_FIELDS = "Modified,Id,GUID,File,Folder"
SERVER_SELECTED_FIELDS ="WikiField, Modified,Id,GUID,File,Folder" 
```

connectors/sources/sharepoint.py
akanshi-elastic: We are extracting the content of aspx(i.e site pages) only for sharepoint connector and if we directly update the list of TIKA_SUPPORTED_FILETYPES in `utils.py` then this will be applicable for all the connectors. That's why we are keeping it specific only for sharepoint only.

vidok: Rendered content is HTML not the .NET source code right? So if you want to extract raw text from HTML then the file is `.html` right? 

Also, could you please point it me where exactly the connector sends `.aspx` files to tika?

seanstory: > We are extracting the content of aspx(i.e site pages) only for sharepoint connector and if we directly update the list of TIKA_SUPPORTED_FILETYPES in utils.py then this will be applicable for all the connectors. That's why we are keeping it specific only for sharepoint only.

I think you're doing exposing this for all connectors anyway - you're not creating a new variable for `TIKA_SUPPORTED_FILETYPES`, you're changing the value of the global variable. As soon as this python file is loaded,  `.aspx` will be in the list of values for any connector that references this global variable.

```
>>> from connectors.utils import TIKA_SUPPORTED_FILETYPES
>>> TIKA_SUPPORTED_FILETYPES
['.txt', '.py', '.rst', '.html', '.markdown', '.json', '.xml', '.csv', '.md', '.ppt', '.rtf', '.docx', '.odt', '.xls', '.xlsx', '.rb', '.paper', '.sh', '.pptx', '.pdf', '.doc']
>>> import connectors.sources.sharepoint
>>> TIKA_SUPPORTED_FILETYPES
['.txt', '.py', '.rst', '.html', '.markdown', '.json', '.xml', '.csv', '.md', '.ppt', '.rtf', '.docx', '.odt', '.xls', '.xlsx', '.rb', '.paper', '.sh', '.pptx', '.pdf', '.doc', '.aspx']
>>> 
```

akanshi-elastic: > Rendered content is HTML not the .NET source code right? So if you want to extract raw text from HTML, the file is `.html` right?

Yes the rendered conetnt is HTML in response but the extension in filename for web pages is `.aspx`. And in our connector we are verifying the supported filetype with there name so that's why we need to add it in `TIKA_SUPPORTED_FILETYPES`
> Also, could you please point me to where exactly the connector sends `.aspx` files to tika?

this will be send in `_attachment` field along with the condition that the value in this field must be a base64 encoded 
binary
for ref: https://crestdatasystems.slack.com/archives/C02AV5C8ZDE/p1682058879139289?thread_ts=1681798035.637829&cid=C02AV5C8ZDE
https://github.com/elastic/connectors-python/blob/main/connectors/sources/sharepoint.py#L248

akanshi-elastic: > > We are extracting the content of aspx(i.e site pages) only for sharepoint connector and if we directly update the list of TIKA_SUPPORTED_FILETYPES in utils.py then this will be applicable for all the connectors. That's why we are keeping it specific only for sharepoint only.
> 
> I think you're doing exposing this for all connectors anyway - you're not creating a new variable for `TIKA_SUPPORTED_FILETYPES`, you're changing the value of the global variable. As soon as this python file is loaded, `.aspx` will be in the list of values for any connector that references this global variable.
> 
> ```
> >>> from connectors.utils import TIKA_SUPPORTED_FILETYPES
> >>> TIKA_SUPPORTED_FILETYPES
> ['.txt', '.py', '.rst', '.html', '.markdown', '.json', '.xml', '.csv', '.md', '.ppt', '.rtf', '.docx', '.odt', '.xls', '.xlsx', '.rb', '.paper', '.sh', '.pptx', '.pdf', '.doc']
> >>> import connectors.sources.sharepoint
> >>> TIKA_SUPPORTED_FILETYPES
> ['.txt', '.py', '.rst', '.html', '.markdown', '.json', '.xml', '.csv', '.md', '.ppt', '.rtf', '.docx', '.odt', '.xls', '.xlsx', '.rb', '.paper', '.sh', '.pptx', '.pdf', '.doc', '.aspx']
> >>> 
> ```
I assumed that for now we can run single connector at a time, and in that case once we add connector_id in config.yml for running a specific connector and if the connector is sharepoint only then `aspx` will append to  `TIKA_SUPPORTED_FILETYPES` . for running a different connector if we change the connector_id then it will consider it as a new sync and `aspx` will be removed from the list.
But if you want to store this in a new variable I can do this change.

seanstory: Let's either add `.aspx` directly to https://github.com/elastic/connectors-python/blob/main/connectors/utils.py#L34-L56, or simply change the file name to be `.html` instead of `.aspx`. Either is fine. But it doesn't make sense to modify this global variable here.

seanstory (approved):
Other than the TIKA_SUPPORTED_FILETYPES nit (in a thread above), this LGTM.

In my testing, this worked as expected for simple site pages:

Screenshot 2023-05-03 at 9 15 31 AM
Screenshot 2023-05-03 at 9 14 41 AM

There's room for enhancements:
- comments aren't indexed
- webpart site pages don't seem to be indexed
- news/activity content (like in Home.aspx) don't seem to be indexed

But this is still a great first pass, and we can add more functionality in future initiatives. [See this slack thread for posterity](https://elastic.slack.com/archives/C01795T48LQ/p1683123250284699).

seanstory (approved):

PR #788, opened 2023-04-18
Move last_sync_status to Job Execution Service

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4159

1. Add the `connector.last_sync_status` check to Job Execution Service, so if the connector is still syncing, don't create a `SyncJobRunner` instance for the sync job.
2. Raise `SyncJobStartError` if a `SyncJobRunner` instance fails to start a sync

Context is here: https://github.com/elastic/connectors-python/pull/768#discussion_r1165566262

#### Pre-Review Checklist

### DIFF
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -6,6 +6,8 @@
 import asyncio
 import time
 
+import elasticsearch
+
 from connectors.byoc import JobStatus
 from connectors.byoei import ElasticServer
 from connectors.es import Mappings
@@ -21,6 +23,10 @@ ES_ID_SIZE_LIMIT = 512
 
 
 class SyncJobRunningError(Exception):
+    pass
+
+
+class SyncJobStartError(Exception):
     pass
 
 
@@ -84,12 +90,7 @@ class SyncJobRunner:
             raise SyncJobRunningError(f"Sync job {self.job_id} is already running.")
         self.running = True
 
-        if not await self.sync_starts():
-            logger.debug(
-                f"Failed to start sync for connector {self.connector_id}, the sync is executed by another connector service."
-            )
-            return
-
+        await self.sync_starts()
         await self.sync_job.claim()
         self._start_time = time.time()
 
@@ -215,16 +216,23 @@ class SyncJobRunner:
     @with_concurrency_control()
     async def sync_starts(self):
         if not await self.reload_connector():
-            return False
+            raise SyncJobStartError(f"Couldn't reload connector {self.connector_id}")
 
         if self.connector.last_sync_status == JobStatus.IN_PROGRESS:
             logger.debug(
-                f"Connector {self.connector_id} is syncing, skip the job {self.job_id}..."
+                f"A sync job is started for connector {self.connector_id} by another connector instance, skipping..."
             )
-            return False
+            raise SyncJobStartError(
+                f"A sync job is started for connector {self.connector_id} by another connector instance"
+            )
 
-        await self.connector.sync_starts()
-        return True
+        try:
+            await self.connector.sync_starts()
+        except Exception as e:
+            if isinstance(e, elasticsearch.ConflictError):
+                raise
+            else:
+                raise SyncJobStartError(f"Unexpected error happened: {e}")
 
     async def prepare_docs(self):
         logger.debug(f"Using pipeline {self.sync_job.pipeline}")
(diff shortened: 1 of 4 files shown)


connectors/sync_job_runner.py
```diff
@@ -215,16 +216,23 @@ async def _sync_done(self, sync_status, sync_error=None):
...
 
-        await self.connector.sync_starts()
-        return True
+        try:
+            await self.connector.sync_starts()
+        except Exception as e:
+            if isinstance(e, elasticsearch.ConflictError):
+                raise
+            else:
+                raise SyncJobStartError(f"Unexpected error happened: {e}")
```
artem-shelkovnikov, line 235: I think it'll be more readable if you have multiple `except` instead of comparing with `isinstance`:

```
except ConflictError as e:
    raise
except Exception as e:
    raise SyncJobStartError(f"Unexpected error happened: {e}")

```

connectors/sync_job_runner.py
```diff
@@ -215,16 +216,23 @@ async def _sync_done(self, sync_status, sync_error=None):
     @with_concurrency_control()
     async def sync_starts(self):
         if not await self.reload_connector():
-            return False
+            raise SyncJobStartError(f"Couldn't reload connector {self.connector_id}")
```
artem-shelkovnikov, line 219: I think it could be a generic error - we have similar logic in multiple places, for jobs for example. We can probably have similar error-handling with error named like `EntryNotFound` or similar.

connectors/sync_job_runner.py
```diff
@@ -215,16 +216,23 @@ async def _sync_done(self, sync_status, sync_error=None):
...
 
         if self.connector.last_sync_status == JobStatus.IN_PROGRESS:
             logger.debug(
-                f"Connector {self.connector_id} is syncing, skip the job {self.job_id}..."
+                f"A sync job is started for connector {self.connector_id} by another connector instance, skipping..."
+            )
```
artem-shelkovnikov, line 224: Do you need to log like this, since the code that catches the exception probably reports error message too

connectors/sync_job_runner.py
```diff
@@ -24,6 +26,10 @@ class SyncJobRunningError(Exception):
     pass
 
 
+class SyncJobStartError(Exception):
+    pass
```
artem-shelkovnikov, line 30:
```suggestion
class SyncJobStartFailed(Exception):
    pass
```

How do you feel about such naming? I've seen that not every error in Python follows the convention SOMETHINGError, we can probably follow the practice of SOMETHINGFailed for some errors, or SOMETHINGMissing.

connectors/sync_job_runner.py
wangch079: > do you think second `raise` can be done with [exception context](https://docs.python.org/3/library/exceptions.html#exception-context)?

TIL

connectors/sync_job_runner.py
wangch079: I have no preference, but just follow the convention of exceptions defined in other places. E.g.

https://github.com/elastic/connectors-python/blob/da8a867218d90e3523624dcc6a3c278bdb767dc9/connectors/byoc.py#L65-L74

connectors/sync_job_runner.py
wangch079: It's handled by `ConcurrentTasks`, and I don't think it properly logs anything if an exception is raised:

https://github.com/elastic/connectors-python/blob/da8a867218d90e3523624dcc6a3c278bdb767dc9/connectors/utils.py#L292

artem-shelkovnikov: Ha, I think it should! I'll add an issue about it, let's keep it as is for now :)

connectors/sync_job_runner.py
artem-shelkovnikov: We can indeed fix it later for all errors, I think `Error` suffix is often redundant and not python-ish

connectors/sync_job_runner.py
wangch079: We do have `ConnectorNotFoundError` and `ConnectorJobNotFoundError` defined, and used in `check_job`:

https://github.com/elastic/connectors-python/blob/da8a867218d90e3523624dcc6a3c278bdb767dc9/connectors/sync_job_runner.py#L280-L285

But I think it makes sense to raise `SyncJobStartError` in method `sync_starts`?

artem-shelkovnikov: All depends on the catcher - if you want to distinguish errors and show different error messages - yes, otherwise probably good to keep only one

PR #793, opened 2023-04-18
Execute the oldest sync job first

wangch079:
This PR adds `sort` to `pending_jobs` of `SyncJobIndex` class, so that the oldest sync job is executed first.

#### Pre-Review Checklist

### DIFF
--- a/connectors/byoc.py
+++ b/connectors/byoc.py
@@ -62,6 +62,11 @@ class JobTriggerMethod(Enum):
     UNSET = None
 
 
+class Sort(Enum):
+    ASC = "asc"
+    DESC = "desc"
+
+
 class ServiceTypeNotSupportedError(Exception):
     pass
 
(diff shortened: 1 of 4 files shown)


connectors/byoc.py
```diff
@@ -710,7 +710,8 @@ async def pending_jobs(self, connector_ids):
                 ]
             }
         }
-        async for job in self.get_all_docs(query=query):
+        sort = [{"created_at": "asc"}]
```
timgrein, line 713: nit: "asc"/"desc" could be an enum, so we don't need to know the exact string value all the time, if we want to resue it

timgrein (approved):
One small nit, otherwise LGTM

connectors/byoc.py
wangch079: Done: https://github.com/elastic/connectors-python/pull/793/commits/572276d742c9297d59f7e582d1805f66e618e8c9

timgrein (approved):
LGTM, thanks for incorporating the change

PR #794, opened 2023-04-18
Use custom docker images for our CI

artem-shelkovnikov:
## Follow-up for https://github.com/elastic/ci-agent-images/pull/132

We'll have our own docker image to run our pipelines against. This PR makes use of this image and cleans up scripts that run tests.

Example before:

https://buildkite.com/elastic/connectors-python/builds/4574#01879df5-c172-4e72-8b53-dd43513f35ff

image

Example after:

https://buildkite.com/elastic/connectors-python/builds/4583#01879e0c-2445-4e4b-a921-0eb8dc12e4e3

image

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 agents:
   provider: "gcp"
   machineType: "n1-standard-8"
-  useVault: true
+  useVault: false
   image: family/enterprise-search-ubuntu-2204-connectors-py
 
 steps:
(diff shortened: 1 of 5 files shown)


.buildkite/pipeline.yml
```diff
@@ -1,7 +1,8 @@
 agents:
   provider: "gcp"
   machineType: "n1-standard-8"
-  useVault: false
+  useVault: true
```
vidok, line 4: Why do we need Vault?

vidok (dismissed):
Great job! This will save a lot of time in the feature.

.buildkite/pipeline.yml
artem-shelkovnikov: We don't seem to, removing

vidok (approved):

PR #798, opened 2023-04-19
Add MongoDB to nightly runs

artem-shelkovnikov:
Minor thing - we have a functional test for MongoDB, but it's not included into nightlies. This PR makes sure we run a functional test for MongoDB during Nightly QA.


seanstory (approved):
lgtm

PR #804, opened 2023-04-20
Make mongodb instance password-protected and use this password

artem-shelkovnikov:
Small problem to fix Nightly QA + `NAME=mongodb make ftest`.

Our MongoDB instance is not password-proteched, but connector recently started requiring username/password. This change updates our functional tests to setup a MongoDB instance with a password + change functional test setup to respect this password.


timgrein (approved):
LGTM

PR #805, opened 2023-04-21
Don't set connector/sync job to None when reload fails

wangch079:
In class `SyncJobRunner`, when reload of `connector` and `sync job` fail, it was set to `None`. This is not necessary, and we want to make sure that `self.sync_job` and `self.connector` are not `None`. This is because in #806, we want to log with `connector`/`sync_job` directly, e.g. 
```python
self.sync_job.info("...")
self.connector.info("...")
```

#### Pre-Review Checklist


timgrein (approved):
LGTM

PR #806, opened 2023-04-21
Log correlation

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/3809

This PR adds log correlations to any subclass of `ESDocument`:
1. Add `id`, `index name` for `Connector`.
2. Add `id`, `connector id` and `index name` for `SyncJob`.

One example:

_Without filebeat_:

Before
```
[FMWK][00:09:55][DEBUG] Scheduling is disabled
```

After
```
[FMWK][00:09:55][DEBUG] [Connector id: rz9ZoocBx7v2jizQfVFT, index name: search-mongodb] Scheduling is disabled
```

_With filebeat_

Before
```
{"@timestamp":"2023-04-21T16:10:26.112Z","log.level":"debug","message":"Scheduling is disabled","ecs":{"version":"1.6.0"},"labels":{"index_date":"2023.04.22"},"log":{"logger":"connectors","origin":{"file":{"line":54,"name":"document.py"},"function":"log_debug"},"original":"Scheduling is disabled"},"process":{"name":"MainProcess","pid":12688,"thread":{"id":140704517596736,"name":"MainThread"}},"service":{"type":"connectors-python","version":"8.8.0.0"}}
```

After
```
{"@timestamp":"2023-04-21T16:10:26.112Z","log.level":"debug","message":"Scheduling is disabled","ecs":{"version":"1.6.0"},"labels":{"connector_id":"rz9ZoocBx7v2jizQfVFT","index_date":"2023.04.22","index_name":"search-mongodb"},"log":{"logger":"connectors","origin":{"file":{"line":54,"name":"document.py"},"function":"log_debug"},"original":"Scheduling is disabled"},"process":{"name":"MainProcess","pid":12688,"thread":{"id":140704517596736,"name":"MainThread"}},"service":{"type":"connectors-python","version":"8.8.0.0"}}
```

#### Pre-Review Checklist


connectors/es/document.py
```diff
@@ -46,3 +49,27 @@ async def reload(self):
         self._seq_no = doc_source.get("_seq_no")
         self._primary_term = doc_source.get("_primary_term")
         self._source = doc_source.get("_source", {})
+
+    def debug(self, msg, *args, **kwargs):
```
timgrein, line 53: Personally I would prefer to name this methods `log_debug`, `log_info`, etc. . IMO it could be confusing seeing a method call like `sync_job.info(...)`, `connector.info(...)`, which someone could misinterpret. With `sync_job.log_info(...)` and `connector.log_info(...)` it's more obvious that the method call triggers a log. Especially for `sync_job.error()`, `connector.error()` these methods had a different meaning, right? (setting the error field).

WDYT?

connectors/es/document.py
```diff
@@ -46,3 +49,27 @@ async def reload(self):
...
+    def exception(self, msg, *args, exc_info=True, **kwargs):
+        logger.exception(self._prefix_msg(msg), *args, exc_info=exc_info, **kwargs)
+
+    def critical(self, msg, *args, **kwargs):
+        logger.critical(self._prefix_msg(msg), *args, **kwargs)
+
+    def fatal(self, msg, *args, **kwargs):
+        logger.fatal(self._prefix_msg(msg), *args, **kwargs)
+
+    def _prefix_msg(self, msg):
```
timgrein, line 74: nit: Maybe a short doc string that overriding this method is what's expected (with an example), if you want to add prefixing/tracing capabilities would be helpful.

timgrein (commented):
Overall very nice change, will greatly improve debugging. Added one suggestion around method names and I've added a small nit to add a docstring to `_prefix_msg`.

connectors/byoc.py
```diff
@@ -274,6 +274,9 @@ async def _terminate(
             doc["metadata"] = connector_metadata
         await self.index.update(doc_id=self.id, doc=doc)
 
+    def _prefix_msg(self, msg):
+        return f"[Sync Job id: {self.id}, connector id: {self.connector_id}, index name: {self.index_name}] {msg}"
```
artem-shelkovnikov, line 278: What do you think of making it shorter and truncating potentially index name?

Something like:
[search-mongodb:wYILo4cBJRjMAFbg745N][Job: wYILo4cBJRjMAFbg745N]

Also can you add an example output of service to the issue?

artem-shelkovnikov (commented):
Can you add example output of connector service to the issue?

Another concern - does it work well with cloud logging? Is it possible to provide correlation ids to cloud without making them a part of the message, so that it's possible to filter logs in Kibana by providing connector_id and such?

connectors/es/document.py
wangch079: Make perfect sense. I will update.

wangch079:
>Another concern - does it work well with cloud logging? 

It will, but as you mentioned, it will be part of the `message`.

>Is it possible to provide correlation ids to cloud without making them a part of the message, so that it's possible to filter logs in Kibana by providing connector_id and such?

This is a great idea. Will do so.

connectors/es/document.py
wangch079: Done in https://github.com/elastic/connectors-python/pull/806/commits/1d89963fe2b4e40d3018fafa15de3b2ea3df8d62

connectors/byoc.py
wangch079: Both `connector id` and `sync job id` are fixed length. `Index name` is from customers, but I don't expect customers to key in super long index name. 

I think it's also important to log the full index name.

Regarding the proposed format, it actually only saves the length of `Connector id` and `index name`. IMO, it's worth to keep them to make the log message more understandable.

timgrein (dismissed):
LGTM, thank you the changes. To me it looks like that @artem-shelkovnikov's suggestions were also applied providing labels for filtering in Kibana. But you can still wait for his approval, if you want 👍

wangch079:
>Sometimes data is duplicated, see line - job id is there twice:

Didn't notice this. Will update.

>I thought that correlation id is hierarchical? So Fetcher, Bulker, Connector itself and all the other log lines that were logged while running this particular job should have the correlation message?

I also observed this before PTO, and I think the only way is to pass a job instance to `ElasticServer` class, just for logging purpuse (Although I don't like it).

wangch079:
>I also observed this before PTO, and I think the only way is to pass a job instance to ElasticServer class, just for logging purpuse (Although I don't like it).

This change is not trivial, and I prefer to address it in a separate PR. 

Before proceeding with the change, I'd like to discuss another place where log correlation doesn't apply: the connector implementation. We can also pass the sync job object to `get_docs` of each connector implementation, but that's a huge change and requires framework-level change. cc @elastic/ingestion-team

artem-shelkovnikov:
Let's maybe make a meeting about it and discuss?

connectors/es/document.py
```diff
@@ -46,3 +49,75 @@ async def reload(self):
         self._seq_no = doc_source.get("_seq_no")
         self._primary_term = doc_source.get("_primary_term")
         self._source = doc_source.get("_source", {})
+
+    def log_debug(self, msg, *args, **kwargs):
```
tarekziade, line 53: You can probably reduce the code here by implementing a single method in a wrapper class

```
class DocumentLogger:
    def __init__(self, logger):
        self._logger = logger
        setattr(self, f'{level}', functools.partial(self._log,level))   

    def _log(self, level, msg, *args, **kwargs):
       func = getattr(logger, level.lower())
       func(
            msg,
            *args,
            prefix=self._prefix(),  
            extra=self._extra(),
            **kwargs,
        )
```

and then in you code:

```
self.logger = DocumentLogger(logger)
```

so you can use:

```
self.logger.debug('message')
```

wangch079: I actually had similar implementation:
At [the end of __init__](https://github.com/elastic/connectors-python/blob/86c030afb61ef51c4f1447c1c9af9dfd62285553/connectors/es/document.py#L32) of `ESDocument`, add:

```python
for level in [
    "debug",
    "info",
    "warning",
    "error",
    "exception",
    "critical",
    "fatal",
]:
    setattr(self, f"log_{level}", functools.partial(self._log, level))
```

and add the same `_log` method as you proposed. But the problem is that pyright will complain when I call something like `connector.log_debug(...`,:
```shell
/Users/chenhui.wang/src/github.com/elastic/connectors-python/connectors/byoc.py:509:18 - error: Cannot access member "log_debug" for type "Connector"
```

I have to add `pyright: ignore` every time I call `log_*`. I'd rather explicitly define the method for each job level.

wangch079:
Hi @tarekziade , need your insights here. I can apply the correlation in `byoei.py` in a separate PR, but I don't feel right to pass the `sync_job` object to each connector implementation.

wangch079:
>One thing we could do is create per-sync logger objects:

I think we want to keep one global logger? 

Probably we can have a wrapper of the global `logger` in `SyncJobRunner` class, and pass it to `ElasticServer` class and source implementations. In class `SyncJobRunner`, `ElasticServer` and the source implementation, e.g. [mysql.py](https://github.com/elastic/connectors-python/blob/main/connectors/sources/mysql.py), logging is like:

```python
class SyncJobRunner:
    def __init__():
        ...
        self._logger = SyncJobRunner(sync_job=self.sync_job)


    def execute():
        self._logger.info("Sync Job starts...")

        self.elastic_server = ElasticServer(elastic_config=self.es_config, logger_=self._logger)

        ...


class ElasticServer:
    def __init(elastic_config, logger_):
        ...
        self._logger = logger_

    async def async_bulk():
        self._logger.info("Async bulk starts...")
```

And in other places(i.e. Job scheduling service, Job execution service), the current implementation is still used (i.e. `connector.log_info(...)`

If you all agree with this approach, we can merge this one, and I will start the sync job runner logger implementation in a separate PR. cc @tarekziade @artem-shelkovnikov @timgrein

tarekziade:
> > One thing we could do is create per-sync logger objects:
> 
> I think we want to keep one global logger?

it would not go away -- a per-sync logger object would just carry extra metadata 


> 
> Probably we can have a wrapper of the global `logger` in `SyncJobRunner` class, and pass it to `ElasticServer` class and source implementations. In class `SyncJobRunner`, `ElasticServer` and the source implementation, e.g. [mysql.py]

yeah that wrapper is what I had in mind. 

(https://github.com/elastic/connectors-python/blob/main/connectors/sources/mysql.py), logging is like:
> 
> ```python
> class SyncJobRunner:
>     def __init__():
>         ...
>         self._logger = SyncJobRunner(sync_job=self.sync_job)

I am confused by this line. did you really mean `SyncJobRunner` ?

wangch079:
>it would not go away -- a per-sync logger object would just carry extra metadata

Agree, if we implement the custom logger as a wrapper, we will only have one global logger, and multiple custom loggers which holds the reference to the global logger.

> I am confused by this line. did you really mean `SyncJobRunner` ?

My bad. It should be `SyncJobRunnerLogger`. Updated in the original comment.

wangch079:
Hi @tarekziade , I eventually added the logger wrapper `DocumentLogger` as you suggested: https://github.com/elastic/connectors-python/pull/806/commits/0de01eb0f494791f2c5772109860f0abbdbdf2a1

In this commit, I repeatedly defined `info`, `debug`, `warning`,... in `DocumentLogger`, to get rid of pyright warnings. 
I also kept the methods `log_debug`, `log_info`,... in `ESDocument` class, because I think it's more convenient to call `connector.log_info(...)` than `connector.logger.info(...)`.

I also applied log correlation to `byoei.py`(https://github.com/elastic/connectors-python/pull/806/commits/fbd11f94e22046e4fc21140d855329ae0710e651) and all the source implementations(https://github.com/elastic/connectors-python/pull/806/commits/065594bd33ef2102a0d1f656d82128baa7c440ee)

tarekziade:
> > it would not go away -- a per-sync logger object would just carry extra metadata
> 
> Agree, if we implement the custom logger as a wrapper, we will only have one global logger, and multiple custom loggers which holds the reference to the global logger.
> 
> > I am confused by this line. did you really mean `SyncJobRunner` ?
> 
> My bad. It should be `SyncJobRunnerLogger`. Updated in the original comment.

nice! yeah, Maybe you don't even need a custom class for this, could be a simple function that adds the metadata on the fly

```
self._logger = create_logger(sync_id=self.sync_id, etc..)
```

tarekziade (dismissed):
Nice!

wangch079:
This PR doesn't introduce significant changes to the framework, and I just manually triggered a nightly QA and it passed: https://buildkite.com/elastic/connectors-python-nightly/builds/625.

I feel it's safe to backport to 8.8. WDYT? @tarekziade @danajuratoni

seanstory:
It's a lot of touched files, after FF, and isn't fixing a bug.
HOWEVER, it will greatly improve the lives of our support team, and would be great to have as connectors GA. 
Have you manually tested too? Looked through the logs to be sure there aren't unexpected stack traces? ECS still works? Are there any outstanding bugfix PRs that this will conflict with that could cause an issue? If you have full confidence in this PR, I think I'd be supportive of sneaking it in. But also think it would be ok to wait until 8.8.1 if you're a little hesitant.

Just my 2 cents. :)

wangch079:
I have manually tested MySQL but not all connectors, that's why I triggered a nightly qa.

ECS worked when I tested MySQL.

>Are there any outstanding bugfix PRs that this will conflict with that could cause an issue?

Once this is merged, we should broadcast to the team and crest, to log properly. Yes, this may affect some opening PRs.

wangch079:
The `DocumentLogger` accepts `prefix` and `extra`, instead of `document`:
https://github.com/elastic/connectors-python/blob/6bf12003808ccf39832fbf4daded086800afdf68/connectors/es/document.py#L14-L16

If the `prefix` or `extra` includes a field `foo`, and the value changes, it won't be updated if you log it with `DocumentLogger`. This is because `document.logger` will be passed to source implementations, and I don't want it to hold a reference of the document object. For now, all fields in the `prefix` and `extra` are not going to change: `connector_id`, `sync_job_id` and `service_type`.

artem-shelkovnikov:
Also - how do you feel if we color log lines differently?

For example, we can color the correlation part differently based on `connector_id` to visually distinguish different connectors running together.

wangch079:
> Also - how do you feel if we color log lines differently?
> 
> For example, we can color the correlation part differently based on `connector_id` to visually distinguish different connectors running together.

For Cloud deployment, we add custom labels, so customers can filter by connector_id in Kibana logging. So I think this only applies to on-prem deployment.

It's not easy to log in different colors for different connectors, you need to implement a custom logger and hold a connector to color map. And I don't feel it worth the effort for now.

tarekziade:
> > I've also noticed that filtering validation does not contain correlation, will it be implemented as a part of this PR?
> 
> Done in commit [5243fc6](https://github.com/elastic/connectors-python/commit/5243fc65ee78fe4a66a3508d9fef492c8d92e5f0)
> 
> I feel this is going out of control. I don't like a logger being passed from object to object 😰

You can also keep the loggers in a central factory, remove it from signatures

seanstory:
> I don't like a logger being passed from object to object 😰

FWIW, this is a pattern I've seen used consistently before.

Another I've seen used frequently is some sort of thread-local (or job-local?) LoggerFactory.

The way you've implemented this feels more straightforward, IMO.

wangch079:
>Another I've seen used frequently is some sort of thread-local (or job-local?) LoggerFactory.

ThreadLocal is the idea comes to my mind, then I realized we don't use multi-threading but asyncio 😅

tarekziade:
> > Another I've seen used frequently is some sort of thread-local (or job-local?) LoggerFactory.
> 
> ThreadLocal is the idea comes to my mind, then I realized we don't use multi-threading but asyncio 😅

We're single threaded. If your goal is to keep a global track of all loggers, you can use a factory and a mapping.
I'd be curious to know what would be the API to get back the right logger ?

```
get_logger(connector_id) ?
```

seanstory:
yep, that's about what I was thinking. But again, that's one option. I don't think what you've done here is bad. Passing loggers at initialize is a pretty common pattern, usually combined with a decorator pattern.

tarekziade:
> yep, that's about what I was thinking. But again, that's one option. I don't think what you've done here is bad. Passing loggers at initialize is a pretty common pattern, usually combined with a decorator pattern.

Yea., agreed, as long as we keep the same signature for all loggers, we're ok -- the benefit of passing it along is that we can add extra metadata along the way without having to worry about how a piece of code can get back the correct instance of loggers -- unless that is obvious

Notice that the logger API in python have a built-in function to get an existing logger instance, `getLogger`

```
>>> import logging
>>> my_logger = logging.getLogger('some_id')
>>> my_logger_again = logging.getLogger('some_id')
>>> my_logger is my_logger_again
True
```

So if you know what to put for `some_id` and that does not make your life complicated, you could do this and not alter signatures

wangch079:
> Notice that the logger API in python has a built-in function to get an existing logger instance, `getLogger`

The `logging` keeps a loggerDict, which means it keeps multiple logger objects.

> So if you know what to put for `some_id` and that does not make your life complicated, you could do this and not alter signatures

It means passing `some_id` around, something related to `sync_job_id`.

wangch079:
Hi @tarekziade @seanstory @artem-shelkovnikov , do you think we can merge it for 8.8? or we can hold this and have a better solution in later release?

artem-shelkovnikov:
My opinion is mild "Let's reassess it in 8.9", if someone has strong reasons to merge this PR and improve later I'm good with it

DianaJourdan:
Lots of good opinions have been shared in this PR about how this can be implemented. It looks like we have the choice to merge Chenhui's PR in 8.8 ( my understanding is the risk should be very low - please correct me if I am wrong) and get some correlations or postpone this work for later. Looking at how 8.9 release looks like, I do not expect there will be a lot more time for this. How would you all feel about getting this iteration merged and having a ticket for 8.10 to follow up? cc @danajuratoni isn't this also part of Starbuck's requirements?

tarekziade:
> How would you all feel about getting this iteration merged and having a ticket for 8.10 to follow up?

+1 let's get the ball rolling and improve gradually

connectors/byoei.py
```diff
@@ -98,6 +99,7 @@ def __init__(
         self.indexed_document_count = 0
         self.indexed_document_volume = 0
         self.deleted_document_count = 0
+        self._logger = logger_ or logger
```
tarekziade, line 102: I would keep a single style across the board, so `logger_` or `_logger`

connectors/es/document.py
```diff
@@ -3,10 +3,86 @@
...
+        self._extra = extra
+
+    def isEnabledFor(self, level):
+        return logger.isEnabledFor(level)
+
+    def debug(self, msg, *args, **kwargs):
+        logger.debug(
+            msg,
+            *args,
+            prefix=self._prefix,  # pyright: ignore
```
tarekziade, line 25: pyright is right, you can't pass `prefix` like this, it will error out. you want to create a new formatter

tarekziade (commented):
I don't think this is going to work because you are introducing a prefix and you need to create a new formatter for this. I will add my thoughts in the issue

tarekziade:
I think we need to step back, we're doing too much work here.

We are discussing creating loggers with a specific prefix and extra info here.

for adding default extras, you can use an adapter -- see https://docs.python.org/3/library/logging.html#loggeradapter-objects and then adapt the prefix

something like:

```
logger_with_extra = logging.LoggerAdapter(logger, extra)
logger_with_extra.handlers[0].setFormatter(_formatter("Worker ID"))
```

and this can be added in functions in the logger module

tarekziade:
> > How would you all feel about getting this iteration merged and having a ticket for 8.10 to follow up?
> 
> +1 let's get the ball rolling and improve gradually

The current code does not work, let's make sure we fix it first (no need to add new classes, we can use an adapter -- sorry I should have pointed that earlier)

connectors/byoei.py
wangch079: `logger_` is to shadow the `logger` in outer space, and `_logger` indicates it's a private variable.

connectors/es/document.py
wangch079: pyright complains `No parameter named "prefix"`. Maybe what I should do is:
```python
def debug(self, msg, *args, **kwargs):
    kwargs["prefix"] = self._prefix
    kwargs["extra"] = self._extra
    logger.debug(
        msg,
        *args,
        **kwargs,
    )
```

But when I test the current code, it works

danajuratoni:
>How would you all feel about getting this iteration merged and having a ticket for 8.10 to follow up? cc @danajuratoni isn't this also part of Starbuck's requirements?

This requirement is part of making features GA. With 2 GA native connectors in 8.8 and 1 more GA connector coming in 8.8, I 100% support the initiative to merge a first iteration for the ACs in the ticket. Any customer that will want to use connectors in production need this, including Starbucks.

We'll want to constantly improve this so a second iteration in 8.10 should be fine. Is it clear what functionality would fall in the second iteration?

connectors/es/document.py
tarekziade:
```
>>> import logging
>>> logger = logging.getLogger('test')
>>> kw = {'prefix': 'ok'}
>>> logger.warn('message', **kw)
<stdin>:1: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 1506, in warn
    self.warning(msg, *args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 1501, in warning
    self._log(WARNING, msg, args, **kwargs)
TypeError: Logger._log() got an unexpected keyword argument 'prefix'
```

wangch079: Hi @tarekziade , I also made this change: https://github.com/elastic/connectors-python/pull/806/files#diff-a34ad6d1aa661c54359acc9e3e994d49e5b7b2d4f30f8db995682fd5d3f436c1

That's why it works.

wangch079:
> for adding default extras, you can use an adapter -- see https://docs.python.org/3/library/logging.html#loggeradapter-objects and then adapt the prefix
> 
> something like:
> 
> ```
> logger_with_extra = logging.LoggerAdapter(logger, extra)
> logger_with_extra.handlers[0].setFormatter(_formatter("Worker ID"))
> ```
> 
> and this can be added in functions in the logger module

I actually looked into `LoggerAdapter` initially, but I thought it couldn't accept dynamic attributes. Turns out it can. I just had a POC PR https://github.com/elastic/connectors-python/pull/931, where I only apply the changes to `JobSchedulingService` and `JobExecutionService`, and it works as expected.

So in the POC, instead of calling
```python
connector.log_info("blah...")
sync_job.log_info("blah...")
```

You need to call like
```python
logger.info("blah...", connector=connector)
logger.info("blah...", sync_job=sync_job)
```

The main difference is that you move all the adapting logic to `logger` module in the POC, while it resides in each `ESDocument` class (i.e. `Connector` and `SyncJob`) in this PR.

The main drawback of the POC PR:
1. When we have more and more `ESDocument` classes, the adapting logic will become more complicated.
2. We can't apply this to source implementations. I don't think it's a good idea to pass `sync_job` object to any source implementation. 

cc @tarekziade

tarekziade (approved):

wangch079:
Hi @tarekziade , this PR uses a totally different approach from #931 . Do you mean we merge this for 8.8 and switch to use `LoggerAdapter` in 8.9?

tarekziade:
> Hi @tarekziade , this PR uses a totally different approach from #931 . Do you mean we merge this for 8.8 and switch to use `LoggerAdapter` in 8.9?

If the logger adapter requires more work and this branch here provides the feature we need, we could do the adapter as an improvement later. up to you

DianaJourdan:
>cc @danajuratoni @DianaJourdan given that 8.8 is postponed, what do you see if we merge this to 8.8?
The new date is tomorrow and a BC8 is already ready. I would not like to take any risks and push this so late into 8.8.

Let's merge this for 8.9 and backport to 8.8.1

DianaJourdan:
@wangch079 I have just noticed this PR was not merged. Should it just go in 8.9?

ppf2:
If it's an easy change, it would be nice to backport to a 8.8.x release as well to aid with troubleshooting and correlation.  Certainly happy to see this in 8.9.x though 😄

connectors/es/document.py
```diff
@@ -46,3 +123,61 @@ async def reload(self):
         self._seq_no = doc_source.get("_seq_no")
         self._primary_term = doc_source.get("_primary_term")
         self._source = doc_source.get("_source", {})
+
+    def log_debug(self, msg, *args, **kwargs):
```
artem-shelkovnikov, line 127: Should the code be removed since you introduced a wrapper?

wangch079: This is syntactic sugar, so you can call `connector.log_debug` instead of `connector.logger.debug`

artem-shelkovnikov: It feels like it does not add much value - it just passes arguments to a different object/function, so IMO it should be inlined

seanstory (changes requested):

connectors/es/document.py
wangch079: I think the value is that when you use `connector.log_debug`, you know you are logging with the context of the connector. Otherwise, you have to remember to use `connector.logger.debug`. Actually `logger` should be made `private`, i.e. `_logger`. It's not private because I have to pass it to each data source.

connectors/sync_job_runner.py
```diff
@@ -110,22 +110,19 @@ async def execute(self):
         self._start_time = time.time()
 
         try:
-            self.data_provider = self.source_klass(self.sync_job.configuration)
+            self.data_provider = self.source_klass(
+                configuration=self.sync_job.configuration, logger_=self.sync_job.logger
+            )
```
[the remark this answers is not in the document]
wangch079: e2e test successful: https://buildkite.com/elastic/connectors-python/builds/6347

seanstory (approved):
LGTM

artem-shelkovnikov:
Example output running service in debug mode:

```
artemshelkovnikov@Artems-MacBook-Pro-2 connectors-py % make run
bin/elastic-ingest
[FMWK][11:00:04][INFO] Running connector service version 8.9.0.0
[FMWK][11:00:04][INFO] Loading config from /Users/artemshelkovnikov/git_tree/connectors-py/connectors/../config.yml
[FMWK][11:00:04][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:04][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:04][INFO] Preflight checks...
[FMWK][11:00:04][DEBUG] Wait for Elasticsearch (max: 120)
[FMWK][11:00:04][INFO] Waiting for NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={}) (so far: 0 secs)
[FMWK][11:00:04][DEBUG] Checking for index .elastic-connectors presence
[FMWK][11:00:04][DEBUG] Checking for index .elastic-connectors-sync-jobs presence
[FMWK][11:00:04][DEBUG] ConnectorIndex connecting to http://localhost:9200
[FMWK][11:00:04][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:04][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:04][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:04][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:04][DEBUG] Native support for
[FMWK][11:00:04][INFO] Service started, listening to events from http://localhost:9200
[FMWK][11:00:04][DEBUG] Polling every 30 seconds
[FMWK][11:00:04][DEBUG] ConnectorIndex connecting to http://localhost:9200
[FMWK][11:00:04][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:04][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:04][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:04][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:04][DEBUG] Native support for
[FMWK][11:00:04][INFO] Service started, listening to events from http://localhost:9200
[FMWK][11:00:04][DEBUG] Polling every 30 seconds
[FMWK][11:00:04][INFO] Successfully started Job cleanup task...
[FMWK][11:00:04][DEBUG] ConnectorIndex connecting to http://localhost:9200
[FMWK][11:00:04][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:04][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:04][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:04][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:04][INFO] Start cleaning up orphaned jobs...
[FMWK][11:00:04][DEBUG] Importing module connectors.sources.mongo
[FMWK][11:00:04][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Sending heartbeat
[FMWK][11:00:04][INFO] No orphaned jobs found. Skipping...
[FMWK][11:00:04][INFO] Start cleaning up idle jobs...
[FMWK][11:00:04][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Status is Status.CONNECTED
[FMWK][11:00:04][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] 'full' sync scheduling is disabled
[FMWK][11:00:04][INFO] No idle jobs found. Skipping...
[FMWK][11:00:04][DEBUG] [Sync Job id: sAwc3YgBGEMX9i2IbCVk, connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Setting '_features' for MongoDB
[FMWK][11:00:04][DEBUG] [Sync Job id: sAwc3YgBGEMX9i2IbCVk, connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Validating configuration
[FMWK][11:00:06][DEBUG] [Sync Job id: sAwc3YgBGEMX9i2IbCVk, connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Existing databases: ['dana-tests', 'sample_airbnb', 'sample_analytics', 'sample_geospatial', 'sample_guides', 'sample_mflix', 'sample_restaurants', 'sample_supplies', 'sample_training', 'sample_weatherdata', 'sean-spacetime', 'admin', 'local']
[FMWK][11:00:06][DEBUG] [Sync Job id: sAwc3YgBGEMX9i2IbCVk, connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Existing collections in sample_airbnb: ['listingsAndReviewsShort', 'listingsAndReviews']
[FMWK][11:00:06][DEBUG] [Sync Job id: sAwc3YgBGEMX9i2IbCVk, connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Pinging the backend
[FMWK][11:00:06][DEBUG] SyncOrchestrator connecting to http://localhost:9200
[FMWK][11:00:06][DEBUG] Host is NodeConfig(scheme='http', host='localhost', port=9200, path_prefix='', headers={}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs=None, client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={})
[FMWK][11:00:06][DEBUG] Connecting using Basic Auth (user: elastic)
[FMWK][11:00:06][DEBUG] Preparing the content index
[FMWK][11:00:06][DEBUG] Checking index search-mongo
[FMWK][11:00:06][DEBUG] search-mongo exists
[FMWK][11:00:06][DEBUG] Index search-mongo already has mappings. Skipping...
[FMWK][11:00:06][INFO] Starting doc lookups
[FMWK][11:00:06][DEBUG] Scanning existing index search-mongo
[FMWK][11:00:07][DEBUG] Found 77 docs in search-mongo (duration 0 seconds)
[FMWK][11:00:07][DEBUG] Size of ids in memory is 0.01MiB
[FMWK][11:00:07][DEBUG] [Sync Job id: sAwc3YgBGEMX9i2IbCVk, connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Using pipeline {'name': 'ent-search-generic-ingestion', 'extract_binary_content': True, 'reduce_whitespace': True, 'run_ml_inference': True}
[FMWK][11:00:30][DEBUG] [get_doc call] 0-_decorate_with_metrics_span took 23.667561292648315 seconds.
[FMWK][11:00:30][INFO] Extractor <create: 0 |update: 0 |delete: 0>
[FMWK][11:00:31][DEBUG] Task 1 - Sending a batch of 154 ops -- 5.0MiB
[FMWK][11:00:31][INFO] Extractor <create: 23 |update: 77 |delete: 0>
[FMWK][11:00:32][DEBUG] Sink stats - no. of docs indexed: 77, volume of docs indexed: 5254000 bytes, no. of docs deleted: 0
[FMWK][11:00:32][DEBUG] [_bulk API call] _batch_bulk took 1.5169999599456787 seconds.
[FMWK][11:00:34][DEBUG] Polling every 30 seconds
[FMWK][11:00:34][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Status is Status.CONNECTED
[FMWK][11:00:34][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] 'full' sync scheduling is disabled
[FMWK][11:01:04][DEBUG] Polling every 30 seconds
[FMWK][11:01:04][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Status is Status.CONNECTED
[FMWK][11:01:05][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] 'full' sync scheduling is disabled
[FMWK][11:01:35][DEBUG] Polling every 30 seconds
[FMWK][11:01:35][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] Status is Status.CONNECTED
[FMWK][11:01:35][DEBUG] [Connector id: rwwb3YgBGEMX9i2IxCVQ, index name: search-mongo] 'full' sync scheduling is disabled
```


Is it indended for lines like `Extractor <create: 23 |update: 77 |delete: 0>` to not have correlation?

wangch079:
>Is it indended for lines like Extractor <create: 23 |update: 77 |delete: 0> to not have correlation?

Nice catch. I think this may be missed when resolving conflicts. Fixed in https://github.com/elastic/connectors-python/pull/806/commits/085701e98d82b4044759bf6d4cd13058389a59e4

seanstory (approved):

PR #808, opened 2023-04-21
Add items that are too big to ingest into the queue anyway

artem-shelkovnikov:
Problem that's found during testing and blocking ABS/GCS/Postgres testing.

Problem:

We have settings that allow us to limit the size of data that we store locally before we send it to Elasticsearch. We collect data into the queue of X megabytes max and then send it to Elasticsearch.

With default config the max queue size is 5MB, but for Cloud deployments it's smaller - 2MB.

Current logic has a flaw - if a document that is too large to even fit the queue is added to the queue it's deadlocked. It's considered "full" but still expects that somehow it'll free up to add this too-large-document. In the end after a timeout the queue says that it's full beyond recovery and crashes the job.

This fix changes the logic to actually add item to the queue if it's bigger than the threshold. This way the item is ingested as fast as possible anyway.

It might not make a lot of sense to discard it - it's already in memory, so we can attempt to ingest it anyway without putting too much pressure on memory.


connectors/utils.py
```diff
@@ -234,6 +234,12 @@ async def _putter_timeout(self, putter):
     async def put(self, item):
         item_size = get_size(item)
 
+        if item_size > self.maxmemsize:
+            logger.debug(
```
parth-elastic, line 238: Can we use warning log as we should aware user about it that we are skipping this file?

artem-shelkovnikov: Makes sense, though with current level of indirection it's hard to tell which document is ignored so it's not gonna be a really actionable message

timgrein (approved):
LGTM, would it make sense to include the doc id in the log message? So you could trace back, which documents were skipped?

artem-shelkovnikov:
> Would it make sense to include the doc id in the log message? So you could trace back, which documents were skipped?

MemQueue is abstract and makes no assumptions about what is stored inside, some tests try to store arrays inside, for example. 

I'll think a bit how it's doable without breaking this abstraction.

timgrein:
> MemQueue is abstract and makes no assumptions about what is stored inside, some tests try to store arrays inside, for example.

Was just wondering as the log message kinda implies that it's not abstract by using `Document too big` (maybe we should use `item` then?). An easy way would be to add an optional `on_item_too_big` field to the constructor, where you provide an anonymous function accepting the item being too big as parameter and let the caller decide what to do, but you'll probably already have something mind 👍

artem-shelkovnikov:
> Was just wondering as the log message kinda implies that it's not abstract by using Document too big (maybe we should use item too big then?). An easy way would be to add an optional on_item_too_big field to the constructor, where you provide an anonymous function accepting the item being too big as parameter and let the caller decide what to do, but you'll probably already have something mind 👍

I thought of just raising an exception with some additional information and catching it in the caller instead, seems easy to do.

However after a discussion I've changed this PR to actually attempt to ingest the document instead of dropping it.

We'll also need to follow up on this later - expose more settings in cloud, add potentially a setting to drop an item too.

artem-shelkovnikov:
Discussion: https://docs.google.com/document/d/1gUDjvxOpHSlKHNjKJPnJ87ltdT1AHB31m7ISPK-1CCs/edit#

tarekziade (approved):

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.7|[](https://github.com/elastic/connectors-python/pull/842)|

### Questions ?

PR #815, opened 2023-04-24
[New Connector] ServiceNow: Ping feature

parth-elastic:
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/servicenow.py
+++ b/connectors/sources/servicenow.py
@@ -41,7 +41,7 @@ class InvalidResponse(Exception):
     pass
 
 
-class ServiceNowClient(BaseDataSource):
+class ServiceNowClient():
     """ServiceNow Client"""
 
     def __init__(self, configuration):
--- a/connectors/sources/servicenow.py
+++ b/connectors/sources/servicenow.py
@@ -1,0 +1,346 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+"""ServiceNow source module responsible to fetch documents from ServiceNow."""
+import json
+import os
+
+import aiohttp
+from aiohttp.client_exceptions import ServerDisconnectedError
+
+from connectors.logger import logger
+from connectors.source import BaseDataSource, ConfigurableFieldValueError
+from connectors.utils import CancellableSleeps, iso_utc
+
+RETRIES = 3
+RETRY_INTERVAL = 2
+FETCH_SIZE = 1000
+MAX_CONCURRENCY_SUPPORT = 10
+
+RUNNING_FTEST = (
+    "RUNNING_FTEST" in os.environ
+)  # Flag to check if a connector is run for ftest or not.
+
+ENDPOINTS = {
+    "TABLE": "/api/now/table/{table}",
+    "ATTACHMENT": "/api/now/attachment",
+    "DOWNLOAD": "/api/now/attachment/{sys_id}/file",
+}
+
+
+class InvalidResponse(Exception):
+    """Raise when retrieving invalid response..
+
+    Args:
+        Exception (InvalidResponse): Invalid response exception.
+    """
+
+    pass
+
+
+class ServiceNowClient(BaseDataSource):
+    """ServiceNow Client"""
+
+    def __init__(self, configuration):
+        """Setup the ServiceNow client.
+
+        Args:
+            configuration (DataSourceConfiguration): Instance of DataSourceConfiguration class.
+        """
+
+        self.session = None
+        self._sleeps = CancellableSleeps()
+        self.configuration = configuration
+        self.services = self.configuration["services"]
+        self.retry_count = self.configuration["retry_count"]
+
+    def _get_session(self):
+        """Generate aiohttp client session with configuration fields.
+
+        Returns:
+            aiohttp.ClientSession: An instance of Client Session
+        """
+
+        if self.session:
+            return self.session
+
+        logger.debug("Generating aiohttp client session")
+        connector = aiohttp.TCPConnector(limit=MAX_CONCURRENCY_SUPPORT)
+        basic_auth = aiohttp.BasicAuth(
+            login=self.configuration["username"],
+            password=self.configuration["password"],
+        )
+        request_headers = {
+            "Content-Type": "application/json",
+            "Accept": "application/json",
+        }
+        timeout = aiohttp.ClientTimeout(total=None)  # pyright: ignore
+
+        self.session = aiohttp.ClientSession(
+            connector=connector,
+            base_url=self.configuration["url"],
+            auth=basic_auth,
+            headers=request_headers,
+            timeout=timeout,
+            raise_for_status=True,
+        )
+        return self.session
+
+    async def _api_call(self, url, params, is_attachment=False):
+        """Handle every api call to ServiceNow instance with retries.
+
+        Args:
+            url (str): ServiceNow query to be executed.
+            params (dict): Params to format the query.
+            is_attachment (bool, False): Flag for handle attachment api call. Defaults to False.
+
+        Raises:
+            InvalidResponse: An instance of InvalidResponse class.
+            Exception: An instance of an exception class.
+
+        Yields:
+            list: Table & Attachment metadata.
+            str: Attachment content.
+        """
+        retry = 1
+        offset = 0
+
+        if not is_attachment:
+            query = "ORDERBYsys_created_on^"
+            if "sysparm_query" in params.keys():
+                query += params["sysparm_query"]
+            params.update({"sysparm_query": query, "sysparm_limit": FETCH_SIZE})
+
+        while True:
+            if not is_attachment:
+                params["sysparm_offset"] = offset
+
+            try:
+                async with self._get_session().get(  # pyright: ignore
+                    url=url, params=params
+                ) as response:
+                    if (
+                        not RUNNING_FTEST
+                        and response.headers.get("Connection") == "close"
+                    ):
+                        raise Exception("Couldn't connect to ServiceNow instance")
+
+                    if is_attachment:
+                        yield response
+                        break
+
+                    fetched_response = await response.read()
+                    if fetched_response == b"":
+                        raise InvalidResponse(
+                            "Request hasn't processed from ServiceNow server"
+                        )
+                    elif not response.headers["Content-Type"].startswith(
+                        "application/json"
+                    ):
+                        raise InvalidResponse(
+                            "Request retrieved with invalid content type to process further"
+                        )
+
+                    json_response = json.loads(fetched_response)
+                    result = json_response["result"]
+
+                    response_length = len(result)
+                    if response_length == 0:
+                        break
+
+                    offset += response_length
+                    retry = 1
+                    yield result
+
+            except Exception as exception:
+                logger.warning(
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}."
+                )
+                await self._sleeps.sleep(RETRY_INTERVAL**retry)
+
+                if isinstance(
+                    exception,
+                    ServerDisconnectedError,
+                ):
+                    await self.session.close()  # pyright: ignore
+
+                if retry == self.retry_count:
+                    raise exception
+                retry += 1
+
+    async def filter_services(self):
+        """Filter services based on service mappings.
+
+        Returns:
+            list, list: Valid service names, Invalid services.
+        """
+
+        try:
+            logger.debug("Filtering services")
+            service_names, invalid_services = [], self.services.copy()
+
+            payload = {"sysparm_fields": "label, name"}
+            async for response in self._api_call(
+                url=ENDPOINTS["TABLE"].format(table="sys_db_object"), params=payload
+            ):
+                for mapping in response:  # pyright: ignore
+                    if mapping["label"] in invalid_services:
+                        service_names.append(mapping["name"])
+                        invalid_services.remove(mapping["label"])
+
+            return service_names, invalid_services
+
+        except Exception as exception:
+            logger.exception(f"Error while filtering services. Exception: {exception}.")
+            raise
+
+    async def ping(self):
+        """Ping to ServiceNow instance."""
+
+        payload = {
+            "sysparm_query": "label=Incident",
+            "sysparm_fields": "label, name",
+        }
+        await anext(
+            self._api_call(
+                url=ENDPOINTS["TABLE"].format(table="sys_db_object"), params=payload
+            )
+        )
+
+    async def close_session(self):
+        """Closes unclosed client session"""
+        self._sleeps.cancel()
+        if self.session is None:
+            return
+        await self.session.close()
+        self.session = None
+
+
+class ServiceNowDataSource(BaseDataSource):
+    """ServiceNow"""
+
+    name = "ServiceNow"
+    service_type = "servicenow"
+
+    def __init__(self, configuration):
+        """Setup the connection to the ServiceNow instance.
+
+        Args:
+            configuration (DataSourceConfiguration): Instance of DataSourceConfiguration class.
+        """
+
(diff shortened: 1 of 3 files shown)


connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+            },
+            "password": {
+                "label": "Password",
+                "sensitive": True,
+                "order": 3,
+                "type": "str",
+                "value": "changeme",
+            },
+            "services": {
+                "label": "List of services",
```
navarone-feekery, line 275:
```suggestion
                "display": "textarea",
                "label": "Comma-separated list of services",
```

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                "order": 4,
+                "type": "list",
+                "value": "*",
+            },
+            "retry_count": {
+                "display_value": RETRIES,
```
navarone-feekery, line 281:
```suggestion
                "default_value": RETRIES,
```

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                "display": "numeric",
+                "label": "Maximum retries per request",
```
navarone-feekery, line 283:
```suggestion
                "label": "Retries per request",
```

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+            payload = {"sysparm_fields": "label, name"}
+            async for response in self._api_call(
+                url=ENDPOINTS["TABLE"].format(table="sys_db_object"), params=payload
+            ):
+                for mapping in response:  # pyright: ignore
+                    if mapping["label"] in invalid_services:
+                        service_names.append(mapping["name"])
+                        invalid_services.remove(mapping["label"])
+
+            return service_names, invalid_services
```
navarone-feekery, line 193: Why are we returning the data from `name` for valid services but `label` for invalid services?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                    fetched_response = await response.read()
+                    if fetched_response == b"":
+                        raise InvalidResponse(
+                            "Request hasn't processed from ServiceNow server"
+                        )
+                    elif not response.headers["Content-Type"].startswith(
+                        "application/json"
+                    ):
+                        raise InvalidResponse(
+                            "Request retrieved with invalid content type to process further"
```
navarone-feekery, line 143:
```suggestion
                            f"Cannot proceed due to unexpected response type '{response.headers["Content-Type"]}'; response type must begin with 'application/json'."
```

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                        raise Exception("Couldn't connect to ServiceNow instance")
+
+                    if is_attachment:
+                        yield response
+                        break
+
+                    fetched_response = await response.read()
+                    if fetched_response == b"":
+                        raise InvalidResponse(
+                            "Request hasn't processed from ServiceNow server"
```
navarone-feekery, line 137: Question: does this mean that ServiceNow is still processing the request, or that they have failed to?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                    if response_length == 0:
+                        break
+
+                    offset += response_length
+                    retry = 1
+                    yield result
+
+            except Exception as exception:
+                logger.warning(
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}."
```
navarone-feekery, line 159:
```suggestion
                    f"Retry attempt {retry} out of {self.retry_count} failed. Exception: {exception}."
```

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                )
+                await self._sleeps.sleep(RETRY_INTERVAL**retry)
+
+                if isinstance(
+                    exception,
+                    ServerDisconnectedError,
+                ):
+                    await self.session.close()  # pyright: ignore
```
navarone-feekery, line 167: Question: Does this need to wait for the `.sleep` command above to close?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+MAX_CONCURRENCY_SUPPORT = 10
+
+RUNNING_FTEST = (
+    "RUNNING_FTEST" in os.environ
+)  # Flag to check if a connector is run for ftest or not.
+
+ENDPOINTS = {
+    "TABLE": "/api/now/table/{table}",
+    "ATTACHMENT": "/api/now/attachment",
+    "DOWNLOAD": "/api/now/attachment/{sys_id}/file",
```
navarone-feekery, line 29: Question: I can't find these used anywhere. Will they be used in the future?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+
+        Returns:
+            aiohttp.ClientSession: An instance of Client Session
+        """
+
+        if self.session:
+            return self.session
+
+        logger.debug("Generating aiohttp client session")
+        connector = aiohttp.TCPConnector(limit=MAX_CONCURRENCY_SUPPORT)
```
navarone-feekery, line 70: Question: Should we use the configuration value `concurrent_downloads` here? Or is this something else?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+        basic_auth = aiohttp.BasicAuth(
+            login=self.configuration["username"],
+            password=self.configuration["password"],
+        )
+        request_headers = {
+            "Content-Type": "application/json",
+            "Accept": "application/json",
+        }
+        timeout = aiohttp.ClientTimeout(total=None)  # pyright: ignore
```
navarone-feekery, line 79: Question: will having `total=None` default to a specific timeout value?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+        self.retry_count = self.configuration["retry_count"]
+
+    def _get_session(self):
+        """Generate aiohttp client session with configuration fields.
+
+        Returns:
+            aiohttp.ClientSession: An instance of Client Session
+        """
+
+        if self.session:
```
timgrein, line 66: I think you can remove this explicit check and use `@cached_property`. Should result in the same functionality. You can probably remove the `self.session` field then also and only use `_get_session`.

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+    """Raise when retrieving invalid response..
+
+    Args:
+        Exception (InvalidResponse): Invalid response exception.
+    """
+
+    pass
+
+
+class ServiceNowClient(BaseDataSource):
```
timgrein, line 43: Why does the `ServiceNowClient` extend the `BaseDataSource`?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                        invalid_services.remove(mapping["label"])
+
+            return service_names, invalid_services
+
+        except Exception as exception:
+            logger.exception(f"Error while filtering services. Exception: {exception}.")
+            raise
+
+    async def ping(self):
+        """Ping to ServiceNow instance."""
```
timgrein, line 200:
```suggestion
```

IMO we should remove doc strings, when they don't add any additional information, which isn't already present in the method name.

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+
+        options["concurrent_downloads"] = self.concurrent_downloads
+
+    @classmethod
+    def get_default_configuration(cls):
+        """Get the default configuration for ServiceNow.
+
+        Returns:
+            dict: Default configuration.
+        """
```
timgrein, line 253:
```suggestion
```

Same rationale as above, the doc string doesn't add any additional information.

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+                "label": "Maximum concurrent downloads",
+                "order": 6,
+                "required": False,
+                "type": "int",
+                "ui_restrictions": ["advanced"],
+                "value": MAX_CONCURRENCY_SUPPORT,
+            },
+        }
+
+    async def _remote_validation(self):
```
timgrein, line 302: Should this be decorated with `@retryable`?

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+            if invalid_services:
+                raise ConfigurableFieldValueError(
+                    f"Services '{', '.join(invalid_services)}' are not available. Available services are: '{', '.join(service_names)}'"
+                )
+
+    async def validate_config(self):
+        """Validates whether user input is empty or not for configuration fields
+        Also validate, if user configured services are available in ServiceNow."""
+
+        logger.info("Validating ServiceNow Configuration")
```
timgrein, line 323:
```suggestion
```

The framework already logs a similar information. We can remove that IMO.

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+        self.configuration.check_valid()
+        await self._remote_validation()
+
+    async def close(self):
+        """Closes servicenow client."""
```
timgrein, line 328:
```suggestion
```

Same as rationale as in the above comments.

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,346 @@
...
+
+        await self.servicenow_client.close_session()
+
+    async def ping(self):
+        """Verify the connection with ServiceNow."""
+
+        try:
+            logger.info("Pinging ServiceNow instance")
```
timgrein, line 336:
```suggestion
```

This is also redundant IMO as `sync_job_runner` logs something similar before creating a sync.

connectors/sources/tests/test_servicenow.py
```diff
@@ -0,0 +1,343 @@
...
+import pytest
+from aiohttp.client_exceptions import ServerDisconnectedError
+
+from connectors.source import ConfigurableFieldValueError, DataSourceConfiguration
+from connectors.sources.servicenow import ServiceNowClient, ServiceNowDataSource
+from connectors.sources.tests.support import create_source
+from connectors.tests.commons import AsyncIterator
+
+
+class MockResponse:
```
timgrein, line 18: Could use an `AsyncMock()` instead of a custom class?

connectors/sources/tests/test_servicenow.py
```diff
@@ -0,0 +1,343 @@
...
+    assert config["password"] == "changeme"
+
+
+@pytest.mark.parametrize("field", ["url", "username", "password", "services"])
+@pytest.mark.asyncio
+async def test_validate_config_missing_fields_then_raise(field):
+    source = create_source(ServiceNowDataSource)
+    source.configuration.set_field(name=field, value="")
+
+    with pytest.raises(Exception):
```
timgrein, line 67:
```suggestion
    with pytest.raises(ConfigurableFieldValueError):
```

IMO we should check for the specific exception when possible

timgrein (commented):
Left some comments, I think we can remove some of the doc strings, if they don't provide any additional information, when compared to the name of the method.

connectors/sources/servicenow.py
parth-elastic: Yes, request was failed and returned empty response

connectors/sources/servicenow.py
parth-elastic: Yes, It will be used for downloading attachment in `get_docs`

connectors/sources/servicenow.py
parth-elastic: For not limiting client creation in the session

connectors/sources/servicenow.py
parth-elastic: It will be helpful in `get_docs` for preparing URLs with `name` & kept `label` for invalid services so that user can easily find out which entered services are invalid

connectors/sources/servicenow.py
parth-elastic: Retry mechanism is already handle internally in `_api_call`

connectors/sources/tests/test_servicenow.py
parth-elastic: Custom class still be needed after extending `AsyncMock`. So created custom class with `aenter` & `aexit`

connectors/sources/servicenow.py
timgrein: I think it's better to just yield the docs from the client "raw" and serialize them in the `ServiceNowDataSource`. When we think of inheritance as a "is a" relationship it's a bit confusing IMO, when the client is also a data source.

moxarth-elastic:
buildkite test this

connectors/sources/servicenow.py
navarone-feekery: Right, but in the configuration there's a field called `concurrent_downloads`. Shouldn't it be used here instead of the const `MAX_CONCURRENCY_SUPPORT`?

connectors/sources/servicenow.py
navarone-feekery: In that case, I suggest:

```suggestion
                            "Request to ServiceNow server returned an empty response."
```

connectors/sources/servicenow.py
```diff
@@ -0,0 +1,332 @@
...
+
+        if self.servicenow_client.services != ["*"]:
+            (
+                service_names,
+                invalid_services,
+            ) = await self.servicenow_client.filter_services()
+            if invalid_services:
+                raise ConfigurableFieldValueError(
+                    f"Services '{', '.join(invalid_services)}' are not available. Available services are: '{', '.join(service_names)}'"
+                )
```
navarone-feekery, line 307: > It will be helpful in get_docs for preparing URLs with name & kept label for invalid services so that user can easily find out which entered services are invalid

In a comment above we were discussing names vs labels, this error here is an example of what is confusing me. It tells the user two things:
- Service _labels_ x, y, and z are unavailable
- Service _names_ a, b, and c are available 

It's comparing different things (labels vs names), so I think it can lead to confusion. Can you rework this error message to be consistent?
For example, either:

`f"Services '{', '.join(invalid_service_labels)}' are not available. Available services are: '{', '.join(service_labels)}'"`

Or

`f"Services '{', '.join(invalid_service_names)}' are not available. Available services are: '{', '.join(service_names)}'"`

connectors/sources/servicenow.py
parth-elastic: If user enters `concurrent_downloads` value as 5 then it will restrict client creation in session only with 5 which may affect performance. Our source handles 10 clients at a time

navarone-feekery: Okay, so it's a different use-case for the `concurrent_downloads` configuration field. What is `concurrent_downloads` used for then, if not this? Is it being added later?

connectors/sources/servicenow.py
parth-elastic: Got it 👍 Updated

connectors/sources/servicenow.py
timgrein: Could you elaborate a bit on the circular inheritance/concurrency problem with an example? Maybe we need to simply redesign the class structure to remove those issues. IMO we definitely shouldn't inherit from `BaseDataSource` in a client class.

connectors/sources/servicenow.py
parth-elastic: Yup, It will be used for handling concurrency for downloading attachment content

navarone-feekery (commented):
Thanks for addressing my comments. LGTM, but I'll wait until @timgrein has finished his review to approve.

connectors/sources/servicenow.py
parth-elastic: Faced this issue while implementing `get_docs` but for now we can remove this inheritance and will redesign class structure according to your suggestion

timgrein: Great, thanks!

moxarth-elastic:
buildkite test this

timgrein (approved):
LGTM, thank you for implementing the proposed changes

PR #816, opened 2023-04-24
Use SIGINT to terminate the process from monitor

artem-shelkovnikov:
Minor change to ftest flow - instead of terminating service with SIGTERM, we give it a chance to gracefully stop with SIGINT.

Why?

Perf8 needs time to compile reports and SIGTERM sometimes prevents it from doing so - if amount of data is reasonably large Perf8 will take several minutes to compile the report. This change potentially makes it so that we'll wait until Perf8 report will be completed.

### DIFF
--- a/connectors/sources/tests/fixtures/fixture.py
+++ b/connectors/sources/tests/fixtures/fixture.py
@@ -88,7 +88,7 @@ def _monitor_service(pid):
         print(f"Failed to monitor the sync job. Something bad happened: {e}")
     finally:
         # the process should always be killed, no matter the monitor succeeds, times out or raises errors.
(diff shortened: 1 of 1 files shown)


timgrein (approved):
LGTM

tarekziade:
@artem-shelkovnikov I wonder why `SIGTERM sometimes prevents it from doing so`

they are treated the same way https://github.com/elastic/perf8/blob/main/perf8/watcher.py#L55

artem-shelkovnikov:
@tarekziade you are right, it was one of small things I wanted to check, but I've created a different PR to address the whole problem: https://github.com/elastic/connectors-python/pull/819

PR #818, opened 2023-04-25
Fix misformatted mongodb connector.json

artem-shelkovnikov:
Small mistake in MongoDB connector.json - somehow there's a dangling comma!


navarone-feekery (approved):
I did this! My bad.

navarone-feekery:
I'm surprised this was allowed to merge. Do we need a check on JSON formatting as well?

artem-shelkovnikov:
We don't check formatting for YML/JSON - maybe it's a good idea indeed, but it would have been caught by FTEST anyway :)

PR #819, opened 2023-04-25
Monitor PERF8 process to make sure it finishes the report

artem-shelkovnikov:
Problem - PERF8 process takes variable amount of time when generating the report. Process of report generation happens in the background, thus if ftest ends faster than report is generated, then report is not actually generated and stored.

This PR changes `ftest.sh`. Instead of sleeping for 2 minutes between sync job runs, we check process of perf8 in the end of execution - if it's running, we're waiting for it to finish.

Later we can add a timeout for it too, but for now it does not seem to be needed.


connectors/tests/ftest.sh
```diff
@@ -74,3 +69,16 @@ NUM_DOCS=`$PYTHON fixture.py --name $NAME --action get_num_docs`
 $PYTHON $ROOT_DIR/scripts/verify.py --index-name search-$NAME --service-type $NAME --size $NUM_DOCS
 $PYTHON fixture.py --name $NAME --action stop_stack
 $PYTHON fixture.py --name $NAME --action teardown
+
```
tarekziade, line 72: This should be added before line 61

artem-shelkovnikov: Probably not - Perf8 generates report in the background, so we can save some time here by doing second ingestion and report generation at the same time. Performance of 2nd ingestion does not matter too much, so we can do it that way IMO

tarekziade (approved):

PR #820, opened 2023-04-25
Fix configuration formatting during preparation

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/4219

During the connector `prepare` function, also do the following:

- Raise error if a field is missing
  - No error is raised if an extra field is present, it is just ignored.
- Add missing field properties with default values

#### Pre-Review Checklist


navarone-feekery:
Converted to a draft because this doesn't work for native connectors, will try to fix that.

connectors/byoc.py
```diff
@@ -650,6 +681,55 @@ async def document_count(self):
...
+            )
+
+        # if a field is missing properties, add them with default values
+        configs_missing_properties = filter_nested_dict_by_keys(
+            DEFAULT_CONFIGURATION.keys(), current_config
+        )
+        if not configs_missing_properties:
+            return
+
+        logger.debug(
```
vidok, line 709: I would make it more verbose. I often run the service without `--debug` flag so I will miss this.

navarone-feekery: I've changed this logger to be `.info`. Do you think more logs could be added, or just this one more visible?

connectors/byoc.py
```diff
@@ -650,6 +681,55 @@ async def document_count(self):
         )
         return result["count"]
 
+    async def check_configuration_formatting(self, fqn, service_type):
```
vidok, line 684: I really like how you structured the code and it's super easy to read. However, the method's name is misleading because it not just checks but can update the configuration. 

**not a blocker**: I would also split the method into 2 because it does 2 things: validation, and modifying the doc.

navarone-feekery: I broke it up into a wrapper function and two sub functions. Also added docstrings so everything was a bit clearer.

connectors/byoc.py
vidok: it looks good to me

vidok (approved):
LGTM

PR #822, opened 2023-04-25
[e2e] add a max rss for each connector

tarekziade:
- Adds a max rss of 200MiB (and more for some)


tarekziade:
BK nightly build https://buildkite.com/elastic/connectors-python-nightly/builds/545

connectors/tests/ftest.sh
```diff
@@ -47,9 +47,9 @@ if [[ $PERF8 == "yes" ]]
 then
     if [[ $PLATFORM == "darwin" ]]
     then
-      $PERF8_BIN --refresh-rate $REFRESH_RATE -t $ROOT_DIR/perf8-report-$NAME --asyncstats --memray --psutil -c $ELASTIC_INGEST --debug & PID=$!
+      $PERF8_BIN --refresh-rate $REFRESH_RATE -t $ROOT_DIR/perf8-report-$NAME --asyncstats --memray --psutil --psutil-max-rss 209715200 -c $ELASTIC_INGEST --debug & PID=$!
```
timgrein, line 50: Could we add a comment, that 209715200 bytes are 200MiB?

artem-shelkovnikov:
200MB is not enough, right? I see some connectors take almost 270 MB, so let's do 300?

connectors/tests/ftest.sh
tarekziade: I think I'll do better : allow `--psutil-max-rss 200MiB`

tarekziade:
> 200MB is not enough, right? I see some connectors take almost 270 MB, so let's do 300?

Yeah I did this on purpose to see some fail. They don't haha. Because the bash script does not get back the exit code. We need to tweak that script

I so want to get rid of that bash file and do a simpler Python script..

tarekziade:
new run with raised limits per connector when needed https://buildkite.com/elastic/connectors-python-nightly/builds/554

navarone-feekery (approved):

PR #829, opened 2023-04-28
[SharePoint] Logic to handle too many request

akanshi-elastic:
#### PR Contains ####

- handle too many request error in `_api_call` method

#### Pre-Review Checklist
- [ x Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

### DIFF
--- a/connectors/sources/sharepoint.py
+++ b/connectors/sources/sharepoint.py
@@ -261,7 +261,7 @@ class SharepointClient:
         Yields:
             data: API response.
         """
-        retry = 0
+        retry = 1
         # If pagination happens for list and drive items then next pagination url comes in response which will be passed in url field.
         if url == "":
             url = URLS[url_name].format(**url_kwargs)
@@ -284,26 +284,33 @@ class SharepointClient:
                         yield await result.json()
                     break
             except Exception as exception:
-                if isinstance(
-                    exception,
-                    ClientResponseError,
-                ) and "token has expired" in exception.headers.get(  # pyright: ignore
-                    "x-ms-diagnostics", ""
-                ):
-                    await self._set_access_token()
-                elif isinstance(
-                    exception,
-                    ServerDisconnectedError,
-                ):
-                    await self.close_session()
-                if retry >= self.retry_count:
-                    raise exception
-                retry += 1
+                if retry > self.retry_count:
+                        raise exception
+                    logger.warning(
+                        f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                    )
+                    retry += 1
 
-                logger.warning(
-                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
-                )
-                await self._sleeps.sleep(RETRY_INTERVAL**retry)
+                    retry_after = RETRY_INTERVAL**retry
+
+                    if getattr(exception, "status", None) == 429:
+                        retry_after = int(
+                            exception.headers.get("Retry-After", 0)  # pyright: ignore
+                        )
+                    elif isinstance(
+                        exception,
+                        ClientResponseError,
+                    ) and "token has expired" in exception.headers.get(  # pyright: ignore
+                        "x-ms-diagnostics", ""
+                    ):
+                        await self._set_access_token()
+                    elif isinstance(
+                        exception,
+                        ServerDisconnectedError,
+                    ):
+                        await self.close_session()
+
+                    await self._sleeps.sleep(retry_after)
 
     async def _fetch_data_with_next_url(self, site_url, list_id, param_name):
         """Invokes a GET call to the SharePoint Server/Online for calling list and drive item API.
--- a/connectors/sources/tests/test_sharepoint.py
+++ b/connectors/sources/tests/test_sharepoint.py
@@ -928,7 +928,7 @@ async def test_api_call_when_token_is_expired(patch_default_wait_multiplier):
     source = create_source(SharepointDataSource)
     mock_response = {"access_token": "test2344", "expires_in": "1234555"}
     async_response = MockResponse(mock_response, 401)
-    source.sharepoint_client.retry_count = 0
+    source.sharepoint_client.retry_count = 1
     async_response.headers = {"x-ms-diagnostics": "token has expired"}
 
     with patch.object(
@@ -946,6 +946,38 @@ async def test_api_call_when_token_is_expired(patch_default_wait_multiplier):
             with pytest.raises(aiohttp.client_exceptions.ClientResponseError):
                 source.sharepoint_client._get_session()
 
+                # Execute
+                await anext(
+                    source.sharepoint_client._api_call(
+                        url_name="attachment", url="abc.com"
+                    )
+                )
+            await source.close()
+
+
+class TooManyRequestException(Exception):
+    status = 429
+    headers = {"Retry-After": 0}
+
+
+@pytest.mark.asyncio
+async def test_api_call_when_status_429_exception():
+    # Setup
+    source = create_source(SharepointDataSource)
+    mock_response = {"access_token": "test2344", "expires_in": "1234555"}
+    async_response = MockResponse(mock_response, 429)
+    source.sharepoint_client.retry_count = 1
+
+    async_response.headers = {}
+
+    with patch.object(
+        aiohttp.ClientSession,
+        "get",
+        side_effect=TooManyRequestException("Something Went Wrong"),
+    ):
+        with patch("aiohttp.request", return_value=async_response):
+            source.sharepoint_client._get_session()
+            with pytest.raises(TooManyRequestException):
                 # Execute
                 await anext(
                     source.sharepoint_client._api_call(


connectors/sources/sharepoint.py
```diff
@@ -284,7 +284,20 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                )
+                retry += 1
+
+                retry_after = RETRY_INTERVAL**retry
+
+                if getattr(exception, "status", None) == 429:
+                    retry_after = int(
+                        exception.headers.get("Retry-After", 0)  # pyright: ignore
+                    )
```
navarone-feekery, line 299: This should be default to the `RETRY_INTERVAL`.

connectors/sources/tests/test_sharepoint.py
```diff
@@ -955,6 +955,38 @@ async def test_api_call_when_token_is_expired(patch_default_wait_multiplier):
             await source.close()
 
 
+class TooManyRequestException(Exception):
+    status = 429
+    headers = {"Retry-After": 0}
+
+
+@pytest.mark.asyncio
+async def test_api_call_when_status_429_exception():
```
navarone-feekery, line 964: If we receive a 429 response, we retry. This test doesn't retry because the retry_count limit starts at `1`.
Can you add tests that checks that we retry the expected number of times?
- 1 test for 429 errors (and if possible, test that we retry using the expected header response as the interval value).
- 1 test for non-429 errors (to make sure retrying works in general).

navarone-feekery (changes requested):
Nice! Some change requests for now.

connectors/sources/sharepoint.py
```diff
@@ -284,7 +284,20 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
             except Exception as exception:
-                if isinstance(
+                if retry > self.retry_count:
+                    raise exception
+                logger.warning(
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                )
+                retry += 1
+
+                retry_after = RETRY_INTERVAL**retry
```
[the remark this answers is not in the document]
akanshi-elastic: IMO we should handle the validation of handling the max value for retry using UI so user can't enter random values like 1000.
And Regarding the exponential back off, SharePoint uses exponential that's why we have also used exponential backoff. We can move to liner back off too but exponential backoff make more sense to handle throttling in async connector

connectors/sources/tests/test_sharepoint.py
akanshi-elastic: I have updated test case for `test_api_call_when_status_429_exception`. And for non 429 errors I have already added test cases for that such as: `test_api_call_when_server_is_down`, `test_api_call_when_token_is_expired`.

connectors/sources/sharepoint.py
```diff
@@ -284,7 +284,25 @@ async def _api_call(self, url_name, url="", **url_kwargs):
                         yield await result.json()
                     break
             except Exception as exception:
```
artem-shelkovnikov, line 286: Should you change this `except Exception` to handle the specific exception that means that request was throttled?

connectors/sources/sharepoint.py
```diff
@@ -284,7 +284,25 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+
+                if getattr(exception, "status", None) == 429:
+                    retry_after = int(
+                        exception.headers.get(  # pyright: ignore
+                            "Retry-After", RETRY_INTERVAL
+                        )
+                    )
+                    logger.warning(
+                        f"Rate limit exceeds. Retrying after {retry_after} seconds."
+                    )
```
artem-shelkovnikov, line 304: Why would we retry if it's not 429?

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
akanshi-elastic: What should be the maximum value for retry to add in vallidation?

connectors/sources/sharepoint.py
akanshi-elastic: we are handling multiple exceptions inside this exception and if handle specific exception here then logic for retry will be duplicated in all exceptions.

connectors/sources/sharepoint.py
akanshi-elastic: There are some errors such as ServerDisconnectedError which can be fixed in after few retries

connectors/sources/sharepoint.py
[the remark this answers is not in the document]
artem-shelkovnikov: Let's do 10 times?

connectors/sources/sharepoint.py
artem-shelkovnikov: You can provide private methods for exception handling shared logic, the way it's done now is really hard to read and understand

connectors/sources/sharepoint.py
artem-shelkovnikov: It should be explicit then - catch `ServerDisconnectedError` and handle it instead of handling `Exception`.

connectors/sources/sharepoint.py
```diff
@@ -283,27 +303,40 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
-            except Exception as exception:
-                if isinstance(
-                    exception,
-                    ClientResponseError,
-                ) and "token has expired" in exception.headers.get(  # pyright: ignore
+            except ClientResponseError as exception:
+                retry, retry_seconds = self._get_retry_after(
+                    retry=retry, exception=exception
+                )
+                if getattr(exception, "status", None) == 429:
```
artem-shelkovnikov, line 310: No need to do this, ClientResponseError always have status

connectors/sources/sharepoint.py
```diff
@@ -283,27 +303,40 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+                    logger.warning("Rate limit exceeded.")
```
artem-shelkovnikov, line 311: Please add additional information about rate limit into warning - what data came from the Sharepoint and how it affected the run?

connectors/sources/sharepoint.py
```diff
@@ -283,27 +303,40 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+                    if "Retry-After" not in exception.headers:  # pyright: ignore
```
artem-shelkovnikov, line 312: Why # pyright: ignore?

I could see extracting to `response_headers = exeception.headers or {}` make it work

connectors/sources/sharepoint.py
```diff
@@ -283,27 +303,40 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+                        logger.warning(
+                            "Exception status is 429 but it doesn't contain Retry-After header"
```
artem-shelkovnikov, line 314:
```suggestion
                            "Response Code from Sharepoint Server is 429 but Retry-After header is not found"
```

connectors/sources/sharepoint.py
```diff
@@ -283,27 +303,40 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+                        )
+                    retry_seconds = int(
+                        exception.headers.get("Retry-After", 30)  # pyright: ignore
+                    )
```
artem-shelkovnikov, line 318: I suggest changing defaulting logic so that you can specify that 429 received, Retry-After is not found therefore we will retry in 30 seconds instead. Also 30 seconds should become a constant.

connectors/sources/sharepoint.py
```diff
@@ -283,27 +303,40 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
                     await self._set_access_token()
-                elif isinstance(
-                    exception,
-                    ServerDisconnectedError,
-                ):
-                    await self.close_session()
-                if retry >= self.retry_count:
-                    raise exception
-                retry += 1
+                logger.debug(f"Retrying after {retry_seconds} seconds")
```
artem-shelkovnikov, line 323:
```suggestion
                logger.debug(f"Will retry in {retry_seconds} seconds")
```

danajuratoni:
Added the 8.8 label seems it was missed. For 8.9 there would be no backport needed.
Please link this to an issue, for visibility.

artem-shelkovnikov (dismissed):

connectors/sources/sharepoint.py
```diff
@@ -283,27 +304,43 @@ async def _api_call(self, url_name, url="", **url_kwargs):
...
+                if exception.status == 429:
+                    if "Retry-After" in response_headers:
+                        retry_seconds = int(response_headers["Retry-After"])
+                    else:
+                        logger.warning(
+                            "Response Code from Sharepoint Server is 429 but Retry-After header is not found"
+                        )
+                        retry_seconds = DEFAULT_RETRY_SECONDS
+                    logger.warning(
+                        f"Rate Limited: have {response_headers['RateLimit-Remaining']} of {response_headers['RateLimit-Limit']} left, opportunity to retry in {retry_seconds} seconds"
```
artem-shelkovnikov, line 321:
```suggestion
                        f"Rate Limited by Sharepoint: have {response_headers['RateLimit-Remaining']} of {response_headers['RateLimit-Limit']} left, opportunity to retry in {retry_seconds} seconds"
```

navarone-feekery (approved):

PR #834, opened 2023-05-01
Fix regex for main labelling for backportrc

artem-shelkovnikov:
Maybe it'll fix our autobackports?

It constantly fails saying that it's unable to backport to the branch that's currently main.

Most likely it's broken cause our `main` is automatically labelled with additional `.0` in the end, for example see label for this PR.

We should account for both `v8.9.0` and `v.8.9.0.0` as main, thus I've updated regex to allow optional `.0` in the end.


seanstory (approved):
LGTM 🤞

PR #836, opened 2023-05-02
Fix perf8 example

ppf2:
PERF8 is enabled using =yes instead of =TRUE.


artem-shelkovnikov (approved):

artem-shelkovnikov:
buildkite test this

artem-shelkovnikov:
buildkite test this

PR #838, opened 2023-05-02
Use vault in nightly tests

artem-shelkovnikov:
I've set `useVault: false` for Nightly pipeline but it's actually needed to build Docker images, see https://buildkite.com/elastic/connectors-python-nightly/builds/587#0187dbb5-92de-4791-9115-4d0db6c1b1e1

This PR returns `useVault: true` for Nightlies


timgrein (approved):
LGTM

PR #845, opened 2023-05-03
[MySQL] Filter table update times, which are None

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4458

This PR fixes the issue, that table update times can be `None` sometimes. This leads to an error, when trying to get the max value of `last_update_times` as `>` is not defined on `NoneType`.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
How is it possible to have it None?

timgrein:
> How is it possible to have it None?

- Different storage engines handle it differently (InnoDB)
- Table just created, but no data in it
- Temporary tables
- and probably some other reasons I'm not aware of

PR #847, opened 2023-05-03
Update catalog-info.yaml

artem-shelkovnikov:
Just minor changes to make it nicer.

I copied its format from CI repo, but it's actually not right. Actual schema can be validated with https://backstage.elastic.dev/entity-validation, and that's what I did - changed the schema and validated it with the mentioned tool.


timgrein (approved):
LGTM

PR #859, opened 2023-05-03
Update catalog-info.yaml

artem-shelkovnikov:
My change broke CI: https://buildkite.com/elastic/terrazzo/builds/10542#0187e1d1-4a5d-43a9-a761-df533b017f06

>  There is already a Construct with name ‘connectors-python’ in Construct

Seems like I need to provide unique name for each entry in `catalog-info.yaml`.


timgrein (approved):
LGTM

PR #871, opened 2023-05-04
Increase FTEST RSS thresholds for 8.7

artem-shelkovnikov:
Increasing RSS thresholds by ~35MB for each connector affected.

Seems like we've improved our memory footprints between 8.7 and 8.8 and RSS limits that work for 8.8 don't work for 8.7.

This. PR just increase thresholds so that our nightlies work well.


timgrein (approved):
LGTM

PR #876, opened 2023-05-04
[ServiceNow] Implementation Sync feature

### LINKED ISSUE #877: ServiceNow: Implementation Sync feature

parth-elastic:
#### Pre-Review Checklist


connectors/sources/servicenow.py
```diff
@@ -4,35 +4,58 @@
...
+    ConcurrentTasks,
+    MemQueue,
+    convert_to_b64,
+    iso_utc,
+)
 
 RETRIES = 3
 RETRY_INTERVAL = 2
+QUEUE_MEM_SIZE = 25 * 1024 * 1024  # Size in Megabytes
+FILE_SIZE_LIMIT = 10485760
```
timgrein, line 33: Can you add a similar comment here as in the line above?

connectors/sources/servicenow.py
```diff
@@ -4,35 +4,58 @@
...
 FETCH_SIZE = 1000
 MAX_CONCURRENCY_SUPPORT = 10
 
+END_SIGNAL = "TASK_FINISHED"
 RUNNING_FTEST = (
     "RUNNING_FTEST" in os.environ
 )  # Flag to check if a connector is run for ftest or not.
 
+DEFAULT_QUERY = "ORDERBYsys_created_on^"
```
timgrein, line 42:
```suggestion
ORDER_BY_CREATION_DATE_QUERY = "ORDERBYsys_created_on^"
```

suggestion: I would be explicit here what the query does exactly

connectors/sources/servicenow.py
```diff
@@ -4,35 +4,58 @@
...
+    "sc_req_item",
+    "incident",
+    "kb_knowledge",
+    "change_request",
+)
 
 
 class InvalidResponse(Exception):
-    """Raise when retrieving invalid response..
+    """Raise when retrieving invalid response.
```
timgrein, line 58:
```suggestion
```

nit: I don't think we need an explicit doc string here. `InvalidResponse` already tells enough IMO.

connectors/sources/servicenow.py
```diff
@@ -194,6 +217,83 @@ async def filter_services(self):
...
+
+        try:
+            async for response in self._api_call(url=url, params=params):
+                for record in response:  # pyright: ignore
+                    yield record
+
+        except Exception as exception:
+            logger.warning(
+                f"Skipping data for {url} with {params}. Exception: {exception}."
+            )
```
timgrein, line 239: question: Should we either re-raise the exception so the data source can also handle it (and maybe set an error?) or simply let the data source handle it by moving the log message?

connectors/sources/servicenow.py
```diff
@@ -194,6 +217,83 @@ async def filter_services(self):
...
+        """
+
+        attachment_size = int(metadata["size_bytes"])
+        if not (doit and attachment_size > 0):
+            return
+
+        attachment_name = metadata["file_name"]
+        if os.path.splitext(attachment_name)[-1] not in TIKA_SUPPORTED_FILETYPES:
+            logger.warning(f"{attachment_name} is not supported by TIKA, skipping.")
+            return
```
timgrein, line 260:
```suggestion
        attachment_extension = os.path.splitext(attachment_name)[-1]
        if  attachment_extension not in TIKA_SUPPORTED_FILETYPES:
            logger.warning(f"Files with the extension {attachment_extension} are not supported by TIKA, skipping {attachment_name}.")
            return
```

suggestion: I would use a variable here and also indicate in the log message, that the file type is not supported.

connectors/sources/servicenow.py
```diff
@@ -194,6 +217,83 @@ async def filter_services(self):
...
+                    f"Skipping content for {attachment_name}. Exception: {exception}."
+                )
+                return
+
+        logger.debug(f"Calling convert_to_b64 for file : {attachment_name}.")
+        await asyncio.to_thread(convert_to_b64, source=temp_filename)
+
+        async with aiofiles.open(file=temp_filename, mode="r") as async_buffer:
+            document["_attachment"] = (await async_buffer.read()).strip()
+        await remove(temp_filename)
```
timgrein, line 294: question: Can we encounter an exception here, where the file wasn't deleted properly? If yes we should either retry and/or log a warning, that the file wasn't removed IMO.

connectors/sources/servicenow.py
```diff
@@ -229,6 +329,13 @@ def __init__(self, configuration):
         self.concurrent_downloads = self.configuration["concurrent_downloads"]
         self.servicenow_client = ServiceNowClient(configuration=configuration)
 
+        self.valid_services = None
+        self.invalid_services = None
```
timgrein, line 333:
```suggestion
        self.valid_services = []
        self.invalid_services = []
```

suggestion: I would use empty lists as default so we're not running into any potential issues, when we try to log/format `None` somewhere else.

connectors/sources/servicenow.py
```diff
@@ -327,6 +434,111 @@ async def ping(self):
...
+        """
+
+        data = {key: value for key, value in data.items() if value}
+        data.update(
+            {
+                "_id": data["sys_id"],
+                "_timestamp": iso_utc(parser.parse(data["sys_updated_on"])),
+            }
+        )
+        return self.serialize(doc=data)
```
timgrein, line 454: nit: I think we should keep the `serialize` call separate from `format` as these are different concerns IMO.

connectors/sources/servicenow.py
```diff
@@ -327,6 +434,111 @@ async def ping(self):
...
+
+    async def _attachment_fetcher(self, table_sys_id):
+        """Method to add attachment metadata to queue
+
+        Args:
+            table_sys_id (str): Table Id for fetching attachment metadata.
+        """
+
+        async for attachment_metadata in self.servicenow_client.fetch_data(
```
timgrein, line 463: suggestion: I think we should have more high-level methods in the `ServiceNowClient` like `fetch_attachments` or `get_attachments` (see: [MySQLClient](https://github.com/elastic/connectors-python/blob/main/connectors/sources/mysql.py#L204)). Makes it easier to read and easier to use.

connectors/sources/servicenow.py
```diff
@@ -327,6 +434,111 @@ async def ping(self):
...
+            url=ENDPOINTS["ATTACHMENT"], params={"table_sys_id": table_sys_id}
+        ):
+            formatted_attachment_metadata = await asyncio.to_thread(
```
timgrein, line 466: question: Do you think it's worth to spawn a separate thread for updating a dict vs. simply running it in a "blocking" way? What's the rationale behind using a separate thread here?

connectors/sources/servicenow.py
```diff
@@ -327,6 +434,111 @@ async def ping(self):
...
+
+    async def _producer(self, service_name):
+        """Fetch data for configured service name.
+
+        Args:
+            service_name (str): Service Name for preparing URL.
+        """
+
+        logger.debug(f"Fetching {service_name} data")
+        async for table_data in self.servicenow_client.fetch_data(
```
timgrein, line 489: suggestion: Again I think it's worth to have a more high-level method like `get_table_data` inside the `ServiceNowClient`.

connectors/sources/servicenow.py
```diff
@@ -327,6 +434,111 @@ async def ping(self):
...
+            url=ENDPOINTS["TABLE"].format(table=service_name), params={}
+        ):
+            formatted_table_data = await asyncio.to_thread(
```
timgrein, line 492: question: Same as above for formatting attachment metadata

connectors/sources/servicenow.py
```diff
@@ -327,6 +434,111 @@ async def ping(self):
...
+            )
+            await self.queue.put(
+                (  # pyright: ignore
+                    formatted_attachment_metadata,
+                    partial(
+                        self.servicenow_client.fetch_attachment_content,
+                        formatted_attachment_metadata,
+                    ),
+                )
+            )
```
timgrein, line 477: question: Do we also need to increment `task_count` here?

connectors/sources/servicenow.py
```diff
@@ -327,6 +434,111 @@ async def ping(self):
...
+
+        Yields:
+            dict: Formatted document.
+        """
+
+        while self.task_count > 0:
+            _, item = await self.queue.get()
+
+            if item == END_SIGNAL:
+                self.task_count -= 1
```
timgrein, line 517: question: For `_producer` we're incrementing the `task_count` for each iteration of the loop iterating over the table data, but we're only putting one `END_SIGNAL` into the queue. Will `task_count` be calculated correctly?

connectors/sources/tests/test_servicenow.py
```diff
@@ -333,3 +333,248 @@ async def test_filter_services_with_exception():
...
+                            },
+                        ]
+                    ]
+                ),
+            ],
+        ):
+            async for response in source.get_docs():
+                response_list.append(response)
+
+    assert (
```
timgrein, line 428: question: Should we assert that both docs are present in `response_list`?

timgrein (commented):
Left some suggestions and also some questions around the usage of separate threads for formatting docs.

connectors/sources/servicenow.py
parth-elastic: Yeah! We can remove overhead for spawning separate thread

connectors/sources/servicenow.py
parth-elastic: No, After putting all attachments in the queue for one specific record we are adding `END_SIGNAL`

connectors/sources/servicenow.py
parth-elastic: Based on fetchers `task_count` are incremented & whenever that fetchers execution will be completed `END_SIGNAL` is added to the queue

connectors/sources/servicenow.py
parth-elastic: Moving exception handling to data source and as user must be warned for that skipped data kept `logger.warning` rather then `logger.error`. Should I replace to error?

connectors/sources/servicenow.py
timgrein:
```
try:
      async for table_data in self.servicenow_client.get_table_data(
          url=ENDPOINTS["TABLE"].format(table=service_name)
      ):
          formatted_table_data = self._format_doc(data=table_data)
          serialized_table_data = self.serialize(doc=formatted_table_data)
          await self.fetchers.put(
              partial(
                  self._attachment_fetcher,
                  serialized_table_data["_id"],
              )
          )
          self.task_count += 1

          await self.queue.put((serialized_table_data, None))  # pyright: ignore
      logger.warning(
          f"Skipping table data for {service_name}. Exception: {exception}."
      )

  await self.queue.put(END_SIGNAL)  # pyright: ignore
```

I may misunderstand something, but if we look at this example for `_producer` I notice, that we increment the `task_count` for every iteration in the async loop, but we're only putting one `END_SIGNAL` into the queue.

So if `servicenow_client.get_table_data(...)` yields for example 5x `table_data` we've a `task_count` of 5 and one `END_SIGNAL`. The `_consumer` runs as long as `task_count` is greater than 0. As we're only having one `END_SIGNAL` in the queue the loop will run forever, right (we only decrement `task_count` once)? Am I missing something?

timgrein (commented):
Added an example for my question regarding the `task_count` increment/decrement.

connectors/sources/servicenow.py
parth-elastic: `task_count` aren't updated based on async loop but based on `fetcher.put` we are incrementing.

timgrein: But `await self.queue.put(END_SIGNAL)` lives outside the loop in the example I've provided. So if `get_table_data` yields N times, we've N increments to `task_count`, but only one `END_SIGNAL`, right? Do we've an explicit test for this behavior?

parth-elastic: `END_SIGNAL` for every `task_count` are added to queue from the method which is going to be executed in `fetcher`

timgrein: Thanks for the explanation, we've agreed in a call to introduce an enumeration so we can differentiate between different `END_SIGNAL`(s). Then it's more obvious, which `END_SIGNAL` belongs to which task.

timgrein (approved):
Thank you for the changes! Added one small improvement (classes should be PascalCase), but otherwise LGTM

timgrein (approved):
Thank you for the changes! Added one small improvement (classes should be PascalCase), but otherwise LGTM

connectors/sources/servicenow.py
```diff
@@ -4,40 +4,62 @@
...
+    "sc_req_item",
+    "incident",
+    "kb_knowledge",
+    "change_request",
+)
 
 
-class InvalidResponse(Exception):
-    """Raise when retrieving invalid response..
+class END_SIGNAL(Enum):
```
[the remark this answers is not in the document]
parth-elastic: It will affect at the other changes as well so pushed a new commit.

PR #882, opened 2023-05-05
Correctly copy over log_level from Enterprise Search config file

artem-shelkovnikov:
Funny problem in Cloud - when user changes `log_level` setting, it has to be lowercase.

However, Connector Service expects upper case log_level and is unable to start. Thus, adding this little hack for now to unblock 8.8 deployment - if we're updating log_level for deployment from Enterprise Search, then upper case it.

We need to follow up on this with a more complex value remapping from Enterprise Search config.

### DIFF
--- a/connectors/config.py
+++ b/connectors/config.py
@@ -38,10 +38,13 @@ def _ent_search_config(configuration):
             continue
 
         connector_field = config_mappings[es_field]
+        es_field_value = ent_search_config[es_field]
 
-        _update_config_field(
-            configuration, connector_field, ent_search_config[es_field]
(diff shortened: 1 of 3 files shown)


tarekziade (dismissed):
It's not really a hack -- it's valid.
 
you could also check that the string is correct:

```
if level == 'WARN':
    level = 'WARNING'
levels = ('CRITICAL', 'FATAL', 'ERROR', 'WARN', 'INFO', 'DEBUG')
if level not in levels:
    raise Something()
```

tarekziade (approved):

PR #885, opened 2023-05-05
Rename `byoc` to `connectors.protocol.*`

tarekziade:
`byoc` represents the protocol database for connectors, `db` is probably a better name


wangch079 (commented):
I agree `byoc` is not easy to understand for developers without context, but I'd like to propose a different name other than `db`, maybe `system`, `core`, or just `connectors`?

timgrein (commented):
I could also see some different names possible dependant on how the structure evolves. With DLS for example we'll need a `PermissionsIndex` class. If we decide to also have that class inside `byoc.py` the proposed name `db.py` (if we want to hide the fact we're using es indices) or `index.py` / `indices.py` (if we don't want to hide that we're using es indices) would make sense IMO. If we call it `connector.py` we should have a separate module `permissions.py`, where the `PermissionsIndex` class would live. 

So I've basically two options in mind:
- Modules by domain: `ConnectorIndex` inside `connector.py` and `PermissionsIndex` inside `permissions.py`
- Modules by functionality: `db.py` / `index.py` containing every index abstraction

Personally I would lean towards `connector.py` and `permissions.py` as `db.py` could be pretty overloaded in the future, but both options are already way better than `byoc.py`.

tarekziade:
> I agree `byoc` is not easy to understand for developers without context, but I'd like to propose a different name other than `db`, maybe `system`, `core`, or just `connectors`?

I am trying to find a name that corresponds to the storage the framework uses to function -- the database to function. So right now it's a couple of ES indices. If we don't call it `db` I'd like to find a name that is related to that function.

> Modules by functionality

That gets my vote :)

seanstory:
What about `protocol.py` or `connectors_protocol.py`? My understanding is that this file is responsible mostly for the business logic of maintaining the state machine that is "the Connectors Protocol."

`db.py` does not make sense to me. For starters, we are not using a database here (marketing prefers "document store" for ES), but we do have databases referenced in our mysql, postgres, and mssql files, and I think that could lead to confusion. Secondly, that will just lead to a need for another refactor when we eventually manage the protocol through REST APIs instead of direct data access.

timgrein:
> a protocol/connectors.rb, protocol/jobs.rb, protocol/permissions.rb might also make sense.

I really like that one, it's the most explicit and least confusing IMHO.

timgrein (approved):
LGTM, really liking `protocol`

PR #890, opened 2023-05-08
[Sharepoint] Make all documents IDs size 36 bytes

vidok:
To be able to measure data volume more precisely we need to store documents' IDs as 36 bytes strings (Sharepoint format)

## Related links
* https://github.com/elastic/enterprise-search-team/issues/4332

### DIFF
--- a/connectors/sources/tests/fixtures/sharepoint/fixture.py
+++ b/connectors/sources/tests/fixtures/sharepoint/fixture.py
@@ -19,11 +19,12 @@ _SIZES = {"small": 1000000, "medium": 2000000, "large": 6000000}
 FILE_SIZE = _SIZES[DATA_SIZE]
 LARGE_DATA = "".join([random.choice(string.ascii_letters) for _ in range(FILE_SIZE)])
 DOC_ID_SIZE = 36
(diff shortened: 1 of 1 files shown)


connectors/sources/tests/fixtures/sharepoint/fixture.py
```diff
@@ -13,11 +13,25 @@
...
+    """
+      This methods make sure that all the documemts ids are min 36 bytes like in Sharepoint
+    """
+
+    bytesize = len(id)
+
+    if bytesize >= DOC_ID_SIZE:
+        return id
+
+    addition = "".join([random.choice(string.ascii_letters) for _ in range(DOC_ID_SIZE-bytesize)])
```
wangch079, line 33: Do you need to make sure the same document always has the same id? considering you may need to execute multiple syncs in the test?

vidok: Yes, you are right. It should not be `random`. Good catch!

wangch079 (approved):

wangch079 (approved):

PR #891, opened 2023-05-09
[POC] statsd UDP support for perf metrics

tarekziade:
Adds a statsd client to send metrics to perf8 via UDP for negligible overhead


tarekziade:
Example on the CI for postgresql :

Screenshot 2023-05-08 at 17 59 49

tarekziade:
Jira bulk operations

Screenshot 2023-05-08 at 18 10 17

vidok:
> cc @vidok if we'll use statsd

I hope we won't as it brings another dependency :)

PR #902, opened 2023-05-11
[Utils] Dynamic retryable function wrapper

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4525

With the current `@retryable` decorator we're not able to set `retries`, `retry_interval` at runtime using fields as decorators are not dynamically updated at runtime. Therefore we need a wrapper function, which supports this behavior to support dynamic advanced configuration values for retrying.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -330,22 +330,6 @@ def generate_id(tables, row, primary_key_columns):
         f"{'_'.join([str(pk_value) for pk in primary_key_columns if (pk_value := row.get(pk)) is not None])}"
     )
 
-    @retryable(
-        retries=RETRIES,
-        interval=RETRY_INTERVAL,
-        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
-    )
-    async def get_last_update_time(self, table):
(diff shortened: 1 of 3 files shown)


tarekziade:
why two distinct decorators? you can merge both and have both mechanism from with one single smart decorator

timgrein:
> why two distinct decorators? you can merge both and have both mechanism from with one single smart decorator

What do you mean by smart decorator? Is there an option in python to have a decorator, which supports something like this: 
```
@retryable(retries=self.retries, ...)
def some_method()
```

Specifically referencing a field of an object, which can be updated during runtime.

tarekziade:
> > why two distinct decorators? you can merge both and have both mechanism from with one single smart decorator
> 
> What do you mean by smart decorator? Is there an option in python to have a decorator, which supports something like this:
> 
> ```
> @retryable(retries=self.retries, ...)
> def some_method()

yeah the `retryable` function can have this signature :

```
def retryable(*args, **kw):
   if callable(args[0]):
       func =  args[0]
       etc..
   else:
      etc..
```

So you don't have two different decorators, it stays an internal implementation detail

artem-shelkovnikov:
@timgrein should we close this PR or follow-up?

PR #905, opened 2023-05-11
renamed byoei.* to connectors.es.*

tarekziade:
- `byoei.Fetcher` -> `connectors.es.sink.Extractor`
- `byoei.Bulker` -> `connectors.es.sink.Sink`
- `byoei.ElasticServer` -> `connectors.es.sink.SyncOrchestrator`

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -4,14 +4,14 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 """
-`ElasticServer` is orchestrating a sync by:
+`SyncOrchestrator` is orchestrating a sync by:
 
 - creating a queue
 - launching a `Extractor`, a wrapper on the top of the documents' generator
 - launching a `Sink`, a class that aggregates documents and run the bulk API
 
 
-                  ElasticServer.async_bulk(generator)
+                  SyncOrchestrator.async_bulk(generator)
                                |
                                |
 Elasticsearch <== Sink <== queue <== Extractor <== generator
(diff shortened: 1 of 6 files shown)


connectors/es/sink.py
```diff
@@ -4,19 +4,17 @@
...
 #
 """
-Implementation of BYOEI protocol (+some ids collecting)
-
 `ElasticServer` is orchestrating a sync by:
 
 - creating a queue
-- launching a `Fetcher`, a wrapper on the top of the documents' generator
-- launching a `Bulker`, a class that aggregates documents and run the bulk API
+- launching a `Extractor`, a wrapper on the top of the documents' generator
```
timgrein, line 10: From the doc string I would derive a name like `DocsGenerator`.

connectors/es/sink.py
```diff
@@ -461,18 +459,18 @@ class ElasticServer(ESClient):
 
     - grabs all ids on Elasticsearch for the index
     - creates a MemQueue to hold documents to stream
-    - runs a `Fetcher` (producer) and a `Bulker` (consumer) against the queue
+    - runs a `Extractor` (producer) and a `Sink` (consumer) against the queue
```
timgrein, line 462: We could also name both `DocsProducer` and `DocsConsumer`, but `Sink` is definitely better than `Bulker`, so also happy with that option.

connectors/es/sink.py
seanstory: "generator" is a python language term, but "DocsGenerator" would read to me as if we're fabricating the data (like Faker) as opposed to pulling the data from somewhere.

seanstory (approved):
"SyncOrchestrator" is 👨‍🍳 👌 

I still think that Extractor and SyncOrchestrator probably don't belong in `connectors.es.sync.py`, but that can be a followup.

PR #908, opened 2023-05-15
Fix problem with incorrect mappings created for content indices

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/4465

When framework attempts to create a search index mappings, it does not actually provide "dynamic" and "dynamic_templates" settings to the `put_mappings` call. This PR changes it and actually starts providing these options.

#### Pre-Review Checklist

### DIFF
--- a/connectors/tests/test_byoei.py
+++ b/connectors/tests/test_byoei.py
@@ -19,6 +19,7 @@ from connectors.byoei import (
     Fetcher,
     IndexMissing,
 )
+from connectors.es.settings import TEXT_FIELD_MAPPING
 from connectors.protocol import Pipeline
 from connectors.tests.commons import AsyncIterator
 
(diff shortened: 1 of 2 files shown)


timgrein (approved):
LGTM

connectors/tests/test_byoei.py
```diff
@@ -84,7 +85,17 @@ async def test_prepare_content_index(mock_responses):
...
+    dynamic_templates = {
+        "data": {
+            "match_mapping_type": "string",
+            "mapping": TEXT_FIELD_MAPPING,
+        }
+    }
+    mappings = {
+        "dynamic": True,
+        "dynamic_templates": dynamic_templates,
+        "properties": {"name": {"type": "keyword"}},
```
seanstory, line 97: I see that this was in here before... but I don't think it should be? A `name` property is not in the ruby connectors' mappings, it has `id`.

https://github.com/elastic/connectors-ruby/blob/main/lib/utility/elasticsearch/index/mappings.rb#L92

```suggestion
        "properties": {"id": {"type": "keyword"}},
```

artem-shelkovnikov: It should be fine as this specific function allows arbitrary mappings and this test verifies that these arbitrary mappings are passed correctly.

If we wanna test that our default mappings are passed properly, it should be in a different place, the one that calls `prepare_content_index`

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.8|[](https://github.com/elastic/connectors-python/pull/913)|

### Questions ?

PR #909, opened 2023-05-15
Add malformed configuration error to document

navarone-feekery:
## Closes https://github.com/elastic/enterprise-search-team/issues/4219

When configuration is missing fields, we raise an error. That information doesn't make it to Kibana, because it is never added to the `error` document field.

This PR adds the missing error information to the document.

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -55,7 +55,6 @@ __all__ = [
     "Connector",
     "Features",
     "Filtering",
-    "MalformedConfigurationError",
     "Sort",
     "SyncJob",
 ]
@@ -115,10 +114,6 @@ class ServiceTypeNotConfiguredError(Exception):
 
 
 class DataSourceError(Exception):
-    pass
-
-
-class MalformedConfigurationError(Exception):
     pass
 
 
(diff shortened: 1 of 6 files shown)


connectors/protocol/connectors.py
```diff
@@ -750,6 +752,19 @@ def validate_configuration_fields(
...
+                "status": "error",
+                "error": f'Configuration fields are missing: {", ".join(missing_fields)}.',
+            }
+            await self.index.update(
+                doc_id=self.id,
+                doc=doc,
+                if_seq_no=self._seq_no,
+                if_primary_term=self._primary_term,
+            )
+            await self.reload()
```
artem-shelkovnikov, line 766: I think it should not be done this way - current validation errors make it to the `error` field if they are specified in connector-specific validation method.

Is it possible to rearrange the sync_job_runner code in such a way, that this error in line 768 is caught by same process that marks job as error?

artem-shelkovnikov (changes requested):

connectors/protocol/connectors.py
navarone-feekery: @artem-shelkovnikov I've moved the field validation to the `sync_job_runner` code, but left the field property validations in the connector's `prepare` method. It doesn't raise an error and just populates missing field properties so I don't think it's necessary to move. WDYT?

artem-shelkovnikov: To me it sounds good, if it would raise an error and needed to report it, then it would need to be moved to sync_job_runner as well :)

navarone-feekery:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.8|[](https://github.com/elastic/connectors-python/pull/914)|

### Questions ?

Issue #915, opened 2023-05-16
GitHub: Ping feature


abhishekjoshi-crest:

abhishekjoshi-crest:
Closes this issue as #916 got merged

PR #916, opened 2023-05-16
[New Connector] GitHub: Ping feature

abhishekjoshi-crest:
## Closes https://github.com/elastic/connectors-python/issues/915

Added Ping Functionality for GitHub Connector
#### Pre-Review Checklist


moxarth-elastic:
buildkite test this

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+                    invalid_repo.append(repo)
+            return invalid_repo
+        except Exception as exception:
+            logger.exception(
+                f"Error while filtering repositories. Exception: {exception}."
+            )
+            raise
+
+    @cached_property
+    def _get_session(self):
```
timgrein, line 133:
```suggestion
    def _get_client(self):
```

nit: Maybe rename it to something different as it could be confusing having `_get_session` referring to the cached GitHub "session" and `session` referring to the http session.

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+
+    async def ping(self):
+        await self.execute_getitem(request=ENDPOINTS["USER"])
+
+    async def close_session(self):
+        self._sleeps.cancel()
+        if self.session is None:
+            return
+        await self.session.close()
+        del self._get_session
```
timgrein, line 149:
```suggestion
        if self.session is not None:
            await self.session.close()
            del self._get_session
```

Small nit

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+        self.repos = self.configuration["repositories"]
+        self.github_token = self.configuration["github_token"]
+        self.headers = {"Authorization": f"Bearer {self.github_token}"}
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def execute_getitem(self, request):
```
timgrein, line 51:
```suggestion
    async def get_item(self, resource):
```

As we're passing a URI I think `request` is a misleading name.

We could also use an enumeration for the different resources + a dictionary:

```
class GitHubResource(Enum):
   USER = ...
   REPOSITORY = ...
   
URIS = {
  REPOSITORY: {
     ALL: "/user/repos?type=all",
     BY_NAME: "repos/{repo_name}"
  }
}
```

Feel free to use this approach or just rename the method.

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+            request (str): API for the request
+
+        Returns:
+            list: Response of the request
+        """
+        status_code, response_headers, data = await self._get_session._request(
+            method="get", url=USER_API, headers=self.headers
+        )
+        if status_code == 200:
+            return response_headers
```
timgrein, line 80: question: Why are we only returning the headers and not the data? Do we only need data from the `response_headers`?

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+        """
+        async for response in self._get_session.getiter(url=request):
+            yield response
+
+    async def filter_repos(self):
+        try:
+            logger.debug("Filtering repositories")
+            public_repos, configured_repos, users_repos = [], [], []
+            for repo in self.repos:
+                if "/" in repo:
```
timgrein, line 109: Is a repository always public, if it contains a `/`?

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+        Args:
+            request (str): API for the request
+
+        Yields:
+            dict: Response of the request
+        """
+        async for response in self._get_session.getiter(url=request):
+            yield response
+
+    async def filter_repos(self):
```
timgrein, line 104:
```suggestion
    async def get_invalid_repos(self):
```

I see we're returning `invalid_repo`. Should this be reflected in the method name?

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+            public_repos, configured_repos, users_repos = [], [], []
+            for repo in self.repos:
+                if "/" in repo:
+                    public_repos.append(repo)
+                else:
+                    configured_repos.append(repo)
+
+            async for response in self.execute_getiter(request=ENDPOINTS["ALL_REPOS"]):
+                users_repos.append(response["name"])
+            invalid_repo = list(set(configured_repos) - set(users_repos))
```
timgrein, line 116:
```suggestion
            invalid_repos = list(set(configured_repos) - set(users_repos))
```

suggestion: For consistency I would use plural here.

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+                "order": 3,
+                "required": False,
+                "type": "int",
+                "ui_restrictions": ["advanced"],
+                "value": RETRIES,
+                "validations": [{"type": "less_than", "constraint": 10}],
+            },
+        }
+
+    async def _remote_validation(self):
```
timgrein, line 203: Should this use `@retryable?`

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+        """Validate scope of the configured Token and accessibility of repositories
+
+        Raises:
+            ConfigurableFieldValueError: Insufficient privileges error.
+        """
+        response_headers = await self.github_client.execute_request(
```
timgrein, line 209: I think it makes sense to have a high-level method here present in the client. For example: `get_user`

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+            method="get", url=USER_API
+        )
+        if "repo" not in response_headers.get("X-OAuth-Scopes"):  # pyright: ignore
+            raise ConfigurableFieldValueError(
+                "Configured token does not have required rights to fetch the content"
+            )
+        if self.github_client.repos != ["*"]:
+            invalid_repos = await self.github_client.filter_repos()
```
timgrein, line 217: suggestion: Maybe it makes sense to have a method fetching all public and private repositories for a user and do the check, if they match with the configured repositories (`list(set(configured_repos) - set(users_repos))`) in the data source and not in the client. To me it feels like only fetching the repositories should be the concern of the client and checking if it makes sense in terms of the configuration is a concern of the data source.

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+        return GitHubAPI(
+            self.session,
+            requester="",
+            oauth_token=self.github_token,
+        )
+
+    async def ping(self):
+        await self.execute_getitem(request=ENDPOINTS["USER"])
+
+    async def close_session(self):
```
timgrein, line 144:
```suggestion
    async def close(self):
```

nit: I think we can call this method `close` as if we're looking strictly at it, it does not only close one session, but two and it also cancels possible pending sleeps.

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+        raise Exception(
+            f"Something went wrong while executing request Exception: {message['message']}"
+        )
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def execute_getiter(self, request):
```
timgrein, line 92:
```suggestion
    async def get_resource(self, resource):
```

suggestion: I think we can call this method `get_resource`, `yield_resource` or `resource_generator` to make it more explicit what's returned by that method.

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+        self.configuration.check_valid()
+        await self._remote_validation()
+
+    async def close(self):
+        await self.github_client.close_session()
+
+    async def ping(self):
+        try:
+            await self.github_client.ping()
+            logger.debug("Successfully connected to the GitHub.")
```
timgrein, line 236:
```suggestion
            logger.debug("Successfully connected to GitHub.")
```

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+
+        except Exception:
+            logger.exception("Error while connecting to the GitHub.")
```
timgrein, line 239:
```suggestion
            logger.exception("Error while connecting to GitHub.")
```

connectors/sources/tests/test_github.py
```diff
@@ -0,0 +1,148 @@
...
+from gidgethub import BadRequest
+
+from connectors.source import ConfigurableFieldValueError, DataSourceConfiguration
+from connectors.sources import github
+from connectors.sources.github import GitHubDataSource
+from connectors.sources.tests.support import create_source
+from connectors.tests.commons import AsyncIterator
+
+
+def test_get_configuration():
```
timgrein, line 20:
```suggestion
def test_get_default_configuration():
```

connectors/sources/tests/test_github.py
```diff
@@ -0,0 +1,148 @@
...
+    ):
+        with pytest.raises(Exception):
+            await source.ping()
+
+
+@pytest.mark.asyncio
+async def test_close_without_session():
+    source = create_source(GitHubDataSource)
+    await source.close()
+    source.github_client._get_session
```
timgrein, line 63: What's the purpose of this line? Do you mean `source.github_client._get_session()`. We should also assert something here, i.e. that the correct closing methods were called.

connectors/sources/tests/test_github.py
```diff
@@ -0,0 +1,148 @@
...
+    await source.close()
+
+
+@pytest.mark.asyncio
+async def test_close_with_session():
+    source = create_source(GitHubDataSource)
+    source.github_client._get_session
```
timgrein, line 70: Same question as above

connectors/sources/tests/test_github.py
```diff
@@ -0,0 +1,148 @@
...
+        await source.validate_config()
+
+
+@pytest.mark.asyncio
+async def test_validate_config_with_unauthorized_user():
+    source = create_source(GitHubDataSource)
+    source.github_client.execute_request = mock.AsyncMock(
+        return_value={"X-OAuth-Scopes": ""}
+    )
+    with pytest.raises(Exception):
```
timgrein, line 93: Why aren't we expecting `ConfigurableFieldValueError`? Should we change something in the implementation?

timgrein (commented):
Mainly comments around naming and where to place things, otherwise looking good!

connectors/sources/github.py
moxarth-elastic: We get the scope of the token in `response_headers` and currently we use the _request method of gidgethub to fetch the headers only to validate the token. In the future, if we use this function and require the response, we will return the whole response.

connectors/sources/github.py
moxarth-elastic: Currently, we have added `@retryable` to the function that makes an API call.

connectors/sources/github.py
moxarth-elastic: Here we need `response_headers` only to validate the github token, and we are not fetching the user's data. So, WDYT, do we need to add the method `get_user`.

connectors/sources/tests/test_github.py
moxarth-elastic: Sure, adding `ConfigurableFieldValueError`

connectors/sources/github.py
abhishekjoshi-crest: We need `response_headers` only to validate the github token, as we are getting scopes of the token in `response_headers`. In the future, if we are planning to use this method, and if required, we will also return a response.

connectors/sources/github.py
abhishekjoshi-crest: If the user wants to fetch repos of a different user, then the user needs to configure the repo name as `owner/repo_name`. As `public_repos` is misleading, I will change the variable name to `external_repos`.

connectors/sources/github.py
abhishekjoshi-crest: Currently, we have added `@retryable` to functions that make API calls.

connectors/sources/github.py
abhishekjoshi-crest: But we want `response_headers` only and `execute_request` returns only `response_headers`. So, WDYT do we need to add `get_user` method.

connectors/sources/github.py
timgrein: We could also move that to the data source, so the caller can decide whether a particular request should be retryable or not.

connectors/sources/github.py
```diff
@@ -0,0 +1,252 @@
...
+USER_API = "https://api.github.com/user"
+ENDPOINTS = {
+    "ALL_REPOS": "/user/repos?type=all",
+    "REPO": "repos/{repo_name}",
+    "USER": "/user",
+}
+
+
+class GitHubClient:
+    """GitHub Client"""
```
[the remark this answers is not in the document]
abhishekjoshi-crest: Done

connectors/sources/github.py
abhishekjoshi-crest: sure, updated

connectors/sources/github.py
abhishekjoshi-crest: We think enum become complex so currently rename the method

connectors/sources/github.py
abhishekjoshi-crest: sounds good

connectors/sources/github.py
abhishekjoshi-crest: Yes, added retryable in data source

connectors/sources/github.py
abhishekjoshi-crest: Done

connectors/sources/tests/test_github.py
abhishekjoshi-crest: Yes to get the client we called `_get_session` and as it is `cached_property` and returns github client we did not call like this `source.github_client._get_session()`

connectors/sources/tests/test_github.py
abhishekjoshi-crest: https://github.com/elastic/connectors-python/pull/916#discussion_r1197692140

connectors/sources/tests/test_github.py
abhishekjoshi-crest: updated `ConfigurableFieldValueError`

Issue #917, opened 2023-05-16
Dropbox: Ping Feature


moxarth-elastic:

vidok:
Could you please fill in the PR description or add references?

PR #918, opened 2023-05-16
[New Connector] Dropbox: Ping feature

moxarth-elastic:
## Closes https://github.com/elastic/connectors-python/issues/917

Added a Ping Functionality for Dropbox connector.

#### Pre-Review Checklist

- [x] New external service dependency added - `Dropbox`


connectors/sources/dropbox.py
```diff
@@ -0,0 +1,173 @@
...
+
+import dropbox
+import requests
+from dropbox.exceptions import ApiError, AuthError, BadInputError
+
+from connectors.logger import logger
+from connectors.source import BaseDataSource, ConfigurableFieldValueError
+from connectors.utils import CancellableSleeps, iso_utc
+
+RETRIES = 3
```
vidok, line 18: Let's rename it to `RETRY_COUNT` to align with the SDK configuration

connectors/sources/dropbox.py
```diff
@@ -0,0 +1,173 @@
...
+            app_key=self.configuration["app_key"],
+            app_secret=self.configuration["app_secret"],
+            oauth2_refresh_token=self.configuration["refresh_token"],
+            max_retries_on_error=self.retry_count,
+            max_retries_on_rate_limit=self.retry_count,
+            session=self._session,
+        )
+
+    def ping(self):
+        self._create_connection.users_get_current_account()
```
vidok, line 47: How do you handle response errors or exceptions?

connectors/sources/dropbox.py
```diff
@@ -0,0 +1,173 @@
...
+        try:
+            if self.dropbox_client.path not in ["", None]:
+                self.dropbox_client.check_path()
+        except BadInputError:
+            raise ConfigurableFieldValueError(
+                "Configured App Key or App Secret is invalid"
+            )
+        except AuthError:
+            raise ConfigurableFieldValueError("Configured Refresh Token is invalid")
+        except ApiError:
```
vidok, line 150: I think is this not the right exception type to validate the path. `ApiError` is a general exception for all API errors. Shouldn't this be `dropbox.exceptions.PathRootError`?

connectors/sources/dropbox.py
```diff
@@ -0,0 +1,173 @@
...
+            raise ConfigurableFieldValueError(
+                f"Configured Path: {self.dropbox_client.path} is invalid"
+            )
+        except Exception:
```
vidok, line 154: It would be useful to pass the exception message?

connectors/sources/dropbox.py
moxarth-elastic: I think we can handle errors in the DataSource's ping method same as we did in `_remote_validation`.

connectors/sources/dropbox.py
vidok: You are right, however, if look at the exception object you can see that it breaks down the error. There is a LookupError inside which indicates what exactly went wrong. 

To be more precise in API error handling, we need to handle this particular error. Other ApiError exceptions can go to the general errors.

moxarth-elastic: I see, let me make it more precise by catching the `LookupError`.

connectors/sources/dropbox.py
```diff
@@ -0,0 +1,179 @@
...
+        self.path = (
+            ""
+            if self.configuration["path"] in ["/", None]
+            else self.configuration["path"]
+        )
+        self.retry_count = self.configuration["retry_count"]
+        self._session = None
+
+    @cached_property
+    def _create_connection(self):
```
vidok, line 36: I would prefer to name the method `_connection`.

vidok (approved):

PR #919, opened 2023-05-16
make sure we switch to the right branch

tarekziade:
### Closes https://github.com/elastic/enterprise-search-team/issues/4488

The docker image was cloning on main only, this fixes it


tarekziade:
Nightly run on the branch https://buildkite.com/elastic/connectors-python-nightly/builds/640
(should pick the branch to build the docker)

artem-shelkovnikov (approved):

seanstory:
This closed https://github.com/elastic/enterprise-search-team/issues/4488, but the Enterprise Search cloud image doesn't use connectors docker images, does it? I don't see how this would fix that issue - I thought that ent-search cloned the repo directly.

seanstory:
Thanks for reopening, @artem-shelkovnikov

PR #924, opened 2023-05-17
[SharePoint] Pagination support in the fake SharePoint server

akanshi-elastic:
#### Pre-Review Checklist


connectors/sources/sharepoint.py
```diff
@@ -36,7 +36,10 @@
 RETRIES = 3
 FILE_SIZE_LIMIT = 10485760
 CHUNK_SIZE = 1024
-TOP = 5000
+if "RUNNING_FTEST" in os.environ:
```
vidok, line 39: This is not needed anymore. Could you please remove it?

tests/sources/fixtures/sharepoint/fixture.py
```diff
@@ -107,21 +109,27 @@ def get_sites(site_collections):
...
-                "Id": adjust_document_id_size("sites-0"),
-                "LastItemModifiedDate": "2023-02-16T06:48:30Z",
-                "ServerRelativeUrl": "/sites/site1",
-                "Title": "site1",
-                "Url": f"http://127.0.0.1:8000/sites/{site_collections}/site1",
-            }
-        )
+    if skip:
+        skip_start = int(skip)
+        if skip_start + 100 >= total_subsites:
```
vidok, line 118: I'm not sure I understand how this works

connectors/sources/sharepoint.py
akanshi-elastic: We have added this condition to decrease the pagination size.. otherwise it will return 5000 response for one api call and if those 5000 response process further for lists so it will return huge number of records which takes more than 10minutes and limit will exceed for ftest in that case.

tests/sources/fixtures/sharepoint/fixture.py
akanshi-elastic: We've used this condition for handling pagination so that if pagination limit is 100 and number of documents we are passing is 150 then in that case this will handle the second api call for the remaining 50 documents.

connectors/sources/sharepoint.py
vidok: I'm not sure I get this. 

There was a configuration `start_lists, end_lists = 0, 450` that generated about 5k docs in total. If I requests TOP=10k but there are only 1k of docs shouldn't the server return only 1k docs?

tests/sources/fixtures/sharepoint/fixture.py
vidok: What it the pagination limit is changed? Shouldn't it be picked up from the URL arguments?

vidok (approved):

PR #927, opened 2023-05-18
fix pyright on latest 3.10

tarekziade:
We have a few warnings on pyright


seanstory (approved):

seanstory:
merging, so it doesn't block other PRs

PR #937, opened 2023-05-23
[DLS] Add "get_permissions" to BaseDataSource

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4551

This PR adds the `get_permissions` method to the `BaseDataSource`, which should be implemented by connectors supporting DLS.

#### Pre-Review Checklist

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -467,6 +467,13 @@ class BaseDataSource:
         """
         pass
 
+    async def get_permissions(self):
+        """Returns an asynchronous iterator on the permission documents present in the backend.
+
+        Each document is a dictionary containing permission data indexed into a corresponding permissions index.
+        """
+        raise NotImplementedError
+
     async def get_docs(self, filtering=None):
         """Returns an iterator on all documents present in the backend
 
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -472,7 +472,7 @@ class BaseDataSource:
 
         Each document is a dictionary containing permission data indexed into a corresponding permissions index.
         """
-        return []
+        raise NotImplementedError
 
     async def get_docs(self, filtering=None):
         """Returns an iterator on all documents present in the backend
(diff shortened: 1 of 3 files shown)


connectors/source.py
```diff
@@ -467,6 +467,13 @@ async def close(self):
         """
         pass
 
+    async def get_permissions(self):
+        """Returns an asynchronous iterator on the permission documents present in the backend.
+
+        Each document is a dictionary containing permission data indexed into a corresponding permissions index.
+        """
+        raise NotImplementedError
```
wangch079, line 475: This implies every connector needs to implement a `get_permissions` method. Is it true?

wangch079 (commented):
What is expected `get_permissions` implementation for a connector which doesn't support permissions

connectors/source.py
timgrein: Returning an empty list now 👍 My idea was to guard the usage of this method with the corresponding feature flag so there's a "hard" error, when DLS is enabled via a manual index update, but not a strong opinion, so changing it.

wangch079: There's very similar scenario in incremental sync as well. Post it in [slack](https://elastic.slack.com/archives/C01795T48LQ/p1684913931191989) for wider discussion

timgrein: @seanstory Very valid points, either way I would guard these calls behind the feature flag, but as you've said people could update the flag manually and we would expect weird and unexpected behavior in the sense that anyone can look at a document. An error would be more safe IMO.

Thoughts @wangch079 @tarekziade ?

timgrein: Update: We had a discussion with @danajuratoni and from a product perspective it makes sense to hide the usage of `get_permissions` behind a feature flag and raise an error, when someone manually update the feature flag.

tarekziade (approved):

PR #947, opened 2023-05-24
[DLS] Add "last_permissions_sync_scheduled_at" mutator

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4612

This PR adds an update method for the new `last_permissions_sync_scheduled_at` field of a connector document. This is needed, when a job scheduler scheduled a permissions sync job, so other connector service instances don't schedule the same permissions sync job multiple times.

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -556,6 +556,14 @@ class Connector(ESDocument):
             if_primary_term=self._primary_term,
         )
 
+    async def update_last_permissions_sync_scheduled_at(self, new_ts):
+        await self.index.update(
+            doc_id=self.id,
+            doc={"last_permissions_sync_scheduled_at": new_ts.isoformat()},
+            if_seq_no=self._seq_no,
+            if_primary_term=self._primary_term,
+        )
+
     async def sync_starts(self):
         doc = {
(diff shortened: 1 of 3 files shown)


connectors/protocol/connectors.py
```diff
@@ -556,6 +556,14 @@ async def update_last_sync_scheduled_at(self, new_ts):
             if_primary_term=self._primary_term,
         )
 
+    async def update_last_permissions_sync_scheduled_at(self, new_ts):
```
tarekziade, line 559: I am wondering if we need an explicit API for this ? (especially given the very long name) as opposed to something like `connector.update(last_permissions_sync_scheduled_at=ts.isoformat())`

wangch079: > as opposed to something like `connector.update(last_permissions_sync_scheduled_at=ts.isoformat())`

One concern about exposing `update` is that you can actually update anything.

`ESDocument` and its subclasses are just data entity class (only read via `@property`), and we guard field mutating by specific methods, e.g. `heartbeat` will only allow you to update `last_seen`.

wangch079 (approved):

PR #956, opened 2023-05-25
Add Confluence advanced sync rules

### LINKED ISSUE #954: add support for advanced sync rules as defined in the document (also referred to as remote filtering)

praveen-elastic:
## Closes #954 

This PR introduces advanced sync rules for Confluence by using CQL (Confluence Query Language) queries to filter out confluence data. The connector utilises `atlassian.py` file for validating the query.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -59,6 +59,7 @@ END_SIGNAL = "FINISHED_TASK"
 
 CONFLUENCE_CLOUD = "confluence_cloud"
 CONFLUENCE_SERVER = "confluence_server"
+WILDCARD = "*"
 
 
 class ConfluenceClient:
@@ -263,7 +264,7 @@ class ConfluenceDataSource(BaseDataSource):
                 "label": "Confluence space keys",
                 "order": 7,
                 "type": "list",
-                "value": "*",
+                "value": WILDCARD,
             },
             "ssl_enabled": {
                 "display": "toggle",
@@ -330,7 +331,7 @@ class ConfluenceDataSource(BaseDataSource):
         await self._remote_validation()
 
     async def _remote_validation(self):
-        if self.spaces == ["*"]:
+        if self.spaces == [WILDCARD]:
             return
         space_keys = []
         async for response in self.confluence_client.paginated_api_call(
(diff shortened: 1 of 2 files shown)


seanstory:
Should this also add sync rule docs for confluence here https://github.com/elastic/connectors-python/tree/main/docs/sync-rules ?

connectors/sources/confluence.py
```diff
@@ -442,6 +449,58 @@ async def fetch_attachments(
...
+                        document.update(
+                            {
+                                "size": entity_details.get("extensions", {}).get(
+                                    "fileSize"
+                                ),
+                                container: entity_details.get("container", {}).get(
+                                    "title"
+                                ),
+                            }
+                        )
```
timgrein, line 498:
```suggestion
                    file_size =  entity_details.get("extensions", {}).get("fileSize")
                    title = entity_details.get("container", {}).get("title")
                    document.update(
                         {
                             "size": file_size,
                             "container": title
                         }
                     )
```

I think using variables would improve readability here. Also container should be in quotes, right?

connectors/sources/confluence.py
```diff
@@ -442,6 +449,58 @@ async def fetch_attachments(
...
+                        document.pop("body")
```
timgrein, line 499:
```suggestion
                        # Removing body as attachment will be downloaded lazily
                        document.pop("body")
```

nit: Maybe a comment would be useful, why we're removing the body

connectors/sources/confluence.py
```diff
@@ -557,25 +616,42 @@ async def get_docs(self, filtering=None):
...
+
         else:
-            quoted_spaces = "','".join(self.spaces)
-            configured_spaces_query = f"cql=space in ('{quoted_spaces}') AND type="
-        await self.fetchers.put(self._space_coro)
-        await self.fetchers.put(
-            partial(
-                self._page_blog_coro,
-                f"{configured_spaces_query}blogpost&{CONTENT_QUERY}",
+            if self.spaces == ["*"]:
```
timgrein, line 635:
```suggestion
            if self.spaces == [WILDCARD]:
```

suggestion: I think we should use a constant here as `*` could mean anything

connectors/sources/confluence.py
praveen-elastic: Agreed. I've updated with variable names to improve readability. Also for container variable I've renamed it to `container_type` as it denotes the type of container.

connectors/sources/confluence.py
praveen-elastic: Good catch, updated [cc07dee](https://github.com/elastic/connectors-python/pull/956/commits/cc07deeb07c1485d5df46659357d0be281cc2a1c)

timgrein (approved):
LGTM!

PR #963, opened 2023-05-26
Update sync_cursor for a completed sync job

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4626

Adds mutating of `sync_cursor` to the `sync_done` method of the `Connector` class, so that the `sync_cursor` will only be updated after a completed sync job.

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -625,7 +625,7 @@ class Connector(ESDocument):
         }
         await self.index.update(doc_id=self.id, doc=doc)
 
-    async def sync_done(self, job):
+    async def sync_done(self, job, sync_cursor=None):
         job_status = JobStatus.ERROR if job is None else job.status
         job_error = JOB_NOT_FOUND_ERROR if job is None else job.error
         if job_error is None and job_status == JobStatus.ERROR:
@@ -641,6 +641,10 @@ class Connector(ESDocument):
             "status": connector_status.value,
             "error": job_error,
         }
+
+        # only update sync cursor after a successful sync job
+        if job_status == JobStatus.COMPLETED:
+            doc["sync_cursor"] = sync_cursor
 
         if job is not None and job.terminated:
             doc["last_indexed_document_count"] = job.indexed_document_count
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -625,7 +625,7 @@ class Connector(ESDocument):
         }
         await self.index.update(doc_id=self.id, doc=doc)
 
-    async def sync_done(self, job, sync_cursor=None):
+    async def sync_done(self, job, cursor=None):
         job_status = JobStatus.ERROR if job is None else job.status
         job_error = JOB_NOT_FOUND_ERROR if job is None else job.error
         if job_error is None and job_status == JobStatus.ERROR:
@@ -644,7 +644,7 @@ class Connector(ESDocument):
 
         # only update sync cursor after a successful sync job
         if job_status == JobStatus.COMPLETED:
-            doc["sync_cursor"] = sync_cursor
+            doc["sync_cursor"] = cursor
 
         if job is not None and job.terminated:
             doc["last_indexed_document_count"] = job.indexed_document_count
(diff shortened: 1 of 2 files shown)


timgrein (dismissed):
LGTM, only hinting at some indentations

connectors/protocol/connectors.py
```diff
@@ -621,6 +621,10 @@ async def sync_done(self, job):
             "error": job_error,
         }
 
+        # only update sync cursor after a successful sync job
+        if job_status == JobStatus.COMPLETED:
```
vidok, line 625: I have 2 comments here: 

1. Would it be possible for a sync job that was not completed with success to report the current progress? So the next job can pick it up? 
2. I have a feeling that `sync_done` should not decide if the cursor needs to be saved or not. This is just a "reporting" method so I would not add any logic into it.

wangch079: >Would it be possible for a sync job that was not completed with success to report the current progress? So the next job can pick it up?

I considered this option when drafting the implementation doc, but I decided to only update the cursor after a successful job.

E.g. A sync job fetched 100 docs and pushed them into the queue, then encountered an error. Let's see 50 out of the 100 docs are indexed into ES. If we update the `sync_cursor` when failing the sync job, it will miss the 50 docs in the queue when running the next sync job.

So it's much safer to update `sync_cursor` only after a successful sync job. We can iterate further in the later release.

wangch079: > I have a feeling that `sync_done` should not decide if the cursor needs to be saved or not. This is just a "reporting" method so I would not add any logic into it.

IMO, `sync_done` is to update the connector when a sync job is done (either successfully or not), so it contains logic to handle different scenarios, e.g. 1) make sure `last_sync_error` is set if the job status is `error`, 2) only update `last_indexed_document_count` and `last_deleted_document_count` when the sync job is terminated. So I feel it makes sense to have this check here.

tests/protocol/test_connectors.py
```diff
@@ -232,7 +232,7 @@ def test_utc():
     ],
 )
 async def test_supported_connectors(
-    native_service_types, connector_ids, expected_connector_count, mock_responses
+        native_service_types, connector_ids, expected_connector_count, mock_responses
```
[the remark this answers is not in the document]
wangch079: reverted

connectors/protocol/connectors.py
vidok: > I considered this option when drafting the implementation doc, but I decided to only update the cursor after a successful job.
E.g. A sync job fetched 100 docs and pushed them into the queue, then encountered an error. Let's see 50 out of the 100 docs are indexed into ES. If we update the sync_cursor when failing the sync job, it will miss the 50 docs in the queue when running the next sync job.
So it's much safer to update sync_cursor only after a successful sync job. We can iterate further in the later release.

Get it. It makes sense to me for this iteration. What do you do when the incremental sync failed? Can we just restart it without changing the cursor?

wangch079: >What do you do when the incremental sync failed? Can we just restart it without changing the cursor?

It just fails. `error` is a terminal status. Customers have to trigger/schedule another incremental sync, which will take the cursor from the connector, i.e. the same cursor as the failed incremental sync.

vidok: > IMO, sync_done is to update the connector when a sync job is done (either successfully or not), so it contains logic to handle different scenarios, e.g. 1) make sure last_sync_error is set if the job status is error, 2) only update last_indexed_document_count and last_deleted_document_count when the sync job is terminated. So I feel it makes sense to have this check here.

`sync_done` is already overloaded with business logic. Adding more if/else conditions to it will make it even more complex. Let's imagine we want to update cursors when sync is failed then you will need to add another if/else to the `error` or `sync_done` methods and so on. In my ideal world it should look like this: 

```
def sync_done(args)
  doc = update_status(doc, args)
  doc = update_error(doc, args)
  doc = update_cursor(doc, args)
  
  store(doc)
```

So I would at least incapsulate updating cursor logic even if you want to call inside `sync_done`. 

_It's not a blocker as we need to move forward._

Issue #940, opened 2023-05-23
[Sharepoint] Exception: 400, message='Bad Request'


llermaly:
## Bug Description
I'm having a bad request with a specific document URL:

[FMWK][10:51:04][WARNING] Retry count: 3 out of 3. Exception: 400, message='Bad Request', url=URL('https://tenant.sharepoint.com/sites/site/_api/web/lists/getbytitle('Site%20Pages')/items?$select=Title,CanvasContent1,FileLeafRef&$filter=FileLeafRef+eq+'Some-File--Name'.aspx'')

My guess is the ".aspx" in the name is messing with the parsing. I tried in Postman and can reproduce the error, then if I fix the single quote:

'Some-File--Name.aspx'

It works.

### To Reproduce
1. Create sharepoint site with list title with .aspx
2. Run sharepoint crawler
3. See error

## Expected behavior
Object should ingest.

## Screenshots
N/A

## Environment
- Elastic Cloud 8.8
- Sharepoint connector main branch

## Additional context
N/A

danajuratoni:
split from original issue cc: @akanshi-elastic

akanshi-elastic:
Hey @llermaly  , I checked the issue by creating a site page with a name having single/double quote: `Some-File--Name'.aspx` 
And it is throwing 400 bad request on postman as well. I tried to encode the single/double quote but still it's giving 400 error. and it is required to add the quotes around the name of file in url so if we pass single/double quote in between name of file then the url considered it as a closing quote  and throws the error. I also tried to add escape before the quote but that is not working in this case.

So it seems that Sharepoint server is itself unable to parse the single quotes in in their endpoint as we are facing this in the postman too.
if the sharepoint server is not able to process the same request so we can't do much from our side to prevent that.
Let me know your thoughts how do we want to proceed further in this case.
cc: @artem-shelkovnikov

llermaly:
hi @akanshi-elastic , I found another similar than this one: 

`$filter=FileLeafRef+eq+'Some-file-&-name.aspx'`

I could make it work in postman doing 

FileLeafRef+eq+'Some-file-%26-name.aspx'

I was about to create a new issue but looks like the problem is related to the same.

I can make the File--name to work in postman using this notation:

![CleanShot 2023-05-24 at 18 20 45](https://github.com/elastic/connectors-python/assets/8575569/29eaf2c5-956b-417b-916f-d0ed26e3d535)

akanshi-elastic:
[...]

PR #968, opened 2023-05-26
[Sharepoint][Bugfix]Encode Site Pages URL

akanshi-elastic:
## Closes https://github.com/elastic/connectors-python/issues/940###
This PR includes:
1.  Encode Site Pages URL for pages having special characters.
2. Encode url field

#### Pre-Review Checklist


connectors/sources/sharepoint.py
```diff
@@ -322,8 +322,12 @@ async def get_site_pages_content(
             return
 
         source_file_name = ""
+        format_filename = filename.replace("'", "''")
```
tarekziade, line 325: this should be under `if self.is_cloud`
so the `'` char will be replaced by `%27%27` (just making sure that is the intent)

connectors/sources/sharepoint.py
```diff
@@ -570,7 +574,7 @@ async def get_list_items(self, list_id, site_url, server_relative_url, **kwargs)
         ):
             for result in list_items_data:
                 if not result.get("Attachments"):
-                    url = f"{self.host_url}{server_relative_url}/DispForm.aspx?ID={result['Id']}&Source={self.host_url}{server_relative_url}/AllItems.aspx&ContentTypeId={result['ContentTypeId']}"
+                    url = f"{self.host_url}{quote(server_relative_url)}/DispForm.aspx?ID={result['Id']}&Source={self.host_url}{quote(server_relative_url)}/AllItems.aspx&ContentTypeId={result['ContentTypeId']}"
```
tarekziade, line 577: can we add in the class a dedicated function for building URLs, I see now places where you use `join` and places where you use string formatting.

Ideally we have one single function that gets all parts and do the quoting work

tarekziade (commented):
We should vectorize the code that builds URLs in this patch for clarity

connectors/sources/sharepoint.py
akanshi-elastic: Yes to ignore any single quote in mid of the url query .I am adding another single quote as a prefix. otherwise we are getting 400 Bad request error for the file name having single quote.

connectors/sources/sharepoint.py
```diff
@@ -184,6 +184,29 @@ def _get_session(self):
...
 
+    def format_url(
+        self,
+        relative_url,
+        list_item_id=None,
+        content_type_id=None,
+        is_list_item_has_attachment=False,
+    ):
+        if is_list_item_has_attachment:
+            return "".join(
```
tarekziade, line 195: if you are adding strings you can just use `+`

artem-shelkovnikov (approved):

Issue #961, opened 2023-05-26
Add timers for bulk and backend operations


tarekziade:
Related to https://github.com/elastic/connectors-python/issues/923

Let's add a couple of debug timers logs at 

https://github.com/elastic/connectors-python/blob/main/connectors/es/sink.py#L116
and 
https://github.com/elastic/connectors-python/blob/main/connectors/es/sink.py#L356

to measure how fast bulk requests and doc fetchings are happening. This can help us diagnose problems when something is getting slow or gets throttled

PR #971, opened 2023-05-26
@tracer.start_as_current_span

tarekziade:
### Closes https://github.com/elastic/connectors-python/issues/961

Adding some timers to track where we spend time

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -29,7 +29,7 @@ from elasticsearch.helpers import async_scan
 
 from connectors.es import ESClient
 from connectors.filtering.basic_rule import BasicRuleEngine, parse
-from connectors.logger import logger
+from connectors.logger import logger, tracer
 from connectors.protocol import Filter
 from connectors.utils import (
     DEFAULT_CHUNK_MEM_SIZE,
@@ -113,6 +113,7 @@ class Sink:
 
         raise TypeError(operation)
 
+    @tracer.start_as_current_span("_bulk API call")
     async def _batch_bulk(self, operations, stats):
         # TODO: treat result to retry errors like in async_streaming_bulk
         task_num = len(self.bulk_tasks)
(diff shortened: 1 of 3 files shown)


tarekziade:
nightly e2e test build with the new timers https://buildkite.com/elastic/connectors-python-nightly/builds/674

tarekziade:
sample from the buildkite output:

```
[FMWK][22:05:48][DEBUG] get a doc took 0.0004730224609375 seconds.
[FMWK][22:05:48][DEBUG] _bulk API call _batch_bulk took 0.6930327415466309 seconds.
```

tarekziade:
we are going to get too many logs with one log per doc. Added a `slow_log` option so we just log the ones that are abnormally slow

https://buildkite.com/elastic/connectors-python-nightly/builds/675

connectors/es/sink.py
```diff
@@ -339,6 +340,8 @@ async def get_docs(self, generator):
         Extraction happens in a separate task, when a document contains files.
         """
         logger.info("Starting doc lookups")
+        generator = TracedAsyncGenerator(generator, "get a doc")
```
vidok, line 343: Why do you need this generator here?

tarekziade: it's used to wrap the generator that fetches documents (get_docs) so we can trace how long each call to that generator takes

vidok: Can you use the same format you use to decorate `_batch_bulk`?

vidok (approved):

connectors/es/sink.py
tarekziade: Ah good point, I can detect it's an async generator, I have not thought about this, thanks

tarekziade: Done. That was fun!

tarekziade:
oh no, it was already merged. Adding a follow up issue

PR #972, opened 2023-05-26
switch to ruff

tarekziade:
Ruff is way faster than flake8 and also catches more issues.

This patch move the project to use ruff -- it's a drop in replacement.


artem-shelkovnikov (approved):
Good stuff!

timgrein (approved):
Nice! lgtm

PR #975, opened 2023-05-30
Add sync_cursor to Sync Job

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4665

This PR:
1. adds `sync_cursor` property to `SyncJob` class
2. adds `sync_cursor` parameter to the `claim` method of `SyncJob` class
3. adds `sync_cursor` to connector protocol

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -229,6 +229,10 @@ class SyncJob(ESDocument):
         return Pipeline(self.get("connector", "pipeline"))
 
     @property
+    def sync_cursor(self):
+        return self.get("connector", "sync_cursor")
+
+    @property
     def terminated(self):
         return self.status in (JobStatus.ERROR, JobStatus.COMPLETED, JobStatus.CANCELED)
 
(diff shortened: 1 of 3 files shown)


navarone-feekery (approved):
LGTM, but a question for my own curiosity:

how does the `sync_cursor` work? The implementation doc says that it will differ between connectors. What could it look like for a specific connector for example?

wangch079:
> how does the `sync_cursor` work? The implementation doc says that it will differ between connectors. What could it look like for a specific connector for example?

As the doc says, it really depends on connectors. That's why the type of `sync_cursor` is `object`.

Take SPO as an example, SPO supports [delta link to query changes](https://learn.microsoft.com/en-us/graph/delta-query-overview). Suppose we only sync sites, the structure can be:
```json
{
    "sites": {
        [site_id] : <delta link>
    }
}
```

The purpose of this PR (adding `sync_cursor` to `.elastic-connectors-sync-jobs`) is to record the `sync_cursor` used to start the sync job. We could also improve this to periodically update this field during job execution, so that when a sync job is suspended, it can be resumed from where it left off.

PR #985, opened 2023-05-31
unify the metrics API

tarekziade:
Follow up from https://github.com/elastic/connectors-python/pull/971 

I implemented the unified API and did not notice the PR was already merged 

@vidok this is the change to have a single API that works on async generators as well (fun to implement)


artem-shelkovnikov (approved):
Nice!

PR #986, opened 2023-05-31
Add incremental sync support to BaseDataSource

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/4635

This PR
1. adds a class method `support_incremental_sync` to indicate whether the data source supports incremental syn
2. adds a method `get_docs_incrementally` to run incremental sync
3. adds a method `sync_cursor`

#### Pre-Review Checklist

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -339,6 +339,7 @@ class BaseDataSource:
 
     name = None
     service_type = None
+    support_incremental_sync = False
 
     def __init__(self, configuration):
         if not isinstance(configuration, DataSourceConfiguration):
@@ -399,11 +400,6 @@ class BaseDataSource:
         """
 
         return hash_id(_id)
-
-    @classmethod
-    def support_incremental_sync(cls):
-        """Indicates whether the data source supports incremental sync"""
-        return False
 
     async def validate_filtering(self, filtering):
         """Execute all basic rule and advanced rule validators."""
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -348,6 +348,8 @@ class BaseDataSource:
 
         self.configuration = configuration
         self.configuration.set_defaults(self.get_default_configuration())
+        # A dictionary, the structure of which is connector dependent, to indicate a point where the sync is at
+        self._sync_cursor = None
 
     def __str__(self):
         return f"Datasource `{self.__class__.name}`"
@@ -397,6 +399,11 @@ class BaseDataSource:
         """
 
         return hash_id(_id)
+
+    @classmethod
+    def support_incremental_sync(cls):
+        """Indicates whether the data source supports incremental sync"""
+        return False
 
     async def validate_filtering(self, filtering):
         """Execute all basic rule and advanced rule validators."""
@@ -501,6 +508,34 @@ class BaseDataSource:
         """
         raise NotImplementedError
 
+    async def get_docs_incrementally(self, sync_cursor, filtering=None):
+        """Returns an iterator on all documents changed since the sync_cursor
+
+        Each document is a tuple with:
+        - a mapping with the data to index
+        - a coroutine to download extra data (attachments)
+        - an operation, can be one of index, update or delete
+
+        The mapping should have least an `id` field
+        and optionally a `timestamp` field in ISO 8601 UTC
+
+        The coroutine is called if the document needs to be synced
+        and has attachments. It needs to return a mapping to index.
+
+        It has two arguments: doit and timestamp
+        If doit is False, it should return None immediately.
+        If timestamp is provided, it should be used in the mapping.
+
+        Example:
+
+           async def get_file(doit=True, timestamp=None):
+               if not doit:
+                   return
+               return {'TEXT': 'DATA', 'timestamp': timestamp,
+                       'id': 'doc-id'}
+        """
+        raise NotImplementedError
+
     def tweak_bulk_options(self, options):
         """Receives the bulk options every time a sync happens, so they can be
         tweaked if needed.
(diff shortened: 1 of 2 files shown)


connectors/source.py
```diff
@@ -398,6 +400,11 @@ def hash_id(cls, _id):
 
         return hash_id(_id)
 
+    @classmethod
```
timgrein, line 403: Will incremental syncs have a feature flag? Then we could guard the usage behind the feature flag removing the need for having a separate flag.

wangch079: Yes, there is one, see https://github.com/elastic/enterprise-search-team/issues/4629

The feature flag is at the connector level, and this is at the data source level (e.g. MySQL).

In 8.9, all SPO connectors will have the feature flag enabled, and SPO data source will have `support_incremental_sync` to return `True`

connectors/source.py
```diff
@@ -549,6 +584,10 @@ def _serialize(value):
 
         return doc
 
+    def sync_cursor(self):
+        """Returns the sync cursor of the current sync"""
+        return self._sync_cursor
```
artem-shelkovnikov, line 589: What's the expectation on this method? Will the framework peridically check this field and store the cursor? Will it only do it once sync ended successfully/unsuccessfully?

connectors/source.py
```diff
@@ -501,6 +508,34 @@ async def get_file(doit=True, timestamp=None):
         """
         raise NotImplementedError
 
+    async def get_docs_incrementally(self, sync_cursor, filtering=None):
```
artem-shelkovnikov, line 511: Not sure about the method name - maybe just `get_docs_incremental`? `get_docs_from_cursor`?

Also I can see that Full Sync can also benefit from cursor especially when sync was suspended. Do you wanna support it now or later?

connectors/source.py
wangch079: >Will the framework peridically check this field and store the cursor?

Not for now. Currently, we do store the `sync_cursor` in sync job when claiming it:
https://github.com/elastic/connectors-python/blob/1f4447bf95312f8aa9a39fd2c01ae73e570398eb/connectors/protocol/connectors.py#L273

But that is mainly for record purposes, so we know an incremental sync job is executed with this cursor.

This is needed when later we want to be able to `resume` a suspended sync job.

>What's the expectation on this method?
Will it only do it once sync ended successfully/unsuccessfully?

So the expectation for now is to return the cursor at the end of the sync (i.e. at the end of `get_docs`).

The sync_cursor will only be written back to `connector` after a successful sync. See discussion here: https://github.com/elastic/connectors-python/pull/963#discussion_r1209077066

connectors/source.py
wangch079: > Not sure about the method name - maybe just `get_docs_incremental`? `get_docs_from_cursor`?

I'm open to renaming it. WDYT `get_delta_docs`?

> Also I can see that Full Sync can also benefit from cursor especially when sync was suspended. Do you wanna support it now or later?

Not for now, see this comment: https://github.com/elastic/connectors-python/pull/986#discussion_r1211961681.

artem-shelkovnikov: I think `get_docs_incrementally` is a bit better, I guess let's keep it like that

connectors/source.py
artem-shelkovnikov: What do you think if we allow connector to set cursor instead, e.g:

```python
def get_docs(self, filtering):
    for page in self.client.get_remote_docs_page():
        for doc in page['docs']:
            yield emit_document(doc), None
       yield emit_cursor(page['cursor']
```

One important thing to keep in mind: some 3rd-party systems have cursors that expire. For example, binlog is limited in size and eventually moves on, so if you're too late, you can't do incremental any more. Same for Sharepoint Online - if your delta cursor is too old, it cannot be used any more.

So it's important even for the failed/stuck sync to move the delta on when possible. I haven't thought too deep into it, but I think this is a super important thing to address.

connectors/source.py
```diff
@@ -398,6 +400,11 @@ def hash_id(cls, _id):
 
         return hash_id(_id)
 
+    @classmethod
+    def support_incremental_sync(cls):
+        """Indicates whether the data source supports incremental sync"""
+        return False
```
tarekziade, line 406: why a classmethod? the `BaseDataSource` class already has two class variables, so it would probably be simpler to have it there too.

class methods are useful if you have to execute some code, here it's just a boolean

wangch079: Makes sense. Will update

connectors/source.py
wangch079: >One important thing to keep in mind: some 3rd-party systems have cursors that expire. For example, binlog is limited in size and eventually moves on, so if you're too late, you can't do incremental any more. Same for Sharepoint Online - if your delta cursor is too old, it cannot be used any more.

Not sure about binlog, but the `deltaLink` for SPO is valid for 7 days. This was considered when designing incremental sync, and the conclusion is to fail the sync job if the cursor expires.

>I haven't thought too deep into it, but I think this is a super important thing to address.

Same, as I mentioned in https://github.com/elastic/connectors-python/pull/963#discussion_r1209077066, I chose the safest and easiest approach. 

>So it's important even for the failed/stuck sync to move the delta on when possible

This should work 99% of the time. The other 1% is due to the async data flow
https://github.com/elastic/connectors-python/blob/15afd8d63000d21e5f809f2af3b492928ae0a77e/connectors/es/sink.py#L14-L17

Data being extracted from a remote source doesn't mean it's ingested into ES. With the original implicit incremental sync design (i.e. run the first sync job as full sync, and incremental sync for all the subsequent sync jobs), we couldn't afford any inconsistency. 

With the current design, I think it's an acceptable edge case. WDYT? cc @tarekziade @vidok

artem-shelkovnikov: Thanks for the explanations, I see that the issue is on your radar and being taken care of. I think it's good to move ahead with the taken approach

artem-shelkovnikov (approved):

PR #988, opened 2023-05-31
Update DEVELOPING.md

artem-shelkovnikov:
Small addition to show that Drive Items are recursive and List Items can point to Drive Items so that the diagram is a bit more precise.


timgrein (approved):
LGTM

PR #995, opened 2023-06-01
Allow to create job with job type

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4627

Allows to create a sync job with job type, defaulted to `full`

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -869,7 +869,7 @@ class SyncJobIndex(ESIndex):
             doc_source=doc_source,
         )
 
-    async def create(self, connector, trigger_method, job_type=JobType.FULL):
+    async def create(self, connector, trigger_method, job_type):
         filtering = connector.filtering.get_active_filter().transform_filtering()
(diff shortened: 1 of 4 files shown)


connectors/protocol/connectors.py
```diff
@@ -858,7 +858,7 @@ def _create_object(self, doc_source):
             doc_source=doc_source,
         )
 
-    async def create(self, connector, trigger_method):
+    async def create(self, connector, trigger_method, job_type=JobType.FULL):
```
timgrein, line 861: Should we've a default value here? Could imagine that the caller code is more explicit, if we always need to specify the `job_type`.

timgrein (approved):
Only one small suggestion/question, but that's only a personal preference so LGTM!

connectors/protocol/connectors.py
wangch079: Make sense. Will update

timgrein (approved):

PR #1000, opened 2023-06-01
[DLS] Access control sync scheduling

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4690

This PR adds scheduling for access control sync jobs. Access control sync jobs can only be scheduled, if the corresponding DLS feature flag is enabled. It uses the same concurrency control mechanisms as other job types.

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -426,6 +426,11 @@ class Features:
 
         self.features = features
 
+    def incremental_sync_enabled(self):
+        return self._nested_feature_enabled(
+            ["incremental_sync", "enabled"], default=False
+        )
+
     def sync_rules_enabled(self):
         return any(
             [
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -507,8 +507,8 @@ class Connector(ESDocument):
         return self.get("scheduling", default={})
 
     @property
-    def permissions_scheduling(self):
-        return self.scheduling.get("permissions", {})
+    def access_control_scheduling(self):
+        return self.scheduling.get("access_control", {})
 
     @property
     def configuration(self):
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -538,6 +538,10 @@ class Connector(ESDocument):
     def last_sync_status(self):
         return JobStatus(self.get("last_sync_status"))
 
+    @property
+    def last_access_control_sync_status(self):
+        return JobStatus(self.get("last_access_control_sync_status"))
+
     def _property_as_datetime(self, key):
         value = self.get(key)
         if value is not None:
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -915,7 +915,6 @@ class SyncJobIndex(ESIndex):
             "deleted_document_count": 0,
             "created_at": iso_utc(),
             "last_seen": iso_utc(),
-            "job_type": job_type.value,
         }
         await self.index(job_def)
 
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -432,6 +432,11 @@ class Features:
             ["incremental_sync", "enabled"], default=False
         )
 
+    def document_level_security_enabled(self):
+        return self._nested_feature_enabled(
+            ["document_level_security", "enabled"], default=False
+        )
+
     def sync_rules_enabled(self):
         return any(
             [
@@ -456,10 +461,6 @@ class Features:
                 return self.features.get("filtering_rules", False)
             case Features.ADVANCED_RULES_OLD:
                 return self.features.get("filtering_advanced_config", False)
-            case Features.DOCUMENT_LEVEL_SECURITY:
-                return self._nested_feature_enabled(
-                    ["document_level_security", "enabled"], default=False
-                )
             case _:
                 return False
 
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -46,6 +46,7 @@ __all__ = [
     "DataSourceError",
     "JobStatus",
     "Pipeline",
+    "JobType",
     "JobTriggerMethod",
     "ServiceTypeNotConfiguredError",
     "ServiceTypeNotSupportedError",
@@ -511,7 +512,7 @@ class Connector(ESDocument):
         return self.get("scheduling", "incremental", default={})
 
     @property
-    def access_control_scheduling(self):
+    def access_control_sync_scheduling(self):
         return self.get("scheduling", "access_control", default={})
 
     @property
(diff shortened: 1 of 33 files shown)


connectors/protocol/connectors.py
```diff
@@ -869,7 +878,7 @@ def _create_object(self, doc_source):
             doc_source=doc_source,
         )
 
-    async def create(self, connector, trigger_method):
+    async def create(self, connector, trigger_method, job_type):
```
wangch079, line 881: This was done in PR #995

connectors/protocol/connectors.py
```diff
@@ -511,7 +512,7 @@ def incremental_sync_scheduling(self):
         return self.get("scheduling", "incremental", default={})
 
     @property
-    def access_control_scheduling(self):
+    def access_control_sync_scheduling(self):
```
wangch079, line 515: This can be a separate PR. But I'm OK to review it in this PR.

connectors/services/job_scheduling.py
```diff
@@ -28,6 +31,44 @@
 from connectors.source import get_source_klass
 
 
+class LastSyncTimeMethods(Enum):
```
wangch079, line 34: Instead of introducing a new class. WDYT if we introduce 2 new methods in `Connector` class?
```python
def last_sync_scheduled_at_by_job_type(job_type):
    match job_type:
        case JobType.ACCESS_CONTROL:
            return self.last_access_control_sync_scheduled_at
        case JobType.INCREMENTAL:
            return self.last_incremental_sync_scheduled_at
        case JobType.FULL:
            return self.last_full_sync_scheduled_at
        case _:
            raise ValueError(f"Unknown job type: {job_type}")

async def update_last_sync_scheduled_at_by_job_type(job_type, new_ts):
    match job_type:
        case JobType.ACCESS_CONTROL:
            await self._update_datetime("last_access_control_sync_scheduled_at", new_ts)
        case JobType.INCREMENTAL:
            await self._update_datetime("last_incremental_sync_scheduled_at", new_ts)
        case JobType.FULL:
            await self._update_datetime("last_sync_scheduled_at", new_ts)
        case _:
            raise ValueError(f"Unknown job type: {job_type}")
```

wangch079 (commented):
For access control sync, do you need to check the feature flag? If it's not enabled, we should skip directly.

connectors/services/job_scheduling.py
timgrein: Makes sense! Increases the cohesion and removes the "magic" lookups. Will adjust it 👍

connectors/protocol/connectors.py
timgrein: Will fetch your changes as soon as they're on main 👍

tests/protocol/test_connectors.py
```diff
@@ -543,24 +543,44 @@ async def test_sync_done(job, expected_doc_source_update):
...
 @patch("connectors.protocol.connectors.next_run")
-async def test_connector_next_sync(next_run, scheduling_enabled, expected_next_sync):
+async def test_connector_next_sync(next_run, scheduling_enabled, expected_next_sync, job_type):
     connector_doc = {
         "_id": "1",
         "_source": {
-            "scheduling": {"full": {"enabled": scheduling_enabled, "interval": "1 * * * * *"}},
+            "scheduling": {
+                "enabled": scheduling_enabled,
+                "interval": "1 * * * * *",
```
wangch079, line 563: related to this comment: https://github.com/elastic/enterprise-search-team/issues/4589#issuecomment-1567909683

Why we want to keep the outer `enabled` and `interval`?

wangch079:
Hi @timgrein , I think you missed this comment: https://github.com/elastic/connectors-python/pull/1000#pullrequestreview-1456737985

tests/protocol/test_connectors.py
timgrein: We don't need it here, but it'll be kept in the mapping to avoid reindex operations on a migration. See the corresponding discussion inside the implementation doc, but I can remove it from this test 👍

timgrein:
> Hi @timgrein , I think you missed this comment: [#1000 (review)](https://github.com/elastic/connectors-python/pull/1000#pullrequestreview-1456737985)

I've merged main into this branch, so the changes are present, but not in the diff (see [job_type param in create method](https://github.com/elastic/connectors-python/pull/1000/files#diff-c2f73fe85d1652beb4af8939b73360a399a3496c1863e34b900e12d694420571R898))

timgrein:
> Hi @timgrein , I think you missed this comment: [#1000 (review)](https://github.com/elastic/connectors-python/pull/1000#pullrequestreview-1456737985)

Wanted to address it in a separate PR originally, but I've added it now to this PR as it's related 👍

tests/protocol/test_connectors.py
wangch079: I think they are removed from both the [mapping](https://github.com/elastic/ent-search/blob/63ce9682bc92b9d57712e7e267550042f4cabca8/shared_togo/lib/shared_togo/crawler2/connector.rb#L156-L177) and the [migration](https://github.com/elastic/ent-search/blob/63ce9682bc92b9d57712e7e267550042f4cabca8/shared_togo/db/migrate/20230526133312_change_scheduling_and_add_job_type_and_sync_cursor_to_connectors.rb#L61-L69)

connectors/services/job_scheduling.py
```diff
@@ -171,48 +173,67 @@ async def _should_schedule_on_demand_sync():
...
                 await connector.reload()
             except DocumentNotFoundError:
                 logger.error(f"Couldn't reload connector {connector.id}")
                 return False
 
-            now = datetime.utcnow()
             if (
-                connector.last_sync_scheduled_at is not None
-                and connector.last_sync_scheduled_at > now
+                job_type == JobType.ACCESS_CONTROL
```
wangch079, line 184: For incremental sync, I will check both the feature flag and this value:
https://github.com/elastic/connectors-python/blob/e4ac7dc0f6b4b674eefc1787868524c48a8085af/connectors/source.py#L342

Not sure whether this applies to DLS.

What do you think we move this check to line 99:
https://github.com/elastic/connectors-python/blob/e4ac7dc0f6b4b674eefc1787868524c48a8085af/connectors/services/job_scheduling.py#L94-L101

This way, we can 1) access `source_klass. support_incremental_sync` directly, 2) avoid `if...else...` in `_should_schedule_scheduled_sync`.

timgrein: Makes sense, adjusted it and did some other simplifications 👍 

> Not sure whether this applies to DLS.

No, this doesn't apply to DLS. We can still discuss it, but this should be a separate change from this PR.

connectors/services/job_scheduling.py
```diff
@@ -171,7 +173,7 @@ async def _should_schedule_on_demand_sync():
                 job_type=JobType.FULL,
             )
 
-    async def _scheduled_sync(self, connector):
+    async def _scheduled_sync(self, connector, job_type=None):
```
wangch079, line 176: I think we can make `job_type` mandatory.

wangch079 (approved):
Nice work

PR #1001, opened 2023-06-02
[DLS] At most one access control sync per connector guard

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4591

This PR adds a guard to check whether a connector has an access control sync already running, so we don't schedule two access control syncs at the same time for one connector.

The assertion will be adapted to use the separate access control syncs task pool in a separate PR addressing the task pool separation.

#### Pre-Review Checklist

### DIFF
--- a/connectors/services/job_execution.py
+++ b/connectors/services/job_execution.py
@@ -57,6 +57,15 @@ class JobExecutionService(BaseService):
                 )
                 return
 
+        if (
+            sync_job.job_type == JobType.ACCESS_CONTROL
+            and connector.last_access_control_sync_status == JobStatus.IN_PROGRESS
+        ):
+            logger.debug(
+                f"Connector {connector.id} is still syncing ('{sync_job.job_type.value}'), skip the job {sync_job.id}..."
+            )
+            return
+
         sync_job_runner = SyncJobRunner(
             source_klass=source_klass,
             sync_job=sync_job,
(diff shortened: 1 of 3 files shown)


connectors/services/job_execution.py
```diff
@@ -57,6 +57,15 @@ async def _sync(self, sync_job):
                 )
                 return
 
+        if (
+            sync_job.job_type == JobType.ACCESS_CONTROL
+            and connector.last_access_control_sync_status == JobStatus.IN_PROGRESS
+        ):
+            logger.debug(
+                f"Connector {connector.id} is still syncing ('{sync_job.job_type.value}'), skip the job {sync_job.id}..."
```
wangch079, line 65: Probably consider reformatting both log messages:

```
Connector {connector.id} is still syncing content, skip the job {sync_job.id}...
Connector {connector.id} is still syncing user identities, skip the job {sync_job.id}...
```

tests/services/test_job_execution.py
```diff
@@ -106,19 +106,23 @@ def sync_job_runner_mock():
         yield sync_job_runner_mock
 
 
-def mock_connector(last_sync_status=JobStatus.COMPLETED):
+def mock_connector(job_type=JobType.FULL, last_sync_status=JobStatus.COMPLETED):
```
wangch079, line 109: Not a big issue, but I'd rather introduce an additional param `last_access_control_sync_status` to avoid the `if...else...`

wangch079 (approved):

PR #1003, opened 2023-06-04
Make Extractor support incremental sync

wangch079:
#### Pre-Review Checklist


connectors/es/sink.py
```diff
@@ -448,6 +455,76 @@ async def get_docs(self, generator):
 
         await self.queue.put("END_DOCS")
 
+    async def get_docs_incrementally(self, generator):
```
wangch079, line 458: Want to make it isolated from `get_docs` (full sync).

Once we use Delete by query API, the `get_docs` method can be deleted, and this method will be renamed to `get_docs`

connectors/sync_job_runner.py
```diff
@@ -238,11 +240,20 @@ async def sync_starts(self):
...
+                sync_cursor=self.connector.sync_cursor,
+                filtering=self.sync_job.filtering,
+            )
+        else:
+            raise UnsupportedJobType
+
+        async for doc, lazy_download, *operation in generator:
             doc_id = str(doc.get("_id", ""))
             doc_id_size = len(doc_id.encode(UTF_8))
+            operation = operation[0] if operation else "index"
```
timgrein, line 256: I would suggest we also use constants for these operations like `INDEX = index` or `OP_INDEX = index` here, so renaming/refactoring will be easier + spelling mistakes are less likely. I guess we would need to move those constants then to avoid circular imports.

connectors/es/sink.py
```diff
@@ -326,9 +331,14 @@ async def _deferred_index(self, lazy_download, doc_id, doc, operation):
             }
         )
 
-    async def run(self, generator):
+    async def run(self, generator, job_type):
         try:
-            await self.get_docs(generator)
+            if job_type == JobType.FULL:
```
timgrein, line 336: Should we use `match/case` here like we did in the `connectors` module for updating and accessing `last_sync_time` for consistency? Ofc nothing blocking.

connectors/es/sink.py
vidok: Can you please create an issue to make sure we won't forget about this?

wangch079: Added: https://github.com/elastic/enterprise-search-team/issues/4712

connectors/es/sink.py
```diff
@@ -576,6 +654,7 @@ async def async_bulk(
         sync_rules_enabled=False,
         content_extraction_enabled=True,
         options=None,
+        job_type=JobType.FULL,
```
vidok, line 657: Would it be more secure if we leave `job_type` empty and then validate so anyone who calls `async_bulk` explicitly knows what they are doing?

connectors/sync_job_runner.py
```diff
@@ -238,11 +240,23 @@ async def sync_starts(self):
     async def prepare_docs(self):
         logger.debug(f"Using pipeline {self.sync_job.pipeline}")
 
-        async for doc, lazy_download in self.data_provider.get_docs(
-            filtering=self.sync_job.filtering
-        ):
+        match self.sync_job.job_type:
```
vidok, line 243: I think it would be good for the code readability if move this block of code to another method like `generator` or `create_generator(job_type)`

wangch079: Done: https://github.com/elastic/connectors-python/pull/1003/commits/d242ed9fb0263de30fab2386949ba405581912cc

vidok (approved):

PR #1005, opened 2023-06-05
[Dropbox] Implement Indexing Feature

### LINKED ISSUE #1004: Dropbox: Indexing Feature
Implement Indexing Feature for Dropbox connector

moxarth-elastic:
## Closes #1004 

Added Document Sync Functionality for Dropbox connector.

#### Pre-Review Checklist

Note: The current implementation for document sync is slow in terms of performance. So, we're trying to use low level APIs instead of SDK. However, the approach will still remain the same.

### DIFF
--- a/connectors/sources/dropbox.py
+++ b/connectors/sources/dropbox.py
@@ -5,22 +5,38 @@
 #
 """Dropbox source module responsible to fetch documents from Dropbox online.
 """
-from functools import cached_property
+import asyncio
+import os
+from functools import cached_property, partial
 
+import aiofiles
 import dropbox
 import requests
+from aiofiles.os import remove
+from aiofiles.tempfile import NamedTemporaryFile
 from dropbox.exceptions import ApiError, AuthError, BadInputError
+from dropbox.files import FileMetadata
+from dropbox.sharing import FileLinkMetadata, SharedFileMetadata
 
 from connectors.logger import logger
 from connectors.source import BaseDataSource, ConfigurableFieldValueError
-from connectors.utils import CancellableSleeps, iso_utc
+from connectors.utils import (
+    TIKA_SUPPORTED_FILETYPES,
+    CancellableSleeps,
+    convert_to_b64,
+    iso_utc,
+)
 
 RETRY_COUNT = 3
+FILE_SIZE_LIMIT = 10485760  # Size in Bytes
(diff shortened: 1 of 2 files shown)


connectors/sources/dropbox.py
```diff
@@ -5,22 +5,38 @@
...
-from connectors.utils import CancellableSleeps, iso_utc
+from connectors.utils import (
+    TIKA_SUPPORTED_FILETYPES,
+    CancellableSleeps,
+    convert_to_b64,
+    iso_utc,
+)
 
 RETRY_COUNT = 3
+FILE_SIZE_LIMIT = 10485760  # Size in Bytes
```
vidok, line 31: How did you come up with 10MB limit?

connectors/sources/dropbox.py
```diff
@@ -164,8 +250,130 @@ async def ping(self):
...
+            return
+
+        logger.debug(f"Downloading {attachment_name}")
+
+        document = {
+            "_id": attachment.id,
+            "_timestamp": attachment.server_modified,
+        }
+        temp_filename = ""
+        async with NamedTemporaryFile(mode="wb", delete=False) as async_buffer:
```
vidok, line 300: `get_content` is doing too much from my point of view. I think this block of code can be defined in its own function.

connectors/sources/dropbox.py
```diff
@@ -164,8 +250,130 @@ async def ping(self):
...
+                _, data = self.dropbox_client.download_paper_files(  # pyright: ignore
+                    path=attachment.path_display
+                )
+                await async_buffer.write(data.content)
+            else:
+                logger.info(f"Skipping the download of a file: {attachment_name}")
+
+            temp_filename = str(async_buffer.name)
+
+        logger.debug(f"Calling convert_to_b64 for file : {attachment_name}")
```
vidok, line 318: base64 encoding can be also moved away from this function.

connectors/sources/dropbox.py
moxarth-elastic: The ElasticSearch is having the default value for indexing a file of 10 MB and so we've kept this as a constant.

connectors/sources/dropbox.py
moxarth-elastic: Yes, this block of code can be moved to another function.

vidok (approved):

Issue #939, opened 2023-05-23
[Sharepoint] mapper [size] cannot be changed from type [long] to [text]


llermaly:
## Bug Description

The first time I run a crawl I see the following error: 

[FMWK][10:49:49][DEBUG] Task 1 - Sending a batch of 202 ops -- 7.12MiB
[FMWK][10:49:50][WARNING] Retry count: 2 out of 3. Exception: 400, message='Bad Request', url=URL('https://tenant.sharepoint.com/sites/site/_api/web/lists/getbytitle('Site%20Pages')/items?$select=Title,CanvasContent1,FileLeafRef&$filter=FileLeafRef+eq+'Some-File--Name'.aspx'')
[FMWK][10:49:50][DEBUG] Will retry in 8 seconds
[FMWK][10:49:53][ERROR] operation index failed, {'type': 'illegal_argument_exception', 'reason': 'mapper [size] cannot be changed from type [long] to [text]'}
[FMWK][10:49:53][ERROR] Exception found for task Task-290: mapper [size] cannot be changed from type [long] to [text]

In the next crawls I only see the bad request one.

### To Reproduce
1. Create a sharepoint site with lists/listitems 
2. Create a sharepoint crawl
3. Run the crawler from the local machine

## Expected behavior
We should not see mapping errors.

## Screenshots
N/A

## Environment
Elastic Cloud 8.8 (latest)
Sharepoint Python Connector main branch

## Additional context
While I write this I notice connector is 8.9 and Elastic Cloud is 8.8 so maybe that's related?

Also, potentially related to this one: 

https://github.com/elastic/connectors-python/issues/940

Thanks

danajuratoni:
split from the old issue cc: @akanshi-elastic

prashant-elastic:
@llermaly Just wanted to understand what do you mean by sharepoint crawl here? Because, we are not using any web crawler to run the connector. Instead we create an index with the help of "Use a connector" and run the connector to index the documents. We tried executing the connector using "Use a connector" option for sharepoint online and were not able to reproduce this issue at our end. Can you please let us know if we are missing anything here?
cc @danajuratoni

akanshi-elastic:
Started working on the issue on 2nd june and raised a [PR](https://github.com/elastic/connectors-python/pull/1006) on github on 5th june

PR #1006, opened 2023-06-05
Fix mapper issue: mapper [size] cannot be changed from type [long] to [text]

akanshi-elastic:
## Closes https://github.com/elastic/connectors-python/issues/939###
This PR includes:
1. Converting all elastic field values to string type
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint.py
+++ b/connectors/sources/sharepoint.py
@@ -84,7 +84,6 @@ SCHEMA = {
     },
     LIST_ITEM: {
         "title": "Title",
-        "author_id": "EditorId",
         "creation_time": "Created",
         "_timestamp": "Modified",
     },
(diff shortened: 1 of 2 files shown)


seanstory (changes requested):
This is going the wrong way. Numeric values _should_ be indexed as numerics, not as strings. If we index as `"0"` instead of `0`, you can't do mathematical aggregations at query time, can't do range queries, and can't leverage runtime fields.

Instead, this PR should be ensuring that we convert the retrieved `size` to a number, or treat it as an error if we get a non-numeric size (like `I0` instead of `10`).

akanshi-elastic:
> This is going the wrong way. Numeric values _should_ be indexed as numerics, not as strings. If we index as `"0"` instead of `0`, you can't do mathematical aggregations at query time, can't do range queries, and can't leverage runtime fields.
> 
> Instead, this PR should be ensuring that we convert the retrieved `size` to a number, or treat it as an error if we get a non-numeric size (like `I0` instead of `10`).

@seanstory We change default size from `0` to `"0"` because we are also getting the Length in string type from API Response. otherwise both type will differ from each other. We kept the type same because it is not advisable to manipulate the API response, But if you want we can manipulate API response length type from str to int. Let me know your thoughts on this?
Screenshot 2023-06-06 144143

seanstory:
> if you want we can manipulate API response length type from str to int

Yes, this is what we should do.

akanshi-elastic:
Started working on the issue on 2nd june  and raised a PR on github on 5th june

PR #1009, opened 2023-06-05
[DLS] Add last access control sync information to sync_starts and sync_done

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/4695

This PR adds the last access control sync information to the `sync_starts` and `sync_done` methods. I didn't introduce separate methods to be consistent with other methods inside the `Connector` class, where we just pass the `job_type` and use control flow inside the method to check what to update or fetch (for `sync_done` we use the `job_type` property of the `job` itself).

#### Pre-Review Checklist


connectors/protocol/connectors.py
```diff
@@ -638,13 +638,24 @@ async def update_last_sync_scheduled_at_by_job_type(self, job_type, new_ts):
...
                 raise ValueError(f"Unknown job type: {job_type}")
 
-    async def sync_starts(self):
+    async def sync_starts(self, job_type=JobType.FULL):
+        last_sync_information = (
+            {
+                "last_access_control_sync_status": JobStatus.IN_PROGRESS.value,
+                "last_access_control_sync_error": None,
+            }
+            if job_type == JobType.ACCESS_CONTROL
```
seanstory, line 647: Nit: This type of structure ("if condition, else default") is going to be more bug prone than one that explicitly disallows unknown states ("if condition_a, else if condition_b, else error"). If we add a 3rd sync type, and someone doesn't notice this code, things will keep working, but with unexpected behavior. I prefer to ensure that we fail-fast if our assumptions ever stop holding.

connectors/protocol/connectors.py
```diff
@@ -662,22 +673,38 @@ async def error(self, error):
...
+        )
+
         doc = {
-            "last_sync_status": job_status.value,
             "last_synced": iso_utc(),
-            "last_sync_error": job_error,
             "status": connector_status.value,
             "error": job_error,
-        }
+        } | last_sync_information
```
seanstory, line 700: TIL you can merge two dicts like this. 🎉

connectors/protocol/connectors.py
```diff
@@ -662,22 +673,38 @@ async def error(self, error):
...
 
         # only update sync cursor after a successful sync job
-        if job_status == JobStatus.COMPLETED:
+        if (
+            job_type is not None
```
seanstory, line 704: `is not None` is a double negative. I'd prefer `job_type in [JobType.FULL, JobType.INCREMENTAL]` to make the logic a little more explicit and clear.

connectors/protocol/connectors.py
```diff
@@ -662,22 +673,38 @@ async def error(self, error):
     async def sync_done(self, job, cursor=None):
         job_status = JobStatus.ERROR if job is None else job.status
         job_error = JOB_NOT_FOUND_ERROR if job is None else job.error
+        job_type = job.job_type if job is not None else None
```
wangch079, line 676: This is an edge case, which I don't know how to handle

E.g. An access control sync runs, and somehow the sync job is deleted. `sync_done` is called with the `None` job. In this case, `last_access_control_sync_status` will not be updated, and it will block other access control sync to be executed.

We could choose to reset both `last_access_control_sync_status` and `last_sync_status`. The worst case is that, there is another full/incremental sync running, and this can possibly execute another full/incremental sync. But I think it's still better the the previous mention case.

connectors/protocol/connectors.py
```diff
@@ -662,22 +673,38 @@ async def error(self, error):
...
-        }
+        } | last_sync_information
 
         # only update sync cursor after a successful sync job
-        if job_status == JobStatus.COMPLETED:
+        if (
+            job_type is not None
+            and job_type is not JobType.ACCESS_CONTROL
+            and job_status == JobStatus.COMPLETED
+        ):
```
wangch079, line 707:
```suggestion
        if (
            job_type != JobType.ACCESS_CONTROL
            and job_status == JobStatus.COMPLETED
        ):
```

or 

```suggestion
        if (
            job_type in (JobType.FULL, JobType.INCREMENTAL)
            and job_status == JobStatus.COMPLETED
        ):
```

connectors/protocol/connectors.py
timgrein: Makes sense, adjusted it 👍

connectors/protocol/connectors.py
timgrein: I think this is a good trade-off, I would also prefer to rerun syncs in the worst case vs. blocking access control syncs 👍

wangch079 (approved):

Issue #1014, opened 2023-06-05
mock server does not support aarch64


tarekziade:
The `aarch64` build https://buildkite.com/elastic/connectors-python-nightly-aarch64/builds/34#01888ca6-e77c-40d3-a049-7122b9dcc2d6 is broken for GCS because the docker for the mock server does not support aarch64. 

https://github.com/mock-server/mockserver/pull/1665

The last release does support arm64 though

PR #1024, opened 2023-06-06
skipping on aarch64

tarekziade:
skips a couple of e2e tests on aarch64


tarekziade:
related https://github.com/elastic/connectors-python/issues/1014

seanstory (approved):
:rubberstamp:

PR #1025, opened 2023-06-07
Add tooltip for fields overridden by Advanced Sync Rules

praveen-elastic:
## Closes https://github.com/elastic/connectors-py/issues/1015, #1016  

The tooltip on the configurable fields informs the user that the field will be ignored when Advanced Sync Rules are used.

#### Pre-Review Checklist


connectors/sources/confluence.py
```diff
@@ -262,6 +262,7 @@ def get_default_configuration(cls):
             "spaces": {
                 "display": "textarea",
                 "label": "Confluence space keys",
+                "tooltip": "This configurable field is ignored when Advanced Sync Rules are used.",
```
navarone-feekery, line 265: Could you keep these alphabetised? `tooltip` should go after `order` here.

connectors/sources/jira.py
```diff
@@ -269,6 +269,7 @@ def get_default_configuration(cls):
             "projects": {
                 "display": "textarea",
                 "label": "Jira project keys",
+                "tooltip": "This configurable field is ignored when Advanced Sync Rules are used.",
```
navarone-feekery, line 272: Location here is fine but I notice there's no `order` for `projects`. Could you also add one while you're doing this PR?

navarone-feekery (changes requested):

connectors/sources/jira.py
praveen-elastic: Updated with commit id: [c44df23](https://github.com/elastic/connectors-python/pull/1025/commits/c44df23c9c71ee41d671df666506142700f46f96)

connectors/sources/confluence.py
praveen-elastic: Updated with commit id: [803c757](https://github.com/elastic/connectors-python/pull/1025/commits/803c7574c4f2fd2ddce1c94e9867accbbb87edd2)

navarone-feekery (approved):
LGTM! Thanks!

PR #1035, opened 2023-06-08
Add extraction service

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/4574

Adds extraction service as a standalone service outside of connectors.

The extraction service is only initialised and used if the user has set `use_text_extraction_service: true`.
If enabled, it will send files to the [data-extraction-service](https://github.com/elastic/data-extraction-service), which will return the extracted text or any errors.
This content is added directly to the `body` of the document before indexing. No `_attachment` is added in this case.

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/data-extraction-service/pull/11

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -23,7 +23,13 @@ from connectors.filtering.validation import (
 )
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import CacheWithTimeout, convert_to_b64, html_to_text, url_encode
+from connectors.utils import (
+    CacheWithTimeout,
+    ExtractionService,
+    convert_to_b64,
+    html_to_text,
+    url_encode,
+)
 
 if "OVERRIDE_URL" in os.environ:
     logger.warning("x" * 50)
@@ -546,6 +552,7 @@ class SharepointOnlineDataSource(BaseDataSource):
         super().__init__(configuration=configuration)
 
         self._client = None
+        self.extraction_service = ExtractionService()
 
     @property
     def client(self):
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -765,7 +765,8 @@ class SharepointOnlineDataSource(BaseDataSource):
                                         "lastModifiedDateTime"
                                     ]
                                     attachment_download_func = partial(
-                                        self.get_attachment_content, list_item_attachment
+                                        self.get_attachment_content,
+                                        list_item_attachment,
                                     )
                                     yield list_item_attachment, attachment_download_func
 
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -723,7 +723,7 @@ class SharepointOnlineDataSource(BaseDataSource):
                                         drive_item.get("name", "")
                                     )[-1]
                                     download_func = partial(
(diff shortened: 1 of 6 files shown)


connectors/utils.py
```diff
@@ -653,3 +655,101 @@ async def aenumerate(asequence, start=0):
...
+    async def extract_with_file_pointer(self, filename):
+        """Sends a request to tika-server to extract text from a file.
+        Sends the filename as a request parameter, which tika will pick up and extract.
+
+        Returns a parsed response
+        """
+
+        params = {"local_file_path": filename}
+
+        async with aiohttp.ClientSession() as session:
```
tarekziade, line 713: you should define a timeout and catch the timeout error

connectors/utils.py
```diff
@@ -653,3 +655,101 @@ async def aenumerate(asequence, start=0):
...
+                return await self.parse_extraction_resp(filename, response)
+
+    async def extract_with_file_send(self, filename):
+        """Sends a request to tika-server to extract text from a file.
+        Sends the file as body data in the request.
+
+        Returns a parsed response
+        """
+
+        async with aiohttp.ClientSession() as session:
```
tarekziade, line 728: use a single session across all functions

connectors/sources/sharepoint_online.py
```diff
@@ -647,105 +666,128 @@ async def get_docs(self, filtering=None):
...
-                                ]
-                                attachment_download_func = partial(
-                                    self.get_attachment_content, list_item_attachment
+                                if (
+                                    max_data_age
+                                    and modified_date
+                                    < datetime.utcnow()
+                                    - timedelta(seconds=max_data_age)
+                                ):
+                                    logger.warning(
```
tarekziade, line 714: I expect that we will have lots of those. do we want to lower the level to debug?

connectors/utils.py
```diff
@@ -653,3 +657,112 @@ async def aenumerate(asequence, start=0):
             yield i, elem
         finally:
             i += 1
+
+
+class ExtractionService:
```
tarekziade, line 662: should we have that in its own module?

tarekziade (approved):
Awesome job!

I think the `get_docs` function is getting too bug. Maybe you should consider splitting it out is smaller pieces

connectors/sources/sharepoint_online.py
```diff
@@ -647,105 +666,128 @@ async def get_docs(self, filtering=None):
...
+                            drive_item["object_type"] = "drive_item"
+                            drive_item["_timestamp"] = drive_item[
+                                "lastModifiedDateTime"
+                            ]
+
+                            download_func = None
+
+                            if "@microsoft.graph.downloadUrl" in drive_item:
+                                modified_date = datetime.strptime(
+                                    drive_item["lastModifiedDateTime"],
```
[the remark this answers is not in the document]
navarone-feekery: I don't know, that was already included from @artem-shelkovnikov's work. It's appearing here as a change because I had to indent it.

connectors/utils.py
navarone-feekery: I want it to be but we discussed in the last projects sync that we should just put it in utils for now.

connectors/sources/sharepoint_online.py
navarone-feekery: It is very noisy, I will lower it to debug

PR #1036, opened 2023-06-08
Support for serverless

tarekziade:
This PR adds support for a serverless e2e test in the CI (using MongoDB) so we can verify that our code is serverless compatible and compare perf

- [x] make it work against serverless (job creation now sets the id with uuid)
- [x] create a docker-compose environment that runs es-stateless and add a fixture that runs it (mongo)
- [x] make all `_refresh` calls optional with a serverless flag
- [x] fix the 2 failing tests
- [x] add Vault support to log in to docker to see the serverless docker images
- [x] wait for https://github.com/elastic/elasticsearch/pull/96752 to land


tarekziade:
not sure about this one 

```
[FMWK][10:26:15][CRITICAL] {'error': {'root_cause': [{'type': 'query_shard_exception', 'reason': 'No mapping found for [created_at] in order to sort on', 'index_uuid': 'ctFO-a__QNa3h1pyy7Ymiw', 'index': '.elastic-connectors-sync-jobs'}], 'type': 'search_phase_execution_exception', 'reason': 'all shards failed', 'phase': 'query', 'grouped': True, 'failed_shards': [{'shard': 0, 'index': '.elastic-connectors-sync-jobs', 'node': 'qXqH6PVDThyfS7amBE33LQ', 'reason': {'type': 'query_shard_exception', 'reason': 'No mapping found for [created_at] in order to sort on', 'index_uuid': 'ctFO-a__QNa3h1pyy7Ymiw', 'index': '.elastic-connectors-sync-jobs'}}]}, 'status': 400}
  File "/Users/tarek/Dev/connectors-python/connectors/es/index.py", line 103, in get_all_docs
    resp = await self.client.search(
  File "/Users/tarek/Dev/connectors-python/lib/python3.10/site-packages/elasticsearch/_async/client/__init__.py", line 3859, in search
    return await self.perform_request(  # type: ignore[return-value]
  File "/Users/tarek/Dev/connectors-python/lib/python3.10/site-packages/elasticsearch/_async/client/_base.py", line 320, in perform_request
    raise HTTP_EXCEPTIONS.get(meta.status, ApiError)(
elasticsearch.BadRequestError: BadRequestError(400, 'search_phase_execution_exception', 'No mapping found for [created_at] in order to sort on')
```

but it's not a blocker, we can ignore for now

connectors/es/index.py
```diff
@@ -64,8 +62,8 @@ async def fetch_response_by_id(self, doc_id):
 
         return resp.body
 
-    async def index(self, doc):
-        return await self.client.index(index=self.index_name, document=doc)
+    async def index(self, doc, doc_id=None):
+        return await self.client.index(index=self.index_name, id=doc_id, document=doc)
```
artem-shelkovnikov, line 66: What about this change? Why is doc_id added?

tarekziade: because serverless does not support `POST /index/_doc` anymore to create a new doc, we want `POST /index/_doc/<new-id>`

seanstory: In that case should `doc_id` be a required param, and not have a default value of `None`?

tarekziade: it was a mistake from serverless gating, I will revert that once they push https://github.com/elastic/elasticsearch/pull/96752

connectors/es/index.py
```diff
@@ -94,8 +92,6 @@ async def get_all_docs(self, query=None, sort=None, page_size=DEFAULT_PAGE_SIZE)
         Returns:
             Iterator
         """
-        await self.client.indices.refresh(index=self.index_name)
```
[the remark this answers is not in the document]
tarekziade: good suggestion -- keeping refresh sounds useful to keep things a little bit faster. I'll find out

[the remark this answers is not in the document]
tarekziade: We'll have an API at some point (see the slacks PMs), but until it's ready maybe we can just use a configuration flag?

tarekziade:
everything is wired 

https://buildkite.com/elastic/connectors-python-nightly/builds/741#0188a233-9087-4071-a8a6-4b6a32386c88

once we have the API back it should work

```


[FMWK][22:11:29][CRITICAL] BadRequestError(400, 'uri [/.elastic-connectors-sync-jobs/_doc] with method [POST] exists but is not available when running in serverless mode', 'uri [/.elastic-connectors-sync-jobs/_doc] with method [POST] exists but is not available when running in serverless mode')
--


elasticsearch.BadRequestError:
 BadRequestError(400, 'uri [/.elastic-connectors-sync-jobs/_doc] with 
method [POST] exists but is not available when running in serverless 
mode', 'uri [/.elastic-connectors-sync-jobs/_doc] with method [POST] 
exists but is not available when running in serverless mode')
```

artem-shelkovnikov (dismissed):

connectors/es/index.py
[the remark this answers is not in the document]
sphilipse: I'd suggest that maybe we don't actually need to refresh here, and it may be an anti-pattern to refresh as much as we're doing. I'm thinking of the issue we had where our frequent refreshes in the crawler were causing issues with fragmentation in the index. I think we should be evaluating a little more consciously whether a refresh is absolutely necessary, or whether it's fine to wait for the index to automatically refresh instead. In Kibana we've removed all refreshes, except for a single `refresh: 'wait_for'` when creating a connector (to make sure it shows up when redirecting the user).

seanstory (approved):

PR #1038, opened 2023-06-08
Attempt to use non-native default and ftest configs

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/4731

- Removes "native" as the default way to start connectors
- enables ftests to run on their on config files, and with pre-determined connector ids (non-native)
- removes `sync_now` usage from ftests
- cleans up some testing logging

#### Pre-Review Checklist
- [x] Considered corresponding documentation changes (separate issue)

## Release Note

Cloned/forked connectors from connectors-python no longer run in "native" mode. Native mode is available exclusively on Elastic Cloud deployments.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -27,26 +27,8 @@ service:
   job_cleanup_interval: 300
   log_level: INFO
 
-native_service_types:
-  - mongodb
-  - mysql
-  - network_drive
-  - s3
-  - google_cloud_storage
-  - azure_blob_storage
-  - postgresql
-  - oracle
-  - confluence
-  - dir
-  - sharepoint
-  - mssql
-  - jira
-  - sharepoint_online
-
-
-# Connector client settings
-#connector_id: 'changeme'
-#service_type: 'changeme'
+connector_id: 'changeme'
+service_type: 'changeme'
 
 sources:
   mongodb: connectors.sources.mongo:MongoDataSource
--- a/connectors/kibana.py
+++ b/connectors/kibana.py
@@ -297,7 +297,9 @@ async def prepare(service_type, index_name, config, connector_definition=None):
         }
 
         logger.info(f"Prepare {CONNECTORS_INDEX}")
-        await upsert_index(es, CONNECTORS_INDEX, docs=[doc], doc_ids=[config["connector_id"]])
+        await upsert_index(
+            es, CONNECTORS_INDEX, docs=[doc], doc_ids=[config["connector_id"]]
+        )
 
         logger.info(f"Prepare {JOBS_INDEX}")
         await upsert_index(es, JOBS_INDEX, docs=[], mappings=JOB_INDEX_MAPPINGS)
@@ -315,7 +317,9 @@ async def prepare(service_type, index_name, config, connector_definition=None):
         await es.close()
 
 
-async def upsert_index(es, index, docs=None, doc_ids=None, mappings=None, settings=None):
+async def upsert_index(
+    es, index, docs=None, doc_ids=None, mappings=None, settings=None
+):
     """Override the index with new mappings and settings.
 
     If the index with such name exists, it's deleted and then created again
--- a/connectors/kibana.py
+++ b/connectors/kibana.py
@@ -354,10 +354,8 @@ async def upsert_index(
     # TODO: bulk
 
     if doc_ids is None:
-        doc_id = 1
-        for doc in docs:
+        for doc_id, doc in enumerate(docs):
             await es.client.index(index=index, id=doc_id, document=doc)
-            doc_id += 1
     else:
         for doc, doc_id in zip(docs, doc_ids, strict=True):
             await es.client.index(index=index, id=doc_id, document=doc)
--- a/connectors/kibana.py
+++ b/connectors/kibana.py
@@ -359,7 +359,7 @@ async def upsert_index(
             await es.client.index(index=index, id=doc_id, document=doc)
             doc_id += 1
     else:
-        for doc, doc_id in zip(docs, doc_ids):
+        for doc, doc_id in zip(docs, doc_ids, strict=True):
             await es.client.index(index=index, id=doc_id, document=doc)
 
 
--- a/connectors/kibana.py
+++ b/connectors/kibana.py
@@ -297,7 +297,7 @@ async def prepare(service_type, index_name, config, connector_definition=None):
         }
 
         logger.info(f"Prepare {CONNECTORS_INDEX}")
-        await upsert_index(es, CONNECTORS_INDEX, docs=[doc])
+        await upsert_index(es, CONNECTORS_INDEX, docs=[doc], doc_ids=[config["connector_id"]])
 
         logger.info(f"Prepare {JOBS_INDEX}")
         await upsert_index(es, JOBS_INDEX, docs=[], mappings=JOB_INDEX_MAPPINGS)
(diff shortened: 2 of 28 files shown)


seanstory:
@elastic/ingestion-team this is ready for review

connectors/kibana.py
```diff
@@ -348,10 +352,15 @@ async def upsert_index(es, index, docs=None, mappings=None, settings=None):
...
         return
     # TODO: bulk
-    doc_id = 1
-    for doc in docs:
-        await es.client.index(index=index, id=doc_id, document=doc)
-        doc_id += 1
+
+    if doc_ids is None:
+        doc_id = 1
+        for doc in docs:
```
tarekziade, line 358: you can use enumerate here instead of a custom variable:

```
for doc_id, doc in enumerate(docs):
    ...
```

tarekziade:
Nice change for the config file. 

I wonder if we could share the same config file base for all the e2e tests to minimize tech debt when we change the config file. e.g. have all tests use the same `base_config.yml` and the ability to inherit from it in specific tests to overrid some parts

seanstory:
> I wonder if we could share the same config file base for all the e2e tests to minimize tech debt when we change the config file. e.g. have all tests use the same base_config.yml and the ability to inherit from it in specific tests to overrid some parts

I'd had a similar thought. However, I don't like the idea of building that kind of config-merging logic just for tests, and it seemed like scope creep to introduce a multi-config-file capability for this issue. Perhaps it's just a matter of us setting all the current default configs _as defaults_ in the code, in case the YAML isn't explicit? That would let us significantly reduce the number of lines in all these yaml files, including the one we ship for customer use.

tarekziade:
> > I wonder if we could share the same config file base for all the e2e tests to minimize tech debt when we change the config file. e.g. have all tests use the same base_config.yml and the ability to inherit from it in specific tests to overrid some parts
> 
> I'd had a similar thought. However, I don't like the idea of building that kind of config-merging logic just for tests, and it seemed like scope creep to introduce a multi-config-file capability for this issue. 

I am talking about doing this in ftest not in the production code:

- use a single central config for all the tests?
- when ftest runs, it simply merges a local yml that contains connector_service, service_type

seanstory:
Nightlies are green! https://buildkite.com/elastic/connectors-python-nightly/builds/754

connectors/kibana.py
```diff
@@ -348,10 +352,13 @@ async def upsert_index(es, index, docs=None, mappings=None, settings=None):
...
-    for doc in docs:
-        await es.client.index(index=index, id=doc_id, document=doc)
-        doc_id += 1
+
+    if doc_ids is None:
+        for doc_id, doc in enumerate(docs, start=1):
+            await es.client.index(index=index, id=doc_id, document=doc)
+    else:
+        for doc, doc_id in zip(docs, doc_ids, strict=True):
+            await es.client.index(index=index, id=doc_id, document=doc)
```
artem-shelkovnikov, line 361: I remember I've seen this change already - is it because of merge by chance?

tests/sources/fixtures/azure_blob_storage/config.yml
```diff
@@ -0,0 +1,33 @@
...
+service:
+  idling: 30
+  heartbeat: 300
+  max_errors: 20
+  max_errors_span: 600
+  max_concurrent_syncs: 1
+  job_cleanup_interval: 300
+  log_level: INFO
+
+connector_id: 'abs'
```
artem-shelkovnikov, line 29: As far as I could see it can be arbitrary value?

config.yml
```diff
@@ -27,26 +27,8 @@ service:
...
-  - dir
-  - sharepoint
-  - mssql
-  - jira
-  - sharepoint_online
-
-
-# Connector client settings
-#connector_id: 'changeme'
-#service_type: 'changeme'
```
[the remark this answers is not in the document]
seanstory: Yep. Not even just default - we're deprecating running in native mode (except for in Cloud)

requirements/tests.txt
```diff
@@ -10,4 +10,4 @@ pytest-randomly>=3.12.0
 git+https://github.com/elastic/perf8#egg=perf8
 freezegun>=0.3.4
 pytest-fail-slow>=0.3.0
-pyright>=1.1.295
+pyright>=1.1.312
```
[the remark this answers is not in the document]
seanstory: I actually had to do it in this, because my usage of `zip` didn't work with the old version.  https://github.com/microsoft/pyright/issues/5172#issuecomment-1579677316

connectors/kibana.py
seanstory: Nope, this was all me 😅 
Because I switched off Native, that meant that the config.yml needed to specify connector IDs. Which meant that the IDs couldn't be non-deterministic. So I had to modify this logic to pass along the pre-configured connector ID.

tests/sources/fixtures/azure_blob_storage/config.yml
seanstory: Yep. Tried to make it kinda meaningful, if short (because I'm lazy). So "abs" is "Azure Blob Store"

config.yml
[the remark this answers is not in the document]
seanstory: See https://github.com/elastic/enterprise-search-team/issues/4731 and https://github.com/elastic/enterprise-search-team/issues/4732

artem-shelkovnikov (approved):
LGTM! Definitely we can improve some things later, but it's already a good change taking us where we wanna be

wangch079:
With the removal of the usage of `sync_now` flag, does the ftest fully rely on the `scheduling`? But I see 2 different schedules:

https://github.com/elastic/connectors-python/blob/dc651cf405514a9d70ede6bf56db3498647a38a2/connectors/kibana.py#L282-L286

https://github.com/elastic/connectors-python/blob/dc651cf405514a9d70ede6bf56db3498647a38a2/tests/sources/fixtures/sharepoint_online/connector.json#L163-L168

and an invalid one:

https://github.com/elastic/connectors-python/blob/dc651cf405514a9d70ede6bf56db3498647a38a2/tests/sources/fixtures/mongodb_serverless/connector.json#L107

seanstory:
> With the removal of the usage of sync_now flag, does the ftest fully rely on the scheduling?

Yes

> But I see 2 different schedules:

I think that's ok? We just want them to run ASAP.

> and an invalid one:

That's probably less ok. I'll fix and re-run the nightly to make sure that's not broken

tests/sources/fixtures/mongodb_serverless/connector.json
```diff
@@ -1,114 +1,125 @@
...
-                "order": 6,
-                "type": "bool",
-                "value": true
-            }
+  "name": "mongodb",
+  "service_type": "mongodb",
+  "index_name": "search-mongodb",
+  "sync_cursor": null,
+  "sync_now": true,
+  "is_native": true,
```
tarekziade, line 7: should be false here, fixes the nightly failure

seanstory:
> WDYT if we keep this action and make it index a new sync job document?

We could do that, modifying kibana.yml to create a sync job record like actual kibana does. But I'd prefer to not make more changes than necessary for this PR, and to just get these changes in. There's lots of room to add incremental improvements here.

> Run it ASAP also indicates it will be run more frequently, is it OK for e2e test? I think the requirement of the ftest is to run a sync explicitly once.

Given that the ftests are green, I don't think it hurts anything to have them on a schedule? But I'll be honest that I haven't dug deep into what the expectations are.

tarekziade (approved):

PR #1047, opened 2023-06-09
Use DTZ rule and fix code that breaks it

artem-shelkovnikov:
Small problem found: we have no linter verifying that we use timezones in consistent manner.

- Adding "DTZ" rule for Ruff: https://beta.ruff.rs/docs/rules/#flake8-datetimez-dtz
- Fixing code that breaks the rules

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,4 +1,4 @@
-select = ["E", "F", "B"]
+select = ["E", "F", "B", "DTZ"]
 ignore = ["E501"]
 
 # Allow autofix for all enabled rules (when `--fix`) is provided.
--- a/connectors/services/job_scheduling.py
+++ b/connectors/services/job_scheduling.py
@@ -200,6 +200,9 @@ class JobSchedulingService(BaseService):
                 )
                 return False
 
+            print("NEXT is {next_sync}")
+            print("NOW is {now}")
+
             next_sync_due = (next_sync - now).total_seconds()
             if next_sync_due - self.idling > 0:
                 logger.debug(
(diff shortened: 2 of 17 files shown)


seanstory (dismissed):
🥇

artem-shelkovnikov:
@seanstory I had to fix more things 😅 Can I get another stamp on this PR?

.ruff.toml
```diff
@@ -1,4 +1,4 @@
-select = ["E", "F", "B"]
+select = ["E", "F", "B", "DTZ"]
```
tarekziade, line 1: yay

tarekziade (dismissed):

seanstory (dismissed):

connectors/services/job_scheduling.py
```diff
@@ -200,6 +200,9 @@ async def _should_schedule_scheduled_sync(job_type):
                 )
                 return False
 
+            print("NEXT is {next_sync}")
+            print("NOW is {now}")
```
wangch079, line 204: debug log?

connectors/services/job_scheduling.py
```diff
@@ -176,7 +176,7 @@ async def _should_schedule_scheduled_sync(job_type):
                 return False
 
             job_type_value = job_type.value
-            now = datetime.utcnow()
+            now = datetime.now(timezone.utc)
```
wangch079, line 179: Should we move all the datetime related functions to `utils` or a new `datetime` module?

connectors/services/job_scheduling.py
artem-shelkovnikov: Yup, good catch!

connectors/services/job_scheduling.py
artem-shelkovnikov: Eventually yes, should be something like `utils/datetime.py`. I'm not planning it in this PR though, I already have a lot of problems with this change alone.

artem-shelkovnikov:
Closing to re-open later, when implemented properly

PR #1052, opened 2023-06-12
Create Jobs index with mappings for ftests

artem-shelkovnikov:
Something has broken our `ftest`:

```
[FMWK][19:46:37][CRITICAL] {'error': {'root_cause': [{'type': 'query_shard_exception', 'reason': 'No mapping found for [created_at] in order to sort on', 'index_uuid': 'r0GfbA4kQ-eWKDDyowBYQQ', 'index': '.elastic-connectors-sync-jobs'}], 'type': 'search_phase_execution_exception', 'reason': 'all shards failed', 'phase': 'query', 'grouped': True, 'failed_shards': [{'shard': 0, 'index': '.elastic-connectors-sync-jobs', 'node': '0gXtkxgKS3aKvIAfH3y91Q', 'reason': {'type': 'query_shard_exception', 'reason': 'No mapping found for [created_at] in order to sort on', 'index_uuid': 'r0GfbA4kQ-eWKDDyowBYQQ', 'index': '.elastic-connectors-sync-jobs'}}]}, 'status': 400}
  File "/opt/buildkite-agent/builds/bk-agent-elastic-gcp-1686338522238055018/elastic/connectors-python-nightly/connectors/es/index.py", line 107, in get_all_docs
    resp = await self.client.search(
  File "/opt/buildkite-agent/builds/bk-agent-elastic-gcp-1686338522238055018/elastic/connectors-python-nightly/lib/python3.10/site-packages/elasticsearch/_async/client/__init__.py", line 3861, in search
    return await self.perform_request(  # type: ignore[return-value]
  File "/opt/buildkite-agent/builds/bk-agent-elastic-gcp-1686338522238055018/elastic/connectors-python-nightly/lib/python3.10/site-packages/elasticsearch/_async/client/_base.py", line 320, in perform_request
    raise HTTP_EXCEPTIONS.get(meta.status, ApiError)(
elasticsearch.BadRequestError: BadRequestError(400, 'search_phase_execution_exception', 'No mapping found for [created_at] in order to sort on')
```

This error does not affect the execution of ftest, but confuses people who don't run it too often.

Why does this error happen? Because we don't run Kibana in ftest and don't create connectors/job indices with any mappings. Connectors mapping is created automatically when a connector record is inserted into the index, but job record is not created.

Since mapping for jobs is not created, service that is responsible to clean up jobs cannot make searches over job index by the field "created_at" because this field is not there and Elasticsearch is unable to determine _how_ to search over field whose type is unknown, hence the error.

I've decided to just add current job index mapping into `kibana.py` so that index is created with some sensible mapping. Later we expect index to be created and managed by Elasticsearch, but for now I think it's okay to have this hack in-place.


artem-shelkovnikov:
Update: I found out that actually `ftest` is not non-deterministicly broken possibly because of change that detached job scheduling from execution.

So from time to time `ftest` would just fail. This change should fix this problem.

timgrein (approved):
LGTM

PR #1059, opened 2023-06-13
Add incremental sync support to SyncJobRunner

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/4637

This PR makes sure the sync cursor is stored to `SyncJob` when claiming the job, and the connector is updated with the latest `sync_cursor` when the sync job completes.

#### Pre-Review Checklist

### DIFF
--- a/tests/test_sync_job_runner.py
+++ b/tests/test_sync_job_runner.py
@@ -46,6 +46,9 @@ def mock_sync_job(job_type, index_name):
     sync_job.job_type = job_type
     sync_job.pipeline = Pipeline({})
     sync_job.filtering = Filter()
+    sync_job.is_content_sync = Mock(
+        return_value=job_type in (JobType.FULL, JobType.INCREMENTAL)
+    )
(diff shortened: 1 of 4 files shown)


tests/test_sync_job_runner.py
```diff
@@ -58,6 +60,8 @@ def create_runner(
     source_changed=True,
     source_available=True,
     validate_config_exception=None,
+    sync_cursor=SYNC_CURSOR,
+    job_type=JobType.FULL,
```
seanstory, line 64: These two don't go together, right? Shouldn't FULL jobs have a None cursor?

wangch079: They are default values. It means, when the test runs, by default, the sync job is a full sync, and the data provider will return the sync cursor `SYNC_CURSOR` at the end of the sync (sync cursor needs to be updated at the end of both full and incremental sync).

vidok (approved):

PR #1063, opened 2023-06-13
deprecate native mode configurably

seanstory:
## part of https://github.com/elastic/enterprise-search-team/issues/4732

- Adds deprecation warnings when running in Native mode
- Adds a new config, `_force_allow_native`, which bypasses ignores these deprecation warnings.

#### Pre-Review Checklist
- [x] Contributed any configuration settings changes to the configuration reference (separate issue)

## Related Pull Requests

* Corresponding ent-search pr: https://github.com/elastic/ent-search/pull/7558

## Release Note

Running self-managed connectors in "native" mode is deprecated starting in 8.9.0.

### DIFF
--- a/connectors/preflight_check.py
+++ b/connectors/preflight_check.py
@@ -41,7 +41,7 @@ class PreflightCheck:
                 logger.critical(f"{self.elastic_config['host']} seem down. Bye!")
                 return False
 
-            self._check_configuration_validity()
+            self._validate_configuration()
             return await self._check_system_indices_with_retries()
         finally:
             self.stop()
@@ -72,7 +72,7 @@ class PreflightCheck:
                     await self._sleeps.sleep(self.preflight_idle)
         return False
 
-    def _check_configuration_validity(self):
+    def _validate_configuration(self):
         # "Native" mode
         configured_native_types = "native_service_types" in self.config
         force_allowed_native = self.config.get("_force_allow_native", False)
--- a/connectors/preflight_check.py
+++ b/connectors/preflight_check.py
@@ -77,11 +77,20 @@ class PreflightCheck:
         configured_native_types = "native_service_types" in self.config
         force_allowed_native = self.config.get("_force_allow_native", False)
         if configured_native_types and not force_allowed_native:
-            logger.warning("The configuration 'native_service_types' has been deprecated. Please remove this configuration.")
-            logger.warning("Native Connectors are only supported internal to Elastic Cloud deployments, which this process is not.")
+            logger.warning(
+                "The configuration 'native_service_types' has been deprecated. Please remove this configuration."
+            )
+            logger.warning(
+                "Native Connectors are only supported internal to Elastic Cloud deployments, which this process is not."
+            )
 
         # Connector client mode
         configured_connector_id = self.config.get("connector_id", None)
         configred_service_type = self.config.get("service_type", None)
-        if not (configured_connector_id and configred_service_type) and not force_allowed_native:
-            logger.warning("Please update your config.yml to explicitly configure a 'connector_id' and a 'service_type'")
+        if (
+            not (configured_connector_id and configred_service_type)
+            and not force_allowed_native
+        ):
+            logger.warning(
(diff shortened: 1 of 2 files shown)


connectors/preflight_check.py
```diff
@@ -69,3 +71,26 @@ async def _check_system_indices_with_retries(self):
                     attempts += 1
                     await self._sleeps.sleep(self.preflight_idle)
         return False
+
+    def _check_configuration_validity(self):
+        # "Native" mode
+        configured_native_types = "native_service_types" in self.config
+        force_allowed_native = self.config.get("_force_allow_native", False)
```
wangch079, line 78: Does a key starting with `_` indicates `for internal use only`?

Will this be set to `True` for a cloud deployment?

I feel an `EnvVar` may be more appropriate if the answers to the above 2 questions are Yes.

connectors/preflight_check.py
```diff
@@ -69,3 +71,26 @@ async def _check_system_indices_with_retries(self):
...
+            logger.warning(
+                "Native Connectors are only supported internal to Elastic Cloud deployments, which this process is not."
+            )
+
+        # Connector client mode
+        configured_connector_id = self.config.get("connector_id", None)
+        configred_service_type = self.config.get("service_type", None)
+        if (
+            not (configured_connector_id and configred_service_type)
+            and not force_allowed_native
```
wangch079, line 92: Should also check whether there's value in `native_service_types`? `force_allowed_native` doesn't mean it runs in native mode.

connectors/preflight_check.py
```diff
@@ -69,3 +71,26 @@ async def _check_system_indices_with_retries(self):
                     attempts += 1
                     await self._sleeps.sleep(self.preflight_idle)
         return False
+
+    def _check_configuration_validity(self):
```
artem-shelkovnikov, line 75:
```suggestion
    def _validate_configuration(self):
```

connectors/preflight_check.py
seanstory: I'm not sure I see the strong benefit of env vars over yaml configuration. Most (all?) of our configuration is done through YAML - why should we divide it into multiple locations? You're right that we _could_, but I don't see the benefit.

connectors/preflight_check.py
seanstory: I want this printed if a customer is running in native mode on self-managed infrastructure. I want it to not only tell them not to configure that, but to tell them what _to_ configure. Otherwise, they have to get errors twice:
1. configured to run native mode
1. get warning to remove `native_service_types`
1. remove `native_service_types` and restart
1. get warning that you have to configure `service_type` and `connector_id`
1. be frustrated that I didn't tell you that to start with.

I want to avoid that last bit of frustration. I also wanted to cover the case where they have both `native_service_types` AND `connector_id`/`service_type` configured, hence why these logs are in two conditionals, and not just one.

connectors/preflight_check.py
wangch079: This certainly will work. But I just feel that the convention is to use `EnvVar` for some unpublished configuration. E.g. `NAME` and `PERF8` in `make ftest`:

https://github.com/elastic/connectors-python/blob/b45bb62b6e7be96196578189c1c08e81bdb09dfd/Makefile#L62

All fields in YAML should be configurable by users?

wangch079 (dismissed):
Only one nit, but not a blocker

connectors/preflight_check.py
seanstory: Ah I see.
`ftest` being a separate bash tool, it makes sense that it needs to get configuration from outside of just our YAML file. But the connector service, when running in production, should be able to rely on YAML, I think.

> All fields in YAML should be configurable by users?

It, depends. In Enterprise Search we have undocumented YAML configs that we use for Feature Flags. In Elastic Cloud, some of our YAML configs are explicitly not accessible to the user (like encryption keys). I think we're ok to stick with this established pattern.

tarekziade (approved):

PR #1066, opened 2023-06-14
Update Sharepoint Online developer documentation

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/4701

This change moves Sharepoint Online dev documentation to `docs/sources/` directory and updates its content to include more information about Sharepoint Online and the connector itself.

#### Pre-Review Checklist


docs/sources/SHAREPOINT_ONLINE.md
```diff
@@ -0,0 +1,234 @@
...
+        "client_secret": client_secret
+    }
+    headers = {"Content-Type": "application/x-www-form-urlencoded"}
+
+    response = requests.post(url,data=data, headers=headers).json()
+
+    access_token = response["access_token"]
+```
+
+What is also important, that tokens expire, so connector needs to be attentive to which token expired and properly refresh it. Both endpoints used for authentication return a field `expires_in` that specifies how many seconds the token expires in.
```
seanstory, line 45:
```suggestion
What is also important, that tokens expire, so the connector needs to be attentive to which token expired and properly refresh it. Both endpoints used for authentication return a field `expires_in` that specifies how many seconds the token expires in.
```

seanstory (commented):
I'm not sure I understand the use case for this document.

A lot of the content is already in https://github.com/elastic/connectors-python/blob/main/docs/DEVELOPING.md#other-considerations, to provide concrete examples for considerations to make when developing a new connector.

Some of this content could also make sense as a reference document for the data source (explaining how to set up auth and what it'll be used for, what it will sync, projecting storage requirements, etc). 

But most of this document feels like developer notes like what we'd usually have in an Implementation Document or some other discovery artifact. Definitely useful, but maybe not necessary to keep in the repo.

What's your vision for how this will be used and how it will evolve over time?

artem-shelkovnikov:
Closing as seems it's not useful

PR #1073, opened 2023-06-15
[DLS] Prefix target index for access control syncs

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4805

This PR prefixes the target index for sync jobs of type `access_control` with `search-acl-filter-`, which is the prefix for access control indices.

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -920,6 +920,8 @@ class SyncJobIndex(ESIndex):
         elastic_config (dict): Elasticsearch configuration and credentials
     """
 
+    ACCESS_CONTROL_INDEX_PREFIX = "search-acl-filter-"
+
     def __init__(self, elastic_config):
         super().__init__(index_name=JOBS_INDEX, elastic_config=elastic_config)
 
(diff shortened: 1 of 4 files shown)


connectors/protocol/connectors.py
```diff
@@ -920,6 +920,8 @@ class SyncJobIndex(ESIndex):
         elastic_config (dict): Elasticsearch configuration and credentials
     """
 
+    ACCESS_CONTROL_INDEX_PREFIX = "search-acl-filter-"
```
artem-shelkovnikov, line 923: Can probably go to Utils

artem-shelkovnikov (approved):

artem-shelkovnikov (approved):

tests/protocol/test_connectors.py
```diff
@@ -190,6 +190,9 @@
 
 SYNC_CURSOR = {"foo": "bar"}
 
+INDEX_NAME = "index_name"
+ACCESS_CONTROL_INDEX_PREFIX = "search-acl-filter-"
```
wangch079, line 194: Shouldn't this be imported?

timgrein: I normally always patch constants, so I'm not depending on their concrete "real" value anywhere in the test.

PR #1076, opened 2023-06-15
Properly populate ids for site pages and list items

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/4701

Found out that some of IDs that were used by the connector before are actually not unique.

For list items I'm mixing in ID of List object that contains items, for Web Pages I switch to `odata.id` - these seem unique.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -728,7 +728,13 @@ class SharepointOnlineDataSource(BaseDataSource):
                     async for list_item in self.client.site_list_items(
                         site["id"], site_list["id"]
                     ):
-                        list_item["_id"] = list_item["id"]
+                        # List Item IDs are unique within list.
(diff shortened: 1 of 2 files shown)


connectors/sources/sharepoint_online.py
```diff
@@ -761,12 +769,10 @@ async def get_docs(self, filtering=None):
                                 )
                                 yield list_item_attachment, attachment_download_func
 
-                        download_func = None
-
-                        yield list_item, download_func
```
navarone-feekery, line 766: Question: Why is it not necessary to have these anyore?

artem-shelkovnikov: I moved it higher to line 751.

Also there's no `download_func` for it, so I just made it `yield list_item, None`

wangch079 (approved):

PR #1077, opened 2023-06-15
Make functional tests poll more often

artem-shelkovnikov:
Attempting to fix functional tests.

Functional tests cron schedule is "1 * * * * *" - meaning at first second of every minute. Service polls every 30 seconds.

If the service starts exactly in the middle of the interval - let's say couple milliseconds over 17:00:31, then there is a chance it'll never work:

next_run(17:00:31.000001) will be reported as "17:01:01" but the service will wake up at "17:01:01.000001" and it's already too late. Service will keep attempting until at some point the rounding will trigger and the job will work.

For now to not spend too much time diving into date times, I'm proposing to just decrease poll interval for functional tests.

When we'll need to support "every minute" schedule, we'll need to address this problem accordingly.

### DIFF
--- a/tests/sources/fixtures/azure_blob_storage/config.yml
+++ b/tests/sources/fixtures/azure_blob_storage/config.yml
@@ -18,7 +18,7 @@ elasticsearch:
   log_level: info
 
 service:
-  idling: 30
+  idling: 5
   heartbeat: 300
   max_errors: 20
   max_errors_span: 600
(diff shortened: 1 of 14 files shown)


seanstory (approved):
I'm fine with this approach. But also think we could just change the scheduling from 1 second to 1 minute?

seanstory:
Alternatively, we could have kibana.yml create the job record for "on demand" instead of relying on schedules.

artem-shelkovnikov:
> I'm fine with this approach. But also think we could just change the scheduling from 1 second to 1 minute?
Scheduling is "every minute" now, so it would be the opposite - change it to every second. I can see this working for now too.

I also like the "on-demand" approach more, but realistically this won't happen for at least a week and just wanna fix this now :)

PR #1079, opened 2023-06-15
[Sharepoint Online] Support incremental sync

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/4704

This PR adds incremental sync support for the Sharepoint Online connector.

The connector uses [Microsoft Graph API Delta queries](https://learn.microsoft.com/en-us/graph/delta-query-overview) to retrieve changes. 

The flow looks like this: 

1. When a customer creates a connector they need to run a full sync. A full sync will store `deltaLink` for each `site drive` in the connectors document (`sync_cursor)`. 
2. When customers run incremental sync it validates the `sync_cursor` property and if it is not correct or one of the tokens is expired, customers will get a message and will be asked to run a full sync to fix the `sync_cursor` object. 

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -1229,6 +1229,9 @@ class SharepointOnlineDataSource(BaseDataSource):
         return self._sync_cursor
 
     def update_drive_delta_link(self, drive_id, link):
+        if not link:
+            return
+
         self._sync_cursor[CURSOR_SITE_DRIVE_KEY][drive_id] = link
 
     def get_drive_delta_link(self, drive_id):
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -1264,9 +1264,7 @@ class SharepointOnlineDataSource(BaseDataSource):
             return None
         elif (
             drive_item["size"] > MAX_DOCUMENT_SIZE
-            and not self.configuration[
-                "use_text_extraction_service"
-            ]
+            and not self.configuration["use_text_extraction_service"]
         ):
             logger.warning(
                 f"Not downloading file {drive_item['name']} of size {drive_item['size']}"
@@ -1274,9 +1272,7 @@ class SharepointOnlineDataSource(BaseDataSource):
 
             return None
         else:
-            drive_item["_original_filename"] = drive_item.get(
-                "name", ""
-            )
+            drive_item["_original_filename"] = drive_item.get("name", "")
             return partial(self.get_drive_item_content, drive_item)
 
     async def get_attachment_content(self, attachment, timestamp=None, doit=False):
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -1012,6 +1012,10 @@ class SharepointOnlineDataSource(BaseDataSource):
                                 "lastModifiedDateTime"
                             ]
 
+                            drive_item = await self._with_drive_item_access_control(
+                                site_drive, drive_item
+                            )
+
                             yield drive_item, self.download_function(
                                 drive_item, max_data_age
                             )
@@ -1074,6 +1078,10 @@ class SharepointOnlineDataSource(BaseDataSource):
                                 drive_item["lastModifiedDateTime"]
                                 if "lastModifiedDateTime" in drive_item
                                 else None
+                            )
+
+                            drive_item = await self._with_drive_item_access_control(
+                                site_drive, drive_item
                             )
 
                             yield drive_item, self.download_function(
@@ -1177,18 +1185,16 @@ class SharepointOnlineDataSource(BaseDataSource):
                         "lastModifiedDateTime"
                     ]
 
-                    list_item_attachment = (
-                        self._decorate_with_access_control(
-                            list_item_attachment,
-                            list_item.get(ACCESS_CONTROL, []),
-                        )
+                    list_item_attachment = self._decorate_with_access_control(
+                        list_item_attachment,
+                        list_item.get(ACCESS_CONTROL, []),
                     )
                     attachment_download_func = partial(
                         self.get_attachment_content, list_item_attachment
                     )
                     yield list_item_attachment, attachment_download_func
 
-            yield list_item
+            yield list_item, None
 
     async def site_lists(self, site):
         async for site_list in self.client.site_lists(site["id"]):
@@ -1208,10 +1214,7 @@ class SharepointOnlineDataSource(BaseDataSource):
             ]  # Apparantly site_page["GUID"] is not globally unique
             site_page["object_type"] = "site_page"
 
-            site_page = await self._with_site_page_access_control(
-                url, site_page
-            )
-
+            site_page = await self._with_site_page_access_control(url, site_page)
 
             for html_field in ["LayoutWebpartsContent", "CanvasContent1"]:
                 if html_field in site_page:
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -6,6 +6,7 @@
 import asyncio
 import os
 import re
+from collections.abc import Iterable, Sized
 from contextlib import asynccontextmanager
 from datetime import datetime, timedelta
 from functools import partial
@@ -17,6 +18,7 @@ from aiofiles.tempfile import NamedTemporaryFile
 from aiohttp.client_exceptions import ClientResponseError
 from fastjsonschema import JsonSchemaValueException
 
+from connectors.es.sink import OP_DELETE, OP_INDEX
 from connectors.filtering.validation import (
     AdvancedRulesValidator,
     SyncRuleValidationResult,
@@ -57,6 +59,15 @@ DEFAULT_RETRY_SECONDS = 30
 FILE_WRITE_CHUNK_SIZE = 1024
 MAX_DOCUMENT_SIZE = 10485760
 WILDCARD = "*"
+DRIVE_ITEMS_FIELDS = "id,content.downloadUrl,lastModifiedDateTime,lastModifiedBy,root,deleted,file,folder,package,name,webUrl,createdBy,createdDateTime,size,parentReference"
+
+CURSOR_SITE_DRIVE_KEY = "site_drives"
+
+# Microsoft Graph API Delta constants
+# https://learn.microsoft.com/en-us/graph/delta-query-overview
+
+DELTA_NEXT_LINK_KEY = "@odata.nextLink"
+DELTA_LINK_KEY = "@odata.deltaLink"
 
 
 class NotFound(Exception):
@@ -96,6 +107,14 @@ class TokenFetchFailed(Exception):
 class PermissionsMissing(Exception):
     """Exception class to notify that specific Application Permission is missing for the credentials used.
     See: https://learn.microsoft.com/en-us/graph/permissions-reference
+    """
+
+    pass
+
+
+class SyncCursorEmpty(Exception):
+    """Exception class to notify that incremental sync can't run because sync_cursor is empty.
+    See: https://learn.microsoft.com/en-us/graph/delta-query-overview
     """
 
     pass
@@ -279,6 +298,19 @@ class MicrosoftAPISession:
             else:
                 break
 
+    async def scroll_url(self, url):
+        scroll_url = url
+
+        while True:
+            graph_data = await self._get_json(scroll_url)
+
+            yield graph_data
+
+            if DELTA_NEXT_LINK_KEY in graph_data:
+                scroll_url = graph_data[DELTA_NEXT_LINK_KEY]
+            else:
+                break
+
     async def _get_json(self, absolute_url):
         async with self._call_api(absolute_url) as resp:
             return await resp.json()
@@ -449,28 +481,25 @@ class SharepointOnlineClient:
             for site_drive in page:
                 yield site_drive
 
-    async def drive_items(self, drive_id):
-        select = ""
+    async def drive_items_delta(self, url):
+        async for response in self._graph_api_client.scroll_url(url):
+            delta_link = (
+                response[DELTA_LINK_KEY] if DELTA_LINK_KEY in response else None
+            )
+            if "value" in response and len(response["value"]) > 0:
+                yield DriveItemsPage(response["value"], delta_link)
 
-        directory_stack = []
-
-        root = await self._graph_api_client.fetch(
-            f"{GRAPH_API_URL}/drives/{drive_id}/root?$select={select}"
+    async def drive_items(self, drive_id, url=None):
+        url = (
+            (
+                f"{GRAPH_API_URL}/drives/{drive_id}/root/delta?$select={DRIVE_ITEMS_FIELDS}"
+            )
+            if not url
+            else url
         )
 
-        directory_stack.append(root["id"])
-        yield root
-
-        while len(directory_stack):
-            folder_id = directory_stack.pop()
-
-            async for page in self._graph_api_client.scroll(
-                f"{GRAPH_API_URL}/drives/{drive_id}/items/{folder_id}/children?$select={select}"
-            ):
-                for drive_item in page:
-                    if "folder" in drive_item:
-                        directory_stack.append(drive_item["id"])
-                    yield drive_item
+        async for page in self.drive_items_delta(url):
+            yield page
 
     async def drive_item_permissions(self, drive_id, item_id):
         return await self._graph_api_client.fetch(
@@ -612,6 +641,37 @@ class SharepointOnlineClient:
         await self._http_session.close()
         self._graph_api_client.close()
         self._rest_api_client.close()
+
+
+class DriveItemsPage(Iterable, Sized):
+    """
+    Container for Microsoft Graph API DriveItem response
+
+    Parameters:
+        items (list<dict>):Represents a list of drive items
+        delta_link (str): Microsoft API deltaLink
+    """
+
+    def __init__(self, items, delta_link):
+        if items:
+            self._items = items
+        else:
+            self._items = []
+
+        if delta_link:
+            self._delta_link = delta_link
+        else:
+            self._delta_link = None
+
+    def __len__(self):
+        return len(self._items)
+
+    def __iter__(self):
+        for item in self._items:
+            yield item
+
+    def delta_link(self):
+        return self._delta_link
 
 
 class SharepointOnlineAdvancedRulesValidator(AdvancedRulesValidator):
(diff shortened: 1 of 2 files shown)


connectors/sources/sharepoint_online.py
```diff
@@ -261,6 +287,19 @@ async def scroll(self, url):
             else:
                 break
 
+    async def scroll_with_delta_link(self, url):
+        scroll_url = url
+
+        while True:
```
artem-shelkovnikov, line 293: Can probably be `while scroll_url` to indicate the condition of breaking the loop better

connectors/sources/sharepoint_online.py
```diff
@@ -406,28 +445,14 @@ async def site_drives(self, site_id):
             for site_drive in page:
                 yield site_drive
 
+    # Use Delta queries
```
artem-shelkovnikov, line 448: I would inline this comment to the method making it a docstring instead.

connectors/sources/sharepoint_online.py
```diff
@@ -511,6 +536,37 @@ def _validate_sharepoint_rest_url(self, url):
     async def close(self):
         await self._http_session.close()
 
+    async def site_drive_items_delta(self, url):
+        async for response in self._graph_api_client.scroll_with_delta_link(url):
+            delta_link = (
+                response[DELTA_LINK_KEY] if DELTA_LINK_KEY in response else None
+            )
+            if "value" in response and len(response["value"]) > 0:
+                yield DriveItemsPage(response["value"], delta_link)
```
artem-shelkovnikov, line 545: I'd move this method to be right above `drive_items` and rename to `drive_items_delta`

connectors/sources/sharepoint_online.py
```diff
@@ -511,6 +536,37 @@ def _validate_sharepoint_rest_url(self, url):
...
+
+
+class DriveItemsPage:
```
artem-shelkovnikov, line 548: The class itself does not seem specifically attached to Drive Items, so can be named like `DeltaResponsePage`.

On the other hand since it works only for Drive Items now it can be okay to keep it like that. Maybe adding a docstring can help?

Also marking this class with `iterable` and `sized` ABC would be good IMO: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
+            site_page["_id"] = site_page["GUID"]
+            site_page["object_type"] = "site_page"
+
+            for html_field in ["LayoutWebpartsContent", "CanvasContent1"]:
+                if html_field in site_page:
+                    site_page[html_field] = html_to_text(site_page[html_field])
+
+            yield site_page
+
+    def initiate_sync_cursor(self):
```
artem-shelkovnikov, line 927:
```suggestion
    def init_sync_cursor(self):
```

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
     async def get_docs(self, filtering=None):
         max_data_age = None
 
+        self.initiate_sync_cursor()
```
artem-shelkovnikov, line 722:
```suggestion
        self.init_sync_cursor()
```

connectors/sources/sharepoint_online.py
artem-shelkovnikov: Thinking of it, this method could be a property instead with get/set accessors hiding this.

Also eventually making SyncCursor a class with methods can help the abstraction imo, but also not a strong opinion.

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
-                    for html_field in ["LayoutWebpartsContent", "CanvasContent1"]:
-                        if html_field in site_page:
-                            site_page[html_field] = html_to_text(site_page[html_field])
+    async def get_docs_incrementally(self, sync_cursor, filtering=None):
+        self._sync_cursor = sync_cursor
 
-                    yield site_page, None
+        if self._sync_cursor is None or self._sync_cursor == {}:
+            raise SyncCursorEmpty(
+                "Incremental sync can't start because it doen't have information about the last full sync job. Please run a full sync to solve this."
```
artem-shelkovnikov, line 777: Idk - what if we make this message less specific just saying that "There's no cursor present"? E.g. not mentioning "full sync" as a reason for this, while still advising to run a full sync.

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
+
+                async for site_drive in self.site_drives(site["id"]):
+                    yield site_drive, None, OP_INDEX
+
+                    delta_link = self.get_drive_delta_link(site_drive["id"])
+
+                    if not delta_link:
+                        raise DeltaLinkCorrupted(
+                            "Incremental sync can't start because one of the delta tokens is missing or expired. Please run a full sync to solve this."
+                        )
```
artem-shelkovnikov, line 803: What about the situation when a new site appeared?

The message says nothing about it and it can be a legit cause.

Also for big customers who do a full sync every several days, it might make sense to make incremental sync pick up newly added site collections, what do you think?

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
+            yield site_page
+
+    def initiate_sync_cursor(self):
+        if not self._sync_cursor:
+            self._sync_cursor = {CURSOR_SITE_DRIVE_KEY: {}}
+
+        return self._sync_cursor
+
+    def update_drive_delta_link(self, drive_id, link):
+        self._sync_cursor[CURSOR_SITE_DRIVE_KEY][drive_id] = link
```
artem-shelkovnikov, line 934: Does it update the cursor in the index?

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
+    def get_drive_delta_link(self, drive_id):
+        return self._sync_cursor.get(CURSOR_SITE_DRIVE_KEY, {}).get(drive_id)
+
+    def drive_item_operation(self, item):
+        if "deleted" in item:
+            return OP_DELETE
+        else:
+            return OP_INDEX
+
+    def download_function(self, drive_item, max_data_age):
```
wangch079, line 953: This extraction itself is worth a PR, to make this PR cleaner.

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
+            for drive_item in page:
+                drive_item["_id"] = drive_item["id"]
+                drive_item["object_type"] = "drive_item"
+                drive_item["_timestamp"] = drive_item["lastModifiedDateTime"]
+
+                yield drive_item, self.download_function(drive_item, max_data_age)
+
+            self.update_drive_delta_link(drive_id=site_drive_id, link=page.delta_link())
+
+    async def site_list_items(
```
wangch079, line 875: This extraction itself is worth a PR, to make this PR cleaner.

connectors/sources/sharepoint_online.py
```diff
@@ -663,125 +719,268 @@ async def validate_config(self):
...
-                        if html_field in site_page:
-                            site_page[html_field] = html_to_text(site_page[html_field])
+                        yield list_item, download_func
 
+                # Sync site pages
+                async for site_page in self.site_pages(site["webUrl"]):
                     yield site_page, None
 
+    async def get_docs_incrementally(self, sync_cursor, filtering=None):
+        self._sync_cursor = sync_cursor
```
wangch079, line 773: I'd suggest assigning `sync_cursor` to a separate variable. `self._sync_cursor` is supposed to be updated to the connector when the sync is done successfully, while the `sync_cursor` is the cursor from which the incremental sync starts.

connectors/sources/sharepoint_online.py
```diff
@@ -663,125 +719,268 @@ async def validate_config(self):
...
+
+        if self._sync_cursor is None or self._sync_cursor == {}:
```
wangch079, line 775:
```suggestion
        if not self._sync_cursor:
```

connectors/sources/sharepoint_online.py
wangch079: Agree, a drive with no cursor should try to retrieve with `root/delta`

connectors/sources/sharepoint_online.py
```diff
@@ -663,125 +719,268 @@ async def validate_config(self):
...
+        return self._sync_cursor
+
+    def update_drive_delta_link(self, drive_id, link):
+        self._sync_cursor[CURSOR_SITE_DRIVE_KEY][drive_id] = link
+
+    def get_drive_delta_link(self, drive_id):
+        return self._sync_cursor.get(CURSOR_SITE_DRIVE_KEY, {}).get(drive_id)
+
+    def drive_item_operation(self, item):
+        if "deleted" in item:
```
wangch079, line 948: It seems we should check `@removed` for delete operation:
>Removed instances are represented by their id and an @removed object. The @removed object may include additional information about why the instance was removed. For example, "@removed": {"reason": "changed"}.

Ref: https://learn.microsoft.com/en-us/graph/delta-query-overview#resource-representation-in-the-delta-query-response

What about `OP_UPSERT`?

wangch079 (commented):
CMIIW, Among all the documents we ingest, namely, site, drive, list, only drive are synced incrementally (I only see drive are populated into the `sync_cursor`)

connectors/sources/sharepoint_online.py
```diff
@@ -406,28 +445,14 @@ async def site_drives(self, site_id):
...
-
-            async for page in self._graph_api_client.scroll(
-                f"{GRAPH_API_URL}/drives/{drive_id}/items/{folder_id}/children?$select={select}"
-            ):
-                for drive_item in page:
-                    if "folder" in drive_item:
-                        directory_stack.append(drive_item["id"])
-                    yield drive_item
+        async for page in self.site_drive_items_delta(url):
+            yield page
```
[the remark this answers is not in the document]
wangch079: I see. The `page` here refers to an `DriveItemsPage` instance. It's an iterable object, which will iterate the drive items, and it also contains the `delta_link`. 

I guess you introduced this new class to defer the call to `update_drive_delta_link`. Are you worrying about the situation where the cursor is updated, but something wrong happened and the drive item is not actually ingested?

If my assumption is true, I think it's not necessary:
1. fetching and indexing are separated, a document is fetched doesn't mean it's indexed into ES.
2. We only update `sync_cursor` to `connector` after a successful sync job. If an error happened, it won't be updated.

connectors/sources/sharepoint_online.py
vidok: Yeah, this one I just forgot to remove. I'll add more comments using docstrings and this one will be just removed.

connectors/sources/sharepoint_online.py
vidok: Yeah, I'd like to keep it as it is right now. We might change it later when `List Items` API will be available. 

I'll have a look at ABC. Thanks.

connectors/sources/sharepoint_online.py
vidok: `sync_cursor` has complex logic so I would still hide getters/setters, especially when we start storing data for `List Items`. 

I would also make a `SyncCursor` class. I'll create a tech debt issue.

connectors/sources/sharepoint_online.py
vidok: This updates just the instance variable then when the job is done, it will be store in the connectors index ([ref](https://github.com/elastic/connectors-python/blob/dmitrii/4704-sharepoint-incremental-sync-2/connectors/protocol/connectors.py#L708))

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
+                async for site_drive in self.site_drives(site["id"]):
                     yield site_drive, None
 
-                    async for drive_item in self.client.drive_items(site_drive["id"]):
-                        drive_item["_id"] = drive_item["id"]
-                        drive_item["object_type"] = "drive_item"
-                        drive_item["_timestamp"] = drive_item["lastModifiedDateTime"]
-
-                        download_func = None
+                    async for page in self.client.drive_items(site_drive["id"]):
```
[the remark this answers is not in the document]
vidok: Because we are scrolling pages not drive_items.

vidok:
> CMIIW, Among all the documents we ingest, namely, site, drive, list, only drive are synced incrementally (I only see drive are populated into the `sync_cursor`)

Yes, you are right. We will need to add `list item` support when the API is ready (it's in beta state right now)

connectors/sources/sharepoint_online.py
vidok: Could you please add more context because I'm not sure I understand what you are saying here.

connectors/sources/sharepoint_online.py
vidok: I agree with you both. What made me do it like that is a possible situation (I don't trust users' input by default) where `sync_cursor` is initiated but doesn't have any delta links in it. In this case, incremental sync job time will be the save as full sync job time.

connectors/sources/sharepoint_online.py
vidok: I tried `OP_UPSERT` but it does not trigger ingest pipelines because ingest pipeline trigger only when you index or re-index documents. Do you a better way to handle this?

vidok: > It seems we should check `@removed` for delete operation:
> 
> > Removed instances are represented by their id and an @removed object. The @removed object may include additional information about why the instance was removed. For example, "@removed": {"reason": "changed"}.
> 
> Ref: https://learn.microsoft.com/en-us/graph/delta-query-overview#resource-representation-in-the-delta-query-response

This is weird because this is what I'm getting when a doc gets deleted:

```
{
'@odata.type': '#microsoft.graph.driveItem', 
  'id': '01WKXJTCYBH6NXECRZ4ZFZVVQIJJ2LPEJD', 
  'size': 0, 
  'parentReference': {'driveType': 'documentLibrary', 'driveId': 'b!1SwS-wjro0ew8puOF4o0UwemLfYyGE9OgLUZ5HpVe-zwJGwDHhGKRaucdMWq3ZAv', 'id': '01WKXJTC56Y2GOVW7725BZO354PWSELRRZ', 'siteId': 'fb122cd5-eb08-47a3-b0f2-9b8e178a3453'}, 
  'deleted': {'state': 'deleted'}, 
  'file': {'hashes': {}}, '_id': '01WKXJTCYBH6NXECRZ4ZFZVVQIJJ2LPEJD', 'object_type': 'drive_item', '_timestamp': None
}
```

connectors/sources/sharepoint_online.py
danajuratoni: I would like to avoid exposing such detailed implementation logic and expand the concepts a user needs to be aware of. _What is a cursor? Why would I, as a customer, care about cursors?_

I think a message like

> Incremental sync can't start because required information is missing. Please run a full sync to allow incremental syncs to run again.

would be useful.
cc: @leemthompo

leemthompo: I agree with @danajuratoni.

Tiny copy tweak suggestion:

`Unable to start incremental sync. Please perform a full sync to re-enable incremental syncs.`

connectors/sources/sharepoint_online.py
wangch079: I meant, this change is to extract a block of code into a new function `download_function`, and this can be a separate PR, so that this PR is smaller and cleaner to review.

connectors/sources/sharepoint_online.py
```diff
@@ -663,118 +719,259 @@ async def validate_config(self):
...
+                if html_field in site_page:
+                    site_page[html_field] = html_to_text(site_page[html_field])
+
+            yield site_page
+
+    def initiate_sync_cursor(self):
+        if not self._sync_cursor:
+            self._sync_cursor = {CURSOR_SITE_DRIVE_KEY: {}}
+
+        return self._sync_cursor
```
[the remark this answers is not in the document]
vidok: Technically, it does make sense to do so because `nil`  can be considered as a successful operation. I assume I will change it anyway based on other feedback.

connectors/sources/sharepoint_online.py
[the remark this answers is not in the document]
vidok: > I guess you introduced this new class to defer the call to `update_drive_delta_link`.

This is correct. I didn't want to check the delta link existence every iteration. Event if I change the cursor state it's will be updated only when sync is successfully finished right?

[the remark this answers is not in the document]
vidok: Do you mean it's safer to assign `cursor` at the end of the method to make everything went well?

artem-shelkovnikov (dismissed):

connectors/sources/sharepoint_online.py
wangch079: > I tried `OP_UPSERT` but it does not trigger ingest pipelines because ingest pipeline trigger only when you index or re-index documents. Do you a better way to handle this?

TIL. I didn't find it documented in the [Bulk API doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html), but it does stated [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html#add-pipeline-to-indexing-request):

>Use the `pipeline` query parameter to apply a pipeline to documents in [individual](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html) or [bulk](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) indexing requests.
You can also use the `pipeline` parameter with the [update by query](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html) or [reindex](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html) APIs.

connectors/sources/sharepoint_online.py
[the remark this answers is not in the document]
vidok: In your code, `drive_items_delta` method is hiding the fact that it's not just retrieving drive items but also updating the cursor. This breaks [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle).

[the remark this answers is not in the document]
wangch079: Or do you think it's better to `update_drive_delta_link` in `scroll_url`? I think it's natural that the `delta_link` should be saved when you have scrolled to the last page, when `@odata.nextLink` is not present and `@odata.deltaLink` is returned.

timgrein (approved):
LGTM!

Issue #1032, opened 2023-06-08
[S3]Replace AWS CLI configuration and add it in s3.py configuration


akanshi-elastic:
### Problem Description
Currently AWS S3 is using `aws cli` to add the values such as `access_key`, `secret_key` and `region`.
Instead of using command-line to store these values, Start adding these values to s3 connector specific configuration.

### Proposed Solution
Connector Should run by adding` access_keys`, `secret_key` and `region` from connector configuration instead of using `aws cli` to add these values

akanshi-elastic:
Started working on this 12th june . It is testing state as of now. Will raise a github PR by eod tommorow.

PR #1082, opened 2023-06-16
[S3]Replace AWS CLI configuration and add it in s3.py configuration

akanshi-elastic:
## Closes https://github.com/elastic/connectors-python/issues/1032###
This PR includes:
1.Add aws keys to connector specific configuration to make it run as native connector
#### Pre-Review Checklist


connectors/sources/s3.py
```diff
@@ -311,11 +315,33 @@ def get_default_configuration(cls):
...
+                "value": "A1B2C3D4",
+                "sensitive": True,
+            },
+            "region": {
+                "display": "textarea",
+                "label": "AWS Region name",
+                "order": 4,
+                "type": "str",
+                "value": "us-east-1",
+            },
```
artem-shelkovnikov, line 339: How does it affect the S3 connector?

Does it restrict which buckets can be synced?

akanshi-elastic: No there is no restrictions with this but as we discussed on call previously that if we are not passing region then we get access denied error.. That's why we need to pass it as of now.. 
We are trying to take update on this:https://github.com/aio-libs/aiobotocore/issues/1005#issuecomment-1530383313
once we get the update on this so we can check confirm on this.

akanshi-elastic: We were facing some exception related to None type region when try to run the connector without passing region name. However it's not reproducible now and looks like there might be some issue in our s3 configuration that time. 

We tested our connector removing the region as you suggested and found that without passing region name our connector is working fine with all type of buckets. 
So we can remove the dependency of region & I've updated PR with the same too

artem-shelkovnikov (approved):

akanshi-elastic:
We need to update readme for s3 as well but we don't have permission to update it. Can you please update these 3 points there?
1.From usage remove: `Set up the AWS CLI tool`
2.Setup everything written under `Set up AWS CLI`
3.In `Confiuration` we have add two new fields `aws_access_key_id` and `aws_secret_access_key` just after the buckets field

artem-shelkovnikov:
cc @danajuratoni - we need to update readme, should we do it ourselves or forward to the tech writers?

artem-shelkovnikov (approved):

PR #1085, opened 2023-06-19
Rename maxDataAge to skipDriveItemsOlderThan

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/4417

Renaming "maxDataAge" to "dontSubextractDriveItemsOlderThan" - might reflect better the logic.

This setting controls subextraction of Drive Items based on last modified timestamp.

`dontSubextractDriveItemsOlderThan: 20` means that Drive Items modified more than 20 days ago will not be subextracted.


timgrein (dismissed):
Makes sense, LGTM!

connectors/sources/sharepoint_online.py
```diff
@@ -682,7 +682,7 @@ class SharepointOnlineAdvancedRulesValidator(AdvancedRulesValidator):
     SCHEMA_DEFINITION = {
         "type": "object",
         "properties": {
-            "maxDataAge": {"type": "integer"},
+            "dontSubextractDriveItemsOlderThan": {"type": "integer"},  # in Days
```
wangch079, line 685: `Days` or `seconds`?

https://github.com/elastic/connectors-python/blob/d7172db01603c068e866c08a50ec7e84d403eeb3/connectors/sources/sharepoint_online.py#L1257-L1259

wangch079 (changes requested):
Should also rename in `download_function`:
https://github.com/elastic/connectors-python/blob/d7172db01603c068e866c08a50ec7e84d403eeb3/connectors/sources/sharepoint_online.py#LL1246C9-L1246C26

connectors/sources/sharepoint_online.py
artem-shelkovnikov: You are right, not sure why seconds are there :/

connectors/sources/sharepoint_online.py
```diff
@@ -1048,11 +1048,11 @@ async def get_docs_incrementally(self, sync_cursor, filtering=None):
                 "Unable to start incremental sync. Please perform a full sync to re-enable incremental syncs."
             )
 
-        max_data_age = None
+        max_drive_item_age = None
 
         if filtering is not None and filtering.has_advanced_rules():
             advanced_rules = filtering.get_advanced_rules()
-            max_data_age = advanced_rules["maxDataAge"]
+            max_drive_item_age = advanced_rules["dontSubextractDriveItemsOlderThan"]
```
wangch079, line 1055: Think this can be extracted to a new method

wangch079 (approved):
LGTM, one potential improvement

connectors/sources/sharepoint_online.py
artem-shelkovnikov: Absolutely, we can address it with subsequent PRs

PR #1086, opened 2023-06-19
[DLS] Set data provider features to corresponding connector features

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4970

This PR sets the data provider features to its corresponding connector features. We need that to adjust the logic inside a data source implementation based on the enabled/disabled features.

#### Pre-Review Checklist


connectors/sync_job_runner.py
```diff
@@ -118,6 +118,12 @@ async def execute(self):
                 await self._sync_done(sync_status=JobStatus.COMPLETED)
                 return
 
+            if hasattr(self.data_provider, "features"):
+                logger.warning(f"{self.data_provider.name} already has 'features' set.")
+            else:
+                logger.debug(f"Setting 'features' for {self.data_provider.name}")
+                self.data_provider.features = self.connector.features
```
wangch079, line 125: Should we have a method `set_features` in `BaseDataSource`?

I'm about to introduce a `set_logger` method as well: https://github.com/elastic/connectors-python/pull/806#discussion_r1231472801

timgrein: Makes sense, it's more consistent 👍

artem-shelkovnikov (approved):
Are you planning to wrap checking features in a function?

Otherwise might make sense to init _features as empty hash/array - depending on the data structure.

timgrein:
> Are you planning to wrap checking features in a function?
> 
> Otherwise might make sense to init _features as empty hash/array - depending on the data structure.

We've an explicit `Features` object, which already provides high-level methods for checking for certain features 👍

PR #1088, opened 2023-06-19
[DLS] Add permissions to SPO docs

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4303

This PR adds the `_allow_permissions` field to the SPO docs fetched.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
Nice job!

connectors/sources/sharepoint_online.py
```diff
@@ -673,6 +770,139 @@ async def validate_config(self):
                 f"The specified SharePoint sites [{', '.join(missing)}] could not be retrieved during sync. Examples of sites available on the tenant:[{', '.join(remote_sites[:5])}]."
             )
 
+    async def _site_access_control(self, site):
```
wangch079, line 773: I wonder why all other methods are `_with_{item}_access_control`, which delegate to `_decorate_with_access_control` in the end, and why this method and `_list_item_access_control` are special?

All the `_with_{item}_access_control` methods are used together with  `if self._dls_enabled()`, do you think we move this check inside the `_with_{item}_access_control` method? E.g.:

```python
async def _with_drive_item_access_control(self, site_drive, drive_item):
    if not self._dls_enabled():
        return drive_item

    ...
```

connectors/sources/sharepoint_online.py
```diff
@@ -555,6 +644,14 @@ async def validate(self, advanced_rules):
             )
 
 
+def _decorate_with_access_control(document, access_control):
+    document[ACCESS_CONTROL] = (
+        document.get(ACCESS_CONTROL, []) + access_control + DEFAULT_GROUPS
```
wangch079, line 649: Do you think you need to uniquify the list?

wangch079 (commented):
Sorry for the late review

connectors/sources/sharepoint_online.py
wangch079: Noted. Because the `site_access_control` and `list_item_access_control` are reused.

connectors/sources/sharepoint_online.py
wangch079: In what scenario, `document.get(ACCESS_CONTROL, [])` returns a non-empty list?

PR #1096, opened 2023-06-21
Make extraction service config optional

navarone-feekery:
### Related to https://github.com/elastic/enterprise-search-team/issues/4972

Make the `extraction_service` config setting optional.
This is to prevent users from needing to manage this configuration even if they aren't using this feature.

- Revert all config changes made in https://github.com/elastic/connectors-python/pull/1089
- Convert configuration check-up to be based on `extraction_service` configuration presence rather than an enabled flag.
- Don't attempt to extract content if `extraction_service` has not been configured.
- Remove the file-pointer extraction method
  - This feature did not make it in time for 8.9 FF on the data-extraction-service side, but we will re-add it in 8.10.

#### Pre-Review Checklist

## Related Pull Requests

This is a follow-up PR based on feedback here: https://github.com/elastic/connectors-python/pull/1089#discussion_r1235272058

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -27,13 +27,6 @@ service:
   job_cleanup_interval: 300
   log_level: INFO
 
-extraction_service:
-  enabled: false
-  host: http://localhost:8090
-  text_extraction:
-    use_file_pointers: false
-    method: 'tika'
-
 connector_id: 'changeme'
 service_type: 'changeme'
 
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -677,16 +677,18 @@ class ExtractionService:
         ) as file:
             config = yaml.safe_load(file)
 
+        self.session = None
+
         self.extraction_config = config.get("extraction_service", None)
         if self.extraction_config is not None:
             self.host = self.extraction_config.get("host", None)
         else:
             self.host = None
-            logger.warning(
-                "Extraction service has been initialised but no extraction service configuration was found."
-            )
 
-        self.session = None
+        if self.host is None:
+            logger.warning(
+                "Extraction service has been initialised but no extraction service configuration was found. No text will be extracted for this sync."
+            )
 
     def _check_configured(self):
         if self.host is not None:
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -677,11 +677,22 @@ class ExtractionService:
         ) as file:
             config = yaml.safe_load(file)
 
-        self.host = config["extraction_service"]["host"]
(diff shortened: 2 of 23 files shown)


connectors/utils.py
```diff
@@ -677,12 +677,25 @@ def __init__(self):
         ) as file:
```
wangch079, line 677: I must have missed the addition of this class, but I do think `ExtractionService` should be instantiated with the config, instead of reading it from a hardcoded path. But we can address it in a separate PR

connectors/utils.py
```diff
@@ -677,12 +677,25 @@ def __init__(self):
...
+        else:
+            self.host = None
+
+        if self.host is None:
+            logger.warning(
+                "Extraction service has been initialised but no extraction service configuration was found. No text will be extracted for this sync."
+            )
+
+    def _check_configured(self):
+        if self.host is not None:
```
wangch079, line 694: Should check `enabled` as well?

connectors/utils.py
navarone-feekery: This is what I attempted to do originally, but it involved too much invasive changes to the connector framework. The best option would be for this to be a stand-alone service but that was out of scope for 8.9, so I used this temporary measure. I also clarified in projects sync that it was okay to read the config file here (only for 8.9).

connectors/utils.py
navarone-feekery: We are removing `enabled` entirely. The presence of the config will determine if it is enabled or not.

config.yml
```diff
@@ -27,13 +27,6 @@ service:
   job_cleanup_interval: 300
   log_level: INFO
 
-extraction_service:
```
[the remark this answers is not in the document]
navarone-feekery: @seanstory had brought up on the PR that introduced the `enabled` flag here that we don't want the users to have to worry about fiddling with this config if they aren't going to use it. It's an entirely optional config for only one connector right now. It is much cleaner and easier on the user to not include the config from the start, and the user can add it if they want to use it (we will have documentation on how to do this).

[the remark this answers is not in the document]
navarone-feekery: This was the discussion: https://github.com/elastic/connectors-python/pull/1089#discussion_r1235272058

seanstory (approved):
LGTM. Thanks for making this change.

connectors/utils.py
wangch079: Noted.

wangch079 (approved):

PR #1097, opened 2023-06-21
bump connector versions

efegurkan:
Bump connector version


.backportrc.json
```diff
@@ -1,13 +1,14 @@
 {
   "targetBranchChoices": [
     { "name": "main", "checked": true },
+    "8.9",
     "8.8",
     "8.7"
```
wangch079, line 6: I think we could remove `8.7`, as there won't be any further patch release for `8.7`

tests/ftest.sh
```diff
@@ -22,7 +22,7 @@ MAX_DURATION=600
 export REFRESH_RATE="${REFRESH_RATE:-5}"
 export DATA_SIZE="${DATA_SIZE:-medium}"
 export RUNNING_FTEST=True
-export VERSION='8.9.0-SNAPSHOT'
+export VERSION='8.10.0-SNAPSHOT'
```
wangch079, line 25: Not sure about this but I think this will make the nightly fail until there's an 8.10 snapshot available? cc @tarekziade

artem-shelkovnikov: Yup this should be bumped in a different PR and merged later

wangch079 (approved):

PR #1107, opened 2023-06-22
Update logging in the Developer's guide

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/3809

This PR updates the Developer's guide on how to log in the connector implementation.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
LGTM, though afaik `_logger` means that it should not be used by anyone at risk of non-compatible changes.

Should we hide logger behind a property `self.logger`?

wangch079:
> LGTM, though afaik `_logger` means that it should not be used by anyone at risk of non-compatible changes.

I think `_logger` means it's a private variable and should only be used inside the object (i.e. data provider). So it's only expected to be used like `self._logger.info(...`, instead of `data_provider._logger.info(...`

> Should we hide logger behind a property `self.logger`?

No, not intended to expose it for public use.

PR #944, opened 2023-05-24
Update connector service on-prem doc

wangch079:
Update the markdown document to reflect the latest UI changes in 8.8

#### Pre-Review Checklist
- [x] this PR has a meaningful title
- [ ] this PR links to all relevant github issues that it fixes or partially addresses
- [ ] if there is no GH issue, please create it. Each PR should have a link to an issue
- [x] this PR has a thorough description
- [x] Covered the changes with automated tests
- [x] Tested the changes locally
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

PR #1109, opened 2023-06-22
Remove references to running natively

seanstory:
## Part of https://github.com/elastic/enterprise-search-team/issues/4734

Removes docs references to running in "native mode" as this is no longer something that end-users should be doing. Native mode is reserved for Elastic Cloud.

#### Pre-Review Checklist


docs/CONFIG.md
```diff
@@ -38,30 +38,23 @@ Configuration lives in [config.yml](../config.yml).
   - `max_concurrent_access_control_syncs`: The maximum number of concurrent access control syncs. Defaults to 1.
   - `job_cleanup_interval`: The interval (in seconds) to run job cleanup task.
   - `log_level`: Connector service log level. Defaults to `INFO`.
-- `native_service_types`: An array of supported native connectors (in service type).
```
wangch079, line 41: So `native_service_types` is supposed to be used only on Cloud, and that's why it's removed. Do you think we should remove `username` and `password` as well? It's more intuitive to use the API key for on-prem connector service.

docs/CONFIG.md
```diff
@@ -38,30 +38,23 @@ Configuration lives in [config.yml](../config.yml).
...
-3. Run the connector service with
-    ```shell
-    make run
-    ```
-
-### Run the connector service for a custom connector
+## Run the connector service on-prem
 
-1. Go to Kibana, _Enterprise Search_ > _Create an Elasticsearch index_. Use the `Build a connector` option for an ingestion method to create an index.
+1. Go to Kibana, _Enterprise Search_ > _Create an Elasticsearch index_. Select the type of connector you wish to use.
```
wangch079, line 57: The on-prem connector service will work for:
1. a connector client
2. a customized connector
3. a converted native connector

Probably we should make this step more general:

```suggestion
When you have a connector client, a customized connector, or a converted native connector, you have to run the connector service on-prem.

1. In Kibana, go to the `Configuration` tab of the connector.
```

wangch079 (commented):
Thanks for the work, I think I can close this PR: https://github.com/elastic/connectors-python/pull/944

docs/CONFIG.md
leemthompo: Driving by here, but I think a connector client is what we call a self-managed connector. So you use the on-prem connector service to run connector clients.

Whether these are customized connectors or converted native connectors is perhaps irrelevant here. 🤷

seanstory: I can work on the phrasing here to capture both points. That any "converted native connector," "net-new custom connector" or "customized/modified connector" must be run as a connector client, on-prem. WDYT?

leemthompo: 👍 

FYI I wonder if we could rename connector clients as self-managed connectors? It would be so much clearer. We've done the hard work of removing confusion around "native"!

seanstory: I would prefer that. But I think that @danajuratoni has been very intentional about our taxonomy thus-far, so I'm hesitant to advocate too strongly.

leemthompo: Indeed, it's not easy to backtrack and rename stuff

docs/CONFIG.md
wangch079: Fair enough. username/password is provided as an additional auth method, while `native_service_types` is something not supposed to be used by customers.

wangch079 (approved):

PR #1114, opened 2023-06-23
populate features for connector clients and customized connectors

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/5007

For connector clients and customized connectors, connector service will populate features available from the data source, if it doesn't match the features in the connector document.

#### Pre-Review Checklist


connectors/protocol/connectors.py
```diff
@@ -762,27 +765,32 @@ async def prepare(self, config):
                 config["sources"][configured_service_type], configured_service_type
             )
 
-            return
-
         doc = {}
+        fqn = config["sources"][configured_service_type]
```
seanstory, line 769: What does `fqn` stand for? I see it came from below, I'm just curious.

connectors/source.py
```diff
@@ -414,6 +413,11 @@ def hash_id(cls, _id):
 
         return hash_id(_id)
 
+    @classmethod
+    def features(cls):
+        """Returns features available for the data source"""
+        raise NotImplementedError
+
```
seanstory, line 420: 👍 
My original instinct was that this should have an implementation, to limit how many concrete data sources needed to override this function. The more I think about it though, the more it makes sense this way, since we can't guarantee that any custom class that extends BaseDataSource would be able to automatically support any new features we add.

seanstory: However, we probably now need to note this function in DEVELOPING.md as one that needs to be implemented. (Can capture that in a follow-up issue if you'd prefer)

connectors/sources/confluence.py
```diff
@@ -313,6 +314,27 @@ def get_default_configuration(cls):
...
+                },
+                "document_level_security": {
+                    "enabled": False,
+                },
+                "incremental_sync": {
+                    "enabled": False,
+                },
+            }
+        )
+
```
seanstory, line 337: I'm a little worried about having this structure bleed into our source code. What will maintenance look like if we decide to change how we structure this object in `.elastic-connectors`? How would we be able to handle needing to change this structure in any custom-made connectors that we don't have in our own source-control?

Part of me thinks that we should have just a list of keys of features that _are_ supported, and let the framework take that list and format into the correct structure. Or another approach is that we could add a `supports_x_feature()` function to the BaseDataSource for each new feature, and have the default impl be false?

WDYT?

connectors/protocol/connectors.py
wangch079: [Fully qualified name](https://en.wikipedia.org/wiki/Fully_qualified_name) 

I think this has confused many people, and we should consider renaming it 😅

connectors/sources/confluence.py
wangch079: >Part of me thinks that we should have just a list of keys of features that are supported, and let the framework take that list and format into the correct structure. Or another approach is that we could add a `supports_x_feature()` function to the BaseDataSource for each new feature, and have the default impl be false?

This is a good idea. Instead of `supports_x_feature()`, I probably will use class variables: https://github.com/elastic/connectors-python/pull/986#discussion_r1213385310

wangch079: Changes were done here: https://github.com/elastic/connectors-python/pull/1114/commits/50b420bc9ef3548918d5fdb7f00c8ce456406b77, which reduced the number of files changed to 12.

connectors/sources/azure_blob_storage.py
```diff
@@ -108,6 +109,27 @@ def get_default_configuration(cls):
             },
         }
 
+    @classmethod
```
timgrein, line 112: Here we probably also want to use the class-level fields?

wangch079: I forgot to remove this 😅

connectors/source.py
wangch079: >However, we probably now need to note this function in DEVELOPING.md as one that needs to be implemented. (Can capture that in a follow-up issue if you'd prefer)

That will be part of https://github.com/elastic/enterprise-search-team/issues/4641

seanstory (approved):
LGTM. Thanks!

PR #1127, opened 2023-06-26
Few little fixes for ACL index prefix

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/5019

Fixes two problems with the ACL index prefix.

- the prefix should be `.search-acl-filter-<name>` (preceeding .)
- the suffix should not start with `search-` (`search-test` -> `.search-acl-filter-test`, not `.search-acl-filter-search-test`)

See also: https://github.com/elastic/kibana/pull/160457/files#diff-26af76102aa2bfc90781bdfc75ca6893573cc8fba97ff4cdcfcf7fa37b50aea5R15-R17

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/kibana/pull/160457/files#diff-26af76102aa2bfc90781bdfc75ca6893573cc8fba97ff4cdcfcf7fa37b50aea5R15-R17


artem-shelkovnikov:
Hi @seanstory - can you update this issue to have meaningful description - listing the changes made and why the changes are made?

connectors/protocol/connectors.py
```diff
@@ -941,7 +942,11 @@ async def create(self, connector, trigger_method, job_type):
         index_name = connector.index_name
 
         if job_type == JobType.ACCESS_CONTROL:
-            index_name = f"{ACCESS_CONTROL_INDEX_PREFIX}{index_name}"
+            index_name = re.sub(
+                r"^(?:search-)?(.*)$",
+                rf"{ACCESS_CONTROL_INDEX_PREFIX}\g<1>",
+                index_name,
+            )
```
wangch079, line 949: This works. But I think we should probably introduce a new field `acl_index_name` in the connector index in the first place, which will involve changes to :
1. Kibana, when creating a connector document
2. ent-search, index mapping and migration
3. connectors-python, assign `connector.acl_index_name` instead of `re.sub`.

Effort for each change should be minimal. Do you think it's too late?

connectors/sync_job_runner.py
```diff
@@ -160,7 +160,7 @@ async def execute(self):
         except ConnectorJobCanceledError:
             await self._sync_done(sync_status=JobStatus.CANCELED)
         except Exception as e:
-            self.sync_job.log_critical(e, exc_info=True)
+            self.sync_job.log_error(e, exc_info=True)
```
seanstory, line 163: Unrelated, just saw this error was the wrong level and sneaking in the fix.

seanstory:
> Hi @seanstory - can you update this issue to have meaningful description - listing the changes made and why the changes are made?

Updated.

connectors/protocol/connectors.py
seanstory: This would be in an effort to not have this logic living in two places (kibana and connectors)? But I don't think that would solve anything.
- kibana would still need to be able to compute what this should be when you create a native connector through Kibana.
- connectors would still need to be able to compute what this should be when registering a connector client

So I'm not sure I see the value of embedding th`acl_index_name` into the connectors document. Especially when the value is deterministic.

wangch079: >kibana would still need to be able to compute what this should be when you create a native connector through Kibana.

Kinana is already doing so for `connector.index_name` by prefixing the value from the customer with `search-`, it can just prefix with `.search-acl-filter-` for `connector.acl_index_name`.

>connectors would still need to be able to compute what this should be when registering a connector client

Do you mean when a connector client is created in Kibana, Kibana won't create the index `.search-acl-filter-{name}`?

wangch079 (approved):
LGTM. I don't think it's necessary to push the proposed change here: https://github.com/elastic/connectors-python/pull/1127#discussion_r1243055644. This is not a breaking change and we can do so in later release if we think it's necessary.

connectors/protocol/connectors.py
seanstory: > Kinana is already doing so for connector.index_name by prefixing the value from the customer with search-, it can just prefix with .search-acl-filter- for connector.acl_index_name.

Fair point.

> Do you mean when a connector client is created in Kibana, Kibana won't create the index .search-acl-filter-{name}?

That is not what I'd meant, but that is correct. The ACL index is defined in an "index template" that allows us to create the index with the correct mappings/settings when it's first added to, regardless of what tries to add to it. Currently, only the connector writes to this index pattern. 

> I don't think it's necessary to push the proposed change here: https://github.com/elastic/connectors-python/pull/1127#discussion_r1243055644. This is not a breaking change and we can do so in later release if we think it's necessary.

Thanks! I'm going to go ahead and merge so that we can have this working in the next BC, but I'm fine with changing it in a future patch if you feel its important.

PR #1131, opened 2023-06-27
[Google Drive] Connector implementation

jedrazb:
## Closes https://github.com/elastic/enterprise-search-team/issues/4992

Implementation of the Google Drive connector based on [Google Drive API v3](https://developers.google.com/drive/api/reference/rest/v3). 

#### Authentication

The connector is authenticated as a service account.
- [Service account overview](https://cloud.google.com/iam/docs/service-account-overview)

#### Ingestion and content extraction

- Connector extracts and ingests same fields as Workplace Search Connector
- Support for downloading Google Suite documents and stored files and folders
- Content extraction is supported
- It uses `aiogoogle` library under the hood for async API calls

#### Testing
- Unit tests, 98% coverage
- Functional test added 
- Manual tests on different sizes of data corpus (up to 85k docs) with:
  - personal account
  - domain (organizational) account Screenshot 2023-06-28 at 11 17 28

- Local run of perf8 is happy ✅ 
Screenshot 2023-07-04 at 16 36 02

#### Documentation

TODO

#### Pre-Review Checklist


connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+from connectors.source import BaseDataSource, ConfigurableFieldValueError
+from connectors.utils import TIKA_SUPPORTED_FILETYPES, convert_to_b64, get_pem_format
+
+DEFAULT_RETRY_COUNT = 3
+DEFAULT_WAIT_MULTIPLIER = 2
+DEFAULT_FILE_SIZE_LIMIT = 10485760
+
+GOOGLE_DRIVE_READ_ONLY_SCOPE = "https://www.googleapis.com/auth/drive.readonly"
+API_NAME = "drive"
+API_VERSION = "v3"
```
artem-shelkovnikov, line 27: Idk if it's worth to put to constants - it seems to be used in a single place, so can either be moved to the only class using it or even inlined IMO

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+            log_message (str): Warning message logged
+
+        Returns:
+            int: increased retry_counter
+        """
+        retry_counter += 1
+        if retry_counter > self.retry_count:
+            raise
+        logger.warning(log_message)
+        await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry_counter)
```
artem-shelkovnikov, line 160: Avoid using `asyncio.sleep` and use `CancellableSleeps` from `utils.py`.

Problem with `asyncio.sleep` is that it will keep the process running if you want to terminate it, for example you do `await asyncio.sleep(60)` and then CTRL+C the host - it'll have to wait until `asyncio.sleep` wakes up, which is 60 seconds - too much. We want to be able to cancel these sleeps as well.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+        """
+        json_credentials = json.loads(self.configuration["service_account_credentials"])
+
+        if (
+            json_credentials.get("private_key")
+            and "\n" not in json_credentials["private_key"]
+        ):
+            json_credentials["private_key"] = get_pem_format(
+                key=json_credentials["private_key"].strip(),
+                max_split=2,
```
artem-shelkovnikov, line 232: Can you also validate the private_key? Will `get_pem_format` raise an error if private key is invalid?

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+
+        return GoogleDriveClient(
+            json_credentials=json_credentials,
+            retry_count=self.configuration["retry_count"],
+        )
+
+    async def ping(self):
+        """""Verify the connection with Google Drive""" ""
+        if RUNNING_FTEST:
+            return
```
artem-shelkovnikov, line 243: I think a logger debug line is worth to add here, that will indicate that the method is skipped for FTEST

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+        self.configuration.check_valid()
+
+        try:
+            json.loads(self.configuration["service_account_credentials"])
+        except ValueError as e:
+            raise ConfigurableFieldValueError(
+                "Google Drive service account is not a valid JSON."
+            ) from e
+
+    async def get_drives(self):
```
artem-shelkovnikov, line 271: Do you think it makes sense to make actually rich client with these methods? E.g. your client will have methods `get_drives`, `retrieve_all_drives` and such and the connector class itself will contain only the logic to call these methods and create view models, etc

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+            yield drive
+
+    async def retrieve_all_drives(self):
+        """Retrieves all shared drives from Google Drive
+
+        Returns:
+            dict: mapping between drive id and its name
+        """
+        drives = {}
+        async for chunk in self.get_drives():
```
artem-shelkovnikov, line 294:
```suggestion
        async for page in self.get_drives():
```

It's probably page, right?

Chunk is more about files IMO

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+            folders_chunk = chunk.get("files", [])
+            for folder in folders_chunk:
+                folders[folder["id"]] = {
+                    "name": folder["name"],
+                    "parents": folder.get("parents", None),
+                }
+
+        return folders
+
+    async def resolve_paths(self):
```
artem-shelkovnikov, line 337: Can we live without this cache? How much memory will it occupy for large drives?

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+            convert_to_b64,
+            source=temp_file_name,
+        )
+
+        blob_size = os.stat(temp_file_name).st_size
+
+        async with aiofiles.open(file=temp_file_name, mode="r") as target_file:
+            attachment = (await target_file.read()).strip()
+
+        logger.debug(f"Downloaded {blob_name} for {blob_size} bytes ")
```
artem-shelkovnikov, line 405:
```suggestion
        logger.debug(f"Downloaded {blob_name} with size of {blob_size} bytes ")
```

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+                fileId=blob_id,
+                mimeType=GOOGLE_MIME_TYPES_MAPPING[blob_mime_type],
+            ),
+        )
+
+        if blob_size > DEFAULT_FILE_SIZE_LIMIT:
+            logger.warning(
+                f"File size {blob_size} of file {blob_name} is larger than {DEFAULT_FILE_SIZE_LIMIT} bytes. Discarding the file content"
+            )
+            return
```
artem-shelkovnikov, line 448: Is it possible to not even download the file if it's too big?

artem-shelkovnikov (commented):
Great start with a connector! I've left some comments, happy to also jump on a call to discuss any time.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+                        logger.debug(
+                            f"Using the google drive emulator at {GOOGLE_DRIVE_EMULATOR_HOST}"
+                        )
+                        # Redirecting calls to fake Google Drive server for e2e test.
+                        drive_client.discovery_document["rootUrl"] = (
+                            GOOGLE_DRIVE_EMULATOR_HOST + "/"
+                        )
+                    resource_object = getattr(drive_client, resource)
+                    method_object = getattr(resource_object, method)
+                    if full_response:
```
abhishekjoshi-crest, line 115: Can we add separate method to handle paginated calls?

abhishekjoshi-crest (commented):
Great work with path lookup implementation. Just minor suggestion regarding new logger and method separation.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+            # Traverse the parents until reaching the root or a missing parent
+            while parent_id and parent_id in folders:
+                parent_folder = folders[parent_id]
+                path.insert(
+                    0, parent_folder["name"]
+                )  # Insert parent name at the beginning
+                parents = parent_folder["parents"]
+                parent_id = parents[0] if parents else None
+
+            folder["path"] = "/".join(path)  # Join path elements with '/'
```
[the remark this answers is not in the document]
jedrazb: It takes ~30s for 15k folders in multiple drives. Yup, it's blocking as we need to build the path lookup before processing the files.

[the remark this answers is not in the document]
artem-shelkovnikov: You can unblock it with sleeping for 0 seconds every now and then - it'll yield to another task and return back eventually

connectors/sources/google_drive.py
artem-shelkovnikov: Good point, thanks for the explanation! Can you add a code comment about it?

connectors/sources/google_drive.py
jedrazb: Takes `0.5MB` of mem for a setup with 15k folders, multiple drives and 85k files in total.

This cache is needed if we want to have the feature parity with Workplace Search. There is no other way to determine the path on drive. See corresponding implementation in: https://github.com/elastic/ent-search/blob/main/connectors/lib/connectors/content_sources/google_drive/extractor.rb#L61 

One reason for keeping this would be the ability to perform search for a given subpath. On the other hand we could just keep track of `drive_name` where a given file exists (that would be much faster as the lookup will just contain drives).

seanstory: The reason Workplace Search built up these paths was to support filtering rules, IIRC. As is, this would be useful with Basic Sync Rules to help accomplish tasks like "filter all documents in the tree under `/unwanted-dir/`". However, Workplace Search didn't have Advanced Sync Rules, and that's really where we'd want to accomplish this kind of task. If we can limit what trees we access at fetch time, we may not need to build up these paths like this. Might be worth checking with Dana and her plans for Advanced Sync Rules with GDrive.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+                logger.error(
+                    f"Error occurred while generating the resource/method object for an API call. Error: {exception}"
+                )
+                raise exception
+            except HTTPError as exception:
+                log_message = f"Retry count: {retry_counter} out of {self.retry_count}. Response code: {exception.res.status_code} Exception: {exception}."
+                retry_counter = await self._retry(retry_counter, log_message)
+            except Exception as exception:
+                log_message = f"Retry count: {retry_counter} out of {self.retry_count}. Exception: {exception}."
+                retry_counter = await self._retry(retry_counter, log_message)
```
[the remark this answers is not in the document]
jedrazb: The `def _retry(retry_counter, log_message)` performs a condition check if current `retry_counter` is bigger than max retries and if true it would re-raise the exception. 

That being said, the `@retryable` decorator seems to be cleaner approach and I will amend the code to use it.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+        """""Verify the connection with Google Drive""" ""
+        if RUNNING_FTEST:
+            return
+
+        try:
+            await anext(
+                self._google_drive_client.api_call(
+                    resource="about", method="get", fields="kind"
+                )
+            )
```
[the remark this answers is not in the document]
jedrazb: It was needed because `api_call()` was returning async generator for both paginated and non-paginated calls, await anext was get the first (and single) object from the async iterator. 

Following suggestions from other comments I'm splitting the api_call function so that the paginated one will be returning async generator whereas the non-paginated func will be returning awaitable result.

connectors/sources/google_drive.py
jedrazb: From what I understand we can pass a query `q` to as parameter to API call to filter results at fetch time. Here is the documentation for reference:
- https://developers.google.com/drive/api/guides/ref-search-terms 

From what I understand it only supports filtering for shared drive membership, it doesn't support path-based filtering. I will confirm with Dana the right way forward.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+                    if RUNNING_FTEST and not sub_method and GOOGLE_DRIVE_EMULATOR_HOST:
+                        logger.debug(
+                            f"Using the google drive emulator at {GOOGLE_DRIVE_EMULATOR_HOST}"
+                        )
+                        # Redirecting calls to fake Google Drive server for e2e test.
+                        drive_client.discovery_document["rootUrl"] = (
+                            GOOGLE_DRIVE_EMULATOR_HOST + "/"
+                        )
+                    resource_object = getattr(drive_client, resource)
+                    method_object = getattr(resource_object, method)
```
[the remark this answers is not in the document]
jedrazb: because the for e.g. `files/list` endpoint we need to access `drive_client.files.list` method object. 

Since `resource` and `method` are api_call params this is the way to programatically access the object. Note that (resource, method) are different for different api calls, they can be:
- (`files`, `list`)
- (`files`, `get`)
- (`drives`, `list`)
- ...

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+
+        Yields:
+            Dictionary: Response returned by the resource method.
+        """
+        retry_counter = 0
+        while True:
+            try:
+                async with Aiogoogle(
+                    service_account_creds=self.service_account_credentials
+                ) as google_client:
```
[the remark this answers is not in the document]
jedrazb: I tried to have a per-client instance of aiogoogle, but I had some warnings/errors about unclosed client sessions. Also this pattern is suggested on [aiogoogle docs](https://aiogoogle.readthedocs.io/en/latest/)

connectors/sources/google_drive.py
jedrazb: Also worth mentioning is that Google Export endpoint will return `403` codes for exports larger than >5MB (that's also good in preventing large downloads), see https://github.com/elastic/enterprise-search-team/issues/1374

connectors/sources/google_drive.py
danajuratoni: Customers have often mentioned that for file based systems, it's extremely valuable to restrict the ingested scope by sharing only specific folders out of an entire GDrive tree structure or the opposite. 

Basic rules, especially with the current first match logic, are "born" to be used for this type of filtering so it would be awesome if we could leverage this. If the performance impact is not worth the cost, I'm open to give Advanced Sync Rules this responsibility completely.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+                    drive_client = await google_client.discover(
+                        api_name=API_NAME, api_version=API_VERSION
+                    )
+                    if RUNNING_FTEST and not sub_method and GOOGLE_DRIVE_EMULATOR_HOST:
+                        logger.debug(
```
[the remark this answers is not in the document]
jedrazb: Thank you for spotting this. 👍  I will use the new logging implementation.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+            "project_id": "project_id",
+            "private_key_id": "abc",
+            "private_key": open(
+                os.path.abspath(DEFAULT_PEM_FILE)
+            ).read(),  # TODO: change this and provide meaningful defaults
+            "client_email": "123-abc@developer.gserviceaccount.com",
+            "client_id": "123-abc.apps.googleusercontent.com",
+            "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+            "token_uri": "http://localhost:10339/token",
+        }
```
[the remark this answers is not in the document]
jedrazb: Yup just realized we can use `connectors.json` to pass the default configuration for ftests. Agreed that it's pretty much useless 👍

connectors/sources/google_drive.py
jedrazb: `get_pem_format` is actually not needed - I copy pasted this from google cloud storage client. It was used to handle default credentials.

I realised we can use `connecotrs.json` for providing credentials for ftests. 

`get_pem_format` just formats the private key into single line format. I would expect that `aiogoogle` or `google api` will error out if private key is invalid. User needs to only copy-paste the generated json here.

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,638 @@
...
+                )
+            )
+            temp_file_name = async_buffer.name
+
+        await asyncio.to_thread(
+            convert_to_b64,
+            source=temp_file_name,
+        )
+
+        blob_size = os.stat(temp_file_name).st_size
```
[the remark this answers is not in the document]
jedrazb: It's fast, it doesn't read the file to determine its size. It seems to be the preferred way for checking file size.

connectors/sources/google_drive.py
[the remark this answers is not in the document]
jedrazb: I timed the blocking part of `resolve_paths()` for 15k folders, with:
```
--- a/connectors/sources/google_drive.py
+++ b/connectors/sources/google_drive.py
@@ -326,6 +326,8 @@ class GoogleDriveDataSource(BaseDataSource):
         folders = await self.retrieve_all_folders()
         drives = await self.retrieve_all_drives()
 
+        import time
+        start_time = time.time()
         # for paths let's treat drives as top level folders
         for id, drive_name in drives.items():
             folders[id] = {"name": drive_name, "parents": []}
@@ -349,6 +351,10 @@ class GoogleDriveDataSource(BaseDataSource):
 
             folder["path"] = "/".join(path)  # Join path elements with '/'
 
+        end_time = time.time()
+
+        self._logger.info(f"Resolving paths took: {end_time - start_time}s")
+
         return folders

```

Here is the result:
```
[...] Resolving paths took: 0.03356814384460449s
```

Do you think I should still introduce periodic `asyncio.sleep(0)` calls?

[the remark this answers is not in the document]
artem-shelkovnikov: I think it's good, thanks for checking :)

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,626 @@
...
+
+        async with aiofiles.open(file=temp_file_name, mode="r") as target_file:
+            attachment = (await target_file.read()).strip()
+
+        self._logger.debug(
+            f"Downloaded {blob_name} with the size of {blob_size} bytes "
+        )
+
+        await remove(str(temp_file_name))
```
wangch079, line 391: Personally, I'm not a fan of temp files. In the [official guide](https://developers.google.com/drive/api/guides/manage-downloads#python), there is an example of using IO streaming. Have you considered that approach?

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,626 @@
...
+
+        Returns:
+            dict: mapping between folder id and its (name, parents, path)
+        """
+        folders = await self.retrieve_all_folders()
+        drives = await self.retrieve_all_drives()
+
+        # for paths let's treat drives as top level folders
+        for id, drive_name in drives.items():
+            folders[id] = {"name": drive_name, "parents": []}
```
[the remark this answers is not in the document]
jedrazb: Thank you for your feedback, the parents can also point to drives (from my testing).  This approach is consistent with what we have in `ent-search` ([link](https://github.com/elastic/ent-search/blob/main/connectors/lib/connectors/content_sources/google_drive/extractor.rb#L66))

Here is a result from internal testing:
Screenshot 2023-07-05 at 10 04 19

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,626 @@
...
+        blob_size = int(blob["size"])
+
+        if blob_size == 0:
+            return
+
+        blob_name, blob_id, blob_extension = (
+            blob["name"],
+            blob["id"],
+            f".{blob['file_extension']}",
+        )
```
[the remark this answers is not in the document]
jedrazb: The framework pops `_id` from referenced blob and sets  its value to `id` key, this is executed before calling `get_content`. I don't have access to `_id` key at this step.

https://github.com/elastic/connectors-python/blob/main/connectors/es/sink.py#L392

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,626 @@
...
+
+        Args:
+            blob (dict): Formatted blob document.
+            timestamp (timestamp, optional): Timestamp of blob last modified. Defaults to None.
+
+        Returns:
+            dict: Content document with id, timestamp & text
+        """
+
+        blob_name, blob_id, blob_mime_type = blob["name"], blob["id"], blob["mime_type"]
```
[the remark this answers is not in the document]
jedrazb: The framework pops `_id` from referenced blob and sets  its value to `id` key, this is executed before calling `get_content`. I don't have access to `_id` key at this step.

https://github.com/elastic/connectors-python/blob/main/connectors/es/sink.py#L392

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,626 @@
...
+
+        for folder in folders.values():
+            path = [folder["name"]]  # Start with the folder name
+
+            parents = folder["parents"]
+            parent_id = parents[0] if parents else None
+
+            # Traverse the parents until reaching the root or a missing parent
+            while parent_id and parent_id in folders:
+                parent_folder = folders[parent_id]
```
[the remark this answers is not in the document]
jedrazb: Good catch, thank you! Will add the improvement :)

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,626 @@
...
+            blob_document["updated_by_email"] = last_modifying_user.get(
+                "emailAddress", None
+            )
+            blob_document["updated_by_photo_url"] = last_modifying_user.get(
+                "photoLink", None
+            )
+
+        # handle folders and shortcuts
+        if blob_document["size"] is None:
+            blob_document["size"] = 0
```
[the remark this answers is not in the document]
jedrazb: We are expecting the key `size` to be present and sometime it point to `None` value, afaik the `get()` method with default value `0` will still return `None` for keys that are present in the dict.

I did a small test:

```
>>> test_blob
{'size': None}
>>> size = test_blob.get('size', 0)
>>> print(size)
None
```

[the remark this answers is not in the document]
wangch079: That's a valid point. Maybe:
```python
size = test_blob.get('size') or 0
```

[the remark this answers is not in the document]
jedrazb: Yeah that's neat! Will include the suggestion

connectors/sources/google_drive.py
```diff
@@ -0,0 +1,670 @@
...
+        if blob_size > FILE_SIZE_LIMIT:
+            self._logger.warning(
+                f"File size {blob_size} of file {blob_name} is larger than {FILE_SIZE_LIMIT} bytes. Discarding the file content"
+            )
+            return
+
+        document = {
+            "_id": blob_id,
+            "_timestamp": blob["_timestamp"],
+        }
```
wangch079, line 521: This will actually be removed here:
https://github.com/elastic/connectors-python/blob/7fbb60ea586c95af4419659b8c25e28d5d022ac2/connectors/es/sink.py#L327-L328

I see the same pattern in other source implementations. Is it intended? cc @elastic/ingestion-team

connectors/sources/google_drive.py
wangch079: Not a blocker. We can revisit later.

connectors/sources/google_drive.py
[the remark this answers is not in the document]
artem-shelkovnikov: I think it's some legacy logic. I'm honestly not a big fan of the fact that once you `yield` something, the object is modified. Let's address this at some point in future

connectors/sources/google_drive.py
artem-shelkovnikov: Yeah afaik it's intended - @tarekziade originally implemented it this way. 

We pop `ids` always because `id` is not a part of the document - it's a part of document metadata. Why `_timestamp` is popped I don't remember. Let's create an issue to address it?

artem-shelkovnikov (approved):

connectors/sources/google_drive.py
jedrazb: We decided to keep this logic, the mem footprint is minimal even for large drives (85k files). The blocking part of the function for large drives takes `~0.3s` and this is acceptable.

connectors/sources/google_drive.py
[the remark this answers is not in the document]
jedrazb: https://github.com/elastic/enterprise-search-team/issues/5095

connectors/sources/google_drive.py
jedrazb: https://github.com/elastic/enterprise-search-team/issues/5095

wangch079: What I meant, it's not necessary to include `_id` and `_timestamp` here. It will eventually be removed from here and not used at all:
https://github.com/elastic/connectors-python/blob/7fbb60ea586c95af4419659b8c25e28d5d022ac2/connectors/es/sink.py#L327-L328

For the `lazy_download` function, it should be enough to just return a document with key `_attachment`.

But this should not block this PR.

wangch079 (approved):
Great work Jedr

PR #1132, opened 2023-06-27
Change Subextract -> Extract for SPO Advanced Filtering

artem-shelkovnikov:
Minor rename after discussion with @danajuratoni - renaming the Advanced Filtering rule to "dontExtractDriveItemsOlderThan"


vidok (approved):

timgrein:
Should this be backported to 8.9?

artem-shelkovnikov:
Good call, yes it should be, thanks Tim!

PR #1135, opened 2023-06-27
Increase SPO connector test coverage

artem-shelkovnikov:
Increases test coverage of `connectors/sources/sharepoint_online.py` from 88% to 93% - this should be sufficient as we have a 92% recommended threshold for test coverage.

Additionally:
- renamed `scroll_url` to `scroll_delta_url` as it's logically attached to delta api
- moved `set_dls_enabled` under `TestSharepointOnlineDataSource` as it's only used there


tests/sources/test_sharepoint_online.py
```diff
@@ -548,10 +707,53 @@ async def test_site_drives(self, client, patch_scroll):
...
+    async def test_drive_items_delta(self, client, patch_fetch, patch_scroll_delta_url):
+        delta_url_input = "https://sharepoint.com/delta-link-lalal"
+        delta_url_next_sync = "https://sharepoint.com/delta-link-lalal/page-2"
+        items_page_1 = ["1", "2"]
+        items_page_2 = ["3", "4"]
 
+        patch_scroll_delta_url.return_value = AsyncIterator(
+            [
+                {"@odata.deltaLink": delta_url_next_sync, "value": items_page_1},
+                {"@odata.deltaLink": delta_url_next_sync, "value": items_page_2},
```
seanstory, line 719: this uses `delta_url_next_sync` for both items, but that variable should only be for page 2, right?

seanstory (dismissed):
lgtm. Left 1 comment that I'm pretty sure is just me being confused.

tests/sources/test_sharepoint_online.py
artem-shelkovnikov: Good one! Tbh I thought it's always returned - @vidok can you help us on that?

vidok: Good catch! There should be `@odata.nextLink` and no `@odata.deltaLink` in the first item.

vidok (approved):

Issue #1136, opened 2023-06-28
[SharePoint Server]Add remote validation for site collection


akanshi-elastic:
### Problem Description

SharePoint server connector doesn't validate the invalid collection while running the code, Also currently the code raise the exception after 3 retries, Which stops the connector to index data in elastic. So instead of raising the exception, we can skip that.
### Proposed Solution
1. Add remote validation for the verifying invalid site collection and raise if there is any collection is not valid
2. Skip the exception after 3 retries.

PR #1137, opened 2023-06-28
[SharePoint Server]Add remote validation for site collection

akanshi-elastic:
## Closes https://github.com/elastic/connectors-python/issues/1136 ###
This PR includes:
1. Verify all invalid collections and raise it  in remote validation
2. Skip the exception after 3 retries instead of raising it.

#### Pre-Review Checklist


connectors/sources/sharepoint_server.py
```diff
@@ -295,26 +295,6 @@ async def get_site_pages_content(
             "_attachment": await self.convert_file_to_b64(source_file_name),
         }
 
-    def _get_retry_after(self, retry, exception):
```
artem-shelkovnikov, line 298: Why is the method deleted?

connectors/sources/sharepoint_server.py
```diff
@@ -375,28 +358,25 @@ async def _fetch_data_with_next_url(
...
         while True:
             if next_url != "":
-                response = await anext(
-                    self._api_call(
-                        url_name=param_name,
-                        url=next_url,
-                        selected_field=selected_field,
-                    )
-                )
+                async for response in self._api_call(
```
artem-shelkovnikov, line 361: Why `async for`? Is result of `_api_call` a collection?

connectors/sources/sharepoint_server.py
```diff
@@ -375,28 +358,25 @@ async def _fetch_data_with_next_url(
...
+                    url_name=param_name,
+                    url=next_url,
+                ):
+                    yield response.get("value", [])  # pyright: ignore
+
+                    next_url = response.get("odata.nextLink", "")  # pyright: ignore
```
artem-shelkovnikov, line 367: What are you ignoring here?

connectors/sources/sharepoint_server.py
```diff
@@ -410,20 +390,19 @@ async def _fetch_data_with_query(self, site_url, param_name):
             Response of the GET call.
         """
         skip = 0
+        response_result = []
```
artem-shelkovnikov, line 393: Why declare it here since you only do re-assigning on L402?

connectors/sources/sharepoint_server.py
```diff
@@ -668,6 +648,33 @@ async def close(self):
...
+                site_collections=collection,
+                host_url=self.sharepoint_client.host_url,
+            ):
+                is_invalid = False
+            if is_invalid:
+                self.invalid_collections.append(collection)
+        if self.invalid_collections:
+            raise ConfigurableFieldValueError(
+                f"Collections {', '.join(self.invalid_collections)} are not available."
+            )
```
artem-shelkovnikov, line 670: Can you fetch all collections in one go and then compare?

E.g.:

```
user_defined_collections = self.config['site_collections']
site_collections_on_server = await self.sharepoint_client.get_all_site_collections()

# Then compare user_defined_collections against site_collections_on_server by name

```

connectors/sources/sharepoint_server.py
akanshi-elastic: In this PR we have remove the logic to raise the exception after 3 retries, instead of raising it now we are skipping that call. And to achieve this we need to add `break`  inside the condition `if retry > self.retry_count:`. 
to add `break` loop is required and that's why we have removed this method and directly use it in `_api_call` to skip the call.

connectors/sources/sharepoint_server.py
akanshi-elastic: On line 406 we are checking the length of response result for peroforming pagination over this, and if we remove from here then line 406 will give `response_result is not defined` error in case line 402 doesn't execute.

connectors/sources/sharepoint_server.py
akanshi-elastic: The expected for _api_call is collection only, but if I use anext then I have to handle `StopAsyncIteration` exception explicitly, that's why I use async for here.

connectors/sources/sharepoint_server.py
akanshi-elastic: There is no such api available which returns list of all collections.

connectors/sources/sharepoint_server.py
akanshi-elastic: not required now

connectors/sources/sharepoint_server.py
artem-shelkovnikov: I see, ideally you should refactor _api_call to two methods: one for a single entity, one for the collection. This is outside of the scope of this PR, however

artem-shelkovnikov (approved):

artem-shelkovnikov:
@akanshi-elastic do you want to backport to 8.9 as well?

akanshi-elastic:
> @akanshi-elastic do you want to backport to 8.9 as well?

The code is well tested so we can add this in 8.9 as well if feature isn't freeze yet

akanshi-elastic:
Worked on adding remote validation

PR #1140, opened 2023-06-28
Use a global variable to hold cli loaded config

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/5005

The ExtractionService had hardcoded to look in `connectors-python/config.yml` for its configuration. But our CLI allows users to specify non-default config files with a `-c /path/to/other/config.yml`. 

This PR changes `cli.py` to call a new function in `config.py` which sets a new global variable when loading the user-specified (or implied) config file. Now, any code that wants to read configurations from our YAML don't need them passed as params, but can access them with `connectors.config.get_framework_config()`.

`ExtractionService` and its tests were then refactored to do just this.

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -19,12 +19,12 @@ import signal
 from argparse import ArgumentParser
 
 from connectors import __version__
-from connectors.config import load_config, set_framework_config
+from connectors.config import load_config
 from connectors.logger import logger, set_logger
 from connectors.preflight_check import PreflightCheck
 from connectors.services import get_services
 from connectors.source import get_source_klass, get_source_klasses
-from connectors.utils import get_event_loop
+from connectors.utils import ExtractionService, get_event_loop
 
 __all__ = ["main"]
 
@@ -140,7 +140,9 @@ def run(args):
     config = {}
     try:
         config = load_config(args.config_file)
-        set_framework_config(config)
+        ExtractionService.set_extraction_config(
+            config.get("extraction_service", None)
+        )  # Not perfect, let's revisit
     except Exception as e:
         # If something goes wrong while parsing config file, we still want
         # to set up the logger so that Cloud deployments report errors to
--- a/connectors/cli.py
+++ b/connectors/cli.py
@@ -19,7 +19,7 @@ import signal
 from argparse import ArgumentParser
 
 from connectors import __version__
-from connectors.config import load_config
+from connectors.config import load_config, set_framework_config
 from connectors.logger import logger, set_logger
 from connectors.preflight_check import PreflightCheck
 from connectors.services import get_services
@@ -140,6 +140,7 @@ def run(args):
     config = {}
     try:
         config = load_config(args.config_file)
+        set_framework_config(config)
     except Exception as e:
         # If something goes wrong while parsing config file, we still want
         # to set up the logger so that Cloud deployments report errors to
--- a/tests/sources/test_sharepoint_online.py
+++ b/tests/sources/test_sharepoint_online.py
@@ -2300,8 +2300,11 @@ class TestSharepointOnlineDataSource:
             ("", None),
             ("abc|", None),
             ("abc|def|", None),
+            ("abc|def|_o", None),
             (f"abc|def|{DOMAIN_GROUP_ID}", DOMAIN_GROUP_ID),
+            (f"abc|def|{DOMAIN_GROUP_ID}_o", DOMAIN_GROUP_ID),
             (f"abc|def|ghi/{DOMAIN_GROUP_ID}", DOMAIN_GROUP_ID),
+            (f"abc|def|ghi/{DOMAIN_GROUP_ID}_o", DOMAIN_GROUP_ID),
         ],
     )
     def test_domain_group_id(self, user_info_name, expected_domain_group_id):
--- a/tests/sources/test_sharepoint_online.py
+++ b/tests/sources/test_sharepoint_online.py
@@ -1662,27 +1662,24 @@ class TestSharepointOnlineDataSource:
 
         with patch(
             "connectors.utils.ExtractionService.extract_text", return_value=message
-        ) as extraction_service_mock:
-            with patch(
-                "connectors.sources.sharepoint_online.get_framework_config",
-                return_value={"extraction_service": {"host": "http://localhost:8090"}},
-            ):
+        ) as extraction_service_mock, patch(
+            "connectors.utils.ExtractionService.get_extraction_config",
+            return_value={"host": "http://localhost:8090"},
+        ):
 
-                async def download_func(attachment_id, async_buffer):
-                    await async_buffer.write(bytes(message, "utf-8"))
+            async def download_func(attachment_id, async_buffer):
+                await async_buffer.write(bytes(message, "utf-8"))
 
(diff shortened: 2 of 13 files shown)


wangch079 (commented):
If we decided to make the change, should we do it as I proposed here: https://github.com/elastic/connectors-python/pull/1096#discussion_r1236898591?

IMO, `utils` should be fully pluggable, and should not have any hard dependency on the framework (the change in this PR is reading the global variable from the framework)

We also avoided having a global config, and you can see `config` are passed around. I think the problem now is that `ExtractionService` is used in `SharepointOnlineDataSource`, where config is passed. Do you think `ExtractionService` should be used inside the `Extractor` class?

seanstory:
> If we decided to make the change, should we do it as I proposed here: https://github.com/elastic/connectors-python/pull/1096#discussion_r1236898591?

I'd missed that thread. I do think that this class is a "utility class" and therefore makes the most sense under utils.py. 

> IMO, utils should be fully pluggable, and should not have any hard dependency on the framework

This I do agree with though. I'll work to refactor this to let ExtractionService take a dict as a constructor param, and then let whatever initializes it be the thing that's tied to the framework.

> the change in this PR is reading the global variable from the framework. We also avoided having a global config, and you can see config are passed around

I'm conflicted about this. On one hand, global variables are (typically) bad. On the other hand, they seem like a really common pattern when it comes to configuration. Like in ent-search `SharedTogo.config` is accessible from anywhere - and that's _very_ convenient. WDYT?

> I think the problem now is that ExtractionService is used in SharepointOnlineDataSource, where config is passed. Do you think ExtractionService should be used inside the Extractor class?

I'm less concerned about this. Since we're past FF and this isn't part of the "bug", I don't plan to change this, but I do expect we'll revisit this in 8.10 as other connectors get access to this feature.

tests/sources/test_sharepoint_online.py
```diff
@@ -1663,20 +1663,26 @@ async def test_get_attachment_with_text_extraction_enabled_adds_body(
         with patch(
             "connectors.utils.ExtractionService.extract_text", return_value=message
         ) as extraction_service_mock:
+            with patch(
```
seanstory, line 1666: This file is easier to review if you hide whitespace changes.

tests/test_utils.py
```diff
@@ -617,10 +617,10 @@ async def mock_responses(self):
         ],
     )
     def test_check_configured(self, mock_config, expected_result):
-        with patch("yaml.safe_load") as mock_safe_load:
```
seanstory, line 620: This file is easier to review if you hide whitespace changes

tests/sources/test_sharepoint_online.py
artem-shelkovnikov: You can actually do both patches in one `with`: https://stackoverflow.com/a/73884949

connectors/cli.py
```diff
@@ -140,6 +140,7 @@ def run(args):
     config = {}
     try:
         config = load_config(args.config_file)
+        set_framework_config(config)
```
artem-shelkovnikov, line 143: I would avoid this at all costs - having globally exposed config will provoke things that we're heavily trying to avoid - such as connector having access to host configuration, which in turn can lead to config files being used for configuration of individual connectors.

Would it make more sense to do something like this:

1. Make ExtractionService have a class variable `configuration` that will store extraction service config. It's not initialized by default and if ExtractionService is created when no config is provided, then it won't work.

2. Add a class method to update the configuration and use it here, e.g.:
```
class ExtractionService:
    configuration = None
    
    @classmethod
    def update_configuration(class, config):
         class.configuration = config # do validity checks here
    
    def __init__(self):
        self.instance_configuration = type(self).configuration # we treat class as a factory

```

This way config does not leak outside globally + we have a working solution that can easily be changed to anything else.

artem-shelkovnikov:
> Do you think ExtractionService should be used inside the Extractor class?

I think it's our logical next step - simplify the `download` functions in the connectors + make usage of extraction service a part of the framework. It should stop being configuration field and become a part of overall connector setting (if needed) or just be handled transparently by the framework.

connectors/cli.py
wangch079: Why not simply use param `config` in `__init__`?

seanstory: >  having globally exposed config will provoke things that we're heavily trying to avoid - such as connector having access to host configuration, which in turn can lead to config files being used for configuration of individual connectors

Why are we trying to avoid that at all costs? We just talked in a sync yesterday about how some individual connector configs should be configurable through the YAML file. I know there was some disagreement on what configs would be a good fit, but I thought there was alignment that this should be possible?

> Make ExtractionService have a class variable configuration that will store extraction service config

If I understand this correctly, the suggestion is still to use a global variable, but to just put it in a place where it's less obviously useful to other code elsewhere? I do not understand the motivation for making it hard for any given code to access our configuration values.

seanstory: > Why not simply use param config in __init__?

I looked at that. But then I'd have to change a ton of method signatures to pass `config` (not just `es_config`) all the way down to whatever is initializing ExtractionService (in this case, `SharepointOnlineDataSource`). 

I'd prefer to be able to look up my configs instead of having to pass them around.

wangch079 (approved):
> I looked at that. But then I'd have to change a ton of method signatures to pass `config` (not just `es_config`) all the way down to whatever is initializing ExtractionService (in this case, `SharepointOnlineDataSource`).

Agree. I don't think we can implement it (including moving `ExtractionService` one level up to `Extractor`) properly after FF. As the purpose of this PR is to load the extraction service config to a global config so that the extraction service in SPO data source can access it, I think this is an acceptable solution for now. cc @artem-shelkovnikov

artem-shelkovnikov:
I'm okay to merge it like this if we address this ASAP after merge, otherwise I'd recommend doing a different solution.

> Why are we trying to avoid that at all costs? We just talked in a sync yesterday about how some individual connector configs should be configurable through the YAML file.

I don't think any connector-specific config should go to the configuration file - we're a distributed system and aim to stay a distributed system. Imagine changing "http.timeout" for one connector on one host, and all other hosts will not be able to sync if 3rd-party system is slow. That's why I oppose this change. It's making some things easier in short term and has a potential of making refactoring/changing things in app much worse in medium-long term.

seanstory:
> I'm okay to merge it like this if we address this ASAP after merge, otherwise I'd recommend doing a different solution.

Artem and I had a chance to briefly sync. While I'm not sure I agree, it became clear to me that there are concerns that by exposing access to the full config across the whole codebase we'll encourage bad patterns. This PR definitely could have far-reaching consequences as is, and it's probably not fair for me to sneak this in after FF without a team discussion.


I'm going to try to refactor to limit the configuration exposure for now, to just fix this bug. The fix will not be perfect, but we know we'll be changing how ExtractionService is used in future releases, so it will be a temporary stopgap.

We can revisit later, as a team, if we should make config more globally accessible.

seanstory:
This should be ready for re-review.

artem-shelkovnikov (approved):
Thank you for taking care of it, Sean!

As discussed, there's no fast and good solution now, we'll reiterate on it later.

Issue #1111, opened 2023-06-23
MSSQL connector not able to verify a valid CA cert


danajuratoni:
## Bug Description
We tested the connection outside the connection outside of our MSSQL Connector using Python directly and it worked.  This is for our own test environment (not customer) so OK to use the credentials and cert file for reproduction.  The cert file used is linked below (it's the same file, except that I have renamed the extension before attaching to this issue).  The full error stack is included in the gist above.  Here's an excerpt.

```
[FMWK][12:49:03][WARNING] Something went wrong while executing query. Exception: [('SSL routines', '', 'certificate verify failed')]
[FMWK][12:49:03][WARNING] Retry count: 1 out of 3. Exception: [('SSL routines', '', 'certificate verify failed')]
[FMWK][12:49:06][WARNING] Something went wrong while executing query. Exception: [('SSL routines', '', 'certificate verify failed')]
[FMWK][12:49:06][WARNING] Retry count: 2 out of 3. Exception: [('SSL routines', '', 'certificate verify failed')]
[FMWK][12:49:11][WARNING] Something went wrong while executing query. Exception: [('SSL routines', '', 'certificate verify failed')]
[FMWK][12:49:11][WARNING] Retry count: 3 out of 3. Exception: [('SSL routines', '', 'certificate verify failed')]
[FMWK][12:49:11][CRITICAL] Can't connect to Microsoft SQL on elastic-sql-connector.database.windows.net
```

- When configuring this using MSSQL connector client using the contents of the [cert file](https://github.com/elastic/sdh-enterprise-search/files/11840263/yourdomain.txt), it doesn’t work so I don’t think the certificate itself is invalid, but it’s likely due to how we are loading the certificate from our configuration (perhaps we are requiring some specific manipulation of the content of the cert before it can be added as the configuration value for ssl_ca from the UI)?
- The following is the cert entry from the elastic connectors index for the configuration:

REDACTED. See the slack thread: https://elastic.slack.com/archives/C02AV5C8ZDE/p1687535501711019?thread_ts=1687519055.139339&cid=C02AV5C8ZDE

I suspect somewhere between adding the cert content to the UI and how we are storing and retrieving the cert content to pass into the Python calls is causing the cert to become invalid.  Is this a potential bug with the cert being a combined cert with a chain?

prashant-elastic:
[...]

PR #1151, opened 2023-06-30
[MS-SQL] Add Support for multiple SSL Certificates

moxarth-elastic:
## Closes https://github.com/elastic/connectors-python/issues/1111

This PR includes:
- Add multiple certificates in `get_pem_format`
- Add condition to avoid unnecessary logger while connector founds tables in database.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/generic_database.py
+++ b/connectors/sources/generic_database.py
@@ -447,9 +447,10 @@ class GenericBaseDataSource(BaseDataSource):
         """
         tables_to_fetch = await self.get_tables_to_fetch(schema)
         if self.database in self.tables_to_skip.keys():
-            tables_to_fetch = [
-                *(set(tables_to_fetch) - set(self.tables_to_skip[self.database]))
-            ]
+            tables_to_fetch = set(tables_to_fetch) - set(
+                self.tables_to_skip[self.database]
+            )
+        tables_to_fetch = list(tables_to_fetch)
         for table in tables_to_fetch:
             self._logger.debug(f"Found table: {table} in database: {self.database}.")
             async for row in self.fetch_documents(
@@ -457,7 +458,7 @@ class GenericBaseDataSource(BaseDataSource):
                 schema=schema,
             ):
                 yield row
-        else:
+        if len(tables_to_fetch) < 1:
             if schema:
                 self._logger.warning(
                     f"Fetched 0 tables for schema: {schema} and database: {self.database}"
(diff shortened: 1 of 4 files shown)


connectors/sources/generic_database.py
```diff
@@ -447,17 +447,18 @@ async def fetch_rows(self, schema=None):
         """
         tables_to_fetch = await self.get_tables_to_fetch(schema)
         if self.database in self.tables_to_skip.keys():
-            tables_to_fetch = [
-                *(set(tables_to_fetch) - set(self.tables_to_skip[self.database]))
-            ]
+            tables_to_fetch = set(tables_to_fetch) - set(
+                self.tables_to_skip[self.database]
+            )
+        tables_to_fetch = list(tables_to_fetch)
```
artem-shelkovnikov, line 453: Why change this?

connectors/sources/mssql.py
```diff
@@ -143,9 +142,28 @@ async def close(self):
...
+            key = key.replace(" ", "\n")
+            key = " ".join(key.split("\n", max_split))
+            key = " ".join(key.rsplit("\n", max_split))
+            all_cert += key + "\n"
+        return all_cert
+
     def create_pem_file(self):
         """Create pem file for SSL Verification"""
-        pem_format = get_pem_format(key=self.ssl_ca, max_split=1)
+        pem_format = self.get_pem_format(keys=self.ssl_ca, max_split=1)
```
artem-shelkovnikov, line 166:
```suggestion
        pem_certificates = self.get_pem_format(keys=self.ssl_ca, max_split=1)
```

connectors/sources/mssql.py
```diff
@@ -143,9 +142,28 @@ async def close(self):
             return
         self.connection.close()
 
+    def get_pem_format(self, keys, max_split=-1):
```
artem-shelkovnikov, line 145: Why create a new function instead of updating the old one?

moxarth-elastic: As customers are waiting for MS-SQL, we have updated it at the connector level so other connectors do not get affected. Also, we are working on it. Once testing is done for all the connectors, we will raise a follow-up PR in which we move these changes to utils and remove them from MS-SQL.

connectors/sources/generic_database.py
moxarth-elastic: As type of `tables_to_fetch` is map, we need to convert it to list. also to remove multiple type conversion we remove it from the L450

artem-shelkovnikov (approved):

PR #1156, opened 2023-06-30
[SPO] Update DLS implementation 

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/4306

This PR updates the SPO DLS implementation to _only_ fetch access control for sharepoint sites. This access control is then propagated down the tree, meaning child items have the same access control as their (transitive) parent site.

Also add a guard to check, if a document is `None`, which can fail the ingestion inside `sink`.
Remove all the access control calls for anything except a site.
#### Pre-Review Checklist

## Release Note

DLS on site level for Sharepoint Online connector.


connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
-        if self._features is None:
-            return False
+        email = user.get("EMail", user.get("mail", None))
+        username = user[username_field]
 
-        if not self._features.document_level_security_enabled():
-            return False
+        prefixed_mail = _prefix_email(email)
+        prefixed_username = _prefix_user(username)
+        id_ = email if email else username
```
seanstory, line 1174: This makes perfect sense to me, but have you communicated this with Ioana?

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
-            "_id": user.get("Id"),
+            # Here we're intentionally using the email/username without the prefix
+            "_id": id_,
             "identity": {
-                "email": _prefix_email(user.get("Email")),
-                "username": _prefix_user(user.get("LoginName")),
+                "email": prefixed_mail,
+                "username": prefixed_username,
             },
+            "_timestamp": timestamp,
```
seanstory, line 1190: We're also expecting a `created_at` field: https://github.com/elastic/ent-search/blob/main/shared_togo/lib/shared_togo/ingestion/access_control_filter_v1.rb#L21

connectors/es/sink.py
```diff
@@ -546,6 +546,10 @@ async def get_access_control_docs(self, generator):
         try:
             async for doc in generator:
                 doc, _, _ = doc
+
+                if doc is None:
+                    continue
```
artem-shelkovnikov, line 551: Should it be here? We should not yield None IMO

connectors/sources/sharepoint_online.py
```diff
@@ -748,6 +798,80 @@ def _prefix_email(email):
...
 
 
+def _postfix_group(group):
+    if group is None:
+        return None
+
+    return f"{group} Members"
+
+
+def is_domain_group(user_fields):
```
artem-shelkovnikov, line 808: Do they need to be in the module?

I see these are functions that are pure, which is nice, but idk if they should be here. I'm also not a Python expert and not sure what are best practices here.

connectors/sources/sharepoint_online.py
```diff
@@ -748,6 +798,80 @@ def _prefix_email(email):
...
+    if len(name_parts) <= 2:
+        return None
+
+    domain_group_id = name_parts[2]
+
+    if len(domain_group_id) == 0:
+        return None
+
+    if "/" in domain_group_id:
+        domain_group_id = domain_group_id.split("/")[1]
```
artem-shelkovnikov, line 841: Could it be that there are multiple slashes here?

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
-        return {"query": {"template": {"params": {"access_control": access_control}}}}
+        if "Modified" in user:
+            timestamp = datetime.strptime(user["Modified"], TIMESTAMP_FORMAT)
+        else:
+            timestamp = iso_utc()
 
-    def _user_access_control_doc(self, user, access_control):
         return {
-            "_id": user.get("Id"),
+            # Here we're intentionally using the email/username without the prefix
```
artem-shelkovnikov, line 1184: why intentionally though? What's the use-case?

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
+                "username": prefixed_username,
             },
+            "_timestamp": timestamp,
         } | self.access_control_query(access_control)
 
     async def get_access_control(self):
+        """Yields an access control document for every user of a site.
+        Note: this method can sometimes fetch some groups and therefore some users multiple times for different sites.
+        This can lead to a higher "indexed" number than actual documents present in the index (documents with the same id won't be indexed multiple times).
+        This is expected as it would otherwise overload memory keeping a set of all users/groups already seen for big Sharepoint Online instances.
```
artem-shelkovnikov, line 1197: Probably would be worth to address this in UI at some point - I'm already constantly confused with number of docs in the index not being equal to number of docs ingested, etc. cc @danajuratoni

connectors/sources/sharepoint_online.py
```diff
@@ -748,6 +798,80 @@ def _prefix_email(email):
...
+
+    if "/" in domain_group_id:
+        domain_group_id = domain_group_id.split("/")[1]
+
+    return domain_group_id
+
+
+async def _emails_and_usernames_of_domain_group(
+    domain_group_id, group_identities_generator, only_usernames=False, prefix=True
+):
```
artem-shelkovnikov, line 848: This method specifically looks like instance-specific cause `generator` is actually doing calls to server. I think this method should not be separated from the instance who uses it so much.

connectors/sources/sharepoint_online.py
```diff
@@ -1148,9 +1254,10 @@ async def get_docs(self, filtering=None):
                 site_collection["siteCollection"]["hostname"],
                 self.configuration["site_collections"],
             ):
-                yield site, None
+                access_control = await self._site_access_control(site)
+                yield self._decorate_with_access_control(site, access_control), None
```
artem-shelkovnikov, line 1258: So in essence every document within site is gonna have same permissions?

Out of curiosity, do we know what's the storage impact of this?

artem-shelkovnikov (commented):
Looks good, left a couple questions/comments

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
-            }
+                    if len(group) > 0:
+                        access_control.add(_prefix_group(group["id"]))
 
-        We're extracting "loginName" or "email" for groups or users inside "grantedToV2" or "grantedTo".
-        """
+                    async for member_email_or_username in _emails_and_usernames_of_domain_group(
+                        domain_group_id, self.client.group_members
+                    ):
+                        access_control.add(member_email_or_username)
```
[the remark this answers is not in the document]
seanstory: After talking, instead of holding these in an in-memory cache - just remove this logic for getting the users from groups. We'll add the explicit groups to `_allow_access_control`, and the explicit users to `_allow_access_control`, but not the implicit users.

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
 
-        access_control = await self._access_control_for_role_assignments(
-            site_web_url, role_assignments
-        )
+        return list(access_control)
 
-        return self._decorate_with_access_control(site_list, access_control)
+    def _dls_enabled(self):
+        if self._features is None:
+            return False
```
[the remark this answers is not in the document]
seanstory: Or might just be that git is confused. Maybe ignore.

connectors/sources/sharepoint_online.py
seanstory: also just may want to remove `_timestamp` since the ACL index doesn't _index_ that field, so it won't be searchable.

connectors/es/sink.py
timgrein: I think it should be there as otherwise every connector implementation needs to never yield `None`, which is something out of the framework's control. Maybe log a warning/debug message? WDYT?

connectors/sources/sharepoint_online.py
timgrein: I think they should as they're specific to SPO. That's also a suggestion by the IDE to create a top-level function, if it's a pure function.

connectors/sources/sharepoint_online.py
timgrein: Didn't see it as of now, but I'll note that down for testing

connectors/sources/sharepoint_online.py
timgrein: We want to query by emails/usernames of AppSearch users, which are not prefixed.

connectors/sources/sharepoint_online.py
timgrein: Anyway addressed by caching emails/usernames, which were already seen -> reduces ingest time and number of calls to ES.

connectors/sources/sharepoint_online.py
timgrein: Per definition it's not instance-specific as it misses the `self` reference, but agree that it looks like one. I can also move it down again, if you prefer so.

connectors/sources/sharepoint_online.py
timgrein: Yes, it's reduced now as we're only adding groups and not individual emails/usernames (only if a user is a person, lol SPO?), if the same information can be derived from an identity being part of a group.

connectors/sources/sharepoint_online.py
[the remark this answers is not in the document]
timgrein: Weird diff

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
-                )
-
-                yield self._user_access_control_doc(user, access_control)
+            async for site in self.client.sites(
+                site_collection["siteCollection"]["hostname"],
+                self.configuration["site_collections"],
+            ):
+                async for user_information in self.client.user_information_list(
+                    site["id"]
+                ):
```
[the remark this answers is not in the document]
timgrein: Implemented the caching 👍

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,240 @@ async def validate_config(self):
...
+                if "EMail" not in user:
+                    continue
 
-        return self._decorate_with_access_control(drive_item, access_control)
+                email = user["EMail"]
 
-    async def _with_site_list_access_control(self, site_web_url, site_list):
-        if not self._dls_enabled():
-            return site_list
+                if email in access_control:
```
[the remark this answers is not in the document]
timgrein: Yes

connectors/es/sink.py
artem-shelkovnikov: `yield None` is weird cause also you yield multiple things: `yield None, download_func, DELETE` - that's also weird. I'd honestly just disable it by default, connectors should do this check themselves.

Where did this situation happen for you in current work?

timgrein: Ran into it, when some SPO system user was yielded. Will add the guards to SPO and remove it from `sink`

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +1003,274 @@ async def validate_config(self):
...
-                        sharepoint_groups,
-                    ),
-                ),
-            )
-        )
+                if domain_group_id:
+                    group = await self.client.group(domain_group_id)
+
+                    if len(group) > 0:
+                        access_control.add(_prefix_group(group["id"]))
```
seanstory, line 1043: Is this redundant? If you have the `domain_group_id` why not just:

```suggestion
                    access_control.add(_prefix_group(domain_group_id))
```
?

seanstory (approved):

seanstory:
My approval lost it's comment. 😡 

TLDR, looks good, and the rest of my comments are nits that probably don't need to block this getting in. I'm comfortable with those being addressed separately if you'd prefer.

connectors/sources/sharepoint_online.py
timgrein: Yes, good catch, many calls less 🥳

seanstory (dismissed):

connectors/sources/sharepoint_online.py
wangch079: Personal preference, if it's only one line, I don't think it makes much sense to make it a function.

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +995,283 @@ async def validate_config(self):
...
+                    access_control.add(_prefix_group(domain_group_id))
+
+            if is_person(user):
+                name = user.get("Name")
+
+                if name and name.startswith("i:0#.f|membership|"):
+                    parts = name.split("|")
+
+                    if len(parts) > 2:
+                        email = parts[2]
```
wangch079, line 1041: You can extract this to a function, similar to `_domain_group_id`

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +995,283 @@ async def validate_config(self):
...
+                        access_control.add(_prefix_email(email))
+                        continue
+
+                if "EMail" not in user:
+                    continue
+
+                email = user["EMail"]
+
+                if _prefix_email(email) in access_control:
```
wangch079, line 1050: Seems not necessary as you uses `set` here.

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +995,283 @@ async def validate_config(self):
...
+
+                    if len(parts) > 2:
+                        email = parts[2]
+                        access_control.add(_prefix_email(email))
+                        continue
+
+                if "EMail" not in user:
+                    continue
+
+                email = user["EMail"]
```
wangch079, line 1048: It should be consistent, I see you use `user["Name"]` and `user.get("Name")` above without checking presence.

connectors/sources/sharepoint_online.py
wangch079: But I think you still use `prefixed_mail` and `prefixed_username`?

connectors/sources/sharepoint_online.py
```diff
@@ -879,258 +995,283 @@ async def validate_config(self):
...
-    def _dls_enabled(self):
-        if self._features is None:
-            return False
+                    if is_domain_group(user):
+                        domain_group_id = _domain_group_id(user["Name"])
 
-        if not self._features.document_level_security_enabled():
-            return False
+                        if domain_group_id:
+                            async for member_username in _emails_and_usernames_of_domain_group(
```
wangch079, line 1226: Can `_emails_and_usernames_of_domain_group` yield a tuple of `email` and `username`, instead of one `email` and one `username`? So that we can check `_already_seen` without calling `self.client.user`?

connectors/sources/sharepoint_online.py
```diff
@@ -1148,9 +1289,10 @@ async def get_docs(self, filtering=None):
                 site_collection["siteCollection"]["hostname"],
                 self.configuration["site_collections"],
             ):
-                yield site, None
+                access_control = await self._site_access_control(site)
+                yield self._decorate_with_access_control(site, access_control), None
 
-                async for site_drive in self.site_drives(site):
+                async for site_drive in self.site_drives(site, access_control):
```
wangch079, line 1295: I don't think it's necessary to pass `access_control` to `site_drives`, as well as `site_lists` and `site_pages`. It can be decorated in `get_docs` and `get_docs_incrementally`.

connectors/sources/sharepoint_online.py
```diff
@@ -459,6 +462,45 @@ async def site_groups(self, site_web_url):
         except NotFound:
             return
 
+    async def user_information_list(self, site_id):
+        expand = "fields"
+        url = f"{GRAPH_API_URL}/sites/{site_id}/lists/User Information List/items?expand={expand}"
```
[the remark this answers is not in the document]
timgrein: It's kinda hidden:

https://learn.microsoft.com/en-us/answers/questions/135386/how-to-get-sharepoint-site-group-name-using-graph

connectors/sources/sharepoint_online.py
timgrein: Yes, I'll address one or two refactorings in separate PRs (this will be one of them).

connectors/sources/sharepoint_online.py
```diff
@@ -748,6 +790,72 @@ def _prefix_email(email):
...
+    Yields:
+        - Prefixed email or email without prefix
+        - Prefixed user principal name or user principal name without prefix
+
+    """
+    async for identity in group_identities_generator(domain_group_id):
+        email = identity.get("mail")
+        username = identity.get("userPrincipalName")
+
+        if prefix:
```
wangch079, line 853: Do we have a use case when we need to prefix `email`/`username`? I think it's better to keep prefix out of the scope of this function.

connectors/sources/sharepoint_online.py
```diff
@@ -1243,26 +1378,32 @@ async def get_docs_incrementally(self, sync_cursor, filtering=None):
...
+                    yield self._decorate_with_access_control(
+                        site_list, access_control
+                    ), None, OP_INDEX
 
                     async for list_item, download_func in self.site_list_items(
                         site_id=site["id"],
                         site_list_id=site_list["id"],
                         site_web_url=site["webUrl"],
                         site_list_name=site_list["name"],
+                        access_control=access_control,
```
wangch079, line 1390: don't need to pass `access_control`

wangch079 (approved):
Overall LGTM. There are some nits which can be addressed later

timgrein:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.9|[](https://github.com/elastic/connectors-python/pull/1171)|

### Questions ?

PR #1158, opened 2023-06-30
Stream files for data extraction

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/5048

Stream files when extracting data instead of sending them multipart.
This allows for file extraction without loading a file into memory.
(See [documentation for aiohttp](https://docs.aiohttp.org/en/stable/client_quickstart.html#streaming-uploads) on how streaming files works).

Streaming the file upload requires some changes to the data-extraction-service ([relevant PR](https://github.com/elastic/enterprise-search-team/issues/5048)). The most notable aspect of this is the new endpoint will return a status code, which is neat (previously all errors were still returning `200` but with an error message).

#### Pre-Review Checklist

## Related Pull Requests

https://github.com/elastic/data-extraction-service/pull/17


artem-shelkovnikov (commented):
Looks good, but curious - did you check memory consumption difference here? I think it's possible to do with perf8 over `communications-enterprise-search-python-connectors-document-corpus` site collection, should give a good hint already!

navarone-feekery:
@artem-shelkovnikov I ran both versions until around 4000 docs had been ingested.
The streaming version has a lower average memory usage at around 380MiB (but did have a crazy 1gb spike for some reason).
The original multipart version averaged around 420MiB.

The numbers feel pretty close, not sure if it's really a significant change or not 🤔 
Graphs attached:

Streamed:

rss

Multipart:

rss

artem-shelkovnikov:
@navarone-feekery something feels off. Have you tried using [async generator](https://docs.aiohttp.org/en/stable/client_quickstart.html#streaming-uploads) instead with specifying the batch size yourself? Curious if batch size changes anything.

Also it's possible to debug it with `memray` to see who used this memory, though in this case it can easily eat all your available memory, so you'd have to prepare the test dataset yourself - it should be smaller but still contain all use-cases you wanna test.

connectors/utils.py
```diff
@@ -759,11 +766,8 @@ async def extract_with_file_send(self, filepath, filename):
         Returns a parsed response
         """
         with open(filepath, "rb") as file:
```
seanstory, line 768: I don't think this is the right approach. IIUC, `with open(...) as file:` is going to read the file into memory. This is what you're trying to avoid.

Tarek had POC'd originally using filepointers, which avoided this, but also constrained the Extraction Service to need to share a filesystem with connectors-python. But the more recent idea was to use "form data" which allows a single request to actually be made up from a lot of smaller requests. TikaServer supports this through `/rmeta` and `/form` endpoints. 

Doing a brief google, I think that python should support this nicely without needing to read in the full file contents: https://stackoverflow.com/a/12385661

seanstory: EDIT:
Friday brain. 🤦 

> IIUC, with open(...) as file: is going to read the file into memory

_<insert "that's not how any of this works gif">_

Doing the bare effort google (I should have done this to start) I see you're following streaming upload as defined here: https://docs.aiohttp.org/en/stable/client_quickstart.html#streaming-uploads 👍 

> But the more recent idea was to use "form data" which allows a single request to actually be made up from a lot of smaller requests. TikaServer supports this through /rmeta and /form endpoints.

This part I think is still accurate, but should be saving us memory on the Tika side, not the connector side. Looks like the relevant docs are right above the streaming docs: https://docs.aiohttp.org/en/stable/client_quickstart.html#post-a-multipart-encoded-file

navarone-feekery: Summary after perf tests and slack discussions: using the `/rmeta` endpoint isn't very different from using a streaming endpoint in terms of tikaserver memory usage, even for files of xGB size. I think it's safe to switch endpoints.

connectors/utils.py
```diff
@@ -760,29 +769,29 @@ async def extract_with_file_send(self, filepath, filename):
...
-            ) as response:
-                return await self.parse_extraction_resp(filename, response)
+        async with self._begin_session().put(
+            f"{self.host}/extract_text/", data=self.file_sender(filepath)
+        ) as response:
+            return await self.parse_extraction_resp(filename, response)
+
+    async def file_sender(self, filepath):
+        async with aiofiles.open(filepath, "rb") as f:
+            chunk = await f.read(64 * 1024)
```
artem-shelkovnikov, line 779: Is it 64 kilobytes?

artem-shelkovnikov: I think it's nice to make this setting configurable via the initializer or configuration

artem-shelkovnikov (commented):
Looks good, left a couple comments :)

connectors/utils.py
navarone-feekery: > Is it 64 kilobytes?

It is!

> I think it's nice to make this setting configurable via the initializer or configuration

Good idea, I'll add that

navarone-feekery: @artem-shelkovnikov added!

navarone-feekery: It would be under `extraction_service.stream_chunk_size` in the config now.
(config is optional for this so I haven't added it to the yml file).

artem-shelkovnikov (approved):

connectors/utils.py
```diff
@@ -693,6 +693,8 @@ def __init__(self):
         self.extraction_config = ExtractionService.get_extraction_config()
         if self.extraction_config is not None:
             self.host = self.extraction_config.get("host", None)
+            self.timeout = self.extraction_config.get("timeout", 30)
+            self.chunk_size = self.extraction_config.get("stream_chunk_size", 65535)
```
artem-shelkovnikov, line 697: Just curious why `65535`, but not `65536`? :D

navarone-feekery: Fixed, could you re-review? Sorry haha

artem-shelkovnikov (approved):

PR #1177, opened 2023-07-04
Dropbox connector code with AioHTTP

moxarth-elastic:
## Closes https://github.com/elastic/connectors-py/issues/1175, https://github.com/elastic/connectors-py/issues/1176

This PR contains the Dropbox connector implementation using `aiohttp` module. This PR aims to provide a better performance than Dropbox connector with SDK implementation (https://github.com/elastic/connectors-python/pull/1005).

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/dropbox_aiohttp.py
+++ b/connectors/sources/dropbox_aiohttp.py
@@ -67,17 +67,17 @@ else:
 
 
 class EndpointName(Enum):
-    ACCESS_TOKEN = "ACCESS_TOKEN"
-    PING = "PING"
-    CHECK_PATH = "CHECK_PATH"
-    FILES_FOLDERS = "FILES_FOLDERS"
-    FILES_FOLDERS_CONTINUE = "FILES_FOLDERS_CONTINUE"
-    RECEIVED_FILES = "RECEIVED_FILES"
-    RECEIVED_FILES_CONTINUE = "RECEIVED_FILES_CONTINUE"
-    RECEIVED_FILE_METADATA = "RECEIVED_FILE_METADATA"
-    DOWNLOAD = "DOWNLOAD"
-    PAPER_FILE_DOWNLOAD = "PAPER_FILE_DOWNLOAD"
-    RECEIVED_FILE_DOWNLOAD = "RECEIVED_FILE_DOWNLOAD"
+    ACCESS_TOKEN = "access_token"
+    PING = "ping"
+    CHECK_PATH = "check_path"
+    FILES_FOLDERS = "files_folders"
+    FILES_FOLDERS_CONTINUE = "files_folders_continue"
+    RECEIVED_FILES = "received_files"
+    RECEIVED_FILES_CONTINUE = "received_files_continue"
+    RECEIVED_FILE_METADATA = "received_files_metadata"
+    DOWNLOAD = "download"
+    PAPER_FILE_DOWNLOAD = "paper_file_download"
+    RECEIVED_FILE_DOWNLOAD = "received_files_download"
 
 
 ENDPOINTS = {
--- a/connectors/sources/dropbox_aiohttp.py
+++ b/connectors/sources/dropbox_aiohttp.py
@@ -194,11 +194,11 @@ class DropboxClient:
         return request_headers
 
     def _get_retry_after(self, retry, exception):
-        if retry > self.retry_count:
-            raise exception
         self._logger.warning(
             f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
         )
+        if retry >= self.retry_count:
+            raise exception
         retry += 1
         return retry, RETRY_INTERVAL**retry
 
@@ -207,34 +207,42 @@ class DropboxClient:
         match exception.status:
             case 401:
                 await self._set_access_token()
+            # For Dropbox APIs, 409 status code is responsible for endpoint-specific errors.
+            # Refer `Errors by status code` section in the documentation: https://www.dropbox.com/developers/documentation/http/documentation
             case 409:
                 raise InvalidPathException(
                     f"Configured Path: {self.path} is invalid or not found."
                 ) from exception
             case 429:
                 response_headers = exception.headers
+                updated_response_headers = {
+                    key.lower(): value for key, value in response_headers.items()
+                }
                 retry_seconds = int(
-                    response_headers.get("Retry-After", DEFAULT_RETRY_AFTER)
+                    updated_response_headers.get("retry-after", DEFAULT_RETRY_AFTER)
                 )
                 self._logger.warning(
                     f"Rate limited by Dropbox. Retrying after {retry_seconds} seconds"
                 )
+            case _:
+                raise
         self._logger.debug(f"Will retry in {retry_seconds} seconds")
         await self._sleeps.sleep(retry_seconds)
         return retry
 
     async def api_call(self, base_url, url_name, data=None, file_type=None, **kwargs):
-        retry = 0
+        retry = 1
         url = parse.urljoin(base_url, url_name)
         while True:
             try:
                 await self._set_access_token()
                 headers = self._get_request_headers(file_type=file_type, kwargs=kwargs)
-                async with self._get_session.post(  # pyright: ignore
+                async with self._get_session.post(
                     url=url, headers=headers, data=data
                 ) as response:
                     yield response
                     break
+            # These errors are handled in `check_errors` method
             except (InvalidClientCredentialException, InvalidRefreshTokenException):
                 raise
             except ClientResponseError as exception:
@@ -247,47 +255,57 @@ class DropboxClient:
                     ServerDisconnectedError,
                 ):
                     await self.close()
-                retry += 1
-                if retry > self.retry_count:
-                    raise exception
-                self._logger.warning(
-                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                retry, retry_seconds = self._get_retry_after(
+                    retry=retry, exception=exception
                 )
-                await self._sleeps.sleep(RETRY_INTERVAL**retry)
+                await self._sleeps.sleep(retry_seconds)
 
-    async def _paginated_api_call(self, is_shared=False, **kwargs):
+    async def _paginated_api_call(
+        self, base_url, breaking_field, is_shared=False, **kwargs
+    ):
         """Make a paginated API call for fetching Dropbox files/folders.
 
         Args:
+            base_url (str): Base URL for dropbox APIs
+            breaking_field (str): Breaking field to break the pagination
             is_shared (bool): Flag to determine the method flow. If `True` then fetching shared files, else normal files/folders.
+
+        Yields:
+            dictionary: JSON response
         """
         data = kwargs["data"]
         url_name = kwargs["url_name"]
         while True:
-            async for response in self.api_call(
-                base_url=kwargs["base_url"],
-                url_name=url_name,
-                data=json.dumps(data),
-            ):
-                json_response = await response.json()
-                yield json_response
+            try:
+                async for response in self.api_call(
+                    base_url=base_url,
+                    url_name=url_name,
+                    data=json.dumps(data),
+                ):
+                    json_response = await response.json()
+                    yield json_response
 
-                # Breaking condition for pagination
-                # breaking_fields:
-                #  - "has_more" for fetching files/folders
-                #  - "cursor" for fetching shared files
-                if not json_response.get(kwargs["breaking_field"]):
-                    return
+                    # Breaking condition for pagination
+                    # breaking_fields:
+                    #  - "has_more" for fetching files/folders
+                    #  - "cursor" for fetching shared files
+                    if not json_response.get(breaking_field):
+                        return
 
-                data = {
-                    BreakingField.CURSOR.value: json_response[
-                        BreakingField.CURSOR.value
-                    ]
-                }
-                if is_shared:
-                    url_name = ENDPOINTS["RECEIVED_FILES_CONTINUE"]
-                else:
-                    url_name = ENDPOINTS["FILES_FOLDERS_CONTINUE"]
+                    data = {
+                        BreakingField.CURSOR.value: json_response[
+                            BreakingField.CURSOR.value
+                        ]
+                    }
+                    if is_shared:
+                        url_name = ENDPOINTS["RECEIVED_FILES_CONTINUE"]
+                    else:
+                        url_name = ENDPOINTS["FILES_FOLDERS_CONTINUE"]
+            except Exception as exception:
+                self._logger.warning(
+                    f"Skipping of fetching files/folders for url: {parse.urljoin(base_url, url_name)}. Exception: {exception}"
+                )
+                return
 
     async def ping(self):
         await anext(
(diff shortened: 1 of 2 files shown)


moxarth-elastic:
buildkite test this

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,659 @@
...
+        self._logger.debug(f"Will retry in {retry_seconds} seconds")
+        await self._sleeps.sleep(retry_seconds)
+        return retry
+
+    async def api_call(self, base_url, url_name, data=None, file_type=None, **kwargs):
+        retry = 0
+        url = parse.urljoin(base_url, url_name)
+        while True:
+            try:
+                await self._set_access_token()
```
vidok, line 231: Can we call this method per session? Let's say we initiate it once before we start syncing. In case of expired token we catch the error and generate a new one.

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,659 @@
...
+                headers = self._get_request_headers(file_type=file_type, kwargs=kwargs)
+                async with self._get_session.post(  # pyright: ignore
+                    url=url, headers=headers, data=data
+                ) as response:
+                    yield response
+                    break
+            except (InvalidClientCredentialException, InvalidRefreshTokenException):
```
vidok, line 238: Can add a user-friendly message to this exception handler?

moxarth-elastic: I already added a message while raising these exceptions in `check_errors` method and that's why i am simply raising exceptions here. Do i need to add a message here also?

connectors/sources/dropbox_aiohttp.py
moxarth-elastic: So, you mean we'll create an access token from `_get_session` method and when the token gets expired we'll create a new session to generate a new token. Is that what you want to do?

connectors/sources/dropbox_aiohttp.py
seanstory, line 1: Is the long-term goal to replace `dropbox.py`, or to offer both? I'm surprised that this is being created in its own source right now.

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+        retry += 1
+        return retry, RETRY_INTERVAL**retry
+
+    async def _handle_client_errors(self, retry, exception):
+        retry, retry_seconds = self._get_retry_after(retry=retry, exception=exception)
+        match exception.status:
+            case 401:
+                await self._set_access_token()
+            case 409:
+                raise InvalidPathException(
```
seanstory, line 211: should this be a 404? 409 Conflict usually doesn't mean that a path was invalid or not-found.

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+                    f"Configured Path: {self.path} is invalid or not found."
+                ) from exception
+            case 429:
+                response_headers = exception.headers
+                retry_seconds = int(
+                    response_headers.get("Retry-After", DEFAULT_RETRY_AFTER)
```
seanstory, line 217: we had a recent bug where we were fetching the header like this, and the casing changed from `Retry-After` to `retry-after` and things broke. Can you make this fetch case insensitive?

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+        return retry
+
+    async def api_call(self, base_url, url_name, data=None, file_type=None, **kwargs):
+        retry = 0
+        url = parse.urljoin(base_url, url_name)
+        while True:
+            try:
+                await self._set_access_token()
+                headers = self._get_request_headers(file_type=file_type, kwargs=kwargs)
+                async with self._get_session.post(  # pyright: ignore
```
seanstory, line 233: what are you needing to ignore from pyright here?

connectors/sources/dropbox_aiohttp.py
seanstory: What if we just don't try to reset the access token unless we first hit a 401 error saying the token is expired?

By doing this at the start of every API call, we're being pretty inefficient.

connectors/sources/dropbox_aiohttp.py
seanstory: I think that's fair. Can you just add an inline comment here saying that these errors are handled in `check_errors`?

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+        """
+        data = kwargs["data"]
+        url_name = kwargs["url_name"]
+        while True:
+            async for response in self.api_call(
+                base_url=kwargs["base_url"],
+                url_name=url_name,
+                data=json.dumps(data),
+            ):
+                json_response = await response.json()
```
seanstory, line 272: Should you check for a 200 response? Or at least add a `try/except` in case `.json()` fails?

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+                yield json_response
+
+                # Breaking condition for pagination
+                # breaking_fields:
+                #  - "has_more" for fetching files/folders
+                #  - "cursor" for fetching shared files
+                if not json_response.get(kwargs["breaking_field"]):
```
seanstory, line 279: this will cause a KeyError if `breaking_field` isn't specified by the caller. If it's a required param, it should be added as an explicit required param, and not implicitly expected through kwargs.

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+    "PING": "users/get_current_account",
+    "CHECK_PATH": "files/get_metadata",
+    "FILES_FOLDERS": "files/list_folder",
+    "FILES_FOLDERS_CONTINUE": "files/list_folder/continue",
+    "RECEIVED_FILES": "sharing/list_received_files",
+    "RECEIVED_FILES_CONTINUE": "sharing/list_received_files/continue",
+    "RECEIVED_FILE_METADATA": "sharing/get_shared_link_metadata",
+    "DOWNLOAD": "files/download",
+    "PAPER_FILE_DOWNLOAD": "files/export",
+    "RECEIVED_FILE_DOWNLOAD": "sharing/get_shared_link_file",
```
seanstory, line 79: should all these keys also be constants, so that the values can be accessed using constants, instead of with string literals? I can typo `"ACCESS_TOKEN"` just as easily as typoing `"oauth2/token"`, so  this currently isn't providing much value over just using the literal strings everywhere they're used.

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+            options (dictionary): Config bulker options
+        """
+        options["concurrent_downloads"] = self.concurrent_downloads
+
+    async def close(self):
+        await self.dropbox_client.close()
+
+    async def ping(self):
+        await self.dropbox_client.ping()
+        self._logger.debug("Successfully connected to Dropbox")
```
seanstory, line 481: this can be info, I think? we don't ping that often, IIRC.

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+        )
+
+    def _pre_checks_for_get_content(
+        self, attachment_extension, attachment_name, attachment_size
+    ):
+        if attachment_extension == "":
+            self._logger.warning(
+                f"Files without extension are not supported by TIKA, skipping {attachment_name}."
+            )
+            return
```
seanstory, line 541:
```suggestion
            self._logger.debug(
                f"Files without extension are not supported, skipping {attachment_name}."
            )
            return
```

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+
+        elif attachment_extension.lower() not in TIKA_SUPPORTED_FILETYPES:
+            self._logger.warning(
+                f"Files with the extension {attachment_extension} are not supported by TIKA, skipping {attachment_name}."
```
seanstory, line 545:
```suggestion
            self._logger.debug(
                f"Files with the extension {attachment_extension} are not supported, skipping {attachment_name}."
```

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,661 @@
...
+    def _adapt_dropbox_doc_to_es_doc(self, response):
+        is_file = response[".tag"] == "file"
+        timestamp = response["server_modified"] if is_file else iso_utc()
+        return {
+            "_id": response["id"],
+            "type": FILE if is_file else FOLDER,
+            "name": response["name"],
+            "file path": response["path_display"],
+            "size": response["size"] if is_file else 0,
+            "_timestamp": timestamp,
```
seanstory, line 612: should use `.get()` instead of `[]` to avoid KeyErrors

tests/sources/test_dropbox_aiohttp.py
```diff
@@ -0,0 +1,809 @@
...
+
+@pytest.fixture
+def patch_default_wait_multiplier():
+    with mock.patch("connectors.sources.dropbox.RETRY_INTERVAL", 0):
+        yield
+
+
+@pytest.mark.asyncio
+@mock.patch("connectors.utils.apply_retry_strategy")
+async def test_api_call_when_token_is_expired(
```
seanstory, line 514: Shouldn't it try to refresh? Isn't there a a configured refresh token?

tests/sources/test_dropbox_aiohttp.py
```diff
@@ -0,0 +1,809 @@
...
+                    base_url=HOST_URLS["FILES_FOLDERS_HOST_URL"],
+                    url_name=PING,
+                    data=json.dumps(None),
+                )
+            )
+        await source.close()
+
+
+@pytest.mark.asyncio
+async def test_api_call_when_status_429_exception():
```
seanstory, line 544: Shouldn't this assert that a retry is made, not that it raises an exception?

tests/sources/test_dropbox_aiohttp.py
```diff
@@ -0,0 +1,809 @@
...
+    with mock.patch("aiohttp.ClientSession.post", return_value=get_stream_reader()):
+        with mock.patch(
+            "aiohttp.StreamReader.iter_chunked",
+            return_value=AsyncIterator([bytes(RESPONSE_CONTENT, "utf-8")]),
+        ):
+            response = await source.get_content(
+                attachment=MOCK_ATTACHMENT,
+                doit=True,
+            )
+            assert response == EXPECTED_CONTENT
```
seanstory, line 620: All these content tests repeat a lot of code, and could be combined into one test with a `@pytest.mark.parametrize`

connectors/sources/dropbox_aiohttp.py
moxarth-elastic: Yes, `dropbox.py` file will be replaced by this file once the review is completed. This new files is created to make a review easier.
cc: @vidok

connectors/sources/dropbox_aiohttp.py
moxarth-elastic: Yes, usually it should be but for dropbox apis, 409 is responsible for endpoint-specific errors. there is no such 404 status code according to the documentation. refer this api documentation - https://www.dropbox.com/developers/documentation/http/documentation#files-search

connectors/sources/dropbox_aiohttp.py
moxarth-elastic: Sure, i'll make all header-keys lower and then fetch them accordingly.

connectors/sources/dropbox_aiohttp.py
moxarth-elastic: I don't recall it correctly, let me check if it still persists.

connectors/sources/dropbox_aiohttp.py
moxarth-elastic: If you look at the `_set_access_token`, we're not resetting the token each time we call a method instead we first checking if the existing access token still alive and if the token is expired, we're creating a new one else using the existing one.

tests/sources/test_dropbox_aiohttp.py
moxarth-elastic: Yes, we can also do that. After retrying once, we can return an actual response 💡

tests/sources/test_dropbox_aiohttp.py
moxarth-elastic: I see! let me give it a try

connectors/sources/dropbox_aiohttp.py
seanstory: Ok thanks. An inline comment to explain this would be good to ensure this doesn't cause confusion in the future.

moxarth-elastic: Sure, i'll add an inline comment stating the same thing.

connectors/sources/dropbox_aiohttp.py
seanstory: I think you misunderstood - I like having all the endpoints in a single place and referencing them easily. I am suggesting just adding some more constants, like:

```
ACCESS_TOKEN = "access_token"
...
ENDPOINTS = {
  ACCESS_TOKEN : "oauth2/token"
  ...
 }
```

So that they can be referenced below as `ENDPOINTS.get(ACCESS_TOKEN)` instead of `ENDPOINTS.get('ACCESS_TOKEN')`. This way, we get an error about the variable not existing if someone typos as `ACESS_TOKEN`, instead of just getting `None`.

tests/sources/test_dropbox_aiohttp.py
```diff
@@ -0,0 +1,758 @@
...
+async def test_ping_with_ssl(
+    mock_post,
+):
+    mock_post.return_value.__aenter__.return_value.status = 200
+    source = create_source(DropboxDataSource)
+
+    source.dropbox_client.ssl_enabled = True
+    source.dropbox_client.certificate = (
+        "-----BEGIN CERTIFICATE----- Certificate -----END CERTIFICATE-----"
+    )
```
seanstory, line 461: custom client certs and enabling/disabling SSL isn't part of the connector configuration. Does it need to be?

seanstory (commented):
A few small nits, but otherwise looks good and very close.

connectors/sources/dropbox_aiohttp.py
moxarth-elastic: Gotcha! I'll create an enum for all the keys and refer them.

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,679 @@
...
+        )
+
+    def _adapt_dropbox_doc_to_es_doc(self, response):
+        is_file = response.get(".tag") == "file"
+        timestamp = response.get("server_modified") if is_file else iso_utc()
+        return {
+            "_id": response.get("id"),
+            "type": FILE if is_file else FOLDER,
+            "name": response.get("name"),
+            "file path": response.get("path_display"),
```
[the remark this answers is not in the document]
moxarth-elastic: Sure, i'll rename the key to `file_path`

connectors/sources/dropbox_aiohttp.py
```diff
@@ -0,0 +1,679 @@
...
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def _set_access_token(self):
+        if self.token_expiration_time and (
+            not isinstance(self.token_expiration_time, datetime)
+        ):
+            self.token_expiration_time = datetime.fromisoformat(
+                self.token_expiration_time
+            )  # pyright: ignore
```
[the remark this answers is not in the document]
moxarth-elastic: I tried removing all ignores, it has fixed now 👍

seanstory (approved):
Looks good. Thanks!

PR #1179, opened 2023-07-04
Make SPO validate configuration before doing custom validation

artem-shelkovnikov:
Found a small missing piece of code: when SPO connector runs validation, it never does backend validation of the configuration (validation of field types, required fields and so on).

This PR adds this logic to make sure basic validation happens as well.


connectors/sources/sharepoint_online.py
```diff
@@ -946,6 +946,8 @@ def get_default_configuration(cls):
         }
 
     async def validate_config(self):
+        self.configuration.check_valid()
```
wangch079, line 949:
```suggestion
        super().validate_config()
```

I think this pattern can also apply to other sources that are using `self.configuration.check_valid()`

artem-shelkovnikov: Yeah but for now I chose the consistent call here - same as in other connectors.

TBH I also think that `configuration.check_valid()` should be called outside - I think all connectors need to have valid configuration to function. If we follow current pattern, some connectors can by mistake or on purpose ignore it and allow invalid config, which is just weird. I think we should revisit this.

artem-shelkovnikov: What do you think if we merge it like it is now and make a tech debt issue to address this?

navarone-feekery (approved):

connectors/sources/sharepoint_online.py
navarone-feekery: ++ on making a tech debt issue

artem-shelkovnikov: Created an issue: https://github.com/elastic/connectors-python/issues/1194

PR #1184, opened 2023-07-04
Remove SHAREPOINT.md

artem-shelkovnikov:
Removing SHAREPOINT.md in root - there's no use to it and info there should be covered by official docs, tests and functional tests.


wangch079 (approved):
Do you know what is the folder [docs/reference](https://github.com/elastic/connectors-python/tree/main/docs/reference) used for? It doesn't include every connector client.

artem-shelkovnikov:
@wangch079 I'm honestly not sure, it looks like the `reference` folder contains information very similar to what's in public docs already, looks like a duplicate to me

seanstory:
> Do you know what is the folder [docs/reference](https://github.com/elastic/connectors-python/tree/main/docs/reference) used for? It doesn't include every connector client.

That's where Crest should be writing reference docs _before_ they get migrated to our official docs in ent-search-pubs. So it's for the tech-preview and under-development connectors.

wangch079:
>That's where Crest should be writing reference docs before they get migrated to our official docs in ent-search-pubs. So it's for the tech-preview and under-development connectors.

If a connector is released as a connector client, it should be available in our official doc and removed from this reference folder. If that's the case, I think we should remove confluence, jira and sharepoint for 8.9

artem-shelkovnikov:
> If a connector is released as a connector client, it should be available in our official doc and removed from this reference folder. If that's the case, I think we should remove confluence, jira and sharepoint for 8.9

Agreed. Do you think we need a team discussion or we can just remove them and notify Crest about it?

wangch079:
I think we can just remove them. Don't feel right that we have duplicated content in the repo and official guide

PR #1186, opened 2023-07-05
Add content extraction support for upper case extensions

### LINKED ISSUE #1185: Add content extraction support for upper case extensions

abhishekjoshi-crest:
## Closes #1185

Currently connectors are skipping files which are supported by TIKA but their extensions are in upper case. So by avoiding that check by lowering extension content extraction will be included.

And fix one minor typo in sharepoint online

#### Pre-Review Checklist


timgrein (commented):
LGTM, but I'm wondering, if we could have one function in utils `supported_by_tika(file_extension)`, which hides the `lower()` call. We could then reuse the method without always keeping mind to change the file extension to lowercase.

WDYT?

abhishekjoshi-crest:
> LGTM, but I'm wondering, if we could have one function in utils `supported_by_tika(file_extension)`, which hides the `lower()` call. We could then reuse the method without always keeping mind to change the file extension to lowercase.
> 
> WDYT?

Was thinking to add all this pre_checks in utils which will cover all cases like size, extensions but after having chat with Artem I dropped that plan as content extraction related changes will be there in framework around some time so it will be rework by moving in utils. So for now applied `lower()` at every extension check.

timgrein:
> > LGTM, but I'm wondering, if we could have one function in utils `supported_by_tika(file_extension)`, which hides the `lower()` call. We could then reuse the method without always keeping mind to change the file extension to lowercase.
> > WDYT?
> 
> Was thinking to add all this pre_checks in utils which will cover all cases like size, extensions but after having chat with Artem I dropped that plan as content extraction related changes will be there in framework around some time so it will be rework by moving in utils. So for now applied `lower()` at every extension check.

Thank you for the explanation!

timgrein (approved):

PR #1193, opened 2023-07-05
Make tests fail on warnings; add some warnings to ignore list

artem-shelkovnikov:
This PR makes 2 things:

1. Makes our regular tests fail if a warning happens
2. Ignores several warnings that are not relevant to us (see comments in the ini file)

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,6 +4,15 @@ addopts =
   -v
 filterwarnings =
   error
+; botocore has this warning that is reported by them to be irrelevant
   ignore:.*urllib3.contrib.pyopenssl.*:DeprecationWarning:botocore.*
+; latest main of aioresponses does not have this problem, but current package uses deprecated pkg_resources API
+  ignore:.*pkg_resources.*:DeprecationWarning
+; SQLAlchemy uses deprecated APIs internally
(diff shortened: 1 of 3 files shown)


jedrazb:
:wave: I digged a bit regarding the inheritance warning
- `ignore:Inheritance class RetryableAiohttpSession from ClientSession is discouraged:DeprecationWarning`
I don't see a straightforward way to fix it - the root of the warning is that the `aiogoogle` extends from `aiohttp.ClientSession` class. The inheritance looks as follows: `RetryableAiohttpSession -> aiogoogle.AiohttpSession -> aiohttp.ClientSession` and the warning is triggered by transitive closure of our class to `aiohttp.ClientSession`.

From `aiohttp` standpoint they prefer aggregation over inheritance, see:
- https://github.com/aio-libs/aiohttp/issues/2691
- https://github.com/aio-libs/aiohttp/issues/3695#issuecomment-495988925

That means that in order to get rid of this warning we would need to re-implement `aiogoogle.AiohttpSession` object with aggregation rather than inheritance - that implies copy-pasting lots of code from `aiohttp.ClientSession`  which can become tech debt immediately when aiogoogle version gets bumped and some potentially breaking changes are introduced.

navarone-feekery (approved):

PR #1198, opened 2023-07-05
Remove unused instance variable _dirty

wangch079:
Just spotted an unused variable `_dirty`.

#### Pre-Review Checklist


wangch079:
>Good one! Do you know if it's possible to catch those with linter/static analysis

I only see [unused global variable](https://pylint.readthedocs.io/en/latest/user_guide/configuration/all-options.html#allow-global-unused-variables) in Pylint and [unused local variable](https://beta.ruff.rs/docs/rules/unused-variable/) in ruff.

artem-shelkovnikov:
Do you know by chance if we run these rules?

PR #1213, opened 2023-07-07
Ignore deleted files when choosing a download function

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/5135

The incremental sync is used to ignore deleted files filtering them by `@microsoft.graph.downloadUrl`. It worked well when we checked that property first. However, the check was moved down so it made the service fail when the `folder` is deleted because deleted items don't have the `name` property.

This PR will add an explicit validation and ignore all deleted items.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -1526,6 +1526,8 @@ class SharepointOnlineDataSource(BaseDataSource):
 
     def download_function(self, drive_item, max_drive_item_age):
         if "deleted" in drive_item:
+            # deleted drive items do not contain `name` property in the payload
+            # so drive_item['id'] is used
             self._logger.debug(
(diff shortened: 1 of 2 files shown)


connectors/sources/sharepoint_online.py
```diff
@@ -1525,6 +1525,13 @@ def drive_item_operation(self, item):
             return OP_INDEX
 
     def download_function(self, drive_item, max_drive_item_age):
+        if "deleted" in drive_item:
+            self._logger.debug(
+                f"Not downloading the item id={drive_item['id']} because it has been deleted"
+            )
```
artem-shelkovnikov, line 1531: Might be worth to add a note that `drive_item['id']` is not available for situation when file is deleted.

artem-shelkovnikov (dismissed):

PR #1219, opened 2023-07-07
Allow file pointers for local data extraction

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/5137

Re-enable filepointers as an option for local data extraction.

_(The following instructions will also be added to our official documentation)_

To set this up:
- connectors-python must be run in docker
- you must also run data-extraction-service (use the branch in [this PR](https://github.com/elastic/data-extraction-service/pull/18) if it isn't merged yet)
- the two docker images must share a network and a volume (see below for instructions)

Sample config (note the `host` is no longer `localhost`):

```
extraction_service:
  host: http://host.docker.internal:8090
  use_file_pointers: true
  shared_volume_dir: '/app/files'
```

First make a volume that will be used by both docker containers:

```bash
$ docker volume create --name connectors-volume
```

Follow the [instructions to run connectors-python in docker](https://github.com/elastic/connectors-python/blob/main/docs/DOCKER.md), but when calling `docker run`, be sure to add the value set for `shared_volume_dir` as a volume. Example:

```bash
$ docker run -i --rm --name connectors-python \
  -v ~/connectors-python-config:/config \
  -v connectors-volume:/app/files \
  --network "elastic" --tty --rm \
  connectors-python \
  /app/bin/elastic-ingest -c /config/config.yml
```

Then run data-extraction-service using the same network and volume:

```
$ docker run -p 8090:8090 -i -t --rm --name extraction-service \
  -v ConnectorsVolume:/app/files \
  --network "elastic" \
  extraction-service
```

Then set up a sharepoint online connector and set `use_text_extraction_service: true` and begin ingesting documents!
This set up will also work for file streaming.

#### Pre-Review Checklist

## Related Pull Requests

https://github.com/elastic/data-extraction-service/pull/18

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -910,8 +910,10 @@ class SharepointOnlineDataSource(BaseDataSource):
 
         if self.configuration["use_text_extraction_service"]:
             self.extraction_service = ExtractionService()
+            self.download_dir = self.extraction_service.get_volume_dir()
         else:
             self.extraction_service = None
+            self.download_dir = None
 
     def _set_internal_logger(self):
         self.client.set_logger(self._logger)
@@ -1687,7 +1689,7 @@ class SharepointOnlineDataSource(BaseDataSource):
 
         try:
             async with NamedTemporaryFile(
-                mode="wb", delete=False, suffix=file_extension
+                mode="wb", delete=False, suffix=file_extension, dir=self.download_dir
             ) as async_buffer:
                 source_file_name = async_buffer.name
 
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -802,9 +802,6 @@ class ExtractionService:
                 f"Text extraction unexpectedly failed for {filename}. Error: {e}"
             )
 
-        if content:
-            logger.info(f"{filepath} WAS EXTRACTED")
-
         return content
 
     async def file_sender(self, filepath):
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -712,7 +712,16 @@ class ExtractionService:
         if self.extraction_config is not None:
             self.host = self.extraction_config.get("host", None)
             self.timeout = self.extraction_config.get("timeout", 30)
-            self.chunk_size = self.extraction_config.get("stream_chunk_size", 65536)
+            self.headers = {"accept": "application/json"}
+
+            self.use_file_pointers = self.extraction_config.get("use_file_pointers", False)
+            if self.use_file_pointers:
+                self.volume_dir = self.extraction_config.get("fileshare_dir", "/app/files")
+                self.chunk_size = None
+            else:
+                self.volume_dir = None
+                self.chunk_size = self.extraction_config.get("stream_chunk_size", 65536)
+                self.headers["content-type"] = "application/octet-stream"
         else:
             self.host = None
 
@@ -734,10 +743,7 @@ class ExtractionService:
         timeout = aiohttp.ClientTimeout(total=self.timeout)
         self.session = aiohttp.ClientSession(
             timeout=timeout,
-            headers={
-                "content-type": "application/octet-stream",
-                "accept": "application/json",
-            },
+            headers=self.headers,
         )
 
     async def _end_session(self):
@@ -745,6 +751,12 @@ class ExtractionService:
             return
 
         await self.session.close()
+
+    def get_volume_dir(self):
+        if self.host is None:
+            return None
+
+        return self.volume_dir
 
     async def extract_text(self, filepath, original_filename):
         """Sends a text extraction request to tika-server using the supplied filename.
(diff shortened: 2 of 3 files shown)


tests/test_utils.py
```diff
@@ -660,6 +660,30 @@ async def test_extract_text(self, mock_responses):
 
             assert response == "I've been extracted!"
 
+    @pytest.mark.asyncio
+    async def test_extract_text_with_file_pointer(self, mock_responses):
+        filepath = "/tmp/notreal.txt"
+        url = "http://localhost:8090/extract_text/?local_file_path=/tmp/notreal.txt"
+        payload = {"extracted_text": "I've been extracted!"}
```
seanstory, line 667:
```suggestion
        payload = {"extracted_text": "I've been extracted from a local file!"}
```

seanstory (dismissed):
Good stuff. Nothing needs to be done to declare a volume on the docker image? Typically I'd expect this would work by requiring both the connectors-python container and the data-extraction-service container to have a "volume" that uses the same dir on the host system.

Also, does this really require connectors-python to be run from docker? It can't be run from just `make run` as long as the dir is set correctly?

connectors/utils.py
```diff
@@ -712,7 +712,18 @@ def __init__(self):
...
-            self.chunk_size = self.extraction_config.get("stream_chunk_size", 65536)
+            self.headers = {"accept": "application/json"}
+
+            self.use_file_pointers = self.extraction_config.get(
+                "use_file_pointers", False
+            )
+            if self.use_file_pointers:
+                self.volume_dir = self.extraction_config.get(
+                    "fileshare_dir", "/app/files"
+                )
```
wangch079, line 723:
```suggestion
                self.volume_dir = self.extraction_config.get(
                    "shared_volume_dir", "/app/files"
                )
```

tests/test_utils.py
navarone-feekery: The fact that this test was passing is concerning, I'll investigate first.

navarone-feekery:
@seanstory that's completely correct actually, I think I was just in docker-land when I wrote this.
When I write the documentation for this, I'll include both methods.

navarone-feekery:
> Nothing needs to be done to declare a volume on the docker image? Typically I'd expect this would work by requiring both the connectors-python container and the data-extraction-service container to have a "volume" that uses the same dir on the host system.

When you run `docker volume create --name connectors-python` docker will assign a local filepath. Afterwards when I inspect I can see the results:

```sh
$ docker volume inspect connectors-python
[
    {
        "CreatedAt": "2023-07-07T14:41:16Z",
        "Driver": "local",
        "Labels": null,
        "Mountpoint": "/var/lib/docker/volumes/connectors-python/_data",
        "Name": "ConnectorsVolume",
        "Options": null,
        "Scope": "local"
    }
]
```

Specifying this with `-v connectors-python:/path/inside/container` when running the docker containers is enough to mount the volume.

connectors/utils.py
```diff
@@ -770,7 +785,13 @@ async def extract_text(self, filepath, original_filename):
         )
 
         try:
-            content = await self.extract_with_file_send(filepath, filename)
+            async with self._begin_session().put(
```
vidok, line 788: `self.filepointer_params(filepath)` I expect that this method always returns `filepointer_params` but not "". 

for the readability sake, I'd like to suggest to split this method into two instead of having 2 if-else statements.

navarone-feekery: I separated them into two functions, it is definitely more readable now.

seanstory:
> Specifying this with -v connectors-python:/path/inside/container when running the docker containers is enough to mount the volume

Hadn't occurred to me that this could be done from the CLI as well as through the dockerfile. I'm fine with the CLI approach as long as we document the relevant steps. Because I won't remember how to do this the first time I get an SDH about it. :)

connectors/sources/sharepoint_online.py
```diff
@@ -910,8 +910,10 @@ def __init__(self, configuration):
 
         if self.configuration["use_text_extraction_service"]:
             self.extraction_service = ExtractionService()
+            self.download_dir = self.extraction_service.get_volume_dir()
         else:
             self.extraction_service = None
+            self.download_dir = None
```
seanstory, line 916: This looked concerning, but after a little digging found in https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp:

> If dir is not None, the file will be created in that directory; otherwise, a default directory is used.

Leaving a comment here in case this startles other reviewers.

connectors/utils.py
```diff
@@ -712,7 +712,19 @@ def __init__(self):
...
+            self.use_file_pointers = self.extraction_config.get(
+                "use_file_pointers", False
+            )
+            if self.use_file_pointers:
+                self.volume_dir = self.extraction_config.get(
+                    "shared_volume_dir", "/app/files"
+                )
+            else:
+                self.volume_dir = None
+                self.headers["content-type"] = "application/octet-stream"
```
seanstory, line 727: Do we need to set the headers to be `application/json` in the `if` clause to match this? Or does it work without them being set explicitly?

seanstory (approved):
LGTM. 

We'll also need to update the CONFIG.md, but that can wait for a follow-up pr if you like.

connectors/utils.py
navarone-feekery: It works as is, I think it isn't necessary to set `content-type` to anything for filesend because the other request type is just a PUT with a URI param.

navarone-feekery:
> We'll also need to update the CONFIG.md, but that can wait for a follow-up pr if you like.

Will make it a followup PR. Thanks!

connectors/utils.py
seanstory: 🤦 oh duh. For some reason I thought it had a payload, but you're totally right.

Issue #1204, opened 2023-07-06
[Sharepoint Online] Improve performance of Identity Sync


artem-shelkovnikov:
As reported, syncing identities takes a long time:

![image](https://github.com/elastic/connectors-python/assets/12238374/1b1ef41c-bead-44d6-91e7-c3f072a744a8)

Minor triaging showed the following:

We fetch members for each group for each user but instead we can fetch groups and its members with $expand query: https://graph.microsoft.com/v1.0/groups?$expand=members . It can reduce number of calls to Sharepoint dramatically.
Right now the number of calls is gonna be:
For each user (N users)
For each group (M groups per user)
Fetch members, so
N * M - for 300 users and 4 groups each it’s 1200 requests.

For 42000 users in the screenshot and assuming at least 10 groups, it's gonna be 420.000 requests each costing 5 resource units. 

We need to optimize this logic to reduce sync time.

seanstory:
☝️ if that PR's approach goes in, I'll need to also:
- [ ] update the Kibana configuration for Native SPO connectors
- [ ] update the configuration reference for the SPO connector docs

seanstory:
closing this, though https://github.com/elastic/enterprise-search-team/issues/5162 is a follow-up, because this change was insufficient.

PR #1220, opened 2023-07-07
Make access control strategy configurable

seanstory:
## Part of https://github.com/elastic/connectors-python/issues/1204

See https://elastic.slack.com/archives/C01795T48LQ/p1688745239319069

Before this change, the access control sync for Sharepoint Online would do something like:
```
for each site collection
  for each site
    for each item in the UserInformationList
      for each person in a group item
        for each group that person belongs to
          add_to_acl_list(group)
```          

This 5 levels of nesting is only efficient if each collection is relatively small. 

This PR adds an alternate approach:

```
for each user in list_users
  for each group that user belongs to
    add_to_acl_list(group)
```

This is likely more efficient at larger scales, but may be less efficient for large enterprises where the connector is only intended to sync a small number of small/private sites.

#### Pre-Review Checklist


connectors/sources/sharepoint_online.py
```diff
@@ -1174,97 +1194,121 @@ async def get_access_control(self):
...
+                return person_access_control_doc
 
-                                update_already_seen(member_email, member_username)
+        if not self.configuration["fetch_users_by_site"]:
+            self._logger.info("Fetching all users")
+            async for user in self.client.users():
+                user_doc = await process_user(user)
+                if not user_doc:
+                    continue
+                yield user_doc
```
timgrein, line 1236:
```suggestion
                if user_doc:
                    yield user_doc
```

nit: super small nit

connectors/sources/sharepoint_online.py
```diff
@@ -1174,97 +1194,121 @@ async def get_access_control(self):
...
+                self._logger.debug(
+                    f"Looking at site collection for location: {site_collection.get('dataLocationCode')}"
+                )
+                async for site in self.client.sites(
+                    site_collection["siteCollection"]["hostname"],
+                    self.configuration["site_collections"],
+                ):
+                    site_id = site.get("id")
+                    self._logger.debug(f"Looking at site: {site_id}")
+                    if _already_seen(site_id):
```
timgrein, line 1249: Nice, so a site can be part of multiple site collections I suppose?

connectors/sources/sharepoint_online.py
```diff
@@ -1174,97 +1194,121 @@ async def get_access_control(self):
...
+                                        await self._user_access_control_doc(owner)
+                                    )
+                                    if owner_access_control_doc:
+                                        yield owner_access_control_doc
+
+                        elif is_person(user):
+                            user_doc = await process_user(user)
+                            if not user_doc:
+                                continue
+                            yield user_doc
```
timgrein, line 1306:
```suggestion
                            if user_doc:
                                yield user_doc
```

nit: just to make it consistent with the other places

timgrein (dismissed):
LGTM, only two nits. Really nice investigation and solution 👏  
I guess we should've a corresponding PR in Kibana for the new RFC?

seanstory:
> I guess we should've a corresponding PR in Kibana for the new RFC?

Yep. I have two other tasks on https://github.com/elastic/connectors-python/issues/1204#issuecomment-1625942321 if this PR is going in. And given that I haven't seen any objections, I think it will.

timgrein (approved):

PR #1226, opened 2023-07-10
[Fix] Specify the job type when starting a sync

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/5133

This PR sets the `job_type` parameter, when calling `sync_starts`. This fixes a bug, where a failed access control sync prevented content syncs from running. 

Reason: the default `job_type` for `sync_starts` is `JobType.FULL`. Therefore the `last_sync_status` (for content syncs) was set to `in_progress` by an access control sync, which is wrong. When the access control sync failed the sync job runner correctly updated `last_access_control_sync_status` and not `last_sync_status`, which stayed `in_progress`.

#### Pre-Review Checklist


wangch079 (approved):

connectors/sync_job_runner.py
```diff
@@ -298,7 +298,7 @@ async def sync_starts(self):
             )
 
         try:
-            await self.connector.sync_starts()
+            await self.connector.sync_starts(job_type=job_type)
```
artem-shelkovnikov, line 301: Should `job_type` become mandatory instead of being a named parameter with a default?

PR #1240, opened 2023-07-10
Do not download unsupported files

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/5155

There is no need to download files that can't be processed successfully in the extraction service or ingest pipeline.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -1667,6 +1667,12 @@ class SharepointOnlineDataSource(BaseDataSource):
         if not doit:
             return
 
+        if not self.is_supported_format(attachment["_original_filename"]):
+            self._logger.debug(
+                f"Not downloading attachment {attachment['_original_filename']}: file type is not supported"
+            )
+            return
+
         # We don't know attachment sizes unfortunately, so cannot properly ignore them
 
         # Okay this gets weird.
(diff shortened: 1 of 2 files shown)


wangch079 (approved):

connectors/sources/sharepoint_online.py
```diff
@@ -1775,3 +1788,13 @@ async def close(self):
 
     def advanced_rules_validators(self):
         return [SharepointOnlineAdvancedRulesValidator()]
+
+    def is_supported_format(self, filename):
```
timgrein, line 1792: nit: that could live in `utils.py` right? `TIKA_SUPPORTED_FILETYPES` also lives in `utils.py`, so would make sense to me to keep things together.

vidok: Totally agree however,  I'm not sure this is the right moment to move to `utils.py`. The similar code lives in `sharepoint_server.py` and `TIKA_SUPPORTED_FILETYPES ` should not live in `utils.py` either :).

connectors/sources/sharepoint_online.py
```diff
@@ -1775,3 +1788,13 @@ async def close(self):
...
+        if "." not in filename:
+            return False
+
+        attachment_extension = list(os.path.splitext(filename))
```
navarone-feekery, line 1796: nit: I think `list` isn't required, because you can use the following `[-1]` on the tuple result.

navarone-feekery (approved):
LGTM! Left some optional nits.

navarone-feekery (approved):
Nice! 🚀

artem-shelkovnikov (approved):
LGTM - I would also add a test that checks if file extension is ALL CAPS to pin the fact that we handle this case too

PR #1205, opened 2023-07-07
Remove duplicate reference documents

wangch079:
Those reference documents are all migrated to the official guide, therefore remove them from the reference folder.

#### Pre-Review Checklist
- [x] this PR has a meaningful title
- [x] this PR has a thorough description
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

PR #1244, opened 2023-07-11
[Dropbox] Fix reference docs relative links 404s

jedrazb:
## [Dropbox] Fix reference docs relative links 404s

Since the docs are nested under `reference` folder we need to tweak relative links. Right now clicking on them yields 404s.

Screenshot 2023-07-11 at 11 03 59

### DIFF
--- a/docs/reference/dropbox.md
+++ b/docs/reference/dropbox.md
@@ -1,6 +1,6 @@
 # Dropbox Connector
 
-The [Elastic Dropbox connector](../connectors/sources/dropbox.py) is provided in the Elastic connectors python framework and can be used via [build a connector](https://www.elastic.co/guide/en/enterprise-search/current/build-connector.html).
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov:
Afaik we're removing docs in `reference` directory - see https://github.com/elastic/connectors-python/pull/1205

Maybe we should delete this reference file too and move all of its content that is applicable to our docs?

jedrazb:
> Afaik we're removing docs in `reference` directory - see #1205
> 
> Maybe we should delete this reference file too and move all of its content that is applicable to our docs?

From what I understand we want to have at least some docs for connectors under development in the repo before the docs are migrated to the official docs page. 

I'm in the process of adding the google drive connector docs in the `reference` folder and just noticed this bug for dropbox  - not sure about the plans to move the documentation for it.

wangch079:
> Afaik we're removing docs in `reference` directory - see #1205

Hi @artem-shelkovnikov , I removed those markdown files from `/doc/reference` because they are all migrated to the official guide. AFAIK, this folder is for a connector in development, and it should be migrated to the official guide once it's promoted to a connector client.

wangch079 (approved):

PR #1255, opened 2023-07-11
Performance improvements to the ACL sync

seanstory:
## Part of https://github.com/elastic/enterprise-search-team/issues/5162

Speeding up the ACL sync.
- fixes an id-vs-username bug
- Utilizes `$expand=transitiveMemberOf` in a bulk request, to reduce request volume
- Utilizes `$top` to cap out page size
- Utilizes `$filter` to fetch only active users, and not disabled users.
- Utilizes `$select` to fetch only the fields we need, which speeds up response time
- Removes fetching users from UserInformationList because that was broken, and we could not extract User Ids (to then fetch transitive group membership)

#### Pre-Review Checklist
- [x] this PR has a meaningful title
- [x] this PR links to all relevant github issues that it fixes or partially addresses
- [x] if there is no GH issue, please create it. Each PR should have a link to an issue
- [x] this PR has a thorough description
- [x] Covered the changes with automated tests
- [x] Tested the changes locally
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)
- [x] Considered corresponding documentation changes (follow-up PR)
- [x] Contributed any configuration settings changes to the configuration reference

## Related Pull Requests

- https://github.com/elastic/kibana/pull/161798
- https://github.com/elastic/enterprise-search-pubs/pull/3627

PR #1259, opened 2023-07-12
[SPO] Add configurable drive item permissions support

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/5164

This PR adds a RFC `fetch_drive_item_permissions`, which specifies whether drive item specific permissions should be fetched or not. Fetching drive item permissions is now also done via scrolling and not via fetching everything at once. 

Using the `$batch` API to batch 20 permission requests for drive items into 1 reducing the number of requests drastically.

Note: We add `siteUser` and `siteGroup` for drive item permissions, but we don't fetch them during an access control sync yet. This will be part of another PR and will also be made configurable + the dependency checks will be added on the RFC side.

A separate PR adding the `fetch_drive_item_permissions` RFC will follow for Kibana.

#### Pre-Review Checklist


seanstory (dismissed):
LGTM. Have you shared the results of this branch yet with Gustavo? May be good to get confirmation of output before declaring victory.

connectors/sources/sharepoint_online.py
```diff
@@ -1005,6 +1022,15 @@ def get_default_configuration(cls):
                 "type": "bool",
                 "value": False,
             },
+            "fetch_drive_item_permissions": {
+                "depends_on": [{"field": "use_document_level_security", "value": True}],
+                "display": "toggle",
+                "label": "Fetch drive item permissions",
+                "order": 9,
+                "tooltip": "Enable this option to fetch drive item specific permissions. Note that this setting can potentially increase sync time.",
```
leemthompo, line 1030:
```suggestion
                "tooltip": "Enable this option to fetch drive item specific permissions. Note that this setting can increase sync time.",
```

A very minor 🪒! 😃

connectors/sources/sharepoint_online.py
```diff
@@ -1460,6 +1494,11 @@ async def get_docs_incrementally(self, sync_cursor, filtering=None):
                                 drive_item, access_control
                             )
 
+                            if self.configuration["fetch_drive_item_permissions"]:
```
wangch079, line 1497: My personal preference would be moving this check into `_with_drive_item_permissions`, so you you don't have to check in multiple places.

connectors/sources/sharepoint_online.py
```diff
@@ -1203,6 +1230,7 @@ async def _user_access_control_doc(self, user):
 
         prefixed_mail = _prefix_email(email)
         prefixed_username = _prefix_user(username)
+        prefixed_user_id = _prefix_user_id(user.get("id"))
         id_ = email if email else username
```
wangch079, line 1234: Wondering can we use the `user.get("id")` here, instead of email/username?

connectors/sources/sharepoint_online.py
```diff
@@ -1005,6 +1022,15 @@ def get_default_configuration(cls):
...
                 "value": False,
             },
+            "fetch_drive_item_permissions": {
+                "depends_on": [{"field": "use_document_level_security", "value": True}],
+                "display": "toggle",
+                "label": "Fetch drive item permissions",
+                "order": 9,
+                "tooltip": "Enable this option to fetch drive item specific permissions. Note that this setting can potentially increase sync time.",
+                "type": "bool",
+                "value": True,
```
wangch079, line 1032: This is expected to be `True` by default, and I don't see `8.9` label in the PR, is migration needed if customers upgrade from `8.9` to `8.10`?

timgrein: Yes, I would expect that we need one.

connectors/sources/sharepoint_online.py
timgrein: As we may want to have cross-system identities in the near future, it's probably better to use the email as the id as this is searchable.

connectors/sources/sharepoint_online.py
timgrein: Adjusted with [e3820ba](https://github.com/elastic/connectors-python/pull/1259/commits/e3820babb0af8a8ad81da7b02cea0501568c298b)

connectors/sources/sharepoint_online.py
```diff
@@ -1005,6 +1022,15 @@ def get_default_configuration(cls):
                 "type": "bool",
                 "value": False,
             },
+            "fetch_drive_item_permissions": {
```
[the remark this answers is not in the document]
timgrein: Added RFC to functional test `connector.json` with [0719134](https://github.com/elastic/connectors-python/pull/1259/commits/0719134663354606e900cb02cc0bee4dfab22b9f)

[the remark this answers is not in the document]
artem-shelkovnikov: Just merged the PR, you can update `tests/sources/fixtures/sharepoint_online/connector.json` to include the field now

connectors/sources/sharepoint_online.py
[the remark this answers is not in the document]
artem-shelkovnikov: I think the order needs to be decremented after https://github.com/elastic/connectors-python/pull/1255#pullrequestreview-1528292491 - it removes item #8

connectors/sources/sharepoint_online.py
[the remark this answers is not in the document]
timgrein: Also added, but I've to decrement the order also here :)

connectors/sources/sharepoint_online.py
```diff
@@ -606,9 +606,14 @@ async def drive_items(self, drive_id, url=None):
...
-        return await self._graph_api_client.fetch(
-            f"{GRAPH_API_URL}/drives/{drive_id}/items/{item_id}/permissions"
-        )
+        try:
+            async for page in self._graph_api_client.scroll(
+                f"{GRAPH_API_URL}/drives/{drive_id}/items/{item_id}/permissions"
+            ):
+                for drive_item in page:
+                    yield drive_item
+        except NotFound:
```
artem-shelkovnikov, line 615:
```suggestion
                for permission in page:
                    yield permission
        except NotFound:
```

connectors/sources/sharepoint_online.py
```diff
@@ -606,9 +606,14 @@ async def drive_items(self, drive_id, url=None):
             yield page
 
     async def drive_item_permissions(self, drive_id, item_id):
```
artem-shelkovnikov, line 608: So there's no way to fetch permissions in batch?

Do you think doing [request batching](https://learn.microsoft.com/en-us/graph/json-batching) could help here?

timgrein: Just for my understanding: We would pile up to 20 drive items and then fetch the permissions for these 20 at once?

artem-shelkovnikov: Yup. Code will get ugly, but that's potentially 20x performance bump?

timgrein: Let's do it, if it improves performance 🚀

connectors/sources/sharepoint_online.py
```diff
@@ -354,14 +366,14 @@ async def scroll_delta_url(self, url):
             else:
                 break
 
-    async def _get_json(self, absolute_url):
+    async def _get_json(self, absolute_url, method=HTTPMethod.GET, payload=None):
```
artem-shelkovnikov, line 369: It's not `_get_json` any more :D

Maybe worth to split into different methods or just do same as I did in `pipe` (do specific logic in the method calling `self._call_api`)?

connectors/sources/sharepoint_online.py
```diff
@@ -373,13 +385,27 @@ async def _call_api(self, absolute_url):
...
+                    async with self._http_session.post(
+                        absolute_url, headers=headers, json=payload
+                    ) as resp:
+                        yield resp
+
+                case _:
+                    self._logger.error(
+                        f"Unsupported http method for SPO client: '{method.value}'."
+                    )
+                    raise InternalServerError("Something unexpected occurred")
```
artem-shelkovnikov, line 406: Is it possible to make it without `switch/case`? It starts to look really complex

connectors/sources/sharepoint_online.py
```diff
@@ -89,7 +98,7 @@ class NotFound(Exception):
 
 
 class InternalServerError(Exception):
-    """Internal exception class to handle 500s from the API, which could sometimes also mean NotFound."""
+    """Exception class to indicate that something went wrong on the server side."""
```
artem-shelkovnikov, line 101: It follows the practice with `NotFound` for example - "Internal" here means, that it's used for internal logic of the clients :)

connectors/sources/sharepoint_online.py
timgrein: We could use `get_attr`, use the lowercase value of the enum and do some `**kwargs` magic, but don't know, if that really makes it more readable or more obscure in less LOC.

artem-shelkovnikov: It can probably be even methods like:

`async def _get(self, absolute_url)`
`async def _post(self, absolute_urll, payload)`

I think it's more verbose - passing booleans or enums that change the logic of the methods is making it hard.

For example, you can pass `payload` that is non-none and have a method=GET and code will have to handle it - too much complexity. Let's break the method!

artem-shelkovnikov: > We could use get_attr, use the lowercase value of the enum and do some **kwargs magic, but don't know, if that really makes it more readable or more obscure in less LOC.

I think that's already going into some really dark stuff, let's not do `get_attr` :D

timgrein: Splitted it up 👍

connectors/sources/sharepoint_online.py
```diff
@@ -351,12 +359,34 @@ async def scroll_delta_url(self, url):
...
             return await resp.json()
 
     @asynccontextmanager
     @retryable_aiohttp_call(retries=DEFAULT_RETRY_COUNT)
-    async def _call_api(self, absolute_url):
+    async def _post(self, absolute_url, payload=None):
+        try:
+            token = await self._api_token.get()
+            headers = {"authorization": f"Bearer {token}"}
+            self._logger.debug(f"Calling Sharepoint Endpoint: {absolute_url}")
```
wangch079, line 371: Worth adding `POST` as method, payload as well.

connectors/sources/sharepoint_online.py
```diff
@@ -351,12 +359,34 @@ async def scroll_delta_url(self, url):
...
+
+            async with self._http_session.post(
+                absolute_url, headers=headers, json=payload
+            ) as resp:
+                yield resp
+
+            return
```
wangch079, line 378: If it only yield one item, do we need to use `yield`?

connectors/sources/sharepoint_online.py
```diff
@@ -368,45 +398,44 @@ async def _call_api(self, absolute_url):
...
-                raise
+                self._logger.warning(
+                    f"Response Code from Sharepoint Server is 429 but Retry-After header is not found, using default retry time: {DEFAULT_RETRY_SECONDS} seconds"
+                )
+                retry_seconds = DEFAULT_RETRY_SECONDS
             self._logger.debug(
                 f"Rate Limited by Sharepoint: retry in {retry_seconds} seconds"
             )
 
+            await self._sleeps.sleep(retry_seconds)  # TODO: use CancellableSleeps
```
wangch079, line 424: `self._sleeps` is an instance of `CancellableSleeps`

connectors/sources/sharepoint_online.py
```diff
@@ -368,45 +398,44 @@ async def _call_api(self, absolute_url):
...
-                raise InternalServerError from e
+    async def _handle_client_response_error(self, absolute_url, e):
+        if e.status == 429 or e.status == 503:
+            response_headers = e.headers or {}
+            if "Retry-After" in response_headers:
+                retry_seconds = int(response_headers["Retry-After"])
             else:
-                raise
+                self._logger.warning(
+                    f"Response Code from Sharepoint Server is 429 but Retry-After header is not found, using default retry time: {DEFAULT_RETRY_SECONDS} seconds"
```
wangch079, line 417:
```suggestion
                    f"Response Code from Sharepoint Server is {e.status} but Retry-After header is not found, using default retry time: {DEFAULT_RETRY_SECONDS} seconds"
```

connectors/sources/sharepoint_online.py
```diff
@@ -368,45 +398,44 @@ async def _call_api(self, absolute_url):
...
             self._logger.debug(
                 f"Rate Limited by Sharepoint: retry in {retry_seconds} seconds"
             )
 
+            await self._sleeps.sleep(retry_seconds)  # TODO: use CancellableSleeps
+            raise ThrottledError from e
+        elif (
+            e.status == 403 or e.status == 401
+        ):  # Might work weird, but Graph returns 403 and REST returns 401
+            raise PermissionsMissing(
```
wangch079, line 429: I just noticed `retryable_aiohttp_call` is defined in `sharepoint_online.py`. Do you think it's worth moving it to `utils.py`, together with `PermissionsMissing`? I don't think we want to retry if it's a permission error.

connectors/sources/sharepoint_online.py
```diff
@@ -1286,29 +1393,36 @@ async def get_docs(self, filtering=None):
...
-
-                            drive_item = self._decorate_with_access_control(
-                                drive_item, access_control
-                            )
-
-                            yield drive_item, self.download_function(
-                                drive_item, max_drive_item_age
-                            )
+                        for drive_items_batch in iterable_batches_generator(
+                            page.items, SPO_API_MAX_BATCH_SIZE
```
wangch079, line 1408: Given `SPO_API_MAX_BATCH_SIZE = 20`, to gain the maximum performance, we'd better get `page` from `drive_items` at a multiple of 20.

The delta query for drive item supports `$top` param, but seems we didn't specify it at all. I did a search in the official doc, but I couldn't find the default and maximum value for this param

connectors/sources/sharepoint_online.py
timgrein: AFAIU yes as it's an `@asynccontextmanager`.

connectors/sources/sharepoint_online.py
timgrein: Good point, I would also keep this for now separate and create a corresponding issue to deal with it later as getting the configurable permissions out is prio #1 right now.

connectors/sources/sharepoint_online.py
timgrein: Adjusted

connectors/sources/sharepoint_online.py
seanstory: ++. The sharepointID isn't typically easy for a customer to look up or reason about. But everyone knows their email address. This is why we decided to use email as the key (ES _id) for SPO.

connectors/sources/sharepoint_online.py
seanstory: Good call, Chenhui. Tim, can you create and link an issue so we don't forget?

timgrein: Created it last week: https://github.com/elastic/enterprise-search-team/issues/5181

connectors/sources/sharepoint_online.py
```diff
@@ -1299,29 +1407,36 @@ async def get_docs(self, filtering=None):
...
-                yield self._decorate_with_access_control(site, access_control), None
+                site_access_control = await self._site_access_control(site)
+                yield self._decorate_with_access_control(
+                    site, site_access_control
+                ), None
 
                 async for site_drive in self.site_drives(site):
                     yield self._decorate_with_access_control(
-                        site_drive, access_control
+                        site_drive, site_access_control
```
seanstory, line 1417: Can drives not be given more granular access than their parent site?

seanstory (approved):

connectors/sources/sharepoint_online.py
timgrein: They're treated as lists AFAIU and they can have unique permissions. Will be addressed in a separate PR 👍 Issue for tracking: https://github.com/elastic/enterprise-search-team/issues/5363

connectors/sources/sharepoint_online.py
```diff
@@ -351,12 +359,34 @@ async def scroll_delta_url(self, url):
...
 
     @asynccontextmanager
     @retryable_aiohttp_call(retries=DEFAULT_RETRY_COUNT)
-    async def _call_api(self, absolute_url):
+    async def _post(self, absolute_url, payload=None):
+        try:
+            token = await self._api_token.get()
+            headers = {"authorization": f"Bearer {token}"}
+            self._logger.debug(
+                f"Posting to Sharepoint Endpoint: {absolute_url}. Payload: {payload}"
```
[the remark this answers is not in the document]
timgrein: I wouldn't bet on it. If we want to be on the secure side I can remove `payload`.

seanstory (approved):

timgrein:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|pre-8.10-stable|[](https://github.com/elastic/connectors-python/pull/1287)|

### Questions ?

PR #1262, opened 2023-07-13
Use API key if both API key and basic auth are configured

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/4187

For Elasticsearch config, if both API key and basic auth are configured, API key will take precedence.

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -51,17 +51,17 @@ class ESClient:
         }
         logger.debug(f"Host is {self.host}")
 
-        if "username" in config:
-            if "api_key" in config:
-                raise KeyError(
-                    "You can't use basic auth ('username' and 'password') and 'api_key' at the same time in config.yml"
+        if "api_key" in config:
+            logger.debug(f"Connecting with an API Key ({config['api_key'][:5]}...)")
+            options["api_key"] = config["api_key"]
+            if "username" in config:
+                logger.debug(
+                    "API key will be used instead of basic auth ('username' and 'password')"
                 )
+        elif "username" in config:
             auth = config["username"], config["password"]
             options["basic_auth"] = auth
             logger.debug(f"Connecting using Basic Auth (user: {config['username']})")
-        elif "api_key" in config:
-            logger.debug(f"Connecting with an Api Key ({config['api_key'][:5]}...)")
-            options["api_key"] = config["api_key"]
 
         if config.get("ssl", False):
             options["verify_certs"] = True
--- a/tests/es/test_client.py
+++ b/tests/es/test_client.py
@@ -13,9 +13,11 @@ from elasticsearch import ConflictError, ConnectionError
 from connectors.es.client import ESClient, License, with_concurrency_control
 
 BASIC_CONFIG = {"username": "elastic", "password": "changeme"}
+API_CONFIG = {"api_key": "foo"}
+BASIC_API_CONFIG = {"username": "elastic", "password": "changeme", "api_key": "foo"}
 
 
-def test_esclient():
+def test_es_client():
     # creating a client with a minimal config should create one with sane
     # defaults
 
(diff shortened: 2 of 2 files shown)


timgrein (approved):

connectors/es/client.py
```diff
@@ -51,17 +51,17 @@ def __init__(self, config):
...
-            if "api_key" in config:
-                raise KeyError(
-                    "You can't use basic auth ('username' and 'password') and 'api_key' at the same time in config.yml"
+        if "api_key" in config:
+            logger.debug(f"Connecting with an API Key ({config['api_key'][:5]}...)")
+            options["api_key"] = config["api_key"]
+            if "username" in config:
+                logger.debug(
+                    "API key will be used instead of basic auth ('username' and 'password')"
                 )
```
artem-shelkovnikov, line 60: Maybe it should be a warning?

artem-shelkovnikov:
@wangch079 just to confirm, what's the purpose of this change?

I personally liked that we're forced to choose auth method, not fallback.

wangch079:
>just to confirm, what's the purpose of this change?

Intuitively, when both `api_key` and `username`/`password` are configured, I think `api_key` should take precedence. We also documented `api_ke` is prefered (but not clearly stated it would take precedence, I added it in PR #1243):

https://github.com/elastic/connectors-python/blob/24249c1c706bcb450cfb8643f3abd76ab2b1574a/docs/CONFIG.md?plain=1#L7

The main purpose of introducing this PR, is that when executing a sync job, we will override the `api_key` from the connector in the `elasticsearch` config, and we want to use `api_key` even when `username` and `password` are configured. 

>I personally liked that we're forced to choose auth method, not fallback.

I don't hold the opinion strongly. I can actually discard this PR, and remove `username` and `password when overriding.

cc @artem-shelkovnikov

artem-shelkovnikov:
> The main purpose of introducing this PR, is that when executing a sync job, we will override the api_key from the connector in the elasticsearch config, and we want to use api_key even when username and password are configured.

But technically this change does not achieve this - you can still have username/password for the framework and api_key for the ingestion?

wangch079:
>But technically this change does not achieve this - you can still have username/password for the framework and api_key for the ingestion?

E.g. the config is like:
```yaml
elasticsearch:
  username: username
  password: password

connectors:
  -
    connector_id: foo
    service_type: mysql
    api_key: bar
```

When syncing data for connector `foo`, I will override and pass the following es config to `SyncJobRunner`:
```json
{
  "username": "username",
  "password": "password",
  "api_key": "bar"
}
```

And I expect it to index data into ES with API key `bar`, rather than raising an error `You can't use basic auth ('username' and 'password') and 'api_key' at the same time in config.yml`

But if you like to enforce the customers to choose one, this PR is not needed, and I can remove `username` and `password` when overriding: https://github.com/elastic/connectors-python/pull/1275/files#diff-73ae8f3cdbe21770bec2ce0a72bb40c0aaf2b1191abfcbc81661d77940f2f656R228-R229

artem-shelkovnikov:
> When syncing data for connector foo, I will override and pass the following es config to SyncJobRunner:

But you can also do override while ignoring username/password?

I'm not against the change, but thinking how usable it will be - I find username/password useful for localhost testing, but once testing elsewhere, especially on cloud, it's mostly api_key for me. I feel like we can make api_key management better and switch to them fully?

Imagine a page in our section that allows you to create api_key that is needed only for service functioning (e.g. access only to connectors/sync job indices) + more granular api_keys that allow access to multiple content indices or only one.

wangch079:
>But you can also do override while ignoring username/password?

Yes, that's what I'm doing here: https://github.com/elastic/connectors-python/pull/1275/files#diff-73ae8f3cdbe21770bec2ce0a72bb40c0aaf2b1191abfcbc81661d77940f2f656R228-R229. i.e. when overriding `api_key`, remove `username` and `password`. In this case, this PR is not needed.

I will close this PR as this is not strictly required for the story [Support multiple connector clients in connector service](https://github.com/elastic/enterprise-search-team/issues/4187). And it worth a separate story if we want to make the change.

PR #1264, opened 2023-07-13
[SPO] Add configurable unique page permissions support

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/5179

This PR enables the SPO connector to fetch unique page permissions.
Unique page permissions ignore the parent site permissions.
Added a new RFC `fetch_unique_page_permissions` to make this behavior configurable.

A separate PR for Kibana will follow adding the RFC to the native SPO connector.

Note: the order `10` is on purpose as https://github.com/elastic/connectors-python/pull/1259 adds another RFC with order `9`.

Batch requests for improving performance will be addressed in a separate PR.

#### Pre-Review Checklist


artem-shelkovnikov:
@timgrein have you done minor performance benchmark by chance, do you know how big is the impact?

connectors/sources/sharepoint_online.py
```diff
@@ -1005,6 +1019,15 @@ def get_default_configuration(cls):
...
             },
+            "fetch_unique_page_permissions": {
+                "depends_on": [{"field": "use_document_level_security", "value": True}],
+                "display": "toggle",
+                "label": "Fetch unique page permissions",
+                "order": 10,
+                "tooltip": "Enable this option to fetch unique page permissions. This setting can increase sync time. If this setting is disabled a page will inherit permissions from its parent site.",
+                "type": "bool",
+                "value": True,
+            },
```
artem-shelkovnikov, line 1030: I've just recently merged functional tests for SPO - can you add this field into the `connector.json` for it?

timgrein: It's already added 👍

artem-shelkovnikov: Additionally (I commented in another issue), you might wanna decrement order by 1 cause of https://github.com/elastic/connectors-python/pull/1255#pullrequestreview-1528292491

timgrein:
> @timgrein have you done minor performance benchmark by chance, do you know how big is the impact?

No I haven't yet . As this is configurable there shouldn't be a performance degrade when having it disabled (there are no additional requests, when having it disabled). I'll do performance tests after having every permission granularity level up as PR and/or merged. As granular permissions are guarded through a RFC I don't expect a performance degrade, when having granular permissions disabled.

connectors/sources/sharepoint_online.py
timgrein: Good point, will adjust it in this PR and in https://github.com/elastic/connectors-python/pull/1259

connectors/sources/sharepoint_online.py
```diff
@@ -1572,12 +1586,90 @@ async def site_lists(self, site):
...
+            user_principal_name = user_.get("UserPrincipalName")
+            login_name = _get_login_name(user_.get("LoginName"))
+
+            if user_principal_name:
+                user_access_control.append(_prefix_user(user_principal_name))
+
+            if login_name:
+                user_access_control.append(_prefix_user(login_name))
+
+            return user_access_control
```
artem-shelkovnikov, line 1614: I'm curious about the practices of making methods method-private or module-private. Why is this one method-private particularly?

connectors/sources/sharepoint_online.py
```diff
@@ -1572,12 +1586,90 @@ async def site_lists(self, site):
...
             ]  # Apparently site_page["GUID"] is not globally unique
             site_page["object_type"] = "site_page"
+            has_unique_role_assignments = site_page.get(
+                "HasUniqueRoleAssignments", False
+            )
+
+            # ignore parent site permissions and use unique per page permissions ("unique permissions" means breaking the inheritance to the parent site)
+            if (
+                self.configuration["fetch_unique_page_permissions"]
+                and has_unique_role_assignments
```
artem-shelkovnikov, line 1651: Nice optimization!

artem-shelkovnikov (dismissed):
Looks good! Left a couple small questions, but overall feels like change does not also introduce high performance impact, if flag is enabled

connectors/sources/sharepoint_online.py
timgrein: Reasoning: I'm keeping it method-private here as I'm only using it inside the method and not outside of it. Don't know, if that's pythonic, but it makes sense to me with regards to "locality".

connectors/sources/sharepoint_online.py
```diff
@@ -867,6 +880,13 @@ def is_domain_group(user_fields):
     )
 
 
+def is_dynamic_group(login_name):
+    if not login_name:
+        return False
+
+    return login_name.startswith("c:0o.c|federateddirectoryclaimprovider")
```
seanstory, line 887: I don't think this is indicative of dynamic groups, but just of "groups". See: https://github.com/elastic/connectors-python/blob/main/connectors/sources/sharepoint_online.py#L862-L867

seanstory (approved):

connectors/sources/sharepoint_online.py
```diff
@@ -929,6 +942,13 @@ def is_domain_group(user_fields):
...
```
jedrazb, line 949: nit: should we stay consistent when using `startswith()` and include `|` at the end of the string pattern (same as in `_get_login_name`)? so it would be:
```
login_name.startswith("c:0o.c|federateddirectoryclaimprovider")
```

jedrazb (approved):
Code-wise LGTM  🚢 (I'm not an expert about SPO logic but I see that Artem and Sean approved already) one small nit but feel free to ignore

connectors/sources/sharepoint_online.py
timgrein: It seems it's the combination of `odata.type` being `SP.User` and the `LoginName` starting with `c:0o.c|federateddirectoryclaimprovider`. We can probably keep it local to the caller method. In the method's context it makes more sense.

timgrein:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|pre-8.10-stable|[](https://github.com/elastic/connectors-python/pull/1292)|

### Questions ?

PR #1266, opened 2023-07-13
Use blocklist for retry policy + make user of aiohttp limiter

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/5180

Two small changes bundled together:

- Inverted the retry policy for MicrosoftAPISession - it specifically does not retry `NotFound` and `ClientResponseError`, retried everything else (before it retried 4 specific error types, see changeset)
- Remove the semaphore code and make use of `aiohttp.TCPConnector(limit=X)` to limit aiohttp client maximum consequent call to X

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -62,7 +62,7 @@ else:
 
 DEFAULT_RETRY_COUNT = 3
 DEFAULT_RETRY_SECONDS = 30
-DEFAULT_PARALLEL_CONNECTION_COUNT=10
+DEFAULT_PARALLEL_CONNECTION_COUNT = 10
 FILE_WRITE_CHUNK_SIZE = 1024
 MAX_DOCUMENT_SIZE = 10485760
 WILDCARD = "*"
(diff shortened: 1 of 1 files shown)


connectors/sources/sharepoint_online.py
```diff
@@ -282,14 +283,11 @@ async def wrapped(*args, **kwargs):
...
                         yield item
                     break
-                except (
-                    ThrottledError,
-                    PermissionsMissing,
-                    InternalServerError,
-                ) as e:
+                except (NotFound, ClientResponseError):
+                    raise
+                except Exception:
```
vidok, line 288: nit: It would be also useful to handle `ClientOSError` and log the error that MS API closed the session.

artem-shelkovnikov: Do you still want to retry it?

artem-shelkovnikov: Do you still want to retry it?

I think the fact that MS API closed the session is a sign, that connector is overthrottling and can potentially be banned from Graph API access actually. I'd probably throw with a nice error message

vidok (approved):

PR #1273, opened 2023-07-13
[SPO] Limit synced fields

seanstory:
## Closes https://github.com/elastic/connectors-python/issues/1268

By using explicit `$select` clauses, we both increase performance and limit what types of fields we index. 

This change adds a few explicit select clauses, and reduces our total fields in a full sync of the koffelab dataset from ~224 to ~70 fields (not counting nested subfields).

There's a good chance that we'll eventually want to make the selected fields configurable.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -579,7 +579,7 @@ class SharepointOnlineClient:
                 yield site
 
     async def site_drives(self, site_id):
-        select = ""
+        select = "createdDateTime,description,id,lastModifiedDateTime,name,webUrl,driveType,createdBy,lastModifiedBy,owner"
 
         async for page in self._graph_api_client.scroll(
             f"{GRAPH_API_URL}/sites/{site_id}/drives?$select={select}"
@@ -618,7 +618,7 @@ class SharepointOnlineClient:
         )
 
     async def site_lists(self, site_id):
-        select = ""
+        select = "createdDateTime,id,lastModifiedDateTime,name,webUrl,displayName,createdBy,lastModifiedBy"
 
         async for page in self._graph_api_client.scroll(
             f"{GRAPH_API_URL}/sites/{site_id}/lists?$select={select}"
@@ -639,8 +639,8 @@ class SharepointOnlineClient:
             return {}
 
     async def site_list_items(self, site_id, list_id):
-        select = ""
-        expand = "fields"
+        select = "createdDateTime,id,lastModifiedDateTime,weburl,createdBy,lastModifiedBy,contentType"
+        expand = "fields($select=Title,Link,Attachments,LinkTitle,LinkFilename,Description,Conversation)"
 
         async for page in self._graph_api_client.scroll(
             f"{GRAPH_API_URL}/sites/{site_id}/lists/{list_id}/items?$select={select}&$expand={expand}"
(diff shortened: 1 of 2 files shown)


connectors/sources/sharepoint_online.py
```diff
@@ -698,6 +698,10 @@ async def site_pages(self, site_web_url):
             # I'm not sure if site can have no pages, but given how weird API is I put this here
             # Just to be on a safe side
             return
+        except ClientResponseError:
+            # Some site pages don't have the required columns, and it's unclear why
```
seanstory, line 702: DM me for examples of this, if you're interested. I'm moderately concerned this is a bug, but IDK.

connectors/sources/sharepoint_online.py
```diff
@@ -1462,7 +1466,6 @@ async def site_list_items(
             list_item_natural_id = list_item["id"]
             list_item["_id"] = f"{site_list_id}-{list_item['id']}"
             list_item["object_type"] = "list_item"
-            list_item["_original_filename"] = list_item.get("FileName", "")
```
seanstory, line 1465: I was getting errors that ListItems could not select `FileName` and that this wasn't a valid field. So I deleted this usage, since it would always lead to a `""` value

timgrein:
> There's a good chance that we'll eventually want to make the selected fields configurable.

Should we just do it directly (did we check with product)? Sounds like a low-hanging fruit, if we only do it for SPO (we could also think about a potential more generic framework-level solution). I could definitely see additional value besides performance.

connectors/sources/sharepoint_online.py
artem-shelkovnikov: Should we just keep the select empty to not break the sync for pages of the site?

connectors/sources/sharepoint_online.py
artem-shelkovnikov: Nice catch!

seanstory:
> > There's a good chance that we'll eventually want to make the selected fields configurable.
> 
> Should we just do it directly (did we check with product)?

We did loop in product. See this thread: https://elastic.slack.com/archives/C04UD57HPA7/p1688999960899709?thread_ts=1688513369.754399&cid=C04UD57HPA7
Specifically:

```
we'll limit the ingested fields to the default ones and ignore custom fields. Custom fields could be ingested again by customizing the connector and specifying the allowed custom fields to index.
```

> Sounds like a low-hanging fruit

I disagree, only because adding RCFs at this stage is a little fraught since it requires backports in Kibana too. This is a good opportunity for enhancement in a future version, especially if we get feedback that customizations are needed. But if we can pick a good set of defaults that don't need customizing, maybe we avoid cluttering our configs with more choicepoints.

connectors/sources/sharepoint_online.py
```diff
@@ -1506,7 +1522,7 @@ async def site_pages(self, url):
             ]  # Apparently site_page["GUID"] is not globally unique
             site_page["object_type"] = "site_page"
 
-            for html_field in ["LayoutWebpartsContent", "CanvasContent1"]:
+            for html_field in ["LayoutWebpartsContent", "CanvasContent1", "WikiField"]:
```
seanstory, line 1525: In DMs with Artem, we realized that some SitePages have `LayoutWebPartsContent` and `CanvasContent1` in their columns, but others explicitly do not have these fields (not even with null values) and instead have their HTML in `WikiField`. So we're just sneaking that bugfix in this PR.

seanstory:
There aren't good unit tests for this, because it doesn't seem valuable to me to try to demonstrate that we're passing specific query params to a 3rd-party rest API. I'd just mock the request and response, and not be testing anything. The only logic change did get an updated test though.

connectors/sources/sharepoint_online.py
```diff
@@ -283,7 +283,7 @@ async def wrapped(*args, **kwargs):
                     async for item in func(*args, **kwargs):
                         yield item
                     break
-                except (NotFound, ClientResponseError):
```
seanstory, line 286: Slack thread: https://elastic.slack.com/archives/C01795T48LQ/p1689349522792619

PR #1289, opened 2023-07-18
[GitHub] GraphQL implementation

abhishekjoshi-crest:
## Closes https://github.com/elastic/connectors-py/issues/1288

This PR contains the GitHub connector implementation using GraphQL API.

**Note:** For GitHub GraphQL queries I have created new file "graphql_queries.py" and created class for it. Also, in future if any connector uses GraphQL then they can add queries in the same file.

**Note:** Also for the simplicity I have created new file github_graphql.py. Once the reviewer approved the changes I replaced github_graphql.py to github.py.

#### Pre-Review Checklist


connectors/sources/graphql_queries.py
```diff
@@ -0,0 +1,322 @@
+from enum import Enum
+
+
+class GithubQuery(Enum):
```
timgrein, line 4: suggestion: As this is specific to GitHub I would keep it in `github.py`. I don't expect this queries to be reused across different modules, but to be only used inside the `github.py` module.

connectors/sources/github_graphql.py
```diff
@@ -0,0 +1,889 @@
...
+                ) from exception
+        except Exception:
+            raise
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def get_response(self, resource):
```
timgrein, line 190: Should we call it `get_item` or `get_github_item` for more specificity?

connectors/sources/github_graphql.py
```diff
@@ -0,0 +1,889 @@
...
+            async for repo in self.github_client.get_user_repos():
+                self.user_repos[repo["nameWithOwner"]] = repo
+            invalid_repos = list(set(configured_repos) - set(self.user_repos.keys()))
+
+            for repo_name in foreign_repos:
+                try:
+                    self.foreign_repos[
+                        repo_name
+                    ] = await self.github_client.get_foreign_repo(repo_name=repo_name)
+                except Exception:
```
timgrein, line 415: Maybe it's worth to add a debug log line like `f"Detected invalid repository: {repo_name}"`

connectors/sources/github_graphql.py
```diff
@@ -0,0 +1,889 @@
...
+    async def ping(self):
+        try:
+            await self.github_client.ping()
+            self._logger.debug("Successfully connected to GitHub.")
+        except Exception:
+            self._logger.exception("Error while connecting to GitHub.")
+            raise
+
+    def adapt_gh_doc_to_es_doc(self, github_document, schema):
+        document = {}
```
timgrein, line 464: nit: that could be done using a dict comprehension. Feel free to leave it like this.

connectors/sources/graphql_queries.py
timgrein: I don't have any concern that the `github.py` module contains a high number of LOC. If we've more connectors using GraphQL it's likely that the `graphql_queries.py` will be huge and contain code from many unrelated data sources/connectors. Keeping high cohesion in mind I think it's fair to say, that github specific code should live inside the same module, if there's no need to share the code.

abhishekjoshi-crest: okay, moved queries in `github.py` file

timgrein (approved):
LGTM, thank for addressing the suggested changes!

PR #1298, opened 2023-07-18
Slack connector

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/5073

Adds a slack connector. Credentials to test can be found in vault. 

This connector syncs only users, public channels, and messages from public channels. The authenticating slack bot must be present in a channel to sync it, or must be authorized and configured to invite itself to channels.

#### Pre-Review Checklist

## Release Note

tech-preview Slack Connector client added


connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+        self._http_session.close()
+        return
+
+    async def list_channels(self, only_my_channels, cursor=None):
+        conversation_batch_limit = 200
+        self._logger.debug("Iterating over all channels")
+        if only_my_channels:
+            self._logger.debug("Will only yield channels the bot belongs to")
+        while True:
+            url = f"{BASE_URL}/{ENDPOINTS[LIST_CONVERSATIONS]}?limit={conversation_batch_limit}"
```
artem-shelkovnikov, line 103: I feel like constants here make it much harder to read, why not inline whole `ENDPOINTS` hash? (`BASE_URL` could be useful still for functional tests)

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    def ping(self):
+        return
+
+    def close(self):
+        self._http_session.close()
+        return
```
artem-shelkovnikov, line 95: No need to return

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+            timeout=aiohttp.ClientTimeout(total=None),
+            raise_for_status=True,
+        )
+        self._sleeps = CancellableSleeps()
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    def ping(self):
+        return
```
artem-shelkovnikov, line 91: Is there a way to ping Slack's health?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def close(self):
+        self._http_session.close()
```
artem-shelkovnikov, line 94: You also need to stop `self._sleeps` here

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+        return
+
+    async def list_channels(self, only_my_channels, cursor=None):
```
artem-shelkovnikov, line 97: I see `cursor` is not passed in this method ever, do we need it here?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+        conversation_batch_limit = 200
```
artem-shelkovnikov, line 98: That could be a good constant/setting

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+                yield channel
+            cursor = self._get_next_cursor(response)
+            if not cursor:
+                break
+
+    async def join_channel(self, channel_id):
+        url = f"{BASE_URL}/{ENDPOINTS[JOIN_CONVERSATIONS]}?channel={channel_id}"
+        return await self._get_json(url)
+
+    async def list_messages(self, channel_id, oldest, latest):
```
artem-shelkovnikov, line 121: I see `latest` is always `time.now`, do we want to expose this parameter?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+                if message.get("type") == "message":
+                    if message.get("reply_count", 0) > 0:
+                        async for thread_message in self.list_thread_messages(
+                            channel_id, message["ts"]
+                        ):
+                            yield thread_message
+                    else:
+                        yield message
+            if not response.get("has_more"):
+                break
```
artem-shelkovnikov, line 140: How does it work? do you issue same request but it gives you back different messages?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+
+    async def list_thread_messages(self, channel_id, parent_ts):
+        message_batch_limit = 200
```
artem-shelkovnikov, line 143: Can be same constant as everywhere

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+            f"Fetching messages between {oldest} and {latest} from channel: '{channel_id}'"
+        )
+        message_batch_limit = 200
+        while True:
+            url = f"{BASE_URL}/{ENDPOINTS[CONVERSATION_HISTORY]}?channel={channel_id}&limit={message_batch_limit}&oldest={oldest}&latest={latest}"
+            response = await self._get_json(url)
+            for message in response.get("messages", []):
+                latest = message.get("ts")
+                if message.get("type") == "message":
+                    if message.get("reply_count", 0) > 0:
```
artem-shelkovnikov, line 132: Hmmm. what about this `retry_count`? What's that logic? Curious!

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+
+    async def join_channel(self, channel_id):
+        url = f"{BASE_URL}/{ENDPOINTS[JOIN_CONVERSATIONS]}?channel={channel_id}"
+        return await self._get_json(url)
+
+    async def list_messages(self, channel_id, oldest, latest):
+        self._logger.info(
+            f"Fetching messages between {oldest} and {latest} from channel: '{channel_id}'"
+        )
+        message_batch_limit = 200
```
artem-shelkovnikov, line 125: Can go to same constant as every other place

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+            resonse = await self._get_json(url)
+            for message in resonse["messages"]:
+                if message["type"] == "message":
+                    yield message
+            cursor = self._get_next_cursor(resonse)
+            if not cursor:
+                break
+
+    async def list_users(self, cursor=None):
+        user_batch_limit = 200
```
artem-shelkovnikov, line 158: Same - can be a constant

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+                yield member
+            cursor = self._get_next_cursor(response)
+            if not cursor:
+                break
+
+    def _add_cursor(self, url, cursor):
+        return f"{url}&{CURSOR}={cursor}"
+
+    def _get_next_cursor(self, response):
+        return response.get(RESPONSE_METADATA, {}).get(NEXT_CURSOR)
```
artem-shelkovnikov, line 174: This can be inlined, IMO a little duplication does not hurt in this case

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+                    break
+                except Exception as e:
+                    if retry >= retries:
+                        raise e
+
+                    retry += 1
+
+        return wrapped
+
+    return wrapper
```
artem-shelkovnikov, line 68: I think this code is identical to the code in `utils.py::retryable`? Can we just use `@retryable` from utils?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+        if user_profile.get("display_name_normalized"):
+            return user_profile["display_name_normalized"]
+        elif user_profile.get("real_name_normalized"):
+            return user_profile["real_name_normalized"]
+        elif user.get("real_name"):
+            return user["real_name"]
+        elif user.get("name"):
+            return user["name"]
+        else:
+            return user["id"]
```
artem-shelkovnikov, line 333: It's possible to have users with no name?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+        channel["_id"] = channel["id"]
+        channel["type"] = "channel"
+        return channel
+
+    def adapt_message(self, message, channel, usernames):
+        user_id = message.get("user", message.get("bot_id"))
+        message["author"] = usernames.get(user_id, user_id)
+        message["text"] = re.sub(
+            r"<@([A-Z0-9]+)>", self.convert_usernames, message["text"]  # type: ignore
+        )
```
artem-shelkovnikov, line 404: Idk how it works in Python 100%, but I think we might wanna compile this regex to speed things up. But might not be a big deal too?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+                "latest_reply",
+                "thread_ts",
+            ),
+        )
+        message["_id"] = message["client_msg_id"]
+        return message | {"channel": channel["name"]}
+
+    def convert_usernames(self, match_obj):
+        if match_obj.group(1) is not None:
+            return f"<@{self.usernames.get(match_obj.group(1), match_obj.group(1))}>"
```
artem-shelkovnikov, line 426: This method is really confusing - comments and variable assignments can help with it. Also it's used in a single place, might be worth to make it an inline function right above the place where it's called.

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+
+    def _dict_slice(self, hsh, keys):
+        return {k: hsh.get(k) for k in keys}
```
artem-shelkovnikov, line 429: Can go to utils if needed

tests/sources/test_slack.py
```diff
@@ -0,0 +1,190 @@
...
+    yield client
+    client.close()
+
+
+@pytest.fixture
+def slack_data_source():
+    source = create_source(SlackDataSource, **configuration)
+    source.set_logger(logger)
+    yield source
+    source.close()
```
artem-shelkovnikov, line 37: You'll need to mark them both with `pytest_asyncio.fixture`, otherwise you'll get tons of warnings. We don't enforce tests to not break when warnings are emitted, but it'll happen really soon

artem-shelkovnikov (commented):
Good stuff! Left a couple comments, but looks pretty clean and nice :)

connectors/sources/slack.py
seanstory: I like using constants for api paths because my chances of typos-that-my-IDE-misses goes way down. Nothing more frustrating than a 404 bug that you track back to a misspelling. 

Would it be more readable if it weren't a dict of endpoints, but just a a bunch of constants, to make it like:
```
f"{BASE_URL}/{LIST_CONVERSATIONS_ENDPOINT}?"
```
?

connectors/sources/slack.py
seanstory: It lets us keep the unix timestamp conversion logic in the calling code, and not the client code. 🤷 But I don't feel strongly about it, if you'd prefer not to have the param.

connectors/sources/slack.py
seanstory: What's changed in the request is the `latest`. See: https://github.com/elastic/connectors-python/pull/1298/files/eda90bee5e6ad4bedd200021cec9d13d7986b3ea#diff-4fc38a422f0713e89a903dce26cc081ef3afee931b5ebfdb8fab684db30c5681R130

Weirdly, slack lets you paginate by cursor, or by timestamp, but not by a combination.

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+
+    async def list_thread_messages(self, channel_id, parent_ts):
+        message_batch_limit = 200
+        cursor = None
+        while True:
+            url = f"{BASE_URL}/{ENDPOINTS[CONVERSATION_REPLIES]}?channel={channel_id}&ts={parent_ts}&limit={message_batch_limit}"
+            if cursor:
+                url = self._add_cursor(url, cursor)
+            resonse = await self._get_json(url)
+            for message in resonse["messages"]:
```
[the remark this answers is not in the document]
seanstory: heh. nice find.

connectors/sources/slack.py
seanstory: I'm confused. I _think_ you were noting that I stole your retry logic from Sharepoint (which I did, shamelessly. And not just that logic). But this is _reply_ count. 

This was the indicator I chose for if a given message is the parent for a thread. If it has replies, then you have to make a follow-up request(s) to get all those replies, since they're not part of `conversation.history`

connectors/sources/slack.py
seanstory: I messed up the logic once and decided that I didn't want to have to retype this invocation multiple times and keep risking making another bug. What's the downside, in your opinion, to having this in a helper function?

connectors/sources/slack.py
seanstory: This is actually copied from the SPO connector. Should it also be deduped in there too?

connectors/sources/slack.py
seanstory: Yep! Bots are an example.

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+        elif user_profile.get("real_name_normalized"):
+            return user_profile["real_name_normalized"]
+        elif user.get("real_name"):
+            return user["real_name"]
+        elif user.get("name"):
+            return user["name"]
+        else:
+            return user["id"]
+
+    def adapt_user(self, user):
```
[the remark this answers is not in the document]
seanstory: > Adapter pattern is about class ~interface~ compatibility

I think of it very much like "power adapters". You see one shape on the wall, you slap an adapter on it, and now it's a different shape.

I like using "adapt" when _reshaping_ data from one data source paradigm to another. And it's the language we used in WS. I think that I saw it being used in another recent connector too, but struggling to find that now.

But that doesn't mean we have to use it, just expressing my rationale. 

I'd go for `reshape_*` or `transform_*` or `remap_*`, if you feel strongly about one?

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+        user_profile = user.get("profile", {})
+        user["real_name_normalized"] = user_profile.get("real_name_normalized")
+        user["profile_real_name"] = user_profile.get("real_name")
+        user["display_name"] = user_profile.get("display_name")
+        user["display_name_normalized"] = user_profile.get("display_name_normalized")
+        user["first_name"] = user_profile.get("first_name")
+        user["last_name"] = user_profile.get("last_name")
```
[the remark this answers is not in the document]
seanstory: There's not a consistent subset that's non-null. So I wanted to grab everything, so that there's _something_ stored, but I don't have to make a judgement call on which is preferred for search.

connectors/sources/slack.py
seanstory: IDK how it works in python either. I stole this approach from SO, but I can definitely put in more research. Then maybe I can remove the `# type: ignore` (pyright was unhappy)

connectors/sources/slack.py
artem-shelkovnikov: Oh wow, I read it as a "retry", my bad 😮‍💨

connectors/sources/slack.py
artem-shelkovnikov: Oh wow, I see. But oldest does not change, so it's not a rolling window?

connectors/sources/slack.py
artem-shelkovnikov: I think it's okay to have this function, I just find one-liners like this confusing me when reading/reviewing, thought train is: "hmmm it's a separate function, probably lots of logic inside". Also I personally don't like methods that only encapsulate field access. It's definitely okay to have this method like this, just being picky.

connectors/sources/slack.py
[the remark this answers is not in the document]
artem-shelkovnikov: I see, I also don't have super hard opinion here, `adapt_` works, but I also like `transform_` or `remap_` a bit better :)

connectors/sources/slack.py
```diff
@@ -0,0 +1,429 @@
...
+            not self.auto_join_channels
+        ):
+            self._logger.info(f"Listed channel: '{channel['name']}'")
+            yield self.adapt_channel(channel, usernames)
+            channel_id = channel["id"]
+            join_response = {}
+            if self.auto_join_channels:
+                join_response = await self.slack_client.join_channel(
+                    channel_id
+                )  # can't get channel content if the bot is not in the channel
```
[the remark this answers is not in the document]
seanstory: we'd get 403's and get error logs for free in such a case.

connectors/sources/slack.py
seanstory: Right - `oldest` is set by configuration (default is the last 180 days). So it needs to be a param so that configuration can have an impact, but for a given run, while paging, `oldest` stays the same, while we keep moving `latest` farther back in time until they are equal (or near enough that there are no messages between them).

connectors/sources/slack.py
seanstory: ha, ok. If you don't feel strongly, I'm going to leave this one, so that the `, {}` doesn't get missed the next time we add page handling, or something.

seanstory:
@artem-shelkovnikov I believe I've addressed and resolved all but two comments. The last two I don't have strong feelings on, but am also not sure if you want changes or are just double-checking logic.

Since I'm headed out on PTO, please feel free to push any additional changes that are necessary, and/or merge before I get back. But there's no rush, if you want to talk over any more points, I'll be back next week.

connectors/sources/slack.py
```diff
@@ -0,0 +1,382 @@
...
+        try:
+            async with self._http_session.get(
+                absolute_url,
+                headers=self._headers(),
+            ) as resp:
+                yield resp
+        except ClientResponseError as e:
+            await self._handle_client_response_error(absolute_url, e)
+
+    async def _handle_client_response_error(self, absolute_url, e):
```
praveen-elastic, line 159: Do we need absolute_url ? It is not being used in the method.

connectors/sources/slack.py
```diff
@@ -0,0 +1,382 @@
...
+            ),
+        )
+
+    def remap_message(self, message, channel):
+        user_id = message.get("user", message.get("bot_id"))
+
+        def convert_usernames(
+            match_obj,
+        ):  # used below in an re.sub to map on matching messages
+            id = match_obj.group(1)
```
praveen-elastic, line 362: I think we should not assign to built-in variable `id`

praveen-elastic (commented):
Just a few nits, great work. The bifurcations in tests was pretty nice.

connectors/sources/slack.py
artem-shelkovnikov: Good call, removed

connectors/sources/slack.py
artem-shelkovnikov: Good call, renamed

artem-shelkovnikov (approved):

connectors/sources/slack.py
```diff
@@ -0,0 +1,393 @@
...
+            "real_name_normalized": user_profile.get("real_name_normalized"),
+            "profile_real_name": user_profile.get("real_name"),
+            "display_name": user_profile.get("display_name"),
+            "display_name_normalized": user_profile.get("display_name_normalized"),
+            "first_name": user_profile.get("first_name"),
+            "last_name": user_profile.get("last_name"),
+        } | dict_slice(
+            user,
+            (
+                "id",
```
wangch079, line 327: Is it duplicated?

connectors/sources/slack.py
```diff
@@ -0,0 +1,393 @@
...
+                headers=self._headers(),
+            ) as resp:
+                yield resp
+        except ClientResponseError as e:
+            await self._handle_client_response_error(e)
+
+    async def _handle_client_response_error(self, e):
+        if e.status == 429:
+            response_headers = e.headers or {}
+            if "Retry-After" in response_headers:
```
wangch079, line 167: Feel this can be added to `retryable`, this will be checked, and sleep if necessary, then apply the retry strategy.

connectors/sources/slack.py
```diff
@@ -0,0 +1,393 @@
...
+            raise_for_status=True,
+        )
+        self._logger = logger
+        self._sleeps = CancellableSleeps()
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    def ping(self):
+        return self.test_auth()
```
wangch079, line 58: Wondering why defining another `test_auth` method.

connectors/sources/slack.py
wangch079: You are not alone.

connectors/sources/slack.py
seanstory: Because the endpoint is "auth.test", so the client should have a "test auth" method. It's just an implementation detail that ping (a framework convention of ours) uses the test_auth method.

connectors/sources/slack.py
seanstory: yes, both in `id` and in `_id`. But this is intentional, so that an ES response of _source still contains the relevant slack metadata.

artem-shelkovnikov (approved):

connectors/sources/slack.py
wangch079: In Extractor, the `_id` is poped and stored as `id`:

https://github.com/elastic/connectors-python/blob/a03cc1c9ba9c0d086ccd079b49c24eec5ec86c79/connectors/es/sink.py#L392C17-L392C52

So the `id` here will be eventually overwritten by `_id`, which is essentially, a no-op.

This is happening today. For any connector, you can check the documents in the index `search-*`, and you will find duplicated `id` field and document id `_id`. I don't feel this is right. Just FYI

connectors/sources/slack.py
seanstory: issue for DRYing up the client: https://github.com/elastic/connectors-python/issues/1304

PR #1299, opened 2023-07-19
[SPO] Reduce number of HasUniqueRoleAssignments requests (guard with config)

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/5382

We're currently fetching `HasUniqueRoleAssignments` for every `site_page` even if the RFC `fetch_unique_page_permissions` is not enabled. This change halves the number of requests for site pages, if `fetch_unique_page_permissions` is disabled.

#### Pre-Review Checklist


artem-shelkovnikov (dismissed):

connectors/sources/mysql.py
```diff
@@ -331,6 +331,22 @@ def generate_id(tables, row, primary_key_columns):
         f"{'_'.join([str(pk_value) for pk in primary_key_columns if (pk_value := row.get(pk)) is not None])}"
     )
 
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def get_last_update_time(self, table):
```
jedrazb, line 339: I see this function `get_last_update_time` added in this PR but I can't see it used anywhere. Is it relevant to the changes for SPO? 

If it results from a merge conflict resolution - I think something is off since we are explicitly adding this code to `main` now?

timgrein: Oops, that was my fault yes, good catch! Reverting it

PR #1300, opened 2023-07-19
[SPO] Deduplicate test name

timgrein:
We had two tests using the same name causing a linter warning. This PR deduplicates the test names.

#### Pre-Review Checklist


seanstory (approved):
my linter used to complain about this if I ran `make lint` while in the venv. No idea why it doesn't catch it outside of venv. 

Thanks for fixing.

timgrein:
> my linter used to complain about this if I ran `make lint` while in the venv. No idea why it doesn't catch it outside of venv.
> 
> Thanks for fixing.

Maybe it was a warning? Would be fixed with https://github.com/elastic/connectors-python/pull/1193.

PR #1306, opened 2023-07-19
[SPO] Add unique list permissions support

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/5363

This PR adds support for unique site list permissions, which broke the inheritance to their parent site. This is guarded through a new RFC `fetch_unique_list_permissions`.

Also added some missing tests around `site_lists` and extended the `AsyncIterator` with some new assertions (which are needed for the `site_lists` tests) and added missing tests for this class, too.

Note: a separate PR will be opened for the Kibana RFC

#### Pre-Review Checklist

### DIFF
--- a/tests/commons.py
+++ b/tests/commons.py
@@ -1,3 +1,8 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
 class AsyncIterator:
     """
     Async documents generator fake class, which records the args and kwargs it was called with.
@@ -8,6 +13,7 @@ class AsyncIterator:
         self.call_args = []
         self.call_kwargs = []
         self.i = 0
+        self.call_count = 0
 
     def __aiter__(self):
         return self
@@ -21,6 +27,8 @@ class AsyncIterator:
         return item
 
     def __call__(self, *args, **kwargs):
+        self.call_count += 1
+
         if args:
             self.call_args.append(args)
 
@@ -28,3 +36,9 @@ class AsyncIterator:
             self.call_kwargs.append(kwargs)
 
         return self
+
+    def assert_not_called(self):
+        return self.call_count == 0
+
+    def assert_called_once(self):
+        return self.call_count == 1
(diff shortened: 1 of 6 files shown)


tests/commons.py
```diff
@@ -1,3 +1,8 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
```
seanstory, line 5: Huh. We probably need a test to ensure that this appears on all our files. 
Nice catch.

tests/commons.py
```diff
@@ -21,10 +27,18 @@ async def __anext__(self):
...
         if kwargs:
             self.call_kwargs.append(kwargs)
 
         return self
+
+    def assert_not_called(self):
+        return self.call_count == 0
+
+    def assert_called_once(self):
+        return self.call_count == 1
```
seanstory, line 44: I think this capability exists already: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

tests/commons.py
```diff
@@ -21,10 +27,18 @@ async def __anext__(self):
...
         if args:
             self.call_args.append(args)
 
         if kwargs:
             self.call_kwargs.append(kwargs)
 
         return self
+
+    def assert_not_called(self):
+        return self.call_count == 0
```
seanstory, line 41: It baffles me that this isn't part of pytest. I've run across https://stackoverflow.com/questions/54838354/python-how-can-i-assert-a-mock-object-was-not-called-with-specific-arguments before. Lots of ways to hack around it I guess.

tests/test_commons.py
seanstory, line 1: Good ol' meta-tests. :D

seanstory (approved):

tests/commons.py
timgrein: Yes but not on our custom `AsyncIterator` (which should probably be called `AsyncGenerator` class :) )

tests/commons.py
timgrein: I think we check it somewhere for _some_ files. Maybe the `tests` directory is ignored?

PR #1315, opened 2023-07-20
[Google Drive] Use async os stat operation to check file size

jedrazb:
## Changes

Use async os stat operation to check file size

While checking implementation of `aiohttp_session` I realized that we can use `aiofiles` to get file size in non-blocking way. Here is the link to open-source repo: https://github.com/omarryhan/aiogoogle/blob/master/aiogoogle/sessions/aiohttp_session.py#L17

Use the same async approach in google drive connector.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/google_drive.py
+++ b/connectors/sources/google_drive.py
@@ -9,7 +9,7 @@ import os
 from functools import cached_property, partial
 
 import aiofiles
-from aiofiles.os import remove
+from aiofiles.os import remove, stat
 from aiofiles.tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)


timgrein (approved):
Nice! Do we've other connectors, where we could benefit from getting the file size in an async manner?

jedrazb:
> Nice! Do we've other connectors, where we could benefit from getting the file size in an async manner?

Yup there are a couple of places where we use sync os operations: https://github.com/elastic/connectors-python/blob/main/connectors/utils.py#L205 - not relates to size tho

Issue #949, opened 2023-05-25
Dropbox Advanced Sync Rules


danajuratoni:
Similarly to the advanced filtering capabilities added to Jira and Confluence e.g. https://github.com/elastic/connectors-python/issues/874, support customized filtering via advanced sync rules for Dropbox.

### Acceptance criteria
- [ ] add support for advanced sync rules
- [ ] add documentation with several examples illustrating the syntax, how a basic use case works, limitations
- [ ] document impact of advanced sync rules on connector configuration and on document indexing
- [ ] add tooltip on the configurable fields that are ignored when advanced sync rules are used, informing the user about the behavior

praveen-elastic:
Updates [8th June to 15th June]:

Explored the GitHub APIs and documentation for search query parameters
Prepared and shared Sync Rules Object Model and Schema as a document: https://docs.google.com/document/d/1bNEvbxJyW_MAmZEcNe9Y8gBv6PbnzbVEhLU7HnEKflg/edit

praveen-elastic:
Updates [8th June to 15th June]:

Explored the Dropbox APIs and documentation for search query parameters
Prepared and shared Sync Rules Object Model and Schema as a document: https://docs.google.com/document/d/1bNEvbxJyW_MAmZEcNe9Y8gBv6PbnzbVEhLU7HnEKflg/edit

PR #1316, opened 2023-07-20
[Dropbox] Advanced Sync Rules

moxarth-elastic:
## Closes https://github.com/elastic/connectors-python/issues/949

This PR is having the implementation of Advanced Sync rules for Dropbox connector.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/dropbox.py
+++ b/connectors/sources/dropbox.py
@@ -15,10 +15,15 @@ from urllib import parse
 
 import aiofiles
 import aiohttp
+import fastjsonschema
 from aiofiles.os import remove
 from aiofiles.tempfile import NamedTemporaryFile
 from aiohttp.client_exceptions import ClientResponseError, ServerDisconnectedError
 
+from connectors.filtering.validation import (
+    AdvancedRulesValidator,
+    SyncRuleValidationResult,
+)
 from connectors.logger import logger
 from connectors.source import BaseDataSource
 from connectors.utils import (
@@ -78,6 +83,8 @@ class EndpointName(Enum):
     DOWNLOAD = "download"
     PAPER_FILE_DOWNLOAD = "paper_file_download"
     RECEIVED_FILE_DOWNLOAD = "received_files_download"
+    SEARCH = "search"
+    SEARCH_CONTINUE = "search_continue"
 
 
 ENDPOINTS = {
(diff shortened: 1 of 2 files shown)


connectors/sources/dropbox.py
```diff
@@ -684,15 +817,33 @@ async def get_docs(self, filtering=None):
...
-                self.get_content, attachment=attachment, is_shared=True
-            )
+        if filtering and filtering.has_advanced_rules():
+            advanced_rules = filtering.get_advanced_rules()
+            for rule in advanced_rules:
+                self._logger.debug(f"Fetching files using advanced sync rule: {rule}")
+
+                async for document, attachment in self.advanced_sync(rule=rule):
+                    if document["type"] == FILE:
+                        if document.get("url"):
```
seanstory, line 827: would `if document.get("url")` also be true for documents that come out of `self._fetch_shared_files()`? If so, I think this new block could be DRY'd up with the below section. Like pulling out a separate function like:
```
def document_tuple(self, document, attachment)
  if document["type"] == FILE:
    if document.get("url"):
      yield document, partial(self.get_content, attachment=attachment, is_shared=True)
    else:
       yield document, partial( self.get_content, attachment=attachment)
   else:
     yield document, None
     
     
...
      async for document, attachment in self.advanced_sync(rule=rule):
        yield document_tuple(document, attachment)
  else:
    async for document, attachment in self._fetch_files_folders(path=self.dropbox_client.path):
      yield document_tuple(document, attachment)
```

seanstory (approved):

connectors/sources/dropbox.py
```diff
@@ -316,6 +325,8 @@ async def _paginated_api_call(
                         url_name = ENDPOINTS.get(
                             EndpointName.RECEIVED_FILES_CONTINUE.value
                         )
+                    elif kwargs.get("search_file") is True:
+                        url_name = ENDPOINTS.get(EndpointName.SEARCH_CONTINUE.value)
```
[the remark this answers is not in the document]
moxarth-elastic: Sounds better!

connectors/sources/dropbox.py
moxarth-elastic: Yes, Thanks for the suggestion! I think making this code snippet DRY'd would be great 👍

PR #1323, opened 2023-07-21
Initialise empty RCF values in type-safe way

navarone-feekery:

These changes will allow configurable fields to be created without specifying a `value`.
If a field doesn't have a `value` it will initially be assigned a value of `None` by default. This is then converted to an empty value that is more "safe" for its type. Converting to type-specific empty values will make sure unexpected typing errors don't occur, particularly in tests or Kibana UI.
It is expected that the user would then update these values either via the Kibana UI or Dev Tools. They won't cause validation errors because validation checks are not performed until a connector state is configured.

| type | empty value |
| --- | --- |
| str / default | `""` |
| int | `None` |
| float | `None` |
| bool | `None` |
| list | `[]` |

`int` and `float` are not converted because it is impossible to have an empty number.
`bool` also is not converted because `None` doesn't always equal `True` or `False`. In practice, none of our boolean values will start as `None`, so this is largely arbitrary.
In these situations, the configuration validator will pick up any problems to alert the user.

#### Pre-Review Checklist

## Related Pull Requests

This PR is a prerequisite for https://github.com/elastic/connectors-python/pull/761


connectors/source.py
```diff
@@ -106,19 +106,27 @@ def value(self, value):
...
+                    return value
+                return bool(value) if value is not None else None
+            case "list":
+                if isinstance(value, list):
+                    return value
+                if isinstance(value, str):
+                    return [item.strip() for item in value.split(",")] if value else []
+                return [value] if value is not None else []
+            case _:  # str
+                return str(value) if value else ""
```
wangch079, line 129: This seems can be simplified to:

```python
from pydoc import locate

casted_type = locate(type_)

if isinstance(value, casted_type):
    return value

return casted_type(value) if value else None
```

But you need to do some special handling like `strip()`, etc..

wangch079 (commented):
>but wanted to just mark this change as non-backwards-compatible. It's probably fine to proceed with it, though.

I think it's still backward-compatible? The `_convert` function only works when loading configuration from ES to memory, and it will not try to make changes to the config in ES.

If the config in the ES works before this change, it should continue work as expected.

connectors/source.py
navarone-feekery: I will try to simplify it, some of these cases could indeed be consolidated into a default case.
But yeah we need to handle lists differently and there's some subtle differences in handling `None` between numbers,strings, etc.

navarone-feekery: @wangch079 I've simplified the logic using `locate` but it required me to add some `pyright:ignore` comments. Let me know if that's okay. It's caused by pyright being unsure of what class exactly `cast_type` becomes. I added a type check clause earlier in the function so it shouldn't be an issue.

Specifics:
- `if isinstance(value, cast_type)` -> `Object of type "object" is not callable`
- `cast_type(value)` -> `Argument of type "object" cannot be assigned to parameter "__class_or_tuple" of type "_ClassInfo" in function "isinstance"`

connectors/source.py
```diff
@@ -105,20 +106,27 @@ def value(self):
...
-        return value
+        if isinstance(value, cast_type):  # pyright:ignore
+            return value
+
+        # list requires special type casting
+        if field_type_ == "list":
+            if isinstance(value, str):
+                return [item.strip() for item in value.split(",")] if value else []
+            else:
+                return [value] if value is not None else []
```
wangch079, line 123: Just curious, do you need to add special handling if `value` is `set` or `dict`?

connectors/source.py
```diff
@@ -105,20 +106,27 @@ def value(self):
...
-        elif field_type == "bool":
-            return value.lower() in ("y", "yes", "true", "1")
-        elif field_type == "list":
-            return [item.strip() for item in value.split(",")]
-        return value
+        if isinstance(value, cast_type):  # pyright:ignore
+            return value
+
+        # list requires special type casting
+        if field_type_ == "list":
```
wangch079, line 119: nit, you can use
```suggestion
        if cast_type is list:
```

connectors/source.py
```diff
@@ -105,20 +106,27 @@ def value(self):
...
+            if isinstance(value, str):
+                return [item.strip() for item in value.split(",")] if value else []
+            else:
+                return [value] if value is not None else []
+
+        # strings don't get nullified
+        if value is None and field_type_ == "str":
+            return ""
+
+        return cast_type(value) if value else None  # pyright:ignore
```
wangch079, line 129: One suggestion, WDYT if we define a constant:
```python
TYPE_DEFAULTS = {
    str: "",
    int: None,
    float: None,
    bool: None,
    list: [],
}
```

So this can be simplified to
```suggestion
        if value is None:
            return TYPE_DEFAULTS[cast_type]

        return cast_type(value)  # pyright:ignore
```

At the beginning of the method, the unsupported type check can be simplified to:
```python
if cast_type not in TYPE_DEFAULTS:
            # unsupported type
            return value
```

One benefit is that if you want to add more type in the future, you only need to add to `TYPE_DEFAULTS`

connectors/source.py
navarone-feekery: This isn't checking if the value is a list, but if the value is supposed to be a list. Let's say the user has created a field like this:

```
{
  "label": "foo",
  "value": "bar",
  "type": "list"
}
```

In this case, the `cast_type` value will be `str` which doesn't align with the `field_type_` which would be `list`.
If the value and type align then it will already be returned on line 116.

navarone-feekery: AH I was getting confused, I thought for a moment that `cast_type` was built from `value` rather than `field_type_`. I just woke up sorry.

connectors/source.py
navarone-feekery: I added a check for if `value` is `set` or `dict` and the `field_type_` is list. They should be converted like this:

`{"foo": "bar"}` -> `[("foo", "bar")]`
`{"foo", "bar"}` -> `["foo", "bar"]`

connectors/source.py
navarone-feekery: Nice idea, I've added this 🚀

connectors/source.py
navarone-feekery: Fixed!

connectors/source.py
```diff
@@ -239,7 +260,7 @@ def __init__(self, config):
 
     def set_defaults(self, default_config):
         for name, item in default_config.items():
-            self._defaults[name] = item["value"]
+            self._defaults[name] = item.get("value", None)
```
artem-shelkovnikov, line 263: Is this line needed?

tests/sources/test_gmail.py
```diff
@@ -119,7 +119,9 @@ def setup_source():
 
 def set_dls_enabled(source, dls_enabled):
     source.set_features(Features({"document_level_security": {"enabled": dls_enabled}}))
-    source.configuration.set_field("use_document_level_security", value=dls_enabled)
+    source.configuration.set_field(
+        "use_document_level_security", field_type="bool", value=dls_enabled
```
artem-shelkovnikov, line 123: Intuitively, it would be nice not to pass `field_type` here because the configuration already knows, that `use_document_level_security` is boolean.

Why do we need to pass it here?

navarone-feekery: The `set_field` func has some default values that need to be specified. This isn't an issue in the app because we only run `set_field` when a connector's configuration is first populated. We also co-opt it for tests like this but it's not a simple _make value `x`_.

We could create an alternative func for testing that only updates the value, like `set_value`. WDYT?

connectors/source.py
navarone-feekery: I was worried that not specifying a `value` would cause an error here, but it doesn't. I removed it and added some tests to cover that.

tests/sources/test_gmail.py
navarone-feekery: Actually the more I look into it the more I feel that `set_field` shouldn't be used for managing configurations in testing, so I will create a new func.

connectors/source.py
artem-shelkovnikov: Since it's internal, I think it's okay - we can address it later

tests/sources/test_gmail.py
navarone-feekery: I've added a new func `set_value` and updated all affected tests.

tests/sources/test_azure_blob_storage.py
```diff
@@ -465,7 +465,7 @@ async def test_validate_config_no_account_name():
 
     # Setup
     async with create_source(AzureBlobStorageDataSource) as source:
-        source.configuration.set_field(name="account_name", value="")
+        source.configuration.set_value(name="account_name", value="")
```
artem-shelkovnikov, line 468: Sorry, I completely misunderstood the need for `set_value`.

You can init source like it's done here: https://github.com/elastic/connectors-python/blob/main/tests/sources/test_jira.py#L462

```python
async with create_source(AzureBlobStorageDataSource), account_name="") as source:
    # something
```

Then you can delete `set_value` function - I'd avoid having functions for testing-only in prod code.

navarone-feekery: @artem-shelkovnikov I looked into removing all iterations of `set_field` in favour of initialising the source with any specified values, but it is simply far too much work for this issue. I would prefer to create another tech debt issue and tackle it separately.

Instead I'd prefer to revert back to when we needed to specify `field_type` when calling `set_field` in tests for now.

navarone-feekery: (Is it even an issue? Now we have to specify `field_type` if it isn't string when calling `set_field` in tests, that's all)

navarone-feekery:
@artem-shelkovnikov after our discussion I've changed all existing `set_field` calls in tests that weren't directly related to creating a field. They will now use `get_value().value` to make use of the `Field` class's value setter.

artem-shelkovnikov (approved):
Thank you for doing this!

Issue #950, opened 2023-05-25
GitHub Advanced Sync Rules


danajuratoni:
Similarly to the advanced filtering capabilities added to Jira and Confluence e.g. https://github.com/elastic/connectors-python/issues/874, support customized filtering via advanced sync rules for GitHub.

### Acceptance criteria
- [x] add support for advanced sync rules
- [x] add documentation with several examples illustrating the syntax, how a basic use case works, limitations
- [x] document impact of advanced sync rules on connector configuration and on document indexing
- [x] add tooltip on the configurable fields that are ignored when advanced sync rules are used, informing the user about the behavior

abhishekjoshi-crest:
All tasks of advanced sync rules was completed. Hence, closing this issue

PR #1324, opened 2023-07-24
[GitHub] Advanced Sync rules

abhishekjoshi-crest:
## Closes https://github.com/elastic/connectors-py/issues/950

1. This PR introduces advanced filtering rules for GitHub by queries to filter out document based on that
2. Added documentation with several examples illustrating the syntax, how a basic use case works, limitations
3. Added documentation regarding impact of advanced sync rules on connector configuration and on document indexing
4. Included tooltip on the configurable fields that are ignored when advanced sync rules are used
5. As GitHub Documentation PR is not merged yet, we covered third point of the #950 in GitHub documentation PR.

#### Pre-Review Checklist


docs/sync-rules/GITHUB.md
```diff
@@ -0,0 +1,59 @@
+### Setting up the Github connector
+
+See the [Developer guide](../../docs/DEVELOPING.md) for setting up connectors.
+
+### Example advanced sync rules
+
+#### Schema for indexing document and files based on branch name configured via branch key
```
timgrein, line 7:
```suggestion
#### Advanced rules for indexing document and files based on branch name configured via branch key
```

docs/sync-rules/GITHUB.md
```diff
@@ -0,0 +1,59 @@
...
+  {
+    "repository": "repo_name",
+    "filter": {
+      "issue": "is:bug"
+    }
+  }
+]
+```
+
+#### Schema for indexing document based on pr query related to open pr's via pr key
```
timgrein, line 33:
```suggestion
#### Advanced rules for indexing document based on pr query related to open pr's via pr key 
```

connectors/sources/github.py
```diff
@@ -1185,28 +1439,86 @@ async def get_docs(self, filtering=None):
...
+                        async for pull_request in self._fetch_pull_requests(
+                            repo_name=repo_name,
+                            response_key=("search",),
+                            filter_query=pull_request_query,
+                            query=GithubQuery.SEARCH_QUERY.value,
+                        ):
+                            yield pull_request, None
+                    else:
+                        self._logger.warning(
+                            f"Skipping pr for rule: {pull_request_query}"
```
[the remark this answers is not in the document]
abhishekjoshi-crest: `query_status` only returns boolean value whenever the query is not related to their object then it will return False otherwise True. Example: When issue query is passed in pr key then `query_status` will be False. IMO `query_status` shouldn't be part of logger as it will also False in skipping logger.

connectors/sources/github.py
```diff
@@ -963,24 +1176,28 @@ async def _extract_pull_request(self, pull_request, owner, repo):
             pull_request.pop(field)
         yield pull_request
 
-    async def _fetch_pull_requests(self, repo_name):
+    async def _fetch_pull_requests(
+        self,
+        repo_name,
+        response_key=(REPOSITORY_OBJECT, "pullRequests"),
+        filter_query=None,
+        query=GithubQuery.PULL_REQUEST_QUERY.value,
```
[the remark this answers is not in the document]
abhishekjoshi-crest: Yup, will hide query via flag whether to fetch all object or fetch with custom query

connectors/sources/github.py
```diff
@@ -373,6 +378,139 @@ class GithubQuery(Enum):
...
+                endCursor
+            }
+            nodes {
+                author {
+                login
+                }
+                body
+            }
+            }
+            reviews(first: 45) {
```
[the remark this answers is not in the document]
abhishekjoshi-crest: In GitHub GraphQL maximum node limit is 500,000 and if we increase it this limit is exceeded.

connectors/sources/github.py
```diff
@@ -231,47 +237,47 @@ class GithubQuery(Enum):
...
+        }}
+    }}
+    }}
     """
     COMMENT_QUERY = """
     query ($owner: String!, $name: String!, $number: Int!, $cursor: String) {{
     repository(owner: $owner, name: $name) {{
         {object_type}(number: $number) {{
-        comments(first: 100, after: $cursor) {{
+        comments(first: {node_size}, after: $cursor) {{
```
seanstory, line 280: Hard to follow in this much diff, but should this be `NODE_SIZE` (capitalized)?

connectors/sources/github.py
```diff
@@ -287,63 +293,63 @@ class GithubQuery(Enum):
...
+        }}
+    }}
+    }}
     """
     LABELS_QUERY = """
     query ($owner: String!, $name: String!, $number: Int!, $cursor: String) {{
     repository(owner: $owner, name: $name) {{
         {object_type}(number: $number) {{
-        labels(first: 100, after: $cursor) {{
+        labels(first: {node_size}, after: $cursor) {{
```
seanstory, line 352: Another instance of missed caps?

connectors/sources/github.py
```diff
@@ -360,7 +366,7 @@ class GithubQuery(Enum):
     query ($owner: String!, $name: String!, $number: Int!, $cursor: String) {{
     repository(owner: $owner, name: $name) {{
         {object_type}(number: $number) {{
-        assignees(first: 100, after: $cursor) {{
+        assignees(first: {node_size}, after: $cursor) {{
```
seanstory, line 369: Another

connectors/sources/github.py
```diff
@@ -533,22 +674,24 @@ async def get_github_item(self, resource):
...
-        """Get page information from the response.
+    def get_data_by_keys(self, response, keys, endKey):
+        """Retrieve data from a nested dictionary using a list of keys and an end key.
 
         Args:
-            response (dict): Response of GitHub
-            Keys (list): List of fields to get pageInfo
+            response (dict): The nested dictionary from which data will be extracted.
+            keys (list): A list of strings representing the keys to navigate the nested dictionary.
+            endKey (str): The final key to retrieve the desired data from the nested dictionary.
```
seanstory, line 683: why have an `endKey` and not just have that be the last key in `keys`?

docs/sync-rules/GITHUB.md
seanstory, line 1: This is really nice, having this guide alongside the code changes for review. Thanks for including these together.

docs/sync-rules/GITHUB.md
```diff
@@ -0,0 +1,80 @@
...
+    "filter": {
+      "issue": "is:bug",
+      "pr": "is:open",
+      "branch": "sync-rules-feature"
+    }
+  }
+]
+```
+
+**NOTE**: All documents pulled by a given rule are indexed regardless of whether the document has already been indexed by the previous rule. In such cases, document duplication won't happen but the count of indexed documents will be different in the logs. The correct count of documents can be referred from the ElasticSearch index.
```
seanstory, line 61:
```suggestion
**NOTE**: All documents pulled by a given rule are indexed regardless of whether the document has already been indexed by the previous rule. In such cases, document duplication will happen, but the count of indexed documents will be different in the logs. The correct count of documents can be referred from the ElasticSearch index.
```

Because Elasticsearch will do the deduplication, right? So it does happen, just after sending duplicates to ES with the same IDs which overwrite one another.

tests/sources/test_github.py
```diff
@@ -895,3 +902,233 @@ async def test_get_docs():
...
+                }
+            ],
+            SyncRuleValidationResult(
+                SyncRuleValidationResult.ADVANCED_RULES,
+                is_valid=False,
+                validation_message=ANY,
+            ),
+        ),
+        (
+            # invalid: invalid key
```
seanstory, line 972:
```suggestion
            # invalid: invalid array value
```

tests/sources/test_github.py
```diff
@@ -895,3 +902,233 @@ async def test_get_docs():
...
+                }
+            ],
+            SyncRuleValidationResult(
+                SyncRuleValidationResult.ADVANCED_RULES,
+                is_valid=False,
+                validation_message=ANY,
+            ),
+        ),
+        (
+            # invalid: repository can be empty
```
seanstory, line 988:
```suggestion
            # invalid: repository can not be empty
```

tests/sources/test_github.py
```diff
@@ -895,3 +902,233 @@ async def test_get_docs():
...
+            ),
+            [
+                REPOS,
+                MOCK_RESPONSE_PULL,
+                MOCK_RESPONSE_ISSUE,
+                MOCK_RESPONSE_ATTACHMENTS[0],
+            ],
+        ),
+        (
+            # Configured invalid queries, without branch
```
seanstory, line 1079:
```suggestion
            # Configured invalid queries, with branch
```

tests/sources/test_github.py
seanstory, line 1: great tests - those comments are super helpful

seanstory (commented):
couple of nits, but largely looks good to me.

connectors/sources/github.py
abhishekjoshi-crest: we will be using `object_type` and `node_size` in format string later so, I kept this variable name in smaller case and in other queries we are using f-string and directly assigning the `NODE_SIZE`. Same for `LABELS_QUERY` and `ASSIGNEES_QUERY`.

connectors/sources/github.py
abhishekjoshi-crest: Because in `_fetch_pull_requests` and in `_fetch_issues` we are passing keys more than one time. one to get `pageInfo` and one to get `nodes`. So if we are remove the `endKey` then we need to update keys multiple times i.e. add `pageInfo` in keys then remove `pageInfo` and add `nodes` etc. So for simplicity I have added endKey argument.

connectors/sources/github.py
seanstory: Is it ever variable? Or are you always just passing that constant as the variable to the format?

seanstory (approved):

connectors/sources/github.py
abhishekjoshi-crest: Yes we are passing constant in format because above query is not f-string as we also needs pass `object_type`.

PR #1332, opened 2023-07-24
Fix warning  messages for respective connectors test cases

akanshi-elastic:
This PR includes:

- Fixing warning messages for test cases for connectors:
  - google_cloud_storage( remove `DEFAULT_PEM_FILE` and set a constant in the same file )
  - confluence
  - jira
  - github
  - sharepoint_server
  - dropbox
- make create_source async context manager

#### Pre-Review Checklist


connectors/sources/google_cloud_storage.py
```diff
@@ -193,14 +193,15 @@ def get_default_configuration(cls):
...
-            "private_key": open(
-                os.path.abspath(DEFAULT_PEM_FILE)
-            ).read(),  # TODO: change this and provide meaningful defaults
+            "private_key": None,
             "client_email": "123-abc@developer.gserviceaccount.com",
             "client_id": "123-abc.apps.googleusercontent.com",
             "auth_uri": "https://accounts.google.com/o/oauth2/auth",
             "token_uri": "http://localhost:4444/token",
         }
+        with open(os.path.abspath(DEFAULT_PEM_FILE)) as f:
```
wangch079, line 202: If it's a fixed file, why bother reading it from file system, instead of setting it a constant in this file?

tests/sources/test_confluence.py
```diff
@@ -403,6 +403,7 @@ async def test_remote_validation_when_space_keys_are_valid():
 
     with mock.patch("aiohttp.ClientSession.get", return_value=async_response):
         await source._remote_validation()
+    await source.close()
```
wangch079, line 406: Adding so many `await source.close()` is a signal to me, that we could make some changes to `create_source`, so that it can be used like:

```python
with create_source(ConfluenceDataSource) as source:
    assert source ...
```

tests/sources/support.py
```diff
@@ -23,3 +23,24 @@ async def assert_basics(klass, field, value):
...
+            if k in config:
+                config[k].update({"value": v})
+            else:
+                config[k] = DEFAULT_CONFIGURATION.copy() | {"value": v}
+
+        self.source = self.klass(configuration=DataSourceConfiguration(config))
+        return self.source
+
+    async def __aexit__(self, exc_type, exc_value, traceback):
+        await self.source.close()
```
artem-shelkovnikov, line 46: I would actually make `create_source` function do this:

```
@asynccontextmanager
def create_source():
    # create source
    yield source
    # close source
```

The code here is unnecessary complicated imo

tests/sources/test_servicenow.py
```diff
@@ -85,7 +86,7 @@ async def test_validate_config_missing_fields_then_raise(field):
 
 @pytest.mark.asyncio
 async def test_validate_configuration_with_invalid_service_then_raise():
-    source = setup_servicenow()
+    source = await setup_servicenow()
```
wangch079, line 89: The `source` be closed already when it's returned.

tests/sources/support.py
```diff
@@ -3,23 +3,28 @@
...
     for k, v in extras.items():
         if k in config:
             config[k].update({"value": v})
         else:
             config[k] = DEFAULT_CONFIGURATION.copy() | {"value": v}
 
-    return klass(configuration=DataSourceConfiguration(config))
+    source = klass(configuration=DataSourceConfiguration(config))
+    yield source
+    await source.close()
```
wangch079, line 22: I think you may want to put it inside `try...finally` block: [ref](https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager)

```suggestion
    source = klass(configuration=DataSourceConfiguration(config))
    try:
        yield source
    finally:
        await source.close()
```

tests/sources/test_servicenow.py
akanshi-elastic: In that case I guess we need to use create_source to every test case to close the sessions properly.

wangch079: You can change `setup_servicenow()` to only set the fields:
```python
def setup_servicenow(source):
    # Set up default config with default values
    source.configuration.set_field(name="username", value="admin")
    source.configuration.set_field(name="password", value="changeme")
```

And then:

```python
async with create_source(ServiceNowDataSource) as source:
    setup_servicenow(source)
    ...
```

artem-shelkovnikov (dismissed):

artem-shelkovnikov (approved):

PR #1339, opened 2023-07-25
removing memray usage

seanstory:
This is a sad attempt to fix our nightly builds

see this thread: https://elastic.slack.com/archives/C04KMMNAHGB/p1689797820693809

### DIFF
--- a/tests/ftest.sh
+++ b/tests/ftest.sh
@@ -58,9 +58,9 @@ then
     $PYTHON fixture.py --name $NAME --action description > description.txt
     if [[ $PLATFORM == "darwin" ]]
     then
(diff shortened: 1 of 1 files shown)


jedrazb (approved):
Should we add a tech-debt ticket to re-add memray once the new version is released? Thank you for keeping our nightlies healthy! 🥇 

(I looked at this PR before it was marked as a draft)

seanstory:
> Should we add a tech-debt ticket to re-add memray once the new version is released? 

Yes. However, this doesn't look to be sufficient. Turns out the aarch64 nightly dies during the install phase of our test deps, so this changes nothing. Now I need to look at actually removing/disabling memray in perf8. :/

seanstory:
https://github.com/elastic/connectors-python/pull/1373 is better.

PR #1341, opened 2023-07-26
[Network Drive] Advanced Sync Rules

praveen-elastic:
## Part of https://github.com/elastic/connectors-py/issues/1290

- This PR introduces advanced filtering rules for Network Drive by glob patterns to filter out document based on that
- Added documentation with several examples illustrating the syntax, how a basic use case works, limitations
- Added documentation regarding impact of advanced sync rules on connector configuration and on document indexing

#### Pre-Review Checklist


connectors/sources/network_drive.py
```diff
@@ -91,6 +166,36 @@ def create_connection(self):
...
+
+        Returns:
+            matched_paths (set): Set of paths that match the advanced rules.
+            invalid_rules (list): List of advanced rules that have no matching paths.
+        """
+        invalid_rules = []
+        matched_paths = set()
+        for rule in self.advanced_rules:
+            rule_valid = False
+            glob_pattern = rule["pattern"].replace("\\", "/")
```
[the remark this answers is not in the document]
praveen-elastic: It's Windows 😅 . Paths in windows are traditionally having backslash between directories. `C:\Program Files\Applications`
Windows does not allow blackslash in their folder/file naming: [reference](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#:~:text=The%20backslash%20divides%20the%20file,separates%20the%20names%20into%20components.)
Similarly, Unix/Linux does not allow forward-slash in their file/folder name

wangch079 (commented):
I don't feel it's necessary to have the instance variable `advanced_rules`. Advanced rules are used in two places:
1. data synchronization (i.e. `get_docs`)
2. filtering validation (i.e. `validate`)

And the instance variable has to be assigned when you call the above two methods. I'd suggest making `find_matching_paths` accept a new parameter `advanced_rules`. I don't think we need to cache `find_matching_paths`, instead we could cache `find_matching_paths`, which is the expensive operation.

wangch079 (dismissed):
LGTM, with one nit.

connectors/sources/network_drive.py
```diff
@@ -200,9 +306,14 @@ async def get_docs(self, filtering=None):
         Yields:
             dictionary: Dictionary containing the Network Drive files and folders as documents
         """
-        directory_details = smbclient.walk(top=rf"\\{self.server_ip}/{self.drive_path}")
 
-        for path, _, _ in directory_details:
+        if filtering and filtering.has_advanced_rules():
+            advanced_rules = filtering.get_advanced_rules()
+            matched_paths, _ = self.find_matching_paths(advanced_rules)
```
wangch079, line 312: Although I don't expect any invalid rules here, because rules should be first validated before calling `get_docs`, I'd suggest adding a warning message here in case something weird happens.

praveen-elastic: Upated in commit: [add a warning](https://github.com/elastic/connectors-python/pull/1341/commits/5dca844a1ad3bffbef0fe9fc07da2c76d15560ea)

connectors/sources/network_drive.py
```diff
@@ -200,9 +306,20 @@ async def get_docs(self, filtering=None):
...
-        for path, _, _ in directory_details:
+        if filtering and filtering.has_advanced_rules():
+            advanced_rules = filtering.get_advanced_rules()
+            try:
+                matched_paths, _ = self.find_matching_paths(advanced_rules)
+            except Exception as exception:
+                self._logger.warning(
+                    f"Something went wrong while running advanced rules. Exception: {exception}"
+                )
+                raise
```
wangch079, line 318: I don't think it will raise exception
```suggestion
            matched_paths, invalid_rules = self.find_matching_paths(advanced_rules)
            if invalid_rules:
                self._logger.error(
                    f"Invalid advanced rules are found: {invalid_rules}"
                )
                raise InvalidRules(...)
```

wangch079 (approved):

PR #1349, opened 2023-07-27
[Salesforce] Add salesforce connector skeleton

navarone-feekery:
## Closes https://github.com/elastic/enterprise-search-team/issues/5438

Adds the skeleton for Salesforce connector.
Includes the following basic functionality:

- configuration with validation
- ping
- get_token

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -1,0 +1,123 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+"""Salesforce source module responsible to fetch documents from Salesforce."""
+import aiohttp
+
+from connectors.source import BaseDataSource
+from connectors.utils import retryable
+
+RETRIES = 3
+RETRY_INTERVAL = 1
+
+BASE_URL = "https://{domain}.my.salesforce.com"
+TOKEN_ENDPOINT = "/services/oauth2/token"
+
+
+class SalesforceClient:
+    def __init__(self, configuration):
+        self.session = None
+        self.token = None
+        self.token_issued_at = None
+
+        self.base_url = BASE_URL.replace("{domain}", configuration["domain"])
+        self.client_id = configuration["client_id"]
+        self.client_secret = configuration["client_secret"]
+        self.token_payload = {
+            "grant_type": "client_credentials",
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+        }
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    def _get_session(self):
+        if self.session is not None:
+            return self.session
+
+        self.session = aiohttp.ClientSession(
+            timeout=aiohttp.ClientTimeout(total=None),
+            raise_for_status=True,
+        )
+        return self.session
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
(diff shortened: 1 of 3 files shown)


connectors/sources/salesforce.py
```diff
@@ -0,0 +1,123 @@
...
+                "value": "",
+            },
+            "domain": {
+                "type": "str",
+                "value": "",
+            },
+        }
+
+    async def validate_config(self):
+        self.configuration.check_valid()
```
wangch079, line 104: Need to call `super` as well.

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,123 @@
...
+            },
+            "client_secret": {
+                "type": "str",
+                "value": "",
+            },
+            "domain": {
+                "type": "str",
+                "value": "",
+            },
+        }
```
wangch079, line 101: Assume this is not the complete configuration and will be updated in later release.

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def _get_session(self):
+        if self.session is not None:
+            return self.session
+
+        self.session = aiohttp.ClientSession(
+            timeout=aiohttp.ClientTimeout(total=None),
+            raise_for_status=True,
+        )
+        return self.session
```
wangch079, line 45:
```suggestion
        if self.session is None:
            self.session = aiohttp.ClientSession(
                timeout=aiohttp.ClientTimeout(total=None),
                raise_for_status=True,
            )
        return self.session
```

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,123 @@
...
+        self.token_payload = {
+            "grant_type": "client_credentials",
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+        }
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    def _get_session(self):
```
wangch079, line 37: WDYT if we rename it to `session`,  make it a `cached_property`, and remove the `session` instance variable?

connectors/sources/salesforce.py
navarone-feekery: No it's not, this is just the basics for auth.
When I merge into `main` this will have more fields and all the missing field properties.

connectors/sources/salesforce.py
navarone-feekery: Are you sure? Every other implementation of `validate_config` doesn't call `super`.

connectors/sources/salesforce.py
navarone-feekery: I like this idea, so I implemented it. Thank you!

connectors/sources/salesforce.py
wangch079: Now I remembered, I had the discussion in another PR. We called `self.configuration.check_valid()` in `validate_config` of all the source implementations, which is also implemented in `validate_config` of `BaseDataSource`:

https://github.com/elastic/connectors-python/blob/e9b4e577b4311d8fd8ca01cddae14ec5528ae9c7/connectors/source.py#L473-L479

Instead, we should call `super` here instead of `self.configuration.check_valid()`. But this is not in the scope of this PR.

wangch079 (approved):
LGTM.

Just confirm, are `get_token` and `_auth_headers` for later use? Does `ping` not require authentication?

connectors/sources/salesforce.py
navarone-feekery: Ah I see, thank you! I will look into fixing that in the next PR.

Issue #1350, opened 2023-07-27
[GitHub] Index only markdown files


abhishekjoshi-crest:

abhishekjoshi-crest:
#1351 is merged. Hence closing this issue.

PR #1351, opened 2023-07-27
[GitHub] Index markdown files only

abhishekjoshi-crest:
## Closes https://github.com/elastic/connectors-python/issues/1350

As discussed on slack, In GitHub we are indexing markdown files only.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -24,7 +24,6 @@ from connectors.filtering.validation import (
 from connectors.logger import logger
 from connectors.source import BaseDataSource, ConfigurableFieldValueError
 from connectors.utils import (
-    TIKA_SUPPORTED_FILETYPES,
     CancellableSleeps,
     RetryStrategy,
     convert_to_b64,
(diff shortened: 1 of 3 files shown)


connectors/sources/github.py
```diff
@@ -1287,39 +1288,32 @@ async def _fetch_files(self, repo_name, default_branch):
             )
 
             for repo_object in file_tree.get("tree", []):
-                last_commit_timestamp = await self._fetch_last_commit_timestamp(
```
seanstory, line 1290: Should we be fetching the last_commit_timestamp for the markdown files? I see that we weren't before when type was `BLOB`, but I don't understand why.

seanstory (commented):
two clarity-seeking comments. But largely looks good.

connectors/sources/github.py
```diff
@@ -1287,39 +1288,32 @@ async def _fetch_files(self, repo_name, default_branch):
...
-                    repo_name=repo_name, path=repo_object["path"]
-                )
                 if repo_object["type"] == BLOB:
```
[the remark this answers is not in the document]
abhishekjoshi-crest: In the response, we are getting two types of documents one is `blob` and the other is `tree`. For the files, we are getting `blob` type and for the folders we are getting `tree` type. So, markdown files fall under `blob` type.

connectors/sources/github.py
abhishekjoshi-crest: Earlier, we were fetching `last_commit_timestamp` for all the documents as we were indexing all the files. but now we are indexing markdown files only, so we are fetching `last_commit_timestamp` for markdown files only in L1299.

seanstory: omg Idk how I missed that. Thanks.

PR #1355, opened 2023-07-28
Add PostgreSQLClient abstraction

wangch079:

This PR adds PostgreSQLClient abstraction, which will handle the connection to PostgreSQL server.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/postgresql.py
+++ b/connectors/sources/postgresql.py
@@ -5,6 +5,7 @@
 #
 """Postgresql source module is responsible to fetch documents from PostgreSQL."""
 import ssl
+from functools import cached_property
 from urllib.parse import quote
 
 from asyncpg.exceptions._base import InternalClientError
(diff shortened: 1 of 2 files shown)


connectors/sources/postgresql.py
```diff
@@ -170,127 +399,16 @@ def get_default_configuration(cls):
...
-            await anext(
-                self.execute_query(
-                    query=self.queries.ping(),
-                )
-            )
+            await self.postgresql_client.ping()
             self._logger.info("Successfully connected to Postgresql.")
         except Exception as e:
-            raise Exception(f"Can't connect to Postgresql on {self.host}") from e
+            raise Exception("Can't connect to Postgresql.") from e
```
navarone-feekery, line 411: Nit: it would still be useful to keep the host in the exception for easier debugging

navarone-feekery (approved):
Nice one! Just an optional nit

connectors/sources/postgresql.py
```diff
@@ -62,6 +63,231 @@ def all_schemas(self):
...
+    def close(self):
+        self._sleeps.cancel()
+
+    @cached_property
+    def engine(self):
+        connection_string = f"postgresql+asyncpg://{self.user}:{quote(self.password)}@{self.host}:{self.port}/{self.database}"
+        return create_async_engine(
+            connection_string,
+            connect_args=self._get_connect_args(),
+        )
```
artem-shelkovnikov, line 109: What's an async engine? Does it need to be closed afterwards?

connectors/sources/postgresql.py
```diff
@@ -62,6 +63,231 @@ def all_schemas(self):
...
+                table=table,
+            ),
+            fetch_many=True,
+            schema=schema,
+            table=table,
+        ):
+            yield data
+
+    async def _execute_query(self, query, fetch_many=False, **kwargs):
+        """Executes a query and yield rows
```
artem-shelkovnikov, line 214: I think it's worth to refactor this `fetch_many` flag, `yield_once` logic is also ultra confusing

wangch079: Agree. I can make it two separate functions:
1. `_execute_query`, to simply run the query and yield all.
2. `_execute_query_by_cursor`, which will yield rows one by one. It will also yield column names first if the parameter `include_column_names` is `True` (what `yield_once` does).

I think they were defined in one function, to re-use the retry logic. I actually tried to reuse the `@retryable` annotation, but it doesn't support instance variables:
```python
@retryable(
    retries=self.retry_count, # this is not supported
    interval=0,
    strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
)
```

But I'd like to refactor this in a follow-up PR, to unblock Crest team working on advanced rules.

artem-shelkovnikov (approved):

PR #1366, opened 2023-07-31
[SPO] Granular permissions for site list items + propagate site admin permissions for site-list and list-item

jedrazb:
## Closes https://github.com/elastic/enterprise-search-team/issues/5480

Add support for list-item granular permissions, this feature can have impact on sync time therefore it's controlled by RCF `fetch_unique_list_item_permissions`. 

Also, propagate the collection site admins to access control lists of site lists and site list items. 

After this change is merged we should have following:

| Entity     | Granular permissions | Collection site admins in ACL |
|------------|----------------------|-------------------------------|
| drive-item | ✅                    | ✅                             |
| site-page  | ✅                    | ✅                             |
| site-list  | ✅                    | ✅ (this PR)                   |
| list-item  | ✅ (this PR)          | ✅ (this PR)                   |

#### Pre-Review Checklist


connectors/sources/sharepoint_online.py
```diff
@@ -1142,6 +1159,15 @@ def get_default_configuration(cls):
                 "type": "bool",
                 "value": True,
             },
+            "fetch_unique_list_item_permissions": {
+                "depends_on": [{"field": "use_document_level_security", "value": True}],
+                "display": "toggle",
+                "label": "Fetch unique list item permissions",
+                "order": 11,
+                "tooltip": "Enable this option to fetch unique list item permissions. This setting can increase sync time. If this setting is disabled a list will inherit permissions from its parent site.",
```
seanstory, line 1167:
```suggestion
                "tooltip": "Enable this option to fetch unique list item permissions. This setting can increase sync time. If this setting is disabled a list item will inherit permissions from its parent site.",
```

connectors/sources/sharepoint_online.py
```diff
@@ -1772,6 +1812,40 @@ async def site_list_items(
...
+                            await self._get_access_control_from_role_assignment(
+                                role_assignment
+                            )
+                        )
+
+                    list_item = self._decorate_with_access_control(
+                        list_item, list_item_access_control
+                    )
+
+            if not has_unique_role_assignments:
```
seanstory, line 1844: This could just be an `else:` and you could remove the first definition of `has_unique_role_assignments = False` above.

connectors/sources/sharepoint_online.py
```diff
@@ -1784,6 +1858,9 @@ async def site_list_items(
                     list_item_attachment[
                         "_original_filename"
                     ] = list_item_attachment.get("FileName", "")
+                    list_item_attachment[ACCESS_CONTROL] = list_item.get(
```
seanstory, line 1861: oh nice catch.

seanstory (dismissed):

connectors/sources/sharepoint_online.py
jedrazb: Yup, I shamelessly copy-pasted this from other place in the code-base. Will fix both.

connectors/sources/sharepoint_online.py
jedrazb: Nice catch!

connectors/sources/sharepoint_online.py
```diff
@@ -699,27 +699,44 @@ async def site_list_role_assignments(self, site_web_url, site_list_name):
...
+        self, site_web_url, site_list_name, list_item_id
+    ):
+        self._validate_sharepoint_rest_url(site_web_url)
 
-        async for page in self._graph_api_client.scroll(
-            f"{GRAPH_API_URL}/sites/{site_id}/lists/{list_id}/items?$select={select}&$expand={expand}"
-        ):
-            for site_list in page:
-                yield site_list
+        url = f"{site_web_url}/_api/lists/GetByTitle('{site_list_name}')/items({list_item_id})/HasUniqueRoleAssignments"
```
artem-shelkovnikov, line 707: Curious - is this call necessary?

I see that you fetch `roleassignments` below, what if `HasUniqueRoleAssignments` is False, will `roleassignments` still return something that will break the logic?

artem-shelkovnikov (commented):
LGTM, left a question.

Have you done performance benchmarks? How slower does it get?

connectors/sources/sharepoint_online.py
jedrazb: This logic is based on the Tim's work - I mirrored existing logic for site pages and site lists. 

If `HasUniqueRoleAssignments` returns `False` means that there are no custom roles defined and we resort to using parent permissions for access control for a given doc as in #1264

jedrazb:
@artem-shelkovnikov 

> Have you done performance benchmarks? How slower does it get?

It gets slower (don't have solid numbers - I just tested with the test instance that mimics data of a particular customer).  The slowdown depends on the % of lists and list items in the dataset. 

Idea: Maybe we should have this option disabled by default?

artem-shelkovnikov (approved):

PR #1373, opened 2023-07-31
Don't run memray for ftest, run only for ftrace

artem-shelkovnikov:
##  Minor CI improvement

Memray is a plugin that's used for tracing the memory usage of the connectors. It uses lots of memory, disk, and provides a report that's not possible to see in our nightlies - it's just too big.

This PR makes it so, that memray does not run when running `make ftest`. To make it run, it's possible to use a separate command - `make ftrace`. `make ftrace` will enable memray plugin for ftest run.

#### Pre-Review Checklist

### DIFF
--- a/tests/ftest.sh
+++ b/tests/ftest.sh
@@ -19,10 +19,18 @@ PLATFORM='unknown'
 MAX_RSS="200M"
 MAX_DURATION=600
 
+export PERF8_TRACE=${PERF8_TRACE:-False}
 export REFRESH_RATE="${REFRESH_RATE:-5}"
 export DATA_SIZE="${DATA_SIZE:-medium}"
 export RUNNING_FTEST=True
 export VERSION='8.10.0-SNAPSHOT'
+
+if [ "$PERF8_TRACE" = true ]; then
+    echo 'Tracing is enabled, memray stats will be delivered'
+    PLUGINS='--asyncstats --memray --psutil'
+else
+    PLUGINS='--asyncstats --psutil'
+fi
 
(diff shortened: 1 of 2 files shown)


seanstory (approved):

tests/ftest.sh
```diff
@@ -19,11 +19,19 @@ PLATFORM='unknown'
 MAX_RSS="200M"
 MAX_DURATION=600
 
+export PERF8_TRACE=${PERF8_TRACE:-False}
 export REFRESH_RATE="${REFRESH_RATE:-5}"
 export DATA_SIZE="${DATA_SIZE:-medium}"
 export RUNNING_FTEST=True
 export VERSION='8.10.0-SNAPSHOT'
 
+if [ "$PERF8_TRACE" = true ]; then
```
wangch079, line 28: TIL, `=` is also used for comparison. From this [answer](https://stackoverflow.com/a/20449556), `=` is prefered, WDYT if we change all `==` to `=`?

artem-shelkovnikov: Good call, I've changed it to `==` for consistency within the file, but we might wanna move to `=` everywhere else.

wangch079 (approved):
>I've changed it to `==` for consistency within the file, but we might wanna move to `=` everywhere else.

Let's make it consistent in this PR. It can be a separate PR to move to `=`.

artem-shelkovnikov:
IMO it's okay if we don't backport it to 8.8, it should not break the builds - they can be slower, but not a big deal.

PR #1381, opened 2023-08-01
Update default configs for dropbox & servicenow

sphilipse:
This updates the ServiceNow and Dropbox configs to be in line with our implicit guidelines (no default values that don't provide meaningful information, no repeating the data source name in field names).


artem-shelkovnikov (dismissed):
Do we want to backport these changes too?

wangch079 (approved):

artem-shelkovnikov (approved):

PR #1389, opened 2023-08-02
[Salesforce] Add Salesforce connector

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/5183

Adds Salesforce as a new connector.
This connector is in beta for 8.10.

- Ingests the SObjects Accounts, Contacts, Leads, Campaigns, Opportunities, Cases
- Ingest ContentDocuments and extracts binary data from their associated files

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -5,7 +5,6 @@
 #
 """Salesforce source module responsible to fetch documents from Salesforce."""
 import asyncio
-from contextlib import contextmanager
 from functools import cached_property
 from itertools import groupby
 
@@ -269,8 +268,8 @@ class SalesforceClient:
 
         query = await self._cases_query()
         async for records in self._yield_non_bulk_query_pages(query):
+            case_feeds_by_case_id = {}
             if await self._is_queryable("CaseFeed") and records:
-                case_feeds_by_case_id = {}
                 case_ids = [x.get("Id") for x in records]
                 case_feeds = await self.get_case_feeds(case_ids)
 
@@ -907,6 +906,7 @@ class SalesforceAPIToken:
         if self._token:
             return self._token
 
+        response_body = {}
         try:
             response = await self.session.post(self.url, data=self.token_payload)
             response_body = await response.json()
@@ -916,9 +916,7 @@ class SalesforceAPIToken:
         except ClientResponseError as e:
             if 400 <= e.status < 500:
                 # 400s have detailed error messages in body
-                error_message = response_body.get(
-                    "error", "No error dscription found."
-                )
+                error_message = response_body.get("error", "No error dscription found.")
                 if error_message == "invalid_client":
                     raise InvalidCredentialsException(
                         f"The `client_id` and `client_secret` provided could not be used to generate a token. Status: {e.status}, message: {e.message}, details: {error_message}"
@@ -1336,7 +1334,9 @@ class SalesforceDataSource(BaseDataSource):
     def __init__(self, configuration):
         super().__init__(configuration=configuration)
         self.base_url = BASE_URL.replace("<domain>", configuration["domain"])
-        self.salesforce_client = SalesforceClient(configuration=configuration, base_url=self.base_url)
+        self.salesforce_client = SalesforceClient(
+            configuration=configuration, base_url=self.base_url
+        )
         self.doc_mapper = SalesforceDocMapper(self.base_url)
 
     def _set_internal_logger(self):
@@ -1414,7 +1414,9 @@ class SalesforceDataSource(BaseDataSource):
 
         # Note: this could possibly be done on the fly if memory becomes an issue
         all_content_docs = self._combine_duplicate_content_docs(all_content_docs)
-        async for content_doc in self.salesforce_client.download_content_documents(all_content_docs):
+        async for content_doc in self.salesforce_client.download_content_documents(
+            all_content_docs
+        ):
             yield self.doc_mapper.map_content_document(content_doc), None
 
     def _combine_duplicate_content_docs(self, content_docs):
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -839,7 +839,6 @@ class SalesforceClient:
             f"ContentDocument.LatestPublishedVersion.{x}"
             for x in queryable_version_fields
         ]
-        [f"'{x[1:]}'" for x in TIKA_SUPPORTED_FILETYPES]
         where_in_clause = ",".join([f"'{x[1:]}'" for x in TIKA_SUPPORTED_FILETYPES])
 
         self._content_document_links_join = (
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -1,0 +1,123 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+"""Salesforce source module responsible to fetch documents from Salesforce."""
+import aiohttp
+
+from connectors.source import BaseDataSource
+from connectors.utils import retryable
+
+RETRIES = 3
+RETRY_INTERVAL = 1
+
+BASE_URL = "https://{domain}.my.salesforce.com"
+TOKEN_ENDPOINT = "/services/oauth2/token"
+
+
+class SalesforceClient:
+    def __init__(self, configuration):
+        self.session = None
+        self.token = None
+        self.token_issued_at = None
+
+        self.base_url = BASE_URL.replace("{domain}", configuration["domain"])
+        self.client_id = configuration["client_id"]
+        self.client_secret = configuration["client_secret"]
+        self.token_payload = {
+            "grant_type": "client_credentials",
+            "client_id": self.client_id,
+            "client_secret": self.client_secret,
+        }
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    def _get_session(self):
+        if self.session is not None:
+            return self.session
+
+        self.session = aiohttp.ClientSession(
+            timeout=aiohttp.ClientTimeout(total=None),
+            raise_for_status=True,
+        )
+        return self.session
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+    )
+    async def get_token(self):
+        resp = await self._api_request(
+            f"{self.base_url}{TOKEN_ENDPOINT}", "post", data=self.token_payload
+        )
+        resp_json = await resp.json()
+        self.token = resp_json["access_token"]
+        self.token_issued_at = resp_json["issued_at"]
+
+    async def ping(self):
+        await self._api_request(self.base_url, "head")
+
+    async def close_session(self):
+        if self.session is not None:
+            await self.session.close()
+
+    def _auth_headers(self):
+        return {"authorization": f"Bearer {self.token}"}
+
+    async def _api_request(self, url, method, data=None):
+        return await getattr(self._get_session(), method)(url=url, data=data)
+
+
+class SalesforceDataSource(BaseDataSource):
+    """Salesforce"""
+
+    name = "Salesforce"
+    service_type = "salesforce"
+
+    def __init__(self, configuration):
+        super().__init__(configuration=configuration)
+        self.salesforce_client = SalesforceClient(configuration=configuration)
+
+    def _set_internal_logger(self):
+        self.salesforce_client.set_logger(self._logger)
+
+    @classmethod
+    def get_default_configuration(cls):
+        return {
+            "client_id": {
+                "type": "str",
+                "value": "",
+            },
+            "client_secret": {
+                "type": "str",
+                "value": "",
+            },
+            "domain": {
+                "type": "str",
+                "value": "",
+            },
+        }
+
+    async def validate_config(self):
+        self.configuration.check_valid()
+
+    async def close(self):
+        await self.salesforce_client.close_session()
+
+    async def ping(self):
+        try:
+            await self.salesforce_client.ping()
+            self._logger.debug("Successfully connected to Salesforce.")
+        except Exception as e:
+            self._logger.exception(f"Error while connecting to Salesforce: {e}")
+            raise
+
+    async def get_content(self, attachment, timestamp=None, doit=False):
+        # TODO implement
+        return
+
+    async def get_docs(self, filtering=None):
+        # TODO implement
+        yield {"foo": "bar"}, None
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -17,7 +17,6 @@ from connectors.utils import CancellableSleeps, retryable
 
 RETRIES = 3
 RETRY_INTERVAL = 1
-QUERY_PAGINATION_LIMIT = 100  # TODO: maybe make the limit configurable
 
 BASE_URL = "https://<domain>.my.salesforce.com"
 API_VERSION = "v58.0"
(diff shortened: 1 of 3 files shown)


connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+
+            download_url = content_version.get("VersionDataUrl")
+            if content_version.get("VersionDataUrl"):
+                doc["_attachment"] = await self._download(download_url)
+
+            yield doc
+
+    async def queryable_sobjects(self):
+        """Cached async property"""
+        if self._queryable_sobjects is not None:
```
wangch079, line 364: This can be simplified by `@cached_property`

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+        url = f"{self.base_url}{QUERY_ENDPOINT}"
+        params = {"q": soql_query}
+
+        while True:
+            response = await self._get_json(
+                url,
+                params=params,
+            )
+            yield response.get("records")
+            if response.get("done", True) is True:
```
wangch079, line 502: Why defaults to `True`? Will there be scenario where `done` is not returned?

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+                break
+
+            url = f"{self.base_url}{response.get('nextRecordsUrl')}"
+            params = None
+
+    async def _execute_non_paginated_query(self, soql_query):
```
wangch079, line 508: I see this is used by `get_case_feeds` only. Why `case_feeds` doesn't need pagination?

wangch079 (commented):
I see there are classes `SalesforceClient`, `SalesforceAPIToken`, `SalesforceSoqlBuilder`, `SalesforceDocMapper` and `SalesforceDataSource`. I think `SalesforceClient` takes too much responsibility. We should have the duty of each class clearly separated.

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+            timeout=aiohttp.ClientTimeout(total=None),
+        )
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        skipped_exceptions=[InvalidCredentialsException],
+    )
+    async def get_token(self):
+        while True:
```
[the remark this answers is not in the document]
navarone-feekery: Ah missed removing it when I added the retryable, nice catch!

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+
+    def __init__(self, configuration):
+        super().__init__(configuration=configuration)
+        self.salesforce_client = SalesforceClient(configuration=configuration)
+
+    def _set_internal_logger(self):
+        self.salesforce_client.set_logger(self._logger)
+
+    @classmethod
+    def get_default_configuration(cls):
```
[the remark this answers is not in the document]
navarone-feekery: required is `True` by default if you don't specify a key

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+            "grant_type": "client_credentials",
+            "client_id": client_id,
+            "client_secret": client_secret,
+        }
+
+    def token(self):
+        return self._token
+
+    async def generate(self):
+        with self._non_blocking_lock():
```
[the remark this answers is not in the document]
navarone-feekery: I may be wrong but I thought our requests were threaded, so there's a chance that multiple requests are attempting to get a token at the same time. Is that not the case?
Looking at it now, the Salesforce connector is not threading anything. I don't know how it works at the client level, though.

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+        self.session = session
+        self.url = f"{base_url}{TOKEN_ENDPOINT}"
+        self.token_payload = {
+            "grant_type": "client_credentials",
+            "client_id": client_id,
+            "client_secret": client_secret,
+        }
+
+    def token(self):
+        return self._token
```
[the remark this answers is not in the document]
navarone-feekery: The token does expire eventually. The user can define the duration for this. Anecdotally my tokens for testing last a few hours but they aren't permanent. I don't know how long a sync of a very large dataset could take, so I'd prefer to be able to "refresh" (i.e. re-get) the token.

connectors/sources/salesforce.py
navarone-feekery: It's asynchronous though. My understanding is that cached_property doesn't support async calls.
https://asyncstdlib.readthedocs.io/en/stable/source/api/functools.html#async-caches

connectors/sources/salesforce.py
navarone-feekery: I'm just being defensive in my coding. I'm pretty sure there will always be a `done` field.

connectors/sources/salesforce.py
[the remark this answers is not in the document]
wangch079: TIL. That means the token is fetched without expiry time, and not renewable. So I guess the only option left is trial and error

connectors/sources/salesforce.py
wangch079: TIL. Didn't realize that.

connectors/sources/salesforce.py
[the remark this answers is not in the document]
wangch079: But still think it might be worth combining methods `token` and `generate`.

connectors/sources/salesforce.py
navarone-feekery: I was copying from the WS connector and thought (wrongly) that case_feeds were not being paginated (partly because we have a limit on the query of 500). I dug deeper and it _is_ paginated, so I'll fix that.

connectors/sources/salesforce.py
wangch079: My point is, should it be defaulted to `False`?

connectors/sources/salesforce.py
navarone-feekery: Sorry I deleted my comment because I think I'm wrong? cached_property doesn't work from functools but the very library I linked has an async cached_property
(linking again) https://asyncstdlib.readthedocs.io/en/stable/source/api/functools.html#async-caches
We don't use this library though so I don't think it's worth introducing. So it IS possible to cache, just not with functools.

connectors/sources/salesforce.py
navarone-feekery: If it's defaulted to `False` it will continue trying to get the next page, which might not exist and will raise an error. But maybe that is better, if their API got weird we probably don't want to silently ingest half their dataset.

Alternative: keep it as `True` but log a warning if the `done` field is missing. WDYT?

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        skipped_exceptions=[RateLimitedException, InvalidQueryException],
+    )
+    async def _get_json(self, url, params=None):
+        response_body = None
+
+        while True:
```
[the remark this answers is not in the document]
navarone-feekery: Same reason as above, will remove 👍🏻

connectors/sources/salesforce.py
wangch079: Or we could ignore it, and just rely on `nextRecordsUrl`.

We stop if there's no `nextRecordsUrl`.

navarone-feekery: That's a lot simpler, I'll do that

connectors/sources/salesforce.py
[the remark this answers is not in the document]
navarone-feekery: I looked further into this and I don't think there will be competition, so I've removed the lock.

connectors/sources/salesforce.py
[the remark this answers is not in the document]
navarone-feekery: I've redone this and now there's a single func called `token`.
There's also no need to pre-emptively call it, calling `_auth_headers` will attempt to fetch a token if one doesn't exist yet.

connectors/sources/salesforce.py
navarone-feekery: case feeds also support pagination now

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1475 @@
...
+        # TODO ping something of value (this could be config check instead)
+        await self.session.head(self.base_url)
+
+    async def close(self):
+        self.api_token.clear()
+        if self.session is not None:
+            await self.session.close()
+            del self.session
+
+    async def get_docs(self):
```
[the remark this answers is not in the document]
navarone-feekery: Moved!

navarone-feekery:
@wangch079 Thanks for the review. I implemented most of your feedback. I also moved around a lot of the logic so it wasn't so heavy in `SalesforceClient`.

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+                "label": "Domain",
+                "order": 3,
+                "sensitive": True,
+                "tooltip": "The domain for your Salesforce instance. If your Salesforce URL is 'foo.my.salesforce.com', the domain would be 'foo'.",
+                "type": "str",
+                "value": "",
+            },
+        }
+
+    async def validate_config(self):
```
jedrazb, line 1373: Should we add some programatic validations here?  Probably good to check that the config fields are non-null. We should throw an Exception otherwise.

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+        self.doc_mapper = SalesforceDocMapper(self.base_url)
+
+    def _set_internal_logger(self):
+        self.salesforce_client.set_logger(self._logger)
+
+    @classmethod
+    def get_default_configuration(cls):
+        return {
+            "client_id": {
+                "label": "Client ID",
```
jedrazb, line 1349: Suggestion: add RCF `validations` to configuration fields, [RCF docs](https://github.com/elastic/connectors-python/blob/main/docs/DEVELOPING.md#rich-configurable-fields)

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+            response_body = await response.json()
+            response.raise_for_status()
+            self._token = response_body["access_token"]
+            return self._token
+        except ClientResponseError as e:
+            if 400 <= e.status < 500:
+                # 400s have detailed error messages in body
+                error_message = response_body.get("error", "No error dscription found.")
+                if error_message == "invalid_client":
+                    raise InvalidCredentialsException(
```
wangch079, line 921: This can be added to `skipped_exceptions`

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+        try:
+            await self.salesforce_client.ping()
+            self._logger.debug("Successfully connected to Salesforce.")
+        except Exception as e:
+            self._logger.exception(f"Error while connecting to Salesforce: {e}")
+            raise
+
+    async def get_docs(self, filtering=None):
+        # We collect all content documents and de-duplicate them before downloading and yielding
+        all_content_docs = []
```
jedrazb, line 1389: > We collect all content documents and de-duplicate them before downloading and yielding

What does collecting all content documents mean exactly here? Is this expected to be a lot for large datasets? Can we cause OOMs with that reference?

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+class SalesforceSoqlBuilder:
+    def __init__(self, table):
+        self.table_name = table
+        self.fields = []
+        self.where = ""
+        self.order_by = ""
+        self.limit = ""
+
+    def with_id(self):
+        self.fields.append("Id")
```
wangch079, line 946:
```suggestion
    def with_id(self):
        self.fields.append("Id")
        return self
```
I was expecting the methods in `SalesforceSoqlBuilder` will return `self`, so that it can be chained:

```python
query_builder = SalesforceSoqlBuilder("CaseFeed")
return query_builder.with_id().with_default_metafields().build()
```

connectors/sources/salesforce.py
navarone-feekery: What unique validations do you think these require?

navarone-feekery: Of the possible custom validation types (`'less_than' | 'greater_than' | 'list_type' | 'included_in' | 'regex'`) I don't see anything that we could validate on here.

connectors/sources/salesforce.py
navarone-feekery: Every SObject can have a x number of related ContentDocuments, but the same ContentDocument can be attached to multiple SObjects. This collection is so we can compare them before downloading and avoid downloading duplicate files.

> Can we cause OOMs with that reference?
It's possible. This is something we will find out in QA when we can test against a very large dataset. We can exchange the memory usage for higher CPU usage by checking duplicates on the fly, but we deemed it wasn't necessary because 1. we need to see if it's actually a problem, and 2. it's still a beta connector so it wasn't urgent to fix.

[Discussion link](https://github.com/elastic/connectors-python/pull/1451#discussion_r1294483719)

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+        self._logger = logger_
+
+    @cached_property
+    def session(self):
+        return aiohttp.ClientSession(
+            timeout=aiohttp.ClientTimeout(total=None),
+        )
+
+    async def ping(self):
+        # TODO ping something of value (this could be config check instead)
```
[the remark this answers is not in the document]
navarone-feekery: What config check are you referring to? `validate_config`?
If so, that's called independently by the [sync job runner here](https://github.com/elastic/connectors-python/blob/5c3db0e12e5105018d30df89c5676a59634e19f1/connectors/sync_job_runner.py#L126).

The ping is then handled a [few steps lower](https://github.com/elastic/connectors-python/blob/5c3db0e12e5105018d30df89c5676a59634e19f1/connectors/sync_job_runner.py#L129). Right now the Salesforce ping is just checking if the `domain` that the user registers even exists.

You're right that we can get rid of the comment 👍🏻

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+        await self.session.head(self.base_url)
+
+    async def close(self):
+        self.api_token.clear()
+        await self.session.close()
+        del self.session
+
+    async def get_accounts(self):
+        if not await self._is_queryable("Account"):
+            self._logger.info(
```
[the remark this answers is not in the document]
navarone-feekery: Good call!

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+        Many sobjects require extra data that is taxing on the rate limiter
+        to repeatedly fetch each request.
+        Instead we cache them on the first request for re-use later.
+        """
+        if self._sobjects_cache_by_type is not None:
+            return self._sobjects_cache_by_type
+
+        self._sobjects_cache_by_type = {}
+        self._sobjects_cache_by_type["Account"] = await self._prepare_sobject_cache(
+            "Account"
```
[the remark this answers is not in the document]
navarone-feekery: The data fetched is from a different endpoint. It made sense in a structural sense to me to include the `Account` cache here instead of from the results of `get_account`. This is where all caching of SObjects is handled. It's easier to mentally track what's going on and is only 1 extra request.

connectors/sources/salesforce.py
navarone-feekery: I didn't even know this was possible, thanks!

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1443 @@
...
+        if not await self._is_queryable("Account"):
+            self._logger.info(
+                "Object Account is not queryable, so they won't be ingested."
+            )
+            return
+
+        query = await self._accounts_query()
+        async for records in self._yield_non_bulk_query_pages(query):
+            for record in records:
+                content_docs = self._parse_content_documents(record)
```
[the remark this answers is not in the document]
navarone-feekery: Not sure why I left this actually, it does feel out of place. I've moved it to `SalesforceDataSource`.

navarone-feekery:
@jedrazb @wangch079 thanks heaps for the reviews! I've applied some further changes.

connectors/sources/salesforce.py
jedrazb: Ahh you are right, I somehow linked `required` property to live in `validations`. 👍

connectors/sources/salesforce.py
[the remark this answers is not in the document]
jedrazb: For context, I referred to the `config check` mentioned in the comment.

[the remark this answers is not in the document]
navarone-feekery: I can't even read my own comments, sorry 😅

connectors/sources/salesforce.py
```diff
@@ -0,0 +1,1453 @@
...
+                "VersionDataUrl",
+                "VersionNumber",
+            ],
+        )
+        queryable_docs_fields = [f"ContentDocument.{x}" for x in queryable_docs_fields]
+        queryable_version_fields = [
+            f"ContentDocument.LatestPublishedVersion.{x}"
+            for x in queryable_version_fields
+        ]
+        [f"'{x[1:]}'" for x in TIKA_SUPPORTED_FILETYPES]
```
jedrazb, line 842: Is this an unused statement? Not sure if it's a github PR formatting issue but I have an impression that we initialise this array without any reference to it.

navarone-feekery: Ah that is also in the line below it, it's probably a reformatting error. Removed, thanks for spotting that.

jedrazb (approved):
🚢 ! I expect that the ftests will be added as a follow up PR. Also saw a few `TODO` in comments that hopefully will be fixed before the release

PR #1390, opened 2023-08-02
Navarone/5441 add opportunities and doc builder

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/5441

Adds the following features:

- Ingest `opportunities`
- Build docs correctly with a dedicated builder class
- Check queryable sobjects and fields, and cache when relevant

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -264,8 +264,9 @@ class SalesforceDocBuilder:
     def build_account(self, account):
         owner = account.get("Owner", {})
 
-        opportunities = account.get("Opportunities", {}).get("records", [])
-        opportunity = opportunities[0] if len(opportunities) > 0 else {}
+        opportunities = account.get("Opportunities")
+        opportunity_records = opportunities.get("records", []) if opportunities else []
+        opportunity = opportunity_records[0] if len(opportunity_records) > 0 else {}
         opportunity_url = (
             f"{self.base_url}/{opportunity.get('Id')}" if opportunity else ""
         )
@@ -283,7 +284,7 @@ class SalesforceDocBuilder:
             "open_activities_urls": "",  # TODO
             "opportunity_name": opportunity.get("Name"),
             "opportunity_status": opportunity_status,
-            "opportunity_url": opportunity_url or "",
+            "opportunity_url": opportunity_url,
             "owner": owner.get("Name"),
             "owner_email": owner.get("Email"),
             "rating": account.get("Rating"),
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -105,38 +105,6 @@ class SalesforceClient:
         for record in resp_json.get("records", []):
             record["_id"] = record["Id"]
             yield record
-        self.token_refresh_enabled = enable_refresh
-        self._logger.debug(f"Salesforce token retrieved.")
-
-    @retryable(
-        retries=RETRIES,
-        interval=RETRY_INTERVAL,
-    )
-    async def get_accounts(self):
-        # TODO prepare cache
-
-        query_builder = SalesforceSoqlBuilder("Account")
-        query_builder.with_id()
-        query_builder.with_default_metafields()
-        query_builder.with_fields(
-            self._select_queryable_fields(
-                [
-                    "Name",
-                    "Description",
-                    "BillingAddress",
-                    "Type",
-                    "Website",
-                    "Rating",
-                ]
-            )
-        )
-
-        query = query_builder.build()
-        resp = await self._yield_non_bulk_query_pages(query)
-        resp_json = await resp.json()
-        for record in resp_json.get("records", []):
-            record["_id"] = record["Id"]
-            yield record
 
     async def _yield_non_bulk_query_pages(self, soql_query):
         return await self._api_request(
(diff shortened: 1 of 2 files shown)


connectors/sources/salesforce.py
```diff
@@ -147,13 +232,101 @@ def with_default_metafields(self):
...
+        query_lines.append(f"FROM {self.table_name}")
+        query_lines.append(self.where)
+        query_lines.append(self.order_by)
+        query_lines.append(self.limit)
+
+        return "\n".join([line for line in query_lines if line != ""])
 
-        return "\n".join(query_lines)
+
+class SalesforceDocBuilder:
```
artem-shelkovnikov, line 260: I'd call it a mapper, `Builder` is a legit pattern

connectors/sources/salesforce.py
```diff
@@ -79,32 +84,57 @@ async def get_token(self, enable_refresh=True):
...
+        await self._cache_queryable_sobjects()
+
+    async def _cache_queryable_sobjects(self):
+        resp = await self._api_request(
+            f"{self.base_url}{DESCRIBE_ENDPOINT}", "get", headers=self._auth_headers()
+        )
+        resp_json = await resp.json()
+        self.queryable_sobjects = [
+            x["name"].lower() for x in resp_json["sobjects"] if x["queryable"] is True
+        ]
```
artem-shelkovnikov, line 122: We probably don't need all objects, right?

Also for better syntax, is it possible to make this some sort of async cached property?

connectors/sources/salesforce.py
```diff
@@ -79,32 +84,57 @@ async def get_token(self, enable_refresh=True):
...
+        """User settings can cause fields to be non-queryable
+        This causes
+        """
+        endpoint = DESCRIBE_SOBJECT_ENDPOINT.replace("<sobject>", sobject)
+        resp = await self._api_request(
+            f"{self.base_url}{endpoint}", "get", headers=self._auth_headers()
+        )
+        resp_json = await resp.json()
+        queryable_fields = [f["name"] for f in resp_json["fields"]]
+        return [f for f in fields if f in queryable_fields]
```
artem-shelkovnikov, line 137: Intuitively, it can be cached too. Would it be a good idea to unite this method with `_cache_queryable_objects` as they kinda prepare the class for working?

connectors/sources/salesforce.py
```diff
@@ -79,32 +84,57 @@ async def get_token(self, enable_refresh=True):
...
 
     async def _yield_non_bulk_query_pages(self, soql_query):
         return await self._api_request(
```
artem-shelkovnikov, line 140: Just noticed it here - method name says `yield_XXX_pages` but actually returns results. Should it be renamed, or yield instead?

artem-shelkovnikov: Also - is there paging implemented already?

connectors/sources/salesforce.py
```diff
@@ -147,13 +232,101 @@ def with_default_metafields(self):
...
+        opportunities = account.get("Opportunities")
+        opportunity_records = opportunities.get("records", []) if opportunities else []
+        opportunity = opportunity_records[0] if len(opportunity_records) > 0 else {}
+        opportunity_url = (
+            f"{self.base_url}/{opportunity.get('Id')}" if opportunity else ""
+        )
+        opportunity_status = opportunity.get("StageName", "")
+
+        return {
+            "_id": f"salesforce_connector_{account.get('Id')}",
```
artem-shelkovnikov, line 276: Is the prefix "salesforce_connector_" necessary here?

Also, are ids between accounts, leads, etc unique?

artem-shelkovnikov (commented):
I also noticed usage of `get_attr` in the connector. It might not be for this PR, but can we get rid of `get_attr` from the code and make calls to methods explicit?

connectors/sources/salesforce.py
```diff
@@ -147,13 +232,101 @@ def with_default_metafields(self):
...
+class SalesforceDocBuilder:
+    def __init__(self, base_url):
+        self.base_url = base_url
+
+    def build_account(self, account):
+        owner = account.get("Owner", {})
+
+        opportunities = account.get("Opportunities")
+        opportunity_records = opportunities.get("records", []) if opportunities else []
+        opportunity = opportunity_records[0] if len(opportunity_records) > 0 else {}
```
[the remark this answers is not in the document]
navarone-feekery: The join has a limit of 1 so there should only be one result. It's just returned to us as an array.

connectors/sources/salesforce.py
navarone-feekery: > Is the prefix "salesforce_connector_" necessary here?

The ids for the ws connector documents are all `salesforce_<sobject_id>`. 

> 

They are apparently. You can use the id without any previous url endpoint to access a unique object. [source](https://developer.salesforce.com/forums/?id=9060G000000UVpzQAG)

connectors/sources/salesforce.py
navarone-feekery: I'll rename it. And there's no paging yet.

connectors/sources/salesforce.py
navarone-feekery: I can't find examples in our code of async cached properties. Do we do this yet?

I think I can do this if I introduce [asyncstdlib](https://asyncstdlib.readthedocs.io/en/stable/index.html#) or [async-property](https://pypi.org/project/async-property/). WDYT?

connectors/sources/salesforce.py
navarone-feekery: Same as above: 
I can do this if I introduce [asyncstdlib](https://asyncstdlib.readthedocs.io/en/stable/index.html#) or [async-property](https://pypi.org/project/async-property/).

connectors/sources/salesforce.py
artem-shelkovnikov: I think we don't do it, and it might be an overkill to use a package - you can do caching on your own by storing it in a private variable and returning it if asked for it, additionally you can move this logic to a different class to encapsulate it better.

navarone-feekery:
@artem-shelkovnikov I can split up the `getattr` method to be more explicit in a followup PR

connectors/sources/salesforce.py
artem-shelkovnikov: Oh that's good to know, TIL!

About ids - I think we don't need prefix here, and can just store `sobject_id`

connectors/sources/salesforce.py
navarone-feekery: I tried moving it to another class because it looked nice, but it needs access to the SalesforceClient to make requests and so they both needed a wrapper and it got messy... So I just tried to clean up the code a bit so the fact that it's cached is more obvious. I also reduced the cache size by limiting it to only sobjects/fields that we actually care about.

connectors/sources/salesforce.py
navarone-feekery: This is now cached 💾

navarone-feekery:
@artem-shelkovnikov so in the end I also split up the `getattr` in this PR. It felt right to do when reworking caching.

artem-shelkovnikov (approved):
Looks good!

PR #1393, opened 2023-08-04
[Salesforce] Improve request handling

navarone-feekery:
## Related to https://github.com/elastic/enterprise-search-team/issues/5441

General improvements to the API request/response logic

- Add exception handling
- Add token refresh (not really a refresh)
- Add pagination handling

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -4,14 +4,16 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 """Salesforce source module responsible to fetch documents from Salesforce."""
+from contextlib import contextmanager
 from functools import cached_property
+from threading import Lock
 
 import aiohttp
 from aiohttp.client_exceptions import ClientResponseError
 
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import retryable
+from connectors.utils import CancellableSleeps
 
 RETRIES = 3
 RETRY_INTERVAL = 1
@@ -38,7 +40,7 @@ RELEVANT_SOBJECT_FIELDS = [
 
 
 class RateLimitedException(Exception):
-    """Notifies that Salesforce has begun rate limiting the current accound"""
+    """Notifies that Salesforce has begun rate limiting the current account"""
 
     pass
 
@@ -67,24 +69,27 @@ class RequestError(Exception):
     pass
 
 
+class LockedException(Exception):
+    """Notifies that the current process is locked, only for token generation"""
+
+    pass
+
+
 class SalesforceClient:
     def __init__(self, configuration):
         self._logger = logger
+        self._sleeps = CancellableSleeps()
 
-        self.token = None
-        self._fetching_token = False
         self._queryable_sobjects = None
         self._queryable_sobject_fields = None
 
         self.base_url = BASE_URL.replace("<domain>", configuration["domain"])
-        self.client_id = configuration["client_id"]
-        self.client_secret = configuration["client_secret"]
-        self.token_payload = {
-            "grant_type": "client_credentials",
-            "client_id": self.client_id,
-            "client_secret": self.client_secret,
-        }
-
+        self.api_token = SalesforceAPIToken(
+            self.session,
+            self.base_url,
+            configuration["client_id"],
+            configuration["client_secret"],
+        )
         self.doc_mapper = SalesforceDocMapper(self.base_url)
 
     def set_logger(self, logger_):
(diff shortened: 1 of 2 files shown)


connectors/sources/salesforce.py
```diff
@@ -80,58 +110,66 @@ async def close(self):
...
-        self._logger.debug("Fetching Salesforce token...")
-        resp = await self.session.post(
-            f"{self.base_url}{TOKEN_ENDPOINT}", data=self.token_payload
-        )
-        resp_json = await resp.json()
-        self.token = resp_json["access_token"]
-        self.token_issued_at = resp_json["issued_at"]
-        self.token_refresh_enabled = enable_refresh
-        self._logger.debug("Salesforce token retrieved.")
+    async def get_token(self):
```
artem-shelkovnikov, line 113: I'd probably move token logic into separate class.

Also, it might make sense to wrap the logic in a mutex.

That also gave me a hint, that we're missing this logic in Sharepoint Online connector!

connectors/sources/salesforce.py
```diff
@@ -80,58 +110,66 @@ async def close(self):
...
+            response = await self._post(
+                f"{self.base_url}{TOKEN_ENDPOINT}", data=self.token_payload
+            )
+            response_body = await response.json()
+            response.raise_for_status()
+            self.token = response_body["access_token"]
+        except ClientResponseError as e:
+            if 400 <= e.status < 500:
+                # 400s have detailed error messages in body
+                error_resp = self._handle_token_response_body_error(response_body)
```
artem-shelkovnikov, line 129: Is `e.message` different than response_body?

connectors/sources/salesforce.py
```diff
@@ -169,26 +206,118 @@ async def _select_queryable_fields(self, sobject, fields):
...
+            page_count = 1
+            next_url = response.get("nextRecordsUrl")
+            while True:
+                next_page = await self._get_json(next_url)
+                yield next_page.get("records")
+
+                page_count += 1
+                if (
+                    next_page.get("done") is True
+                    or page_count >= QUERY_PAGINATION_LIMIT
```
artem-shelkovnikov, line 227: I believe usage of `nextRecordsUrl` eliminates the need to have QUERY_PAGINATION_LIMIT? I thought it applies only to skip/take pagination

connectors/sources/salesforce.py
```diff
@@ -169,26 +206,118 @@ async def _select_queryable_fields(self, sobject, fields):
...
 
     def _auth_headers(self):
         return {"authorization": f"Bearer {self.token}"}
 
-    async def get_request(self, endpoint, params=None):
-        # TODO improve error handling
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+    )
```
artem-shelkovnikov, line 242: I think you'll need custom retryable logic here - no need to retry on invalid query, probably on throttling too

connectors/sources/salesforce.py
```diff
@@ -169,26 +206,118 @@ async def _select_queryable_fields(self, sobject, fields):
...
+    async def _get_json(self, url, params=None):
+        response_body = None
+
+        try:
+            response = await self._get(url, params=params)
+            response_body = await response.json()
+            # We get the response body before raising for status as it contains vital error information
+            response.raise_for_status()
```
artem-shelkovnikov, line 250: Hmmm, I haven't used this method before - I thought response will raise by default? Or is it only if you use it in context manager style?

connectors/sources/salesforce.py
```diff
@@ -169,26 +206,118 @@ async def _select_queryable_fields(self, sobject, fields):
...
+    async def _handle_client_response_error(self, response_body, e):
+        exception_details = f"status: {e.status}, message: {e.message}"
+
+        if e.status == 401:
+            self._logger.warning(
+                f"Token expired, attemping to fetch new token. Status: {e.status}, message: {e.message}"
+            )
+            # The user can alter the lifetime of issued tokens, so we don't know when they expire
+            # Therefore we fetch the token when we encounter an error rather than when it expires
+            await self.get_token()
```
artem-shelkovnikov, line 276: Might it be worth to actually wipe the token before fetching?

connectors/sources/salesforce.py
navarone-feekery: I just didn't want an endless loop, in case _something_ went wrong. It's really just a safeguard.

connectors/sources/salesforce.py
navarone-feekery: If you set `raise_for_status=True` in either the session or the request, then it will raise (it's `False` by default)

connectors/sources/salesforce.py
navarone-feekery: Yeah they are different, for example if the credentials are invalid it throws a 400 with the message a `bad_request`. The actual cause (`invalid_client`) is then inside the response body.

connectors/sources/salesforce.py
navarone-feekery: mutex + separate class sounds really clean, I like it! Thanks!

connectors/sources/salesforce.py
```diff
@@ -35,13 +37,42 @@
 ]
 
 
+class RateLimitedException(Exception):
+    """Notifies that Salesforce has begun rate limiting the current accound"""
```
wangch079, line 41:
```suggestion
    """Notifies that Salesforce has begun rate limiting the current account"""
```

connectors/sources/salesforce.py
```diff
@@ -169,26 +206,118 @@ async def _select_queryable_fields(self, sobject, fields):
...
-    async def _get_non_bulk_query_pages(self, soql_query):
-        return await self.get_request(
-            QUERY_ENDPOINT,
+    async def _yield_non_bulk_query_pages(self, soql_query):
+        """loops through query response pages and yields lists of records"""
+        response = await self._get_json(
+            f"{self.base_url}{QUERY_ENDPOINT}",
             params={"q": soql_query},
         )
+        yield response.get("records")
```
wangch079, line 215: Why not use `while True` and `break`?

connectors/sources/salesforce.py
```diff
@@ -169,26 +206,118 @@ async def _select_queryable_fields(self, sobject, fields):
...
+                    error in error_codes
+                    for error in [
+                        "INVALID_FIELD",
+                        "INVALID_TERM",
+                        "MALFORMED_QUERY",
+                    ]
+                )
+                in error_codes
+            ):
+                raise InvalidQueryException(
```
wangch079, line 299: If it's `RateLimitedException` or `InvalidQueryException`, we should not retry

wangch079: You can use this when it's merged: https://github.com/elastic/connectors-python/pull/1395

navarone-feekery: That was a really convenient update, thank you for adding it!

navarone-feekery:
@artem-shelkovnikov @wangch079 I think I've addressed your comments. I need to merge this to unblock another PR, so if there's anything else that you think requires changing let me know and I'll make a followup PR 🙂

PR #1394, opened 2023-08-04
Try out buildkite diff plugin to run ftest once connector changes

artem-shelkovnikov:
Idea - if any of connector source files or fixture files change, we want to run ftest for this PR to check that nothing broke. 

We had nightly builds failing recently due to changes pushed that break functional tests, and this PR aims to reduce the size of this issue.


artem-shelkovnikov:
Example of PR without connector changes: https://buildkite.com/elastic/connectors-python/builds/7746
Example of PR with connector changes that break ftest: https://buildkite.com/elastic/connectors-python/builds/7744

jedrazb (approved):
Great idea to keep the ftests green 💚

PR #1395, opened 2023-08-07
Add skipped exceptions to retryable

wangch079:

The `retryable` decorator will retry when an exception is raised. But sometimes, we may want to skip some exceptions, e.g. `RateLimitingException` means it reached the rate limit, and `ClientError` may indicate the request body was formed in the wrong format.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
Good stuff!

connectors/utils.py
```diff
@@ -412,25 +412,41 @@ class UnknownRetryStrategyError(Exception):
     pass
 
 
-def retryable(retries=3, interval=1.0, strategy=RetryStrategy.LINEAR_BACKOFF):
+def retryable(
+    retries=3,
+    interval=1.0,
+    strategy=RetryStrategy.LINEAR_BACKOFF,
+    skipped_exceptions=None,
```
praveen-elastic, line 419: I believe there can be a better way to implement this. How about accepting the exceptions that you want to retry. The logic would be clean for that. Also you can add another list for those exceptions where you want to retry without sleep. 
```
def retryable(
    retries=3,
    interval=1.0,
    strategy=RetryStrategy.LINEAR_BACKOFF,
    exception_list=(Exception,),
    exception_with_sleep=(Exception,),
):
    def wrapper(func):
        if inspect.isasyncgenfunction(func):
            return retryable_async_generator(
                func, retries, interval, strategy, exception_list, exception_with_sleep
            )
        else:
            return retryable_async_function(
                func, retries, interval, strategy, exception_list, exception_with_sleep
            )

    return wrapper
```

and further you can do: 

```
def retryable_async_function(
    func, retries, interval, strategy, exception_list, exception_with_sleep
):
    @functools.wraps(func)
    async def wrapped(*args, **kwargs):
        retry = 1
        while retry <= retries:
            try:
            except exception_list as e:
                if retry >= retries:
                    raise e
                if exception_list in exception_with_sleep:
                    await apply_retry_strategy(strategy, interval, retry)
                retry += 1

    return wrapped
```

artem-shelkovnikov: I would avoid introducing `retryable` to the places that use Cancellable Sleeps - now retryable prevents app from stopping/restarting because of its global Sleep

artem-shelkovnikov: `exception_with_sleep` IMO makes function unnecessary complex too - it's supposed to be a simple wrapper, but adding more and more arguments to it just makes it harder to use and understand

wangch079: What if I want to raise without retry for some exception? E.g.:

https://github.com/elastic/connectors-python/blob/a03cc1c9ba9c0d086ccd079b49c24eec5ec86c79/connectors/sources/postgresql.py#L264-L273

praveen-elastic: Hey, for that we can just avoid passing those in`exception_list`. But In this case of postgresql, `retryable` should work out of the box, as here retry is only being used for General Exception which is what `retryable` is implemented with currently, i.e. retry on General Exception.

praveen-elastic (approved):

connectors/utils.py
wangch079: > But In this case of postgresql, `retryable` should work out of the box

If you see the code snippet, it tries to retry for all exceptions except `InternalClientError` and `ProgrammingError`, and the current `retryable` couldn't handle it.


> for that we can just avoid passing those in`exception_list`.

You need to define something like `all exceptions except x, y, z`, which is hard if you only define `retriable exceptions`.

praveen-elastic: > If you see the code snippet, it tries to retry for all exceptions except InternalClientError and ProgrammingError, and the current retryable couldn't handle it.
Won't General Exception catch all the raised exceptions ?

PR #1396, opened 2023-08-07
Implement Jira DLS

### LINKED ISSUE #1398: add DLS support for JIRA as defined in the discovery

moxarth-elastic:
## Closes #1398 

Adds the document level security for Jira connector.

#### Pre-Review Checklist


connectors/sources/jira.py
```diff
@@ -56,11 +61,43 @@
     ISSUE_DATA: "/rest/api/2/issue/{id}",
     ATTACHMENT_CLOUD: "/rest/api/2/attachment/content/{attachment_id}",
     ATTACHMENT_SERVER: "/secure/attachment/{attachment_id}/{attachment_name}",
+    USERS: "/rest/api/3/users/search",
+    PROJECT_PERMISSIONS: "/rest/api/2/user/permission/search?projectKey={project_key}&permissions=BROWSE_PROJECTS",
+    PROJECT_ROLE_MEMBERS: "/rest/api/3/project/{project_key}/role/{role_id}",
+    ISSUE_SECURITY_LEVEL: "/rest/api/2/issue/{issue_key}?fields=security",
+    SECURITY_LEVEL_MEMBERS: "/rest/api/3/issuesecurityschemes/level/member?maxResults={max_results}&startAt={start_at}&levelId={level_id}&expand=user,group,projectRole",
```
timgrein, line 68: suggestion: I think functions called `project_permissions`, `project_role_members` etc. would be more explicit. Here it's not obvious that these format string constants expect arguments. A function/method would make it obivous that something needs to be passed (and especially what).

connectors/sources/jira.py
```diff
@@ -56,11 +61,43 @@
...
+def _prefix_account_id(user_id):
+    return _prefix_identity("account_id", user_id)
+
+
+def _prefix_group_id(group_id):
+    return _prefix_identity("group", group_id)
+
+
+def _prefix_role_key(role_key):
+    return _prefix_identity("application_role", role_key)
```
timgrein, line 99: (more like a note to myself): We should probably move some generic functionality into a separate `access_control.py` module to increase reusability. I think I've started that for the GMail connector before my holidays, so you can keep it here for this PR.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+                    }
+                },
+            }
+        }
+
+    def _is_active_atlassian_user(self, user_info):
+        user_url = user_info.get("self")
+        if not user_url:
+            self._logger.debug(
+                f"Skipping user: {user_info.get('displayName', 'user')} due to the absence of a personal URL."
```
timgrein, line 440: nit: As we're using `user_info.get('displayName', 'user')` at three different places I think it would be useful to have a variable `user_name`, which can be reused.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+
+    def _decorate_with_access_control(self, document, access_control):
+        if self._dls_enabled():
+            document[ACCESS_CONTROL] = list(
+                set(document.get(ACCESS_CONTROL, []) + access_control)
+            )
+
+        return document
+
+    async def _user_information_list(self, project):
```
timgrein, line 466: suggestion: I would suggest that we move this method to the `JiraClient` class. I would also hide the low-level methods `api_call` and `paginated_api_call` methods then.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+        async for actors in self._user_information_list(project=project):
+            for actor in actors:
+                if actor["accountType"] == ATLASSIAN:
+                    access_control.add(
+                        _prefix_account_id(user_id=actor.get("accountId"))
+                    )
+                    access_control.add(_prefix_username(user=actor.get("displayName")))
+        return list(access_control)
+
+    async def _issue_security_level(self, issue_key):
```
timgrein, line 486: Same as above: can probably live inside the `JiraClient`.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+        async for response in self.jira_client.api_call(
+            url_name=ISSUE_SECURITY_LEVEL, issue_key=issue_key
+        ):
+            yield await response.json()
+
+    async def _issue_security_level_members(self, level_id):
```
timgrein, line 492: Same as above: can probably live inside the `JiraClient`.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+        async for response in self.jira_client.paginated_api_call(
+            url_name=SECURITY_LEVEL_MEMBERS, level_id=level_id
+        ):
+            yield response
+
+    async def _project_role_members(self, project, role_id, access_control):
```
timgrein, line 498: Same as above: can probably live inside the `JiraClient`.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+
+        access_control = set()
+        async for response in self._issue_security_level(issue_key=issue_key):
+            if security := response.get("fields", {}).get("security"):
+                level_id = security.get("id")
+                async for members in self._issue_security_level_members(
+                    level_id=level_id
+                ):
+                    for actor in members["values"]:
+                        if actor.get("holder", {}).get("type") == "user":
```
timgrein, line 533: suggestion: `actor.get("holder", {}).get("type")` is used at three different places. I would suggest to use a variable here.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+                                # Project Role - `atlassian-addons-project-access` with id 10003 is not needed for DLS
+                                if role_id == 10003:
+                                    continue
+                                access_control = await anext(
+                                    self._project_role_members(
+                                        project=project,
+                                        role_id=role_id,
+                                        access_control=access_control,
+                                    )
+                                )
```
timgrein, line 562: Small readability nit:

```suggestion
                                is_addons_projects_access = role_id == 10003
                                if not is_addons_projects_access:
                                  access_control = await anext(
                                      self._project_role_members(
                                          project=project,
                                          role_id=role_id,
                                          access_control=access_control,
                                      )
                                  )
```

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+        The document contains all groups of a user and his email and/or username under `query.template.params.access_control`.
+
+        Returns:
+            dict: dictionary representing an user access control document
+            {
+                "_id": "some.user@spo.com",
+                "identity": {
+                    "email": "email:some.user@spo.com",
+                    "username": "user:some.user",
+                    "user_id": "user_id:some user id"
```
timgrein, line 586: Here you're using `email`, `username` and `user_id` as the fields for `identity`, but I only see `account_id` and `display_name` down below. Should we adjust that?

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+        _prefixed_role_keys = {
+            _prefix_role_key(role_key=role.get("key", ""))
+            for role in user.get("applicationRoles", {}).get("items", [])
+        }
+
+        user_document = {
+            "_id": account_id,
+            "identity": {
+                "account_id": _prefixed_account_id,
+                "display_name": _prefixed_user_name,
```
timgrein, line 620: question: Do we've the email of a user? The email is usually the same for users across different systems.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+            },
+            "created_at": iso_utc(),
```
timgrein, line 622: question: is there no `creationDate` field or something similar returned from the Jira API?

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
 
+        access_control = (
+            [_prefixed_account_id]
+            + list(_prefixed_group_ids)
+            + list(_prefixed_role_keys)
+        )
+
+        return user_document | self.access_control_query(access_control=access_control)
+
+    async def _fetch_all_users(self):
```
timgrein, line 633: Should also live in the `JiraClient` IMO.

connectors/sources/jira.py
```diff
@@ -320,8 +371,310 @@ def get_default_configuration(cls):
...
+        async for users in self.jira_client.api_call(url_name=USERS):
+            yield await users.json()
+
+    async def _fetch_user(self, user_url):
```
timgrein, line 637: Should also live in the `JiraClient` IMO.

tests/sources/test_jira.py
```diff
@@ -155,6 +155,110 @@
...
+                "groupId": "d614bbb8-703d-4d41-a51a-0c3cfcef9318",
+            },
+        },
+    ],
+}
+
+ACCESS_CONTROL = "_allow_access_control"
+
 
 class mock_ssl:
```
timgrein, line 263:
```suggestion
class MockSsl:
```

nit: class names should be PascalCase.

connectors/sources/jira.py
moxarth-elastic: Sure, I'll tweak a name of these constants. Renaming them to `PROJECT_PERMISSIONS_BY_KEY` and `PROJECT_ROLE_MEMBERS_BY_ROLE_ID`.

connectors/sources/jira.py
moxarth-elastic: Sure, let me wrap it in a variable.

connectors/sources/jira.py
moxarth-elastic: Yes, this should be adjusted.

connectors/sources/jira.py
moxarth-elastic: No, we don't. We will only get an email of the authenticated user, not for all users, so email is not included in the identity metadata.

connectors/sources/jira.py
moxarth-elastic: No, there's no `creationDate` or similar field getting from the Jira API response.

connectors/sources/jira.py
moxarth-elastic: I'll address this change in a separate PR. Created an issue regarding the same - https://github.com/elastic/connectors-python/issues/1417

connectors/sources/jira.py
moxarth-elastic: https://github.com/elastic/connectors-python/issues/1417

tests/sources/test_jira.py
```diff
@@ -216,6 +320,25 @@ def side_effect_function(url, ssl):
...
+        return get_json_mock(mock_response=MOCK_USER)
+    elif url == f"{HOST_URL}/rest/api/2/issue/TP-1?fields=security":
+        return get_json_mock(mock_response=MOCK_SECURITY_LEVEL)
+    elif (
+        url
+        == f"{HOST_URL}/rest/api/3/issuesecurityschemes/level/member?maxResults=100&startAt=0&levelId=level-1&expand=user,group,projectRole"
+    ):
+        return get_json_mock(mock_response=MOCK_SECURITY_LEVEL_MEMBERS)
+    elif url == f"{HOST_URL}/rest/api/3/project/TP/role/10002":
+        return get_json_mock(mock_response=MOCK_PROJECT_ROLE_MEMBERS)
```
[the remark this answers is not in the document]
moxarth-elastic: https://github.com/elastic/connectors-python/issues/1417

connectors/sources/jira.py
timgrein: I would still prefer methods as nearly every mature IDE can easily autocomplete method parameters. I think a formatted string doesn't have a clear interface, which makes it harder to understand. Having FF in mind I can definitely live with the current solution, but would prefer creating a follow-up issue for making it cleaner using explicit methods.

moxarth-elastic: Sure, even feel free to add it in the existing issue https://github.com/elastic/connectors-python/issues/1417 👍

timgrein (approved):

PR #1401, opened 2023-08-08
Refactor execute_query function

wangch079:

This PR refactor the `execute_query` function to `fetch` function, which
1. yields column names first if `fetch_columns` is true
2. fetches rows by `fetch_size` and yields rows one by one
3. uses `@retryable` to handle retry logic

#### Pre-Review Checklist


artem-shelkovnikov:
Curious - why do we need `fetch_all` even? Having only one `fetch` should be sufficient.

wangch079:
>Curious - why do we need fetch_all even? Having only one fetch should be sufficient.

You are absolutely right. I will make the change to expose only one `fetch` function.

PR #1402, opened 2023-08-08
Add MSSQL client abstraction

wangch079:

This PR adds MSSQLClient abstraction, which will handle the connection to MSSQL server.

#### Pre-Review Checklist


connectors/sources/mssql.py
```diff
@@ -75,25 +313,26 @@ def __init__(self, configuration):
...
+            ssl_enabled=self.configuration["ssl_enabled"],
+            ssl_ca=self.configuration["ssl_ca"],
+            validate_host=self.configuration["validate_host"],
+            retry_count=self.configuration["retry_count"],
+            fetch_size=self.configuration["fetch_size"],
+            logger_=self._logger,
+        )
+
+    def _set_internal_logger(self):
+        self.mssql_client.set_logger(self._logger)
```
artem-shelkovnikov, line 335: Probably not needed since you set it beforehands?

connectors/sources/mssql.py
```diff
@@ -62,6 +62,244 @@ def all_schemas(self):
...
+        return await anext(
+            self._execute_query(
+                query=self.queries.table_last_update_time(
+                    schema=self.schema,
+                    table=table,
+                ),
+            )
+        )
+
+    async def data_streamer(self, table):
```
artem-shelkovnikov, line 223: Maybe rename to `yield_table_data`?

connectors/sources/mssql.py
```diff
@@ -62,6 +62,244 @@ def all_schemas(self):
...
+            fetch_many (bool): Flag to use fetchmany method. Defaults to False.
+
+        Raises:
+            exception: Raise an exception after retrieving
+
+        Yields:
+            list: Column names and query response
+        """
+        retry = 1
+        yield_once = True
```
artem-shelkovnikov, line 260: Should this be removed? Function should explicitly return or yield, not change behaviour on a flag.

connectors/sources/mssql.py
```diff
@@ -62,6 +62,244 @@ def all_schemas(self):
...
+                        for row in rows:
+                            yield row
+
+                        rows_fetched += rows_length
+                        await asyncio.sleep(0)
+                else:
+                    yield cursor.fetchall()  # pyright: ignore
+                break
+            except (InternalClientError, ProgrammingError):
+                raise
```
artem-shelkovnikov, line 292: Why only these two?

connectors/sources/mssql.py
```diff
@@ -62,6 +62,244 @@ def all_schemas(self):
...
+            except Exception as exception:
+                self._logger.warning(
+                    f"Retry count: {retry} out of {self.retry_count}. Exception: {exception}"
+                )
+                if retry == self.retry_count:
+                    raise exception
```
artem-shelkovnikov, line 298:
```suggestion
                    raise
```

connectors/sources/mssql.py
```diff
@@ -62,6 +62,244 @@ def all_schemas(self):
...
+                await asyncio.sleep(DEFAULT_WAIT_MULTIPLIER**retry)
```
artem-shelkovnikov, line 299: Is it possible to use CancellableSleeps here?

connectors/sources/mssql.py
wangch079: When a data source is instantiated, `self._logger` is set to the global `logger`:
https://github.com/elastic/connectors-python/blob/ed88820df1b622c68a5865baba152a47420221bf/connectors/source.py#L347-L349

and set by `set_logger`:
https://github.com/elastic/connectors-python/blob/ed88820df1b622c68a5865baba152a47420221bf/connectors/sync_job_runner.py#L113-L116

The `set_logger` will subsequently call `_set_internal_logger`:

https://github.com/elastic/connectors-python/blob/ed88820df1b622c68a5865baba152a47420221bf/connectors/source.py#L364-L366

We do this because we don't want to change the signature of `BaseDataSource`
https://github.com/elastic/connectors-python/blob/ed88820df1b622c68a5865baba152a47420221bf/connectors/source.py#L347

connectors/sources/mssql.py
wangch079: This name is actually from the original code:
https://github.com/elastic/connectors-python/blob/e9b4e577b4311d8fd8ca01cddae14ec5528ae9c7/connectors/sources/generic_database.py#L396-L404

WDYT about `table_data_generator`?

connectors/sources/mssql.py
wangch079: The name `yield_once` is misleading. It means only yield column names first, then rows.

I don't want to make changes here, because once #1401 is merged, this PR will rebase and remove this function.

connectors/sources/mssql.py
wangch079: These are from the original code:
https://github.com/elastic/connectors-python/blob/e9b4e577b4311d8fd8ca01cddae14ec5528ae9c7/connectors/sources/generic_database.py#L247-L248

They are also caught here:

https://github.com/elastic/connectors-python/blob/e9b4e577b4311d8fd8ca01cddae14ec5528ae9c7/connectors/sources/generic_database.py#L434-L437

I don't know excatly why

connectors/sources/mssql.py
artem-shelkovnikov: I see, might be worth to keep it the way it was before too, not a biggie

connectors/sources/mssql.py
wangch079: Hi @moxarth-elastic, git blame told me it was added by you. Why these 2 exceptions are not retryable?

connectors/sources/mssql.py
wangch079: This is not needed once #1401 is merged.

connectors/sources/mssql.py
wangch079: Same, this is not needed once #1401 is merged.

artem-shelkovnikov: @retryable wrapper does not use CancellableSleeps and that's why I'm always reluctant to use it. Should we have a tech debt case for it?

artem-shelkovnikov (dismissed):

connectors/sources/mssql.py
wangch079: >https://github.com/retryable wrapper does not use CancellableSleeps and that's why I'm always reluctant to use it. Should we have a tech debt case for it?

If you want to use `CancellableSleeps` in `retryable`, you have to pass a `CancellableSleeps` instance to it.

connectors/sources/mssql.py
moxarth-elastic: hey @wangch079, `InternalClientError` is raised by asyncpg lib when it does not process the request like "invalid datatype" and same goes for `ProgrammingError`, in sqlalchemy when we pass wrong query then it raises ProgrammingError. i.e; table not found. So, I don't think we can retry for this type of exceptions.

tests/sources/test_generic_database.py
```diff
@@ -76,16 +76,35 @@ def fetchmany(self, size):
...
+                        "abcd",
+                    ),
+                    (
+                        2,
+                        "xyz",
+                    ),
+                ]
         return []
 
     def fetchall(self):
```
wangch079, line 110: This is kept for `test_oracle` to pass the test. Will be removed in the subsequent PR.

artem-shelkovnikov (approved):

PR #1404, opened 2023-08-08
move default configs to code

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/5056

First pass at moving default configs from YAML to our codebase. This will declutter our YAML file for the average user.

The only real configs we need are:
- Elasticsearch credentials (and only if `elastic:changeme` is insufficient)
- list of `connectors` (ids+service_type+api_key)

Everything else should be optional, since we have defaults.

Out of scope for now:
- leveraging a more complex schema, like the ent-search ConfigParser. This issue is about simplifying the config process, not making bad configs more easily detectable.
- Constantizing all the config names and default values. We don't currently have a need to refactor how we're fetching configs - we just want defaults encoded.

#### Pre-Review Checklist

## Release Note

The connectors-python YAML config file no longer needs to be fully explicit, and can omit any values where you wish to use the Elastic default.


tests/sources/fixtures/jira/config.yml
```diff
@@ -1,30 +1,4 @@
...
-  idling: 5
-  heartbeat: 300
-  max_errors: 20
-  max_errors_span: 600
-  max_concurrent_syncs: 1
-  job_cleanup_interval: 300
-  log_level: INFO
 
 connectors:
   -
```
navarone-feekery, line 4: It seems below this you've left one `sources` section

navarone-feekery (dismissed):

wangch079 (commented):
What do you think we keep those configs (with default values) in `config.yml` and commented out? Like what we did in https://github.com/elastic/ent-search/blob/main/dist/enterprise-search/package_assets/config/enterprise-search.yml

So whenever a user wants to customize the config, he/she can easily uncomment and edit.

seanstory:
> What do you think we keep those configs (with default values) in `config.yml` and commented out?



I waffled back and forth on this. Ent-search also has a configuration reference doc, but it's not in the same repo, and isn't present in the artifact when you download/install it. So it definitely makes sense that you'd want to have some reference available for someone who's poking at the artifact through the terminal. 


However, we have the [config.md](https://github.com/elastic/connectors-python/blob/main/docs/CONFIG.md) which is also there in the repo, and has exactly the same info as what we'd put in the YAML, commented out. My loosely-held instinct was that this would be too redundant, and it would be better to have a minimal YAML file. But I'm open to being convinced otherwise. WDYT, @wangch079?

wangch079:
>My loosely-held instinct was that this would be too redundant

I agree we shouldn't have redundant info.

As a developer, I'd prefer to have commented-out configs with comments (better if the config name is self-explanatory) in `config.yml`, instead of looking for docs in another markdown file.

Sorry, I don't know why I added `CONFIG.md` in the first place. I should add them in the `config.yml`.

navarone-feekery (approved):

vidok (commented):
This is awesome! 

nit: While reading `config.yml` I caught myself thinking that I don't know if a comment belongs to the configuration option below or above. Would it be possible to add extra empty lines between the config options?

vidok (approved):

seanstory:
@vidok addressed in https://github.com/elastic/connectors-python/pull/1456

config.yml
```diff
@@ -1,56 +1,214 @@
...
+# NOTE: Elastic Connectors comes with reasonable defaults.
+#       Before adjusting the configuration, make sure you understand what you
+#       are trying to accomplish and the consequences.
+#
+# ------------------------------- Connectors -------------------------------
+#
+#  The list of connector clients/customized connectors configurations.
+#    Each object in the list requires `connector_id` and `service_type`.
+#    An example is:
+#    connectors:
```
wangch079, line 12: The example should be:
```yaml
#    connectors:
#      - 
#        connector_id: changeme # the ID of the connector.
#        service_type: changeme # The service type of the connector.
#        api_key: changeme # The Elasticsearch API key used to write data into the content index.
```

config.yml
```diff
@@ -1,56 +1,214 @@
...
+#
+#  The ID of the connector.
+#    (Deprecated. Configure the connector client in an object in the `connectors` list)
+#
+#connector_id: null
+#
+#  The service type of the connector.
+#    (Deprecated. Configure the connector client in an object in the `connectors` list)
+#
+#service_type: null
```
wangch079, line 27: Do you think it's better to use `changeme` as the default value?

wangch079 (commented):
Hi @seanstory, left two comments.

Although it's strictly required, do you think we want to remove some configs from [here](https://github.com/elastic/ent-search/blob/main/dist/shared/cloud/coservices/python-connectors.yml).

PR #1425, opened 2023-08-10
[Refactoring] Remove shadow variables

timgrein:
(No issue, since this is super trivial/just three renamings)

It seems to be a bad practice in python to shadow built-in names like `id` and `filter`. Variables with these names should end with `_` for being able to distinguish them from built-in functions and to not confuse many IDEs.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
Good stuff!

What do you think if we enable the linter rules for it? https://beta.ruff.rs/docs/rules/#flake8-builtins-a

timgrein:
> https://beta.ruff.rs/docs/rules/#flake8-builtins-a

Let's add it 🚀

Issue #1435, opened 2023-08-11
[SPO] Some entities do not populate `webUrl` field


artem-shelkovnikov:
## Bug Description
Some of the entities synced by Sharepoint Online connector do not sync `webUrl` field - this field is a reference to the resource (document/webpage) on Sharepoint Online instance.

It's important, cause not having it makes it impossible to use found object, essentially.

Objects that were reported to have this problem:
- Site Pages
- List Item Attachments

### To Reproduce
1. Create a SPO connector 
2. Sync a Sharepoint Online instance that has some site pages and list item attachments
3. Check data synced into the index
4. See that some records have no `webUrl` field

## Expected behavior
All entities have `webUrl` field

PR #1440, opened 2023-08-11
Populate webUrl for site pages and list item attachments

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors-python/issues/1435

This PR updates the logic that yields site pages and list item attachments. These fields are missing from objects as they are loaded with Rest API.

For site pages webUrl is populated from `EncodedAbsUrl` that is requested additionally.
For attachments it's populated from existing fields.

#### Pre-Review Checklist

## Release Note

SPO connector now populates webUrl for Site Pages and List Item Attachments

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -764,27 +764,25 @@ class SharepointOnlineClient:
         self._validate_sharepoint_rest_url(site_web_url)
 
         # select = "Id,Title,LayoutWebpartsContent,CanvasContent1,Description,Created,AuthorId,Modified,EditorId"
-        select = ""  # ^ is what we want, but site pages don't have consistent schemas, and this causes errors. Better to fetch all and slice
+        select = "*,EncodedAbsUrl"  # ^ is what we want, but site pages don't have consistent schemas, and this causes errors. Better to fetch all and slice
         url = f"{site_web_url}/_api/web/lists/GetByTitle('Site%20Pages')/items?$select={select}"
 
         try:
             async for page in self._rest_api_client.scroll(url):
                 for site_page in page:
                     yield {
-                        k: site_page.get(k, None)
-                        for k in (
-                            "Id",
-                            "Title",
-                            "LayoutWebpartsContent",
-                            "CanvasContent1",
-                            "WikiField",
-                            "Description",
-                            "Created",
-                            "AuthorId",
-                            "Modified",
-                            "EditorId",
-                            "odata.id",
-                        )
+                        "Id": site_page.get("Id"),
+                        "Title": site_page.get("Title"),
+                        "webUrl": site_page.get("EncodedAbsUrl"),
+                        "LayoutWebpartsContent": site_page.get("LayoutWebpartsContent"),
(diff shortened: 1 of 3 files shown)


timgrein (dismissed):

connectors/sources/sharepoint_online.py
```diff
@@ -764,27 +764,25 @@ async def site_pages(self, site_web_url):
         self._validate_sharepoint_rest_url(site_web_url)
 
         # select = "Id,Title,LayoutWebpartsContent,CanvasContent1,Description,Created,AuthorId,Modified,EditorId"
-        select = ""  # ^ is what we want, but site pages don't have consistent schemas, and this causes errors. Better to fetch all and slice
+        select = "*,EncodedAbsUrl"  # ^ is what we want, but site pages don't have consistent schemas, and this causes errors. Better to fetch all and slice
```
seanstory, line 767: TIL `*` is an option.

We're sure that `EncodedAbsUrl` is going to be available always, and not cause errors on some differently-schema'd pages?

connectors/sources/sharepoint_online.py
```diff
@@ -1846,6 +1845,18 @@ async def site_list_items(
                     list_item_attachment[
                         "_original_filename"
                     ] = list_item_attachment.get("FileName", "")
+                    if (
+                        "ServerRelativePath" in list_item_attachment
+                        and "DecodedUrl" in list_item_attachment["ServerRelativePath"]
```
seanstory, line 1850:
```suggestion
                        and "DecodedUrl" in list_item_attachment.get("ServerRelativePath", "")
```
if it's `None` for some reason, this would error.

```
>>> "foo" in None
TypeError: argument of type 'NoneType' is not iterable
```

connectors/sources/sharepoint_online.py
```diff
@@ -1780,8 +1776,11 @@ async def drive_items(self, site_drive, max_drive_item_age):
                 yield drive_item, self.download_function(drive_item, max_drive_item_age)
 
     async def site_list_items(
-        self, site_id, site_list_id, site_web_url, site_list_name, site_access_control
+        self, site, site_list_id, site_list_name, site_access_control
     ):
+        site_id = site["id"]
+        site_web_url = site["webUrl"]
+        site_collection = site["siteCollection"]["hostname"]
```
seanstory, line 1783:
```suggestion
        site_id = site.get("id")
        site_web_url = site.get("webUrl")
        site_collection = site.get("siteCollection", {}).get("hostname")
```

connectors/sources/sharepoint_online.py
artem-shelkovnikov: Looks like it's available for any list item object, so it's available for all Site Page objects because of it.

connectors/sources/sharepoint_online.py
artem-shelkovnikov: Yeah, but I check it in the line above - `"ServerRelativePath" in list_item_attachment`, is it not sufficient?

navarone-feekery (approved):

Issue #1443, opened 2023-08-14
SPO connector fetches granular permissions even if DLS is disabled


artem-shelkovnikov:
## Bug Description
By default SPO connector is created with the following configuration:
- DLS **is disabled**
- Fetch drive item permissions **is enabled**
- Fetch unique page permissions **is enabled**
- Fetch unique list permissions **is enabled**
- Fetch unique list item permissions **is enabled**

Some of the "enabled" above is handled incorrectly - e.g. unique permissions are fetched for the items even if DLS is disabled.

That causes connector to take tremendous amount of time to fetch data even if DLS is fully disabled due to API calls done to the permission endpoints.

### To Reproduce
1. Create a Sharepoint Online connector, connect a service to it that has `log level: DEBUG`
2. Make sure DLS is disabled
3. Run a sync
4. Observe logs and requests - you'll notice messages like `Fetching unique page permissions for page with id` and lots of throttling

## Expected behavior
Connector is not throttled as much, no permissions are fetched for items when DLS is disabled

PR #1446, opened 2023-08-14
Add _dls_enabled() before fetching unique permissions

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors-python/issues/1443

Adding `_dls_enabled()` condition check for logic that fetches unique DLS permissions. Without this check, permissions are fetched even if DLS is disabled, making connector really slow (permission calls are expensive for DLS, plus some of them call for permissions in N+1 style: per item, which slows thing down more)

#### Pre-Review Checklist


timgrein (approved):
LGTM, nice catch!

Issue #1194, opened 2023-07-05
Make framework validate connector configuration instead of making connectors do it


artem-shelkovnikov:
### Problem Description

Currently basic validation of connector configurable fields is done by each individual connector, for example see [MongoDB validation logic](https://github.com/elastic/connectors-python/blob/main/connectors/sources/mongo.py#L205-L231):

```python
    async def validate_config(self):
        self.configuration.check_valid() # <<<<<<<< THIS LINE

        client = self.client
        configured_database_name = self.configuration["database"]
        configured_collection_name = self.configuration["collection"]

        existing_database_names = await client.list_database_names()

        self._logger.debug(f"Existing databases: {existing_database_names}")

        if configured_database_name not in existing_database_names:
            raise ConfigurableFieldValueError(
                f"Database ({configured_database_name}) does not exist. Existing databases: {', '.join(existing_database_names)}"
            )

        ...
```

### Proposed Solution

Framework should call this method on the connectors instead - the call should be pulled up from all the connectors and the BaseDataSource `validate_config` function should raise `NotImplementedError` instead of doing basic validation.

PR #1450, opened 2023-08-15
Make validate_config of data source call its super method first

wangch079:
In the `validate_config` of data source classes, it will explicitly call `self.configuration.check_valid()`, which is already implemented in `validate_config` of `BaseDataSource`. This PR makes the changes to data source classes to call `validate-config` of parent class first.

#### Pre-Review Checklist


artem-shelkovnikov:
This PR is good, but also we need to make sure data sources don't need to even call this method and validation is done outside by the framework.

Are you planning to address this, @wangch079 ?

artem-shelkovnikov (approved):

wangch079:
>but also we need to make sure data sources don't need to even call this method and validation is done outside by the framework.

Sorry, don't quite understand this statement. could you help elaborate? The config is now validated by the framework:

https://github.com/elastic/connectors-python/blob/34063b2d2597c67d2d46be8d344a9d8f85e9e2a1/connectors/sync_job_runner.py#L126

artem-shelkovnikov:
> Sorry, don't quite understand this statement. could you help elaborate? The config is now validated by the framework:
> 
> [connectors-python/connectors/sync_job_runner.py](https://github.com/elastic/connectors-python/blob/34063b2d2597c67d2d46be8d344a9d8f85e9e2a1/connectors/sync_job_runner.py#L126)
> 
> Line 126 in [34063b2](https://github.com/elastic/connectors-python/commit/34063b2d2597c67d2d46be8d344a9d8f85e9e2a1)
> 
>  await self.data_provider.validate_config()

It would be nice to make our DataSource classes to not rely on calling `super().something`, in this case `super.validate_config()` - it's error-prone and just weird design-wise.

Instead, would be nice that the class that facilitates the work of data sources does it by itself:

```python
configuration = Configuration(connector_record['configurable_fields']) # for example

# Validate configuration fields
await configuration.validate()
data_provider = create_connector(service_type, configuration)

# Now ask connector to do its special validation
await self.data_provider.validate_config()
```

artem-shelkovnikov:
Found this issue that I filed that explains it: https://github.com/elastic/connectors-python/issues/1194

wangch079:
Noted. Let's address it in that issue.

PR #1451, opened 2023-08-15
Download ContentDocuments and extract binary content

navarone-feekery:
## Closes https://github.com/elastic/enterprise-search-team/issues/5442

Download `ContentDocument` SObjects that have a connection to existing SObjects.

A single `ContentDocument` can be connected to many SObjects through `ContentDocumentLinks`. This means duplicates are common. To prevent downloading or ingesting duplicates, we hold on to all `ContentDocument` ids until ingestion of all other SObjects is complete. We then ensure there are no duplicates among the files to download. 

When mapping the new doc for a `ContentDocument`, we retain a list of related SObject ids in the field `linked_ids`. These SObject ids are also the ids used in Elastic docs. (Happy to change the `linked_ids` field name).

`ContentDocuments` also have multiple `ContentVersions`. We ingest only the latest version through the relation `LatestPublishedVersion`.

All `ContentDocuments` for a SObject are fetched with a join query alongside the original SObjects, rather than in a dedicated `ContentDocument` query. This reduces the number of queries being made.

- [ContentDocument reference docs](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_contentdocument.htm)
- [ContentVersion reference docs](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_contentversion.htm)

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -4,20 +4,30 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 """Salesforce source module responsible to fetch documents from Salesforce."""
+import asyncio
 from contextlib import contextmanager
 from functools import cached_property
 from itertools import groupby
 from threading import Lock
 
+import aiofiles
 import aiohttp
+from aiofiles.os import remove
+from aiofiles.tempfile import NamedTemporaryFile
 from aiohttp.client_exceptions import ClientResponseError
 
 from connectors.logger import logger
 from connectors.source import BaseDataSource
-from connectors.utils import CancellableSleeps, retryable
+from connectors.utils import (
+    TIKA_SUPPORTED_FILETYPES,
+    CancellableSleeps,
+    convert_to_b64,
+    retryable,
+)
 
 RETRIES = 3
 RETRY_INTERVAL = 1
+CHUNK_SIZE = 1024
 
 BASE_URL = "https://<domain>.my.salesforce.com"
 API_VERSION = "v58.0"
@@ -33,6 +43,9 @@ RELEVANT_SOBJECTS = [
     "CaseComment",
     "CaseFeed",
     "Contact",
+    "ContentDocument",
+    "ContentDocumentLink",
+    "ContentVersion",
     "EmailMessage",
     "FeedComment",
     "Lead",
@@ -48,6 +61,7 @@ RELEVANT_SOBJECT_FIELDS = [
     "CommentBody",
     "CommentCount",
     "Company",
+    "ContentSize",
     "ConvertedAccountId",
     "ConvertedContactId",
     "ConvertedDate",
@@ -56,6 +70,7 @@ RELEVANT_SOBJECT_FIELDS = [
     "Description",
     "Email",
     "EndDate",
+    "FileExtension",
     "FirstOpenedDate",
     "FromAddress",
     "FromName",
@@ -65,6 +80,7 @@ RELEVANT_SOBJECT_FIELDS = [
     "LastEditById",
     "LastEditDate",
     "LastModifiedById",
+    "LatestPublishedVersionId",
     "LeadSource",
     "LinkUrl",
     "MessageDate",
@@ -83,6 +99,8 @@ RELEVANT_SOBJECT_FIELDS = [
     "Title",
     "ToAddress",
     "Type",
+    "VersionDataUrl",
+    "VersionNumber",
     "Website",
 ]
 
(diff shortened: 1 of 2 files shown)


connectors/sources/salesforce.py
```diff
@@ -137,6 +155,8 @@ def __init__(self, configuration):
         self._queryable_sobjects = None
         self._queryable_sobject_fields = None
         self._sobjects_cache_by_type = None
+        self._supported_file_types = [x[1:] for x in TIKA_SUPPORTED_FILETYPES]
```
artem-shelkovnikov, line 158: Nit: Can probably be inlined

connectors/sources/salesforce.py
```diff
@@ -183,23 +203,35 @@ async def close(self):
...
-            yield case, None
+        async for campaign, content_docs in self.get_campaigns():
+            all_content_docs.extend(content_docs)
+            yield campaign
+
+        async for case, content_docs in self.get_cases():
+            all_content_docs.extend(content_docs)
+            yield case
+
+        all_content_docs = self._combine_duplicate_content_docs(all_content_docs)
```
artem-shelkovnikov, line 232: How big are the objects?

How much is the memory growth when all of that is stored in-memory?

If it's big enough, then we probably need to do it on the fly.

connectors/sources/salesforce.py
```diff
@@ -398,6 +452,43 @@ async def _select_queryable_fields(self, sobject, fields):
...
+        Here we ensure that we don't download any duplicates while retaining links"""
+        grouped = {}
+
+        for content_doc in content_docs:
+            content_doc_id = content_doc["Id"]
+            if content_doc_id in grouped:
+                linked_id = content_doc["linked_sobject_id"]
+                linked_ids = grouped[content_doc_id]["linked_ids"]
+                if linked_id not in linked_ids:
+                    linked_ids.append(linked_id)
```
artem-shelkovnikov, line 480: Does it actually append to `grouped[content_doc_id]["linked_ids"]`?

I'm not 100% familiar with how ref variables work in Python

artem-shelkovnikov (approved):
Looks good, left a couple minor questions

connectors/sources/salesforce.py
navarone-feekery: it does!
Apparently in python `linked_ids` here isn't technically its own variable, its just a reference to that dict key. It threw me for a loop when I first realised.
(I think this is why calling `.copy()` is so important in python)

connectors/sources/salesforce.py
navarone-feekery: I guess we have no idea, a user could conceivably have a massive amount of files.

> we probably need to do it on the fly.

Do you mean check for duplicates as we're downloading files?

connectors/sources/salesforce.py
navarone-feekery: Sorry, what do you mean? (I thought inlined means a single line)

connectors/sources/salesforce.py
artem-shelkovnikov: > I guess we have no idea, a user could conceivably have a massive amount of files.

That's something good to test, when we have a proper test instance.

> Do you mean check for duplicates as we're downloading files?

I mean that right now you do the following: collect ids for objects A, B, C, D...Z, then combine and de-dupe. To optimise it (if duplicates are actually happening often, of course) you can do it on the fly, imagine the following class:

```python
class IdDeduplicator:
    def __init__(self):
        self.ids = {}

    def add(obj):
        if self._has_id(obj):
            self._append_to_existing(obj)
        else:
            self._append_non_existing(obj)

    def get():
        return self.ids
```

So every time an object is added, deduplication happens. We'll trade memory vs CPU here, but it's efficient as long as there are a lot of duplicate objects. I believe there's no need to touch it now, but good to keep this in mind for later testing. Also, something to collect for telemetry if we're not sure about actual numbers of duplicates even after testing.

navarone-feekery: Ah I see! Thanks for the summary. I'm going to open a few improvement issues for this connector so I'll include both of your points here (optimising this and ensuring we get telemetry on it).

connectors/sources/salesforce.py
artem-shelkovnikov: It's similar to this: https://www.jetbrains.com/help/resharper/Refactorings__Inline_Field.html

If you have a field that's used once in one place, you can instead of assigning `[x[1:] for x in TIKA_SUPPORTED_FILETYPES]` to a field, change the place that uses `self._supported_file_types` to use `TIKA_SUPPORTED_FILETYPES` directly with a filter statement.

navarone-feekery: I didn't realise I was only using it once, that makes sense. Thanks!

PR #1456, opened 2023-08-15
reformat config yaml

seanstory:
follow-up from https://github.com/elastic/connectors-python/pull/1404#issuecomment-1679150103, just reformats commented yaml

Note that now the comments should not be separated from the configs they describe, but configs have two lines of whitespace between them.

Also, configs are prefixed with a single `#` but comments are prefixed with `## `, which means that you can highlight the whole file and have your IDE "uncomment" it, and you still get valid yaml and meaningful comments.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -22,7 +22,6 @@
 #connector_id: null
 #
 #
-#
 ##  The service type of the connector.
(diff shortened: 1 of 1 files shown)


navarone-feekery (approved):
Nice improvement!

Issue #1468, opened 2023-08-16
Impossible to run connectors with an api_key


artem-shelkovnikov:
### To Reproduce
1. Pull latest connector framework
2. Set up a connector in Kibana and use its API KEY for `elasticsearch.api_key` setting. Don't fill-in `elasticsearch.username` or `elasticsearch.password`.
3. Run connector
4. See error: "KeyError: "You can't use basic auth ('username' and 'password') and 'api_key' at the same time in config.yml""

## Expected behavior
Connector runs well with `api_key` specified.

llermaly:
Nice. I had this exact same issue when trying to fix the warning and use the new config.yml syntax.

PR #1470, opened 2023-08-16
default to api_key, fallback on username/password

seanstory:
## Closes https://github.com/elastic/connectors-python/issues/1468

When I moved default configs to the code instead of the YAML, I added logic that merges the default config dict with the config-file provided config dict. This introduced an edge case where the default provided username/password auth for Elasticsearch, but the user provided API key auth for Elasticsearch, and we raised an error.

However - there's really no reason for this to be an error state. If both are configured, we'll just default to using API Key. That's the one we want folks using.

Alternatively, we _could_ put effort into trying to reconcile the two configurations, and discount certain default configs if other explicit configs are provided. But I'm not sure this is worth the effort.

#### Pre-Review Checklist


wangch079 (commented):
did the same thing here: https://github.com/elastic/connectors-python/pull/1262 😆 

Worth adding some tests

connectors/es/client.py
```diff
@@ -50,17 +50,13 @@ def __init__(self, config):
         }
         logger.debug(f"Host is {self.host}")
 
-        if "username" in config:
```
timgrein, line 53: We could also add a warning log that we found both in the config and use the `api_key`, so someone could trace this behavior during support/debugging, if needed.

timgrein (commented):
++ on adding a test here (and maybe a warning log), this error looks familiar as @wangch079 said

connectors/es/client.py
seanstory: warn log is a good idea

artem-shelkovnikov (approved):
Good stuff!

artem-shelkovnikov (approved):

PR #1477, opened 2023-08-17
Use BoundedSemaphore instead of asyncio.Event to control the max concurrency

wangch079:
### What's the change:

This PR uses `asyncio.BoundedSemaphore` instead of `asyncio.Event` to control the max concurrency of `ConcurrentTasks`

### Why the change:

1. `BoundedSemaphore` is a better fit for this use case.
2. There's a [bug](https://github.com/elastic/connectors-python/pull/1477#issuecomment-1690159544) in the original implementation, and you can verify it by adding and running this test in [test_utils.py](https://github.com/elastic/connectors-python/blob/main/tests/test_utils.py):
```python
@pytest.mark.asyncio
async def test_concurrent_runner():
    async def coroutine(i, sleep_time):
        await asyncio.sleep(sleep_time)
        print(f"Complete task {i}")

    runner = ConcurrentTasks(max_concurrency=1)
    # push a short-running task a and wait for its completion
    await runner.put(functools.partial(coroutine, "a", 0.1))
    print()
    print(f"task a was pushed into pool at {datetime.now()}")
    await runner.join()

    # push a long-running task b and a short-running task c, c was supposed to block until b is done
    await runner.put(functools.partial(coroutine, "b", 5))
    print(f"task b was pushed into pool at {datetime.now()}")
    await runner.put(functools.partial(coroutine, "c", 0.1))
    print(f"task c was pushed into pool at {datetime.now()}")
    await runner.join()
```

The output in `main` branch:
```shell
task a was pushed into pool at 2023-08-24 21:40:34.742396
Complete task a
task b was pushed into pool at 2023-08-24 21:40:34.844073
task c was pushed into pool at 2023-08-24 21:40:34.844197 # task c was pushed into the pool at the same time with task b, but it's supposed to block and wait for the completion of task b (5 seconds)
Complete task c # task c was completed first
Complete task b
```

If you run the same test in this PR, you get:
```shell
task a was pushed into pool at 2023-08-24 21:48:27.405617
Complete task a
task b was pushed into pool at 2023-08-24 21:48:27.506222
Complete task b
task c was pushed into pool at 2023-08-24 21:48:32.508061
Complete task c
```

#### Pre-Review Checklist

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -258,7 +258,7 @@ async def test_concurrent_runner():
 
 
 @pytest.mark.asyncio
-async def test_concurrent_runner_fails():
+async def test_concurrent_runner_fails(patch_logger):
     results = []
 
     def _results_callback(result):
@@ -276,6 +276,7 @@ async def test_concurrent_runner_fails():
 
     await runner.join()
     assert 5 not in results
+    patch_logger.assert_present("I FAILED")
 
 
 @pytest.mark.asyncio
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -258,6 +258,30 @@ async def test_concurrent_runner():
 
 
 @pytest.mark.asyncio
+async def test_concurrent_runner_canceled():
+    results = []
+    tasks = []
+
+    def _results_callback(result):
+        results.append(result)
+
+    async def coroutine(i):
+        await asyncio.sleep(1)
+        return i
+
+    runner = ConcurrentTasks(max_concurrency=10, results_callback=_results_callback)
+    for i in range(10):
+        tasks.append(await runner.put(functools.partial(coroutine, i)))
+
+    runner.cancel()
+    await runner.join()
+    assert len(tasks) == 10
+    for task in tasks:
+        assert task.cancelled()
+    assert len(results) == 0
+
+
+@pytest.mark.asyncio
 async def test_concurrent_runner_fails():
     results = []
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -258,7 +258,7 @@ async def test_concurrent_runner():
 
 
 @pytest.mark.asyncio
-async def test_concurrent_runner_fails(patch_logger):
+async def test_concurrent_runner_fails():
     results = []
 
     def _results_callback(result):
@@ -276,7 +276,6 @@ async def test_concurrent_runner_fails(patch_logger):
 
     await runner.join()
     assert 5 not in results
-    patch_logger.assert_present("I FAILED")
 
 
 @pytest.mark.asyncio
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -282,6 +282,32 @@ async def test_concurrent_runner_canceled():
 
 
 @pytest.mark.asyncio
+async def test_concurrent_runner_canceled_with_waiting_task():
+    results = []
+
+    def _results_callback(result):
+        results.append(result)
+
+    async def coroutine(i, sleep_time):
+        await asyncio.sleep(sleep_time)
+        return i
+
+    runner = ConcurrentTasks(max_concurrency=10, results_callback=_results_callback)
+    for i in range(10):
+        await runner.put(functools.partial(coroutine, i, 1))  # long-running task
+
+    # create a task to put a waiting task so that it won't block
+    asyncio.create_task(runner.put(functools.partial(coroutine, 100, 0.1)))
+    runner.cancel()
+    # wait for the first 10 tasks to be canceled
+    await runner.join()
+    # wait for the 11th task to complete
+    await runner.join()
+    assert len(results) == 1
+    assert results[0] == 100
+
+
+@pytest.mark.asyncio
 async def test_concurrent_runner_fails():
     results = []
 
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -359,7 +359,7 @@ class ConcurrentTasks:
     def __init__(self, max_concurrency=5, results_callback=None):
         self.tasks = []
         self.results_callback = results_callback
-        self._sem = asyncio.Semaphore(max_concurrency)
+        self._sem = asyncio.BoundedSemaphore(max_concurrency)
 
     def __len__(self):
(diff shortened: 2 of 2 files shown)


timgrein (commented):
Could we run into a potential problem, if we call `cancel` and `_callback` tasks get cancelled before releasing the semaphore? Should we "reset" the semaphore, when calling `cancel`?

artem-shelkovnikov (approved):
+1 to Tim's question.

Can we also add more tests for this class? This class is one of the core classes that make the framework do its thing, so it should be covered 200% + for every bug there should be a regression test. Otherwise we can end up in subtle breakages that are hard to trace and troubleshoot.

wangch079:
> Could we run into a potential problem, if we call `cancel` and `_callback` tasks get cancelled before releasing the semaphore?

The `_callback` function is executed as a done callback, and cancel is one of [the `done` scenarios](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.done):
>A Task is _done_ when the wrapped coroutine either returned a value, raised an exception, or the Task was cancelled.

So `_callback` itself won't be canceled.

Actually there's a more strict Semaphore class [BoundedSemaphore](https://docs.python.org/3/library/asyncio-sync.html#asyncio.BoundedSemaphore), which will

>raises a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) in [release()](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore.release) if it increases the internal counter above the initial value.

Initially, I think this is not needed as I don't see how this can happen. But I think it's not a bad idea to be more defensive.

>Can we also add more tests for this class?

Agree. Will do so.

wangch079:
Update it to use `BoundedSemaphore` class, and added tests to cover `cancel` function. 

Let me know if you think there are more test cases need to be covered, cc @artem-shelkovnikov @timgrein

timgrein:
> The _callback function is executed as a done callback, and cancel is one of [the done scenarios](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.done)

I see, that makes sense. Could you maybe add a short docs line referencing the link you've provided. For me this behavior is not too obvious without having this explicit information. Rest looks good 👍

tests/test_utils.py
```diff
@@ -233,6 +233,30 @@ async def coroutine(i):
     assert results == list(range(10))
 
 
+@pytest.mark.asyncio
```
timgrein, line 236: Could we've additional tests, testing the behavior, when we submit more tasks than `max_concurrency` allows? Just that we don't introduce regressions, when someone changes something in `ConcurrentTasks` later on. 

`test_concurrent_runner_high_concurrency` addresses this partially, but I guess there are some more edge cases (max_concurrency being 10 -> submit 11 long running tasks -> cancel -> what happens with the 11th task?). Just that we've the intention how this works in a test for later reference. (or we expand the class doc string to describe these scenarios).

timgrein (approved):
One additional comment for expanding the class doc string or adding more tests for edge cases, but otherwise LGTM.

tests/test_utils.py
wangch079: >but I guess there are some more edge cases (max_concurrency being 10 -> submit 11 long running tasks -> cancel -> what happens with the 11th task?).

In this case, the first 10 jobs will be canclled, and the 11th task will be completed. I will add this test.

wangch079: This is more complicated than I thought. I'm trying to get the test working or prove that `ConcurrentTasks` doesn't work as expected.

wangch079: Updated: https://github.com/elastic/connectors-python/pull/1477/commits/d8ef6705005fc53b00fb32d4bcc48cf9346bbf60

wangch079:
> One additional question, you said: `Semaphore is a better fit for this use case.`
> 
> Why is it a better use? Can we still work with older implementation of Event?

When you want to control the total number of items, Semaphore is the first idea comes to my mind.

There also might be a bug in the old `Event` solution (Although I didn't verify it):

Assuming the `max_concurrency` is 5, we first add a short-running task, which will casue the `Event` to be set when it's done:
https://github.com/elastic/connectors-python/blob/c0366e0cf6eee0c11d47a8fff903cb8924961a6b/connectors/utils.py#L345

Then we add 5 long-running tasks, and now the task pool is full, and the `Event` is still set.

When we add the 6th task, it executes the below code:
https://github.com/elastic/connectors-python/blob/c0366e0cf6eee0c11d47a8fff903cb8924961a6b/connectors/utils.py#L364-L368

`await self._task_over.wait()` will return `True` immediately, because:
>If the event is set, return `True` immediately. Otherwise block until another task calls [set()](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event.set). ([Ref](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event.wait))

wangch079:
Hi @elastic/ingestion-team , updated the [PR description](https://github.com/elastic/connectors-python/pull/1477#issue-1854473400)

artem-shelkovnikov:
Btw original design of ConcurrentTasks was meant to _not reuse_ the class for adding new tasks after a join. So there could be more things going the wrong way when we start re-using it. Might it be worth to create a different abstraction for it, like a coroutine pool? Maybe there's already a built-in abstraction for it?

artem-shelkovnikov (approved):

artem-shelkovnikov:
While I approved the PR, I feel like this whole topic (concurrency changes) should be a work in the open as much as possible - with gists, PoCs and vast description on how and why it works.

I know that we're moving to serverless and there are many topic flowing just around this, so I suggest approaching it with "let's agree on design first, then implement".

artem-shelkovnikov:
Let's maybe collect requirements first and discuss @wangch079? I feel like this and consequent PRs are not going to be reviewed well/approved if we as a team (all of us) do not know what's going on.

wangch079:
Hi @artem-shelkovnikov , totally agree. I'm not in a hurry as we are at the beginning of the new release cycle. Let's find some time next week to discuss further on this topic.

artem-shelkovnikov (approved):

Issue #1473, opened 2023-08-16
[Google Drive] Sync crashes when can't process a file


llermaly:
## Bug Description
When the connector can not sync a single file (e.g: no permissions) the entire sync crashes.

Error example:

```
Forbidden

Content:
{'code': 403,
 'errors': [{'domain': 'global',
             'message': 'The user does not have sufficient permissions for '
                        'this file.',
             'reason': 'insufficientFilePermissions'}],
 'message': 'The user does not have sufficient permissions for this file.'}

Request URL:
https://www.googleapis.com/drive/v3/files/1FvYMLbOnRuT4BwCLgjZQHHGJFDzHbiYt/permissions?fields=permissions%28type%2CemailAddress%2Cdomain%29%2CnextPageToken&supportsAllDrives=True&pageSize=100.
```

### To Reproduce
1. Create index
2. Give incorrect permissions to one file
3. Run sync

## Expected behavior
If something happens with a single file this file should be skipped and the sync must continue.

## Screenshots
![image](https://github.com/elastic/connectors-python/assets/8575569/d459dda0-1d27-4590-b091-f13a43948b76)

Issue #1474, opened 2023-08-16
[Google Drive] Service account needs Editor permissions for sync


llermaly:
### Problem Description

The Service account needs editor permissions to be able to sync files. I don't think we should be asking for editor permissions if we are only going to read the files.

This is problematic because if we are using a group then we must give Editor permissions to the entire group.

This is related to #1473 , if you give "View" permission to the service account the sync will crash.

### Proposed Solution
Only require "View" permissions to sync the files.

### Additional Context
![CleanShot 2023-08-16 at 18 49 11](https://github.com/elastic/connectors-python/assets/8575569/a51767a1-f9cc-4bfb-850b-4be5468b236f)

jedrazb:
Update: For files on shared drives we are able to list all permissions using the [permissions.list](https://developers.google.com/drive/api/reference/rest/v3/permissions/list) endpoint - that works only for shared drives.

For other "drives" e.g. `my drive`, `shared with me` - permissions are returned as a part of response in [files.list](https://developers.google.com/drive/api/reference/rest/v3/files/list) but the permissions are only populated if the requesting user (here SA) can further share the file (search for phrase `Output only. The full list of permissions for the file. This is only available if the requesting user can share the file. Not populated for items in shared drives.` in [REST Resource: files](https://developers.google.com/drive/api/reference/rest/v3/files))

So as for me, we are correctly handling both cases (shared and non-shared drives), and the viewer permission for non-shared drives doesn't return permissions as expected (since view-only permissions probably doesn't grant sharing rights).

llermaly:
@jedrazb thanks! 

How this behave for domain-wide strategy using admin account? can see everything? or still needs to be shared with the admin?

seanstory:
@jedrazb is this issue still relevant?

jedrazb:
No longer relevant. Google Drive connector now supports using a domain-wide delegation for syncing data. For anyone reading this issue refer to: https://www.elastic.co/guide/en/enterprise-search/current/connectors-google-drive.html

PR #1478, opened 2023-08-17
[Google Drive] Handle failures when fetching file ACLs

jedrazb:
## Closes https://github.com/elastic/connectors-python/issues/1473

In certain scenarios the `permissions.list` endpoint will result in `403 insufficientFilePermissions` error. 

For example: someone else (e.g. from outside org) shares a file with you, you then go to drive UI to the "shared with me" tab and share this with the SA with view-only access. This chain of permissions is insufficient according to google to fetch ACL for that file. (Surprisingly the editor access allows to fetch that). See full scenario described in: https://github.com/elastic/connectors-python/issues/1474

Before, the whole sync would fail if we failed to fetch permissions even for a single file. In this PR I'm adding a logic to handle this gracefully and log problematic filenames and corresponding exceptions. 

Also, for the sake of completeness I added `https://www.googleapis.com/auth/drive.metadata.readonly` scope to drive client - to achieve parity with scopes used by older Workplace Search connector and this connector. 

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/google_drive.py
+++ b/connectors/sources/google_drive.py
@@ -1110,9 +1110,11 @@ class GoogleDriveDataSource(BaseDataSource):
                 except HTTPError as exception:
                     # Gracefully handle scenario when the service account does not
                     # have permission to fetch ACL for a file.
-                    self._logger.warning(
-                        f"Unable to fetch permission list for the file {file_name}. Exception: {exception}."
-                    )
+                    exception_log_msg = f"Unable to fetch permission list for the file {file_name}. Exception: {exception}."
+                    if exception.res.status_code == 403:
+                        self._logger.warning(exception_log_msg)
+                    else:
+                        self._logger.error(exception_log_msg)
 
             file_document[ACCESS_CONTROL] = self._process_permissions(permissions)
 
--- a/connectors/sources/google_drive.py
+++ b/connectors/sources/google_drive.py
@@ -242,7 +242,10 @@ class GoogleDriveClient(GoogleAPIClient):
             json_credentials=json_credentials,
             api_name="drive",
             api_version="v3",
-            scopes=["https://www.googleapis.com/auth/drive.readonly"],
+            scopes=[
+                "https://www.googleapis.com/auth/drive.readonly",
+                "https://www.googleapis.com/auth/drive.metadata.readonly",
+            ],
         )
 
     async def ping(self):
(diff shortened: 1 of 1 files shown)


connectors/sources/google_drive.py
```diff
@@ -1100,9 +1103,16 @@ async def prepare_file(self, file, paths):
...
             permissions = file.get("permissions", [])
             if not permissions:
-                permissions = await self._get_permissions_on_shared_drive(
-                    file_id=file_id
-                )
+                try:
+                    permissions = await self._get_permissions_on_shared_drive(
+                        file_id=file_id
+                    )
+                except HTTPError as exception:
```
vidok, line 1110: not blocking: I wonder if some of the HTTPError exceptions can be fixed by retrying the request or reauthenticate it?

jedrazb: > I wonder if some of the HTTPError exceptions can be fixed by retrying the request
Nope, we already have a re-try logic at a lower level. This `try/except` method is only supposed to gracefully handle failure scenarios after all retries are exhausted. The underlying issue is with scenario where user doesn't grant sufficient permissions to a file, enable dls, and so permissions fetching will always fail. 

> if you handle only a specific case like 403 error then it's worth adding an extra `try:` where you log that specific error.
Yes, exactly I only handle 403s in this clause. I can't really add extra logging at the route of the exception as this is handled by a library. Not sure if I understand what you mean here, can you elaborate. For context the logged `exception` will print the status code, and a detailed error message.

vidok: > I can't really add extra logging at the root of the exception as this is handled by the aiogoogle library. Not sure if I understand what you mean here, can you elaborate.

what I mean is: 
```python
   try:
    .....
   except HTTPError as e:
     if e.reg.status == 403:
       self._logger.warning("Unable to fetch permissions due to 403 error. Exception {exception}")
    else:
      self._logger.error("Unable to fetch permissions due exception {exception}")
```

In this case, we can use different log levels or even handle those errors differently if we need. 
"Need" - is the main question here :)

vidok (approved):

connectors/sources/google_drive.py
jedrazb: That's a good point, we might want to make the non-403s more visible to the user. I updated the PR! :)

vidok (approved):

PR #1479, opened 2023-08-17
Add non-blocking try_put method to ConcurrentTasks

wangch079:
### What's the change

This PR adds a `try_put` method to `ConcurrentTasks`. Compared to the `put` method, `try_put` will return a `None` task immediately if the `ConcurrentTasks` are not ready to accept a new task.

### Why the change

The new method will be used in #1865, by both `AccessControlSyncJobExecutionService` and `ContentSyncJobExecutionService`. 

The existing `put` method is blocking, which will prevent the loop of both services from going to the next iteration.

While `try_put` will return immediately if there's no space available, allowing the service to try in the next iteration instead of blocking.

#### Pre-Review Checklist

### DIFF
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -380,11 +380,11 @@ class ConcurrentTasks:
 
         # put a task into pool
         # it will block until the task was put successfully
-        task = task_pool.put(a_coroutine)
+        task = await task_pool.put(coroutine)
 
         # put a task without blocking
         # it will try to put the task, and return None if it can't be put immediately
-        task = task_pool.try_put(a_coroutine)
+        task = task_pool.try_put(coroutine)
 
         # call join to wait for all tasks in pool to complete
         # this is not required to execute the tasks in pool
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -372,6 +372,25 @@ class ConcurrentTasks:
 
     - `max_concurrency`: max concurrent tasks allowed, default: 5
     - `results_callback`: when provided, synchronous function called with the result of each task.
+
+    Examples:
+
+        # create a task pool with the default max concurrency
+        task_pool = ConcurrentTasks()
+
+        # put a task into pool
+        # it will block until the task was put successfully
+        task = task_pool.put(a_coroutine)
+
+        # put a task without blocking
+        # it will try to put the task, and return None if it can't be put immediately
+        task = task_pool.try_put(a_coroutine)
+
+        # call join to wait for all tasks in pool to complete
+        # this is not required to execute the tasks in pool
+        # a task will be automatically scheduled to execute once it's put successfully
+        # call join() only when you need to do something after all tasks in pool complete
+        await task_pool.join()
     """
 
     def __init__(self, max_concurrency=5, results_callback=None):
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -428,6 +428,7 @@ class ConcurrentTasks:
 
         if self._sem.try_acquire():
             return self._add_task(coroutine, result_callback=result_callback)
+        return None
 
     async def join(self):
         """Wait for all tasks to finish."""
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -395,15 +395,7 @@ class ConcurrentTasks:
         if self.results_callback is not None:
             self.results_callback(task.result())
 
-    async def put(self, coroutine, result_callback=None):
-        """Adds a coroutine for immediate execution.
-
-        If the number of running tasks reach `max_concurrency`, this
-        function will block and wait for a free slot.
-
-        If provided, `result_callback` will be called when the task is done.
-        """
-        await self._sem.acquire()
+    def _add_task(self, coroutine, result_callback=None):
         task = asyncio.create_task(coroutine())
         self.tasks.append(task)
         # _callback will be executed when the task is done,
@@ -414,7 +406,18 @@ class ConcurrentTasks:
         )
         return task
 
-    async def try_put(self, coroutine, result_callback=None):
+    async def put(self, coroutine, result_callback=None):
+        """Adds a coroutine for immediate execution.
+
+        If the number of running tasks reach `max_concurrency`, this
+        function will block and wait for a free slot.
+
+        If provided, `result_callback` will be called when the task is done.
+        """
+        await self._sem.acquire()
+        return self._add_task(coroutine, result_callback=result_callback)
+
+    def try_put(self, coroutine, result_callback=None):
         """Tries to add a coroutine for immediate execution.
 
         If the number of running tasks reach `max_concurrency`, this
@@ -423,10 +426,8 @@ class ConcurrentTasks:
         If provided, `result_callback` will be called when the task is done.
         """
 
-        if self._sem.locked():
-            return None
-
-        return await self.put(coroutine, result_callback=result_callback)
+        if self._sem.try_acquire():
+            return self._add_task(coroutine, result_callback=result_callback)
 
     async def join(self):
         """Wait for all tasks to finish."""
--- a/connectors/utils.py
+++ b/connectors/utils.py
@@ -414,6 +414,20 @@ class ConcurrentTasks:
         )
         return task
 
+    async def try_put(self, coroutine, result_callback=None):
+        """Tries to add a coroutine for immediate execution.
+
+        If the number of running tasks reach `max_concurrency`, this
+        function return a None task immediately
+
+        If provided, `result_callback` will be called when the task is done.
+        """
+
+        if self._sem.locked():
+            return None
+
+        return await self.put(coroutine, result_callback=result_callback)
+
     async def join(self):
         """Wait for all tasks to finish."""
         await asyncio.gather(*self.tasks, return_exceptions=True)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -324,7 +324,7 @@ async def test_concurrent_runner_canceled_with_waiting_task():
 
 
 @pytest.mark.asyncio
-async def test_concurrent_runner_fails():
+async def test_concurrent_runner_fails(patch_logger):
     results = []
 
     def _results_callback(result):
@@ -343,6 +343,7 @@ async def test_concurrent_runner_fails():
 
     await runner.join()
     assert 5 not in results
+    patch_logger.assert_present("I FAILED")
 
 
 @pytest.mark.asyncio
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -374,6 +374,27 @@ async def test_concurrent_runner_high_concurrency():
     assert second_results == [3]
 
 
+@pytest.mark.asyncio
+async def test_concurrent_runner_try_put():
+    results = []
+
+    def _results_callback(result):
+        results.append(result)
+
+    async def coroutine(i):
+        await asyncio.sleep(0.1)
+        return i
+
+    runner = ConcurrentTasks(1, results_callback=_results_callback)
+    await runner.put(functools.partial(coroutine, 0))
+    task = await runner.try_put(functools.partial(coroutine, 1))
+
+    await runner.join()
+
+    assert task is None
+    assert 1 not in results
+
+
 @contextlib.contextmanager
 def temp_file(converter):
     if converter == "system":
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -324,7 +324,7 @@ async def test_concurrent_runner_canceled_with_waiting_task():
 
 
 @pytest.mark.asyncio
-async def test_concurrent_runner_fails(patch_logger):
+async def test_concurrent_runner_fails():
     results = []
 
     def _results_callback(result):
@@ -343,7 +343,6 @@ async def test_concurrent_runner_fails(patch_logger):
 
     await runner.join()
     assert 5 not in results
-    patch_logger.assert_present("I FAILED")
 
 
 @pytest.mark.asyncio
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -374,8 +374,18 @@ async def test_concurrent_runner_high_concurrency():
     assert second_results == [3]
 
 
+@pytest.mark.parametrize(
+    "initial_capacity, expected_result",
+    [
+        (0, True),  # when pool is empty
+        (2, True),  # when pool is almost empty
+        (5, True),  # when pool is half full
+        (8, True),  # when pool is almost full
+        (10, False),  # when pool is full
+    ],
+)
 @pytest.mark.asyncio
-async def test_concurrent_runner_try_put():
+async def test_concurrent_runner_try_put(initial_capacity, expected_result):
     results = []
 
     def _results_callback(result):
@@ -385,18 +395,22 @@ async def test_concurrent_runner_try_put():
         await asyncio.sleep(0.1)
         return i
 
-    runner = ConcurrentTasks(1, results_callback=_results_callback)
-    # the first task is put successfully
-    task_0 = await runner.put(functools.partial(coroutine, 0))
-    # the second task is not put because the pool is full
-    task_1 = runner.try_put(functools.partial(coroutine, 1))
+    runner = ConcurrentTasks(10, results_callback=_results_callback)
+    # fill the pool with initial capacity
+    for i in range(initial_capacity):
+        await runner.put(functools.partial(coroutine, i))
+
+    # try to put an additional task
+    task = runner.try_put(functools.partial(coroutine, 100))
 
     await runner.join()
 
-    assert task_0 is not None
-    assert task_1 is None
-    assert 0 in results
-    assert 1 not in results
+    if expected_result:
+        assert task is not None
+        assert 100 in results
+    else:
+        assert task is None
+        assert 100 not in results
 
     # try_put the third task
     task_2 = runner.try_put(functools.partial(coroutine, 2))
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -387,7 +387,7 @@ async def test_concurrent_runner_try_put():
 
     runner = ConcurrentTasks(1, results_callback=_results_callback)
     await runner.put(functools.partial(coroutine, 0))
-    task = await runner.try_put(functools.partial(coroutine, 1))
+    task = runner.try_put(functools.partial(coroutine, 1))
 
     await runner.join()
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -412,14 +412,6 @@ async def test_concurrent_runner_try_put(initial_capacity, expected_result):
         assert task is None
         assert 100 not in results
 
-    # try_put the third task
-    task_2 = runner.try_put(functools.partial(coroutine, 2))
-
-    await runner.join()
-
-    assert task_2 is not None
-    assert 2 in results
-
 
 @contextlib.contextmanager
 def temp_file(converter):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -298,32 +298,6 @@ async def test_concurrent_runner_canceled():
 
 
 @pytest.mark.asyncio
-async def test_concurrent_runner_canceled_with_waiting_task():
-    results = []
-
-    def _results_callback(result):
-        results.append(result)
-
-    async def coroutine(i, sleep_time):
-        await asyncio.sleep(sleep_time)
-        return i
-
-    runner = ConcurrentTasks(max_concurrency=10, results_callback=_results_callback)
-    for i in range(10):
-        await runner.put(functools.partial(coroutine, i, 1))  # long-running task
-
-    # create a task to put a waiting task so that it won't block
-    asyncio.create_task(runner.put(functools.partial(coroutine, 100, 0.1)))
-    runner.cancel()
-    # wait for the first 10 tasks to be canceled
-    await runner.join()
-    # wait for the 11th task to complete
-    await runner.join()
-    assert len(results) == 1
-    assert results[0] == 100
-
-
-@pytest.mark.asyncio
 async def test_concurrent_runner_fails():
     results = []
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -298,6 +298,32 @@ async def test_concurrent_runner_canceled():
 
 
 @pytest.mark.asyncio
+async def test_concurrent_runner_canceled_with_waiting_task():
+    results = []
+
+    def _results_callback(result):
+        results.append(result)
+
+    async def coroutine(i, sleep_time):
+        await asyncio.sleep(sleep_time)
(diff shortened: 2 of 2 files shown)


artem-shelkovnikov (commented):
What's the use-case for it?

I would avoid doing so, cause it complicates the class that is pretty sensitive.

wangch079:
>What's the use-case for it?

The [JobExecutionService](https://github.com/elastic/connectors-python/blob/main/connectors/services/job_execution.py) executes both content sync and permissions sync in the same loop (with 2 task pools), if the pool is full, it will block (likely happen for content sync), which will also block permissions sync. With the non-blocking put, we can try to put and fail fast, and the service can move to the next iteration.

connectors/utils.py
```diff
@@ -361,18 +360,28 @@ async def put(self, coroutine, result_callback=None):
...
+    async def try_put(self, coroutine, result_callback=None):
+        """Tries to add a coroutine for immediate execution.
+
+        If the number of running tasks reach `max_concurrency`, this
+        function return a None task immediately
+
+        If provided, `result_callback` will be called when the task is done.
+        """
+
+        if self._sem.locked():
```
timgrein, line 380: Should we add a debug log line here? Could be interesting for support as there's a potential action attached to it (increasing/decreasing `max_concurrency` config). We could also add a debug log line further up in the call stack and make it specific for the sync job (or whoever the caller is).

timgrein (dismissed):
LGTM, you'll probably address using this method in a separate PR?

wangch079:
>LGTM, you'll probably address using this method in a separate PR?

Yes, as mentioned here: https://github.com/elastic/connectors-python/pull/1479#issuecomment-1685780873

But I will need to merge https://github.com/elastic/connectors-python/pull/1477 first before I can merge this

connectors/utils.py
wangch079: > Could be interesting for support as there's a potential action attached to it (increasing/decreasing `max_concurrency` config).

This won't take effect until you restart the service.

If `self._sem.locked()` returns `True`, it means the task queue is full: https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore.locked

artem-shelkovnikov:
Reposting my comment from a different PR:

> Hmmm, would it be worth to have 2 task queues instead - "normal" priority and "high" priority with different thresholds/limits. I think putting these 2 concepts into same class might backfire.

> I honestly like to look at it from perspective of priority queues, changes in this PR make me worried (also because I'm still not ultra comfortable with some async/task code, it's a very complex thing)

connectors/utils.py
timgrein: > This won't take effect until you restart the service.
It still could be something you want to notice as you can do something about it (restarting and increasing `max_concurrency`), right? Just that it's more obvious, if the service starts to become slower than expected due to reaching the `max_concurrency` limit. Or do we log it somewhere else in a different form?

wangch079: >Or do we log it somewhere else in a different form?

I'd like to log it in the caller of this method, as `ConcurrentTasks` doesn't have the context.

E.g. We can log in `JobExecutionService`:
> The Sync job 123 is skipped and will retry in 30 seconds because the task pool is full.

vidok:
@wangch079 I'm trying to understand the problem you are solving so let me try to summarise it (please correct me if I'm wrong): 

The current concurrency implementation is limited by a single iteration of this [while-loop](https://github.com/elastic/connectors-python/blob/main/connectors/services/job_execution.py#L163). In other words, a new sync job can't start if the connector is already executing other sync jobs. So it's not only about access sync jobs but also content sync jobs. What you are trying to do is to unblock the locking mechanism and move it outside the loop (I guess this is the final plan).

If I look at the bigger picture, it's clear that Content Syncs and Access Control Syncs (and any other sync types) should be separated. Technically speaking, there should be different `JobExecutionService` abstractions for each job type (this brings me back to 8.6 where I introduced Consumer/Producer and wanted to have different consumers for each sync job type). This can give a huge improvement in:
* Better concurrency control (can be different for different types of jobs)
* Code complexity reduction (base JobExecution class will be small and clean)
* Easy to write tests for jobs because async logic can be easily mocked
* Ability to define different autoscaling configurations for different JobExecutors in a k8s environment (Serverless)

This separation + moving queues out of the while-loop is a very important step. 

Happy to chat more about this.

wangch079:
> I'm trying to understand the problem you are solving so let me try to summarise it

Hi @vidok , I also had a discussion with @artem-shelkovnikov , and I realized that the team needs more context to understand and review the PR. I have updated the issue description in [Improve the concurrency of the Job execution service](https://github.com/elastic/enterprise-search-team/issues/5193)

>I guess this is the final plan

That's right. The PR [Use BoundedSemaphore instead of asyncio.Event to control the max concurrency](https://github.com/elastic/connectors-python/pull/1477) and this PR are for preparation of the real change (and they are also easier to review). That's why you see `try_put` method is added but never used. But @artem-shelkovnikov suggested it was hard to review this PR without the usage of the new method, so I will probably add the changes to `JobExecutionService` in this PR.

>If I look at the bigger picture, it's clear that Content Syncs and Access Control Syncs (and any other sync types) should be separated.

You are absolutely right, but that requires more changes and may be worth a separate tech-debt issue. My plan is that `JobExecutionService` will work in a fire-and-forget (i.e. non-blocking) mechanism. With this, I don't feel different consumers for each sync job type is absolutely necessary.

>Better concurrency control (can be different for different types of jobs)

We have different task pools for content and access control sync jobs:
https://github.com/elastic/connectors-python/blob/13d923ebee1426d93d61625b21d79932c0d4f215/connectors/services/job_execution.py#L165-L170

>Code complexity reduction (base JobExecution class will be small and clean)

Personally I'd favor composition over inheritance:

https://github.com/elastic/connectors-python/blob/13d923ebee1426d93d61625b21d79932c0d4f215/connectors/services/job_execution.py#L130-L137

>Easy to write tests for jobs because async logic can be easily mocked

I don't see the difference. Can you elaborate more?

>Ability to define different autoscaling configurations for different JobExecutors in a k8s environment (Serverless)

TBH I have no idea how it works. In k8s, isn't the smallest unit to scale the connector service? Interested to hear more.

wangch079:
Hi @artem-shelkovnikov , I still think it makes sense to break changes to PRs as meaningfully small as possible. I added the usage of `try_put` in a new PR [Improve the job execution concurrency](https://github.com/elastic/connectors-python/pull/1534) (Pointing base branch to this PR).

As discussed, I will need more time to test the edge cases in the above PR, so it's better to merge this PR first if it's good to go.

Let me know if you still have concerns, and I can merge that PR so all the changes will apear in this PR.

vidok:
> My plan is that `JobExecutionService` will work in a fire-and-forget (i.e. non-blocking) mechanism. With this, I don't feel different consumers for each sync job type is absolutely necessary.

Is my assumption correct that with `fire-and-forget` the while-loop will keep iterating (querying ES) even if the queue is full? If yes, this is an irrational use of resources. Also, `fire-and-forget` will look like 

> My plan is that JobExecutionService will work in a fire-and-forget (i.e. non-blocking) mechanism. 

I would make this logic more explicit, more verbose, and easy to read than just `fire-and-forget`: 

```python
def put(self, coroutine, result_callback=None):

  if self._sem.locked():
    raise ConcurrencyError('Max concurrency has reached. Can't executed a job')
  
  return await self.put(coroutine, result_callback=result_callback)
```

and then a usage:
```python
   try: 
     puts(coroutine, result_callback=result_callback)
   except ConcurrencyError as e:
     logger.warning(e.message)
   except:
     logger.error(e.message)      
```

> Easy to write tests for jobs because async logic can be easily mocked
> I don't see the difference. Can you elaborate more?

Sure, if you have good abstractions (either inheritance or composition) and the connection between them, mocking becomes easier because you don't need to mock a lot of different methods, constants, or objects. You can easily isolate the class/method you are testing. Usually, a code smells when you need to mock a lot of things to isolate its behavior. 

> TBH I have no idea how it works. In k8s, isn't the smallest unit to scale the connector service? Interested to hear more.

I understand that many folks from our team don't have experience with Kubernetes, so this is why I bring up these topics. We need to keep this in mind as serverless is approaching we can have a difficult time making connectors suitable for k8s environment. Happy to chat anytime.

wangch079:
> I would make this logic more explicit, more verbose, and easy to read than just `fire-and-forget`:

I think this is similar to my implementation in PR https://github.com/elastic/connectors-python/pull/1534. Instead of raise and catch. I checked the returned task:

https://github.com/elastic/connectors-python/blob/e7967a509e8c221cba4de69ecc5a4cfa98410d72/connectors/utils.py#L383-L384

https://github.com/elastic/connectors-python/blob/e7967a509e8c221cba4de69ecc5a4cfa98410d72/connectors/services/job_execution.py#L85-L88

I think @artem-shelkovnikov also had concerns, I will schedule a meeting next week to discuss this further.

connectors/utils.py
```diff
@@ -396,6 +396,20 @@ async def put(self, coroutine, result_callback=None):
...
+            return None
+
+        return await self.put(coroutine, result_callback=result_callback)
```
jedrazb, line 411: IMO there is a potential for a race condition here, if the semaphore becomes locked between the check `if self._sem.locked():` and the `return await self.put(..)`. 

I think it should still be safe because it essentially means that we will be less efficient at handling tasks in job execution service. Is there any potential for unwanted impact in this race condition scenario?

artem-shelkovnikov: +++ to addressing this. I think it's possible that once `await self.put` is called, execution is given to another coroutine that can in turn call `try_put`.

Since semaphore is set inside `self.put`, it won't yet be set in this case for both calls and both will go through.

wangch079: First of all, I want to make it clear that `ConcurrentTasks` is not designed to be thread-safe. [asyncio.BoundedSemaphore](https://docs.python.org/3/library/asyncio-sync.html#boundedsemaphore) is not either.

But I think your concern is valid. I assumed it would pass the control back to the main loop when it encountered the first `await` in `put` method (i.e. `await self._sem.acquire()`), but I coudn't find the evidence.

I will make some changes to handle this case.

wangch079: I tested with the following script:
```python
import asyncio
from datetime import datetime

async def put(sem, i):
  print(f"{datetime.now()} - putting {i}")
  # use asyncio.sleep(0) to release control
  await asyncio.sleep(0)
  await sem.acquire()
  print(f"{datetime.now()} - Successfully put {i}")

  # sleep for 2 second and release
  await asyncio.sleep(2)
  print(f"{datetime.now()} - Successfully done {i}")
  sem.release()

async def try_put(sem, i):
  # wait for other tasks
  await asyncio.sleep(0.5)
  print(f"{datetime.now()} - try_put {i}")
  if sem.locked():
    print(f"{datetime.now()} - try_put {i}, couldn't put the task, return")
    return

  await put(sem, i)


async def main():
  task_pool_limit = 5
  num_of_tasks = task_pool_limit + 1
  sem = asyncio.BoundedSemaphore(task_pool_limit)
  tasks = []
  for i in range(num_of_tasks):
    tasks.append(asyncio.create_task(try_put(sem, i+1)))

  await asyncio.gather(*tasks)

asyncio.run(main())

```
and here is the output:
```
2023-11-09 17:54:40.309128 - try_put 1
2023-11-09 17:54:40.309180 - putting 1
2023-11-09 17:54:40.309209 - try_put 2
2023-11-09 17:54:40.309221 - putting 2
2023-11-09 17:54:40.309237 - try_put 3
2023-11-09 17:54:40.309247 - putting 3
2023-11-09 17:54:40.309266 - try_put 4
2023-11-09 17:54:40.309276 - putting 4
2023-11-09 17:54:40.309291 - try_put 5
2023-11-09 17:54:40.309300 - putting 5
2023-11-09 17:54:40.309314 - try_put 6
2023-11-09 17:54:40.309323 - putting 6 --> trying to put task 6, blocking instead of returning
2023-11-09 17:54:40.309365 - Successfully put 1
2023-11-09 17:54:40.309396 - Successfully put 2
2023-11-09 17:54:40.309419 - Successfully put 3
2023-11-09 17:54:40.309440 - Successfully put 4
2023-11-09 17:54:40.309460 - Successfully put 5
2023-11-09 17:54:42.310391 - Successfully done 1
2023-11-09 17:54:42.310486 - Successfully done 2
2023-11-09 17:54:42.310520 - Successfully done 3
2023-11-09 17:54:42.310548 - Successfully done 4
2023-11-09 17:54:42.310574 - Successfully done 5
2023-11-09 17:54:42.310632 - Successfully put 6 --> task 6 is successfully put after blocking for 2 seconds
2023-11-09 17:54:44.311775 - Successfully done 6
```

But if I comment out `await asyncio.sleep(0)`, it works as expected:
```
2023-11-09 17:56:52.230397 - try_put 1
2023-11-09 17:56:52.230523 - putting 1
2023-11-09 17:56:52.230562 - Successfully put 1
2023-11-09 17:56:52.230647 - try_put 2
2023-11-09 17:56:52.230688 - putting 2
2023-11-09 17:56:52.230710 - Successfully put 2
2023-11-09 17:56:52.230768 - try_put 3
2023-11-09 17:56:52.230809 - putting 3
2023-11-09 17:56:52.230823 - Successfully put 3
2023-11-09 17:56:52.230857 - try_put 4
2023-11-09 17:56:52.230873 - putting 4
2023-11-09 17:56:52.230886 - Successfully put 4
2023-11-09 17:56:52.230917 - try_put 5
2023-11-09 17:56:52.230932 - putting 5
2023-11-09 17:56:52.230945 - Successfully put 5
2023-11-09 17:56:52.230973 - try_put 6
2023-11-09 17:56:52.230987 - try_put 6, couldn't put the task, return --> trying to put task 6, return immediately
2023-11-09 17:56:54.231842 - Successfully done 1
2023-11-09 17:56:54.231925 - Successfully done 2
2023-11-09 17:56:54.231960 - Successfully done 3
2023-11-09 17:56:54.231992 - Successfully done 4
2023-11-09 17:56:54.232019 - Successfully done 5
```

So it seems the current implementation does not release control to the loop unless you do so explicitly. 

I don't find an easy way to prevent this from happening. Considered using [asyncio.Lock](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock), so that there can only be one party executing `put`/`try_put`, so that the context switching won't happen between the lock is acquired and released. But this will not work when a user calls `put` and is blocked, then another user calls `try_put` and will be blocked again because it can't acquire the lock.

wangch079: I think I found an easy workaround. WDYT if we extend the `asyncio.BoundedSemaphore` class and introduce a `try_acquire` method:

```python
def try_acquire(self):
  if self.locked():
    return False

  self._value -= 1
  return True
```

The `locked` implementation is [here](https://github.com/python/cpython/blob/97c4c06d0d235aad00e5b6b10af8b8d68c889b9b/Lib/asyncio/locks.py#L358-L361). It will return `True` if either there's no space or there are other tasks blocking to acquire it.

The code
```
self._value -= 1
return True
```
is borrowed from the implementation of [acquire](https://github.com/python/cpython/blob/97c4c06d0d235aad00e5b6b10af8b8d68c889b9b/Lib/asyncio/locks.py#L363). The logic for `try_acquire` is simple because we don't need to add it to the waiting pool when there's no space available immediately.

artem-shelkovnikov: Intuitively it looks okay, but I need to think a bit more about it, thanks for sharing the thought process!

wangch079: I implemented the above change in a separate PR: #1889

artem-shelkovnikov: I've tried an approach with keeping the `Event` working and it seemed to be a little cleaner:

```python
+    async def try_put(self, coroutine, result_callback=None):
+        """Adds a coroutine for immediate execution.
+
+        If the number of running tasks reach `max_concurrency`, this
+        function will block and wait for a free slot.
+
+        If provided, `result_callback` will be called when the task is done.
+        """
+        # If self.tasks has reached its max size, we wait for one task to finish
+        if len(self.tasks) >= self.max_concurrency:
+            try:
+                await asyncio.wait_for(self._task_over.wait(), 1) # Wait for 1 second and if we can't add a task error out
+                self._task_over.clear()
+            except asyncio.TimeoutError:
+                return None
+        # rearm
+        task = asyncio.create_task(coroutine())
+        self.tasks.append(task)
+        task.add_done_callback(
+            functools.partial(self._callback, result_callback=result_callback)
+        )
+        return task
```

artem-shelkovnikov: Do you think it can work?

wangch079: Replied in anther PR: https://github.com/elastic/connectors/pull/1889#discussion_r1395312172, to make discussions in a single place.

connectors/utils.py
```diff
@@ -414,6 +406,29 @@ async def put(self, coroutine, result_callback=None):
...
+        """Tries to add a coroutine for immediate execution.
+
+        If the number of running tasks reach `max_concurrency`, this
+        function return a None task immediately
+
+        If provided, `result_callback` will be called when the task is done.
+        """
+
+        if self._sem.try_acquire():
+            return self._add_task(coroutine, result_callback=result_callback)
```
artem-shelkovnikov, line 430: I believe you want to indicate to the caller, that put failed?

E.g. raise an error or return True/False?

artem-shelkovnikov: Ah, I see it returns `None` in case nothing is added.

tests/test_utils.py
```diff
@@ -374,6 +374,27 @@ def _second_callback(result):
     assert second_results == [3]
 
 
+@pytest.mark.asyncio
+async def test_concurrent_runner_try_put():
```
artem-shelkovnikov, line 378: I see another good test here that verifies, that `try_put` will succeed after the task actually finishes.

wangch079: Yes, as we discussed yesterday, I will add more tests

connectors/utils.py
wangch079: Good point. Will do

wangch079: > Ah, I see it returns `None` in case nothing is added.

I made it explicitly return `None` in commit https://github.com/elastic/connectors/pull/1479/commits/bf54ba1364a05f74d5755778bb0da503c4c34551

I also think that `put` and `try_put` should not return the task, which should be managed internally by `ConcurrentTasks`. For all usages of `put` and `try_put`, the returned task is not used, but it's used in tests. I think it's safe to make `put` and `try_put` to return `True/False`, but I'd like to address it in a separate PR

tests/test_utils.py
wangch079: I updated the test here: https://github.com/elastic/connectors/pull/1479/commits/8d033e040007d5ada52a6e9661fca170175e3555

I think the scenario you mentioned yesterday is not practical:
When the pool is full and another task is waiting by calling `put`, I don't think another task can jump the queue by calling `try_put`, because `try_put` will check `self.locked()` first, which will [take those waiting tasks into consideration](https://github.com/python/cpython/blob/9e56eedd018e1a4681789e634016cbb7699dcb8a/Lib/asyncio/locks.py#L358-L361).

wangch079:
Hi @artem-shelkovnikov , as discussed yesterday, added some examples in the comments: https://github.com/elastic/connectors/pull/1479/commits/151145b5ba14a08547302177905f5f6ad4c410cb

tests/test_utils.py
```diff
@@ -374,6 +374,39 @@ def _second_callback(result):
...
+    results = []
+
+    def _results_callback(result):
+        results.append(result)
+
+    async def coroutine(i):
+        await asyncio.sleep(0.1)
+        return i
+
+    runner = ConcurrentTasks(1, results_callback=_results_callback)
```
navarone-feekery, line 388: This is an absolute nit, but I could see some benefit in having a test with a higher ConcurrentTasks limit. Right now there's no test to confirm that `try_put` will work if the pool is 1/2 full, for example.

navarone-feekery (approved):
LGTM. I see a lot of conversation here and they appear to be mostly resolved. I can't think of anything else that should block merging.

connectors/utils.py
```diff
@@ -372,6 +372,25 @@ class ConcurrentTasks:
...
     - `results_callback`: when provided, synchronous function called with the result of each task.
+
+    Examples:
+
+        # create a task pool with the default max concurrency
+        task_pool = ConcurrentTasks()
+
+        # put a task into pool
+        # it will block until the task was put successfully
+        task = task_pool.put(a_coroutine)
```
artem-shelkovnikov, line 383:
```suggestion
        task = await task_pool.put(a_coroutine)
```

tests/test_utils.py
wangch079: Updated: https://github.com/elastic/connectors/pull/1479/commits/63675a9ed825ad8def6abd2077376ed24b936d26

connectors/utils.py
wangch079: Updated: https://github.com/elastic/connectors/pull/1479/commits/24de66a61886b4d05be294eaf4bfc063ce7f2f38

connectors/utils.py
```diff
@@ -372,6 +372,25 @@ class ConcurrentTasks:
...
+
+        # put a task without blocking
+        # it will try to put the task, and return None if it can't be put immediately
+        task = task_pool.try_put(a_coroutine)
+
+        # call join to wait for all tasks in pool to complete
+        # this is not required to execute the tasks in pool
+        # a task will be automatically scheduled to execute once it's put successfully
+        # call join() only when you need to do something after all tasks in pool complete
```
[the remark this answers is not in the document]
wangch079: What kind of weird results do you expect? 

Calling `join` will just wait for all tasks (at the moment) to be completed before returning.

Assuming there are 5 tasks in the pool, and you call `join`, it will wait for the 5 tasks to complete. Before it returns, if you add the 6th task, the 6th task should remain in the pool.

Let me see if I can add a test for this case.

tests/test_utils.py
wangch079: I think I accidentally forgot to remove those code: https://github.com/elastic/connectors/pull/1479/commits/0e539b01de8fdb9cf1bcf6f74567a70e9e0d60c4

connectors/utils.py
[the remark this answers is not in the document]
artem-shelkovnikov: Yeah exactly this case - adding 5 tasks, joining, then adding another one and expecting all 6 to finish. We had this bug and it was actually a part of implementation/tests at some point, as far as I remember.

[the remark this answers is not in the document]
wangch079: Hi @artem-shelkovnikov , added the test case here: https://github.com/elastic/connectors/pull/1479/commits/e2e0197e6974004b9603d15456c8638e1f43dae0

artem-shelkovnikov (approved):

PR #1485, opened 2023-08-17
[Microsoft SQL] Advanced Sync rules 

parth-elastic:
## Closes https://github.com/elastic/connectors-python/issues/1484

1. This PR introduces advanced filtering rules for MS-SQL by queries to filter out document based on that
2. Added documentation with several examples illustrating the syntax, how a basic use case works
4. Included tooltip on the configurable fields that are ignored when advanced sync rules are used

#### Pre-Review Checklist

## Release Note

For existing Microsoft SQL connectors, the next sync after the upgrade will delete all the existing documents and reindex them with different document IDs.


artem-shelkovnikov (commented):
Intuitively it looks like it's better to allow specify query per each table synced. Do you think it makes sense @danajuratoni ?

connectors/sources/mssql.py
```diff
@@ -419,55 +563,88 @@ async def fetch_documents(self, table):
...
+            docs_generator = (
+                self._yield_docs_custom_query(tables, query)
+                if query is not None
+                else self._yield_all_docs_from_tables(tables[0])
+            )
+            async for doc in docs_generator:
+                yield doc
+        except (InternalClientError, ProgrammingError) as exception:
+            self._logger.warning(
+                f"Something went wrong while fetching document for table {', '.join(tables)}. Error: {exception}"
```
wangch079, line 575: Might want to log `query` as well. Or maybe make it two methods:
1. fetch_documents_from_table
2. fetch_documents_from_query

connectors/sources/mssql.py
```diff
@@ -209,22 +288,26 @@ async def get_table_row_count(self, table):
...
-                        schema=self.schema,
-                        table=table,
-                    ),
-                ),
-                fetch_size=self.fetch_size,
-                retry_count=self.retry_count,
+    async def get_table_primary_key(self, tables):
+        primary_keys = []
+        for table in tables:
+            primary_keys.extend(
```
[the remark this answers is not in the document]
parth-elastic: Here at [L586](https://github.com/elastic/connectors-python/pull/1485/files#diff-e24c08f31e763759f680b48f1a3bf27650b49fcefa87acdedc94fe9bfbfaaf23R586) we are checking if more than one table have same primary key.

connectors/sources/mssql.py
parth-elastic: Sure, I have created two methods

connectors/sources/mssql.py
```diff
@@ -419,55 +548,105 @@ async def fetch_documents(self, table):
...
+                    for table in tables
+                ],
+            )
+        )
+        async for row in self.yield_rows_for_query(
+            primary_key_columns=primary_key_columns, tables=tables, query=query
+        ):
+            last_update_time = (
+                max(last_update_times) if len(last_update_times) else iso_utc()
+            )
```
wangch079, line 606: This can be done outside the `async for` generator

connectors/sources/mssql.py
```diff
@@ -409,7 +496,49 @@ async def ping(self):
...
+        )
+
+        if not set(primary_key_columns) - set(column_names):
+            async for row in streamer:
+                row = dict(zip(column_names, row, strict=True))
+                yield row
+        else:
+            self._logger.warning(
+                f"Skipping query for tables {', '.join(tables)} as primary key column name is not present in query."
+            )
```
wangch079, line 539: Do you want to log `query` as well?

connectors/sources/generic_database.py
```diff
@@ -47,15 +47,33 @@ def is_wildcard(tables):
...
+        primary key values: 1, 42
+        table1_table2_1_42
+    """
+
+    if not isinstance(tables, list):
+        tables = [tables]
+
+    return (
+        f"{'_'.join(sorted(tables))}_"
+        f"{'_'.join([str(pk_value) for pk in primary_key_columns if (pk_value := row.get(pk)) is not None])}"
```
wangch079, line 73: The original implementation is not `join`. E.g. there are 2 columns `foo` and `bar` forming the primary key. Assuming the value for `foo` is `a`, and `bar` is `b`. Previously the id will be `database_schema_table_a_b_`, now it will be `database_schema_table_a_b`

parth-elastic: Yes, I think in last _ is not required in ID. So, I have change ID in both the document with sync rules and without sync rules. So it is uniform in both the cases.

connectors/sources/mssql.py
parth-elastic: Yes, added query in the logger 👍

wangch079 (approved):
Please add a release note as discussed [here](https://elastic.slack.com/archives/C02AV5C8ZDE/p1693310281295159?thread_ts=1692285570.129979&cid=C02AV5C8ZDE)

PR #1501, opened 2023-08-22
[Outlook] Implement Ping + Indexing

moxarth-elastic:
## Related to https://github.com/elastic/connectors-py/issues/1421

Ping and Indexing is covered in this PR.

Outlook objects covered in the connector:
- Mails (inbox, sent, archive and junk)
- Contacts
- Tasks
- Calendar events

Total test coverage is 91%.

#### Pre-Review Checklist

- [x] New external service dependencies added.
  - `ldap3` is used to fetch AD users for exchange server.


connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+        return Connection(
+            server=server,
+            user=self.user,
+            password=self.password,
+            client_strategy=SAFE_SYNC,
+            auto_bind=True,  # pyright: ignore
+        )
+
+    async def get_users(self):
+        warnings.filterwarnings("ignore")
```
jedrazb, line 213: Do we really need to disable the warnings here? 

As a side effect we will disable warning for the whole framework (I think) because we never revert back to default level:
```
# Revert back to the default warning behavior
warnings.filterwarnings("default")
```

But if possible I would get rid of this statement and address the root cause of warnings.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+
+        status, _, response, _ = self._create_connection(
+            server=Server(host=self.ad_server)
+        ).search(
+            search_query,
+            SEARCH_FILTER_FOR_NORMAL_USERS,
+            attributes=["mail"],
+        )
+
+        status, _, response_for_admin, _ = self._create_connection(
```
jedrazb, line 225: We use overwrite `status` variable here. The variable is declared twice and it's used for detecting errors. IMO we should have separate variables or add a check after normal users search. 

Also are there other scenarios where we fail? E.g. can status be 404 or sth else that indicates an error?

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+            return
+
+        self._logger.debug(f"Downloading {attachment_name}")
+
+        document = {
+            "_id": attachment.attachment_id.id,
+            "_timestamp": ews_format_to_datetime(
+                datetime=attachment.last_modified_time, timezone=timezone
+            ),
+            "_attachment": get_base64_value(attachment.content),
```
jedrazb, line 747: Could `get_base64_value` block the event loop for large attachments (not sure if we are dealing just with email binary content or actual mail attachments)

connectors/sources/outlook.py
moxarth-elastic: Warning logs are there in case the SSL is disabled, so we're ignoring that warnings. If the ssl is enabled, warnings won't be there (for SSL). Other than that i am not sure which warnings are there so let's keep a "default" behaviour.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+
+    def __init__(self, ad_server, domain, exchange_server, user, password, ssl_enabled):
+        self.ad_server = ad_server
+        self.domain = domain
+        self.exchange_server = exchange_server
+        self.user = user
+        self.password = password
+        self.ssl_enabled = ssl_enabled
+
+    def _create_connection(self, server):
```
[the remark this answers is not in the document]
moxarth-elastic: IG it's worth to make it cached 👍

connectors/sources/outlook.py
moxarth-elastic: Ah I see! need to rename `status`. 

> Also are there other scenarios where we fail? E.g. can status be 404 or sth else that indicates an error?

No, `status` basically returns something that can be a boolean, dict, etc. so there is a chance of 0 and 1, either we get a value  or not.

❗ Here, `status` is not referring to the response status code.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+    async def get_docs(self, filtering=None):
+        """Executes the logic to fetch outlook objects in async manner
+
+        Args:
+            filtering (Filtering): Object of class Filtering
+
+        Yields:
+            dictionary: dictionary containing meta-data of the files.
+        """
+        async for account in self.client._get_client.get_user_accounts():
```
wangch079, line 887: Why do we need to call `get_user_accounts` via `_get_client`?

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+        await self.client.ping()
+        self._logger.info("Successfully connected to Outlook")
+
+    async def _consumer(self):
+        """Async generator to process entries of the queue
+
+        Yields:
+            dictionary: Documents from Outlook.
+        """
+        while self.tasks > 0:
```
wangch079, line 871: is it possible that the consumer is faster and has processed all items in the queue, but fetchers are not done yet?

wangch079 (commented):
I see you defined `ExchangeUsers` and `Office365Users`. They basically both serve as clients. Why are they using different connection methods? Can they be merged into one class `OutlookClient`?

connectors/sources/outlook.py
moxarth-elastic: Yes, that needs to be fixed.

moxarth-elastic:
> I see you defined `ExchangeUsers` and `Office365Users`. They basically both serve as clients. Why are they using different connection methods? Can they be merged into one class `OutlookClient`?

Both the products use different authentication mechanisms to fetch users. `ExchangeUsers` uses `ldap3` whereas `Office365Users` uses graph apis, so to separate the concerns we kept different classes.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+
+            credentials = Credentials(
+                username=self.user,
+                password=self.password,
+            )
+            configuration = Configuration(
+                credentials=credentials,
+                server=self.exchange_server,
+                retry_policy=FaultTolerance(max_wait=120),
+            )
```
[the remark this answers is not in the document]
moxarth-elastic: Yes, I think they can be outside of the loop.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+        await self._get_client.get_users()
+
+    async def get_mails(self, account):
+        for mail_type in MAIL_TYPES:
+            if "archive" in mail_type["folder"]:
+                # If 'Archive' folder is not present, skipping the iteration
+                try:
+                    folder_object = (
+                        account.root / "Top of Information Store" / "Archive"
+                    )
```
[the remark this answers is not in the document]
moxarth-elastic: This is a pathlib notation - https://ecederstrand.github.io/exchangelib/#contacts

connectors/sources/outlook.py
```diff
@@ -0,0 +1,916 @@
...
+            "ssl_enabled": {
+                "depends_on": [{"field": "data_source", "value": OUTLOOK_SERVER}],
+                "display": "toggle",
+                "label": "Enable SSL",
+                "order": 10,
+                "type": "bool",
+                "value": False,
+            },
+            "ssl_ca": {
+                "depends_on": [{"field": "ssl_enabled", "value": True}],
```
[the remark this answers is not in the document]
moxarth-elastic: No, that's not happening. I've checked this by switching data_source from server to cloud and nothing went wrong.

connectors/sources/outlook.py
moxarth-elastic: I think the method name - `_get_client` is confusing you. Here, `_get_client` returns an instance of `ExchangeUsers` or `Office365Users` and calling `get_user_accounts` through it. Let me rename it to something like `_get_user_instance`.

connectors/sources/outlook.py
moxarth-elastic: To overcome that case, we're first incrementing a task by 1 and once that task is completed then only we're putting an `END SIGNAL` to the queue. 

On the other hand in consumer, we're checking tasks count as well as the END SIGNAL so it won't be possible that the consumer has completed its job first and fetchers are not yet done.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,944 @@
...
+            user=self.user,
+            password=self.password,
+            client_strategy=SAFE_SYNC,
+            auto_bind=True,  # pyright: ignore
+        )
+
+    async def get_users(self):
+        warnings.filterwarnings("default")
+        ldap_domain_name_list = ["DC=" + domain for domain in self.domain.split(".")]
+        search_query = ",".join(map(str, ldap_domain_name_list))
```
wangch079, line 233: Why do you need `map` here?

connectors/sources/outlook.py
```diff
@@ -0,0 +1,944 @@
...
+            SEARCH_FILTER_FOR_ADMIN,
+            attributes=["mail"],
+        )
+
+        if not (has_value_for_normal_users or has_value_for_admin_users):
+            raise UsersFetchFailed(
+                "Error while fetching users from Exchange Active Directory."
+            )
+
+        response.extend(response_for_admin)
```
wangch079, line 257: I didn't check the documentation, just want to make sure:

1. The `search` operation will not raise exceptions but always return something in the first return value (e.g. `has_value_for_normal_users`) to indicate the search result.
2. The code proceeds as long as either of  `has_value_for_normal_users` and `has_value_for_admin_users` is `True`. Make sure `response` is not `None` if `has_value_for_normal_users` is `False`.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,944 @@
...
+        for user in response:
```
wangch079, line 258: If you just want to yield users, you can simply do so without `extend`
```python
for user in response:
    yield user

for user in response_for_admin:
    yield user
```

connectors/sources/outlook.py
```diff
@@ -0,0 +1,944 @@
...
+            case _:
+                raise
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+        skipped_exceptions=UnauthorizedException,
+    )
+    async def _fetch_token(self):
```
wangch079, line 323: Can the token be cached?

connectors/sources/outlook.py
```diff
@@ -0,0 +1,944 @@
...
+                },
+            ) as response:
+                json_response = await response.json()
+                users_mails = [
+                    user.get("mail")
+                    for user in json_response.get("value")
+                    if user.get("mail")
+                ]
+                for user in users_mails:
+                    yield user
```
wangch079, line 361: This can be simplified to:
```suggestion
                for user in json_response.get("value", []):
                    if "mail" in user:
                        yield user.get("mail")
```

But I think the methods in `Client` class should be general. If you name the method name `get_users`, you should simply yield `user`, and let the caller handle it as appropriate.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,944 @@
...
+
+            for mail in folder_object.all().only(*MAIL_FIELDS):
+                yield mail, mail_type
+
+    async def get_calendars(self, account):
+        for calendar in account.calendar.all().only(*CALENDAR_FIELDS):
+            yield calendar
+
+    async def get_child_calendars(self, account):
+        for child_calendar in account.calendar.children:
```
wangch079, line 576: I didn't do any tests. Just want to make sure do you need to call something like `.all()` to make it iterate.

connectors/sources/outlook.py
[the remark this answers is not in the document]
wangch079: When you switched `data_source` from server to cloud, did it reset `ssl_enabled` to `False`? If `ssl_enabled` is still `True`, I expect `ssl_ca`wills till present

connectors/sources/outlook.py
moxarth-elastic: Oh Yes! I think map is not required. Thanks!

connectors/sources/outlook.py
moxarth-elastic: > The search operation will not raise exceptions but always return something in the first return value (e.g. has_value_for_normal_users) to indicate the search result.

Yes, it will not raise an exception instead return a status to indicate the result

> The code proceeds as long as either of has_value_for_normal_users and has_value_for_admin_users is True. Make sure response is not None if has_value_for_normal_users is False.

Sure

connectors/sources/outlook.py
moxarth-elastic: it's used once so i did not make it cached

connectors/sources/outlook.py
moxarth-elastic: Agreed 👍

connectors/sources/outlook.py
moxarth-elastic: No, this is another property of calendar object so there is no need of `.all()` to make it iterate.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,992 @@
...
+        search_query = ",".join(ldap_domain_name_list)
+
+        for user in self._fetch_normal_users(search_query=search_query):
+            yield user
+
+        for user in self._fetch_admin_users(search_query=search_query):
+            yield user
+
+    async def get_user_accounts(self):
+        global global_dns_name
```
wangch079, line 300: Is this newly added? Why you need a global variable here?

connectors/sources/outlook.py
```diff
@@ -0,0 +1,992 @@
...
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def get_users(self):
+        try:
+            access_token = await self._fetch_token()
+            async with self._get_session.get(
```
wangch079, line 392: Didn't notice this initially. It's not paginated?

connectors/sources/outlook.py
```diff
@@ -0,0 +1,992 @@
...
+                    "Authorization": f"Bearer {access_token}",
+                    "Content-Type": "application/json",
+                },
+            ) as response:
+                yield await response.json()
+        except Exception:
+            raise
+
+    async def get_user_accounts(self):
+        async for users in self.get_users():
```
wangch079, line 404: It seems `get_users` is not a generator, instead, it only yields a json.

wangch079 (commented):
Outlook Cloud and Server are using different authentication, different clients. Do you think it makes sense to have two separate connectors (Outlook Cloud and Outlook Server)? cc @danajuratoni

connectors/sources/outlook.py
moxarth-elastic: Yes, this is newly added for SSL stuffs.

moxarth-elastic:
> Outlook Cloud and Server are using different authentication, different clients. Do you think it makes sense to have two separate connectors (Outlook Cloud and Outlook Server)? cc @danajuratoni

IMO, just having different authentication methods for server and cloud, we should not create two connector. Also, client methods for fetching outlook document are similar for both server and cloud so i think we should go with the single connector.

We've also followed the same architecture in other connectors like Jira, Confluence.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,999 @@
...
+            return Office365Users(
+                client_id=self.configuration["client_id"],
+                client_secret=self.configuration["client_secret"],
+                tenant_id=self.configuration["tenant_id"],
+            )
+
+        self.ssl_enabled = self.configuration["ssl_enabled"]
+        self.certificate = self.configuration["ssl_ca"]
+
+        global global_cert
```
jedrazb, line 589: Is the global variable necessary here? I feel like we can just pass the cerficate to  `ExchangeUsers` class

jedrazb (changes requested):

connectors/sources/outlook.py
```diff
@@ -0,0 +1,999 @@
...
+class NotFound(Exception):
+    pass
+
+
+class SSLFailed(Exception):
+    pass
+
+
+global_dns_name = ""
+global_cert = ""
```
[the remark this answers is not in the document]
moxarth-elastic: Yes, i knew it but if you look at the usage of these global variable, they are used in `RootCAAdapter` and that class is directly used as below
```python
BaseProtocol.HTTP_ADAPTER_CLS = (
    RootCAAdapter if self.ssl_enabled else NoVerifyHTTPAdapter
)
``` 
So, we can't initialise a class here and pass the value that's why i made them global and access it.

connectors/sources/outlook.py
moxarth-elastic: Yes, we can also do that and in that case the global variable will be set in ExchangeUsers.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,999 @@
...
+
+
+class RootCAAdapter(requests.adapters.HTTPAdapter):
+    """Class to verify SSL Certificate for Exchange Servers"""
+
+    def cert_verify(self, conn, url, ssl_certificate_file, cert):
+        try:
+            with NamedTemporaryFile(mode="w", delete=False, suffix=".cer") as file:
+                file.write(global_cert)
```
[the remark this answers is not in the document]
jedrazb: Hmm, I see. From what I understand we only store `global_cert` as a global variable to later write it to a file in `cert_verify()`. Should we perhaps write the certificate to a file when instantiating the Outlook client and avoid storing the certificate twice?

Also I don't fully understand this part of code:
```
ssl_certificate_file = {
    global_dns_name: file.name,
}[urlparse(url).hostname]
``` 
The `global_dns_name` will always be the same?  Why do we need to first build a dict and then access the key? Is this logic suppose to check for some condition? I would add some more comments as it's not straightforward what's the purpose of this line.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,1003 @@
...
+
+            for user in response_for_admin:
+                yield user
+        except Exception as e:
+            raise UsersFetchFailed(
+                f"Something went wrong while fetching users. Error: {e}"
+            ) from e
+
+    async def get_users(self):
+        warnings.filterwarnings("default")
```
jedrazb, line 293: This line has no effect, `warnings.filterwarnings("default")` resets warning filtering to default python settings. If we want to filter some specific warnings refer to my comment from yesterday. If we can leave it as it is without warning filtering this line is redundant and can be deleted.

jedrazb (commented):
I replied to a few comment threads from yesterday.

connectors/sources/outlook.py
moxarth-elastic: Yes, i think let's not ignore any kind of warnings.

connectors/sources/outlook.py
```diff
@@ -0,0 +1,1003 @@
...
+
+class SSLFailed(Exception):
+    pass
+
+
+class ManageCertificate:
+    cert_file = "outlook_cert.cer"
+
+    def store_certificate(self, certificate):
+        with open(self.cert_file, "w") as file:
```
jedrazb, line 198: can use to make this non-blick
```
import aiofiles
...
async with aiofiles.open(...)
```

connectors/sources/outlook.py
```diff
@@ -0,0 +1,1003 @@
...
+            file.write(certificate)
+
+    def get_certificate_path(self):
+        return os.path.join(os.getcwd(), self.cert_file)
+
+    def remove_certificate_file(self):
+        if os.path.exists(self.cert_file):
+            os.remove(self.cert_file)
```
jedrazb, line 206: we can use to make this non-blocking
```
from aiofiles.os import remove
await remove(...)
```

jedrazb (commented):
Looks good, just leave a small request to use `aiofiles` where possible when operating on files

connectors/sources/outlook.py
jedrazb: Then the method signatures should be async

connectors/sources/outlook.py
moxarth-elastic: `aiofiles` module does not have `os` attribute, so need to keep this.

jedrazb (approved):
LGTM, let's wait for @wangch079 feedback as well

connectors/sources/outlook.py
```diff
@@ -0,0 +1,1006 @@
...
+                tenant_id=self.configuration["tenant_id"],
+            )
+
+        self.ssl_enabled = self.configuration["ssl_enabled"]
+        self.certificate = self.configuration["ssl_ca"]
+
+        if self.ssl_enabled and self.certificate:
+            ssl_ca = get_pem_format(self.certificate)
+        else:
+            ssl_ca = ""
```
wangch079, line 599: Shouldn't this part be done in `__init__`?

connectors/sources/outlook.py
```diff
@@ -0,0 +1,1006 @@
...
+    async def store_certificate(self, certificate):
+        async with aiofiles.open(self.cert_file, "w") as file:
+            await file.write(certificate)
+
+    def get_certificate_path(self):
+        return os.path.join(os.getcwd(), self.cert_file)
+
+    async def remove_certificate_file(self):
+        if os.path.exists(self.cert_file):
+            await remove(self.cert_file)
```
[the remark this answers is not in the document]
moxarth-elastic: @jedrazb suggested to maintain a class for SSL stuffs to make the code clean.

connectors/sources/outlook.py
moxarth-elastic: This only required when `data_source` is server and so i kept it in this methods because `data_source` bifurcation is happening in this method

moxarth-elastic: Yes, we can also do that 👍

wangch079 (approved):

PR #1504, opened 2023-08-22
[New Connector] Zoom Implementation

### LINKED ISSUE #1503: Zoom Implementation

abhishekjoshi-crest:
#### Pre-Review Checklist


connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+                raise NotFound from exception
+            else:
+                raise
+
+    async def fetch(self, url):
+        try:
+            async with self._get(absolute_url=url) as response:
+                return await response.json()
+        except Exception:
+            self._logger.warning(f"Data for {url} is being skipped.")
```
seanstory, line 169: Should probably log why (check to see if the exception has a message, and at least log the class, maybe debug-log the stack trace)

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+        )
+        async for participants in self.api_client.scroll(url=url):
+            for participant in participants.get("participants", []) or []:
+                yield participant
+
+    async def get_recordings(self, user_id):
+        # Zoom recording does not retrieve data that is more than 4 months old.
+        now = datetime.utcnow()
+        for _ in range(4):
+            past_duration = now + timedelta(days=-30)
```
seanstory, line 256: "duration" usually means a delta. This variable is holding a specific date. Can we rename this variable and its usages to something like `start_date` and rename the `now` variable to be `end_date`?

IMO
```
end_date = start_date
```

is more logical as a loop iterates than:
```
now = past_duration
```

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+            url = APIS["RECORDING"].format(
+                base_url=BASE_URL,
+                user_id=user_id,
+                date_from=format_recording_date(date=past_duration),
+                date_to=format_recording_date(date=now),
+                page_size=MEETING_PAGE_SIZE,
+            )
+            async for recordings in self.api_client.scroll(url=url):
+                for recording in recordings.get("meetings", []) or []:
```
seanstory, line 265: would recordings ever have:
```
{
  "meetings": None
}
```
?

If not,  the `or []` feels redundant.

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+
+        previous_meeting["participants"] = participants
+        return previous_meeting
+
+    async def fetch_recordings(self, user_id):
+        async for recording in self.client.get_recordings(user_id=user_id):
+            yield recording
+
+        async for recording in self.client.get_trash_recordings(user_id=user_id):
+            yield recording
```
seanstory, line 473: Is syncing trashed recordings a requirement? This strikes me as something that customers would probably not want by default. At the very least, I'd think this should be configurable. CC @danajuratoni

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+            self._logger.warning(
+                f"Files with the extension {attachment_extension} are not supported, skipping {attachment_name}."
+            )
+            return False
+
+        if attachment_size > FILE_SIZE_LIMIT:
+            self._logger.warning(
+                f"File size {attachment_size} of file {attachment_name} is larger than {FILE_SIZE_LIMIT} bytes. Discarding file content"
+            )
+            return
```
seanstory, line 396: should this return `False`?

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+
+        await asyncio.to_thread(
+            convert_to_b64,
+            source=temp_filename,
+        )
+        async with aiofiles.open(file=temp_filename, mode="r") as target_file:
+            # base64 on macOS will add a EOL, so we strip() here
+            document["_attachment"] = (await target_file.read()).strip()
+        try:
+            await remove(temp_filename)
```
seanstory, line 422: This needs to be in a `finally` block. See other examples in other sources, like: https://github.com/elastic/connectors-python/blob/main/connectors/sources/s3.py#L195-L231

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+
+    async def get_content(self, doc, timestamp=None, doit=False):
+        attachment_size = doc["file_size"]
+        if not (doit and attachment_size > 0):
+            return
+
+        attachment_name = doc["file_name"]
+
+        attachment_extension = (
+            attachment_name[attachment_name.rfind(".") :]  # noqa
```
seanstory, line 438: Why does this need a `# noqa` comment?

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+    async def fetch_recordings(self, user_id):
+        async for recording in self.client.get_recordings(user_id=user_id):
+            yield recording
+
+        async for recording in self.client.get_trash_recordings(user_id=user_id):
+            yield recording
+
+    async def get_docs(self, filtering=None):
+        async for user in self.client.get_users():
+            yield self._format_doc(doc=user, doc_time=iso_utc()), None
```
seanstory, line 477: are we sure we always want to sync user records? I expect that the most value for a zoom connector comes from the chat and from the transcript. Should customers be able to opt out of syncing users @danajuratoni ?

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+                    doc=chat_message, doc_time=chat_message.get("date_time")
+                ), None
+
+            async for chat_file in self.client.get_chats(
+                user_id=user.get("id"), chat_type="file"
+            ):
+                chat_file["id"] = chat_file.get("file_id")
+                yield self._format_doc(
+                    doc=chat_file, doc_time=chat_file.get("date_time")
+                ), partial(self.get_content, doc=chat_file.copy())
```
seanstory, line 540: file should end with an empty line

tests/sources/test_zoom.py
```diff
@@ -0,0 +1,600 @@
...
+        "_timestamp": "2023-02-09T00:00:00Z",
+    },
+]
+
+# File document
+SAMPLE_FILE = {
+    "next_page_token": None,
+    "messages": [
+        {
+            "file_id": "file1",
```
seanstory, line 277: in practice, are these `file_id` values unique? If we're using these for our `_id` field, we'll want to be sure that similarly named files across different meetings don't overwrite one another.

seanstory (commented):
Few nits, and a few items I'd like Dana's input on. But overall, looks pretty good.

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+    "CHAT": "{base_url}/chat/users/{user_id}/messages?page_size={page_size}&search_key=%20&search_type={chat_type}&from={date_from}&to={date_to}",
+}
+
+
+def format_recording_date(date):
+    return date.strftime("%Y-%m-%d")
+
+
+def format_chat_date(date):
+    return date.strftime("%Y-%m-%dT%H:%M:%SZ")
```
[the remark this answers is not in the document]
abhishekjoshi-crest: Yes it is intended as chat api [Reference](https://developers.zoom.us/docs/api/rest/reference/chat/methods/#operation/getChatMessages:~:text=start%20date%2C%20in-,yyyy%2DMM%2Ddd%27T%27HH%3Amm%3Ass%27Z%27,-format.) expects this format

connectors/sources/zoom.py
abhishekjoshi-crest: Yes faced for multiple objects which returns `None` so kept `or []`

connectors/sources/zoom.py
abhishekjoshi-crest: Missed it 😞

tests/sources/test_zoom.py
```diff
@@ -0,0 +1,600 @@
...
+            "aiohttp.ClientSession.post",
+            return_value=get_mock(mock_response=TokenError),
+        ):
+            with pytest.raises(TokenError):
+                await source.ping()
+
+
+@pytest.mark.asyncio
+@freeze_time("2023-03-09T00:00:00")
+async def test_get_docs():
```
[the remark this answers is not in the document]
abhishekjoshi-crest: Sure adding more granular tests

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+            ):
+                yield self._format_doc(
+                    doc=upcoming_meeting,
+                    doc_time=upcoming_meeting.get("created_at"),
+                ), None
+
+            async for previous_meeting in self.client.get_meetings(
+                user_id=user.get("id"), meeting_type="previous_meetings"
+            ):
+                if self.configuration["fetch_past_meeting_details"]:
```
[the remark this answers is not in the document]
abhishekjoshi-crest: Currently `live_meeting`, `upcoming_meeting` & `previous_meeting` are indexed for all fetched_user and regarding `fetch_past_meeting_details` as per [slack discussion](https://crestdatasystems.slack.com/archives/C02AV5C8ZDE/p1692191817237099?thread_ts=1692005081.034619&cid=C02AV5C8ZDE) it will be fetching other information like [Details](https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/pastMeetingDetails) & [Participants](https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/pastMeetingParticipants) for past meetings as more API calls needed

connectors/sources/zoom.py
```diff
@@ -0,0 +1,540 @@
...
+
+            async for channel in self.client.get_channels(user_id=user.get("id")):
+                yield self._format_doc(doc=channel, doc_time=iso_utc()), None
+
+            async for chat_message in self.client.get_chats(
+                user_id=user.get("id"), chat_type="message"
+            ):
+                yield self._format_doc(
+                    doc=chat_message, doc_time=chat_message.get("date_time")
+                ), None
```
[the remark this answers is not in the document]
abhishekjoshi-crest: Yes it's a good idea to have single document for whole conversation but API fetch chat messages based on `user_id` [Reference](https://developers.zoom.us/docs/api/rest/reference/chat/methods/#operation/getChatMessages) not on `meeting_id`

connectors/sources/zoom.py
danajuratoni: In the future, it makes sense to make this configurable for all connectors with Trash/Recycle Bin function. 
Right now, we should stay consistent with the other connectors e.g. GDrive, Salesforce and Gmail which are ignoring trashed items and ignore trashed items for Zoom too.

connectors/sources/zoom.py
danajuratoni: I assume user records are the meeting's attendees. I see use cases where search users would want to filter / search by attendees. So we should include this information, making it configurable would be a nice flexibility improvement. Are there any performance concerns if we add them by default?

danajuratoni: Participants field references only display names, let's ingest users to also have [better feature similarity with the WS](https://github.com/elastic/enterprise-search-zoom-connector#data-extraction-and-syncing) connector

danajuratoni: that's great @abhishekjoshi-crest, we should ingest the participants info returned by the linked API

tests/sources/test_zoom.py
```diff
@@ -0,0 +1,805 @@
...
+
+@pytest.mark.asyncio
+async def test_validate_config():
+    async with create_source(ZoomDataSource) as source:
+        setup_zoom(source)
+        with mock.patch(
+            "aiohttp.ClientSession.post",
+            return_value=mock_token_response(),
+        ):
+            await source.validate_config()
```
seanstory, line 700: missing an assert

tests/sources/test_zoom.py
```diff
@@ -0,0 +1,805 @@
...
+
+@pytest.mark.asyncio
+async def test_ping_for_successful_connection():
+    async with create_source(ZoomDataSource) as source:
+        setup_zoom(source)
+        with mock.patch(
+            "aiohttp.ClientSession.post",
+            return_value=mock_token_response(),
+        ):
+            await source.ping()
```
seanstory, line 721: missing an assert

abhishekjoshi-crest:
buildkite test this

tests/sources/test_zoom.py
abhishekjoshi-crest: Method doesn't returns anything, asserting with None?

tests/sources/test_zoom.py
```diff
@@ -400,14 +401,16 @@ def get_mock(mock_response):
...
+@asynccontextmanager
+async def create_zoom_source(fetch_past_meeting_details=False):
+    async with create_source(ZoomDataSource) as source:
+        source.configuration.get_field("account_id").value = "123"
+        source.configuration.get_field("client_id").value = "id@123"
+        source.configuration.get_field("client_secret").value = "secret#123"
+        source.configuration.get_field(
+            "fetch_past_meeting_details"
+        ).value = fetch_past_meeting_details
+        yield source
```
navarone-feekery, line 413: I notice this is already merged so this is a bit late, but this method is still creating the source before specifying configuration. For a brief moment you have an empty configuration. It's cleaner to do it like this:

```
@asynccontextmanager
async def create_zoom_source(fetch_past_meeting_details=False):
    async with create_source(
        ZoomDataSource,
        account_id="123",
        client_id="id@123",
        client_secret="secret#123",
        fetch_past_meeting_details=fetch_past_meeting_details
    ) as source:
        yield source
```

PR #1514, opened 2023-08-23
Fix mongodb tests by removing mocking of _get_socket + patching admin ping call

artem-shelkovnikov:
MongoDB connector tests broke with new release of MongoDB client - `motor`C  lib depends on it with `>=4.4,<5` expression.

This PR updates the failing tests to not override `_get_socket` private method that is not defined in the new library version any more.

PR #1522, opened 2023-08-23
lock pymongo version

seanstory:
See https://elastic.slack.com/archives/C01795T48LQ/p1692781505612279

The pymongo lib is pulled in not through our framework requirements.txt, but through the OS Arch requirements. It was pinned already in `arm64.txt`, but was pulled in transitively through `motor` from `arch64.txt` and `x86_64.txt`. Motor does not pin the version of pymongo, which means when installing from scratch, this dep can auto increment. 

At some point, we should look into why we're using different versions of pymongo depending on the architecture, but that's out of scope at the moment, as this PR is about creating stability, not opening a can of worms.

#### Pre-Review Checklist

## Related Pull Requests

CI was breaking and was resolved with https://github.com/elastic/connectors-python/pull/1514. This PR just prevents us from getting in a similar spot in the future.

### DIFF
--- a/requirements/aarch64.txt
+++ b/requirements/aarch64.txt
@@ -3,6 +3,7 @@
 
 aiomysql==0.0.22
 aioboto3==10.3.0
+pymongo==3.13.0
 motor==2.5.1
 smbprotocol==1.9.0
 pymongo[srv]==3.13.0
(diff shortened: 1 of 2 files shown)


wangch079 (approved):
LGTM

PR #1526, opened 2023-08-24
[GMail] Pass subject RCF value to Google Directory client

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/5685

This PR fixes a bug, where the value of the `subject` RCF is only passed to the GMail client and not to the Google Directory client, which leads to a 403 Forbidden error (it worked before, because the `subject` field was explicitly part of the `service_account_credentials` json). In reality we've a separate `subject` RCF and we add the `subject` to the json credentials in the backend.

#### Pre-Review Checklist


connectors/sources/gmail.py
```diff
@@ -184,6 +184,7 @@ def _google_directory_client(self):
         return GoogleDirectoryClient(
             json_credentials=self._service_account_credentials,
             customer_id=self.configuration["customer_id"],
+            subject=self.configuration["subject"],
```
navarone-feekery, line 187: Question: why not add the `subject` to the json credentials when you create the `self._service_account_credentials` cache? Is it only required in specific cases?

timgrein: We need to re-update it for every user for the GMail client due to this issue https://issuetracker.google.com/issues/290567932?pli=1. If we would cache `subject` in `self._service_account_credentials` it wouldn't fetch messages for users different than the initial `subject`. That's also the reason why the GMail client cannot be cached until this is fixed.

navarone-feekery (approved):

PR #1531, opened 2023-08-24
[DLS] Remove must_not exists clause from access control query

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/5539

We don't need this part of the query from a technical standpoint and should remove it therefore to avoid confusion. 

#### Pre-Review Checklist


artem-shelkovnikov (commented):
Why was it there originally?

timgrein:
> Why was it there originally?

It was specified like this in the original implementation document and we've noticed after the implementation, that the original reason wasn't valid anymore. There's a slack thread about this, I can search for it, if you're interested.

artem-shelkovnikov (approved):

PR #1543, opened 2023-08-29
[New Connector] Zoom Functional test

### LINKED ISSUE #1542: Zoom Functional tests

abhishekjoshi-crest:
#### Pre-Review Checklist


tests/sources/test_zoom.py
```diff
@@ -0,0 +1,805 @@
...
+    async_mock = mock.AsyncMock()
+    async_mock.__aenter__ = mock.AsyncMock(
+        return_value=ZoomAsyncMock(data=mock_response)
+    )
+    return async_mock
+
+
+def setup_zoom(source, fetch_past_meeting_details=False):
+    # Set up default config with default values
+    source.configuration.set_field(name="account_id", value="123")
```
navarone-feekery, line 405: We're moving away from using `set_field` in tests as it doesn't update a Field, it creates an entirely new one. This loses all of the configuration data set by the default configuration.

A better way to change configuration is `source.configuration.get_field("foo").value = "bar"`

navarone-feekery: Actually, I notice that you're calling `setup_zoom` for almost every test.

You could streamline this by initialising the `create_source` with a wrapper like `create_zoom_source` and then pass in the configuration changes immediately. This would mean the source won't initialise with empty data.

I've done this in the Salesforce test files, [example](https://github.com/elastic/connectors-python/blob/52e4c25bd85b11029d63bdb941068141710ccf08/tests/sources/test_salesforce.py#L423-L445).

connectors/sources/zoom.py
```diff
@@ -0,0 +1,542 @@
...
+        return ZoomClient(configuration=self.configuration)
+
+    @classmethod
+    def get_default_configuration(cls):
+        return {
+            "account_id": {
+                "label": "Account ID",
+                "order": 1,
+                "type": "str",
+                "value": "",
```
navarone-feekery, line 332: We can remove `value` fields entirely now.
```suggestion
```

navarone-feekery: But leave the `fetch_past_meeting_details` value as it's a bool type, and bool types still need an initial value.

tests/sources/test_zoom.py
abhishekjoshi-crest: Forget to mention this is follow up pr, Suggested changes are made in #1504 and on this [commit](https://github.com/elastic/connectors-python/pull/1504/commits/9de198335c2e01befc718589c387ea33ee8a2379)

seanstory (approved):

tests/sources/test_zoom.py
navarone-feekery: Thanks for addressing it in the other PR. I still had a follow-up nit (but at the end of the day it's just a nit). Nice work!

PR #1544, opened 2023-08-29
Add ftest.Dockerfile that's used for all functional tests

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/3397

There's the same docker image that's used for all functional tests, but copied all over. This PR removes it in favour of a single docker image that's used for all functional tests - `ftest.Dockerfile`.

Old Dockerfile is renamed to `app.Dockerfile`.

What's the benefit that I'm looking for here:

I need to be able to share python code between all docker images that we create for functional tests.

- ftest.Dockerfile is in root of repository -> it can copy the whole repo into the docker image
- This way I can use code from connectors / connector tests inside `fixture.py` files that are used to run functional tests.
- Since copy of repo will be available in the container, current code can be ran for functional tests (that in essence means, that changes to ftests will be properly picked up by CI)

#### Pre-Review Checklist


navarone-feekery (dismissed):
This makes sense, nice 🚀

requirements/ftest.txt
seanstory, line 1: The different ftests actually have different requirements, right? Different libraries? How is that being managed now if you've deleted all their specific requirements files?

Makefile
```diff
@@ -73,7 +71,7 @@ default-config: install
 	bin/elastic-ingest --action config --service-type $(SERVICE_TYPE)
 
 docker-build:
-	docker build -t docker.elastic.co/enterprise-search/elastic-connectors:$(VERSION)-SNAPSHOT .
+	docker build -t docker.elastic.co/enterprise-search/elastic-connectors:$(VERSION)-SNAPSHOT -f app.Dockerfile .
```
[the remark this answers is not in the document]
artem-shelkovnikov: I've checked around and saw different suggestions:

1. Do `Dockerfile.{something}` - `Dockerfile.app`, `Dockerfile.ftest` (see https://stackoverflow.com/questions/62195202/multiple-dockerfiles)
2. Do `{something}.Dockerfile` - `app.Dockerfile`, `ftest.Dockerfile` (see https://stackoverflow.com/questions/27409761/docker-multiple-dockerfiles-in-project) - I did this cause of VSCode syntax highlighting
3. Do `Dockerfile_{something}` - `Dockerfile_app`, `Dockerfile_ftest` (see https://stackoverflow.com/questions/62195202/multiple-dockerfiles)

Overall there does not seem to be a standard way to do so so I decided to be very specific - `app.Dockerfile` looks a bit more verbose to me IMO. Additionally, we provide instructions everywhere about it.

I also feel it would be weird to have `Dockerfile` and `ftest.Dockerfile` together, but that's purely to my taste.

Do you feel we can rename app one to `Dockerfile` and leave the other one with prefix/suffix @seanstory?

requirements/ftest.txt
artem-shelkovnikov: Yup. It's all managed now the same way it's managed for the framework - `ftest.txt` requirements file contains requirements for all functional tests.

When we go to direction of plugins, this needs to change, but for now I just need a good way to share code between ftests while also making it work for branches/PRs. Eventually this common setup for tests can be a shared python package.

Makefile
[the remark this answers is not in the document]
seanstory: > Do you feel we can rename app one to Dockerfile and leave the other one with prefix/suffix @seanstory?

Yes, that's my loosely-held suggestion. I don't have the experience to say if that's the "standard" way to do things, but it feels more clear to me that way which one the customer wants, vs which is "sure, a dockerfile, but not what you're looking for."

requirements/ftest.txt
seanstory: Sneaky bugger, you added this in https://github.com/elastic/connectors-python/pull/1544/commits/a43802f5c98529ec0954a3eba3fae2a9e2ec3eff, after I commented. :P

artem-shelkovnikov: Well, functional tests were broken too, so they would not let me sneak that through anyway :D

Makefile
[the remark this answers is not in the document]
vidok: It's very common to have multiple docker files in projects. I personally got used to `Dockerfile.{something_extra}` e.g. there might be different docker files for different platforms or different environments. 

> it feels more clear to me that way which one the customer wants, vs which is "sure, a dockerfile, but not what you're looking for."

Part of me agrees with this so I think it's okay to leave `Dockerfile` as a main docker file and have an extra one which is called `ftest.Dockerfile`.

vidok (approved):

PR #1545, opened 2023-08-29
Fix slack time range error

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/5684

The slack connector will sync messages of past n days (n is configurable), and the time range was incorrect built which cased the latest messages are not synced. This PR fix the bug.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/slack.py
+++ b/connectors/sources/slack.py
@@ -261,9 +261,8 @@ class SlackDataSource(BaseDataSource):
             yield message, None
 
     async def channels_and_messages(self):
-        delta = timedelta(days=self.n_days_to_fetch)
-        past_unix_timestamp = time.mktime((datetime.utcnow() - delta).timetuple())
-        current_unix_timestamp = time.mktime(datetime.utcnow().timetuple())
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov (dismissed):
Out of curiousity - how wrong were the timestamps?

wangch079:
>Out of curiousity - how wrong were the timestamps?

I believe it depends on your local timezone. I'm at UTC+8 and it only syncs data until `now - 16 hours`

timgrein (commented):
Should we add a test, so this doesn't happen again? Could also be a good candidate for a utility function living inside `utils.py`.

wangch079:
> Should we add a test, so this doesn't happen again? 

Added test in https://github.com/elastic/connectors-python/pull/1545/commits/27cdc60cc1aea765d0dbb37df100a460a04430a7

>Could also be a good candidate for a utility function living inside `utils.py`.

I'd like to move it to `utils.py` if there's another usage.

seanstory (approved):

PR #1546, opened 2023-08-30
[Setup] Fix team name

timgrein:
The team is now called `Search Extract and Transform` and not `Ingestion`.

#### Pre-Review Checklist


setup.py
```diff
@@ -89,7 +89,7 @@ def read_reqs(req_file):
     packages=find_packages(),
     description=("Elastic Search Connectors."),
     long_description=long_description,
-    author="Ingestion Team",
+    author="Search Extract and Transform Team",
     author_email="tarek@ziade.org",
```
navarone-feekery, line 93: I wonder if the `author_email` should be updated at any point.

navarone-feekery (approved):

setup.py
timgrein: Yeah, I had the same question on my mind, wanted to check with PM.

PR #1548, opened 2023-08-30
[Setup] Use team email as author email

timgrein:
This PR adapts the `author_email` in `setup.py` to use the team email.

#### Pre-Review Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,7 @@ setup(
     description=("Elastic Search Connectors."),
     long_description=long_description,
     author="Search Extract and Transform Team",
-    author_email="tarek@ziade.org",
+    author_email="enterprise-search-ingestion-team@elastic.co",
     include_package_data=True,
     zip_safe=False,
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov (approved):

setup.py
```diff
@@ -90,7 +90,7 @@ def read_reqs(req_file):
     description=("Elastic Search Connectors."),
     long_description=long_description,
     author="Search Extract and Transform Team",
-    author_email="tarek@ziade.org",
+    author_email="enterprise-search-ingestion-team@elastic.co",
```
vidok, line 93: should this be `extract-and-transform-team`? I'm not sure if it exists though

timgrein: Yes, this should be renamed. I'll follow-up with @DianaJourdan after her PTO. I double-checked and `enterprise-search-ingestion-team@elastic.co` seems to be the correct address for now. I'll change it as soon as we've a new one.

PR #1541, opened 2023-08-29
Remove value field from connector default RCFs (batch 4)

navarone-feekery:
## Related to https://github.com/elastic/connectors-python/issues/606

Remove `value` field from connector default RCFs, allowing them to be initialised as a type-safe empty value.

- s3
- salesforce
- sharepoint_online
- sharepoint_server
- slack

#### Pre-Review Checklist
- [x] this PR has a meaningful title
- [x] this PR links to all relevant github issues that it fixes or partially addresses
- [x] this PR has a thorough description
- [x] Covered the changes with automated tests
- [x] Tested the changes locally
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)
- [x] Considered corresponding documentation changes
- [x] Contributed any configuration settings changes to the configuration reference

## Related PRs

- https://github.com/elastic/connectors-python/pull/1323
- https://github.com/elastic/connectors-python/pull/761
- https://github.com/elastic/connectors-python/pull/1539
- https://github.com/elastic/connectors-python/pull/1540

PR #1552, opened 2023-08-30
Revert removed boolean `value` fields

navarone-feekery:

When removing not-useful `value` fields from connector configurable fields, I mistakenly removed some bool-type values.
Fields with boolean type should always have an initial value.

## Related PRs

- https://github.com/elastic/connectors-python/pull/761
- https://github.com/elastic/connectors-python/pull/1539
- https://github.com/elastic/connectors-python/pull/1540
- https://github.com/elastic/connectors-python/pull/1541

### DIFF
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -287,6 +287,7 @@ class ConfluenceDataSource(BaseDataSource):
                 "label": "Enable SSL",
                 "order": 8,
                 "type": "bool",
+                "value": False,
             },
             "ssl_ca": {
                 "depends_on": [{"field": "ssl_enabled", "value": True}],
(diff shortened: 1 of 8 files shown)


timgrein (approved):

artem-shelkovnikov (approved):

artem-shelkovnikov:
But why do they need initial value? Is it a technical limitation?

navarone-feekery:
@artem-shelkovnikov there are two reasons to keep booleans:

1. If they are null then the Kibana toggle will assume they are `False`, which is not technically correct
2. Some of the values should "default" to `True` (not any in this PR however). An example is DLS for gmail, and some boolean fields for SPO.

Issue #1521, opened 2023-08-23
Override default basic auth with single connector API key


seanstory:
## Bug Description
The implicit default configuration for the connectors framework is:
```
elasticsearch:
  host: http://localhost:9200
  username: elastic
  password: changeme
  api_key: null
connectors: []
```
and the default explicit configuration is:
```
```

Kibana provides a block to paste when you create a new connector, of:
```
connectors:
  -
    service_type: <service_type>
    connector_id: <id>
    api_key: <api_key>
```

If you paste that block, and only that block, into the explicit config (`config.yml`)

You provide Elasticsearch credentials for the _single connector_ but not the whole framework. The top-level `elasticsearch.username` and `elasticsearch.password` will still be using the defaults.

### To Reproduce
1. Start with a fresh clone
2. create a connector on cloud
3. paste only what kibana tells you to paste
4. get a 401 error because `elastic:changeme` doesn't work for your deployment

## Expected behavior
Since the user has provided an API Key, we should use that API key.
This is not true if there are multiple connectors configured. In that case, an explicit set of top-level credentials will be required and will fall back to the defaults if necessary. 

## Environment
8.10.0-BC.1

## Additional context
[slack thread](https://elastic.slack.com/archives/C01795T48LQ/p1692806214826899?thread_ts=1692789088.115089&cid=C01795T48LQ)

seanstory:
This missed 8.10.0. Removing version target and milestone.

seanstory:
We've never had a customer complain about this, and it's getting quite stale. Going to close it.

PR #1553, opened 2023-08-30
Override default auth with connector api key

seanstory:
## Closes https://github.com/elastic/connectors-python/issues/1521

Ensures that if a single connector is configured, and provides an API key, the connector will use that credential as its top-level credential, over the default of `elastic:changeme`. This does not occur if more than one connector is configured, or if the basic auth for the top-level is non-default.

#### Pre-Review Checklist

## Additional Context
More thorough documentation will be covered in: https://github.com/elastic/enterprise-search-team/issues/5681

## Release Note

Allows the credentials of a single configured connector to be used for the framework as a whole.


wangch079 (commented):
I feel this is not consistent: https://elastic.slack.com/archives/C01795T48LQ/p1692858701197249?thread_ts=1692789088.115089&cid=C01795T48LQ

If we leave config `elasticsearch` as it is, the connector service will behave differently based on configuration of `connectors`:
1. If there is only one connector and it's configured with `api_key`, it will work as expected
2. If there's no connector configured or multiple connectors configured, it will break as there's no elasticsearch credential configured.

Although I'd prefer to simply exit the service if users fail to configure the elasticsearch credentials, I agree that we should make the ingesting process as smooth as possible.

Why not copy the very first API key available in `connectors`?

BTW, it seems `_handle_config_conflicts` is not used anywhere.

seanstory:
> I feel this is not consistent

This was the behavior that @danajuratoni and I landed on after discussing some alternatives. I agree it's not consistent, but of the options we have available, it makes the first-touch experience with our connectors the most painless. We want to optimize for folks who are trying out their very first connector. Once they are successful, we can start introducing extra complexity in configuration, because at that point, they're hopefully invested enough to not be overwhelmed or intimidated.

> If there's no connector configured or multiple connectors configured, it will break as there's no elasticsearch credential configured.

There is, though. The default is `elastic:changeme`. We rely on this for our local dev envs and tests. It's not ideal, but changing that is outside of the scope of a minor bugfix change we can make for 8.10 at this point. It's frustrating to users when we say, "paste these creds, and it'll work!" and they paste those creds, and then it doesn't work, because `changeme` isn't the password for their cloud deployment.

> BTW, it seems _handle_config_conflicts is not used anywhere.

🤦 good catch. Thank you

wangch079:
>but of the options we have available, it makes the first-touch experience with our connectors the most painless.

Do you have any concerns about this suggestion?

>Why not copy the very first API key available in `connectors`?

I feel this is even easier to implement and cover more edge cases.

seanstory:
> Do you have any concerns about this suggestion?

Not enough. The risk/reward ratio is low, IMO.

> Why not copy the very first API key available in connectors?

We are doing that too: https://github.com/elastic/enterprise-search-team/issues/5680

But this will let us keep users from getting hung up if they miss copying that new block, and will tide us over until we can make the UX clearer in kibana to ensure that this step isn't missed/skipped.

wangch079:
> > Why not copy the very first API key available in connectors?
> 
> We are doing that too: [elastic/enterprise-search-team#5680](https://github.com/elastic/enterprise-search-team/issues/5680)

I think you misunderstood. Code explains it better:
```python
elasticsearch = configuration.get("elasticsearch", {})
default_elasticsearch = _default_config().get("elasticsearch")
has_default_auth = (
    elasticsearch.get("username") == default_elasticsearch.get("username")
    and elasticsearch.get("password") == default_elasticsearch.get("password")
    and elasticsearch.get("api_key") == default_elasticsearch.get("api_key")
)

if has_default_auth:
    connectors = configuration.get("connectors", [])
    api_key = None
    for connector in connectors:
        if connector.get("api_key", None):
            api_key = connector["api_key"]
            break

    if api_key:
        elasticsearch["api_key"] = connectors[0]["api_key"]
        configuration["elasticsearch"] = elasticsearch
        return configuration
    else:
        raise NoElasticsearchCredentialConfiguredError()
```

seanstory:
> I think you misunderstood. Code explains it better:

Ok, so if I understand the code here, that is what I've implemented, except it would raise an error for a config like:
```
connectors:
  - service_type: foo
    connector_id: bar
```

instead of using the default auth of `elastic:changeme`. And that would mean I'd need to change all our ftest configs to use an API key or use a password other than `changeme`. Which I don't really want to do.

But otherwise, the logic is the same.

seanstory:
>  my proposal is to replace it with any API key available in connectors.

In that case, what do you think the expected behavior should be for a config like:

```
connectors:
  - connector_id: foo
    service_type: bar
    api_key: baz
  - connector_id: scoop
    service_type: doop
    api_key: whoop
```
?

I think this should be an error state, since we don't know if `baz` or `whoop` should be used at the top level, and either way, the key is likely under-privileged to manage both connectors. WDYT?

wangch079:
>I think this should be an error state, since we don't know if baz or whoop should be used at the top level

`baz` will be used. Actually, it doesn't matter. The API key should have sufficient privileges to access system indices, i.e. `.elastic-connectors` and `.elastic-connectors-sync-jobs`, and the top-level API key is used for this purpose only.

>the key is likely under-privileged to manage both connectors.

When running a sync job, i.e. write to the `search-*` index, we will replace it with the particular API key of that connector:

https://github.com/elastic/connectors-python/blob/524a5cd9c894c72970c0fd789dd3372c6b33cbf2/connectors/services/job_execution.py#L216-L229

artem-shelkovnikov:
@seanstory, does this PR still make sense?

PR #1558, opened 2023-08-31
[GMail] Validate GMail authentication

timgrein:
## Related to https://github.com/elastic/enterprise-search-team/issues/5592

This PR adds remote authentication validation for GMail (basically looking for 401 errors indicating authentication problems, like wrong field values or missing OAuth2 scopes).

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/gmail.py
+++ b/connectors/sources/gmail.py
@@ -23,7 +23,12 @@ from connectors.sources.google import (
     UserFields,
     validate_service_account_json,
 )
-from connectors.utils import base64url_to_base64, iso_utc
+from connectors.utils import (
+    EMAIL_REGEX_PATTERN,
+    base64url_to_base64,
+    iso_utc,
+    validate_email_address,
+)
 
 CUSTOMER_ID_LABEL = "Google customer id"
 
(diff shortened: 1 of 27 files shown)


connectors/sources/gmail.py
```diff
@@ -24,6 +25,12 @@
 )
 from connectors.utils import base64url_to_base64, iso_utc
 
+CUSTOMER_ID_LABEL = "Google customer id"
+
+SUBJECT_LABEL = "Subject"
+
+SERVICE_ACCOUNT_CREDENTIALS_LABEL = "GMail service account JSON"
+
```
artem-shelkovnikov, line 33: Why make it constants? It makes it harder to copy-paste configuration into JSON, for example, and does not add any value IMO

timgrein: Because I'm using these values at two places in the code, so if it's changed it's reflected at both places.

artem-shelkovnikov: Ah, overlooked it

artem-shelkovnikov (approved):

artem-shelkovnikov (approved):

PR #1513, opened 2023-08-23
[Confluence] Move common code to Atlassian module

abhishekjoshi-crest:
#### Pre-Review Checklist
- [x] this PR has a meaningful title
- [x] this PR links to all relevant github issues that it fixes or partially addresses
- [x] if there is no GH issue, please create it. Each PR should have a link to an issue
- [x] this PR has a thorough description
- [x] Covered the changes with automated tests
- [x] Tested the changes locally
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

PR #1560, opened 2023-08-31
[Jira] Move common code to Atlassian module

moxarth-elastic:
## Relates to https://github.com/elastic/connectors-python/pull/1513

There were few method which were common across Jira and Confluence. Those common methods are already moved to `atlassian.py` in https://github.com/elastic/connectors-python/pull/1513 PR. So, this PR contains the changes regarding the Jira connector with the respective changes to it.

#### Pre-Review Checklist

## Performance Stats

 - drive link: https://docs.google.com/document/d/1t4qd4AHWa3eT2xCnuHwh7che33e24uE0


connectors/sources/jira.py
```diff
@@ -292,38 +322,44 @@ def get_default_configuration(cls):
                 "label": "Jira Server username",
                 "order": 2,
                 "type": "str",
+                "value": "admin",
```
artem-shelkovnikov, line 325: These values were deleted on purpose by @navarone-feekery. They should not return

navarone-feekery: As way of an explanation: we have removed all arbitrary `value` fields, as it's much cleaner in the UI to not have preset values. All values will now default to a type-safe empty value, so you can safely leave them out of configurable fields now.

tests/sources/test_jira.py
```diff
@@ -261,22 +292,6 @@
...
-        password="changeme",
-        account_email="me@example.com",
-        api_token="abc#123",
-        jira_url="http://127.0.0.1:8080",
-        projects="*",
-        ssl_enabled=False,
-    ) as source:
-        yield source
-
-
```
navarone-feekery, line 279: Please re-implement this wrapper. This is how we are trying to set up unit test configurations going forward now that we don't set any default `value` fields.

navarone-feekery (changes requested):

connectors/sources/jira.py
moxarth-elastic: Oh! seems like some code snippets got reverted. Thanks for pointing this out @artem-shelkovnikov @navarone-feekery. Let me check and get it synced with the original code.

tests/sources/test_jira.py
moxarth-elastic: Got it!

tests/sources/test_jira.py
```diff
@@ -453,8 +586,8 @@ async def test_ping_for_failed_connection_exception(client_session_get):
 @pytest.mark.asyncio
 async def test_validate_config_for_ssl_enabled_when_ssl_ca_empty_raises_error():
     """This function test _validate_configuration when certification is empty when ssl is enabled"""
-    async with create_jira_source() as source:
-        source.configuration.get_field("ssl_enabled").value = True
+
+    async with create_source(JiraDataSource, ssl_enabled=True) as source:
```
navarone-feekery, line 590: Instead of calling `create_source` here, you could create an arg in `create_jira_source` for `ssl_enabled` so that you can still initiate all of the missing configuration values. Otherwise your configuration is initialised here empty except for `ssl_enabled`.

e.g.

```
async def create_jira_source(ssl_enabled=False):
```

and then call it like

```
 create_jira_source(ssl_enabled=True)
```

connectors/sources/jira.py
```diff
@@ -159,6 +163,38 @@ async def close_session(self):
...
+            else:
+                self._logger.warning(
+                    f"Rate Limited but Retry-After header is not found, using default retry time: {DEFAULT_RETRY_SECONDS} seconds"
+                )
+            self._logger.debug(f"Rate Limit reached: retry in {retry_seconds} seconds")
+
+            await self._sleeps.sleep(retry_seconds)
+            raise ThrottledError
+        elif exception.status == 404:
+            raise NotFound
```
timgrein, line 186: Should we log an error here, which url was not found?

connectors/sources/jira.py
```diff
@@ -159,6 +163,38 @@ async def close_session(self):
...
+        elif exception.status == 500:
```
timgrein, line 187: We could/should probably also log a warning/error here

timgrein (commented):
Comments for two places, where we could log something improving debugability.

timgrein (approved):
LGTM, @navarone-feekery are you requested changes fully addressed?

navarone-feekery (approved):

PR #1569, opened 2023-09-05
[GMail] Improve class docs and logging

timgrein:
## Relates to https://github.com/elastic/enterprise-search-team/issues/5591

This PR improves the class documentation and logging for the GMail connector for better debugability.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/gmail.py
+++ b/connectors/sources/gmail.py
@@ -100,6 +100,10 @@ def _message_doc(message):
     return es_doc
 
 
+def _filtering_enabled(filtering):
+    return filtering is not None and filtering.has_advanced_rules()
+
+
 class GMailDataSource(BaseDataSource):
     """GMail"""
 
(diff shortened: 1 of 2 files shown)


connectors/sources/gmail.py
```diff
@@ -100,6 +100,10 @@ def _message_doc(message):
     return es_doc
 
 
+def _filtering_enabled(filtering):
+    return filtering is not None and filtering.has_advanced_rules()
```
navarone-feekery, line 104: What's the reasoning behind moving this to be a module-level func?

navarone-feekery (approved):
LGTM, left a question

connectors/sources/gmail.py
timgrein: It seems to be a "pythonic", that a method, which doesn't use the `self` reference can/should be either a static/class method (if it uses something from the class) or a module level/pure function. PyCharm also warns/hints at it.

PR #1572, opened 2023-09-05
[Zoom] Moving connector code to main

### LINKED ISSUE #1571: Zoom Connector
Moving zoom connector to main branch

abhishekjoshi-crest:
#### Pre-Review Checklist


tests/sources/test_zoom.py
```diff
@@ -0,0 +1,801 @@
...
+@asynccontextmanager
+async def create_zoom_source(fetch_past_meeting_details=False):
+    async with create_source(ZoomDataSource) as source:
+        source.configuration.get_field("account_id").value = "123"
+        source.configuration.get_field("client_id").value = "id@123"
+        source.configuration.get_field("client_secret").value = "secret#123"
+        source.configuration.get_field(
+            "fetch_past_meeting_details"
+        ).value = fetch_past_meeting_details
+        yield source
```
navarone-feekery, line 413:
```suggestion
@asynccontextmanager
async def create_zoom_source(fetch_past_meeting_details=False):
    async with create_source(
        ZoomDataSource,
        account_id="123",
        client_id="id@123",
        client_secret="secret#123",
        fetch_past_meeting_details=fetch_past_meeting_details
    ) as source:
        yield source
```

You can create the data source with the values already instead of applying them after creation.

connectors/sources/zoom.py
```diff
@@ -0,0 +1,539 @@
...
+def format_chat_date(date):
+    return date.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+class TokenError(Exception):
+    pass
+
+
+class NotFound(Exception):
+    pass
```
navarone-feekery, line 73: Nit: I think this could be more clearly named. I have to search through the code to understand what could raise this error.
Can you add a prefix, e.g. `ZoomResourceNotFound` (so we can understand it's an API error and not a connector error), or a comment explaining what could cause this error?

abhishekjoshi-crest:
buildkite test this

connectors/sources/zoom.py
```diff
@@ -0,0 +1,539 @@
...
+                "order": 4,
+                "tooltip": "Enable this option to fetch past past meeting details. This setting can increase sync time.",
+                "type": "bool",
+                "value": False,
+            },
+        }
+
+    async def validate_config(self):
+        await super().validate_config()
+        await self.client.api_token.get()
```
navarone-feekery, line 356: Question: do you call this here to check that the credentials are valid? Or are you actually fetching the token for an upcoming sync job?

If the for validation, a token is already being fetched in `ping`, and `ping` is run alongside `validate_config` ([see here](https://github.com/elastic/connectors-python/blob/ff8ff703c9275c76616a0963ed9ba57a490ab42c/connectors/sync_job_runner.py#L124-L129)) so it's not necessary to check the token again.
If it's for an upcoming sync job, it feels awkward to run during validation.

In either case, I don't think it's necessary to have here.

abhishekjoshi-crest: Yes, self.client.api_token.get() kept here as it will create and validate all config params if any value is invalid it will throw an exception.

> If the for validation, a token is already being fetched in ping, and ping is run alongside validate_config ([see here](https://github.com/elastic/connectors-python/blob/ff8ff703c9275c76616a0963ed9ba57a490ab42c/connectors/sync_job_runner.py#L124-L129)) so it's not necessary to check the token again.
Shouldn't configured values remotely validated before sync job or pinging?

navarone-feekery: I can see now that the token is being cached so I'm not as concerned about the duplicate calls.

tests/sources/fixtures/zoom/Dockerfile
seanstory, line 1: I think this is no longer necessary, after: https://github.com/elastic/connectors-python/pull/1544

tests/sources/fixtures/zoom/requirements.txt
```diff
@@ -0,0 +1,2 @@
+flask==2.3.3
```
seanstory, line 1: also not required after https://github.com/elastic/connectors-python/pull/1544

seanstory (approved):

Issue #1422, opened 2023-08-10
Teams connector client (8.11 Preview) 


danajuratoni:
## New Teams connector
New connector client to be released with 8.11 in Preview.
Preliminary date for 8.11 Feature Freeze: Tue, Oct 03 2023

### Acceptance criteria
- [x] Updated [connector timeline](https://docs.google.com/spreadsheets/d/1ZRK-ospXAbE624m8g9UGKUGXbyWECAjW4401b0-ymQw/edit#gid=1132430678) 
- [x] reuse old TDD
- [x] update QA plan in [GDrive](https://drive.google.com/drive/folders/1DN2y4TRXIuybX8vawHljkWozoyPEOYaf)
- [x] Code merged in main following the defined PR process
- [x] Functional tests performed
- [x] Connector client documentation merged in repo 
- [x] Demo of the connector in latest builds in Elastic Cloud in [GDrive](https://drive.google.com/drive/folders/1DN2y4TRXIuybX8vawHljkWozoyPEOYaf)
- [x] Sign off given from an engineer in Ingestion
     - [x] Code backported to the correct version 
     - [x] Automated tests to 90% coverage
- [x] Regression test suite performed for the release along with QA day for all supported OS in https://www.elastic.co/support/matrix 
   - [ ] if no new libs are needed, skip compatibility tests

akanshi-elastic:
Working on ping, indexing and unit test cases for microsoft teams.

akanshi-elastic:
Raised ping PR on github, working on indexing feature.

akanshi-elastic:
- ping for ms teams merged
-Raised PR for indexing
-working on e2e for teams

akanshi-elastic:
indexing for ms teams merged
-Raised  e2e pr for teams

akanshi-elastic:
E2E PR merged for teams
- Working on teams documentation
- Fixing few bugs while testing the connector internally. Will raise PR in few days

khushbu-elastic:
@danajuratoni  @danajuratoni  Demo of the connector in latest builds in Elastic Cloud in [GDrive](https://drive.google.com/file/d/1Jsu07HQDxXgyYodCuqbx_Ynz0AIjuNG-/view?usp=drive_link) is uploaded

jedrazb:
Engineering sign off given ✅

jedrazb:
[...]

PR #1574, opened 2023-09-05
[New Connector] Microsoft Teams: Indexing feature

akanshi-elastic:
## Related to https://github.com/elastic/connectors-python/issues/1422###
This PR includes the indexing functionality for microsoft teams:
PR includes object types:
-USER_CHATS_MESSAGE
-USER_CHAT_TABS
-USER_CHAT_ATTACHMENT
-USER_CHAT_MEETING_RECORDING
-USER_MEETING
-TEAMS
-TEAM_CHANNEL
-CHANNEL_TAB
-CHANNEL_MESSAGE
-CHANNEL_MEETING
-CHANNEL_ATTACHMENT

#### Pre-Review Checklist


connectors/sources/microsoft_teams.py
```diff
@@ -186,25 +366,59 @@ def set_logger(self, logger_):
...
+                    for substring in [
+                        "/drive",
+                        "/items/root/children",
+                        "/children?$filter=name eq",
+                        "/events",
+                    ]
+                ):
+                    token = await self._api_token.get_with_client()
+                else:
+                    token = await self._api_token.get_with_username_password()
```
navarone-feekery, line 410: This could be abstracted into a token-generating func to make the `_get`  func cleaner.

connectors/sources/microsoft_teams.py
```diff
@@ -214,7 +428,22 @@ async def _get(self, absolute_url):
...
+            }
+            retry_seconds = int(
+                updated_response_headers.get("retry-after", RETRY_SECONDS)
+            )
+            self._logger.debug(
+                f"Rate limited by Microsoft Teams. Retrying after {retry_seconds} seconds"
+            )
+            await self._sleeps.sleep(retry_seconds)
+            raise ThrottledError(
+                f"Received Too many request error for {absolute_url}: Exception: {e}"
```
navarone-feekery, line 444:
```suggestion
                f"Service is throttled because too many requests have been made to {absolute_url}: Exception: {e}"
```

connectors/sources/microsoft_teams.py
```diff
@@ -214,7 +428,22 @@ async def _get(self, absolute_url):
...
+                key.lower(): value for key, value in response_headers.items()
+            }
+            retry_seconds = int(
+                updated_response_headers.get("retry-after", RETRY_SECONDS)
+            )
+            self._logger.debug(
+                f"Rate limited by Microsoft Teams. Retrying after {retry_seconds} seconds"
+            )
+            await self._sleeps.sleep(retry_seconds)
+            raise ThrottledError(
```
navarone-feekery, line 443: I think this retry logic doubling up with the `@retryable` decorator. Is this intended?

connectors/sources/microsoft_teams.py
```diff
@@ -266,6 +627,84 @@ def client(self):
...
+        if "" in attachment_extension:
+            self._logger.debug(
+                f"Files without extension are not supported, skipping {filename}."
+            )
+            return
+        if "." not in filename:
+            self._logger.debug(
+                f"Files without extension are not supported by TIKA, skipping {filename}."
+            )
+            return
```
navarone-feekery, line 655:
```suggestion
        if "" in attachment_extension:
            self._logger.debug(
                f"Files without extension are not supported, skipping {filename}."
            )
            return
```
From what I can understand, if the `filename` is missing `.`, then `attachment_extension` will always have an empty value, so the second check here is redundant.

Sample testing:

```
>>> os.path.splitext("test.txt")
('test', '.txt')

>>> os.path.splitext("tmp/test.txt")
('tmp/test', '.txt')

>>> os.path.splitext("tmp.txt/test.txt")
('tmp.txt/test', '.txt')

>>> os.path.splitext(".txt")
('.txt', '')

>>> os.path.splitext(".")
('.', '')

>>> os.path.splitext("test")
('test', '')
```

connectors/sources/microsoft_teams.py
```diff
@@ -266,6 +627,84 @@ def client(self):
...
+        Args:
+            document(dictionary): Modified document
+            item (dictionary): Document from Microsoft Teams.
+            document_type(string): Name of function to be called for fetching the mapping.
+
+        Returns:
+            dictionary: Modified document with the help of adapter schema.
+        """
+        for elasticsearch_field, sharepoint_field in document_type().items():
+            document[elasticsearch_field] = item[sharepoint_field]
```
navarone-feekery, line 706: No change request, I just wanted to point out that this is a really neat way to deal with mappings. I will personally try to use this in the Salesforce connector sometime. Nice!

navarone-feekery: Although one followup question: is it possible for `item[sharepoint_field]` to throw a `KeyError` or are we always sure that the API returns the expected response format?
If we aren't sure, using `item.get(sharepoint_field)` might be safer.

connectors/sources/microsoft_teams.py
```diff
@@ -281,6 +720,475 @@ async def ping(self):
...
+
+            # If type of meeting  is monthly so body will be: Recurrence: Occurs every month on day 5 of march
+            # starting {date} until {enddate}
+            elif pattern.get("type", "") in ["absoluteMonthly", "relativeMonthly"]:
+                days_pattern = (
+                    f"on day {pattern['dayOfMonth']}"
+                    if pattern.get("dayOfMonth")
+                    else "on {pattern['index']} {','.join(pattern['daysOfWeek'])}"
+                )
+                days = f"{occurrence} month {days_pattern}"
```
navarone-feekery, line 775: `pattern.get("type", "")` is called a lot, I think it could be a variable, something like;

```suggestion
            pattern_type = pattern.get("type", "")

            # In case type of meeting is daily so body will be: Recurrence: Occurs every 1 day starting {startdate}
            # until {enddate}
            if pattern_type == "daily":
                days = f"{occurrence} day"

            # If type of meeting  is yearly so body will be: Recurrence: Occurs every year on day 5 of march starting
            # {date} until {enddate}
            elif pattern_type in ["absoluteYearly", "relativeYearly"]:
                day_pattern = (
                    f"on day {pattern['dayOfMonth']}"
                    if pattern.get("dayOfMonth")
                    else "on {pattern['index']} {','.join(pattern['daysOfWeek'])}"
                )
                days = f"year {day_pattern} of {month_name[pattern['month']]}"

            # If type of meeting  is monthly so body will be: Recurrence: Occurs every month on day 5 of march
            # starting {date} until {enddate}
            elif pattern_type in ["absoluteMonthly", "relativeMonthly"]:
                days_pattern = (
                    f"on day {pattern['dayOfMonth']}"
                    if pattern.get("dayOfMonth")
                    else "on {pattern['index']} {','.join(pattern['daysOfWeek'])}"
                )
                days = f"{occurrence} month {days_pattern}"
```

connectors/sources/microsoft_teams.py
```diff
@@ -281,6 +720,475 @@ async def ping(self):
...
+            }
+        )
+        return document
+
+    def get_calendar_detail(self, calendar):
+        body = ""
+        organizer = calendar.get("organizer", {}).get("emailAddress").get("name")
+        if calendar.get("recurrence"):
+            recurrence_range = calendar.get("recurrence", {}).get("range")
+            pattern = calendar.get("recurrence", {}).get("pattern")
```
navarone-feekery, line 749: `calendar.get("recurrence")` is called a lot, it could be a variable. You wouldn't have to worry about safe-calling it into a dict this way either.

```suggestion
        calendar_recurrence = calendar.get("recurrence")
        if calendar_recurrence:
            recurrence_range = calendar_recurrence.get("range")
            pattern = calendar_recurrence.get("pattern")
```

connectors/sources/microsoft_teams.py
```diff
@@ -281,6 +720,475 @@ async def ping(self):
...
+            if pattern.get("type", "") == "daily":
+                days = f"{occurrence} day"
+
+            # If type of meeting  is yearly so body will be: Recurrence: Occurs every year on day 5 of march starting
+            # {date} until {enddate}
+            elif pattern.get("type", "") in ["absoluteYearly", "relativeYearly"]:
+                day_pattern = (
+                    f"on day {pattern['dayOfMonth']}"
+                    if pattern.get("dayOfMonth")
+                    else "on {pattern['index']} {','.join(pattern['daysOfWeek'])}"
```
navarone-feekery, line 763:
```suggestion
                    else f"on {pattern['index']} {','.join(pattern['daysOfWeek'])}"
```

connectors/sources/microsoft_teams.py
```diff
@@ -281,6 +720,475 @@ async def ping(self):
...
+                )
+                days = f"year {day_pattern} of {month_name[pattern['month']]}"
+
+            # If type of meeting  is monthly so body will be: Recurrence: Occurs every month on day 5 of march
+            # starting {date} until {enddate}
+            elif pattern.get("type", "") in ["absoluteMonthly", "relativeMonthly"]:
+                days_pattern = (
+                    f"on day {pattern['dayOfMonth']}"
+                    if pattern.get("dayOfMonth")
+                    else "on {pattern['index']} {','.join(pattern['daysOfWeek'])}"
```
navarone-feekery, line 773:
```suggestion
                    else f"on {pattern['index']} {','.join(pattern['daysOfWeek'])}"
```

connectors/sources/microsoft_teams.py
```diff
@@ -281,6 +720,475 @@ async def ping(self):
...
+            ):
+                document.update(
+                    {
+                        "title": event.get("callRecordingDisplayName"),
+                        "recording_url": event.get("callRecordingUrl"),
+                    }
+                )
+        return document
+
+    async def get_user_chat_messages(self, chat, message):
```
navarone-feekery, line 884: To avoid confusion API methods

```suggestion
    async def format_user_chat_messages(self, chat, message):
```

connectors/sources/microsoft_teams.py
```diff
@@ -281,6 +720,475 @@ async def ping(self):
...
+        self._logger.debug(f"Downloading {filename}")
+        attachment_content = await self._download_content_for_attachment(
+            partial(self.client.download_item, download_url),
+            original_filename=filename,
+        )
+        document["_attachment"] = attachment_content
+
+        return document
+
+    def get_attachment_names(self, attachments):
```
navarone-feekery, line 968:
```suggestion
    def format_attachment_names(self, attachments):
```

connectors/sources/microsoft_teams.py
```diff
@@ -281,6 +720,475 @@ async def ping(self):
...
+                "chatType": chat.get("chatType"),
+                "sender": message["from"]["user"].get("displayName")
+                if message.get("from") and message["from"].get("user")
+                else "",
+                "message": html_to_text(html=message.get("body", {}).get("content")),
+            }
+        )
+        return message
+
+    async def get_user_chat_attachments(self, **kwargs):
```
navarone-feekery, line 906:
```suggestion
    async def format_user_chat_attachments(self, **kwargs):
```

connectors/sources/microsoft_teams.py
akanshi-elastic: I have already verified for the fields, they will come everytime in `item` but still we can use `.get` to make it more safe.

connectors/sources/microsoft_teams.py
```diff
@@ -144,9 +319,14 @@ async def _fetch_token(self):
             client_credential=self.client_secret,
             authority=authority,
         )
-        token_metadata = auth_context.acquire_token_by_username_password(
-            username=self.username, password=self.password, scopes=SCOPE
-        )
+        if is_acquire_for_client:
+            token_metadata = auth_context.acquire_token_for_client(
+                scopes=[GRAPH_ACQUIRE_TOKEN_URL]
+            )
```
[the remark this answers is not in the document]
akanshi-elastic: there are different type of scopes required for `acquire_token_for_client` and `acquire_token_by_username_password`. For `acquire_token_for_client`  we need [".default"] scope to be added and for `acquire_token_by_username_password` : ["User.Read.All", "TeamMember.Read.All", "ChannelMessage.Read.All", "Chat.Read", "Chat.ReadBasic", "Calendars.Read"] must be passed to generate token.. To make it more clear I can pass  [".default"] instead of the complete url. Both will perform the same operation.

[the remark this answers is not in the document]
navarone-feekery: If you're sure, no need to change it. I just noticed they were different and wanted to confirm why.

connectors/sources/microsoft_teams.py
akanshi-elastic: But in that case we need to add this condition at 3 places. In `get_with_client` method we need to add one condition for checking cached_value, one for fetch_token and one for setting token to cache.
To remove it from here we can include one more fucntion where it can decide which function needs to call for generating token `get_with_client` or `get_with_username_password` what do you think on this?

connectors/sources/microsoft_teams.py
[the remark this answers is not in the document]
akanshi-elastic: The scopes are different only. For `acquire_token_for_client` we need to pass `default` in scope and for `acquire_token_by_username_password` we need to pass the scope which is related to fetching chats and teams metadata.
Passing [".default"] or [""https://graph.microsoft.com/.default"] for `acquire_token_for_client` is same thing.

connectors/sources/microsoft_teams.py
akanshi-elastic: Can you please elaborate more on this? How it is doubling up the logic? 
I guess this will retry 3 times(i.e. number of RETRY_COUNT)

navarone-feekery: On line 452 the code is being slept for x seconds, and then the error is raised. After raising the error, the `@retryable` decorator will make the code sleep again for `RETRY_INTERVAL` seconds. So it's doubling up in retry time. 

I can see that we will need to respect the `retry-after` value in the response headers, so I don't see an easy workaround. The code will just pause for longer than expected because it will pause twice.

It's okay like this, I just wanted to make sure we're all aware that this is intended.

connectors/sources/microsoft_teams.py
navarone-feekery: If changing this has such an impact, let's leave it where it is. The token isn't needed anywhere so abstraction is just a preference, not a requirement.

navarone-feekery (approved):
LGTM

PR #1576, opened 2023-09-05
Ensure indices are created on startup

sphilipse:
This ensures the system indices are created by the connector on startup if not already present. 

**Release Note**
Connectors now create the connectors indices by indexing an empty document to trigger the Elasticsearch-installed index template.

### DIFF
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -10,7 +10,7 @@ import time
 from enum import Enum
 
 from elastic_transport.client_utils import url_to_node_config
-from elasticsearch import ApiError, AsyncElasticsearch, ConflictError, NotFoundError
+from elasticsearch import ApiError, AsyncElasticsearch, ConflictError
 from elasticsearch import ConnectionError as ElasticConnectionError
 
 from connectors import __version__
@@ -171,8 +171,10 @@ class ESClient:
         for index in indices:
             logger.debug(f"Checking index {index}")
             if not await self.client.indices.exists(index=index):
-                await self.client.index(index=index, document={}, id='.connectors-create-doc')
-                await self.client.delete(index=index, id='.connectors-create-doc')
+                await self.client.index(
+                    index=index, document={}, id=".connectors-create-doc"
+                )
+                await self.client.delete(index=index, id=".connectors-create-doc")
                 logger.debug(f"Created index {index}")
 
     async def delete_indices(self, indices):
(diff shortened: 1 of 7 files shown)


connectors/es/client.py
```diff
@@ -164,23 +164,16 @@ async def wait(self):
         await self.close()
         return False
 
-    async def check_exists(self, indices=None, pipelines=None):
+    async def ensure_exists(self, indices=None):
         if indices is None:
             indices = []
-        if pipelines is None:
```
vidok, line 170: do the indices templates define the pipelines? is there still a need to check them?

connectors/preflight_check.py
```diff
@@ -92,8 +92,10 @@ async def _check_system_indices_with_retries(self):
         while self.running:
             try:
                 # Checking the indices/pipeline in the loop to be less strict about the boot ordering
-                await self.es_client.check_exists(
-                    indices=[CONNECTORS_INDEX, JOBS_INDEX]
+                # Using concrete write index to create these to ensure ES-installed template takes
+
+                await self.es_client.ensure_exists(
+                    indices=[CONNECTORS_INDEX + '-v1', JOBS_INDEX +'-v1']
```
vidok, line 98: Could you please add a comment on why `-v1` is being used? Or maybe move the aliases to the constants?

vidok (commented):
Should this also work for dedicated deployments? 

Could you please link the changes you've made in Elasticsearch so we can always refer to that information?

connectors/preflight_check.py
sphilipse: There's a comment there! We're using the concrete write index because the template is triggered when targeting the concrete write index. We cannot use an alias to create the index (the alias is created by the template!), but we can use it for everything else.

sphilipse:
> Should this also work for dedicated deployments?
Yes, this is not specific to Serverless. 
 
> Could you please link the changes you've made in Elasticsearch so we can always refer to that information?
The templates are installed [here](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/ConnectorTemplateRegistry.java) and located [here](https://github.com/elastic/elasticsearch/tree/main/x-pack/plugin/core/template-resources/src/main/resources/entsearch/connector). I'll add these as a comment to this PR.

seanstory (dismissed):

seanstory (dismissed):

seanstory (dismissed):

seanstory (approved):

PR #1590, opened 2023-09-06
Add common perftest classes for data generation + use in confluence to demonstrate

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/3397

Adding a series of classes that can be used to have common setup for performance tests. `WeightedFakeProvider` class allows to have a class that returns random small/medium/large/extra_large text/html files with given distribution.

For example, by default distribution is `[0.58, 0.3, 0.1, 0.02]`, meaning that small items will be returned with 58% chance, medium-sized items with 30% chance, large with 10% and extra_large with 2%.

The idea is to have this class used in all functional tests to have uniform test setup, so that we could do better job at estimating the memory consumption of each connector.

One thing to consider: I had to change all `python <FIXTURE_FILE.py>` to `bin/python <FIXTURE_FILE.py>` for it to work, because the shared class is inside of the connectors library, and this library is not installed locally - it's installed inside the venv. So to use it, we need to use python that's distributed in venv as well. IMO it makes sense.

#### Pre-Review Checklist


tests/sources/fixtures/confluence/fixture.py
```diff
@@ -7,26 +7,50 @@
...
 
-DATA_SIZE = os.environ.get("DATA_SIZE", "small").lower()
-_SIZES = {"small": 1000000, "medium": 2000000, "large": 6000000}
-FILE_SIZE = _SIZES[DATA_SIZE]
-LARGE_DATA = "".join([random.choice(string.ascii_letters) for _ in range(FILE_SIZE)])
+from tests.commons import WeightedFakeProvider
+
+fake_provider = WeightedFakeProvider()
+
+DATA_SIZE = os.environ.get("DATA_SIZE", "medium")
```
seanstory, line 18: missing a `.lower()`?

seanstory: also, intentional that the default went from `"small"` to `"medium"`?

tests/sources/fixtures/confluence/fixture.py
```diff
@@ -7,26 +7,50 @@
...
+
+match DATA_SIZE:
+    case "small":
+        SPACE_COUNT = 100
+        SPACE_OBJECT_COUNT = 100
+        ATTACHMENT_COUNT = 3
+    case "medium":
+        SPACE_COUNT = 100
+        SPACE_OBJECT_COUNT = 200
+        ATTACHMENT_COUNT = 5
```
seanstory, line 28: These look like a pretty significant change from what the ftest is currently doing. Why not add new `cases` with. these numbers, instead of replacing the current test case?

If I'm reading the code right, before this change, it was effectively:
```
SPACE_COUNT = 4000
SPACE_OBJECT_COUNT = 50
ATTACHMENT_COUNT = 6
```

tests/sources/fixtures/confluence/fixture.py
artem-shelkovnikov: Good catch! Will do `.lower()`.

As about data_size - all defaults should be consistent, so swapped here to "medium".

Our nightly and on-commit tests use explicit DATA_SIZE anyway, so it's the change for local testing without providing DATA_SIZE.

tests/sources/fixtures/confluence/fixture.py
artem-shelkovnikov: Yes, but the data changed - before it was static size files, now they are different - some are small, some are bigger.

So before sync took 59 seconds and synced 1MB files.

Now sync takes around 78 seconds and syncs files from 100KB to 25MB files.

On the screenshots you see how the memory profiles changed:

Before:

image

After:

image

TDLR: if I use the old setup, the sync will take significantly more time - approximately `(4000 * 50 * 6) / (200 * 100 * 5)=12` times more making it ~12-15 minutes.

artem-shelkovnikov:
Started the migration of other sources in https://github.com/elastic/connectors-python/pull/1607. Did it for ABS and Dropbox there

seanstory (dismissed):

seanstory:
Unclear to me why CI is failing, we'll need to dig into that.

navarone-feekery (approved):
LGTM!

PR #1593, opened 2023-09-07
[Sharepoint Online] Debug pretty print group names for ACL validation

jedrazb:
Put human-readable group names into `_allow_access_control` prefixed with `debug_group_pretty`. It won't affect how DLS works. This patch should be only used for verfication purposes.  

Leaving it as draft PR. More discussion is needed if/how we want to productionize it in the future.

Note: some tests are failing, this is expected. Won't be fixing them unless we plan to merge it at some point.

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -919,8 +919,10 @@ class SharepointOnlineAdvancedRulesValidator(AdvancedRulesValidator):
 def _prefix_group(group):
     return prefix_identity("group", group)
 
+
 def _prefix_group_pretty(group):
     return prefix_identity("debug_group_pretty", group)
+
 
 def _prefix_user(user):
     return prefix_identity("user", user)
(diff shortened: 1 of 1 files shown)


timgrein:
Could we hide it behind a RFC under the `advanced` section? Sounds like a useful feature. cc @danajuratoni

seanstory:
As long as we don't put this on the data's `_allow_permissions`, I don't see an issue with putting them on the identities in the ACL index. Though I might suggest that we map the group name to the id, like: 
```
def _prefix_group_pretty(group_name, group_id):
    return prefix_identity("debug_group_pretty", f"{group_name}:{group_id}")

...
user_access_control.add(_prefix_group(domain_group_id))
user_access_control.add(_prefix_group_pretty(user["Title"], domain_group_id))
```

sadusumisbux:
👍

artem-shelkovnikov:
@jedrazb does this PR still make sense?

jedrazb:
> does this PR still make sense?
@artem-shelkovnikov this PR was meant to serve as an example for a customer. I think we can close the PR for now and revisit this functionality if product team finds this important

PR #1619, opened 2023-09-12
Improve OneDrive Connector Performance

praveen-elastic:
## Closes https://github.com/elastic/connectors-py/issues/1618

With the following change, the execution time is reduced by half 🎉 so it now takes 55 minutes in place of 2 hours to index 51k documents.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/onedrive.py
+++ b/connectors/sources/onedrive.py
@@ -46,15 +46,14 @@ FILE = "file"
 FOLDER = "folder"
 
 USERS = "users"
-CONTENT = "content"
 DELTA = "delta"
 PING = "ping"
+ITEM_FIELDS = "id,name,lastModifiedDateTime,content.downloadUrl,createdDateTime,size,webUrl,parentReference,file,folder"
 
 ENDPOINTS = {
     PING: "drives",
-    USERS: "users",
-    DELTA: "users/{user_id}/drive/root/delta",
-    CONTENT: "users/{user_id}/drive/items/{item_id}/content",
+    USERS: "users?$top={page_size}",
+    DELTA: "users/{user_id}/drive/root/delta?$select={item_fields}&$top={page_size}",
 }
 
 if "OVERRIDE_URL" in os.environ:
@@ -278,7 +277,6 @@ class OneDriveClient:
 
     async def paginated_api_call(self, url_name, **url_kwargs):
         url = parse.urljoin(BASE_URL, ENDPOINTS[url_name].format(**url_kwargs))
-        url = f"{url}?$top={FETCH_SIZE}"
 
         while True:
             try:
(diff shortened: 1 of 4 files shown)


connectors/sources/onedrive.py
```diff
@@ -303,12 +301,19 @@ async def paginated_api_call(self, url_name, **url_kwargs):
                 break
 
     async def list_users(self):
-        async for response in self.paginated_api_call(url_name=USERS):
+        async for response in self.paginated_api_call(
+            url_name=USERS, page_size=FETCH_SIZE
```
wangch079, line 305: I think it's better to make `FETCH_SIZE` as the default value of `page_size`, so you can call it with `self.paginated_api_call(url_names=USERS)

connectors/sources/onedrive.py
```diff
@@ -535,23 +558,25 @@ async def get_docs(self, filtering=None):
...
                     user_id = user_mail_id_map.get(mail)
-                    async for entity in self.client.get_owned_files(
+                    async for entity, download_url in self.client.get_owned_files(
                         user_id, skipped_extensions, pattern
                     ):
                         entity = self.prepare_doc(entity)
                         if entity["type"] == FILE:
                             yield entity, partial(
-                                self.get_content, entity.copy(), user_id
+                                self.get_content, entity.copy(), download_url
```
wangch079, line 567: `download_url` can be easily checked here, instead of yielding it blindly. 

Although it returns immediately in `get_content`, it will still be wrapped and executed as a task, which is not necessary.

connectors/sources/onedrive.py
```diff
@@ -303,12 +301,19 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
             for user_detail in response:
                 yield user_detail
 
     async def get_owned_files(self, user_id, skipped_extensions=None, pattern=""):
-        async for response in self.paginated_api_call(url_name=DELTA, user_id=user_id):
+        async for response in self.paginated_api_call(
+            url_name=DELTA,
+            user_id=user_id,
+            page_size=FETCH_SIZE,
+            item_fields=ITEM_FIELDS,
```
wangch079, line 315: We should probably make `paginated_api_call` as general as possible.

From the name `paginated_api_call`, I'm only expecting parameters like `url` and `page_size`.

connectors/sources/onedrive.py
```diff
@@ -276,10 +275,7 @@ async def get(self, url):
             else:
                 raise
 
-    async def paginated_api_call(self, url_name, **url_kwargs):
-        url = parse.urljoin(BASE_URL, ENDPOINTS[url_name].format(**url_kwargs))
-        url = f"{url}?$top={FETCH_SIZE}"
-
+    async def paginated_api_call(self, url):
```
wangch079, line 278: Just one nit, I do think the method paginated_api_call should allow callers to specify the `fetch_size`.

```suggestion
    async def paginated_api_call(self, url, params=None, fetch_size=FETCH_SIZE):
        if params is None:
            params = {}
        params["$top"] = fetch_size
        params = "&".join(f"{key}={val}" for key, val in params.items())
        url = url + "?" + params # should apply url escape here?
```

praveen-elastic: You're right. Updated with commit [f8cce96](https://github.com/elastic/connectors-python/pull/1619/commits/f8cce9667efb0fc7a29eb7ae454f3189c857dc0b)

wangch079 (approved):
Nice work

PR #1623, opened 2023-09-12
[Zoom] Add recording age as config param

### LINKED ISSUE #1624: [Zoom] Recording age as config param
Add support for fetching all duration recordings #1420

abhishekjoshi-crest:
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/zoom.py
+++ b/connectors/sources/zoom.py
@@ -341,12 +341,6 @@ class ZoomDataSource(BaseDataSource):
                 "sensitive": True,
                 "type": "str",
             },
-            "recording_age": {
-                "label": "Recording Age Limit (Months)",
-                "order": 5,
-                "tooltip": "Set the maximum age (in months) for retaining recordings.",
-                "type": "int",
-            },
             "fetch_past_meeting_details": {
                 "display": "toggle",
                 "label": "Fetch past meeting details",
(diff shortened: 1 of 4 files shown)


connectors/sources/zoom.py
```diff
@@ -341,6 +341,12 @@ def get_default_configuration(cls):
                 "sensitive": True,
                 "type": "str",
             },
+            "recording_age": {
+                "label": "Recording Age Limit (Months)",
+                "order": 5,
```
seanstory, line 346: This is pedantic, but let's put this below `fetch_past_meeting_details` so that the `order` matches the order they are in this list.

connectors/sources/zoom.py
```diff
@@ -341,6 +341,12 @@ def get_default_configuration(cls):
...
+                "tooltip": "Set the maximum age (in months) for retaining recordings.",
```
seanstory, line 347: This isn't about _retaining_ recordings (zoom is what's _retaining_ things), but about how far back in history to search. For slack, we worded this like:
```
            "fetch_last_n_days": {
                "display": "numeric",
                "label": "Days of message history to fetch",
                "order": 2,
                "tooltip": "How far back in time to request message history from slack. Messages older than this will not be indexed.",
                "type": "int",
            },
```

Could we do something similar here?

docs/reference/zoom.md
```diff
@@ -75,6 +75,10 @@ The "Client Secret" refers to a confidential piece of information generated when
 
 - `eieiUJRsiH543P5NbYadavczjkqgdRTw`
 
+#### `Recording Age Limit (Months)`  (required)
+
+Set the maximum age (in months) for retaining recordings. This configuration parameter allows you to define a time limit, measured in months, for which recordings will be indexed.
```
seanstory, line 80: same comment as above around "retaining" not being the right term, IMO

seanstory (commented):
small nits, but mostly LGTM.

connectors/sources/zoom.py
abhishekjoshi-crest: Sure!

seanstory (approved):

PR #1631, opened 2023-09-13
[New Connector] Box Implementation

parth-elastic:
## Closes https://github.com/elastic/connectors-py/issues/1630

This PR includes Indexing of Box connector.

Note: In this PR we have used OAuth2.0 as Box free account does not support JWT or Client Credentials Grant. Once we get the Paid Box instance we raised follow up PR to add support of JWT/Client Credentials Grant.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/box.py
+++ b/connectors/sources/box.py
@@ -1,0 +1,377 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+"""Box source module responsible to fetch documents from Box"""
+import asyncio
+from datetime import datetime, timedelta
+from functools import cached_property, partial
+
+import aiofiles
+import aiohttp
+from aiofiles.os import remove
+from aiofiles.tempfile import NamedTemporaryFile
+from aiohttp.client_exceptions import ClientResponseError
+
+from connectors.logger import logger
+from connectors.source import BaseDataSource
+from connectors.utils import (
+    TIKA_SUPPORTED_FILETYPES,
+    CacheWithTimeout,
+    CancellableSleeps,
+    ConcurrentTasks,
+    MemQueue,
+    RetryStrategy,
+    convert_to_b64,
+    retryable,
+)
+
+ENDPOINTS = {
+    "TOKEN": "/oauth2/token",
+    "PING": "/2.0/users/me",
+    "FOLDER": "/2.0/folders/{folder_id}/items",
+    "CONTENT": "/2.0/files/{file_id}/content",
+}
+RETRIES = 3
+RETRY_INTERVAL = 2
+FILE_SIZE_LIMIT = 10485760
+CHUNK_SIZE = 1024
+FETCH_LIMIT = 1000
+QUEUE_MEM_SIZE = 5 * 1024 * 1024  # ~ 5 MB
+MAX_CONCURRENCY = 2000
+CONCURRENT_DOWNLOADS = 15
+FIELDS = "name,modified_at,download_url,size,type,sequence_id,etag,created_at,modified_at,content_created_at,content_modified_at,description,created_by,modified_by,owned_by,parent,item_status"
+FILE = "file"
+
+refresh_token = None
+
+
+class TokenError(Exception):
+    pass
+
+
+class NotFound(Exception):
+    pass
+
+
+class AccessToken:
+    def __init__(self, configuration, http_session):
+        global refresh_token
+        self.client_id = configuration["client_id"]
+        self.client_secret = configuration["client_secret"]
+        self._http_session = http_session
+        if refresh_token is None:
+            refresh_token = configuration["refresh_token"]
+        self._token_cache = CacheWithTimeout()
+
+    async def get(self):
+        if cached_value := self._token_cache.get_value():
+            return cached_value
+        await self._set_access_token()
+        return self.access_token
+
+    async def _set_access_token(self):
+        global refresh_token
+        data = {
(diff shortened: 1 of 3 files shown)


connectors/sources/box.py
```diff
@@ -0,0 +1,445 @@
...
+        )
+        await self._sleeps.sleep(retry_after)
+        raise Exception("Rate limit exceeded.")
+
+    async def _handle_client_errors(self, exception):
+        match exception.status:
+            case 401:
+                await self.token._set_access_token()
+                raise
+            case 429:
```
vidok, line 157: What's the purpose of using the custom `_put_to_sleep` method instead of relying on  `@retryable` wrapper?

connectors/sources/box.py
```diff
@@ -0,0 +1,445 @@
...
+    def client(self):
+        return BoxClient(configuration=self.configuration)
+
+    def tweak_bulk_options(self, options):
+        """Tweak bulk options as per concurrent downloads support by Box
+
+        Args:
+            options (dict): Config bulker options.
+        """
+        options["concurrent_downloads"] = CONCURRENT_DOWNLOADS
```
vidok, line 229: What is the reason for rewriting the bulk options?

connectors/sources/box.py
```diff
@@ -0,0 +1,445 @@
...
+                if offset >= total_count:
+                    break
+                offset += FETCH_LIMIT
+        except Exception:
+            raise
+
+    async def ping(self):
+        await self.get(url=ENDPOINTS["PING"], headers={})
+
+    async def close(self):
```
vidok, line 199: Should we also clean up the queue and the fetchers?

connectors/sources/box.py
```diff
@@ -0,0 +1,445 @@
...
+            self._logger.exception("Error while connecting to Box.")
+            raise
+
+    async def get_users_id(self):
+        async for user in self.client.paginated_call(
+            url=ENDPOINTS["USERS"], params={}, headers={}
+        ):
+            yield user.get("id")
+
+    async def _fetch_files_folders(self, doc_id, user_id=None):
```
vidok, line 293: Does this method sync all folders' files? Or both files and folders?

connectors/sources/box.py
```diff
@@ -0,0 +1,445 @@
...
+class BoxDataSource(BaseDataSource):
+    name = "Box"
+    service_type = "box"
+
+    def __init__(self, configuration):
+        super().__init__(configuration=configuration)
+        self.configuration = configuration
+        self.tasks = 0
+        self.is_enterprise = configuration["is_enterprise"]
+        self.queue = MemQueue(maxmemsize=QUEUE_MEM_SIZE, refresh_timeout=120)
```
vidok, line 213: What is the need to have a queue?

connectors/sources/box.py
```diff
@@ -0,0 +1,445 @@
...
+            base_url="https://api.box.com", raise_for_status=True
+        )
+        self.token = AccessToken(
+            configuration=configuration, http_session=self._http_session
+        )
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    async def _put_to_sleep(self, exception):
```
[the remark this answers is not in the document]
parth-elastic: Done

connectors/sources/box.py
parth-elastic: We are using `_put_to_sleep` because at the time of `429` we need to wait for some specific amount of time, which we get from the retry-after header, but how can we pass it to `@retryable`?

connectors/sources/box.py
parth-elastic: Box has 1000 API requests per minute. So at a time, we can download approximately 15 files. So, I have changed the bulk options to 15.

connectors/sources/box.py
parth-elastic: In the consumer method, we are getting all the documents from the queue. So, it will empty at the end, and fetchers are nothing but tasks, so i think there is no need to clean it.

connectors/sources/box.py
parth-elastic: Yes this method sync both files and folders.

connectors/sources/box.py
parth-elastic: We are fetching files in a recursive manner in the `_fetch_files_folder`. So, without fetchers, it's fetching files in a synchronized manner only. But with the fetchers, when the `_fetch_files_folder` function is called, we put it into the fetchers, and the fetchers add that document to the queue, and parallelly consumer function will yield the data from the queue.

connectors/sources/box.py
vidok: Alright, there is no way to pass it to @retryable right now.

connectors/sources/box.py
vidok: When the framework cancels the sync job it should clean up all the resources (HTTP connections, queues, etc.). The framework will be waiting for the job to be canceled only for a specific time (5 seconds by default) so cleaning the queue might be useful to avoid "kill" stop and do everything gracefully.

connectors/sources/box.py
vidok: Concurrent downloads can be part of the global framework configuration or connector configuration. It cannot be a "hardcoded" constant. If you want a limit, you should introduce `MAX_CONCURRENT_DOWNLOADS` and lower the number if it's higher than the limit.

connectors/sources/box.py
parth-elastic: I have cleaned the queue at the `close` method of `BoxDataSource` as we are using queue in `BoxDataSource` [here](https://github.com/elastic/connectors-python/pull/1631/commits/e5e9edd9a3956e13ba02709106a2b601c1f9ee11). Let me know if I missed something.

vidok (approved):

PR #1635, opened 2023-09-13
[OneDrive] Document level security 

### LINKED ISSUE #1628: Add DLS support for OneDrive as defined in the discovery

abhishekjoshi-crest:
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/onedrive.py
+++ b/connectors/sources/onedrive.py
@@ -646,11 +646,6 @@ class OneDriveDataSource(BaseDataSource):
         if not self._dls_enabled():
             return []
 
-        def _get_id(permission, identity):
-            if identity not in permission:
-                return
-            return permission.get(identity, {}).get("id")
-
         permissions = []
         async for permission in self.client.list_file_permission(
             user_id=user_id, file_id=file_id
@@ -660,8 +655,8 @@ class OneDriveDataSource(BaseDataSource):
 
             if identities := permission.get("grantedToIdentitiesV2"):
                 for identity in identities:
-                    user_permission = _get_id(identity, "user")
-                    group_permission = _get_id(identity, "group")
+                    user_permission = identity.get("user", {}).get("id")
+                    group_permission = identity.get("group", {}).get("id")
 
                     if user_permission:
                         permissions.append(_prefix_user_id(user_permission))
--- a/connectors/sources/onedrive.py
+++ b/connectors/sources/onedrive.py
@@ -646,6 +646,11 @@ class OneDriveDataSource(BaseDataSource):
         if not self._dls_enabled():
             return []
 
+        def _get_id(permission, identity):
+            if identity not in permission:
+                return
+            return permission.get(identity, {}).get("id")
+
         permissions = []
         async for permission in self.client.list_file_permission(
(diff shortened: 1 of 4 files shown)


connectors/sources/onedrive.py
```diff
@@ -529,6 +582,90 @@ def prepare_doc(self, file):
...
+            yield await self._user_access_control_doc(user=user)
+
+    async def get_entity_permission(self, user_id, file_id):
+        if not self._dls_enabled():
+            return []
+
+        def _get_id(permission, identity):
+            if identity not in permission:
+                return
+            return permission.get(identity).get("id")
```
wangch079, line 647:
```suggestion
            return permission.get(identity, {}).get("id")
```

connectors/sources/onedrive.py
```diff
@@ -572,9 +709,14 @@ async def get_docs(self, filtering=None):
 
                 async for entity, download_url in self.client.get_owned_files(user_id):
                     entity = self.prepare_doc(entity)
+                    entity_permissions = await self.get_entity_permission(
+                        user_id=user_id, file_id=entity.get("_id")
+                    )
```
wangch079, line 714: Although `get_entity_permission`  is guarded by `if not self._dls_enabled():`, do you think it's better to wrap it in `_decorate_with_access_control`? 

There shouldn't be duplicated calls for a single entity unless the logic itself generating duplicated entity, for which the current code will result in duplicated calls to `get_entity_permission` as well.

connectors/sources/onedrive.py
```diff
@@ -305,12 +331,30 @@ async def paginated_api_call(self, url, params=None, fetch_size=FETCH_SIZE):
                 break
 
     async def list_users(self):
```
wangch079, line 333: Consider make it optional to include groups
```suggestion
    async def list_users(self, include_groups=False):
```

wangch079 (commented):
It seems entity permission only applies when advanced rules are not configured.

connectors/sources/onedrive.py
abhishekjoshi-crest: Yes! Better to wrap in `_decorate_with_access_control`

connectors/sources/onedrive.py
```diff
@@ -529,6 +584,93 @@ def prepare_doc(self, file):
...
+        async for user in self.client.list_users(include_groups=True):
+            yield await self._user_access_control_doc(user=user)
+
+    async def get_entity_permission(self, user_id, file_id):
+        if not self._dls_enabled():
+            return []
+
+        def _get_id(permission, identity):
+            if identity not in permission:
+                return
```
wangch079, line 651: With the change, you don't need these two lines.

Actually, I don't feel it's necessary to define this function, and it's a bit confusing, because you pass in `identity` to param `permission`:
```python
user_permission = _get_id(identity, "user")
group_permission = _get_id(identity, "group")
```

I think you can simply remove this function, and do
```python
user_permission = identity.get("user", {}).get("id")
group_permission = identity.get("user", {}).get("group")
```

wangch079 (approved):

Issue #1648, opened 2023-09-14
[SPO] RCF for individual site lookup by path


seanstory:
### Problem Description
Today, the SPO connector does a `GET /sites?search=*`, and pages through _all_ sites to make sure that the configured site collections exist. 

This is efficient when the number of configured site collections is larger than the number of pages of sites (total sites / 200).

However, when just a single site is configured, this is very inefficient.

In addition, we've been noticing inconsistencies in the results returned from `GET /sites?search=*`, where sometimes some sites don't seem to be included in the full listing of sites retrieved in this manner. 

We can look up sites by ID (which would be hard for end users to locate) or by _path_, which looks a lot like `"name"` (which is what we have them configure today. The API to look up sites by path is: `GET /sites/{hostname}:/{server-relative-path}`.  I've got a POC branch up that illustrates this capability: https://github.com/elastic/connectors-python/pull/1639

### Proposed Solution
- [ ] add an RCF that lets you toggle behavior
- [ ] add functionality to look up sites by path, rather than by name out of "all sites"
- [ ] document the new approach, and recommend folks use it .

PR #1653, opened 2023-09-15
[SPO] individual site lookup RCF

seanstory:
## Closes https://github.com/elastic/connectors-python/issues/1648

Sites are currently looked up by name, out of the paginated results of enumerating all sites. This is efficient when large numbers of sites should be synced relative to the total number of sites. However, if a tenant has many sites, but only a small number are configured for syncing, it may be more efficient to look up just the individually configured sites.

This PR adds an RCF to allow the user to pick which strategy to use, and implements looking up sites based on their path.

Additionally, this PR modifies the `validate_config` logic to check sites not by their `"name"` but by the path they use in their URL. This is more reliable and easier for customers to reason about. 

#### Pre-Review Checklist

## Related Pull Requests

* original POC: https://github.com/elastic/connectors-python/pull/1639/commits

## Release Note

Adds the ability for the Sharepoint Online connector to fetch sites individually, rather than by paging through all sites. This can be advantageous when syncing a relatively small number of sites.

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -785,15 +785,17 @@ class Connector(ESDocument):
             self.log_debug(f"Populated service type {configured_service_type}")
 
         simple_config = source_klass.get_simple_configuration()
+        current_config = self.configuration.to_dict()
+        missing_keys = simple_config.keys() - current_config.keys()
         if self.configuration.is_empty():
             # sets the defaults and the flag to NEEDS_CONFIGURATION
             doc["configuration"] = simple_config
             doc["status"] = Status.NEEDS_CONFIGURATION.value
             self.log_debug("Populated configuration")
-        elif simple_config.keys() - self.configuration.to_dict().keys():
-            missing_keys = simple_config.keys() - self.configuration.to_dict().keys()
-            self.log_warning(f"Detected an existing connector: {self.id} ({self.service_type}) that was previously {Status.CONNECTED.value} but is now missing configuration: {missing_keys}. Values for the new fields will be automatically set. Please review these configuration values as part of your upgrade.")
-            doc["configuration"] = {k:simple_config[k] for k in missing_keys}
+        elif missing_keys:
+            doc["configuration"] = self._updated_configuration(
+                missing_keys, current_config, simple_config
+            )
             # doc["status"] = Status.NEEDS_CONFIGURATION.value # not setting status, because it may be that default values are sufficient
 
         if self.features.features != source_klass.features():
@@ -810,6 +812,30 @@ class Connector(ESDocument):
             if_primary_term=self._primary_term,
         )
         await self.reload()
+
+    def updated_configuration(
+        self, missing_keys, current_config, simple_default_config
+    ):
+        self.log_warning(
+            f"Detected an existing connector: {self.id} ({self.service_type}) that was previously {Status.CONNECTED.value} but is now missing configuration: {missing_keys}. Values for the new fields will be automatically set. Please review these configuration values as part of your upgrade."
+        )
+        draft_config = {
+            k: simple_default_config[k] for k in missing_keys
+        }  # add missing configs as they exist in the simple default
+
+        # copy any differences (excluding differences in "value") to the draft
+        # the contents of simple_default_config are used unless they are missing
+        for config_name, config_obj in current_config.items():
+            for k, v in config_obj.items():
+                simple_default_value = simple_default_config.get(config_name, {}).get(k)
+                if k != "value" and simple_default_value != v:
+                    draft_config_obj = draft_config.get(config_name, {})
+                    if simple_default_value is None:
+                        draft_config_obj[k] = v
+                    else:
+                        draft_config_obj[k] = simple_default_value
+                    draft_config[config_name] = draft_config_obj
+        return draft_config
 
     @with_concurrency_control()
     async def validate_filtering(self, validator):
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -784,11 +784,17 @@ class Connector(ESDocument):
             doc["service_type"] = configured_service_type
             self.log_debug(f"Populated service type {configured_service_type}")
 
+        simple_config = source_klass.get_simple_configuration()
         if self.configuration.is_empty():
             # sets the defaults and the flag to NEEDS_CONFIGURATION
-            doc["configuration"] = source_klass.get_simple_configuration()
+            doc["configuration"] = simple_config
             doc["status"] = Status.NEEDS_CONFIGURATION.value
             self.log_debug("Populated configuration")
+        elif simple_config.keys() - self.configuration.to_dict().keys():
+            missing_keys = simple_config.keys() - self.configuration.to_dict().keys()
+            self.log_warning(f"Detected an existing connector: {self.id} ({self.service_type}) that was previously {Status.CONNECTED.value} but is now missing configuration: {missing_keys}. Values for the new fields will be automatically set. Please review these configuration values as part of your upgrade.")
+            doc["configuration"] = {k:simple_config[k] for k in missing_keys}
+            # doc["status"] = Status.NEEDS_CONFIGURATION.value # not setting status, because it may be that default values are sufficient
 
         if self.features.features != source_klass.features():
             doc["features"] = source_klass.features()
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -830,10 +830,7 @@ class Connector(ESDocument):
                 simple_default_value = simple_default_config.get(config_name, {}).get(k)
                 if k != "value" and simple_default_value != v:
                     draft_config_obj = draft_config.get(config_name, {})
-                    if simple_default_value is None:
-                        draft_config_obj[k] = v
-                    else:
-                        draft_config_obj[k] = simple_default_value
+                    draft_config_obj[k] = simple_default_value or v
                     draft_config[config_name] = draft_config_obj
         return draft_config
 
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -793,7 +793,7 @@ class Connector(ESDocument):
             doc["status"] = Status.NEEDS_CONFIGURATION.value
             self.log_debug("Populated configuration")
         elif missing_keys:
-            doc["configuration"] = self._updated_configuration(
+            doc["configuration"] = self.updated_configuration(
                 missing_keys, current_config, simple_config
             )
             # doc["status"] = Status.NEEDS_CONFIGURATION.value # not setting status, because it may be that default values are sufficient
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -1200,7 +1200,10 @@ class SharepointOnlineDataSource(BaseDataSource):
                 "value": 3,
                 "order": 8,
                 "validations": [{"type": "greater_than", "constraint": 0}],
-                "depends_on": [{"field": "fetch_subsites", "value": True}, {"field": "enumerate_all_sites", "value": False}],
+                "depends_on": [
+                    {"field": "fetch_subsites", "value": True},
+                    {"field": "enumerate_all_sites", "value": False},
+                ],
             },
             "use_text_extraction_service": {
                 "display": "toggle",
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -685,13 +685,13 @@ class SharepointOnlineClient:
             f"{GRAPH_API_URL}/sites/{sharepoint_host}:/sites/{allowed_site}"
         )
 
-    async def _fetch_subsites_by_parent_id(self, parent_site_id):
+    async def _scroll_subsites_by_parent_id(self, parent_site_id):
         self._logger.debug(f"Scrolling subsites of {parent_site_id}")
         async for page in self._graph_api_client.scroll(
             f"{GRAPH_API_URL}/sites/{parent_site_id}/sites?expand=sites"
         ):
             for site in page:
-                async for subsite in self._recurse_sites(site):
+                async for subsite in self._recurse_sites(site):  # pyright: ignore
                     yield subsite
 
     async def _recurse_sites(self, site_with_subsites):
@@ -699,7 +699,7 @@ class SharepointOnlineClient:
         site_with_subsites.pop("sites@odata.context", None)  # remove unnecesary field
         yield site_with_subsites
         if subsites:
-            async for site in self._fetch_subsites_by_parent_id(
+            async for site in self._scroll_subsites_by_parent_id(
                 site_with_subsites["id"]
             ):
                 yield site
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -694,7 +694,7 @@ class SharepointOnlineClient:
         site_with_subsites.pop("sites@odata.context", None)  # remove unnecesary field
         yield site_with_subsites
         for site in subsites:
-            async for item in self._recurse_sites(site):
+            async for item in self._recurse_sites(site):  # pyright: ignore
                 yield item
 
     def _subsite_expansion_str(self, depth):
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -590,17 +590,23 @@ class SharepointOnlineClient:
         except NotFound:
             return
 
-    async def sites(self, parent_site_id, allowed_root_sites):
+    async def sites(self, parent_site_id, allowed_root_sites, enumerate_all_sites=True):
         select = ""
 
-        if allowed_root_sites == [WILDCARD]:
+        if allowed_root_sites == [WILDCARD] or enumerate_all_sites:
             async for page in self._graph_api_client.scroll(
-                f"{GRAPH_API_URL}/sites/{parent_site_id}/sites?search=*&$select={select}" # TODO: why search?
+                f"{GRAPH_API_URL}/sites/{parent_site_id}/sites?search=*&$select={select}"
             ):
                 for site in page:
+                    # Filter out site collections that are not needed
+                    if (
+                            WILDCARD not in allowed_root_sites
+                            and site["name"] not in allowed_root_sites
+                    ):
+                        continue
                     yield site
         else:
-            self._logger.debug(f"Looking up sites: {allowed_root_sites}")
+            self._logger.debug(f"Looking up sites: {allowed_root_sites} individually")
             for allowed_site in allowed_root_sites:
                 try:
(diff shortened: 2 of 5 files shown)


seanstory:
This probably can't be merged as-is, since it seems to be backwards incompatible. We may need a migration to introduce the new config, or code changes to make the missing config be handled gracefully.

seanstory:
> This probably can't be merged as-is

Migration capability added. Still needs tests

connectors/sources/sharepoint_online.py
```diff
@@ -605,22 +605,83 @@ async def site_users(self, site_web_url):
...
                 yield site
 
+    async def _fetch_site_and_subsites(
+        self, parent_site_id, allowed_site, expand_clause
+    ):
+        self._logger.debug(
+            f"Requesting site '{allowed_site}' and subsites by relative path in parent site: {parent_site_id}"
+        )
+        site_with_subsites = await self._graph_api_client.fetch(
+            f"{GRAPH_API_URL}/sites/{parent_site_id}:/sites/{allowed_site}?{expand_clause}"
```
wangch079, line 658: WDYT if we [enumerate subsites](https://learn.microsoft.com/en-us/graph/api/site-list-subsites) separately? This will definitely increase the number of requests, but it does have some benefits:
1. Easier to understand
2. Less configuration (no need to config the depth)
3. Confirm all subsites are fetched, assuming there's limit in expand result.

connectors/protocol/connectors.py
```diff
@@ -784,11 +784,17 @@ async def prepare(self, config, sources):
...
             self.log_debug(f"Populated service type {configured_service_type}")
 
+        simple_config = source_klass.get_simple_configuration()
         if self.configuration.is_empty():
             # sets the defaults and the flag to NEEDS_CONFIGURATION
-            doc["configuration"] = source_klass.get_simple_configuration()
+            doc["configuration"] = simple_config
             doc["status"] = Status.NEEDS_CONFIGURATION.value
             self.log_debug("Populated configuration")
+        elif simple_config.keys() - self.configuration.to_dict().keys():
```
jedrazb, line 793: We repeat the expression `simple_config.keys() - self.configuration.to_dict().keys()` twice

connectors/protocol/connectors.py
```diff
@@ -784,11 +784,17 @@ async def prepare(self, config, sources):
...
+            missing_keys = simple_config.keys() - self.configuration.to_dict().keys()
+            self.log_warning(f"Detected an existing connector: {self.id} ({self.service_type}) that was previously {Status.CONNECTED.value} but is now missing configuration: {missing_keys}. Values for the new fields will be automatically set. Please review these configuration values as part of your upgrade.")
+            doc["configuration"] = {k:simple_config[k] for k in missing_keys}
```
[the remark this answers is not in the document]
jedrazb: What are the expected side effects of adding and changing order of RCFs in connector configuration (except for the errors mentioned in this PR description)? I will need to do something similar for google drive connector soon

i think we should also check for updates in "editable" fields like label, tooltip or order. I asked chatgpt to generate some code so that it's clear what I mean :D 

```python
simple_config = source_klass.get_default_configuration()

# If configuration is empty, set the default
if self.configuration.is_empty():
    doc["configuration"] = simple_config
    doc["status"] = Status.NEEDS_CONFIGURATION.value
    self.log_debug("Populated configuration")
else:
    current_config = self.configuration.to_dict()

    # Check for missing keys
    missing_keys = simple_config.keys() - current_config.keys()
    if missing_keys:
        self.log_warning(f"Detected an existing connector: {self.id} ({self.service_type}) that was previously {Status.CONNECTED.value} but is now missing configuration: {missing_keys}. Values for the new fields will be automatically set. Please review these configuration values as part of your upgrade.")
        doc["configuration"] = {k: simple_config[k] for k in missing_keys}

    # Check for changes in order, label, and tooltip properties
    for key, value in simple_config.items():
        if key in current_config:
            # List of properties we want to check
            properties = ["order", "label", "tooltip"]
            for prop in properties:
                if prop in value and (prop not in current_config[key] or value[prop] != current_config[key][prop]):
                    self.log_warning(f"Detected a change in '{prop}' for key '{key}' in connector: {self.id} ({self.service_type}). Updating it to the latest value.")
                    doc["configuration"][key][prop] = value[prop]
```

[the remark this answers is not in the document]
navarone-feekery: If the values of existing field properties are changed in a release (e.g. label, order, etc.), they aren't migrated over for existing users. This isn't breaking or anything. These values are only used to initialise the connector configuration, so it just means the user may have an outdated tooltip or the order of their configuration will be different.

Because it isn't breaking, I don't know if there needs to be a priority for including changes to update existing field properties.

Also a possible reason not to do this is the user may have changed tooltips or labels to suit themselves, we'd be effectively removing all freedom there each time we validate.

connectors/protocol/connectors.py
```diff
@@ -784,11 +784,17 @@ async def prepare(self, config, sources):
...
+        simple_config = source_klass.get_simple_configuration()
         if self.configuration.is_empty():
             # sets the defaults and the flag to NEEDS_CONFIGURATION
-            doc["configuration"] = source_klass.get_simple_configuration()
+            doc["configuration"] = simple_config
             doc["status"] = Status.NEEDS_CONFIGURATION.value
             self.log_debug("Populated configuration")
+        elif simple_config.keys() - self.configuration.to_dict().keys():
+            missing_keys = simple_config.keys() - self.configuration.to_dict().keys()
+            self.log_warning(f"Detected an existing connector: {self.id} ({self.service_type}) that was previously {Status.CONNECTED.value} but is now missing configuration: {missing_keys}. Values for the new fields will be automatically set. Please review these configuration values as part of your upgrade.")
```
navarone-feekery, line 795: How do you intend to get a default value for a new field? I recently did work to [remove not-useful default values from RCFs](https://github.com/elastic/connectors-python/pull/761), so they aren't typicallly available unless the field is a boolean.

connectors/protocol/connectors.py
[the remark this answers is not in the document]
navarone-feekery: To clarify on un-migrated behaviour: if two RCFs have the same `order` value, they will show next to each other in Kibana in random order, but always in the same place in relation to the other RCFs.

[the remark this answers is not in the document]
seanstory: I definitely hit this behavior when manually testing, and just forgot to look into why the ordering didn't work like I'd expected. :) 

I'm not opposed to including all updates/changes to configurations as part of this migration, as long as the cofniguration key names and the values stay the same as what the user had. I don't think we need to support a user customizing their tooltip/label directly through ES without having made a customized connector with a custom implementation of `get_default_configuration()`.

connectors/protocol/connectors.py
seanstory: Yep. I guess I can precompute it, though it wouldn't be used in the first `if`. Ugh. :P

connectors/protocol/connectors.py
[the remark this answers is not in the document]
seanstory: updated

connectors/protocol/connectors.py
seanstory: well in this case, these are booleans and one `int` where I added a default `value`.

We shouldn't be adding new RCFs that need non-trivial values, or that have initial values that will cause a behavior change from the previous version. Such RCFs would need a "real" migration, I think. In a discussion with @sphilipse and @danajuratoni , we felt that we still need a more comprehensive plan for migrations in the future, but that this small subset of "migrations" are something we could whip up pretty quickly and without a lot of risk.

seanstory: And I just removed the depth (int) config, so we're back to just booleans

connectors/sources/sharepoint_online.py
seanstory: Ok, I've implemented a recursion like this that doesn't limit depth, but also doesn't make useless requests at leaf nodes. WDYT?

connectors/protocol/connectors.py
```diff
@@ -805,6 +813,30 @@ async def prepare(self, config, sources):
...
+        # the contents of simple_default_config are used unless they are missing
+        for config_name, config_obj in current_config.items():
+            for k, v in config_obj.items():
+                simple_default_value = simple_default_config.get(config_name, {}).get(k)
+                if k != "value" and simple_default_value != v:
+                    draft_config_obj = draft_config.get(config_name, {})
+                    if simple_default_value is None:
+                        draft_config_obj[k] = v
+                    else:
+                        draft_config_obj[k] = simple_default_value
```
wangch079, line 836:
```suggestion
                    draft_config_obj[k] = simple_default_value or v
```

connectors/protocol/connectors.py
```diff
@@ -784,11 +784,19 @@ async def prepare(self, config, sources):
...
             # sets the defaults and the flag to NEEDS_CONFIGURATION
-            doc["configuration"] = source_klass.get_simple_configuration()
+            doc["configuration"] = simple_config
             doc["status"] = Status.NEEDS_CONFIGURATION.value
             self.log_debug("Populated configuration")
+        elif missing_keys:
+            doc["configuration"] = self.updated_configuration(
+                missing_keys, current_config, simple_config
+            )
+            # doc["status"] = Status.NEEDS_CONFIGURATION.value # not setting status, because it may be that default values are sufficient
```
wangch079, line 799: I do think we should set it to `NEEDS_CONFIGURATION`, which can prompt the user to update configuration. 

`NEEDS_CONFIGURATION` status can also stop new sync job from being scheduled:

https://github.com/elastic/connectors-python/blob/88a0abbad169dfc5f87ce890759a0835bffbaf5c/connectors/services/job_scheduling.py#L81-L85

connectors/sources/sharepoint_online.py
```diff
@@ -626,20 +626,82 @@ async def site_users(self, site_web_url):
...
-    async def sites(self, parent_site_id, allowed_root_sites):
-        select = ""
+    async def sites(
+        self,
+        sharepoint_host,
+        allowed_root_sites,
+        enumerate_all_sites=True,
+        fetch_subsites=False,
+    ):
+        if allowed_root_sites == [WILDCARD] or enumerate_all_sites:
```
wangch079, line 636: I'd suggest to make it consistent by checking

```python
WILDCARD in allowed_root_sites
```

This can work with the edge case when customers configure something like:
```
Sean'steamsite, *, Chenhui'scommunicationsite
```

connectors/sources/sharepoint_online.py
```diff
@@ -1223,21 +1302,33 @@ async def validate_config(self):
...
-                site_collection["siteCollection"]["hostname"], [WILDCARD]
+                site_collection["siteCollection"]["hostname"],
+                configured_root_sites,
+                self.configuration["enumerate_all_sites"],
             ):
-                remote_sites.append(site["name"])
+                if self.configuration["enumerate_all_sites"]:
+                    retrieved_sites.append(site["name"])
+                else:
+                    retrieved_sites.append(self._site_path_from_web_url(site["webUrl"]))
```
wangch079, line 1316: Why do we need this? Won't the `get site by path` API return `name` field?

connectors/sources/sharepoint_online.py
```diff
@@ -626,20 +626,82 @@ async def site_users(self, site_web_url):
...
+        allowed_root_sites,
+        enumerate_all_sites=True,
+        fetch_subsites=False,
+    ):
+        if allowed_root_sites == [WILDCARD] or enumerate_all_sites:
+            self._logger.debug(f"Looking up all sites to fetch: {allowed_root_sites}")
+            async for site in self._all_sites(sharepoint_host, allowed_root_sites):
+                yield site
+        else:
+            self._logger.debug(f"Looking up individual sites: {allowed_root_sites}")
```
wangch079, line 641: Probably you want to add a [get root site call](https://learn.microsoft.com/en-us/graph/api/site-get?view=graph-rest-1.0&tabs=http#get-the-tenants-root-site) in the else loop, assuming we always want to fetch root site.

Adding a new RCF `Fetch root site?` may be confusing. Imagine customers configured `*` in `site collections, and toggled off `Fetch root site?`, you need to exclude the root site in the above `if branch`.

connectors/sources/sharepoint_online.py
```diff
@@ -626,20 +626,82 @@ async def site_users(self, site_web_url):
...
-            f"{GRAPH_API_URL}/sites/{parent_site_id}/sites?search=*&$select={select}"
+            f"{GRAPH_API_URL}/sites/{sharepoint_host}/sites?search=*&$select={select}"
         ):
             for site in page:
                 # Filter out site collections that are not needed
-                if (
-                    WILDCARD not in allowed_root_sites
-                    and site["name"] not in allowed_root_sites
-                ):
+                if [WILDCARD] != allowed_root_sites and site[
```
jedrazb, line 664: following on Chenhui comment: should we check `if WILDCARD not in allowed_root_sites ...`

connectors/protocol/connectors.py
```diff
@@ -805,6 +813,30 @@ async def prepare(self, config, sources):
...
+        draft_config = {
+            k: simple_default_config[k] for k in missing_keys
+        }  # add missing configs as they exist in the simple default
+
+        # copy any differences (excluding differences in "value") to the draft
+        # the contents of simple_default_config are used unless they are missing
+        for config_name, config_obj in current_config.items():
+            for k, v in config_obj.items():
+                simple_default_value = simple_default_config.get(config_name, {}).get(k)
+                if k != "value" and simple_default_value != v:
```
jedrazb, line 831: nit: Feel free to ignore, the `"value"` feels a bit like a magic string. Should we instead add `class RichConfigurableField(enum) ... ` and have sth like: `RichConfigurableField.value`

connectors/protocol/connectors.py
seanstory: @danajuratoni, @sphilipse, and I discussed this option, and felt that it would be less interruptive for a customer to not be asked to re-configure. We want migrations to preserve current behavior, so as long as our default values on upgrade keep the same behavior as before the upgrade, there's no need to interrupt any scheduled syncs and force a user to manually configure anything.

connectors/sources/sharepoint_online.py
seanstory: I'd assumed that was a logical bug. Why would you configure `*` AND anything else? The star covers all other sites, right?

connectors/sources/sharepoint_online.py
seanstory: No, not always. This is actually the root cause of some issues we've been seeing, where `/sites/outer/foo` and `/sites/foo` might both be named `foo`, and then you might have a `/sites/bar` named `Bar` (uppercase) and a `/sites/baz` named `my-baz`. There are no rules, and this is the wild west, aparantly. 💀

connectors/sources/sharepoint_online.py
seanstory: `*` is currently the only way to fetch a root site. Using a comma-separated list today would not fetch the root site either. I don't think we need to address that edge case in this PR, to avoid scope creep.

connectors/protocol/connectors.py
navarone-feekery: Thinking over it, it's not _such_ a problem if there is no `value`. This migrator will add the new RCF with a type-safe empty value. The configuration won't be flagged as invalid, but if the field is required the validator will flag that the new field is empty and inform the user.

navarone-feekery: I do agree that migrations need to be worked out in the future, but with this change we can effectively migrate already, it's just unclear to the user that a migration happened.
In any case, my initial concern has been abated.

connectors/protocol/connectors.py
danajuratoni: @seanstory's comment is aligned with my expectations. If there is a possibility to make this non-blocking we should take that

navarone-feekery (approved):
I see other comments on this PR, but mine was addressed so here's my approval

tests/protocol/test_connectors.py
```diff
@@ -1974,3 +1974,75 @@ def test_pipeline_properties(key, value, default_value):
...
+            "sensitive": True,
+            "type": "str",
+        },
+        "some_toggle": {"label": "toggle", "order": 6, "type": "bool", "value": True},
+    }
+    missing_configs = ["new_config"]
+    connector = Connector(elastic_index=Mock(), doc_source={"_id": "test"})
+    result = connector.updated_configuration(missing_configs, current, simple_default)
+
+    # all keys included where there are changes (excludes 'tenant_name')
```
jedrazb, line 2032: One question, in protocol/connector.py we have:
```
doc["configuration"] = self.updated_configuration(
                missing_keys, current_config, simple_config
            )
```
(it might be because it's friday afternoon) I don't fully understand why the `tenant_name` is expected to be missing since it's included in current config and simple_default config in the test case

Would it result in the actual connector doc miss that field in the index?

connectors/protocol/connectors.py
seanstory: This is a valid nit, but I think is outside the scope of this PR to address. I filed us an issue to follow up: https://github.com/elastic/connectors-python/issues/1673

connectors/sources/sharepoint_online.py
seanstory: See https://elastic.slack.com/archives/C01795T48LQ/p1695396448194609?thread_ts=1695395634.669799&cid=C01795T48LQ

I'm going to resolve this for now, but we can open up a follow-up issue if you still have concerns.

connectors/sources/sharepoint_online.py
seanstory: I think we're in a good state of compromise for now. It can always be improved in a future effort.

connectors/sources/sharepoint_online.py
seanstory: Diana indicated in a [slack thread](https://elastic.slack.com/archives/C01795T48LQ/p1695390301379779?thread_ts=1695311571.172939&cid=C01795T48LQ) that this is a good candidate for a follow-up effort. I've filed: https://github.com/elastic/connectors-python/issues/1675

tests/protocol/test_connectors.py
seanstory: > I don't fully understand why the tenant_name is expected to be missing since it's included in current config and simple_default config in the test case

The two are identical in all fields, so there's no need to include it in the partial update request, since it wouldn't lead to a change.

> Would it result in the actual connector doc miss that field in the index?

I think this is the gap/misunderstanding. We're using the `_update` elasticsearch API here, which does partial updates. You actually can't remove objects/keys with this API, you can only add new keys and edit existing leaf values.

seanstory:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.10|[](https://github.com/elastic/connectors-python/pull/1676)|
|✅|pre-8.10-stable|[](https://github.com/elastic/connectors-python/pull/1677)|

### Questions ?

PR #1662, opened 2023-09-20
Retrieve individual sites when site collections are configured

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/5739

When Site Collections are configured, SPO connector will try to fetch the site by site name one by one.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -299,7 +299,7 @@ def retryable_aiohttp_call(retries):
                     async for item in func(*args, **kwargs, retry_count=retry):
                         yield item
                     break
-                except (NotFound, BadRequestError):
+                except (NotFound, BadRequestError, PermissionsMissing):
                     raise
                 except Exception:
                     if retry >= retries:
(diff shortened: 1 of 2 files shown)


connectors/sources/sharepoint_online.py
```diff
@@ -299,7 +299,7 @@ async def wrapped(*args, **kwargs):
                     async for item in func(*args, **kwargs, retry_count=retry):
                         yield item
                     break
-                except (NotFound, BadRequestError):
+                except (NotFound, BadRequestError, PermissionsMissing):
```
vidok, line 302: It seems like adding `PermissionsMissing` is not related to the current PR. Could you please expand on this?

vidok:
Should this also need to be backported to pre-8.10?

connectors/sources/sharepoint_online.py
wangch079: Correct, not related. Just feel we shouldn't retry if there's a PermissionsMissing error.

wangch079:
>Should this also need to be backported to pre-8.10?

I think 8.10 is already released?

connectors/sources/sharepoint_online.py
```diff
@@ -618,22 +618,24 @@ async def site_users(self, site_web_url):
...
-                # Filter out site collections that are not needed
-                if (
-                    WILDCARD not in allowed_root_sites
-                    and site["name"] not in allowed_root_sites
-                ):
-                    continue
-
                 yield site
 
+    async def site_by_site_name(self, parent_site_id, site_name):
```
seanstory, line 628: This is specifically not happening by "name" but by "relative path".

seanstory:
There's definitely overlap between https://github.com/elastic/connectors-python/pull/1653 and this, and that's on me, for not realizing that we were working on related stuff earlier. I started on mine during spacetime last week, and had intentionally not made noise about it then, but I should have been vocal come Monday. :(  Sorry!

Preferences on how to resolve the discrepancies?

wangch079:
Close this PR as this is a duplicate of https://github.com/elastic/connectors-python/pull/1653

PR #1668, opened 2023-09-22
Create a new content index if it doesn't exist 

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/5767

Thhis PR brings back the [original](https://github.com/elastic/connectors-python/commit/320d50453e86f83faf6e9470cecce0ecdd27927d#diff-349332898f86f6f4901c5d9533f1c5ec244f231f81ee092e5b14a68a780a18f1R38) functionality that creates content indices.

This handles a use-case when customers use Connectors without Kibana. 

#### Pre-Review Checklist


connectors/es/sink.py
```diff
@@ -660,29 +663,50 @@ async def prepare_content_index(self, index, *, mappings=None):
...
+
         if exists:
+            # Update the index mappings if needed
             self._logger.debug(f"{index} exists")
-            response = await self.client.indices.get_mapping(
-                index=index, expand_wildcards=expand_wildcards
+            await self._ensure_content_index_mappings(index, mappings, expand_wildcards)
+        else:
+            # Create a new index
+            self._logger.info("Attempt to create a new index")
```
wangch079, line 675: It will be helpful to log the index name as well.

connectors/es/sink.py
```diff
@@ -660,29 +663,50 @@ async def prepare_content_index(self, index, *, mappings=None):
...
-                    properties=mappings.get("properties", {}),
-                    expand_wildcards=expand_wildcards,
-                )
-                self._logger.debug("Index %s mappings added", index)
-            else:
-                self._logger.debug("Index %s already has mappings. Skipping...", index)
-            return
+            self._logger.info(f"Index {index} has been successfully created.")
+
+        return
```
wangch079, line 681: `return` is not necessary

connectors/es/sink.py
```diff
@@ -660,29 +663,50 @@ async def prepare_content_index(self, index, *, mappings=None):
...
+
+    async def _ensure_content_index_mappings(self, index, mappings, expand_wildcards):
```
wangch079, line 683: Why the `expand_wildcards` is needed here? I think we always use the exact index name instead of wildcard here?

connectors/es/sink.py
```diff
@@ -660,29 +663,50 @@ async def prepare_content_index(self, index, *, mappings=None):
...
+                index=index,
+                dynamic=mappings.get("dynamic", False),
+                dynamic_templates=mappings.get("dynamic_templates", []),
+                properties=mappings.get("properties", {}),
+                expand_wildcards=expand_wildcards,
+            )
+            self._logger.debug("Index %s mappings added", index)
         else:
-            raise IndexMissing(f"Index {index} does not exist!")
+            self._logger.debug("Index %s already has mappings. Skipping...", index)
```
wangch079, line 702: I see you mentioned
https://github.com/elastic/connectors-python/blob/6b700bcdbc80e0344dc4b809db9491657df0e3fa/connectors/es/sink.py#L670

Do you want to handle mapping changes here? In normal cases, the index should be created by either Kibana or Connector service. If we find a discrepancy, should we 1) create a new index with the correct mapping or 2) raise an error or 3) simply log a warning message and continue?

connectors/es/sink.py
vidok: I'm not aware of any use cases where `expand_wildcards` is required but this option has been used from the beginning. In this PR, I'm not changing the index mapping (only moved it to a separate method) so this is why `expand_wildcards` stays. 

This is a good question but not in this PR scope.

connectors/es/sink.py
vidok: Good question:

If the connector can't apply mappings to the index it should fail immediately. This can already happen, see this [issue](https://github.com/elastic/enterprise-search-team/issues/5765). 

I guess it would be nice to handle it and pass a human-readable error. Is this what you also meant?

connectors/es/sink.py
```diff
@@ -660,29 +663,48 @@ async def prepare_content_index(self, index, *, mappings=None):
...
+
         if exists:
+            # Update the index mappings if needed
             self._logger.debug(f"{index} exists")
-            response = await self.client.indices.get_mapping(
-                index=index, expand_wildcards=expand_wildcards
+            await self._ensure_content_index_mappings(index, mappings, expand_wildcards)
+        else:
+            # Create a new index
+            self._logger.info(f"Attempt to create {index} index")
```
artem-shelkovnikov, line 675:
```suggestion
            self._logger.info(f"Creating content index: {index}")
```

connectors/es/sink.py
```diff
@@ -660,29 +663,48 @@ async def prepare_content_index(self, index, *, mappings=None):
...
-                    dynamic=mappings.get("dynamic", False),
-                    dynamic_templates=mappings.get("dynamic_templates", []),
-                    properties=mappings.get("properties", {}),
-                    expand_wildcards=expand_wildcards,
-                )
-                self._logger.debug("Index %s mappings added", index)
-            else:
-                self._logger.debug("Index %s already has mappings. Skipping...", index)
-            return
+            self._logger.info(f"Index {index} has been successfully created.")
```
artem-shelkovnikov, line 679:
```suggestion
            self._logger.info(f"Contend index successfully created:  {index}")
```

connectors/es/sink.py
```diff
@@ -660,29 +663,48 @@ async def prepare_content_index(self, index, *, mappings=None):
...
+                index=index,
+                dynamic=mappings.get("dynamic", False),
+                dynamic_templates=mappings.get("dynamic_templates", []),
+                properties=mappings.get("properties", {}),
+                expand_wildcards=expand_wildcards,
+            )
+            self._logger.debug("Index %s mappings added", index)
         else:
-            raise IndexMissing(f"Index {index} does not exist!")
+            self._logger.debug("Index %s already has mappings. Skipping...", index)
```
artem-shelkovnikov, line 700:
```suggestion
            self._logger.debug("Index %s already has mappings, skipping mappings creation", index)
```

artem-shelkovnikov, line 700:
```suggestion
            self._logger.debug("Index %s already has mappings, skipping mappings creation", index)
```

connectors/es/sink.py
```diff
@@ -660,29 +663,48 @@ async def prepare_content_index(self, index, *, mappings=None):
...
+                "Index %s has no mappings or it's empty. Adding mappings...", index
+            )
+            await self.client.indices.put_mapping(
+                index=index,
+                dynamic=mappings.get("dynamic", False),
+                dynamic_templates=mappings.get("dynamic_templates", []),
+                properties=mappings.get("properties", {}),
+                expand_wildcards=expand_wildcards,
+            )
+            self._logger.debug("Index %s mappings added", index)
```
artem-shelkovnikov, line 698:
```suggestion
            self._logger.debug("Successfully added mappings for index %s", index)
```

artem-shelkovnikov (approved):
Looks good, left a couple suggestions for log lines.

I personally find log lines ending with "..." suitable only for lines that imply some lengthy process after, e.g:

```
Creating something...
Created!
```

artem-shelkovnikov (approved):

PR #1674, opened 2023-09-22
Add docker release docs

navarone-feekery:
Add documentation for releasing the docker image manually.


docs/RELEASING.md
```diff
@@ -27,6 +27,20 @@ After the Elastic unified release is complete
 
 - Update the **BUILD** version ([example PR](https://github.com/elastic/connectors-python/pull/122)). Note that the Connectors project does not immediately bump to the next **PATCH** version. That won't happen until that patch release's FF date.
 
+## Releasing docker images
+
+To release the docker image, follow these steps:
+
+1. Make sure you're on the right tagged release commit, e.g. `v8.10.2.0`
+   - If no release tag exists yet, use the release branch, e.g. `8.10`
```
seanstory, line 35:
```suggestion
   - If no release tag exists yet, create it with `git tag <tagid> && git push origin <tagid>`
```

seanstory (commented):
one nit, but looks good

docs/RELEASING.md
```diff
@@ -27,6 +27,20 @@ After the Elastic unified release is complete
...
+2. Make sure the version in [VERSION](../connectors/VERSION) is correct
+3. Run `docker login -u <username> -p <password> docker.elastic.co` with credentials that allow for release
+4. Edit the [Makefile](../Makefile) to remove `-SNAPSHOT` from the `docker-build`, `docker-run` and `docker-push` steps
```
wangch079, line 38: Should we add a `docker-release` target?

navarone-feekery: @wangch079 we aren't expecting to need to run this manually because it's automated, so I thought some instructions would be enough rather than a new makefile target.

navarone-feekery:
@seanstory @wangch079 could either of you give this an approval so I can merge it?

docs/RELEASING.md
```diff
@@ -27,6 +27,20 @@ After the Elastic unified release is complete
...
 - Update the **BUILD** version ([example PR](https://github.com/elastic/connectors-python/pull/122)). Note that the Connectors project does not immediately bump to the next **PATCH** version. That won't happen until that patch release's FF date.
 
+## Releasing docker images
+
+To release the docker image, follow these steps:
+
+1. Make sure you're on the right tagged release commit, e.g. `v8.10.2.0`
+   - If no release tag exists yet, create it with `git tag <tagid> && git push origin <tagid>`
+2. Make sure the version in [VERSION](../connectors/VERSION) is correct
+3. Run `docker login -u <username> -p <password> docker.elastic.co` with credentials that allow for release
```
artem-shelkovnikov, line 37: Is it just `allow releasing` instead of `allow for release`?

Asking as a non-native, who barely sees "allow for" 😅

artem-shelkovnikov (approved):

docs/RELEASING.md
artem-shelkovnikov: Is there a ticket for automation already?

docs/RELEASING.md
navarone-feekery: As a native speaker with a weird "it just works to me" understanding; I took a look to see why `allow for` felt right to me and found this post: https://english.stackexchange.com/questions/30069/whats-the-difference-between-allow-and-allow-for

So in this sentence, `that allow for releasing` is referring to `credentials` rather than the run command. Having the credentials does not automatically release anything, so `allows for` works.
The difference is very small, but to me `allow releasing` would work if it was referencing the command rather than the credentials, as running the command directly results in a release.

docs/RELEASING.md
navarone-feekery: @sphilipse do you know if automating this in 8.11 has a ticket?

docs/RELEASING.md
artem-shelkovnikov: Thanks for the explanation!

PR #1680, opened 2023-09-25
Fix issues of Microsoft Teams

akanshi-elastic:
## Related to [#1422](https://github.com/elastic/connectors-python/issues/1422)
This PR includes fixes for:
1. Adding value in `attached_documents` field for every attachment present in channel chat
2. For Channel tabs and Channel Attachments, added Channel name for better readability.
3. For team, creation time field is always null, so removed it from elastic field
4. For the fields having chat message as null, indexing can be skipped for those users.
5. For user chat tabs and user chat attachments, added name of members attached with it in elastic search.
6. change the API to fetch attachment for channels( reason: For private and shared channel attachments were not coming in api response)
7. Added support for fetching multiple records from same thread.
#### Pre-Review Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -252,3 +252,13 @@ steps:
                 - ".buildkite/run_nigthly.sh zoom small"
               artifact_paths:
                 - "perf8-report-*/**/*"
+          
+          - path: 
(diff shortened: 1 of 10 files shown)


navarone-feekery (approved):
LGTM

PR #1682, opened 2023-09-25
Navarone/5857 expand extraction service 1

navarone-feekery:

- Adds the data extraction service functionality to Azure Blob Storage connectors.
- Moves ContentExtraction initializing logic to BaseDataSource so other connectors can easily have extraction service added

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/azure_blob_storage.py
+++ b/connectors/sources/azure_blob_storage.py
@@ -101,6 +101,14 @@ class AzureBlobStorageDataSource(BaseDataSource):
                     {"type": "less_than", "constraint": MAX_CONCURRENT_DOWNLOADS + 1}
                 ],
             },
+            "use_text_extraction_service": {
+                "display": "toggle",
+                "label": "Use text extraction service",
+                "order": 6,
+                "tooltip": "Requires a separate deployment of the Elastic Text Extraction Service. Requires that pipeline settings disable text extraction.",
+                "type": "bool",
+                "value": False,
+            },
         }
 
     def _configure_connection_string(self):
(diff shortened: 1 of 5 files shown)


connectors/sources/azure_blob_storage.py
```diff
@@ -191,16 +200,34 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
+                else:
+                    self._logger.debug(f"Calling convert_to_b64 for file : {blob_name}")
+                    await asyncio.to_thread(convert_to_b64, source=temp_filename)
+                    async with aiofiles.open(
+                        file=temp_filename, mode="r"
+                    ) as async_buffer:
+                        # base64 on macOS will add a EOL, so we strip() here
+                        document["_attachment"] = (await async_buffer.read()).strip()
+            finally:
+                await remove(str(temp_filename))
```
artem-shelkovnikov, line 229: Is this possible to move up, so that each source does not have to do this?

Can be done later, probably together with making downloads in framework more generic

artem-shelkovnikov (approved):
Looks good, do we need a migration to add this field to existing clients' sources?

connectors/sources/azure_blob_storage.py
navarone-feekery: Good idea, I'll move it up

navarone-feekery:
> do we need a migration to add this field to existing clients' sources?

@artem-shelkovnikov we won't need to do this because @seanstory recently added a configuration field check. If a field is missing it will be added on connector startup. [Here's the PR with that change](https://github.com/elastic/connectors-python/pull/1653).

PR #1683, opened 2023-09-25
Forcibly cancel a sync job after waiting for 5 seconds

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/5810

This PR makes the changes so that
1. when the connector cancels the sync job, it will wait for (hardcoded) 5 seconds for cancelation to be done
2. If the sync job is not canceled successfully after 5 seconds, it will try to forcibly cancel the sync job, by removing the queue object, so the whole flow from the remote source to Elasticsearch is cut off.
3. The connector then continue to mark the sync job as `canceled`

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -201,7 +201,7 @@ class Sink:
             raise
         except ForceCanceledError:
             self._logger.error(
-                "Sink cannot be stopped after certain period of time, force cancel the task."
+                f"Sink did not stop within {CANCELATION_TIMEOUT} seconds of cancelation, force-canceling the task."
             )
 
     async def _run(self):
@@ -386,10 +386,10 @@ class Extractor:
             raise
         except ForceCanceledError:
             self._logger.error(
-                "Extractor cannot be stopped after certain period of time, force cancel the extractor."
+                f"Extractor did not stop within {CANCELATION_TIMEOUT} seconds of cancelation, force-canceling the task."
             )
         except Exception as e:
-            self._logger.critical("Document fetcher failed", exc_info=True)
+            self._logger.critical("Document extractor failed", exc_info=True)
             await self.put_doc("FETCH_ERROR")
             self.fetch_error = e
 
@@ -758,7 +758,7 @@ class SyncOrchestrator(ESClient):
                 return
 
         self._logger.error(
-            f"Couldn't stop the sync job after {CANCELATION_TIMEOUT} seconds, force canceling."
+            f"Sync job did not stop within {CANCELATION_TIMEOUT} seconds of canceling. Force-canceling."
         )
         self._sink.force_cancel()
         self._extractor.force_cancel()
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -57,6 +57,7 @@ OP_INDEX = "index"
 OP_UPSERT = "update"
 OP_DELETE = "delete"
 TIMESTAMP_FIELD = "_timestamp"
+CANCELATION_TIMEOUT = 5
 
 
 def get_mb_size(ob):
@@ -65,6 +66,10 @@ def get_mb_size(ob):
 
 
 class UnsupportedJobType(Exception):
+    pass
+
+
+class ForceCanceledError(Exception):
     pass
 
 
@@ -178,12 +183,26 @@ class Sink:
             f"Sink stats - no. of docs indexed: {self.indexed_document_count}, volume of docs indexed: {round(self.indexed_document_volume)} bytes, no. of docs deleted: {self.deleted_document_count}"
         )
 
+    def force_cancel(self):
+        # reset self.queue so that no document will flow to Elasticsearch index
+        self.queue = None
+
+    async def fetch_doc(self):
+        if self.queue:
+            return await self.queue.get()
+
+        raise ForceCanceledError
+
     async def run(self):
         try:
             await self._run()
         except asyncio.CancelledError:
             self._logger.info("Task is canceled, stop Sink...")
             raise
+        except ForceCanceledError:
+            self._logger.error(
+                "Sink cannot be stopped after certain period of time, force cancel the task."
+            )
 
     async def _run(self):
         """Creates batches of bulk calls given a queue of items.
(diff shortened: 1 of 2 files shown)


connectors/es/sink.py
```diff
@@ -352,6 +381,14 @@ async def run(self, generator, job_type):
         except asyncio.CancelledError:
             self._logger.info("Task is canceled, stop Extractor...")
             raise
+        except ForceCanceledError:
+            self._logger.error(
+                "Extractor cannot be stopped after certain period of time, force cancel the extractor."
+            )
+        except Exception as e:
+            self._logger.critical("Document fetcher failed", exc_info=True)
```
seanstory, line 389:
```suggestion
            self._logger.critical("Document extractor failed", exc_info=True)
```
Not related, but we can fix this while we're here.

connectors/es/sink.py
```diff
@@ -691,19 +711,33 @@ def done(self):
...
+            await asyncio.sleep(1)
+            cancelation_timeout -= 1
+            if not self._sink_task_running() and not self._extractor_task_running():
+                self._logger.info(
+                    "Both Extractor and Sink tasks are successfully stopped."
+                )
+                return
+
+        self._logger.error(
+            f"Couldn't stop the sync job after {CANCELATION_TIMEOUT} seconds, force canceling."
```
seanstory, line 737:
```suggestion
            f"Sync job did not stop within {CANCELATION_TIMEOUT} seconds of canceling. Force-canceling."
```

connectors/es/sink.py
```diff
@@ -54,6 +54,7 @@
 OP_UPSERT = "update"
 OP_DELETE = "delete"
 TIMESTAMP_FIELD = "_timestamp"
+CANCELATION_TIMEOUT = 5
```
seanstory, line 57: This feels like a good candidate to add to service configuration in our YAML file, so that it can be easily increased if necessary.

seanstory (commented):
Few comments, but seems like a simple approach.

I remember Tarek talking about putting `sleep(0)` calls in the code, and I found https://stackoverflow.com/a/68760162/2479282, I wonder if we should also just throw a few `sleep(0)` calls into the SPO connector? Looks like this bug may happen if we're actually not sending anything to ES because two full syncs are run one right after another, so there are no updates or deletes to execute?

Or maybe we should just add the `sleep(0)` calls in sync.py to the top of their loops, instead of the bottom, so no `continue` calls skip them? 

Or am I misunderstanding?

connectors/es/sink.py
```diff
@@ -175,12 +180,26 @@ def _populate_stats(self, stats, res):
...
 
+    def force_cancel(self):
+        # reset self.queue so that no document will flow to Elasticsearch index
+        self.queue = None
+
+    async def fetch_doc(self):
+        if self.queue:
+            return await self.queue.get()
+
+        raise ForceCanceledError
```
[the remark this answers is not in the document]
wangch079: > Why not have a specific variable like `self.canceled` or something?

That's a good idea 👍

connectors/es/sink.py
wangch079: The reason why I didn't do so:

1. Although flexibility is always good, I don't feel customers want to configure this.
2. We currently pass `elastic_config` to `SyncOrchestrator`, and apparently this is not part of `elastic_config`, and I'm hesitate to introduce a new param to `SyncOrchestrator`:
https://github.com/elastic/connectors-python/blob/0c69b9f5cf87d004f5f67422fd7b57953a287daf/connectors/es/sink.py#L629-L640

wangch079:
>I wonder if we should also just throw a few sleep(0) calls into the SPO connector?

`asyncio.sleep(0)` forces the current task to release control, and I think `await` does the same.

We do have `asyncio.sleep(0)` in `Extractor` class, after processing each document yielded:
https://github.com/elastic/connectors-python/blob/0c69b9f5cf87d004f5f67422fd7b57953a287daf/connectors/es/sink.py#L442

https://github.com/elastic/connectors-python/blob/0c69b9f5cf87d004f5f67422fd7b57953a287daf/connectors/es/sink.py#L513

https://github.com/elastic/connectors-python/blob/0c69b9f5cf87d004f5f67422fd7b57953a287daf/connectors/es/sink.py#L586

At the framework level, we can't guarantee the cancellation always succeed, so we must have a mechanism to forcibly cancel the sync job.

connectors/es/sink.py
```diff
@@ -175,12 +181,25 @@ def _populate_stats(self, stats, res):
...
+
     async def run(self):
         try:
             await self._run()
         except asyncio.CancelledError:
             self._logger.info("Task is canceled, stop Sink...")
             raise
+        except ForceCanceledError:
+            self._logger.error(
+                f"Sink did not stop within {CANCELATION_TIMEOUT} seconds of cancelation, force-canceling the task."
```
vidok, line 201:
```suggestion
                f"Sync job did not stop within {CANCELATION_TIMEOUT} seconds of cancelation, force-canceling the task."
```

navarone-feekery (approved):
LGTM

connectors/es/sink.py
seanstory: Ok, we can leave this for https://github.com/elastic/enterprise-search-team/issues/5757

connectors/es/sink.py
```diff
@@ -546,60 +573,53 @@ async def get_access_control_docs(self, generator):
...
-            return
+            await self.put_doc(
+                {
+                    "_op_type": operation,
+                    "_index": self.index,
+                    "_id": doc_id,
+                    "doc": doc,
+                }
+            )
+            await asyncio.sleep(0)
```
timgrein, line 614: Do we need `await asyncio.sleep(0)` here? We're already inside an async generator, which should yield control back to the event loop, if possible/needed, right?

wangch079: I'm not adding this line. This is from line 589 (see left).

seanstory:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.10|[](https://github.com/elastic/connectors-python/pull/1706)|
|✅|pre-8.10-stable|[](https://github.com/elastic/connectors-python/pull/1707)|

### Questions ?

PR #1685, opened 2023-09-26
Abstract download file extraction and conversion

navarone-feekery:

This PR moves the logic for downloading files to temp files, and their extraction/conversion to b64 to BaseDataSource.
The new tool is very generic, so it can be easily added to other connectors in future PRs.
If this is approved, I will also generalise/abstract the file size and file extension check.

This PR uses two connectors as examples for how this tool would be used:

1. Confluence

This is a "standard" download and extraction case. The download func uses `response.content.iter_chunked`, which is the same pattern that most connectors use. This download func is now abstracted onto the `BaseDataSource` class as `generic_chunked_download_func`.

1. Azure Blob Storage

ABS has an irregular download method (it uses a library and has its own chunking method). As a result, its download func can't be abstracted. Instead the download func is defined in the ABS connector and is passed as a partial arg.

#### Pre-Review Checklist

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -6,14 +6,19 @@
 """ Helpers to build sources + FQN-based Registry
 """
 
+import asyncio
 import importlib
 import re
+from contextlib import asynccontextmanager
 from datetime import date, datetime
 from decimal import Decimal
 from enum import Enum
 from functools import cache
 from pydoc import locate
 
+import aiofiles
+from aiofiles.os import remove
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov (approved):
Good stuff! Will you follow-up with using it in other sources, or we create an issue and address later?

navarone-feekery:
@artem-shelkovnikov Thanks! I'm intending to use this in all sources with file downloads, I'll be making some batch PRs for that now :)

Issue #1465, opened 2023-08-16
Scheduling is failing when poll interval = time to schedule a sync for new job


artem-shelkovnikov:
## Bug Description

Sometimes scheduler skips scheduling new jobs due to flaws in logic that checks whether a job can be scheduled.

Currently, scheduling service polls for connectors in connectors index every X seconds. Then, imaging the connector has a sync that has to be scheduled in Y seconds.

If X > Y, then a job is scheduled.

However, if X == Y (taking rounding in account), then no job is scheduled and connector waits for a next interval to schedule a job.

A good illustrative example is [this CI job](https://buildkite.com/elastic/connectors-python-nightly/builds/1035#0189fe20-2467-487c-90c0-22a40f6f0ccc):

```
[FMWK][11:39:00][DEBUG] [Connector id: mssql, index name: search-mssql] Populated features
--
  | [FMWK][11:39:00][DEBUG] [Connector id: mssql, index name: search-mssql] Sending heartbeat
  | [FMWK][11:39:01][DEBUG] [Connector id: mssql, index name: search-mssql] Status is Status.CONFIGURED
  | [FMWK][11:39:01][DEBUG] [Connector id: mssql, index name: search-mssql] Filtering is in state valid, skipping...
  | [FMWK][11:39:01][DEBUG] [Connector id: mssql, index name: search-mssql] Next 'full' sync due in 60 seconds
  | [FMWK][11:39:30][DEBUG] Polling every 30 seconds for Job Execution Service
  | [FMWK][11:39:31][DEBUG] Polling every 30 seconds for Job Scheduling Service
  | [FMWK][11:39:31][DEBUG] [Connector id: mssql, index name: search-mssql] Status is Status.CONFIGURED
  | [FMWK][11:39:31][DEBUG] [Connector id: mssql, index name: search-mssql] Filtering is in state valid, skipping...
  | [FMWK][11:39:31][DEBUG] [Connector id: mssql, index name: search-mssql] Next 'full' sync due in 30 seconds
  | [FMWK][11:40:00][DEBUG] Polling every 30 seconds for Job Execution Service
  | [FMWK][11:40:01][DEBUG] Polling every 30 seconds for Job Scheduling Service
  | [FMWK][11:40:01][DEBUG] [Connector id: mssql, index name: search-mssql] Status is Status.CONFIGURED
  | [FMWK][11:40:01][DEBUG] [Connector id: mssql, index name: search-mssql] Filtering is in state valid, skipping...
[...]

PR #1689, opened 2023-09-27
Fix unreliable job scheduling by making it look-behind instead of look-ahead

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors-python/issues/1465

Problem: we inconsistently define what is "now" when doing scheduling for jobs + we use look-ahead approach (schedule jobs NO LATER THAN THEY SHOULD RUN). Example:

Let's say, that now it's 11:59:29 PM and jobs trigger every 24 hours exactly at 12:00:00 PM. Our server wakes up every 30 seconds. 

So at the moment of 11:59:29 PM, the next job is in 31 seconds and should not be scheduled yet. However, we don't wake up in exactly 30 seconds. We wake up in 30 seconds + overhead (event loop lag, we execute code before sleeping too), so in reality between the scheduling checks there's always an interval of 30+X seconds, where X can be between 1 and 2 seconds in best cases. So if we wait for 31 seconds and wake up, the job will be skipped and never scheduled.

This PR changes the approach to be look-behind (schedule the job NO EARLIER THAN THEY SHOULD RUN). Example:

Let's say, that now it's 11:59:29 PM and jobs trigger every 24 hours exactly at 12:00:00 PM. Our server wakes up every 30 seconds. 

With look-behind approach, when the service wakes up it checks if any job should have been scheduled after it woke up last time. So if server wakes up in 30 seconds, it checks if any jobs should have been scheduled in exact last 30 seconds. Same if it wakes up with a lag - in 35, 40, 77 or any other number of seconds. It eliminates all potential problems of not scheduling the task because of setting the expectations incorrectly.

Additionally, IMO it makes sense to always use look-behind approach: it's always better that the sync starts _a bit later_ than _a bit earlier_ when doing capacity planning and scheduling for the syncs, especially if the sync process is multistep (e.g. before sync happens, operators of 3rd-party systems do some step, for example turn off the replication and wait).

#### Pre-Review Checklist

## Release Note

Fixed a problem with scheduling when jobs sometimes were not correctly scheduled.

Changed scheduling mechanism to be look-behind and schedule jobs no earlier than expected time as opposed to no later than expected time before.


artem-shelkovnikov:
Gonna do a couple CI runs in the meantime, but also test locally and think more about the problem

connectors/protocol/connectors.py
```diff
@@ -604,7 +604,7 @@ async def heartbeat(self, interval):
             self.log_debug("Sending heartbeat")
             await self.index.heartbeat(doc_id=self.id)
 
-    def next_sync(self, job_type):
+    def next_sync(self, job_type, now):
```
artem-shelkovnikov, line 607: This is a major change - before "now" was arbitrary - connector wakes up at 11:59:59, but when checks the schedule it's already 12:00:01 and the moment of scheduling is lost

artem-shelkovnikov:
TODO: still add tests, but wanted to open this PR for discussion

connectors/services/job_scheduling.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, config):
         self.source_list = config["sources"]
         self.connector_index = None
         self.sync_job_index = None
+        self.last_wake_up_time = datetime.utcnow()
```
artem-shelkovnikov, line 42: What this means, if that if server starts 1 second after something should be scheduled, it won't be scheduled. It's probably fine, but we can also do `datetime.utcnow() - timedelta(seconds=IDLING)` to still execute jobs that should have started IDLING seconds ago

connectors/services/job_scheduling.py
```diff
@@ -167,29 +170,40 @@ async def _run(self):
...
         return 0
 
-    async def _scheduled_sync(self, connector, job_type):
+    async def _try_schedule_sync(self, connector, job_type):
+        last_wake_up_time = self.last_wake_up_time
+        this_wake_up_time = datetime.utcnow()
+
+        print(
+            f"Last time woke up at {last_wake_up_time}, woke up at {this_wake_up_time}"
+        )
```
timgrein, line 179:
```suggestion
```

I guess this was used for local debugging?

connectors/services/job_scheduling.py
```diff
@@ -167,29 +170,40 @@ async def _run(self):
...
+            ):
                 connector.log_debug(
                     f"A scheduled '{job_type_value}' sync is created by another connector instance, skipping..."
                 )
                 return False
 
             try:
-                next_sync = connector.next_sync(job_type)
+                next_sync = connector.next_sync(job_type, last_wake_up_time)
+                print(f"Next sync is at {next_sync}")
```
timgrein, line 206:
```suggestion
```

Probably also a debug statement? Or should we change this to use `logger`?

connectors/services/job_scheduling.py
```diff
@@ -167,29 +170,40 @@ async def _run(self):
                 await self.sync_job_index.close()
         return 0
 
-    async def _scheduled_sync(self, connector, job_type):
+    async def _try_schedule_sync(self, connector, job_type):
+        last_wake_up_time = self.last_wake_up_time
```
timgrein, line 174: Question: Do we need this assignment? Is the rationale to shorten the name throughout the method by removing the `self` prefix?

timgrein (commented):
Looks good from a functional perspective. Left some comments around two `print` statements and a variable assignment.

connectors/services/job_scheduling.py
artem-shelkovnikov: Yup, need to change it to logger.debug!

connectors/services/job_scheduling.py
artem-shelkovnikov: Yup, good call

connectors/services/job_scheduling.py
artem-shelkovnikov: It's just personal taste - since it's used several times, I just put it to variable and use down the code. So purely personal style, not opposed to inlining it.

seanstory:
> IMO it makes sense to always use look-behind approach: it's always better that the sync starts a bit later than a bit earlier when doing capacity planning and scheduling for the syncs

Agree, and I like this approach/idea

timgrein (approved):
LGTM, nice change 👏

PR #1693, opened 2023-09-27
Expand extraction service to more connectors #2

navarone-feekery:

- Expand extraction service to more connectors:
  - Jira
  - Dropbox
  - S3
- Create utils func for retrieving file extension
- Abstract file size and file extension pre-download check
- Fix some tests

Extraction service will be expanded to more connectors in further PRs

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/connectors-python/pull/1685


connectors/source.py
```diff
@@ -684,6 +689,32 @@ def is_premium():
         """
         return False
 
+    def get_file_extension(self, filename):
+        return get_file_extension(filename)
```
artem-shelkovnikov, line 693: Looks redundant to me, why not use utils method directly?

connectors/source.py
```diff
@@ -684,6 +689,32 @@ def is_premium():
...
+
+    def can_file_be_downloaded(self, file_extension, filename, file_size):
```
artem-shelkovnikov, line 695: Would it be better to pass the file name (or path?) and calculate extension from it?

I see that now `filename` is used only for debug lines, so method signature feels a bit weird from this perspective.

connectors/source.py
```diff
@@ -684,6 +689,32 @@ def is_premium():
...
+            )
+            return False
+
+        if file_extension.lower() not in TIKA_SUPPORTED_FILETYPES:
+            self._logger.debug(
+                f"Files with the extension {file_extension} are not supported, skipping {filename}."
+            )
+            return False
+
+        if file_size > FILE_SIZE_LIMIT and not self.configuration.get(
```
artem-shelkovnikov, line 708: In some cases (Sharepoint REST API) we don't know file sizes, it would be nice to have default handling for it (e.g. not passing file_size always downloads it or something similar)

connectors/sources/azure_blob_storage.py
```diff
@@ -165,34 +162,28 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
+        if not self.can_file_be_downloaded(file_extension, filename, file_size):
             return
 
-        if blob_size > DEFAULT_FILE_SIZE_LIMIT:
+        if blob["tier"] == "Archive":
             self._logger.warning(
-                f"File size {blob_size} of file {blob_name} is larger than {DEFAULT_FILE_SIZE_LIMIT} bytes. Discarding the file content"
+                f"{filename} can't be downloaded as blob tier is archive"
             )
             return
```
artem-shelkovnikov, line 179: This can probably go upper on line 168 - that's just more from aesthetic side :)

artem-shelkovnikov (dismissed):

connectors/source.py
navarone-feekery: I didn't want to call the import on every source. AFAIU I would need to import it multiple times. This way it's imported once on `BaseDataSource` and I can just call back up to get to the utils func.

I agree it looks redundant. Do you think it's better coding standards to just import and call on each source directly?

connectors/source.py
navarone-feekery: I actually have a draft PR that will allow that! A connector will be able to call the file type or the file size check individually if needed.
https://github.com/elastic/connectors-python/pull/1694

It's dependent on this PR though so I haven't opened it for review yet.

connectors/source.py
navarone-feekery: The file extension is needed for other operations (specifically creating a tempfile with the correct suffix), so I extract earlier and pass it on to this function. Also, some connectors get their file extensions from other sources. For example Salesforce responses have a separate field just for the file extension, and it isn't included in the file name.

> I see that now filename is used only for debug lines, so method signature feels a bit weird from this perspective.

Is there something else we can do to fix this?

navarone-feekery: Would it help aesthetically if `filename` was the last arg?

connectors/sources/azure_blob_storage.py
navarone-feekery: Aesthetics are good, I'll move this

connectors/source.py
artem-shelkovnikov: I'm on the fence here, I think it's okay to keep it and then inline if not needed.

I just feel that a method that pipes arguments to another method without any modifications is a code-smell, but IMO not a big one and okay to keep.

connectors/source.py
artem-shelkovnikov: It could be that the method returns an object instead of just True/False, e.g.:

```
if file_extension == "":
    return FileDownloadValidationResult(False, SomeEnum.NoExtension) 
```

Names of files and choice of enum/string with reason is up to discussion here, but not relying on `self._logger` here is a nice bonus IMO.

However, it might be a change that can be done only in future when almost the whole download function is in the base source, so my comment can be ignored for now.

artem-shelkovnikov (approved):

connectors/source.py
navarone-feekery: I think an alternative is to not have the file extension tool in utils. It's only used in BaseDataSource anyway. In my next PR I'll just move the code to this function so it's not piping anymore.

PR #1694, opened 2023-09-27
Expand extraction service to more connectors #3

navarone-feekery:

- Expand extraction service to more connectors:
  - Google Cloud Storage
  - Google Drive
  - Onedrive
  - Salesforce
  - Servicenow

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/connectors-python/pull/1685
- https://github.com/elastic/connectors-python/pull/1693

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -724,7 +724,12 @@ class BaseDataSource:
         return True
 
     async def download_and_extract_file(
-        self, doc, source_filename, file_extension, download_func, return_doc_if_failed=False
+        self,
+        doc,
+        source_filename,
+        file_extension,
+        download_func,
+        return_doc_if_failed=False,
     ):
         """
         Performs all the steps required for handling binary content:
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -760,7 +760,8 @@ class BaseDataSource:
             return doc
         except Exception as e:
             self._logger.warning(
-                f"File download and extraction or conversion for file {source_filename} failed: {e}", exc_info=True
+                f"File download and extraction or conversion for file {source_filename} failed: {e}",
+                exc_info=True,
             )
             if return_doc_if_failed:
                 return doc
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -724,14 +724,24 @@ class BaseDataSource:
         return True
 
     async def download_and_extract_file(
-        self, doc, source_filename, file_extension, download_func
+        self, doc, source_filename, file_extension, download_func, return_doc_if_failed=False
     ):
-        # 1 create tempfile
+        """
+        Performs all the steps required for handling binary content:
+        1. Make temp file
+        2. Download content to temp file
+        3. Extract using local service or convert to b64
+
+        Will return the doc with either `_attachment` or `body` added.
+        Returns `None` if any step fails.
+
+        If the optional arg `return_doc_if_failed` is `True`,
+        will return the original doc upon failure
+        """
         try:
             async with self.create_temp_file(file_extension) as async_buffer:
                 temp_filename = async_buffer.name
 
-                # 2 download to tempfile
                 await self.download_to_temp_file(
                     temp_filename,
                     source_filename,
@@ -739,7 +749,6 @@ class BaseDataSource:
                     download_func,
                 )
 
-                # 3 extract or convert content
                 doc = await self.handle_file_content_extraction(
                     doc, source_filename, temp_filename
                 )
@@ -748,7 +757,10 @@ class BaseDataSource:
             self._logger.warning(
                 f"File download and extraction or conversion for file {source_filename} failed: {e}"
             )
-            return
+            if return_doc_if_failed:
+                return doc
+            else:
+                return
 
     @asynccontextmanager
     async def create_temp_file(self, file_extension):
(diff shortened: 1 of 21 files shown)


connectors/source.py
```diff
@@ -693,6 +693,11 @@ def get_file_extension(self, filename):
         return get_file_extension(filename)
 
     def can_file_be_downloaded(self, file_extension, filename, file_size):
+        return self.is_valid_file_type(
+            file_extension, filename
+        ) and self.is_file_size_within_limit(file_size, filename)
```
seanstory, line 698: I thought one of the point for the extraction service is to remove this file size limitation?

connectors/source.py
```diff
@@ -716,26 +724,48 @@ def can_file_be_downloaded(self, file_extension, filename, file_size):
...
+                    download_func,
+                )
+
+                doc = await self.handle_file_content_extraction(
+                    doc, source_filename, temp_filename
+                )
+            return doc
+        except Exception as e:
+            self._logger.warning(
+                f"File download and extraction or conversion for file {source_filename} failed: {e}"
```
seanstory, line 763: let's add `exc_info` to this, so we get a stack trace from `e`

connectors/sources/google_cloud_storage.py
```diff
@@ -370,50 +375,40 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
+        file_extension = self.get_file_extension(filename)
+        if not self.can_file_be_downloaded(file_extension, filename, file_size):
             return
 
-        if blob_size > DEFAULT_FILE_SIZE_LIMIT:
-            self._logger.warning(
-                f"File size {blob_size} of file {blob_name} is larger than {DEFAULT_FILE_SIZE_LIMIT} bytes. Discarding the file content"
-            )
-            return
-        self._logger.debug(f"Downloading {blob_name}")
```
seanstory, line 387: can you move this logging and other valuable logging to the base class?

connectors/sources/google_cloud_storage.py
```diff
@@ -370,50 +375,40 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
         document = {
             "_id": blob["id"],
             "_timestamp": blob["_timestamp"],
         }
-        source_file_name = ""
-        async with NamedTemporaryFile(mode="wb", delete=False) as async_buffer:
+
+        # gcs has a unique download method so we can't utilize
+        # the generic download_and_extract_file func
```
seanstory, line 393: Useful comment 👍 

Is there anything we could generify, so that we don't have usages of `create_temp_file` in multiple places? Like would it help if the base function could take a proc as an optional arg or something? Not for this PR, but something we can think about.

connectors/source.py
navarone-feekery: In the func `is_file_size_within_limit` we are checking if the user has `use_text_extraction_service` enabled or not. If they have it enabled, we don't care about file size.

This pre-check is a generic gateway before any file downloading.

connectors/sources/salesforce.py
```diff
@@ -1201,7 +1163,6 @@ def map_content_document(self, content_document):
 
         return {
             "_id": content_document.get("Id"),
-            "_attachment": content_document.get("_attachment"),
```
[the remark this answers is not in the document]
navarone-feekery: This will be fine. Salesforce is the only connector where the flow changed to support the new download funcs.

Before:
Download file -> map to elastic doc -> return doc

Now:
map to elastic doc -> download file -> return doc

The `_attachment` (or `body` if local extraction is used`) is added to the doc during the "download file" phase.

connectors/sources/google_cloud_storage.py
navarone-feekery: The func `download_to_temp_file` has debug logs identical to this. So this exists on the base class now.

connectors/source.py
navarone-feekery: Added! Didn't know about this, I naively thought it would happen automatically.

connectors/sources/google_cloud_storage.py
seanstory: My concern with calling `create_temp_file` is that every time it's called, there's a chance that the author does it in such a way that the temp file won't be cleaned up. We've had this issue before, where like 8/10 connectors were cleaning up their tempfiles appropriately, but due to copy-paste errors, occasionally some wouldn't. These types of bugs can be hard to catch, and its easier to keep them from propagating if you just don't have numerous usages of the risky code. 

But again, not necessary to solve right now.

seanstory (approved):

connectors/sources/google_cloud_storage.py
navarone-feekery: @seanstory I was also concerned about that. It may alleviate your concerns but I _think_ I have this covered with the way tempfiles are being created now. If a connector uses `create_temp_file` it will clean itself up after everything is done, including deleting the file and outputting an error if the file deletion failed.

The code in question: https://github.com/elastic/connectors-python/blob/b9fe37744bd9724b3b4b82104f0c124d70bf3b02/connectors/source.py#L771-L783

Of course we should properly check to see if this is actually the case.

Issue #1721, opened 2023-10-03
[SPO] incremental sync isn't incremental except for drive items


seanstory:
## Bug Description
The expectation is that Incremental Syncs should fetch only recent changes, and should be significantly faster than full syncs. However, it looks like the SPO connector's incremental sync only fetches drive items incrementally, but is still fetching all Sites, Site Drives, Site Lists, Site List Items, and Site Pages. This means that while incremental syncs are faster, they're not _as_ fast as expected, and are redundantly re-fetching unmodified objects.

### To Reproduce
1. run a full sync
2. run an incremental sync
3. Notice the incremental sync was almost as long as the full sync

## Expected behavior
Incremental syncs should be very limited in what they retrieve - only pulling back documents that have been changed since the last sync was run.

## Environment
8.11.0-SNAPSHOT

Issue #1664, opened 2023-09-20
[SPO] Incremental Syncs might not be picking up changes in roleassignments


jedrazb:
## Bug Description

The `roleassignments` endpoint allows us to capture access controls. This endpoint provides `sitegroups` with the associated users in a single response, meaning we no longer need to track the tricky relations between sitegroups (with non-global id) and users during the access control sync. Instead, we can directly include all users (from those groups) in the document ACL.

However, this approach makes the access control updates tricky. If a new user is added or removed from a site group, we must wait for the content sync to detect this change in ACLs.

There's uncertainty about whether a change in the site's groups  would cause the site (or other object types) to appear in the delta API. If it doesn't, then changes in access controls (e.g., adding or removing a user from a site group) might not be detected promptly by the incremental syncs.

### To Reproduce
1. Run full sync
2. Turn on incremental syncs
3. Update sitegroup membership
4. Check ACL for site got picked up by the incremental sync

## Expected behavior
1. Ideally ACL should be updated in incremental sync as delta API should flag a change

seanstory:
I think this bug is wider than just incremental syncs, as I'm seeing that in full syncs we do not update documents with the access_control values are changed if the document doesn't record that it has been changed. In this case we never send the updated ACL to Elasticsearch. I think this relates to https://github.com/elastic/connectors-python/issues/1372

seanstory:
One option we have is to, instead of expanding group memberships during the full/incremental sync, is to instead expand the other way during the ACL sync. This would make sure we caught group membership changes more quickly, even if it doesn't help us with catching document sharing changes.

We originally thought this wasn't possible, since the graph API's `transitiveMemberOf` API didn't include SPO Site Groups. However, we've found that you can request a user's site groups with the Sharepoint REST API. Response from a Microsoft Engineer:

```
Using username

{siteUrl}/_api/web/siteusers(@v)?@v=’username’)&$expand=groups

Example /_api/web/siteusers(@v)[?@v='i%3A0%23.f%7Cmembership%7Cuser2%40sspxtesting.onmicrosoft.com'&$expand=groups](mailto:?@v=%27i%3A0%23.f%7Cmembership%7Cuser2%40sspxtesting.onmicrosoft.com%27&$expand=groups)

@v needs to be encoded. Its value in SPO will be ’i:0#.f|membership|[username@tenantname.onmicrosoft.com](mailto:username@tenantname.onmicrosoft.com)’.
```

[...]

PR #1731, opened 2023-10-04
Truncate branches and leafs in incremental sync

seanstory:
## Closes https://github.com/elastic/connectors-python/issues/1721

Adds conditionals for Sites, Site Drives, Site Lists, Site List Items, and Site Pages to be used in incremental syncs. These conditionals check if the object has a modified date that is greater than the stored cursor date (new with this PR). The object is not yielded if it is not more recent than the stored date.

This has pretty significant performance improvements in initial testing, probably because it allows us to skip any processing and network traffic that we might otherwise do to enrich these objects (like DLS lookups for each List Item).

Initial experiments on our test instance show an incremental sync (with no changes, with DLS enabled) going from taking 18 min to 4 seconds.

#### Pre-Review Checklist
- [ x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

## Release Note

Improves performance of Sharepoint Online incremental syncs by skipping Sites, Site Drives, Site Lists, Site List Items, and Site Pages that have a `lastModifiedDateTime` or `Modified` field that is older than the most recent sync's start time.


timgrein (commented):
Nice improvement 👏  This is basically key-based incremental replication, right?

connectors/sources/sharepoint_online.py
```diff
@@ -2182,6 +2224,18 @@ def update_drive_delta_link(self, drive_id, link):
...
+        return self._sync_cursor.get(CURSOR_SYNC_TIMESTAMP, self.epoch_timestamp())
+
+    def current_timestamp(self):
+        return datetime.now().strftime(TIMESTAMP_FORMAT)
+
+    def epoch_timestamp(self):
+        return strftime(TIMESTAMP_FORMAT, time.gmtime(0))
+
+    def update_sync_timestamp_cursor(self, timestamp):
+        self._sync_cursor[CURSOR_SYNC_TIMESTAMP] = timestamp
```
timgrein, line 2237: These methods look like good candidates for moving them to the framework (outside of the PR scope, but probably worth creating an issue).

seanstory:
> This is basically key-based incremental replication, right?

Kinda. In the perfect world, we'd query SPO based on that key. Like:
```
sites = get_sites_modified_since(timestamp)
for site in sites:
  yield site
```

instead of 
```
sites = get_all_sites()
for site in sites:
  if site.modified_date > timestamp
    yield site
```

but the end result _should_ be the same.

connectors/sources/sharepoint_online.py
```diff
@@ -1750,38 +1758,49 @@ async def get_docs_incrementally(self, sync_cursor, filtering=None):
...
+                async for site_page in self.site_pages(
+                    site, site_access_control, check_timestamp=True
+                ):
                     # Always include site admins in site page access controls
                     site_page = self._decorate_with_access_control(
                         site_page, site_admin_access_control
                     )
                     yield site_page, None, OP_INDEX
 
+        self.update_sync_timestamp_cursor(timestamp)
```
vidok, line 1770: Should we also update the `timestamp_cursor` when a full sync job is done?

seanstory: I'm actually updating it when a full sync starts. This is also updating the cursor to the value at the _start_ of the sync, but I have to wait till the end to update it, so that I don't throw things off.

connectors/sources/sharepoint_online.py
```diff
@@ -1750,38 +1756,50 @@ async def get_docs_incrementally(self, sync_cursor, filtering=None):
...
-            site["object_type"] = "site"
+            if not check_timestamp or (
+                check_timestamp
+                and site["lastModifiedDateTime"] >= self.last_sync_time()
+            ):
+                site["_id"] = site["id"]
+                site["object_type"] = "site"
 
-            yield site
+                yield site
```
artem-shelkovnikov, line 1791: Just to double check - if an object is uploaded to Site's Drive, the root Site's modified date is not changed.

Because of it, the approach here will ignore resyncing the Site Drive?

seanstory: > if an object is uploaded to Site's Drive, the root Site's modified date is not changed.

It actually _does_ change the root Site's modified date. I tested this with drive items, lists, list items, and site page edits. So for everything we care about, the change date propagates up. List item changes update the List's modified date, too.

The only case where nesting updates don't seem to apply is subsite changes don't modify site dates. However, since we're enumerating sites and subsites together in the same generator, this isn't problematic - I can exclude a top-level site but still enumerate the subsite.

connectors/utils.py
```diff
@@ -73,13 +74,25 @@
...
         when = datetime.now(timezone.utc)
     return when.isoformat()
 
 
+def iso_zulu():
+    """Returns the current time in ISO Zulu format"""
+    return datetime.now(timezone.utc).strftime(ISO_ZULU_TIMESTAMP_FORMAT)
+
+
+def epoch_timestamp():
```
timgrein, line 91:
```suggestion
def epoch_timestamp_zulu():
```

nit: Just for consistency purposes as the method above also specifies `zulu` explicitly in its method name.

timgrein (approved):
LGTM, added one small nit, feel free to skip it or just ping me, if you decide to add it, so I can re-approve :)

connectors/sources/sharepoint_online.py
artem-shelkovnikov: What about permission changes? Changes in dynamic groups?

I have a weird feeling, that it's possible that there will be weird edge-cases that will be problematic :/

timgrein (approved):

connectors/sources/sharepoint_online.py
vidok: Yeah, you are updating it when `sync_cursor` is empty. I'm thinking of a use case when:

1.  you run a full sync -> sync_cursor is initialized with the timestamp
2. you run a couple of incremental syncs -> sync cursor is updated as well
3. you run a full sync again -> sync cursor is not updated
4. you run an incremental sync and it uses the timestamp from the step 2 not the step 3

seanstory:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.10|[](https://github.com/elastic/connectors-python/pull/1751)|
|✅|pre-8.10-stable|[](https://github.com/elastic/connectors-python/pull/1752)|

### Questions ?

connectors/sources/sharepoint_online.py
seanstory: Sharing changes also resulted in changing the `lastModifiedDateTime`. Changes to dynamic groups won't. But today in a full sync, those wouldn't get picked up either, since we'd detect that the `_id` and `_ts` have not changed. It's something we need to address in : https://github.com/elastic/connectors-python/issues/1664

Issue #1715, opened 2023-10-02
[SPO] Site Collection Admins not in ACLs


seanstory:
## Bug Description

When looking at site permissions, you can have "Site Onwers" (which is a special group for each site) but you can also have Site Collection Admins who do _not_ have to be included in the Owners group. These Admins also seem to not be included in the Role Assignments API, and therefore we're not including them in the ACLs for Sites and Site Pages, despite the fact that they should have read access to these documents.

### To Reproduce
1. Create a site with minimal users who can see the site
2. run a DLS sync and verify that a given user cannot search that site (UI or Elasticsearch)
3. Add that user as a Site Collection Administrator
4. run a DLS sync again
5. They still cannot search the site through Elasticsearch, even though they can access it through the SPO UI

## Expected behavior
Any user that can access content through the SPO UI should be able to search it in Elasticsearch

## Screenshots
<img width="1147" alt="Screenshot 2023-10-02 at 1 16 00 PM" src="https://github.com/elastic/connectors-python/assets/5288246/31ba68e2-645a-4f68-8abc-b41e991e8e49">
<img width="903" alt="Screenshot 2023-10-02 at 1 16 50 PM" src="https://github.com/elastic/connectors-python/assets/5288246/bfdc30e5-faf1-4664-a1b5-712af52dd5b4">

## Environment
8.11.0-SNAPSHOT

## Additional context
slack thread: https://elastic.slack.com/archives/C01795T48LQ/p1696268325300329

seanstory:
Microsoft support had this to say:

> we can use `/_api/web/SiteUsers?$filter=isSiteAdmin eq true` or `/_api/web/getuserbyid` and check isSiteAdmin property. We don’t recommend directly querying user info list.

seanstory:
See https://github.com/elastic/connectors-python/pull/1758#issuecomment-1758069219 - there's a chance that a customer may ask for this to be backported to `pre-8.10-stable`, but as they have not yet confirmed this, we'll go ahead and close with it only being addressed for `8.12`.

PR #1758, opened 2023-10-09
fetch site admins separate from role assignments

seanstory:
## Closes https://github.com/elastic/connectors-python/issues/1715

This fetches site collection admins with a `GET /_api/web/SiteUsers?$filter=isSiteAdmin eq true`, and appends the results to the ACLS fetched from role assignments that had `isSiteAdmin: true`. This ensures that _all_ the site admins, not just the ones in the site's "Owners" group, get stored and granted access through DLS.

Additionally, this PR ensures that `_o`-suffixed group IDS get expanded into the _owners_ of the group. This happens when a site's "members" includes a group, that group's owners automatically get added to the site's "owners" group.

#### Pre-Review Checklist
  - needs confirmation from stakeholders that it does or does not need backporting.

## Release Note

Ensures that all Site Collection Administrators are present in Site, Site Page, Site List, Site List Item, and Drive Item access control lists for Document Level Security on the Sharepoint Online connector.

### DIFF
--- a/connectors/sources/sharepoint_online.py
+++ b/connectors/sources/sharepoint_online.py
@@ -615,6 +615,19 @@ class SharepointOnlineClient:
             self._logger.debug(f"No role assignments found for site: '{site_web_url}'")
             return
 
+    async def site_admins(self, site_web_url):
+        self._validate_sharepoint_rest_url(site_web_url)
+        filter = url_encode("isSiteAdmin eq true")
+        url = f"{site_web_url}/_api/web/SiteUsers?$filter={filter}"
+
+        try:
+            async for page in self._rest_api_client.scroll(url):
+                for member in page:
+                    yield member
+        except NotFound:
+            self._logger.debug(f"No site admins found for site: '${site_web_url}'")
+            return
+
     async def site_groups_users(self, site_web_url, site_group_id):
         self._validate_sharepoint_rest_url(site_web_url)
 
@@ -656,7 +669,8 @@ class SharepointOnlineClient:
             return
 
     async def group_owners(self, group_id):
-        url = f"{GRAPH_API_URL}/groups/{group_id}/owners"
+        select = "id,mail,userPrincipalName"
+        url = f"{GRAPH_API_URL}/groups/{group_id}/owners?$select={select}"
 
         try:
             async for page in self._graph_api_client.scroll(url):
@@ -1365,6 +1379,9 @@ class SharepointOnlineDataSource(BaseDataSource):
 
             access_control |= member_access_control
 
+        async for member in self.client.site_admins(site["webUrl"]):
+            site_admins_access_control.update(await self._access_control_for_member(member))
+
         return list(access_control), list(site_admins_access_control)
 
     def _dls_enabled(self):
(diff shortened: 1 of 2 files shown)


connectors/sources/sharepoint_online.py
```diff
@@ -656,7 +669,8 @@ async def group_members(self, group_id):
             return
 
     async def group_owners(self, group_id):
-        url = f"{GRAPH_API_URL}/groups/{group_id}/owners"
+        select = "id,mail,userPrincipalName"
```
jedrazb, line 672: Nice optimisation 👍

connectors/sources/sharepoint_online.py
```diff
@@ -1365,6 +1379,11 @@ def _is_site_admin(user):
 
             access_control |= member_access_control
 
+        async for member in self.client.site_admins(site["webUrl"]):
```
jedrazb, line 1382: Reading the code above, it can be a bit confusing to the dev that we construct site_admins_access_control by iterating over 2 loops - both of them instinctively should return the same set of admins (I know they don't in reality).

Can you add a comment explaining that briefly for non-SPO experts 😄  ?

connectors/sources/sharepoint_online.py
```diff
@@ -2457,4 +2481,18 @@ def _access_control_for_user(self, user):
...
+
         return user_access_control
+
+    async def _access_control_for_group_id(self, group_id):
+        if group_id.endswith("_o"):
+            # this isn't the group id, it's a reference to the _owners_ of this ID
+            real_group_id = group_id[0:-2]
+            access_control = []
+            async for owner in self.client.group_owners(real_group_id):
+                access_control.extend(await self._access_control_for_member(owner))
```
jedrazb, line 2495: Just raising this for visibility: `_access_control_for_group_id` can be invoked by `_access_control_for_member` which in turns can call `_access_control_for_group_id` - in certain edge case it can lead to infinite recursion, if: 
- Group A is owned by Group B.
- Group B is owned by Group A.

This doesn't sound possible but .. it's SPO, so would it be worth adding a check to completely eliminate the possibility of infinite recursion?

connectors/sources/sharepoint_online.py
```diff
@@ -2457,4 +2481,18 @@ def _access_control_for_user(self, user):
         if email:
             user_access_control.append(_prefix_email(email))
 
+        if user_id:
+            user_access_control.append(_prefix_user_id(user_id))
+
         return user_access_control
+
+    async def _access_control_for_group_id(self, group_id):
+        if group_id.endswith("_o"):
```
jedrazb, line 2490: `group_id.endswith("_o")` it looks like a candidate for a separate function with a meaningful name and a comment explaining the logic behind the check

connectors/sources/sharepoint_online.py
seanstory: > This doesn't sound possible but .. it's SPO

This is good justification. :)

seanstory: I got a solution all coded up to prevent infinite recursion, and then while writing a test, realized that it wasn't necessary, because the `/groups/<group_id>/owners` API [can only return Users](https://learn.microsoft.com/en-us/graph/api/group-list-owners?view=graph-rest-1.0&tabs=http#response), and not the generic "members" (which can be groups OR users). This let me swap the call to `_access_control_for_member` to `_access_control_for_user` which breaks the potential cycle of infinite recursion. 👍

jedrazb (approved):
LGTM 🚀

seanstory:
At this time, it is not clear if this fix is required as a backport for any customers, so we are intentionally merging only to `main` (8.12). If a customer expresses a severe need for this in an earlier version, we should separately evaluate the risk/reward tradeoff before backporting, as this is not strictly fixing a bug, but adding an enhancement, and we typically do not backport enhancements. The risks here are minimal, but do include the addition of a new API call (SPO REST API, which is not ideal for rate limiting).

CC @DianaJourdan

PR #1778, opened 2023-10-13
[Outlook] Add a new field for `max concurrent tasks`

moxarth-elastic:
This PR is adding a new RCF called `Maximum concurrent tasks` to update the fetchers value. The default value for the field is 2000.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/outlook.py
+++ b/connectors/sources/outlook.py
@@ -767,7 +767,7 @@ class OutlookDataSource(BaseDataSource):
             "use_text_extraction_service": {
                 "display": "toggle",
                 "label": "Use text extraction service",
-                "order": 7,
+                "order": 12,
                 "tooltip": "Requires a separate deployment of the Elastic Text Extraction Service. Requires that pipeline settings disable text extraction.",
                 "type": "bool",
                 "ui_restrictions": ["advanced"],
@@ -777,7 +777,8 @@ class OutlookDataSource(BaseDataSource):
                 "default_value": MAX_CONCURRENCY,
                 "display": "numeric",
                 "label": "Maximum concurrent tasks",
-                "order": 8,
+                "tooltip": "This value denotes the number of tasks that run in parallel. It depends on the number of accounts in the Azure AD.",
+                "order": 13,
                 "required": False,
                 "type": "int",
             },
--- a/connectors/sources/outlook.py
+++ b/connectors/sources/outlook.py
@@ -45,7 +45,7 @@ RETRIES = 3
 RETRY_INTERVAL = 2
 
 QUEUE_MEM_SIZE = 5 * 1024 * 1024  # Size in Megabytes
-MAX_CONCURRENCY = 10
+MAX_CONCURRENCY = 2000
 
 OUTLOOK_SERVER = "outlook_server"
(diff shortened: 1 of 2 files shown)


connectors/sources/outlook.py
```diff
@@ -771,6 +773,14 @@ def get_default_configuration(cls):
                 "ui_restrictions": ["advanced"],
                 "value": False,
             },
+            "max_concurrent_tasks": {
+                "default_value": MAX_CONCURRENCY,
+                "display": "numeric",
+                "label": "Maximum concurrent tasks",
+                "order": 8,
```
wangch079, line 780: There is already `"order": 8`, and it may not be displayed as expected when `Outlook Server` is selected.

connectors/sources/outlook.py
```diff
@@ -771,6 +773,14 @@ def get_default_configuration(cls):
...
+                "required": False,
+                "type": "int",
```
wangch079, line 782: Consider adding `tooltips` to guide users on how to set a reasonable value.

connectors/sources/outlook.py
```diff
@@ -664,7 +664,9 @@ def __init__(self, configuration):
 
         self.tasks = 0
         self.queue = MemQueue(maxmemsize=QUEUE_MEM_SIZE, refresh_timeout=120)
-        self.fetchers = ConcurrentTasks(max_concurrency=MAX_CONCURRENCY)
+        self.fetchers = ConcurrentTasks(
+            max_concurrency=self.configuration["max_concurrent_tasks"]
```
wangch079, line 668: The new field is with `required: False`, so it can be empty even with `default_value`.

WDYT if we make it mandatory?

connectors/sources/outlook.py
moxarth-elastic: Oh i see. Actually the above config field `use_text_extraction_service` was at 7th order so i thought this can be at 8th. I'll update the order for both of them.

connectors/sources/outlook.py
moxarth-elastic: > The new field is with required: False, so it can be empty even with default_value.

@wangch079  If user keep it empty, default value will be treated as max concurrent tasks. And I'd suggest to keep it optional.

danajuratoni:
This field looks like an advanced config field. ("ui_restrictions": ["advanced"])
What would be a reasonable value to be used if the connector would run e.g. as native connector?

moxarth-elastic:
> This field looks like an advanced config field. ("ui_restrictions": ["advanced"]) What would be a reasonable value to be used if the connector would run e.g. as native connector?

@wangch079 suggested to keep it a normal config field.

> What would be a reasonable value to be used if the connector would run e.g. as native connector?

In that case, use the default value of 2000.

wangch079 (approved):

wangch079:
>@wangch079 suggested to keep it a normal config field.

Hi @moxarth-elastic , I suggested making it a basic config because I expect users to configure a reasonable value.

As we discussed it's hard to find an appropriate value for this field, I agree with @danajuratoni that we can make it an advanced field, and use the default 2000 if it's not configured.

PR #1785, opened 2023-10-13
[DEVELOPING] Point to config on main branch instead of 8.6 branch

timgrein:
Two links inside `DEVELOPING.md` pointed to the `8.6` config and config docs. It points to the latest config and config docs on `main` instead.

#### Pre-Review Checklist


docs/DEVELOPING.md
```diff
@@ -26,7 +26,7 @@
 
 ## General Configuration
 
-The details of Elastic instance and other relevant fields such as `service` and `source` needs to be provided in the [config.yml](https://github.com/elastic/connectors-python/blob/8.6/config.yml) file. For more details check out the following [documentation](https://github.com/elastic/connectors-python/blob/8.6/docs/CONFIG.md).
+The details of Elastic instance and other relevant fields such as `service` and `source` needs to be provided in the [config.yml](https://github.com/elastic/connectors-python/blob/master/config.yml) file. For more details check out the following [documentation](https://github.com/elastic/connectors-python/blob/master/docs/CONFIG.md).
```
seanstory, line 29:
```suggestion
The details of Elastic instance and other relevant fields such as `service` and `source` needs to be provided in the [config.yml](https://github.com/elastic/connectors-python/blob/main/config.yml) file. For more details check out the following [documentation](https://github.com/elastic/connectors-python/blob/main/docs/CONFIG.md).
```

Looks like this redirects, so probably not a big deal, but `main` is correct.

timgrein: Right, good catch!

seanstory (dismissed):

wangch079 (commented):
Is it better to use relative path?

timgrein:
> Is it better to use relative path?

I think we're using mainly absolute paths in the docs. I think it makes sense to always point at `main` as this is the tested version. If you're on a different branch a relative path wouldn't lead to `main` AFAIU. WDYT?

wangch079:
> If you're on a different branch a relative path wouldn't lead to `main` AFAIU. WDYT?

It will lead to the docs in the current branch, and I think this is expected? Like in our official docs, if you choose a non-current version (e.g. `8.9`), and you click around, you always stay in this version.

artem-shelkovnikov (approved):

timgrein:
> > If you're on a different branch a relative path wouldn't lead to `main` AFAIU. WDYT?
> 
> It will lead to the docs in the current branch, and I think this is expected? Like in our official docs, if you choose a non-current version (e.g. `8.9`), and you click around, you always stay in this version.

I think we need to revisit this then for all paths in our internal repo docs. I'll bring this up to the team and check with the docs-team, what we should do.

PR #1800, opened 2023-10-17
Use VersionData endpoint to download content

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/5912

In Salesforce, when a file is uploaded, a [ContentDocument](https://developer.salesforce.com/docs/atlas.en-us.244.0.object_reference.meta/object_reference/sforce_api_objects_contentdocument.htm) object is created, and it links to multiple [ContentVersion](https://developer.salesforce.com/docs/atlas.en-us.244.0.object_reference.meta/object_reference/sforce_api_objects_contentversion.htm) objects, which keeps all the history of the file.

The `ContentDocument` has a `LatestPublishedVersionId` field linking to the current version of the file, and we used `VersionDataURL` of the `ContentVersion` to download the file:
>The URL used to fetch a file from the binary data endpoint. This field is only populated on direct queries to ContentVersion, and not when queried through a related entity’s foreign key to ContentVersion.

The URL works in browsers but it will only download a js file when used in REST API call.

This PR switches to use [sObject Blob Get](https://developer.salesforce.com/docs/atlas.en-us.244.0.api_rest.meta/api_rest/resources_sobject_blob_retrieve.htm) to download files.

```
/services/data/vXX.X/sobjects/<sObject>/<id>/<blobField>
```

It tries to get `VersionData` of `ContentVersion`object:
```
/services/data/vXX.X/sobjects/ContentVersion/<content_version_id>/VersionData
```

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -33,9 +33,8 @@ TOKEN_ENDPOINT = "/services/oauth2/token"
 QUERY_ENDPOINT = f"/services/data/{API_VERSION}/query"
 DESCRIBE_ENDPOINT = f"/services/data/{API_VERSION}/sobjects"
 DESCRIBE_SOBJECT_ENDPOINT = f"/services/data/{API_VERSION}/sobjects/<sobject>/describe"
-DOWNLOAD_ENDPOINT = (
-    f"/services/data/{API_VERSION}/sobjects/ContentVersion/<file_id>/VersionData"
-)
+# https://developer.salesforce.com/docs/atlas.en-us.244.0.api_rest.meta/api_rest/resources_sobject_blob_retrieve.htm
+CONTENT_VERSION_DOWNLOAD_ENDPOINT = f"/services/data/{API_VERSION}/sobjects/ContentVersion/<content_version_id>/VersionData"
 
 RELEVANT_SOBJECTS = [
     "Account",
@@ -444,8 +443,10 @@ class SalesforceClient:
             params=params,
         )
 
-    async def _download(self, file_id):
-        endpoint = DOWNLOAD_ENDPOINT.replace("<file_id>", file_id)
+    async def _download(self, content_version_id):
+        endpoint = CONTENT_VERSION_DOWNLOAD_ENDPOINT.replace(
+            "<content_version_id>", content_version_id
+        )
         response = await self._get(f"{self.base_url}{endpoint}")
         yield response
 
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov:
@wangch079 can you add the information about why the endpoint was wrong and why new endpoint is the one to use?

wangch079:
>can you add the information about why the endpoint was wrong and why new endpoint is the one to use?

Updated PR description

connectors/sources/salesforce.py
```diff
@@ -33,6 +33,8 @@
 QUERY_ENDPOINT = f"/services/data/{API_VERSION}/query"
 DESCRIBE_ENDPOINT = f"/services/data/{API_VERSION}/sobjects"
 DESCRIBE_SOBJECT_ENDPOINT = f"/services/data/{API_VERSION}/sobjects/<sobject>/describe"
+# https://developer.salesforce.com/docs/atlas.en-us.244.0.api_rest.meta/api_rest/resources_sobject_blob_retrieve.htm
```
timgrein, line 36:
```suggestion
# https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_sobject_blob_retrieve.htm
```

I guess we want to always point to the latest documentation

timgrein (dismissed):
LGTM. Only one nit, that the comment referencing the salesforce docs points to a specific version and not latest.

connectors/sources/salesforce.py
wangch079: I did this on purpose, because we hardcoded it to `v58.0`:

https://github.com/elastic/connectors/blob/56ad4513b3591f725452749c52e0ed5ca2c2e48c/connectors/sources/salesforce.py#L31

artem-shelkovnikov: v58 is also arbitrary, maybe we can bump it too?

It's IMO much worse to use 2 apis with different versions vs using newer API

wangch079: I can bump the version.

Actually, there is a [List Available REST API versions](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_versions.htm) API, but I think it may be overkill, and I also feel scary if we allow it to bump by itself.

wangch079: I bumped the version in https://github.com/elastic/connectors/pull/1800/commits/f018d551672c15dda81fe3182ae61a374816fe40, and then reverted it in https://github.com/elastic/connectors/pull/1800/commits/7596d654acc7bf766a6a695c1cba93fe08b0fd74.

Because there are many hardcoded `v58.0` in the test file, I think it's worth a separate PR. Drafting...

wangch079: A separate PR: https://github.com/elastic/connectors/pull/1807

timgrein (approved):

PR #1802, opened 2023-10-17
[Filtering] Deduplicate semantic duplicate errors

timgrein:
## Closes https://github.com/elastic/enterprise-search-team/issues/6062

The `FilterValidationError` API allows for multiple `ids`. I've changed the API of `SyncRuleValidationResult` to allow to also reference multiple ids as one result could correlate to multiple sync rules. I've adapted the constructor of `SyncRuleValidationResult` that it can still work with a single value (but now also with a list).

This enables the duplication of the semantic duplicate errors messages:

**Note:** _The ids in the box heading require a Kibana change. This change adapts the message inside the box details._

**Before** (two times the same error just flipped): 
image

**After**:
image

#### Pre-Review Checklist

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -14,7 +14,14 @@
 ##      - connector_id: changeme # the ID of the connector.
 ##        service_type: changeme # The service type of the connector.
 ##        api_key: changeme # The Elasticsearch API key used to write data into the content index.
-#connectors: []
+connectors:
+-
+  connector_id: "BdDmZYsBMwlGH8qyGO3r"
+  service_type: "mongodb"
+  api_key: "QjlEbVpZc0JNd2xHSDhxeUx1MVc6aXBubndhc1dScC1lbWVYTndZbWFjZw=="
+elasticsearch:
+  host: "http://localhost:9200"
+  api_key: "QjlEbVpZc0JNd2xHSDhxeUx1MVc6aXBubndhc1dScC1lbWVYTndZbWFjZw=="
 #
 #
 ##  The ID of the connector.
@@ -126,7 +133,7 @@
 #
 #
 ##  The interval (in seconds) to poll connectors from Elasticsearch.
-#service.idling: 30
+service.idling: 5
 #
 #
 ##  The interval (in seconds) to send a new heartbeat for a connector.
--- a/connectors/filtering/validation.py
+++ b/connectors/filtering/validation.py
@@ -10,6 +10,7 @@ import fastjsonschema
 
 from connectors.filtering.basic_rule import BasicRule, Policy, Rule
 from connectors.logger import logger
+from connectors.utils import Format
 
 
 class ValidationTarget(Enum):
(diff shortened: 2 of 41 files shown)


connectors/filtering/validation.py
```diff
@@ -28,15 +29,21 @@ class SyncRuleValidationResult:
...
     ADVANCED_RULES = "advanced_snippet"
 
-    def __init__(self, rule_id, is_valid, validation_message):
-        self.rule_id = rule_id
+    def __init__(self, rule_ids, is_valid, validation_message):
+        # `str` and `bytes` are also iterable
+        if not isinstance(rule_ids, Iterable) or isinstance(rule_ids, (str, bytes)):
+            rule_ids = [rule_ids]
+        else:
+            rule_ids = list(rule_ids)
```
timgrein, line 37: I think this is okish for now, as we sometimes use the `rule_ids` keyword explicitly and sometimes we omit it. So this makes sure that everything is working without scanning through every usage of `SyncRuleValidationResult`. Maybe we should revisit using type hints again.

artem-shelkovnikov:
Visually the change still looks non-actionable, is there some follow-up planned for this change?

artem-shelkovnikov:
Ah I see, there's now one message instead of two, which is nice. We can address the readability of it separately, I guess?

timgrein:
> Ah I see, there's now one message instead of two, which is nice. We can address the readability of it separately, I guess?

Yes exactly. The heading is set within Kibana, that's a separate change. The message can be adjusted. WDYT about `Basic rule 1 (id: xyz) is semantically equivalent to Basic rule 2 (id: abc). Please delete one of those rules.`?

artem-shelkovnikov:
> WDYT about Basic rule 1 (id: xyz) is semantically equivalent to Basic rule 2 (id: abc). Please delete one of those rules.?

I think ids are not very useful - they are not exposed in UI at all. Is it possible to instead specify the rule - e.g. `The rule #1 is semantically equal to rule #4` and have #1 and #4 displayed in UI somehow (like the first column, for example)

config.yml
```diff
@@ -14,7 +14,14 @@
...
 ##        api_key: changeme # The Elasticsearch API key used to write data into the content index.
-#connectors: []
+connectors:
+-
+  connector_id: "BdDmZYsBMwlGH8qyGO3r"
+  service_type: "mongodb"
+  api_key: "QjlEbVpZc0JNd2xHSDhxeUx1MVc6aXBubndhc1dScC1lbWVYTndZbWFjZw=="
+elasticsearch:
+  host: "http://localhost:9200"
+  api_key: "QjlEbVpZc0JNd2xHSDhxeUx1MVc6aXBubndhc1dScC1lbWVYTndZbWFjZw=="
```
seanstory, line 24: I expect you didn't mean to to commit this

artem-shelkovnikov:
@timgrein do you still want to pursue these changes?

seanstory:
Bump @timgrein can you either close this or get it cleaned up and ready for re-review?

timgrein:
Can be closed, there's probably a better solution for this issue

PR #1812, opened 2023-10-18
Rename connectors-python to connectors in pr pipeline

wangch079:
This PR renames the PR pipeline from `connectors-python` to `connectors` because it couldn't reflect the pipeline status:

image

#### Pre-Review Checklist


timgrein:
That's the same change as https://github.com/elastic/connectors/pull/1811, right?

PR #1817, opened 2023-10-18
[Linter] Add ASYNC rules and fix issues

timgrein:
I skimmed through the `ruff` docs and saw that there are `ASYNC` rules:

- No blocking http calls
- No blocking sleep, open or subprocess calls
- No blocking os method calls

I think these can be pretty useful to avoid blocking the event loop. We already had examples in the past, where we used blocking `os` calls.

Also fixed the issues the linter found. Added `aiofiles` as requirement to enable local async file operations.

#### Pre-Review Checklist


artem-shelkovnikov (approved):
Good stuff!

PR #1823, opened 2023-10-19
Improve PostgreSQL query for identifying primary keys

acrewdson:
According to the [postgres wiki](https://wiki.postgresql.org/wiki/Retrieve_primary_key_columns), this is the optimal query to use for getting a table's primary key.

The previous query we were using had a limitation insofar as: a user with restricted permissions (_eg_ only the `pg_read_all_data` role) can't see data in `information_schema.table_constraints`, and thus cannot determine the primary key for a table.

The new query being used here does not appear to have that limitation.

-----

Related to the above: we are also updating the Postgres functional tests to configure the connector with a user that has reduced (read-only) permissions. Thanks @artem-shelkovnikov for the suggestion.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/postgresql.py
+++ b/connectors/sources/postgresql.py
@@ -53,7 +53,11 @@ class PostgreSQLQueries(Queries):
 
     def table_primary_key(self, **kwargs):
         """Query to get the primary key"""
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov (dismissed):

artem-shelkovnikov:
@acrewdson one thing that we can potentially improve is also change our functional test to not use superuser to access the DB.

Happy to help you with it, functional test itself is [here](https://github.com/elastic/connectors/blob/main/tests/sources/fixtures/postgresql/fixture.py) and the run result for your PR is [here](https://buildkite.com/elastic/connectors/builds/10188#018b4a10-80cf-40d5-bf45-46bf8caee64f)

artem-shelkovnikov (approved):
Looks great, thanks for the fix!

acrewdson:
@artem-shelkovnikov looks like the backport failed above since there is no `8.12` branch. Should I have added labels for other versions to be backported to?

artem-shelkovnikov:
@acrewdson 8.12 is current main, so backport skipped it.

I think it's a good improvement to backport as far as possible, I see that we don't have plans for releases up to 8.11, so we should backport only there.

PR #1830, opened 2023-10-23
Fix configuration migrations to work for peripheral connectors

navarone-feekery:
This PR is to address a bug with RCF validations/migrations.

When a connector is missing a configuration field, it should be added automatically when it is validated. However, this doesn't work for connectors that are not the "main" connector (defined by id in by config file). This bug most often appears for native connectors, but can also occur for self-managed connectors.

The code here was complicated and I think prone to bugs, so the fix also required a refactor. Any connector being sent through the `prepare` function should now have its configuration validated for both fields and field properties. Connectors are not treated too differently, so they can all follow the same logic path.

This fix:

- Combines the field and field property validations into one section
- Removes the pre-emptive doc update for field property validations (there is only one doc update call now)
- Ensures all connectors follow the same logic path to keep things DRY
- Updates existing tests to check if validation works for connectors that are not the "main" connector (set in the config file)


sphilipse (approved):
Approving because this generally looks good to me and we want to get this into today's BC.

connectors/protocol/connectors.py
```diff
@@ -813,7 +792,41 @@ async def prepare(self, config, sources):
...
+        if self.configuration.is_empty():
+            # sets the defaults and the flag to NEEDS_CONFIGURATION
+            self.log_debug("Populated configuration")
+            return {
+                "configuration": simple_config,
+                "status": Status.NEEDS_CONFIGURATION.value,
+            }
+
+        missing_fields = simple_config.keys() - current_config.keys()
+        fields_missing_properties = filter_nested_dict_by_keys(
```
wangch079, line 808: I think we can avoid checking both `missing_fields` and `fields_missing_properties`:

Instead of checking
```python
for config_name, config_obj in current_config.items():
    for k, v in config_obj.items():
        simple_default_value = simple_default_config.get(config_name, {}).get(k)
```

We can check the default config first:
```python
for config_name, config_obj in simple_default_config.items():
    for k, v in config_obj.items():
        current_value = current_config.get(config_name, {}).get(k)
```

This will make sure it covers both missing fields and missing properties

PR #1836, opened 2023-10-24
Add flake8-comprehentions(C4) ruff rule

artem-shelkovnikov:
Adding new linter rule: https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4

This rule helps with simplifying the code working with comprehentions. All the changes in the code are done with `make autoformat` and remove redundant usage of `list`, `set` and such.


timgrein (approved):
👏

PR #1837, opened 2023-10-24
Add T10 and T20 rules + fix problems with them

artem-shelkovnikov:
Adds 3 linting rules and fixes or ignores places that are violating it:

- T100 is usage of debugger statements
- T201 is usage of `print`
- T203 is usage of `pprint`

I personally often forget `print` statements and adding this rule will help me a lot :)


timgrein (approved):
Nice! I really like that we're leveraging the linter more and more 🚀

timgrein (approved):

PR #1840, opened 2023-10-25
Break JobExecutionSerivce to ContentSyncJobExecutionService and AccessControlSyncJobExecutionService

wangch079:
## Part of https://github.com/elastic/enterprise-search-team/issues/5193

This PR separates the JobExecutionSerivce to
1. `ContentSyncJobExecutionService`, which only handles the content sync job execution, i.e. full and incremental sync job
2. `AccessControlSyncJobExecutionService`, which only handles the access control sync job execution.

Further refactor can be done after the `ConcurrentTasks` object is moved out of the loop.

#### Pre-Review Checklist


artem-shelkovnikov (commented):
It's a pretty big change, but I don't quite understand the logic behind this on a bigger scale.

Additionally, there seems to be copy-paste between the services that is not trivial to review/understand.

Can you add the info on why it's done and what's the bigger plan behind this, what other changes will be there and etc?

For now it's hard to review from this perspective

connectors/cli.py
```diff
@@ -37,8 +37,20 @@ def _parser():
...
+        default=[
+            "schedule",
+            "execute_content_sync_job",
+            "execute_access_control_sync_job",
+            "cleanup",
+        ],
+        choices=[
+            "schedule",
+            "execute_content_sync_job",
+            "execute_access_control_sync_job",
```
artem-shelkovnikov, line 49: IMO it's too verbose.

We have `schedule`, this one could be `sync_content`, `sync_acs`?

connectors/services/__init__.py
```diff
@@ -3,7 +3,10 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+# fmt: off
+from connectors.services.access_control_sync_job_execution import AccessControlSyncJobExecutionService  # NOQA
```
artem-shelkovnikov, line 7: Why NOQA?

connectors/services/access_control_sync_job_execution.py
```diff
@@ -31,71 +29,23 @@ def load_max_concurrent_access_control_syncs(config):
...
     def __init__(self, config):
         super().__init__(config)
         self.idling = self.service_config["idling"]
-        self.max_concurrent_content_syncs = load_max_concurrent_content_syncs(
-            self.service_config
-        )
+        self.source_list = config["sources"]
         self.max_concurrent_access_control_syncs = (
             load_max_concurrent_access_control_syncs(self.service_config)
         )
```
artem-shelkovnikov, line 41: `load_max_concurrent_access_control_syncs` can be inlined

artem-shelkovnikov (commented):
There's a lot of shared copy-pasted code between two services, is it possible to reuse it instead of copy-pasting somehow?

connectors/services/__init__.py
wangch079: `# NOQA` is to skip line check from [ruff](https://pypi.org/project/ruff/).

They are present in the original code. If I don't include them for the two new lines, they will be removed after running `make autoformat`

`# fmt: off/on` is to skip line check from [black](https://pypi.org/project/black/).

connectors/services/access_control_sync_job_execution.py
wangch079: Did you mean
```python
self.max_concurrent_access_control_syncs = load_max_concurrent_access_control_syncs(self.service_config)
```

It will be formatted like this after running `make autoformat`.

connectors/cli.py
wangch079: Renamed to `sync_content` and `sync_access_control`: https://github.com/elastic/connectors/pull/1840/commits/f5bd9e98da6d9f01e43299564d0f6922694cd90e

The latter one can be further renamed to `sync_ac`, but I think it's not as easy to understand as `sync_access_control`.

connectors/services/access_control_sync_job_execution.py
artem-shelkovnikov: By inlined, I meant [method inlining](https://refactoring.guru/inline-method). `load_max_concurrent_access_control_syncs` just gets a key from config and IMO it does not make a sense to have it a separate method

artem-shelkovnikov:
@wangch079 thank you for addressing the comments.

What about the duplication of the code between the services?

wangch079:
>What about the duplication of the code between the services?

As mentioned in the PR description:

>Further refactor can be done after the `ConcurrentTasks` object is moved out of the loop.

wangch079:
>Is it different from what I've described?

Hi @artem-shelkovnikov , indeed this is as you described.

The current logic is like:
```python
while True:
  # common code

  self.access_control_syncs = ConcurrentTasks()

  # common code

  finally:
    await self.access_control_syncs.join()

  self.access_control_syncs = None
```

For content sync, it's `self.content_syncs`.

I will move `self.access_control_syncs`/`self.content_syncs` out of the loop in the subsequent PR, then it will be easier for me to move the common code to the parent class.

PR #1862, opened 2023-11-02
Remove convert_file_to_b64 in SP Server (unused)

artem-shelkovnikov:
Minor cleanup: method `convert_file_to_b64`:

- Does not convert anything
- Is not used at all

Removing the method


navarone-feekery (approved):
🔥

PR #1863, opened 2023-11-02
Clean up unused code

artem-shelkovnikov:
This PR cleans up unused code - constants, methods, variables.

Additionally, tests that test unused code are removed too.


connectors/es/sink.py
```diff
@@ -148,22 +145,16 @@ async def _bulk_api_call():
...
-                for item in res["items"]:
-                    for op, data in item.items():
-                        if "error" in data:
-                            self._logger.error(
-                                f"operation {op} failed, {data['error']}"
-                            )
-                            raise Exception(data["error"]["reason"])
-
-            self._populate_stats(stats, res)
+        time.time()
```
seanstory, line 148: is this line still needed?

seanstory (approved):

connectors/es/sink.py
artem-shelkovnikov: Ha, good catch, idk why I left it there

seanstory (approved):

PR #1864, opened 2023-11-02
Increase test coverage for connectors.utils module

artem-shelkovnikov:
We're super close to 92% test coverage and minor refactors that I started in https://github.com/elastic/connectors/pull/1857 and https://github.com/elastic/connectors/pull/1863 started to fail due to low code coverage.

This PR improves code coverage for `connectors.utils` module to make space for refactorings to continue.


tests/test_utils.py
```diff
@@ -832,3 +927,69 @@ def test_validate_email_address(email_address, is_valid):
...
+        get_event_loop(uvloop=True)
+        set_event_loop_policy_mock.assert_called()
+
+
+@pytest.mark.asyncio
+async def test_get_event_loop_uvloop_when_exception():
+    with patch("asyncio.set_event_loop_policy", side_effect=Exception("welp")):
+        # Event if there's an exception, the loop is created
+        get_event_loop(uvloop=True)
+        assert True
```
seanstory, line 944: do you need to close it?

tests/test_utils.py
```diff
@@ -832,3 +927,69 @@ def test_validate_email_address(email_address, is_valid):
...
+
+
+@pytest.mark.parametrize(
+    "strategy, interval, retry, expected_sleep",
+    [
+        (RetryStrategy.CONSTANT, 10, 2, 10),  # Constant 10 seconds
+        (RetryStrategy.LINEAR_BACKOFF, 10, 0, 0),  # 10 * 0 = 0
+        (RetryStrategy.LINEAR_BACKOFF, 10, 1, 10),  # 10 * 1 = 10
+        (RetryStrategy.LINEAR_BACKOFF, 10, 2, 20),  # 10 * 2 = 20
+        (RetryStrategy.EXPONENTIAL_BACKOFF, 10, 0, 1),  # 10 ^ 0 = 1
```
seanstory, line 982: is `0` a valid value for a retry counter? Just checking that we're not doing a `sleep(1)` before our first try for every request if we choose an exponential strategy.

seanstory (dismissed):

tests/test_utils.py
```diff
@@ -308,6 +311,24 @@ def _second_callback(result):
     assert second_results == [3]
 
 
+@pytest.mark.asyncio
+async def test_concurrent_runner_cancel():
```
wangch079, line 315: A similar test was added [here](https://github.com/elastic/connectors/pull/1477/files#diff-33c13e0b177bacd2f02e29bcb8aea5b49e7ce34901fd8f41fefb65defba1bd33R261): but it's not merged to `main` yet.

tests/test_utils.py
artem-shelkovnikov: I'm quite confused by this code TBH, in the two tests above I did not need to close the loop, in the tests below I did, most likely because there's current event loop running for test (it's marked as `async`). 

Inside of the test I create a new event loop, and then I need to close it. If I would not create a new one, there would be no need to close anything.

However, I close 2 loops and this is something I'm confused about, I'll do a bit more research here.

tests/test_utils.py
artem-shelkovnikov: Thanks, I'll remove this test!

tests/test_utils.py
```diff
@@ -374,6 +377,53 @@ def test_convert_to_b64_no_overwrite(converter):
...
 
+@pytest.fixture
+def patch_file_ops():
+    with patch("connectors.utils.open"):
+        with patch("os.remove"):
+            with patch("os.rename"):
+                yield
+
+
+@pytest.mark.parametrize("converter", ["system"])
```
vidok, line 388: Is there a way to [parametrize](https://docs.pytest.org/en/7.1.x/example/parametrize.html) these 3 tests? They look the same just `platform.mac_ver` and `platform.system` change.

artem-shelkovnikov: Good call, parametrized!

vidok (approved):

PR #1865, opened 2023-11-03
Improve job execution concurrency

wangch079:
### What's the change

This PR improves the job execution concurrency by:
1. making the task pool as an instance variable, instead of initializing in every iteration. The task pool is responsible for executing the sync job, and the sync job task status (e.g. whether the sync job is put into the task pool successfully, whether the sync job is completed) will not affect the main loop of the job execution service
2. using `fire-and-forget` mechanism to execute sync jobs, i.e. whenever there's a sync job ready to be executed, the service puts the sync job into the task pool and doesn't wait for its completion.
3. using non-blocking `try_put` method for job execution. If the task pool is ready to take any new sync job, the service will give up immediately and retry in the next iteration. 

### Why the change

Under the current implementation, in each iteration, the job execution service
1. initializes a new task pool
2. uses a blocking `put` method to add sync jobs into the task pool
3. calls `join` method at the end of the iteration to wait for the completion of all the tasks

The drawback is that the job execution blocks the job execution service, which prevents it from reaching its full concurrency potential:

Assuming the max concurrency is 3, and there is one pending and long-running sync job, the job execution service will execute it and be blocked until the sync job completes. If there's another pending sync job added, it won't be executed until the first sync job is completed, even though there are resources available.

#### Pre-Review Checklist


jedrazb (commented):
To make it easier for me to review I have a couple of questions?

To me it makes sense that:
- Instead of blocking on `await self.job_pool.put(sync_job_runner.execute)` we use `try_put` and keep on checking with interval set to `idling`  👍 
- Access control and content sync job execution logic were abstracted to the common class 👍 

❓  I have issues figuring out what bit in the code is responsible for `fire-and-forget`, are there any implications w.r.t. job progress tracking, logging or surfacing errors in UI?

connectors/services/job_execution.py
```diff
@@ -0,0 +1,147 @@
...
+
+        if not self.should_execute(connector, sync_job):
+            return
+
+        sync_job_runner = SyncJobRunner(
+            source_klass=source_klass,
+            sync_job=sync_job,
+            connector=connector,
+            es_config=self._override_es_config(connector),
+        )
```
jedrazb, line 84: Should we add a debug line to track adding sync jobs to `ConcurrentTasks`?

wangch079:
> I have issues figuring out what bit in the code is responsible for `fire-and-forget`, are there any implications w.r.t. job progress tracking, logging or surfacing errors in UI?

In the previous implementation, in each iteration, it `put` the sync job to the task pool, and wait for all tasks to complete at the end of the iteration:

https://github.com/elastic/connectors/blob/d606a5384f6185b349a4f0e329326436c790829d/connectors/services/job_execution.py#L198-L199

This means the `JobExecutionService` will not go to the next iteration until all jobs are complete.

In this PR, The `JobExecutionService` just `try_put` the sync jobs, and sleep before going to the next iteration. It doesn't care about the real execution of the sync jobs, which are handled by `ConcurrentTasks` and `SyncJobRunner`

connectors/services/job_execution.py
```diff
@@ -0,0 +1,146 @@
...
+        if not self.sync_job_pool.try_put(sync_job_runner.execute):
+            sync_job.log_warning(
+                f"{self.display_name.capitalize()} has reached the maximum concurrency of {self.max_concurrency}. Retry in {self.idling} seconds."
+            )
```
artem-shelkovnikov, line 87: I think this message can be improved to indicate some points to the user:

1. What's up with "maximum concurrency" in this case? We can say that service X is already running Y jobs and cannot run more at this point.
2. What's the action user can take? For example, increase "Z" setting in the connector framework to allow more jobs to run.
3. Should it be a warning or info? I can imagine this being spammed every 30 seconds if jobs take too long to finish, and warnings that often look scary. To me this logic is instead a part of normal flow, so info/debug might suit better

connectors/services/job_execution.py
```diff
@@ -0,0 +1,146 @@
...
+            (
+                is_platinum_license_enabled,
+                license_enabled,
+            ) = await self.connector_index.has_active_license_enabled(License.PLATINUM)
+
+            if not is_platinum_license_enabled:
+                sync_job.log_error(
+                    f"Minimum required Elasticsearch license: '{License.PLATINUM.value}'. Actual license: '{license_enabled.value}'."
+                )
+                return
```
artem-shelkovnikov, line 73: Probably not for this PR, but I'd move this logic around a bit so that we would not show this error if the job is supposed to execute.

Cause now if you're using connector that supports permissions, but have no license + no permissions enabled you still see this error in console.

Should I create an issue for it?

artem-shelkovnikov (approved):
Good step ahead! Left a couple small comments/suggestions

connectors/services/job_execution.py
wangch079: In the happy path (not considering users manually index a sync job document), this can happen in a very rare case, when the license expires after the acl sync job is scheduled/manually triggered (We also check licence when scheduling an acl sync job). 

When it happens, we have no way to mark this job as error and the same job will be picked up in every iteration. I think we should move this licence check into `SyncJobRunner`, which will mark the job as error if it doesn't have the minimum licence required.

I have created the issue: https://github.com/elastic/connectors/issues/1924

connectors/services/job_execution.py
wangch079: >We can say that service X is already running Y jobs and cannot run more at this point.

Will rephrase it, although I think they are more or less the same 😆 

>What's the action user can take? For example, increase "Z" setting in the connector framework to allow more jobs to run.

That's a good point, I will add this to the log message.

>To me this logic is instead a part of normal flow, so info/debug might suit better

👍

wangch079: Updated in https://github.com/elastic/connectors/pull/1865/commits/b87c970c16f23abfb6aeb9a6297ecc893f67f178

The message is like:

>Content sync job execution service is already running 4 sync jobs and can't run more at this poinit. It will retry in 30 seconds. Increase 'service.max_concurrent_content_syncs' in config if you want the service to run more sync jobs.


>Access control sync job execution service is already running 4 sync jobs and can't run more at this poinit. It will retry in 30 seconds. Increase 'service.max_concurrent_access_control_syncs' in config if you want the service to run more sync jobs.

PR #1871, opened 2023-11-06
Improve handling of username and password in Mongo Connector

dcampillo:
## Closes  https://github.com/elastic/connectors/issues/1870 ###

Code changes:
The changes validated IF the variable 'user' and/or 'password' are NOT of NoneType. If True, then the variables 'user' and/or 'password' are appened to the config_params dict.

#### Pre-Review Checklist
- [X] Considered corresponding documentation changes


artem-shelkovnikov:
@dcampillo there was a different PR with some questions not addressed, do you mind returning back to the discussion in the previous PR?

We cannot reproduce this problem on recent versions of connectors and the framework should handle such scenarios by default - e.g. if the field is of type `String`, then when `None` is provided we transform it to an empty string.

dcampillo:
Hi @artem-shelkovnikov ,

The fix proposed is related to the cusstom connector for MongoDB, not the Framework itself. In the code, username and password where validated by checking the length of the string, which raise an exception when the username and password are not provided in the UI. In this case, the len(None) trigger an exception.

In " def get_default_configuration(cls):", the username and password are not required, which is ok IF the user wants to connect to MongoDB without username/password, but at the same time the validation of the username and password generated an exception.

Either we improve the handling of the username and password, or we make the username and password required. I would prefer to go with the latter.

Extract of the schema in mongo.py
...
 "user": {
                "label": "Username",
                "order": 2,
                **"required": False,**
                "type": "str",
            },
            "password": {
                "label": "Password",
                "order": 3,
                **"required": False,**
                "sensitive": True,
                "type": "str",
...

David

seanstory:
I'm going to close this PR, as this is not the approach we want to take to address this issue. @dcampillo please continue the conversation with us in https://github.com/elastic/connectors/issues/1870 if this is still something you'd like to see addressed.

PR #1874, opened 2023-11-06
[Draft] Add support for DLS in ServiceNow

parthpuri-crest:
Draft PR for ServiceNow DLS. Add RCF for Roles mapping

### DIFF
--- a/connectors/sources/servicenow.py
+++ b/connectors/sources/servicenow.py
@@ -16,6 +16,11 @@ import aiohttp
 import dateutil.parser as parser
 import fastjsonschema
 
+from connectors.access_control import (
+    ACCESS_CONTROL,
+    es_access_control_query,
+    prefix_identity,
+)
 from connectors.filtering.validation import (
     AdvancedRulesValidator,
     SyncRuleValidationResult,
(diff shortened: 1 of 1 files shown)


connectors/sources/servicenow.py
```diff
@@ -470,15 +530,101 @@ def get_default_configuration(cls):
...
+        if (
+            self._features is None
+            or not self._features.document_level_security_enabled()
+        ):
+            return False
+
+        return self.configuration["use_document_level_security"]
+
+    async def _user_access_control_doc(self, user):
+        user_id = user.get("_id", "")
```
timgrein, line 565: I'm not sure, if this will work. If the `user_id` is empty the framework will probably use an empty `_id` when ingesting the doc, which would lead to Elasticsearch auto-generating a value for `_id`. So for every access control sync N new documents would be added (which conceptually refer to the same identity), if you've N documents without an `_id`.

Can you verify that behavior? If that's the case we should probably log a warning, if the `_id` is missing without yielding it to the framework.

timgrein (changes requested):
Left a comment around a potentially empty `_id`. Also you need to sign the contributor agreement. Thanks!

praveen-elastic:
Closing this as the other one merged #1989

PR #1880, opened 2023-11-07
bump version for 8.11.0

seanstory:
incrementing the BUILD version after releasing 8.11.0.0


connectors/VERSION
```diff
@@ -1 +1 @@
-8.11.0.0
+8.11.0.1
```
artem-shelkovnikov, line 1: Should it be like this?
```suggestion
8.11.1.0
```

seanstory: Glad I'm not the only one that thought this!
https://github.com/elastic/connectors/blob/main/docs/RELEASING.md explicitly says no:

> Update the BUILD version (https://github.com/elastic/connectors/pull/122). Note that the Connectors project does not immediately bump to the next PATCH version. That won't happen until that patch release's FF date.

Which I suppose makes sense. If we wanted to make another release today, it would be for 8.11.0 still, not for 8.11.1

artem-shelkovnikov: Ohhhh it's something new! TIL!

artem-shelkovnikov (approved):

PR #1883, opened 2023-11-08
Make full sync timestamp optimizations configurable

seanstory:
## Part of https://github.com/elastic/connectors/issues/1372

One way we could make full syncs be able to fully reset data. This implementation uses two new configs in the YAML file to separately configure
- if metadata should be updated on Full syncs even when the timestamp shows it hasn't changed (`service.enable_full_sync_meta_ts_optimization`)
- if binary downloads should be done on Full syncs even when the timestamp shows it hasn't been changed (`service.enable_full_sync_download_ts_optimization`)

It also provides a new configuration section for source-specific configuration overrides (`service.source_overrides`). The structure for this is an array of objects, like:
```
service:
  source_overrides:
    - source: source_1
       config_key_1: overwritten_val_1
       config_key_2: overwritten_val_2
     - source: source_2
       config_key_1: overwritten_val_3
       config_key_2: overwritten_val_4
```

The only supported keys for config overwriting right now are the new ts_optimization fields

This change is not, strictly speaking, backwards compatible, and may adversely impact performance of full syncs. However, it also makes upgrading and iterating on ingest pipeline changes significantly easier, and provides an escape hatch for opting back in to the backwards-compatible behavior.

#### Pre-Review Checklist

## Related pull requests
- [ ] cloud allow-list PR TODO
- [ ] ent-search config for native connectors TODO

## Release Note

Introduces two new configs, `service.enable_full_sync_meta_ts_optimization` (default is `false`) and `service.enable_full_sync_download_ts_optimization` (default is `true`). This change causes full syncs to always send all metadata documents to Elasticsearch, even if no obvious data change was detected. This is helpful for ensure that code changes are applied to all documents with a full sync, whether that code change comes from a Connectors upgrade or a change to your ingest pipelines. Also provides a config `service.source_overrides` where you can configure specific settings depending on the connector source type.

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -278,6 +278,8 @@ class Extractor:
         display_every=DEFAULT_DISPLAY_EVERY,
         concurrent_downloads=DEFAULT_CONCURRENT_DOWNLOADS,
         logger_=None,
+        optimize_on_meta_timestamp=False,
+        optimize_on_download_timestamp=True,
     ):
         if filter_ is None:
             filter_ = Filter()
@@ -299,6 +301,8 @@ class Extractor:
         self.concurrent_downloads = concurrent_downloads
         self._logger = logger_ or logger
         self._canceled = False
+        self.optimize_on_meta_timestamp = optimize_on_meta_timestamp
+        self.optimize_on_download_timestamp = optimize_on_download_timestamp
 
     async def _get_existing_ids(self):
         """Returns an iterator on the `id` and `_timestamp` fields of all documents in an index.
@@ -431,11 +435,17 @@ class Extractor:
                     if TIMESTAMP_FIELD in doc and ts == doc[TIMESTAMP_FIELD]:
                         # cancel the download
                         if (
-                            self.content_extraction_enabled
+                            self.optimize_on_download_timestamp
+                            and self.content_extraction_enabled
                             and lazy_download is not None
                         ):
                             await lazy_download(doit=False)
-                        continue
+                            lazy_download = (
+                                None  # we don't want to execute this again below
+                            )
+
+                        if self.optimize_on_meta_timestamp:
+                            continue
 
                     self.total_docs_updated += 1
                 else:
@@ -444,7 +454,11 @@ class Extractor:
                         doc[TIMESTAMP_FIELD] = iso_utc()
 
                 # if we need to call lazy_download we push it in lazy_downloads
-                if self.content_extraction_enabled and lazy_download is not None:
+                if (
+                    self.optimize_on_download_timestamp
+                    and self.content_extraction_enabled
+                    and lazy_download is not None
+                ):
                     await lazy_downloads.put(
                         functools.partial(
                             self._deferred_index, lazy_download, doc_id, doc, operation
@@ -775,6 +789,8 @@ class SyncOrchestrator(ESClient):
         sync_rules_enabled=False,
         content_extraction_enabled=True,
         options=None,
+        optimize_on_meta_timestamp=False,
+        optimize_on_download_timestamp=True,
     ):
         """Performs a batch of `_bulk` calls, given a generator of documents
 
@@ -819,6 +835,8 @@ class SyncOrchestrator(ESClient):
             display_every=display_every,
             concurrent_downloads=concurrent_downloads,
             logger_=self._logger,
+            optimize_on_meta_timestamp=optimize_on_meta_timestamp,
+            optimize_on_download_timestamp=optimize_on_download_timestamp,
         )
         self._extractor_task = asyncio.create_task(
             self._extractor.run(generator, job_type)
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -134,9 +134,7 @@ class SyncJobRunner:
             bulk_options = self.bulk_options.copy()
             self.data_provider.tweak_bulk_options(bulk_options)
 
-            self.elastic_server = SyncOrchestrator(
-                self.es_config, self.service_config, self.sync_job.logger
-            )
+            self.elastic_server = SyncOrchestrator(self.es_config, self.sync_job.logger)
 
             if job_type in [JobType.INCREMENTAL, JobType.FULL]:
                 self.sync_job.log_info(f"Executing {job_type.value} sync")
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -79,6 +79,7 @@ class SyncJobRunner:
         sync_job,
         connector,
         es_config,
+        service_config,
     ):
         self.source_klass = source_klass
         self.data_provider = None
@@ -90,6 +91,7 @@ class SyncJobRunner:
         self.bulk_options = self.es_config.get("bulk", {})
         self._start_time = None
         self.running = False
+        self.service_config = service_config
 
     async def execute(self):
         if self.running:
@@ -132,7 +134,9 @@ class SyncJobRunner:
             bulk_options = self.bulk_options.copy()
             self.data_provider.tweak_bulk_options(bulk_options)
 
-            self.elastic_server = SyncOrchestrator(self.es_config, self.sync_job.logger)
+            self.elastic_server = SyncOrchestrator(
+                self.es_config, self.service_config, self.sync_job.logger
+            )
 
             if job_type in [JobType.INCREMENTAL, JobType.FULL]:
                 self.sync_job.log_info(f"Executing {job_type.value} sync")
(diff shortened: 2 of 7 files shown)


connectors/es/sink.py
```diff
@@ -444,7 +454,11 @@ async def get_docs(self, generator):
                         doc[TIMESTAMP_FIELD] = iso_utc()
 
                 # if we need to call lazy_download we push it in lazy_downloads
-                if self.content_extraction_enabled and lazy_download is not None:
+                if (
+                    self.optimize_on_download_timestamp
```
jedrazb, line 458: Is the `optimize_on_download_timestamp` check needed here? This `if` statement affects both existing and new docs. The `optimize_on_download_timestamp` should only relate to existing docs. Would checking if `self.content_extraction_enabled and lazy_download is not None:` be sufficient here?

jedrazb: Can disabling the optimisation prevent us from downloading any content?

connectors/sync_job_runner.py
```diff
@@ -214,8 +216,30 @@ async def _execute_content_sync_job(self, job_type, bulk_options):
             sync_rules_enabled=sync_rules_enabled,
             content_extraction_enabled=content_extraction_enabled,
             options=bulk_options,
+            optimize_on_meta_timestamp=self._lookup_service_config(
+                "enable_full_sync_meta_ts_optimization", False
```
jedrazb, line 220: Nit, we keep all of our default config values in `connectors/config.py`. Should this fallback be defined there?

connectors/sync_job_runner.py
```diff
@@ -214,8 +216,30 @@ async def _execute_content_sync_job(self, job_type, bulk_options):
...
 
+    def _lookup_service_config(self, config_key, fallback):
+        """
+        Retrieve the source override config if present, otherwise return the top-level service config
+        :param config_key: the configuration to look up
+        :param fallback: the default value to use if the config_key is not found in the service config
+        :return: the configured value for config_key
+        """
+        default_value = self.service_config.get(config_key, fallback)
+        overrides = self.service_config.get("source_overrides", [])
```
artem-shelkovnikov, line 235: We spoke about it - we don't have source-specific config options now at all, and it's done on-purpose.

Content sources/connectors in current (and especially future) designs are supposed to be distributed. This change prevents this.

Our case is:

In current or serverless world, we don't have guarantee that a connector sync is executed on the same host. If hosts have different per-connector configs, then syncs will produce different results over time which is really hard to debug, support and grasp.

Config options like this should be done on the level of connector, IMO, and expressed in UI. Exposing them in config is a very slippery slope

artem-shelkovnikov (changes requested):

artem-shelkovnikov:
> Also provides a config service.source_overrides where you can configure specific settings depending on the connector source type.

Can we postpone this change? If not, can we publicly discuss it? We never had this option, and it's for a good reason IMO, I'd love to postpone such configs for as long as it's possible.

connectors/sync_job_runner.py
seanstory: The fallback of `False` is defined there. I'm hesitant to put the function in `config.py` because right now we treat config as just a dict, not as a class/object. And we frequently pass around just a sub-dict. So I expect we'd end up needing to implement this outside of config.py anyway.

connectors/sync_job_runner.py
seanstory: This was discussed at some length in our [project sync from Nov 9th](https://docs.google.com/document/d/1nV1CMdJVOrusbb53NoFrr4p0ee37To8wicyPfPOTCxs/edit#heading=h.4hda5l2bllxq). The issue is that a configuration through the UI is more involved, so we wanted to look at a solution that didn't require Kibana changes, and also didn't shoehorn this into an RCF that would need to be added individually to every connector.  We also discarded doing connector-specific YAML configs because that would require a process like:
1. create the connector in kibana
2. partially configure it
3. change the YAML for connectors to specify the optimization behavior
4. restart connectors (on Native, restart the whole node)
5. finish configuring and sync

which is not an ideal flow. The compromise we landed on was configuring by service type.

I expect in a multi-tenant serverless world we simply wouldn't expose this configuration - just like we wouldn't expose any `service` configs (like how many threads should be used).

connectors/sync_job_runner.py
```diff
@@ -214,8 +216,30 @@ async def _execute_content_sync_job(self, job_type, bulk_options):
...
             options=bulk_options,
+            optimize_on_meta_timestamp=self._lookup_service_config(
+                "enable_full_sync_meta_ts_optimization", False
+            ),
+            optimize_on_download_timestamp=self._lookup_service_config(
+                "enable_full_sync_download_ts_optimization", True
+            ),
         )
 
+    def _lookup_service_config(self, config_key, fallback):
```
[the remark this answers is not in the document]
seanstory: I'd added some unit tests that I felt were sufficient - are there edge cases you feel are not covered and should be?

seanstory:
> Can we postpone this change? 

Not really. A fix for this problem is needed urgently by an important customer. I'm open to be flexible on _how_ we solve it, but not really on postponing a solution. 

> If not, can we publicly discuss it? We never had this option, and it's for a good reason IMO, I'd love to postpone such configs for as long as it's possible.

We did discuss it in the Project Sync on Nov 9th, but I know you weren't able to attend. Would you like to schedule a team discussion during your hours (ok if I can't attend that one, I'll catch up) so that your position can be addressed? I don't want you to feel left out. It's funny, in the meeting, I'd said, "wasn't someone really opposed to service-type specific configs?" and none of us could remember who that had been or why, so we moved on. Guess it was you 😅

connectors/sync_job_runner.py
wangch079: I agree we shouldn't expose any connector-specific or even source/service type-specific config in YAML, to make sync consistent. But this should only happen to cloud deployments of which we have control. 

For connector clients, developers can even customize the logic in whatever way they want, and I see no harm in giving the flexibility via config. If a developer runs a connector client, he/she should know what he/she is doing and be responsible for that.

connectors/sync_job_runner.py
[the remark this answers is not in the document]
wangch079: I'd suggest we override the service config before passing it to `SyncJobRunner`, similar as the `es_config`:
https://github.com/elastic/connectors/blob/9e53db273a2d85250efbdd095dd89c97e28c1728/connectors/services/job_execution.py#L80-L81

connectors/sync_job_runner.py
artem-shelkovnikov: > For connector clients, developers can even customize the logic in whatever way they want, and I see no harm in giving the flexibility via config. If a developer runs a connector client, he/she should know what he/she is doing and be responsible for that.

That is true and we don't prevent forks, but on the level of framework this should not be supported by us - if anyone introduces unsupported logic then they are bound to support it, not us.

For me harm from flexibility via config (apart from problems with serverless) is also unnecessary complexity of connector setup.

So now you set up connector by:
1. First do setup via UI
2. Some features are behind a feature flag, so you go to dev tools and update setting there

If we allow customisation on config.yaml level, that's the third place. How do you now decide which config options go where? It's not very obvious to starrt.

But most importantly tp me it's just _too complex_ to manage -  maybe if you did your setup in a very simplistic way, everything works, you're happy. 

If the setup becomes more complex (you add redundancy, k8s, docker, tons of connector and instances running) then managing it becomes wildly more complex. Additionally, if something goes wrong in sync you'll have to dig into debugging much deeper. Let's say you have a problem with some sync that happened 4 days ago. You need to find out _who_ ran the sync, then if the host is still alive find it, then check its config file (if it was not edited along the way), etc, etc. It becomes a rabbit hole very fast and it's opposite of simplicity that we're trying to have.

So to me adding official support to config.yaml configuration per connector opens up tremendous complexity, which turns to slower development, much harder SDH and more potential errors/mistakes.

sphilipse: The current idea is that we only support this for (customized) connector clients, which have fewer support demands than native connectors (or future multi-tenant connectors). To me this is just another way for users to customize their connector to fit their needs. 

It also fits where we want to go with connectors, which is to make them capable of running as standalone processes without managing their config through Elasticsearch. The advantages of this in the long run is that we can tell users concerned about secrets to store their secrets in the config.yml (populated through some other process probably).

seanstory:
After team discussion, we decided that the complexity of making this configurable was not justified, and that the better solution is to remove this over-optimization as it causes unintended bugs when upgrading and iterating. Instead, in future versions we will revisit our approach to syncs, either providing more configurability options for behaviors or adding new types of syncs that can be scheduled and triggered.

PR #1892, opened 2023-11-13
[DOCS] Delete prerelease refs that now live in Elastic docs

leemthompo:
These docs are now live on Elastic.co


timgrein (approved):
LGTM, one question: We're keeping `docs/reference/README.md` so that people, who assumed it's still there to find it again?

leemthompo:
>one question: We're keeping docs/reference/README.md so that people, who assumed it's still there to find it again?

@timgrein Right, I figured that if someone switched branches, for example, it would be jarring if that folder didn't exist and 404'd. WDYT?

timgrein:
Makes sense to me 👍

PR #1893, opened 2023-11-13
Bump version to 8.11.1 + fix version in ftest

artem-shelkovnikov:
Bumping VERSION to 8.11.2 as a part of release https://github.com/elastic/enterprise-search-team/issues/5972


tests/ftest.sh
```diff
@@ -23,7 +23,7 @@ export PERF8_TRACE=${PERF8_TRACE:-False}
 export REFRESH_RATE="${REFRESH_RATE:-5}"
 export DATA_SIZE="${DATA_SIZE:-medium}"
 export RUNNING_FTEST=True
-export VERSION='8.10.0-SNAPSHOT'
+export VERSION='8.11.0-SNAPSHOT'
```
seanstory, line 26:
```suggestion
export VERSION='8.11.1-SNAPSHOT'
```
?

seanstory (approved):
rubber stamp, but I'm unsure if these are the right versions?

tests/ftest.sh
artem-shelkovnikov: Theoretically yeah, in practice we don't care much as we provide compatibility on major/minor version? 🤷

PR #1921, opened 2023-11-24
[S3] Advanced Sync rules

akanshi-elastic:
Closes https://github.com/elastic/connectors/issues/1885
This PR introduces advanced filtering rules for S3 by queries to filter out document based on that
Added documentation with several examples illustrating the syntax, how a basic use case works
Included tooltip on the configurable fields that are ignored when advanced sync rules are used
#### Pre-Review Checklist


connectors/sources/s3.py
```diff
@@ -126,7 +133,12 @@ async def get_bucket_objects(self, bucket):
                 bucket_obj = await s3.Bucket(bucket)
                 await asyncio.sleep(0)
 
-                async for obj_summary in bucket_obj.objects.page_size(page_size):
+                if "prefix" in kwargs:
```
wangch079, line 136: This will be evaluated to `True` when you pass `prefix=""`

connectors/sources/s3.py
```diff
@@ -126,7 +133,12 @@ async def get_bucket_objects(self, bucket):
...
+                    objects = bucket_obj.objects.filter(Prefix=kwargs["prefix"])
```
wangch079, line 137: Do we need to apply `page_size` here?

akanshi-elastic: We can add but that will not make a big difference in terms of performance as filter is already added here for filtering out the data.

connectors/sources/s3.py
akanshi-elastic: `prefix=""` is the default behaviour for filtering out the data. So when `prefix=""` that means it will fetch all the object for defined bucket.

akanshi-elastic: This will not throw any error as the default value for prefix is `""`. but yes let me change it a bit to make it more clear.

connectors/sources/s3.py
```diff
@@ -272,6 +351,7 @@ def get_default_configuration(cls):
                 "display": "textarea",
                 "label": "AWS Buckets",
                 "order": 1,
+                "tooltip": "buckets are ignored when Advanced Sync Rules are used.",
```
wangch079, line 354:
```suggestion
                "tooltip": "AWS Buckets are ignored when Advanced Sync Rules are used.",
```

connectors/sources/s3.py
wangch079: Just curious if this is supported?

```python
objects = bucket_obj.objects.filter(Prefix=kwargs["prefix"]).page_size(page_size)
```

connectors/sources/s3.py
```diff
@@ -105,7 +111,7 @@ async def get_bucket_list(self):
             buckets = self.configuration["buckets"]
         return buckets
 
-    async def get_bucket_objects(self, bucket):
+    async def get_bucket_objects(self, **kwargs):
```
wangch079, line 114: Sorry didn't realize it in the first review. I think this change makes the `bucket` parameter optional.

What do you think if we change it to 
```suggestion
    async def get_bucket_objects(self, bucket, prefix=None):
```

or 

```suggestion
    async def get_bucket_objects(self, bucket, **filters):
```

connectors/sources/s3.py
akanshi-elastic: yes it is.

wangch079 (approved):

PR #1923, opened 2023-11-24
Added DLS For GitHub

parth-elastic:
## Related to https://github.com/elastic/connectors-py/issues/1886

Add support of Document Level Security in GitHub.

#### Pre-Review Checklist


connectors/sources/github.py
```diff
@@ -919,49 +1080,106 @@ def get_default_configuration(cls):
...
+    async def _user_access_control_doc(self, user):
+        user_id = user.get("id", "")
+        user_name = user.get("login", "")
+        user_email = user.get("email", "")
+
+        _prefixed_user_id = _prefix_user_id(user_id=user_id)
+        _prefixed_user_name = _prefix_username(user=user_name)
+        _prefixed_email = _prefix_email(email=user_email)
+        return {
+            "_id": user_id,
```
timgrein, line 1164: What happens if you get multiple users with an empty id? Will they override each other or will new documents created with subsequent syncs, if ES generates new unique ids for users having an empty id?

connectors/sources/github.py
```diff
@@ -1397,6 +1643,46 @@ def is_previous_repo(self, repo_name):
...
+        owner, repo = self.github_client.get_repo_details(repo_name)
+        collaborator_variables = {
+            "orgName": owner,
+            "repoName": repo,
+            "cursor": None,
+        }
+        access_control = []
+        if len(self.members) <= 0:
+            async for user in self.github_client._fetch_all_members():
+                self.members.add(user.get("id"))
```
timgrein, line 1663: What happens if `id` is `None`? Should we skip it then?

connectors/sources/github.py
```diff
@@ -1463,7 +1749,20 @@ async def get_docs(self, filtering=None):
             async for repo in self._fetch_repos():
                 if self.is_previous_repo(repo["nameWithOwner"]):
                     continue
-                yield repo, None
+                access_control = []
+                if self._dls_enabled():
+                    if repo.get("visibility").lower() == "public":
```
timgrein, line 1754:
```suggestion
                    is_public_repo = repo.get("visibility").lower() == "public"
                    if is_public_repo:
```

Small nitpick

timgrein (commented):
Left some comments, otherwise LGTM

connectors/sources/github.py
parth-elastic: I don't think we get an empty ID for any document, as we have used the ID that was generated by GraphQL to identify the records. Only emails can be empty for some users. We always get a unique ID and username for all the users.

connectors/sources/github.py
parth-elastic: As mentioned above I don't think we can get empty ID. In our testing we did not faced this scenario.

timgrein (approved):

PR #1925, opened 2023-11-27
Improve the concurrency of the Job execution service

wangch079:
### What's the change

1. Break `JobExecutionService` to `ContentSyncJobExecutionService` and `AccessControlSyncJobExecutionService` to handle content and access control sync job execution separately.
2. Do not let `ConcurrentTasks` block the execution of the `JobExecutionService`.

### Why the change

1. The execution of a sync job should not affect the execution of `JobExecutionService`, e.g. if a sync job is taking a long time to run, it should not affect another sync job to be executed, unless the service reaches its max concurrency
2. The execution of content sync jobs should not affect the execution of access control sync jobs. e.g. if there are too many content sync jobs to be executed, access control sync jobs should be executed as well.

### related PRs

This PR has been broken into multiple PRs and they have been reviewed:

1. #1477 
2. #1840 
3. #1889 
4. #1479 
5. #1865

#### Pre-Review Checklist


connectors/cli.py
```diff
@@ -36,8 +36,20 @@ def _parser():
     parser.add_argument(
         "--action",
         type=str,
-        default=["schedule", "execute", "cleanup"],
-        choices=["schedule", "execute", "list", "config", "cleanup"],
+        default=[
```
artem-shelkovnikov, line 39: cc @vidok - do we need to be careful along with the CLI change when adding options here?

connectors/services/__init__.py
```diff
@@ -3,7 +3,10 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+# fmt: off
```
artem-shelkovnikov, line 6: Why tho?

artem-shelkovnikov: I think it's a sign our names for classes are too long, can we shorten them maybe?

connectors/services/access_control_sync_job_execution.py
```diff
@@ -0,0 +1,38 @@
...
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+from functools import cached_property
+
+from connectors.protocol import JobStatus, JobType
+from connectors.services.job_execution import JobExecutionService
+
+DEFAULT_MAX_CONCURRENT_ACCESS_CONTROL_SYNCS = 1
```
artem-shelkovnikov, line 11: Should it go to default config instead?

connectors/services/access_control_sync_job_execution.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class AccessControlSyncJobExecutionService(JobExecutionService):
+    name = "sync_access_control"
+    display_name = "access control sync job execution"
+    max_concurrency_config = "service.max_concurrent_access_control_syncs"
```
artem-shelkovnikov, line 17: To my taste passing these explicitly in initialiser would be better, so just wondering, how do you feel about it in init method vs via class variables?

connectors/services/access_control_sync_job_execution.py
```diff
@@ -0,0 +1,38 @@
...
+    JOB_TYPES = JobType.ACCESS_CONTROL.value
```
artem-shelkovnikov, line 18: Also why is this one full capital, but previous ones are lower case?

connectors/services/job_execution.py
```diff
@@ -20,81 +20,32 @@
...
 
-        sync_job_runner = SyncJobRunner(
-            source_klass=source_klass,
-            sync_job=sync_job,
-            connector=connector,
-            es_config=self._override_es_config(connector),
-        )
-        await self.access_control_syncs.put(sync_job_runner.execute)
+    def should_execute(self, connector, sync_job):
+        return False
```
artem-shelkovnikov, line 48: Why not raise here too?

connectors/services/job_execution.py
```diff
@@ -121,20 +72,18 @@ async def _sync(self, sync_job):
...
-                f"Unsupported job type '{job_type}'. Skipping sync job execution..."
+        sync_job_runner = SyncJobRunner(
+            source_klass=source_klass,
+            sync_job=sync_job,
+            connector=connector,
+            es_config=self._override_es_config(connector),
+        )
+        if not self.sync_job_pool.try_put(sync_job_runner.execute):
+            sync_job.log_info(
+                f"{self.display_name.capitalize()} service is already running {self.max_concurrency} sync jobs and can't run more at this poinit. It will retry in {self.idling} seconds. Increase '{self.max_concurrency_config}' in config if you want the service to run more sync jobs."  # pyright: ignore
```
artem-shelkovnikov, line 86: This has potential as being spammy (e.g. every 30 seconds by default it'll apper) - should it be debug?

It's also too long IMO, can we shorten it? I think the part with `self.idling` can go.

artem-shelkovnikov (commented):
Looks good overall, left a couple questions/comments.

Looking forward to improved concurrency!

connectors/services/access_control_sync_job_execution.py
wangch079: This is the default value when `service.max_concurrent_access_control_syncs` is not configured.

Is the `default config` somewhere we keep all the default values?

connectors/services/access_control_sync_job_execution.py
wangch079: Because these are not going to change from instance to instance, so I feel it's more appropriate to keep them at class level.

connectors/services/access_control_sync_job_execution.py
wangch079: Good point. It should be a class variable, lower case. Will update.

connectors/services/__init__.py
artem-shelkovnikov: Oh wow, why is it removed?

connectors/services/access_control_sync_job_execution.py
artem-shelkovnikov: IMO class-level variables are more suitable for metaprogramming (specifically cause you don't need to instantiate the class to get these values).

(this point is just to chat about it, not to change the code)

connectors/services/__init__.py
wangch079: Now I understand why.

`make autoformat` will run `black` first, then `ruff`:
https://github.com/elastic/connectors/blob/d1d7f9a72ab6cdc5cefc0d65c05390a9a5b39bd3/Makefile#L45-L53

`black` will change
```python
from connectors.services.access_control_sync_job_execution import (
    AccessControlSyncJobExecutionService,  # NOQA
)
```

to
```python
from connectors.services.access_control_sync_job_execution import (
    AccessControlSyncJobExecutionService, 
)  # NOQA
```
Because `# NOQA` is not applied at the 2nd line, and `AccessControlSyncJobExecutionService` is imported but not used, `ruff` will remove it.

wangch079: https://github.com/elastic/connectors/pull/1925/commits/a435d797c1c6ed303628e6e961c7e07879580792 should make it work

connectors/services/access_control_sync_job_execution.py
wangch079: Good point. I will address it as a follow-up PR

connectors/services/access_control_sync_job_execution.py
wangch079: >IMO class-level variables are more suitable for metaprogramming (specifically cause you don't need to instantiate the class to get these values).

Agreed. That's how the variable `name` comes from (access the variable without instantiation of the class).

For the rest, probably, we could use method:

```python
@cached_property
def display_name():
	return "access control sync job execution"
```

WDYT?

artem-shelkovnikov: Rather than `@cached_property` it can be `@staticmethod` IMO, but it's also just me being purist about it. On the other hand cached property is fine cause you can use it like `self.display_name` rather than `self.class.display_name`, which is looking a bit cleaner.

So yeah, cached_property? Also, we can keep it as is too, that's an option, just wanted to hear your thoughts :)

wangch079: I want to keep it as a class-level thing, as it won't change from instance to instance. But it was mainly used by instance, so it has to be referenced as `self.__class__.display_name`, which is annoying.

And I think @cached_property is a better solution here. Will update.

wangch079: Updated in https://github.com/elastic/connectors/pull/1925/commits/d66692b480664b42f6ae8f6c7b957993bbb6215d

connectors/services/__init__.py
artem-shelkovnikov: Note to self: we can probably get rid of `black` as `ruff` is already doing its job

artem-shelkovnikov (approved):

PR #1928, opened 2023-11-28
Update ACL index name pattern

navarone-feekery:
## Part of https://github.com/elastic/enterprise-search-team/issues/6256

This PR changes how an ACL filter is found in Connectors.
ACL filters will now have a simplified naming pattern of `.search-acl-filter-<index name>`. See related Kibana changes for reference.

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/kibana/pull/172057

## Release Notes
ACL indices are being migrated in 8.12 to a new naming pattern. If a user is using a Connector Client without an Enterprise Search node, this migration will not run for them. This means that DLS will not work after upgrading. Users can remedy this by doing the following manual steps:

- Manually delete all ACL indices from dev tools
- Run ACL syncs for each connector affected
- Re-issue API keys for these connectors and add them to the config.yml


navarone-feekery:
Note about tests: there were no tests for testing if the ACL index name was correctly found from the previous naming pattern.
The existing tests already cover my changes. [See here](https://github.com/elastic/connectors/blob/1acac91719709b2acf233e4ef3a55c7fea1dc9c9/tests/protocol/test_connectors.py#L1903-L1951).

seanstory (approved):

leemthompo:
@seanstory @navarone-feekery is starting up an Ent Search node for the purposes of migrating an option here for connector clients that don't want to do all the manual work?

navarone-feekery:
@leemthompo yes that should be totally fine

PR #1931, opened 2023-11-28
Skip empty permission batches

seanstory:
## Closes https://github.com/elastic/connectors/issues/1930

Skips running a `post` request to the `$batch` endpoint if the requests list is empty.

I can't actually see how we could get in this situation, but the linked ticket shows that we _are_ getting into this spot. I'd rather track down the root cause, but figured a patch for the symptom was better than not solving anything.

#### Pre-Review Checklist

## Release Note

Fixes a bug where the Sharepoint Online connector might sometimes fail syncs when DLS was enabled due to a batching error.


tests/test_utils.py
```diff
@@ -820,22 +820,24 @@ def batch_size(value):
...
+        # ),
+        # (
+        #     [1, 2, 3],
+        #     batch_size(1000),
+        #     [
+        #         [1, 2, 3],
+        #     ],
+        # ),
+        # ([], batch_size(20), []),
+        ([[]], batch_size(20), [[[]]]),
```
wangch079, line 840: Did you include this change accidentally?

leemthompo:
@seanstory I'm adding this to release notes, let me know if anything changes

tests/test_utils.py
seanstory: The commented out test cases were a mistake 🙈 
I did intend to add these other two test cases, to show that the issue doesn't seem to be a bug in `iterable_batches_generator`

wangch079 (approved):

PR #1935, opened 2023-11-30
Fix issue for skipping buckets for GCS and ABS if it doesn't have access

akanshi-elastic:
## Part of https://github.com/elastic/connectors/issues/1831###
PR includes:
1. Fix issue for skipping buckets for GCS and ABS if it doesn't have access

#### Pre-Review Checklist


connectors/sources/google_cloud_storage.py
```diff
@@ -296,14 +276,19 @@ async def fetch_blobs(self, buckets):
...
+            try:
+                async for blob in self._google_storage_client.api_call(
+                    resource="objects",
+                    method="list",
+                    full_response=True,
+                    bucket=bucket["id"],
+                    userProject=self._google_storage_client.user_project_id,
+                ):
+                    yield blob
+            except Exception as exception:
```
vidok, line 288: Let's handle the access error separately. It will be much easier to find it in the logs and it will look much better in the UI.

PR #1937, opened 2023-11-30
[Mongo] Add Support of SSL

parthpuri-elastic:
## Related to https://github.com/elastic/connectors/issues/1887

Add support of SSL in MongoDB.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -3,8 +3,10 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+import os
 from copy import deepcopy
 from datetime import datetime
+from tempfile import NamedTemporaryFile
 
 import fastjsonschema
 from bson import DBRef, Decimal128, ObjectId
@@ -16,6 +18,7 @@ from connectors.filtering.validation import (
     SyncRuleValidationResult,
 )
 from connectors.source import BaseDataSource, ConfigurableFieldValueError
+from connectors.utils import get_pem_format
 
 
 class MongoAdvancedRulesValidator(AdvancedRulesValidator):
@@ -98,6 +101,9 @@ class MongoDataSource(BaseDataSource):
         host = self.configuration["host"]
         user = self.configuration["user"]
         password = self.configuration["password"]
+        ssl_ca = self.configuration["ssl_ca"]
+        tlsinsecure = self.configuration["tlsinsecure"]
+        self.certfile = ""
 
         if self.configuration["direct_connection"]:
             client_params["directConnection"] = True
@@ -105,6 +111,18 @@ class MongoDataSource(BaseDataSource):
         if len(user) > 0 or len(password) > 0:
             client_params["username"] = user
             client_params["password"] = password
+
+        if self.configuration["ssl_enabled"]:
+            client_params["tls"] = True
+            if ssl_ca:
+                pem_certificates = get_pem_format(key=ssl_ca)
+                with NamedTemporaryFile(mode="w", suffix=".pem", delete=False) as cert:
+                    cert.write(pem_certificates)
+                    self.certfile = cert.name
+                client_params["tlsCAFile"] = self.certfile
+            client_params["tlsInsecure"] = tlsinsecure
+        else:
+            client_params["tls"] = False
 
         self.client = AsyncIOMotorClient(host, **client_params)
 
@@ -145,6 +163,28 @@ class MongoDataSource(BaseDataSource):
                 "type": "bool",
                 "value": False,
             },
+            "ssl_enabled": {
+                "display": "toggle",
+                "label": "SSL/TLS Connection",
+                "order": 7,
+                "type": "bool",
+                "value": False,
+            },
+            "ssl_ca": {
+                "depends_on": [{"field": "ssl_enabled", "value": True}],
+                "label": "Certificate Authority (.pem)",
+                "order": 8,
+                "required": False,
+                "type": "str",
+            },
+            "tlsinsecure": {
+                "display": "toggle",
+                "depends_on": [{"field": "ssl_enabled", "value": True}],
+                "label": "tlsInsecure",
+                "order": 9,
+                "type": "bool",
+                "value": False,
+            },
         }
 
     def advanced_rules_validators(self):
@@ -152,6 +192,15 @@ class MongoDataSource(BaseDataSource):
 
     async def ping(self):
         await self.client.admin.command("ping")
+
+    async def close(self):
+        if os.path.exists(self.certfile):
+            try:
+                os.remove(self.certfile)
+            except Exception as exception:
+                self._logger.warning(
+                    f"Something went wrong while removing temporary certificate file. Exception: {exception}"
+                )
 
     # TODO: That's a lot of work. Find a better way
     def serialize(self, doc):
--- a/tests/sources/test_mongo.py
+++ b/tests/sources/test_mongo.py
@@ -21,7 +21,9 @@ from tests.sources.support import create_source
 
 
 @asynccontextmanager
-async def create_mongo_source(database="db", collection="col"):
+async def create_mongo_source(
+    database="db", collection="col", ssl_enabled=False, ssl_ca="", tlsinsecure=False
+):
     async with create_source(
         MongoDataSource,
         host="mongodb://127.0.0.1:27021",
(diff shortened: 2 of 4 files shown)


connectors/sources/mongo.py
```diff
@@ -106,6 +112,18 @@ def __init__(self, configuration):
             client_params["username"] = user
             client_params["password"] = password
 
+        if self.configuration["ssl_enabled"]:
+            client_params["tls"] = True
+            if ssl_ca:
+                pem_certificates = get_pem_format(key=ssl_ca)
+                with NamedTemporaryFile(mode="w", suffix=".pem", delete=False) as cert:
+                    cert.write(pem_certificates)
+                    self.certfile = cert.name
```
artem-shelkovnikov, line 121: Can we also organise cleanup of certificates?

This is gonna spam file system with certificates every time connector class is created - in reality if you have 1 connector configured, you'll have 1 temporary file every 30 seconds -> 720 files per day -> 262800 files per year per connector. That's too much.

connectors/sources/mongo.py
```diff
@@ -145,6 +163,28 @@ def get_default_configuration(cls):
...
+                "value": False,
+            },
+            "ssl_ca": {
+                "depends_on": [{"field": "ssl_enabled", "value": True}],
+                "label": "Certificate Authority (.pem)",
+                "order": 8,
+                "required": False,
+                "type": "str",
+            },
+            "tlsinsecure": {
```
wangch079, line 180: Do you think we need this config?

From the [doc](https://pymongo.readthedocs.io/en/4.6.0/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient):
>tlsInsecure: (boolean) Specify whether TLS constraints should be relaxed as much as possible. Setting `tlsInsecure=True` implies `tlsAllowInvalidCertificates=True` and `tlsAllowInvalidHostnames=True`. Defaults to `False`. **Think very carefully before setting this to `True` as it dramatically reduces the security of TLS.**

connectors/sources/mongo.py
wangch079: Can we pass `delete_on_close=True` to `NamedTemporaryFile`? I wonder, is passing a file the only way? We already have the CA content, can we just pass it to `AsyncIOMotorClient`?

connectors/sources/mongo.py
parthpuri-elastic: For a user who does not have certificate data and is not aware of TLS, they can enable it. In Mongo Comapass, we have this option in the SSL/TLS connection. So, I added it to our connector. but, Yes, we can remove this if it lowers security. so, what do you suggest?

tests/sources/test_mongo.py
```diff
@@ -339,3 +344,19 @@ async def test_validate_config_when_configuration_valid_then_does_not_raise():
 async def test_serialize(raw, output):
     async with create_mongo_source() as source:
         assert source.serialize(raw) == output
+
+
+@pytest.mark.asyncio
+@mock.patch("ssl.SSLContext.load_verify_locations")
+async def test_ssl_validation(mock_ssl):
```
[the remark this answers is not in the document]
parthpuri-elastic: I have added testcases with invalid certificate and valid certificate example.

connectors/sources/mongo.py
```diff
@@ -145,6 +163,28 @@ def get_default_configuration(cls):
...
+                "display": "toggle",
+                "depends_on": [{"field": "ssl_enabled", "value": True}],
+                "label": "tlsInsecure",
```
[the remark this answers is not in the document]
parthpuri-elastic: I have added this label based on Mongo Compass, but yes, we can change it for better understanding.
image

connectors/sources/mongo.py
```diff
@@ -145,6 +163,28 @@ def get_default_configuration(cls):
                 "type": "bool",
                 "value": False,
             },
+            "ssl_enabled": {
```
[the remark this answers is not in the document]
parthpuri-elastic: done

connectors/sources/mongo.py
seanstory: Is the `close` method reliably called everywhere that a DataSource instance is instantiated? I think that's Artem's concern - that every heartbeat/ping/rule-validation instance is not calling `close()`

connectors/sources/mongo.py
```diff
@@ -145,6 +163,31 @@ def get_default_configuration(cls):
...
+                "required": False,
+                "tooltip": "Specifies the root certificate from the Certificate Authority. The value of the certificate is used to validate the certificate presented by the MongoDB instance.",
+                "type": "str",
+            },
+            "tls_insecure": {
+                "display": "toggle",
+                "depends_on": [{"field": "ssl_enabled", "value": True}],
+                "label": "Disable certificate validations",
+                "order": 9,
+                "tooltip": "This option disables certificate validation for TLS/SSL connections to your MongoDB server. We strongly recommend setting this option to 'disable'. Enable it if certificates are missing.",
```
seanstory, line 187:
```suggestion
                "tooltip": "This option disables certificate validation for TLS/SSL connections to your MongoDB server. We strongly recommend setting this option to 'disable'.",
```

I think it's sufficient to say what it does, and to tell them not to do it.

connectors/sources/mongo.py
seanstory: What about making the config "advanced"? 
We've definitely found users who want to be able to ignore certificate validation, especially when just doing POCs on their laptops. But I agree with you, Chenhui, that we should give every signal that this setting should not be adjusted.

connectors/sources/mongo.py
artem-shelkovnikov: @seanstory we need to double check, last time I was checking, call of `validate` on connector never called `close` after validation is ended.

Additionally, all connectors with SSL that I've checked are affected with this resource leakage.

connectors/sources/mongo.py
parthpuri-elastic: For now, I have made this field `advanced`. @wangch079 Please share your reviews about this field.

connectors/sources/mongo.py
wangch079: Hi @parthpuri-elastic , could you investigate whether the temp file can be safely deleted after the `AsyncIOMotorClient` is instantiated?

connectors/sources/mongo.py
```diff
@@ -145,6 +163,32 @@ def get_default_configuration(cls):
...
+                "required": False,
+                "tooltip": "Specifies the root certificate from the Certificate Authority. The value of the certificate is used to validate the certificate presented by the MongoDB instance.",
+                "type": "str",
+            },
+            "tls_insecure": {
+                "display": "toggle",
+                "depends_on": [{"field": "ssl_enabled", "value": True}],
+                "label": "Disable certificate validations",
+                "order": 9,
+                "tooltip": "This option disables certificate validation for TLS/SSL connections to your MongoDB server. We strongly recommend setting this option to 'disable'.",
```
wangch079, line 187: The label is `Disable certificate validations`, and the tooltip says `We strongly recommend setting this option to 'disable'.`, which is confusing.

connectors/sources/mongo.py
```diff
@@ -106,6 +112,18 @@ def __init__(self, configuration):
...
             client_params["password"] = password
 
+        if self.configuration["ssl_enabled"]:
+            client_params["tls"] = True
+            if ssl_ca:
+                pem_certificates = get_pem_format(key=ssl_ca)
+                with NamedTemporaryFile(mode="w", suffix=".pem", delete=False) as cert:
+                    cert.write(pem_certificates)
+                    self.certfile = cert.name
+                client_params["tlsCAFile"] = self.certfile
```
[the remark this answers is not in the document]
parthpuri-elastic: AFAIK `tlsCertificateKeyFile` contains a client key, i.e., "-----BEGIN PRIVATE KEY----- PrivateKey -----END PRIVATE KEY-----". and it validates the same with Mongo, but in SSL Support, we are validating the RootCA of the certificate authority like we have in all database connectors.

moxarth-elastic:
buildkite test this

artem-shelkovnikov (approved):

seanstory:
@praveen-elastic / @parthpuri-elastic / @moxarth-elastic  can you link the Kibana PR that adds these configs for native connectors? See: https://github.com/elastic/connectors/pull/1937#issuecomment-1836775116

PR #1938, opened 2023-11-30
[Salesforce] Add custom objects and fields support via Advanced Sync Rules

moxarth-elastic:
## Relates to https://github.com/elastic/connectors/issues/1859

This PR contains the following Salesforce connector changes:
- Include 1:1 Salesforce object representation in ElasticSearch doc
- Ingest all custom objects and their associated fields
- Ingest specific objects and fields by Advanced Sync Rules

Refer this Advanced Sync Rules discussion documentation for the schema: [link](https://docs.google.com/document/d/1cJqRiQK2qc676hozu58LPIP3SpkAG-VPnrSLDi17D_Y/edit?pli=1)

#### Pre-Review Checklist

## Release Note

This version of the Salesforce connector will change the elastic search document structure. The previous version uses the custom mapping for the salesforce objects but now we do 1:1 representation of Salesforce object in elastic search. The new document structure will be as following:

```
{
  "_id:: <Id>,
  "_timestamp": <LastModifiedDate>,
  "type": <type>,
  "owner": <ownerName>,
  "owner_url": <OwnerURL>,
  "title": <objectName>,
  "description": <description>
  "CreatedDate": <CreatedDate>,
  "Name": <objectName>,
  "Description": <description>,
  ...
}
```


connectors/sources/salesforce.py
```diff
@@ -177,6 +188,96 @@ async def close(self):
...
+            # If FIELDS function is present in SOQL query, LIMIT/OFFSET is used for pagination
+            else:
+                soql_query = self.modify_soql_query(query=query)
+                async for records in self._yield_soql_query_pages_with_fields_function(
+                    soql_query=soql_query
+                ):
+                    for record in records:
+                        yield record
+
+        else:
```
seanstory, line 249: Should this be an `elif` to check for `sosl`, and then add an `else` that raises if it's neither `soql` or `sosl`?

tests/sources/test_salesforce.py
```diff
@@ -1121,6 +1555,138 @@ async def test_get_all_with_content_docs_and_extraction_service(mock_responses):
...
+            TEST_QUERY_MATCH_URL, status=200, payload=SOSL_RESPONSE_PAYLOAD
+        )
+
+        resultant_docs = []
+        async for record, _ in source.get_docs(filtering):
+            resultant_docs.append(record)
+
+
+@pytest.mark.asyncio
+async def test_remote_validation():
```
seanstory, line 1670: Can you add a case where remote validation passes?

tests/sources/test_salesforce.py
```diff
@@ -1121,6 +1555,138 @@ async def test_get_all_with_content_docs_and_extraction_service(mock_responses):
...
+            "SELECT FIELDS(CUSTOM) FROM Account LIMIT 200",
+            "SELECT FIELDS(CUSTOM) FROM Account LIMIT 200 OFFSET 0",
+        ),
+        (
+            "SELECT FIELDS(ALL) FROM Account LIMIT 200 OFFSET 0",
+            "SELECT FIELDS(ALL) FROM Account LIMIT 200 OFFSET 0",
+        ),
+    ],
+)
+async def test_modify_query(soql_query, modified_query):
```
seanstory, line 1577: Can you add a case to check the logic around adding the `LastModifiedDate` clause?

connectors/sources/salesforce.py
```diff
@@ -1179,113 +1139,102 @@ def map_content_document(self, content_document):
...
+        owner_url = f"{self.base_url}/{owner_id}" if owner_id else ""
 
-        feeds = case.get("Feeds", []) or []
-        for feed in feeds:
-            participants.append(feed.get("CreatedBy", {}))
+        return {
+            "_id": _object.get("Id"),
+            "created_at": _object.get("CreatedDate"),
+            "_timestamp": _format_datetime(datetime_=_object.get("LastModifiedDate")),
+            "source": "salesforce",
```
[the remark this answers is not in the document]
moxarth-elastic: Yes, correct. we can remove this metadata.

connectors/sources/salesforce.py
```diff
@@ -402,6 +517,44 @@ async def _yield_non_bulk_query_pages(self, soql_query):
...
+        """loops through SOQL query response pages and yields lists of records"""
+
+        def modify_offset(query, new_offset):
+            offset_pattern = r"OFFSET (\d+)"
+            new_query = re.sub(offset_pattern, f"OFFSET {new_offset}", query)
+
+            return new_query
+
+        url = f"{self.base_url}{QUERY_ENDPOINT}"
+        offset = 200
```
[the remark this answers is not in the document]
moxarth-elastic: we can create a constant for offset value and use it.

connectors/sources/salesforce.py
```diff
@@ -948,213 +1115,6 @@ class SalesforceDocMapper:
     def __init__(self, base_url):
         self.base_url = base_url
 
-    def map_account(self, account):
```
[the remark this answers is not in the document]
moxarth-elastic: Agreed on adding the release note

tests/sources/test_salesforce.py
moxarth-elastic: Added a positive test case 👍

connectors/sources/salesforce.py
```diff
@@ -1179,113 +1139,102 @@ def map_content_document(self, content_document):
...
+            "created_at": _object.get("CreatedDate"),
+            "_timestamp": _format_datetime(datetime_=_object.get("LastModifiedDate")),
+            "source": "salesforce",
+            "type": _object.get("attributes", {}).get("type"),
+            "url": f"{self.base_url}/{_object.get('Id')}",
+            "owner": owner.get("Name"),
+            "owner_url": owner_url,
+            "title": _object.get("Name"),
+            "description": _object.get("Description"),
+            "object": _object,
```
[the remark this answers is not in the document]
seanstory: I think you misunderstand me - I'm suggesting still adding all fields to elasticsearch, just at the top-level instead of in a nested `object` field, and to not include the redundant non-saleseforce fields like `created_at`, `title`, `type`, `owner` and `description` since the salesforce object contains that information already.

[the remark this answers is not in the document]
moxarth-elastic: Oh, got it! You want to merge `_object` to the top level and remove `_object`.

connectors/sources/salesforce.py
```diff
@@ -1179,113 +1140,106 @@ def map_content_document(self, content_document):
...
+        owner_url = f"{self.base_url}/{owner_id}" if owner_id else ""
 
-        email_messages = case.get("EmailMessages", {}) or {}
-        participants.extend(self._collect_created_by(email_messages))
+        doc = {
+            "_id": _object.get("Id"),
+            "_timestamp": _format_datetime(datetime_=_object.get("LastModifiedDate")),
+            "type": _object.get("attributes", {}).get("type"),
+            "url": f"{self.base_url}/{_object.get('Id')}",
+            "owner": owner.get("Name"),
```
seanstory, line 1159:
```suggestion
```

This results in 
```
{
   ...
  "owner" : <name>
  "Owner": {
    "Name": <name>
   }
   ...
}
```

I think we can get rid of the `"owner"` and just leave the built-in `"Owner"` object.

connectors/sources/salesforce.py
```diff
@@ -1179,113 +1140,106 @@ def map_content_document(self, content_document):
...
-        for email in email_messages.get("records", []):
-            names.append(email.get("FromName"))
-            emails.append(email.get("FromAddress"))
-            emails.extend(str(email.get("ToAddress", "")).split(";"))
-            emails.extend(str(email.get("CcAddress", "")).split(";"))
-            emails.extend(str(email.get("BccAddress", "")).split(";"))
+        # remove redundant fields from salesforce object
+        for field in ["attributes", "Id", "Name", "Description", "LastModifiedDate"]:
+            if _object.get(field):
+                del _object[field]
```
seanstory, line 1168:
```suggestion
```

We don't want to delete the properties that come from Salesforce. We want to remove the redundant properties that we're defining in addition to what's coming from salesforce.

connectors/sources/salesforce.py
```diff
@@ -1179,113 +1140,106 @@ def map_content_document(self, content_document):
...
-        email_messages = case.get("EmailMessages", {}) or {}
-        participants.extend(self._collect_created_by(email_messages))
+        doc = {
+            "_id": _object.get("Id"),
+            "_timestamp": _format_datetime(datetime_=_object.get("LastModifiedDate")),
+            "type": _object.get("attributes", {}).get("type"),
+            "url": f"{self.base_url}/{_object.get('Id')}",
+            "owner": owner.get("Name"),
+            "owner_url": owner_url,
+            "title": _object.get("Name"),
```
seanstory, line 1161:
```suggestion
```
We don't need `title` if the object has `Name`

connectors/sources/salesforce.py
```diff
@@ -1179,113 +1140,106 @@ def map_content_document(self, content_document):
...
+            "description": _object.get("Description"),
```
seanstory, line 1162:
```suggestion
```
We don't need `description` if the object has `Description`

connectors/sources/salesforce.py
```diff
@@ -1179,113 +1140,106 @@ def map_content_document(self, content_document):
...
+        owner = _object.get("Owner", {}) or {}
+        owner_id = owner.get("Id")
+        owner_url = f"{self.base_url}/{owner_id}" if owner_id else ""
 
-        email_messages = case.get("EmailMessages", {}) or {}
-        participants.extend(self._collect_created_by(email_messages))
+        doc = {
+            "_id": _object.get("Id"),
+            "_timestamp": _format_datetime(datetime_=_object.get("LastModifiedDate")),
+            "type": _object.get("attributes", {}).get("type"),
```
seanstory, line 1157:
```suggestion
```

we don't need `type` if we have `attributes.type`

seanstory (commented):
Better! Left some suggestions to finish removing duplicate properties but to default to the Salesforce names for fields, rather than our own names.

seanstory (approved):
Thanks!

leemthompo:
@seanstory is this a breaking change?

seanstory:
@leemthompo yes, the structure changes would break any existing search experience or client code that relied on specific field names in the documents. Since this connector was not yet GA, PM determined that such a change was acceptable.

leemthompo:
Cool just wondering if we need to explicitly call it out as such. cc @ppf2

ppf2:
I think it can be helpful to add something to the release notes, maybe under enhancements instead of breaking change (since the connector is not GA)?  

What will happen when upgrading to 8.12?  Will the existing search index fields be migrated to the new schema or will users have to recreate the connector and search index once on 8.12 and re-sync? @seanstory

leemthompo:
cc @danajuratoni

PR #1945, opened 2023-12-04
Ensure that Exceptions with empty error messages at least state the exception

seanstory:
## Closes https://github.com/elastic/connectors/issues/1603

Sometimes, an Exception is raised that has no `message` property. This is common for errors like `NoneTypeError`, `NotFound` or any internal exception we use like:
```
class SomeError(Exception)
  pass
...

raise SomeError() 
```

Our Extractor sets the `fetch_error` to be the instance of the caught exception, and we log it, but the connector sync job was getting updated with `str(e)`, which ends up being the empty string for these types of exceptions.

This PR adds some conditionals to attempt to generate a good error message based on what is passed to the `fail` function.

#### Pre-Review Checklist

## Release Note

Fixes a bug where some sync jobs might fail with an empty error message

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -300,9 +300,17 @@ class SyncJob(ESDocument):
             JobStatus.COMPLETED, None, ingestion_stats, connector_metadata
         )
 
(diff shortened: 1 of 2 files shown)


wangch079 (approved):
Great work 👍

PR #1946, opened 2023-12-04
Fix sharepoint server SSL-related config options

acrewdson:
Currently in the SharePoint Server 2019 connector, SSL certificate verification is disabled by default when configuring the connector. What's more, if you enable cert verification, you also have to provide a self-signed cert.

Coupling these options together does not make sense. We should:

* always verify SSL certs (self-signed or otherwise) by default, unless explicitly disabled
* not require the addition of a self-signed cert just because we want certificate verification enabled; these two configuration options should be configurable independent of one another

**NB:** if this PR is accepted, we'll need to address the docs and configuration reference items in the list below, too

#### Pre-Review Checklist
- [ ] ❗ Considered corresponding documentation changes
- [ ] ❗ Contributed any configuration settings changes to the configuration reference

### DIFF
--- a/connectors/sources/sharepoint_server.py
+++ b/connectors/sources/sharepoint_server.py
@@ -95,10 +95,10 @@ class SharepointServerClient:
         self.site_collections = self.configuration["site_collections"]
 
         self.session = None
-        if self.ssl_enabled and self.certificate:
+        if self.certificate:
             self.ssl_ctx = ssl_context(certificate=self.certificate)
         else:
-            self.ssl_ctx = False
+            self.ssl_ctx = None if self.ssl_enabled else False
 
     def set_logger(self, logger_):
         self._logger = logger_
@@ -480,16 +480,16 @@ class SharepointServerDataSource(BaseDataSource):
             },
             "ssl_enabled": {
                 "display": "toggle",
-                "label": "Enable SSL",
+                "label": "Enable SSL certificate verification",
                 "order": 5,
                 "type": "bool",
-                "value": False,
+                "value": True,
             },
             "ssl_ca": {
-                "depends_on": [{"field": "ssl_enabled", "value": True}],
-                "label": "SSL certificate",
+                "label": "Self-signed SSL certificate",
                 "order": 6,
                 "type": "str",
+                "required": False,
             },
             "retry_count": {
                 "default_value": RETRIES,
--- a/tests/sources/test_sharepoint_server.py
+++ b/tests/sources/test_sharepoint_server.py
@@ -158,13 +158,6 @@ async def test_validate_config_for_ssl_enabled_when_ssl_ca_not_empty_does_not_ra
 
 
 @pytest.mark.asyncio
-async def test_validate_config_for_ssl_enabled_when_ssl_ca_empty_raises_error():
-    async with create_sps_source(ssl_enabled=True) as source:
-        with pytest.raises(ConfigurableFieldValueError):
-            await source.validate_config()
-
-
-@pytest.mark.asyncio
 async def test_api_call_for_exception():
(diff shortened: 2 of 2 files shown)


connectors/sources/sharepoint_server.py
```diff
@@ -480,16 +480,16 @@ def get_default_configuration(cls):
             },
             "ssl_enabled": {
                 "display": "toggle",
-                "label": "Enable SSL",
+                "label": "Enable SSL certificate verification",
                 "order": 5,
                 "type": "bool",
-                "value": False,
+                "value": True,
```
artem-shelkovnikov, line 486: Why True by default? It's always on normally?

connectors/sources/sharepoint_server.py
```diff
@@ -480,16 +480,16 @@ def get_default_configuration(cls):
...
             },
             "ssl_ca": {
-                "depends_on": [{"field": "ssl_enabled", "value": True}],
```
artem-shelkovnikov, line 489: Why remove depends_on?

artem-shelkovnikov (commented):
Sorry for the delay in review - I've asked a couple questions

connectors/sources/sharepoint_server.py
acrewdson: yep, it's a bad practice to _not_ verify certificates – you'd only want to do that under special circumstances, otherwise you're circumventing one of the core benefits of what TLS provides

connectors/sources/sharepoint_server.py
acrewdson: Because right now, with that `depends_on`, if you want to enable SSL then you must also provide your own cert (see screenshot below), which to me does not make sense. A quite common scenario would be (1) enable SSL but (2) do not provide your own cert.

I'm trying in this PR to decouple those two things, since I don't understand why this connector was written in such a way that providing your own cert is coupled to also having SSL enabled.

image

artem-shelkovnikov:
@elastic/ingestion-team - since this change changes existing connector fields, what's our procedure here? Do we create migrations manually? Do we need to implement migration procedure for RCF updates?

seanstory:
> @elastic/ingestion-team - since this change changes existing connector fields, what's our procedure here? Do we create migrations manually? Do we need to implement migration procedure for RCF updates?

I'd thought this would "just work", but reading through our code, I think we currently only handle adding new fields in configs or wholly net-new configs. But changing the values of existing fields in the configs isn't something we'll automagically address.

If we need these changes to apply for old connectors, I think the way to do that today is to add a migration in ent-search. :( 

I don't think we want to automatically align the `.elastic-connectors` record with whatever the source code says should be in these fields - what if a customer explicitly changed a value on purpose? 🤔 I could be convinced otherwise, though.

artem-shelkovnikov:
In this case I think the change needs a migration change + kibana change merged approximately together to make it work, right?

artem-shelkovnikov (commented):
I don't have objections to merging it, there are a couple small things to do/test:

1. Test that migration from previous version to version with this change works as expected, updating the configurable fields. It _should_ work, but it's best to test
2. Create a PR in Kibana to also change these fields and merge somewhat simultaneously to avoid weird problems

acrewdson:
@artem-shelkovnikov thanks Artem 👍 I don't have the bandwidth right now to return to this, but could possibly return to at some point next month. Though I'd probably only want to devote further time to it if you or the team feel it's a valuable fix, _ie_ that we can identify a stronger motivation to get it merged than just the absence of objections 😅 🙏

acrewdson:
I think this PR is pretty stale now, and the changeset may be too out of date to be helpful. Closing, but please feel free to re-open.

PR #1947, opened 2023-12-04
Working Docker Compose Stack for Connectors

markjhoy:
This PR provides a set of example scripts and Docker compose configurations for creating a sellf-managed Elasticsearch, Kibana, and Connectors stack. Provisions are made within the code to also prompt for Connectors configuration using the Connectors CLI.

For documentation and options, see the included [README.md file](scripts/stack/README.md)

#### Pre-Review Checklist


scripts/stack/README.md
```diff
@@ -0,0 +1,106 @@
+# Full Elastic Stack with Connectors
```
artem-shelkovnikov, line 1: Might be worth to mention this page on the main repo README.md

scripts/stack/README.md
```diff
@@ -0,0 +1,106 @@
...
+* [Prerequisites](#prerequisites)
+* [Running the Stack](#running-the-stack)
+* [Prompted Configuration](#prompted-configuration)
+* [Manual Configuration](#manual-configuration)
+* [Script command reference](#script-command-reference)
+  * [run-stack.sh](#run-stacksh)
+  * [stop-stack.sh](#stop-stacksh)
+
+## Prerequisites
+* Linux or MacOS (Windows is not currently supported)
```
artem-shelkovnikov, line 18: Curious - why Windows not supported? I think some ppl run connectors container on windows

scripts/stack/connectors-config/config.yml
```diff
@@ -0,0 +1,226 @@
+## ================= Elastic Connectors Configuration ==================
```
artem-shelkovnikov, line 1: I personally really dislike that now we have 3 config files in the repository - is there a way to avoid this?

Additionally, we don't follow the `config.yml.example` pattern with our regular config, why have one here?

scripts/stack/README.md
```diff
@@ -0,0 +1,106 @@
...
+    1. Run `./scripts/stack/run-stack.sh --no-connectors` to run the full stack without the Connectors service
+    2. From the main menu, go to "Content", "Indices", and create a new index for your connector
+    3. Choose the "Connector" index type
+    4. Select the connector type
+    5. Set the index name
+    6. Generate a new API Key, and copy it somewhere safe
+    7. Set Connector name and description (optional)
+    8. Copy the resulting connector_id, service_type, and api_key into a new entry in the `./scripts/stack/connectors-config/config.yml` fille
+    9. Repeat as necessary for all the connectors you wish too create
+2. when complete, temporarily stop the stack via `./scripts/stack/stop-stack.sh`
```
artem-shelkovnikov, line 70: Is there also an option to run stack first, and then run connectors second without having to restart whole stack? It feels very counter-intuitive to me, that to run lightweight connectors docker I need to restart Kibana and Elasticsearch

scripts/stack/configure-stack.sh
```diff
@@ -0,0 +1,56 @@
...
+
+PYTHON_EXECUTABLE=""
+
+if which python3 > /dev/null; then
+    PY_VERSION=`python3 --version`
+    PYTHON_EXECUTABLE="python3"
+elif which python > /dev/null; then
+    PY_VERSION=`python --version`
+    PYTHON_EXECUTABLE="python"
+fi
```
artem-shelkovnikov, line 18: We actually don't run from system python, rather from venv version. Do you think it can cause trouble?

scripts/stack/configure-stack.sh
```diff
@@ -0,0 +1,56 @@
...
+    fi
+
+    CONNECTORS_EXE="${PROJECT_ROOT}/bin/connectors"
+    if [ ! -f "$CONNECTORS_EXE" ]; then
+        echo "Could not find a connectors executable, running 'make clean install'"
+
+        if [ $PYTHON_EXECUTABLE == "" ]; then
+            echo "Could not find a suitable Python 3 executable..."
+            exit 2
+        fi
```
artem-shelkovnikov, line 32: Should this happen before everything else?

scripts/stack/docker/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+
+networks:
+  connectors_stack_net:
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:${ELASTICSEARCH_VERSION}
+    environment:
+      - "discovery.type=single-node"
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
```
artem-shelkovnikov, line 11: I think 512m is too small for us - we often have problems even with 1GB deployment. Can we make it 2GB here?

scripts/stack/run-stack.sh
```diff
@@ -0,0 +1,67 @@
...
+export CURDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+if ! which docker-compose > /dev/null; then
+    echo "Could not find 'docker-compose'. Make sure it is installed and available via your PATH"
+    exit 2
+fi
+
+pushd "$CURDIR"
+
+export SECURE_STATE_DIR="$(mktemp -d)"
+trap 'rm -rf -- "$SECURE_STATE_DIR"' EXIT
```
artem-shelkovnikov, line 14: What does it do?

scripts/stack/configure-stack.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
artem-shelkovnikov, line 1: While it's called `configure-stack` it only configures connectors, right? Does it make sense to rename it to reflect this?

scripts/stack/README.md
markjhoy: Windows is not supported for running these particular scripts, but it is for the Connectors service... I'll clarify this. Thanks!

scripts/stack/configure-stack.sh
markjhoy: This is only used if a `make clean install` is needed (if the user never built the `bin/connectors` from the source) - which, the `make` will create a venv when it does its magic.

scripts/stack/docker/docker-compose.yml
markjhoy: Thanks for the tip. Will do. I'll also update the README to reflect these constraints too.

scripts/stack/run-stack.sh
markjhoy: Ah - thanks for pointing this out... this was some old code that set up a temporary directory for saving configurations that would be automatically deleted upon the script exit... but, it's not needed anymore. I'll remove this.

scripts/stack/configure-stack.sh
markjhoy: Yep - good point. Thanks!

scripts/stack/configure-stack.sh
markjhoy: This is only needed in the case of it needing to run `make clean install`... so, it's not needed overall.

scripts/stack/connectors-config/config.yml
markjhoy: Ah - good point. Since I have it set for the Connectors-CLI to modify the config file in place, I copied my own version of the config... _but_ I can check and copy the regular config and not keep my own... thanks for this.

scripts/stack/README.md
markjhoy: I can add in a `--connectors-only` option that should cover this case... thanks

scripts/stack/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Prerequisites
+* Linux or MacOS (Although Windows can run the Connectors service, it is not currently supported via these scripts)
+* Docker with Docker Compose Installed
+* It is recommended to run Docker with at least 4GB of available RAM.
+
+## Running the Stack
+
+If you do not wish to use the default Elasticsearch username and password, you must set the username and password (or the API key)
+in the default [config.yml](./config.yml) file before running the `run-stack.sh` script. You can also
```
wangch079, line 25: I think the `config.yml` links to a wrong path

scripts/stack/README.md
```diff
@@ -0,0 +1,108 @@
...
+the configurator, it will take you through a set of prompts using the
+[Connectors CLI](../../connectors/connectors_cli.py) to create a new index and connector.
+
+The resulting configuration will be saved to the [config.yml](./connectors-config/config.yml)
+file that is used to start the Connectors container. Note that using the configurator,
+it will modify the [connectors config.yml](./connectors-config/config.yml) file. You can restore it
+by copying over the [example config.yml.example](./connectors-config/config.yml.example) file
+to `config.yml` in the directory.
+
+If you configured a new connector, you will need to go into Kibana to start or schedule a sync.
```
wangch079, line 46: I think CLI is designed to be an alternative to Kibana, and we can document how to schedule/trigger a sync via Connectors CLI, and make Kibana optional (e.g. only start Kibana if `-k` is specified)

scripts/stack/update-kibana-user-password.sh
```diff
@@ -0,0 +1,20 @@
...
+    export CURDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+fi
+
+if [[ "${ELASTIC_PASSWORD:-}" == "" ]]; then
+    source $CURDIR/read-env.sh $CURDIR/.env
+fi
+
+echo "Updating Kibana password in Elasticsearch running on $ELASTICSEARCH_URL"
+change_data="{ \"password\": \"${ELASTIC_PASSWORD}\" }"
+curl -u elastic:$ELASTIC_PASSWORD "$@" -X POST "${ELASTICSEARCH_URL}/_security/user/kibana_system/_password?pretty" -H 'Content-Type: application/json' -d"${change_data}"
```
wangch079, line 20: This tries to change the password of `kibana_system` to `$ELASTIC_PASSWORD` but authenticate user `elastic` with password `$ELASTIC_PASSWORD`?

scripts/stack/.env
```diff
@@ -0,0 +1,3 @@
+ELASTICSEARCH_VERSION=8.11.2
+KIBANA_VERSION=8.11.2
+CONNECTORS_VERSION=8.11.2.0
```
wangch079, line 3: Do we want to use `SNAPSHOT` here?

scripts/stack/run-stack.sh
```diff
@@ -0,0 +1,76 @@
...
+
+export CURDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+if ! which docker-compose > /dev/null; then
+    echo "Could not find 'docker-compose'. Make sure it is installed and available via your PATH"
+    exit 2
+fi
+
+pushd "$CURDIR"
+
+# Make sure ES_VERSION and ENT_VERSION are set
```
wangch079, line 13:
```suggestion
# Make sure ELASTICSEARCH_VERSION, KIBANA_VERSION and CONNECTORS_VERSION are set
```

scripts/stack/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+* [Prerequisites](#prerequisites)
+* [Running the Stack](#running-the-stack)
+* [Prompted Configuration](#prompted-configuration)
+* [Manual Configuration](#manual-configuration)
+* [Script command reference](#script-command-reference)
+  * [run-stack.sh](#run-stacksh)
+  * [stop-stack.sh](#stop-stacksh)
+
+## Prerequisites
```
oli-g, line 17: We should include Python here, as the bash script invokes it locally (not in container process).

scripts/stack/configure-connectors.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
oli-g, line 1: Should we move this script into `docker.elastic.co/enterprise-search/elastic-connectors` image and thus run this initialization inside the container, as an entrypoint? Docker would be the only pre-requisite on the local machine, in this case.

I read it seems we need an entrypoint (see https://github.com/elastic/enterprise-search-team/issues/6469), so we may think about adding a "smart" entrypoint, which could also generate connectors configuration before starting them (in case the container is created by a human user).

Thoughts?

scripts/stack/docker/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+  kibana:
+    image: docker.elastic.co/kibana/kibana:${KIBANA_VERSION}
+    ports:
+      - 5601:5601
+    depends_on:
+      - elasticsearch
+    networks:
+      - connectors_stack_net
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
oli-g, line 37: Probably we no longer need this, do we?

scripts/stack/docker/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  elastic-connectors:
+    image: docker.elastic.co/enterprise-search/elastic-connectors:${CONNECTORS_VERSION}
+    depends_on:
+      - elasticsearch
+      - kibana
+    volumes:
+      - ${CURDIR}/connectors-config:/config
+    command: /app/bin/elastic-ingest -c /config/config.yml
+    network_mode: "host"
```
oli-g, line 53: Could you clarify why we need this? I think we no longer need it, but I may be wrong.

markjhoy: This is needed as internally, the connectors config can use the other service names to communicate (e.g. "elasticsearch"), however, since we use the same configuration for configuring via the script as what is used in the container itself, we need this to be "localhost:9200" for communicating with the Elasticsearch node - having this makes it possible to use "localhost:9200" from both within the container as well as from our scripts.

scripts/stack/docker/docker-compose.yml
markjhoy: Hmmm... when I had testing things yesterday, Kibana wasn't able to communicate to Elasticsearch without out... just tried again now, and it works without it... 🤷  -- It can be removed for that.

scripts/stack/.env
markjhoy: Probably best -- I had used 8.11 to test with, but for the main - that's a good idea... I was also thinking about tying this to the ./connectors/VERSION file as well

scripts/stack/README.md
markjhoy: True, but these scripts are meant to provide a quick gateway to getting a Connectors framework up and running, and thus, the usage of the CLI is very "lite" here and only used for configuration - but the instructions in this particular part is for if the user want to manually set things up... I'll add in some extra documentation though around the CLI if the user wants to take that route.

scripts/stack/update-kibana-user-password.sh
markjhoy: This is just to keep things consistent when running the stack.

artem-shelkovnikov:
@markjhoy the change looks good, but I see that CI is red - I think just running `make autoformat` will fix it and then we can stamp approval :)

markjhoy:
> @markjhoy the change looks good, but I see that CI is red - I think just running `make autoformat` will fix it and then we can stamp approval :)

Thanks @artem-shelkovnikov - should be all set now.

acrewdson (commented):
@markjhoy awesome work, I kicked the tires locally and it was really smooth 💯 

added a few comments/suggestions, all very minor

scripts/stack/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+The scripts used a default password for the `elastic` user when running the scripts.
+If you wish to use a different password for the `elastic` user, you can specify the environment variable `ELASTIC_PASSWORD` before running the script. E.g.:
+
+```bash
+$ ELASTIC_PASSWORD="my_new_password" ./scripts/stack/run-stack.sh
+```
+
+Alternatively, if you do not wish to a different set of credentials, or an API key for authentication, you can run the `./copy-config.sh` script to create a
+copy of the default [config.yml](../../config.yml) file, and edit the resulting file that will be created in the `scripts/stack/connectors-config` folder.
```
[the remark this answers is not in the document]
markjhoy: Oooo - good catch. This was originally when it used a copy of the config when starting before I opted to use the single / default config and copy that in... will fix.

scripts/stack/README.md
```diff
@@ -0,0 +1,130 @@
...
+    8. Copy the resulting connector_id, service_type, and api_key into a new entry in the `./scripts/stack/connectors-config/config.yml` fille
+    9. Repeat as necessary for all the connectors you wish too create
+3. run `./scripts/stack/run-stack.sh --no-configuration --connectors-only` again to run the Connectors service without prompting for configuration. Once running:
+    1. From the main menu, choose "Content" under the "Search" heading
+    2. Click on your index name
+    3. Click on the "Configuration" tab
+    4. Complete the configuration and optionally start your sync
+    5. Repeat these steps for all the connectors you created
+
+## Script command reference
```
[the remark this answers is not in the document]
markjhoy: Good idea -- I don't think we need another script for this as that might muddy the waters, but a mention of how to view the logs will be useful.

scripts/stack/README.md
```diff
@@ -0,0 +1,130 @@
...
+the configurator, it will take you through a set of prompts using the
+[Connectors CLI](../../connectors/connectors_cli.py) to create a new index and connector.
+
+The resulting configuration will be saved to the [config.yml](./connectors-config/config.yml)
+file that is used to start the Connectors container. Note that using the configurator,
+it will modify the [connectors config.yml](./connectors-config/config.yml) file. You can restore it
+by copying over the [example config.yml.example](./connectors-config/config.yml.example) file
+to `config.yml` in the directory.
+
+If you configured a new connector, you will need to go into Kibana to start or schedule a sync.
```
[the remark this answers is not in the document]
acrewdson: I still think it'd be good to explain how to login to Kibana

scripts/stack/README.md
```diff
@@ -0,0 +1,134 @@
...
+You can also specify an alternative base version by setting the environment variable `CONNECTORS_VERSION` before running the script. e.g.:
+
+```bash
+CONNECTORS_VERSION=8.11.2.0 ./scripts/stack/run-stack.sh --no-snapshot
+```
+
+When running the script, the Elasticsearch and Kibana versions will use the same stack version as the `CONNECTORS_VERSION`, however, you can additionally specify these before running:
+
+```bash
+ELASTICSERACH_VERSION=8.11.2 KIBANA_VERSION=8.11.2 CONNECTORS_VERSION=8.11.2.0 ./scripts/stack/run-stack.sh --no-snapshot
```
acrewdson, line 46:
```suggestion
ELASTICSEARCH_VERSION=8.11.2 KIBANA_VERSION=8.11.2 CONNECTORS_VERSION=8.11.2.0 ./scripts/stack/run-stack.sh --no-snapshot
```

markjhoy: ouch. Great catch 😁

acrewdson (dismissed):
🏆

seanstory (dismissed):
LGTM. I left a bunch of questions/comments, but totally fine if you want to address them in a follow-up. I tested locally, and things largely seem to work, and I'm very excited to have this available to use. 🦙 ❗

scripts/stack/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Prerequisites
+* Linux or macOS (Although Windows can run the Connectors service, it is not currently supported via these scripts)
+* Python 3.10 or higher
+* Docker with Docker Compose Installed
+* It is recommended to run Docker with at least 4GB of available RAM.
+
+## Running the Stack
+
+The scripts use a default password for the `elastic` user, however, if you wish to use a different password, you can specify the environment variable `ELASTIC_PASSWORD` before running the script. E.g.:
```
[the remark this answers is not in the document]
markjhoy: I think it's OK as mentioned for a POC, but I will add a note about this in a production environment and some cautions around that (and suggestions for avoiding it)

scripts/stack/README.md
```diff
@@ -0,0 +1,135 @@
...
+```bash
+ELASTIC_PASSWORD="my_new_password" ./scripts/stack/run-stack.sh
+```
+
+Alternatively, if you do not wish to use a different set of credentials, or an API key for authentication, you can run the `./copy-config.sh` script to create a
+copy of the default [config.yml](../../config.yml) file, and edit the resulting file that will be created in the `scripts/stack/connectors-config` folder.
+
+By default, these Connector scripts will use a `SNAPSHOT` version of the Docker image.
+The version is defined in the [VERSION file](../../connectors/VERSION).
+If you do not wish to use a `SNAPSHOT` version, add the `--no-snapshot` flag when you run the script.
```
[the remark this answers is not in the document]
markjhoy: it just removes the `-SNAPSHOT` tag at the moment... I have some trickery I can use to check for release versions if they are available... I might be able to add that in.

scripts/stack/README.md
```diff
@@ -0,0 +1,135 @@
...
+You can also specify an alternative base version by setting the environment variable `CONNECTORS_VERSION` before running the script. e.g.:
+
+```bash
+CONNECTORS_VERSION=8.11.2.0 ./scripts/stack/run-stack.sh --no-snapshot
+```
+
+When running the script, the Elasticsearch and Kibana versions will use the same stack version as the `CONNECTORS_VERSION`, however, you can additionally specify these before running:
+
+```bash
+ELASTICSEARCH_VERSION=8.11.2 KIBANA_VERSION=8.11.2 CONNECTORS_VERSION=8.11.2.0 ./scripts/stack/run-stack.sh --no-snapshot
```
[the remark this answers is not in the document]
markjhoy: Originally, yes, but to be more consistent with the rest of the Connectors repository, I opted to infer items from the `VERSION` file

scripts/stack/README.md
```diff
@@ -0,0 +1,135 @@
...
+Alternatively, if you do not wish to use a different set of credentials, or an API key for authentication, you can run the `./copy-config.sh` script to create a
+copy of the default [config.yml](../../config.yml) file, and edit the resulting file that will be created in the `scripts/stack/connectors-config` folder.
+
+By default, these Connector scripts will use a `SNAPSHOT` version of the Docker image.
+The version is defined in the [VERSION file](../../connectors/VERSION).
+If you do not wish to use a `SNAPSHOT` version, add the `--no-snapshot` flag when you run the script.
+You can also specify an alternative base version by setting the environment variable `CONNECTORS_VERSION` before running the script. e.g.:
+
+```bash
+CONNECTORS_VERSION=8.11.2.0 ./scripts/stack/run-stack.sh --no-snapshot
```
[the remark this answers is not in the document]
markjhoy: For the other items in the stack if needed (Elasticsearch and Kibana)... I'll make this more clear...

scripts/stack/README.md
```diff
@@ -0,0 +1,135 @@
...
+### stop-stack.sh
+
+This script is used to stop the stack, and optionally remove any data volumes and reset the configuration file.
+
+```bash
+./scripts/stack/stop-stack.sh
+```
+
+Command line options:
+* `-v | --remove-volumes`: delete all data volumes on stop
```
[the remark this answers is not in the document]
markjhoy: I can change that - but for the end user, it _should_ be ok (as they won't be using the `ent-search` stack scripts... but to be sure, I'll rename them for Connectors

scripts/stack/README.md
```diff
@@ -0,0 +1,135 @@
...
+    8. Copy the resulting connector_id, service_type, and api_key into a new entry in the `./scripts/stack/connectors-config/config.yml` fille
+    9. Repeat as necessary for all the connectors you wish too create
+3. run `./scripts/stack/run-stack.sh --no-configuration --connectors-only` again to run the Connectors service without prompting for configuration. Once running:
+    1. From the main menu, choose "Content" under the "Search" heading
+    2. Click on your index name
+    3. Click on the "Configuration" tab
+    4. Complete the configuration and optionally start your sync
+    5. Repeat these steps for all the connectors you created
+
+## Script command reference
```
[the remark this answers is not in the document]
markjhoy: @acrewdson had mentioned about adding a script to conveniently accessing the logs earlier - since you're the second person who mentioned along these lines, I'll add one... but basically, you can do:
```
docker-compose -f ./scripts/stack/docker/docker-compose.yml logs -f elastic-connectors
```
As the container name ("elastic-connectors") should be able to be used as an alias for the id

scripts/stack/README.md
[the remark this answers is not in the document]
seanstory: Totally fine if it's a follow-up enhancement. But I just don't see a realistic scenario where you're on X.Y branch with X.Y.Z-SNAPSHOT in the `VERSION` file, and you want to run (and have access to) X.Y.Z as opposed to X.Y-1 or X.Y.Z-1

scripts/stack/README.md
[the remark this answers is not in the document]
seanstory: I ended up googling and finding that `docker logs docker-elastic-connectors-1 --follow` worked, too 🤷

markjhoy:
This looks good now. Thanks again @acrewdson and @seanstory for all your reviews and comments... going to merge when ready

acrewdson (approved):

PR #1953, opened 2023-12-05
Add some logging to github connector for debugging

seanstory:
## related to https://github.com/elastic/sdh-enterprise-search/issues/1237

During a support request I noticed that the customer was having to add their own debug log statements, because our connector didn't have much. This should help

#### Pre-Review Checklist


wangch079 (dismissed):

connectors/sources/github.py
```diff
@@ -711,6 +712,9 @@ async def post(self, query_data, need_headers=False):
             dictionary: Client response
         """
         url = f"{self.github_url}/graphql"
+        self._logger.debug(
+            f"Sending POST to {url} with body: '{json.dumps(query_data)}'"
+        )
```
artem-shelkovnikov, line 717: Is there a chance to leak anything dangerous here?

connectors/sources/github.py
```diff
@@ -1282,6 +1290,7 @@ async def _get_org_repos(self):
             yield repo_object
 
     async def _get_configured_repos(self, configured_repos):
+        self._logger.info(f"Fetching configured repos: '{configured_repos}'")
```
artem-shelkovnikov, line 1293: Might be worth to put it one line lower and log per repo name?

connectors/sources/github.py
```diff
@@ -1436,6 +1447,9 @@ async def _fetch_pull_requests(
         response_key=(REPOSITORY_OBJECT, "pullRequests"),
         filter_query=None,
     ):
+        self._logger.info(
```
artem-shelkovnikov, line 1450: What's a `response_key`? Is it a token?

seanstory: I think it's which fields from the response object we want back? I dont' really follow the logic and am not familiar with the graphql github API, but I wanted to log everything relevant for us to be able to follow the logic flow and re-create the request if necessary (if DEBUG logs aren't enabled)

connectors/sources/github.py
seanstory: Shouldn't be. This won't have credentials - that's in the Authorization header. And it would just have file paths and repo names and file types and such - so identifiers, but not data contents

seanstory:
@artem-shelkovnikov can I get an approval if you're ok with my comments?

PR #1955, opened 2023-12-06
[8.12] Bump stack version

saarikabhasi:
Related to [RM tasks](https://github.com/elastic/enterprise-search-team/issues/5971)

Bump stack versions


.backportrc.json
```diff
@@ -1,6 +1,7 @@
 {
   "targetBranchChoices": [
     { "name": "main", "checked": true },
+    "8.12",
     "8.11",
     "8.10",
```
wangch079, line 6: I think `8.9` and `8.10` can be removed, as there's no planned minor release.

`pre-8.10-stable` as well? Not 100% sure but I remember this is a label for some SPO features backported to 8.10. Can we remove this as well? cc @elastic/ingestion-team

wangch079 (approved):

.backportrc.json
wangch079: Approved the PR. This can be addressed in a separate PR.

PR #1992, opened 2023-12-19
Ignore errors when Sink or Extractor is forcibly canceled

wangch079:
### What's the change

1. Changed the log level to `warning` when a sync job is forcibly canceled
2. Ignore any error thrown in `Sink` or `Extractor` class if the `Sink` or `Extractor` object is forcibly canceled

### Why the change

In the last release, we [Forcibly cancel a sync job after waiting for 5 seconds](https://github.com/elastic/connectors/pull/1683). This should not be treated as `ERROR`, therefore we want to log it at `warning` level.

When we forcibly cancel the `Sink` or `Extractor` task, it will throw a `ForceCanceledError`. But sometimes there can be other errors thrown due to task cancellation, and be caught first. Before we log those errors as `critical`, we want to check if the task is forcibly canceled. If so, we will ignore those errors, and instead log a `warning` message that the task is forcibly canceled.

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -195,6 +195,7 @@ class Sink:
                     f"Sink did not stop within {CANCELATION_TIMEOUT} seconds of cancelation, force-canceling the task."
                 )
                 return
+            raise
 
     async def _run(self):
         """Creates batches of bulk calls given a queue of items.
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -189,10 +189,12 @@ class Sink:
         except asyncio.CancelledError:
             self._logger.info("Task is canceled, stop Sink...")
             raise
-        except ForceCanceledError:
-            self._logger.error(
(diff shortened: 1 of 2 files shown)


connectors/es/sink.py
```diff
@@ -189,10 +189,12 @@ async def run(self):
...
-        except ForceCanceledError:
-            self._logger.error(
-                f"Sink did not stop within {CANCELATION_TIMEOUT} seconds of cancelation, force-canceling the task."
-            )
+        except Exception as e:
+            if isinstance(e, ForceCanceledError) or self._canceled:
+                self._logger.warning(
+                    f"Sink did not stop within {CANCELATION_TIMEOUT} seconds of cancelation, force-canceling the task."
+                )
+                return
```
artem-shelkovnikov, line 197: I think you need to re-raise here, if error is not ForceCancelledError or self._canceled

wangch079: Good catch

artem-shelkovnikov (approved):

seanstory (approved):
Thanks for the quick fix!

PR #1993, opened 2023-12-19
Make test fixture retry 429 errors

artem-shelkovnikov:

This PR changes functional tests fixture to handle 429 errors from Elasticsearch. This change for now is isolated to fixture.py. In subsequent PRs I'm planning to do the following:

1. Make retry a part of ESClient class and retry transparently
2. Get rid of places calling `ESClient.client` and make `ESClient` a rich client.

#### Pre-Review Checklist

### DIFF
--- a/tests/sources/fixtures/fixture.py
+++ b/tests/sources/fixtures/fixture.py
@@ -5,28 +5,31 @@
 #
 # ruff: noqa: T201
 # ruff: noqa: T203
+import asyncio
+import functools
 import importlib
 import importlib.util
 import logging
 import os
 import pprint
 import signal
-import functools
 import sys
 import time
-import asyncio
 from argparse import ArgumentParser
 
 import requests
 from elastic_transport import ConnectionTimeout
-from elasticsearch import ApiError, AsyncElasticsearch, ConflictError
+from elasticsearch import ApiError
 from requests.adapters import HTTPAdapter, Retry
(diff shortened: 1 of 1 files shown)


tests/sources/fixtures/fixture.py
```diff
@@ -76,62 +84,118 @@ def _parser():
...
 
-                logger.info(f"{CONNECTORS_INDEX} not present, waiting...")
-                time.sleep(1)
-            else:
-                logger.info(f"{CONNECTORS_INDEX} detected")
-                break
+@retrying_transient_errors()
+async def _fetch_connector_metadata(es_client):
+    # There's only one connector, so we just fetch it always
+    response = await es_client.client.search(index=CONNECTORS_INDEX, size=1)
```
seanstory, line 149: this may come back to bite us one day. What if we did `size=2` and raise an exception if `response["hits"]["hits"].size > 1`? That way we can track down where this assumption lives if/when we break it.

seanstory (approved):
LGTM. This is a good fix.

tests/sources/fixtures/fixture.py
artem-shelkovnikov: Yeah I can do this, I tried just keeping original code without modifications in this PR (thus, not fixing such things), but I think this one is small enough to address

vidok (approved):
Awesome fix!

PR #1999, opened 2023-12-20
Refactor ES Client methods

artem-shelkovnikov:

Current structure of our classes that call Elasticsearch is neat, but there are some minor problems here and there as I see them. This PR attempts to refactor out these problems.

It contains several connected changes, might make it a bit harder to review, but they all follow the same logic.

Changes are:

### Change 1: Instead of having one ESClient, there's now additionally ESManagementClient.

A lot of classes inherit from ESClient, for example ConnectorIndex class. ConnectorIndex class is used to do certain operations on connectors - get all connectors, read one connector, get supported connectors, etc.

ESClient, however, had methods that are touching things that logically should not be available for _specialized_ classes that inherit from it, for example this method:

```
async def delete_indices(self, indices):
    # logic here
```

It seems illogical that `ConnectorIndex` class that's a rich wrapper for operations with connectors index can have access to deleting arbitrary indices.

Based on that logic, I've moved all methods that allow for arbitrary index management to `ESManagementClient` class - now it's responsible for management operations - create indices if they don't exit, clean their content, delete indices, etc.

### Change 2: Move ESClient.get_connector_doc to ConnectorIndex.get_connector_by_index

It's a logical change as this is a connector access operation, it does not make sense to have it in `ESClient` class that's a base class for lots of other non-connector-specific classes

### Change 3: ensure_exists method does not create/delete documents now, just uses built-in client method

Self-explanatory - instead of creating indices with `PUT {_index}/_doc` and then deleting this doc, we can just call `client.indices.create` method.

This change could have been separated, but I squeezed it here, sorry.

All other changes are pretty much code/test changes to accomodate the above 3 changes.

#### Pre-Review Checklist

### DIFF
--- a/connectors/services/job_cleanup.py
+++ b/connectors/services/job_cleanup.py
@@ -7,6 +7,7 @@
 A task periodically clean up orphaned and idle jobs.
 """
 
+from connectors.es.client import ESManagementClient
 from connectors.es.index import DocumentNotFoundError
 from connectors.logger import logger
 from connectors.protocol import ConnectorIndex, SyncJobIndex
@@ -27,6 +28,7 @@ class JobCleanUpService(BaseService):
     async def _run(self):
         logger.debug("Successfully started Job cleanup task...")
         self.connector_index = ConnectorIndex(self.es_config)
+        self.es_management_client = ESManagementClient(self.es_config)
         self.sync_job_index = SyncJobIndex(self.es_config)
 
         try:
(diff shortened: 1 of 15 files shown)


connectors/services/job_cleanup.py
```diff
@@ -27,6 +28,7 @@ def __init__(self, config):
     async def _run(self):
         logger.debug("Successfully started Job cleanup task...")
         self.connector_index = ConnectorIndex(self.es_config)
+        self.es_management_client = ESManagementClient(self.es_config)
```
vidok, line 31: We use different variable (`es_client`) names in other places. It's a nit request but I would make it the same for all the classes.

vidok (approved):
LGTM

connectors/services/job_cleanup.py
artem-shelkovnikov: So use `es_client` for `ESManagementClient` too, or have `es_management_client` variable?

vidok: I don't have a strong opinion on that :). I think `es_management_client`  is a little more explicit so I would use it but it's up to you. I'm fine with `es_client` too.

artem-shelkovnikov: Makes sense! I've addressed this in https://github.com/elastic/connectors/pull/1999/commits/bf1ef636637bbe307d2e81d8b0b1237db02b7a4c

vidok (approved):
Great work!

PR #2000, opened 2023-12-20
Introduce automated testing CLI

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/5881

This PR will add a new way of running functional tests. 

To give it a try follow the [installation guide](https://github.com/elastic/connectors/blob/dmitrii/5881-automated-testing/scripts/testing/README.md#L41) and run a sample test. 

#### Pre-Review Checklist

### DIFF
--- a/scripts/testing/README.md
+++ b/scripts/testing/README.md
@@ -2,7 +2,7 @@
 
 ### Installation
 
-Atomated testing CLI depends on `gcloud` and `vault` CLIs so you need to install them. 
+Automated testing CLI depends on `gcloud` and `vault` CLIs so you need to install them. 
 
 #### Google cloud CLI
 Follow the [offical installation](https://cloud.google.com/sdk/docs/install) guide. Once installed, run `gcloud auth login` and authorize using SSO. 
@@ -52,11 +52,11 @@ To automatically delete the VM you need to use `--delete` option.
 `./bin/test-connectors run-test my-testing-environment-name --es-version 8.12-SNAPSHOT --connectors-ref 8.12 --test-case {path to the test case file} --delete`
 
 #### Using different machine type
-All new VMs are based on a predefined image which is in turn based on `ubuntu-2204-lts` with python3 and docker installed. Cusome images are not supported. You can change a machine type by providing `--vm-type`. Visit [the official GCP documentation](https://cloud.google.com/compute/docs/general-purpose-machines) to get more information. 
+All new VMs are based on a predefined image which is in turn based on `ubuntu-2204-lts` with python3 and docker installed. Custome images are not supported. You can change a machine type by providing `--vm-type`. Visit [the official GCP documentation](https://cloud.google.com/compute/docs/general-purpose-machines) to get more information. 
 
 #### Adding a new test case
 
-A test case should be presented as a YAML file.
+A test case should be present as a YAML file.
 
 ```YAML
 ---
@@ -101,12 +101,12 @@ Example:
 }
 ```
 
-Make sure that all the required fields are presented in the secret. The CLI will get all the secrets and upload them to the VM.
+Make sure that all the required fields are present in the secret. The CLI will get all the secrets and upload them to the VM.
 
-Consider adding new secrets only for new connector sources. Rich out to the Search Productivity team if you need any help with Vault secrets. 
+Consider adding new secrets only for new connector sources. Reach out to the Search Productivity team if you need any help with Vault secrets. 
 
 ### Can I use gcloud?
 Yes, it's already installed on your machine and the cli does not limit your use. Visit [the official documentation page](https://cloud.google.com/sdk/docs) for more information.
 
 ### Hygiene of use
-Since the CLI creates virtual machines in GCP it's recommended to keep them running only when you need them. Let's some trees 🌳🌳🌳. 
+Since the CLI creates virtual machines in GCP it's recommended to keep them running only when you need them. Let's save some trees 🌳🌳🌳. 
--- a/scripts/testing/README.md
+++ b/scripts/testing/README.md
@@ -108,3 +108,4 @@ Yes, it's already installed on your machine and the cli does not limit your use.
 
 ### Hygiene of use
 Since the CLI creates virtual machines in GCP it's recommended to keep them running only when you need them. Let's save some trees 🌳🌳🌳. 
+In case of pipeline errors check if your VM is still running `gcloud compute instances list` and then delete it manually `gcloud compute instances delete <your-vm-name>`
--- a/scripts/testing/README.md
+++ b/scripts/testing/README.md
@@ -69,11 +69,11 @@ scenarios:
     tests:
       - name: Full sync job is performed without errors
         job_type: full
-        max_timeout: 60
+        timeout: 60
         match: { status: 'completed'}
       - name: Incremental sync job is performed without errors
         job_type: incremental
-        max_timeout: 20
+        timeout: 20
         match: { status: 'completed'}
 ```
 
--- a/scripts/testing/README.md
+++ b/scripts/testing/README.md
@@ -18,9 +18,6 @@ Run `make clean install` to generate executable files in `./bin` folder
 
 ### Usage
 
-***Note:*** The working directory for `./bin/test-connectors` is `scripts/testing` so when you need to point to a file you should use it as a root directory. 
-e.g. `./bin/test-connectors run-test my-testing-environment-name --es-version 8.12-SNAPSHOT --test-case scenarios/clients/spo_full_and_incremental_syncs.yml`
-
 Run `./bin/test-connectors --help` or `./bin/test-connectors {command name} --help` to get more information about the cli. 
 
 #### Running test with Elastic cloud deployment
--- a/scripts/testing/README.md
+++ b/scripts/testing/README.md
@@ -1,0 +1,112 @@
+## Automated testing
+
+### Installation
+
+Atomated testing CLI depends on `gcloud` and `vault` CLIs so you need to install them. 
+
+#### Google cloud CLI
+Follow the [offical installation](https://cloud.google.com/sdk/docs/install) guide. Once installed, run `gcloud auth login` and authorize using SSO. 
+
+#### Vault
+Download and install the Hashicorp Vault client version 1.14.1 or newer or via brew with `brew install vault`
+1. `brew install vault`.
+2. [Generate](https://github.com/elastic/infra/tree/master/docs/vault#github-auth) a github token to authorize in Vault.
+3. Run `vault login --method github` to authorize your Vault cli.
+
+#### Connector service
+Run `make clean install` to generate executable files in `./bin` folder
+
+### Usage
+
+***Note:*** The working directory for `./bin/test-connectors` is `scripts/testing` so when you need to point to a file you should use it as a root directory. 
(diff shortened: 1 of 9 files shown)


scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
+## Automated testing
+
+### Installation
+
+Atomated testing CLI depends on `gcloud` and `vault` CLIs so you need to install them.
```
artem-shelkovnikov, line 5:
```suggestion
Automated testing CLI depends on `gcloud` and `vault` CLIs so you need to install them. 
```

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+1. `brew install vault`.
+2. [Generate](https://github.com/elastic/infra/tree/master/docs/vault#github-auth) a github token to authorize in Vault.
+3. Run `vault login --method github` to authorize your Vault cli.
+
+#### Connector service
+Run `make clean install` to generate executable files in `./bin` folder
+
+### Usage
+
+***Note:*** The working directory for `./bin/test-connectors` is `scripts/testing` so when you need to point to a file you should use it as a root directory.
```
artem-shelkovnikov, line 21: It's a bit confusing, can we support absolute paths by chance?

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+1. Create a cloud deployment
+2. Download a credentials file (or create a new user)
+3. Run `./bin/test-connectors run-test my-testing-environment-name --es-host {host} --es-username {user name} --es-password {password} --test-case {path to the test case file}`
+
+#### Running test with local Elasticsearch
+If you want to run your tests with local Elasticsearch you need to specify `--es-version` option. Like
+
+`./bin/test-connectors run-test my-testing-environment-name --es-version 8.12-SNAPSHOT --test-case {path to the test case file}`
+
+In this case, the cli will deploy an Elasticsearch instance in the same VM where the connector service will be running.
```
artem-shelkovnikov, line 37: VM here is VM in GCS?

artem-shelkovnikov: Is it possible to use local machine as a target?

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+#### Running test with a specific Connector service version
+
+You can use any git reference such as commit sha, a tag, or a branch name. The cli will pull the defined git reference and run `make clean install`. 
+
+Example: `./bin/test-connectors run-test my-testing-environment-name --es-version 8.12-SNAPSHOT --connectors-ref 8.12 --test-case {path to the test case file}`
+
+#### Keeping your VM running when the tests passed
+Sometimes it's useful to get access to the logs or make some changes in the code and run tests again. The CLI will print a list of useful commands you can use to access the VM resources like: 
+
+`Access logs: gcloud compute ssh {my-testing-environment-name} --zone {VM zone} --command "tail -f ~/service.log"`
```
artem-shelkovnikov, line 48: Really nice thing!

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+To automatically delete the VM you need to use `--delete` option. 
+
+`./bin/test-connectors run-test my-testing-environment-name --es-version 8.12-SNAPSHOT --connectors-ref 8.12 --test-case {path to the test case file} --delete`
+
+#### Using different machine type
+All new VMs are based on a predefined image which is in turn based on `ubuntu-2204-lts` with python3 and docker installed. Cusome images are not supported. You can change a machine type by providing `--vm-type`. Visit [the official GCP documentation](https://cloud.google.com/compute/docs/general-purpose-machines) to get more information.
```
artem-shelkovnikov, line 55:
```suggestion
All new VMs are based on a predefined image which is in turn based on `ubuntu-2204-lts` with python3 and docker installed. Custom images are not supported. You can change a machine type by providing `--vm-type`. Visit [the official GCP documentation](https://cloud.google.com/compute/docs/general-purpose-machines) to get more information. 
```

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+scenarios:
+  - index_name: search-demo-index-001
+    service_type: sharepoint_online
+    index_language: en
+    connector_configuration: scenarios/clients/spo_sean_site.json
+    native: false
+    tests:
+      - name: Full sync job is performed without errors
+        job_type: full
+        max_timeout: 60
```
artem-shelkovnikov, line 72: Should `max_timeout` be just `timeout`?

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+  "fetch_unique_list_item_permissions": true,
+  "fetch_unique_list_permissions": true,
+  "site_collections": "CustomWebSiteCollection",
+  "use_document_level_security": false
+}
+```
+
+Make sure that all the required fields are presented in the secret. The CLI will get all the secrets and upload them to the VM.
+
+Consider adding new secrets only for new connector sources. Rich out to the Search Productivity team if you need any help with Vault secrets.
```
artem-shelkovnikov, line 106:
```suggestion
Consider adding new secrets only for new connector sources. Reach out to the Search Productivity team if you need any help with Vault secrets. 
```

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+  "secret_value": "vault:path-to-spo-secret:SecretValue",
+  "enumerate_all_sites": false,
+  "fetch_unique_list_item_permissions": true,
+  "fetch_unique_list_permissions": true,
+  "site_collections": "CustomWebSiteCollection",
+  "use_document_level_security": false
+}
+```
+
+Make sure that all the required fields are presented in the secret. The CLI will get all the secrets and upload them to the VM.
```
artem-shelkovnikov, line 104:
```suggestion
Make sure that all the required fields are present in the secret. The CLI will get all the secrets and upload them to the VM.
```

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+Consider adding new secrets only for new connector sources. Rich out to the Search Productivity team if you need any help with Vault secrets. 
+
+### Can I use gcloud?
+Yes, it's already installed on your machine and the cli does not limit your use. Visit [the official documentation page](https://cloud.google.com/sdk/docs) for more information.
+
+### Hygiene of use
+Since the CLI creates virtual machines in GCP it's recommended to keep them running only when you need them. Let's some trees 🌳🌳🌳.
```
artem-shelkovnikov, line 112:
```suggestion
Since the CLI creates virtual machines in GCP it's recommended to keep them running only when you need them. Let's save some trees 🌳🌳🌳. 
```

scripts/testing/README.md
```diff
@@ -0,0 +1,112 @@
...
+You can use any git reference such as commit sha, a tag, or a branch name. The cli will pull the defined git reference and run `make clean install`. 
+
+Example: `./bin/test-connectors run-test my-testing-environment-name --es-version 8.12-SNAPSHOT --connectors-ref 8.12 --test-case {path to the test case file}`
+
+#### Keeping your VM running when the tests passed
+Sometimes it's useful to get access to the logs or make some changes in the code and run tests again. The CLI will print a list of useful commands you can use to access the VM resources like: 
+
+`Access logs: gcloud compute ssh {my-testing-environment-name} --zone {VM zone} --command "tail -f ~/service.log"`
+
+To automatically delete the VM you need to use `--delete` option.
```
artem-shelkovnikov, line 50: Curious - do you think it should delete by default instead, so that it's harder to miss that part that you need to delete stuff manually?

scripts/testing/cli.py
```diff
@@ -0,0 +1,603 @@
...
+    """
+    Reads the connector configuration file and replaces all the values that start with `vault:`
+    with the values from Vault
+    """
+    configuration = {}
+    with open(file_path, "r") as f:
+        configuration = json.loads(f.read())
+        for key, item in configuration.items():
+            if type(item) is str and item.startswith("vault:"):
+                configuration[key] = read_from_vault(item)
```
artem-shelkovnikov, line 282: Would be nice to have error handling here to include cases like file is not there, no access to file, no valid json is there, there are no `vault:` keys.

Also which keys can be there? Why that `vault:` prefix check?

scripts/testing/cli.py
```diff
@@ -0,0 +1,603 @@
...
+                else:
+                    click.echo(click.style(f"Test {test['name']} passed", fg="green"))
+                    break
+
+                if timeout >= test["max_timeout"]:
+                    click.echo(click.style(f"Test {test['name']} failed", fg="red"))
+                    break
+
+
+# TODO Read all the fields in one call
```
artem-shelkovnikov, line 571: Should we address this before merging, or is it okay for later?

How many keys do we expect?

scripts/testing/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: "2"
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/${NAMESPACE:-elasticsearch}/elasticsearch:${ES_VERSION}
+    environment:
+      - "discovery.type=single-node"
+      - "ES_JAVA_OPTS=-Xms1024m -Xmx1024m"
```
artem-shelkovnikov, line 8: Is 1GB enough? I'd make it 2 - it's our setup for regular ftests

scripts/testing/cli.py
artem-shelkovnikov: Ah I see - you open connector config and some values are `vault: <key>` and then you replace these with values from the vault. Amazing!

How are 403 messages shown for these? Just JSON of vault response?

artem-shelkovnikov (commented):
Great changeset!

Left a couple questions/comments :)

scripts/testing/README.md
vidok: > VM here is VM in GCS?

Yes. 

> Is it possible to use local machine as a target?

no, only GCP is supported for now.

scripts/testing/README.md
vidok: I assumed that when run a test and it fails you will want to SSH to the VM or at least look at the logs. All of this will be impossible if VM is gone.  One of the next steps is to have scheduled pipelines in Buildkite and one of them will be responsible for cleaning up the old VMs.

scripts/testing/cli.py
vidok: Yes, if there is an error it will be propagated from Vault.  Like: 

```
No value found at secret/unexisted_secret
```

scripts/testing/cli.py
vidok: I would address it in a separate PR if there will be significant performance issues.  I think each test can have 5-7 calls which is okay. Timewise it's less than 1% of the entire test.

scripts/testing/docker-compose.yml
vidok: In this case, we will need to change the default VM type. I assume some of the tests will not require 2GBs so we should probably make it configurable. Or another idea, we can group heavy tests in run them on a bigger VM. 

Anyway, it's too early to adjust it because we don't have many test cases.

scripts/testing/README.md
vidok: This confuses me too. I came up with this solution because Python executable scripts use `./bin` as they root. It's not the best user experience. I will change that.

artem-shelkovnikov (approved):
Tried it out, amazing tool!

Issue #1966, opened 2023-12-08
[DLS][Network Drive] Deny ACEs should take precedence over allowed ACEs 


praveen-elastic:
### To Reproduce

Assign permissions to a file in Windows Network Drive, with explicit `allow write` and `deny read` for a given user. Run the connector with DLS enabled. The document is not accessible to the user in Network Drive but it is visible in search application.

## Expected behavior

In the search application the document should not be visible since in Windows: Access-denied ACEs are placed before access-allowed ACEs. For reference: https://learn.microsoft.com/en-us/windows/win32/secauthz/order-of-aces-in-a-dacl

## Environment

Windows Network Drive

praveen-elastic:
Raised a PR that would fix this issue.

praveen-elastic:
Progress Summary: Resolved comments and merged the PR

PR #2004, opened 2023-12-21
[Network Drive] [DLS] Fix precedence for Deny permissions

### LINKED ISSUE #1963: [DLS][Network Drive] Document is visible if Group has GRANT access but user has an explicit DENY
## Bug Description

Steps to Reproduce: 

A group `A` has been provided read/write access to a file in Windows Network Drive but a given user `user1` of that group `A` has an explicit DENY permissions for a document.

## Expected Behaviour:
In that case the document is not accessible to `user1` in the source (network drive) since in Windows explicit deny ACEs are evaluated before any explicit allowed ACEs. Reference: https://learn.microsoft.com/en-us/windows/win32/secauthz/dacls-and-aces

## Actual Results:
The document is visible in the search applications to the user.

## Environment

Windows Network Drive

praveen-elastic:
## Closes #1963, #1966 

The Deny permissions takes precedence over Allow permissions in Windows Network Drive. So the following two cases have been addressed:

1. If a user has Explicit Deny permissions along with another Explicit allow permission, then the user is denied the access. #1966 
2. If a Group has Deny permission but a user of the group has Explicit allow permission, then the user is denied the access. #1963 

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/network_drive.py
+++ b/connectors/sources/network_drive.py
@@ -532,7 +532,8 @@ class NASDataSource(BaseDataSource):
         rid_groups = []
 
         for group_sid in groups_info or []:
-            rid_groups.append(_prefix_rid(group_sid.split("-")[-1]))
+            rid = group_sid.split("-")[-1]
+            rid_groups.append(_prefix_rid(rid))
 
         access_control = [rid_user, prefixed_username, *rid_groups]
 
@@ -617,6 +618,13 @@ class NASDataSource(BaseDataSource):
                 raise ConnectionError(msg) from exception
 
     async def get_entity_permission(self, file_path, file_type, groups_info):
+        """Processes permissions for a network drive, focusing on key terms:
+
+        - SID (Security Identifier): The unique identifier for a user or group, it undergoes revision.
+        - RID (Relative Identifier): The last part of a SID, uniquely identifying a user or group within a domain.
+        - ACE (Access Control Entry): An entry in `an Access Control List (ACL), defining permissions for a specific user or group. 0 is allowed ACE, 1 is denied ACE.
+        - Mask: A bit field representing allowed or denied permissions within an ACE. Example: Deny write, Allow read, etc.
+        """
         if not self._dls_enabled():
             return []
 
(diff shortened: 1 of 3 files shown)


tests/sources/test_network_drive.py
```diff
@@ -969,3 +985,26 @@ async def test_list_file_permissions_with_inaccessible_file():
...
             assert result is None
+
+
+@pytest.mark.asyncio
+@mock.patch.object(
+    NASDataSource,
+    "list_file_permission",
+    return_value=[
+        mock_permission(sid="S-2-21-211-411", ace=0),
+        mock_permission(sid="S-1-11-10", ace=1),
```
seanstory, line 996: If you add a user here that _does_ have access and does belong to the group, in addition to there being one user who was explicitly denied, what happens?

I expect that the document ends up with the group being granted allow permissions. Right?

connectors/sources/network_drive.py
```diff
@@ -639,13 +639,29 @@ async def get_entity_permission(self, file_path, file_type):
                 access=DirectoryAccessMask.READ_CONTROL,
             )
         for permission in list_permissions or []:
+            mask = permission["mask"].value
```
seanstory, line 642: Can you add some inline comments? As someone who hasn't spent a lot of time with network drives, I'm getting lost in the "mask", "sid", "rid", "ace", verbiage. Some comments to explain what's happening would be helpful.

tests/sources/test_network_drive.py
praveen-elastic: Hey, In that case the second user _won't_ be able to access the document since group permissions are also counted as explicit ACEs. Now Explicit Deny (from group membership) has precedence over Explicit allow (individual), so document is not accessible. 

I'll update the test case to add the second user to the group highlighting this scenario.

connectors/sources/network_drive.py
praveen-elastic: Apologies, you are right, without much idea of network drives, it is a bit complicated to review this PR. I'll update the file with relevant inline comments 👍

tests/sources/test_network_drive.py
seanstory: I thought this case was testing a group with ALLOW access, but a single user with explicit DENY access?

What I'd like to validate is the scenario where:

```
Users
-----
ID | First | Last
1 | John | Doe
2 | Jane | Fonda

Groups
-------
ID | Users
1 | [1, 2]
```

where Group 1 is given ALLOW access, User 1 is given explicit ALLOW access, and User 2 is given explicit DENY access. What is the resulting document `_allow_access_control` and the `permissions` for each of the two users?

praveen-elastic: In that scenario the resulting document's `_allow_access_control` is [`id-user1`], and the document will be accessible to User 1, and inaccessible to User 2. Should I add another test case it ?

seanstory: Please do - because reading the code, I'm worried the document will end up with `_allow_access_control` of `[group1]`, which would end up giving both users access, in error

seanstory: Thanks - I'm glad to see I was wrong, and I'm glad we'll have this test to protect against regressions.

seanstory (approved):

PR #2009, opened 2023-12-22
Make fixtures async

artem-shelkovnikov:
## Fixes failing functional tests

Current builds are failing because PostgreSQL fixture is using event loop to run async postgresql client. With recent change I've made `tests/sources/fixture.py` class async (to reuse code) and it caused this one fixture to fail, because it's running `run_until_complete` on current event loop that is already doing this.

So far the situation is:

1. Postgresql fixture uses async code
2. MongoDB/MySQL connectors use async code, their fixtures do not use async code (should be fixed)
3. PostgreSQL, MSSQL and OracleDB connectors and their fixtures do not use async code (there are manual wrappers), this should be fixed

So for now this async change just makes it work for PostgreSQL's fixture, but it also opens up way for unification of fixtures with connectors for MongoDB/MySQL (they can use same libraries/methods to connect to remote DB and manage it)

This code reflects the modification:

```python
    if hasattr(module, args.action):
        func = getattr(module, args.action)
        if args.action in ("setup", "load", "teardown", "remove"):
            return await func()
        return func()
```

So methods `setup`, `load`, `teardown` and `remove` are async, all others are sync.

If the method is not defined in the fixture, it's not gonna be called at all

### DIFF
--- a/tests/sources/fixtures/mongodb/fixture.py
+++ b/tests/sources/fixtures/mongodb/fixture.py
@@ -18,11 +18,7 @@ fake = Faker()
 client = MongoClient("mongodb://admin:justtesting@127.0.0.1:27021")
 
 
-def setup():
-    pass
-
-
-def load():
+async def load():
     def _random_record():
         return {
             "id": bson.ObjectId(),
@@ -45,7 +41,7 @@ def load():
     collection.insert_many(data)
 
 
-def remove():
+async def remove():
     db = client.sample_database
     collection = db.sample_collection
 
@@ -54,7 +50,3 @@ def remove():
 
     query = {"_id": {"$in": doc_ids}}
     collection.delete_many(query)
-
-
-def teardown():
-    pass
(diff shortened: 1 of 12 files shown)


tests/sources/fixtures/mongodb/fixture.py
```diff
@@ -54,7 +50,3 @@ def remove():
 
     query = {"_id": {"$in": doc_ids}}
     collection.delete_many(query)
-
-
-def teardown():
-    pass
```
navarone-feekery, line 60: Is `teardown` ever called? Just confirming that it's safe to remove entirely.

artem-shelkovnikov: `fixture.py` checks if the method is there before calling: 

```
    if hasattr(module, args.action): # <<<<< this
        func = getattr(module, args.action)
        if args.action in ("setup", "load", "teardown", "remove"):
            return await func()
        return func()
```

A lot of our fixtures don't have any of these methods implemented actually (like for SPO, Box, Github, Salesforce) :)

navarone-feekery (approved):
LGTM! 🚀

PR #2016, opened 2023-12-22
Replace more places accessing Elasticsearch client directly with proxied calls

artem-shelkovnikov:

This PR changes more places to use rich Elasticsearch clients instead of direct calls to python Elasticsearch client.

Example methods that were introduced or changed with this PR:
- `ESManagementClient.ensure_exists` method got "expand_wildcards" argument that's default to `open` - some places use "all", but most of the places use "open". "open" is also more conservative, thus the choice
- `ESManagementClient.create_content_index` - new method to create content index. Normally kibana does it, but in some cases (mostly CI or CLI) we wanna create the index ourselves. _This method is just extracted from the original place that was calling non-rich Elasticsearch client with no changes_
- `ESManagementClient.ensure_content_index_mappings` - new method to ensure that content index has our search mappings. _This method is just extracted from the original place that was calling non-rich Elasticsearch client with no changes_
- `ESManagementClient.ensure_ingest_pipeline_exists` - new method to ensure that ingest pipeline with provided id exists, if not it's created with provided arguments. Used by `kibana.py`. _This method is just extracted from the original place that was calling non-rich Elasticsearch client with no changes_
- `ESManagementClient.delete_indices` - new method to delete indices. Used by `kibana.py`. _This method is just extracted from the original place that was calling non-rich Elasticsearch client with no changes_
- `ESManagementClient.upsert` - new method to upsert a record into target index. Used by `kibana.py`. _This method is just extracted from the original place that was calling non-rich Elasticsearch client with no changes_
- `ESManagementClient.yield_existing_documents_metadata` - formely `Extractor._get_existing_ids`

There is still one change I haven't moved - it's call to `bulk`, I will address it in later PR(s).

All other changes faciliate the changes outlined above.

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli/connector.py
+++ b/connectors/cli/connector.py
@@ -2,7 +2,7 @@ import asyncio
 from collections import OrderedDict
 
 from connectors.es.client import ESManagementClient
-from connectors.es.settings import DEFAULT_LANGUAGE, Mappings, Settings
+from connectors.es.settings import DEFAULT_LANGUAGE
 from connectors.protocol import (
     CONCRETE_CONNECTORS_INDEX,
     CONCRETE_JOBS_INDEX,
@@ -30,7 +30,8 @@ class Connector:
         # TODO move this on top
         try:
             await self.es_management_client.ensure_exists(
-                indices=[CONCRETE_CONNECTORS_INDEX, CONCRETE_JOBS_INDEX]
+                indices=[CONCRETE_CONNECTORS_INDEX, CONCRETE_JOBS_INDEX],
+                expand_wildcards="all",
             )
 
             return [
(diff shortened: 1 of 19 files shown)


connectors/cli/connector.py
```diff
@@ -30,7 +30,8 @@ async def list_connectors(self):
         # TODO move this on top
         try:
             await self.es_management_client.ensure_exists(
-                indices=[CONCRETE_CONNECTORS_INDEX, CONCRETE_JOBS_INDEX]
+                indices=[CONCRETE_CONNECTORS_INDEX, CONCRETE_JOBS_INDEX],
+                expand_wildcards="all",
```
vidok, line 34: I think don't need `expand_wildcards` anymore because connectors work only with one index. 

I'm getting rid of the wildcards in this [PR](https://github.com/elastic/connectors/pull/1959/files)

connectors/cli/connector.py
```diff
@@ -87,25 +88,15 @@ async def __create(
...
+        await self.es_management_client.create_content_index(index_name, language)
 
     async def __create_connector(
         self, index_name, service_type, configuration, is_native, language, from_index
     ):
         try:
             await self.es_management_client.ensure_exists(
-                indices=[CONCRETE_CONNECTORS_INDEX, CONCRETE_JOBS_INDEX]
+                indices=[CONCRETE_CONNECTORS_INDEX, CONCRETE_JOBS_INDEX],
+                expand_wildcards="all",
```
vidok, line 99: The same. I don't think we need this.

connectors/cli/connector.py
```diff
@@ -87,25 +88,15 @@ async def __create(
             await self.es_management_client.close()
 
     async def __create_search_index(self, index_name, language):
```
vidok, line 90: Do we need this method? It's just one line of code :).

vidok:
Great work! I left a couple of comments. 

Also, I don't know if it's a Pythonic way, but does it make sense to move ESManagementClient to a new file?

connectors/cli/connector.py
artem-shelkovnikov: But this one is creating jobs and connectors indices that are hidden, I believe?

vidok: I think `expand_wildcards` is needed only when you want to update multiple indices that match the pattern.  

Here, we operate only with concrete indices.

vidok (approved):
Great work!

PR #2018, opened 2023-12-28
[MongoDB] Validate & Ensure consistency of enabling SSL 

parthpuri-elastic:
## Related to https://github.com/elastic/connectors/issues/1973

Add validation to check consistency of SSL and raise `ConfigurableFieldValueError` if conflicting values found

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -284,7 +284,8 @@ class MongoDataSource(BaseDataSource):
             msg = f"Database ({configured_database_name}) does not exist. Existing databases: {', '.join(existing_database_names)}"
             raise ConfigurableFieldValueError(msg)
 
-        database = client[configured_database_name]
+        if client is not None:
+            database = client[configured_database_name]
 
         existing_collection_names = await database.list_collection_names()
         self._logger.debug(
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -4,6 +4,8 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 import os
+import urllib.parse
+from contextlib import contextmanager
 from copy import deepcopy
 from datetime import datetime
 from tempfile import NamedTemporaryFile
@@ -12,7 +14,6 @@ import fastjsonschema
 from bson import DBRef, Decimal128, ObjectId
 from fastjsonschema import JsonSchemaValueException
 from motor.motor_asyncio import AsyncIOMotorClient
-from pymongo.errors import InvalidURI
 
 from connectors.filtering.validation import (
     AdvancedRulesValidator,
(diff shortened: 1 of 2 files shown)


connectors/sources/mongo.py
```diff
@@ -191,10 +190,24 @@ def get_default_configuration(cls):
...
+        try:
+            self.client = AsyncIOMotorClient(self.host, **self.client_params)
+            self.db = self.client[self.configuration["database"]]
+            self.collection = self.db[self.configuration["collection"]]
+        except InvalidURI as exception:
+            msg = f"Found Invalid URL, please check configuration. Error: {exception}"
+            raise InvalidURI(msg) from exception
+        except Exception as exception:
+            msg = f"Somthing went wrong while creating a client. Error: {exception}"
+            raise Exception(msg) from exception
```
artem-shelkovnikov, line 203: I advice, if possible, to make client a context - so it could be used with something like:

```
with self.get_client() as client:
    # call MongoDB
```

Otherwise the problems are:

1. What if you `_get_client` twice? First one will be overwritten and never disposed of
2. Hard to track who created client, cause it could have been created in different places.
3. You have to check in all places `if self.client is None`

connectors/sources/mongo.py
```diff
@@ -96,38 +97,36 @@ class MongoDataSource(BaseDataSource):
...
+                self.client_params["tlsCAFile"] = self.certfile
+            self.client_params["tlsInsecure"] = tls_insecure
         else:
-            client_params["tls"] = False
-
-        self.client = AsyncIOMotorClient(host, **client_params)
-
-        self.db = self.client[self.configuration["database"]]
-        self.collection = self.db[self.configuration["collection"]]
+            self.client_params["tls"] = False
```
artem-shelkovnikov, line 129: This code can go to `_get_client`

artem-shelkovnikov:
Additionally, this PR's title is non-informative, it should explain what the change does to the connector behaviour wise - can you fix it too?

artem-shelkovnikov:
Additionally, how does it address the original issue?

connectors/sources/mongo.py
parthpuri-elastic: As we have moved this code to `get_client`. I have added a code for removing the temp file in the `finally` block. So, every time it deletes the file when context is over so i guess its fix this [issue](https://github.com/elastic/connectors/issues/1948) as well

parthpuri-elastic:
> Additionally, how does it address the original issue?

I have added validation and raised `ConfigurableFieldValueError` if conflicting values were found. The client throws an error, which was initiated in __init__ (before validation). So, now we have created client separately and add validation.

artem-shelkovnikov:
@parthpuri-elastic I've opened a discussion with the team about this PR, as it looks like overkill (having it possible to configure SSL via connection string and rich configurable field).

We'll get back to this PR after the discussion

seanstory (approved):
> I've opened a discussion with the team about this PR [...]
> We'll get back to this PR after the discussion

After team discussion, we're aligned on taking this approach. Thanks for your work and your patience, apologies for the delay.

parthpuri-elastic:
@seanstory, I have resolved the merge conflicts and added validation for TLS as well. So can you please re-review it?

seanstory (approved):

PR #2023, opened 2024-01-03
Make mongodb work without admin privelege

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/1922

This PR makes it possible to use MongoDB connector with user that has _only_ privilege to read from a collection.

Before this PR, privileges to list databases/collections was also needed.

#### Pre-Review Checklist

Validation logic got more complex and I feel that it can be simplified.

What we do now in steps:

1. Check if collection is accessible by current user
2. If it is, validation is successful
3. If not, try to read list of databases and collections and show the user available dbs/collections in UI
4. If any of operations in 3. fail, we display a message that user has no access to the collection/database and they manually need to ensure that these collection, database exist and user has access to them

Is it getting too complex?

We can safely drop steps 3/4 and always show validation error without availables dbs/collections, code gets much much simpler, but UI gets less user friendly. Would it be worth to delete the code in 3 and 4 to make connector simpler?

### DIFF
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -270,8 +270,9 @@ class MongoDataSource(BaseDataSource):
             )
             return
         except OperationFailure:
-            self._logger.warning(f"Unable to access '{configured_database_name}.{configured_collection_name}' as user '{user}'")
-
+            self._logger.warning(
+                f"Unable to access '{configured_database_name}.{configured_collection_name}' as user '{user}'"
+            )
 
         # If it's not accessible, try to make a good user-friendly error message
         try:
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -265,10 +265,11 @@ class MongoDataSource(BaseDataSource):
         collection_valid = True
 
         try:
+            # This works on both standalone and Managed mongo in the same way
             await client[configured_database_name].validate_collection(
                 configured_collection_name
             )
-        except OperationFailure:  # If the collection doesn't exist
+        except OperationFailure:
             collection_valid = False
 
         if collection_valid:
@@ -276,6 +277,9 @@ class MongoDataSource(BaseDataSource):
 
         # If it's not accessible, try to make a good user-friendly error message
         try:
+            # We will try to access some databases/collections to give a friendly message
+            # That will suggest the name of existing collection - but only if the user
+            # that we use to log in into MongoDB has access to it
             existing_database_names = await client.list_database_names()
 
             self._logger.debug(f"Existing databases: {existing_database_names}")
@@ -295,7 +299,9 @@ class MongoDataSource(BaseDataSource):
                 msg = f"Collection '{configured_collection_name}' does not exist within database '{configured_database_name}'. Existing collections: {', '.join(existing_collection_names)}"
                 raise ConfigurableFieldValueError(msg)
         except OperationFailure as e:
-            user = self.configuration["user"]
             # This happens if the user has no access to operations to list collection/database names
+            # Managed MongoDB never gets here, but if we're running against a standalone mongo
+            # Then this code can trigger
+            user = self.configuration["user"]
             msg = f"Database '{configured_database_name}' or collection '{configured_collection_name}' is not accessible by user '{user}'. Verify that these database and collection exist, and specified user has access to it"
             raise ConfigurableFieldValueError(msg) from e
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -265,7 +265,9 @@ class MongoDataSource(BaseDataSource):
         collection_valid = True
 
         try:
-            await client[configured_database_name].validate_collection(configured_collection_name)
+            await client[configured_database_name].validate_collection(
+                configured_collection_name
+            )
         except OperationFailure:  # If the collection doesn't exist
             collection_valid = False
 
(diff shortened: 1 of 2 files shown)


connectors/sources/mongo.py
```diff
@@ -256,24 +257,51 @@ async def validate_config(self):
...
+        # First check if collection is accessible
+        collection_valid = True
+
+        try:
+            # This works on both standalone and Managed mongo in the same way
+            await client[configured_database_name].validate_collection(
+                configured_collection_name
+            )
+        except OperationFailure:
+            collection_valid = False
```
seanstory, line 273: Should we log anything here to help the customer troubleshoot what went wrong?

connectors/sources/mongo.py
```diff
@@ -256,24 +257,51 @@ async def validate_config(self):
...
+
+        if collection_valid:
+            return
```
seanstory, line 276: why not return as a last step in the above `try` block?

connectors/sources/mongo.py
```diff
@@ -256,24 +257,51 @@ async def validate_config(self):
...
-            raise ConfigurableFieldValueError(msg)
+            if configured_collection_name not in existing_collection_names:
+                msg = f"Collection '{configured_collection_name}' does not exist within database '{configured_database_name}'. Existing collections: {', '.join(existing_collection_names)}"
+                raise ConfigurableFieldValueError(msg)
+        except OperationFailure as e:
+            # This happens if the user has no access to operations to list collection/database names
+            # Managed MongoDB never gets here, but if we're running against a standalone mongo
+            # Then this code can trigger
+            user = self.configuration["user"]
+            msg = f"Database '{configured_database_name}' or collection '{configured_collection_name}' is not accessible by user '{user}'. Verify that these database and collection exist, and specified user has access to it"
```
seanstory, line 306: Shouldn't this message be that the user wasn't allowed to list databases? IDK a lot about mongo, but I would assume that there's a scenario where a user could list contents of a known database, but couldn't list all databases or even list databases and get back only the database for which they have content access.

connectors/sources/mongo.py
artem-shelkovnikov: Omg yes, will do it

connectors/sources/mongo.py
artem-shelkovnikov: Good suggestion, will add a line

connectors/sources/mongo.py
artem-shelkovnikov: Yes, but it's not the important bit for this code.

We list databases/collections to give a suggestion in the validation error message:

> Can't find DB "toys", found dbs: ["pets", "dolls", "bicycles"]

Or:

> Can't find collection "road bikes" in database "bicycles". Existing collections: ["road_bikes", "mtb", "kids_bikes"]

If we fail to list collections/dbs, it's not a critical problem - we don't really need these permissions for anything except suggestions in validation message. If we fail to list collections/databases, we just cannot suggest, thus we just say that user has no access to the specified db/collection.

connectors/sources/mongo.py
artem-shelkovnikov: Addressed in https://github.com/elastic/connectors/pull/2023/commits/4c51560fa6c1d8de997cea0b9d580207cc3cd806

connectors/sources/mongo.py
artem-shelkovnikov: Addressed in https://github.com/elastic/connectors/pull/2023/commits/4c51560fa6c1d8de997cea0b9d580207cc3cd806

connectors/sources/mongo.py
seanstory: I don't think I understand.
I'm concerned that 
```
existing_database_names = await client.list_database_names()
```
could fail, because the user might not be allowed "list" permissions on "database names". But the connector doesn't actually need that permission - what we need is "list" permissions on "X database, Y collection contents"

So the error message of `Database 'X' or Collection 'Y' is not accessible by user` may be incorrect - those might be perfectly accessible, it was just the `list_database_names()` operation that failed.

seanstory (approved):
Left a follow-up comment, but assuming it's just a matter of me misunderstanding, otherwise LGTM. :)

connectors/sources/mongo.py
wangch079: Agreed. A customer can specify a valid database `foo` with an inaccessible collection `bar`, and we are expected to display something like
>You are not authorized to access collection `bar` in database `foo`

I'd suggest we avoid doing operations like `list_database_names()` and `list_collection_names()`, as this is not required to sync data. We should be able to sync as long as the user has access to the configured collection of the configured database.

wangch079: In other words, I'd suggest removing 3 and 4, and simply checking with something like

```python
try:
    database = client[configured_database_name]
    collection = database[configured_collection_name]
except OperationFailure as e:
    msg = f"User '{user}' does not have access to collection '{configured_collection_name}' in database '{configured_database_name}'. Verify that the database and the collection exist, and specified user has access to it"
    raise ConfigurableFieldValueError(msg) from e
```

artem-shelkovnikov: The code above `existing_database_names = await client.list_database_names()` already determined, that either DB or collection is not available, so at this point we already know - either DB or collection is not accessible. And then this code kicks in, trying to either make a user-friendly message that says which collections/dbs are accessible OR if user has no permission to list DB names to just give a generic message.

That's why I also feel like this whole section in try/except can be removed - it does a lot, confuses writer/maintainer for the price of the message that's slightly user-friendlier.

artem-shelkovnikov: > I'd suggest we avoid doing operations like list_database_names() and list_collection_names(), as this is not required to sync data. 

While I partially agree, I remember that one of the feedback for MongoDB connector was that it's really hard to find problems when you specify wrong db/collection, so user experience improved a lot with this validation.

seanstory: > The code above existing_database_names = await client.list_database_names() already determined, that either DB or collection is not available, so at this point we already know - either DB or collection is not accessible. And then this code kicks in, trying to either make a user-friendly message that says which collections/dbs are accessible OR if user has no permission to list DB names to just give a generic message.

Totally misunderstood/misread. You even have that in a comment:
```
# If it's not accessible, try to make a good user-friendly error message
```

I'm going to resolve this comment, we can pretend it never happened. :P

wangch079: >The code above existing_database_names = await client.list_database_names() already determined, that either DB or collection is not available

👍 

>I remember that one of the feedback for MongoDB connector was that it's really hard to find problems when you specify wrong db/collection, so user experience improved a lot with this validation.

I agree. Let's keep it. Even when the user does not have privilege to list database/collections, we will still display a friendly message:
>Database '{configured_database_name}' or collection '{configured_collection_name}' is not accessible by user '{user}'

wangch079 (approved):

PR #2029, opened 2024-01-05
Make clean_index method proceed on conflicts

artem-shelkovnikov:
Minor improvement for `clean_index` - added `conflicts=proceed` to not fail the cleaning of the index if conflicts happened during clearing.

Normally conflicts can happen if document was updated while the delete_by_query is executed (this includes ingest pipelines updating a document after it was ingested)

### DIFF
--- a/connectors/es/index.py
+++ b/connectors/es/index.py
@@ -72,6 +72,7 @@ class ESIndex(ESClient):
             index=self.index_name,
             body={"query": {"match_all": {}}},
             ignore_unavailable=True,
+            conflicts="proceed"
         )
 
     async def update(self, doc_id, doc, if_seq_no=None, if_primary_term=None):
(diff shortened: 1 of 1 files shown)


seanstory (approved):

connectors/es/index.py
```diff
@@ -72,6 +72,7 @@ async def clean_index(self):
             index=self.index_name,
             body={"query": {"match_all": {}}},
             ignore_unavailable=True,
+            conflicts="proceed",
```
seanstory, line 75: Bonus points if you add a unit test

artem-shelkovnikov: Do you have a pattern of testing in mind? I can't think of reliable way of testing it, except for checking that function was called with an arg

seanstory: ah shoot that's right, we don't run tests with a real ES. :/ 
OK yeah, probably not super valuable to write a test that validates mocks.

PR #2042, opened 2024-01-09
Raise an error if our ES client fails to execute a _search

seanstory:
## related to https://github.com/elastic/enterprise-search-team/issues/6075

We have a theory that we've been hitting some bugs when ES is unhealthy and returns non-200 responses for `_search` requests. This PR changes our client behavior to raise the ApiError if we hit one, instead of implicitly returning an empty result set for the executed search.

#### Pre-Review Checklist

## Related Pull Requests

- see also: https://github.com/elastic/connectors/pull/2038

## Release Note

Fixes a bug where Elasticsearch health issues might be passed over. Previously, if Elasticsearch returned a 500 error when connectors executed a `_search`, this error was not propagated correctly.

### DIFF
--- a/connectors/es/index.py
+++ b/connectors/es/index.py
@@ -123,9 +123,11 @@ class ESIndex(ESClient):
                     seq_no_primary_term=True,
                 )
             except ApiError as e:
-                logger.critical(f"The server returned {e.status_code}")
-                logger.critical(e.body, exc_info=True)
-                return
+                logger.error(
+                    f"Elasticsearch returned {e.status_code} for 'GET {self.index_name}/_search' with body:"
+                )
+                logger.error(e.body, exc_info=True)
+                raise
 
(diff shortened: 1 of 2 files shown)


sphilipse (approved):

wangch079 (approved):

artem-shelkovnikov:
IMO it's nice to test it around locally - theoretically we should be safe, but some things can really break. Probably running functional test suite a couple times should be sufficient (to do so I normally add a blank line into the Dockerfile for functional tests)

seanstory:
> IMO it's nice to test it around locally

Yes! I wanted to get the PR up at my EOB yesterday, but hadn't tested yet so left it unchecked. I'd planned to just run some syncs, but ftests is a good idea. I'll run the nightly on this branch! https://buildkite.com/elastic/connectors-nightly/builds/1717

PR #2044, opened 2024-01-10
Code of Redis connector

akanshi-elastic:
## Part of [https://github.com/elastic/connectors/issues/1941](https://github.com/elastic/connectors/issues/1941)
PR includes:
1. Redis ping+indexing
2. Redis sync rules
#### Pre-Review Checklist


connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+        self._logger = logger_
+
+    @asynccontextmanager
+    async def client(self):
+        pool = redis.ConnectionPool.from_url(
+            f"redis://{self.username}:{self.password}@{self.host}:{self.port}",
+            decode_responses=True,
+            max_connections=MAX_POOL_SIZE,
+        )
+        yield redis.Redis(connection_pool=pool)
```
wangch079, line 51: Should you do the cleanup in the `finally`?

In the official [example](https://redis.readthedocs.io/en/stable/examples/asyncio_examples.html), if you create client with `from_pool` class method, you will only need to call `await client.close()` to close it.

But if you use `redis.Redis(connection_pool=pool)`, you need to explicitly close the client and the pool.

connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+                database = await redis_client.config_get("databases")
+                databases = list(range(int(database.get("databases", 1))))
+        except Exception as exception:
+            self._logger.warning(
+                f"Something went wrong while fetching database list. Error: {exception}",
+                exc_info=True,
+            )
+        return databases
+
+    async def _get_paginated_keys(self, redis_client, pattern, _type=None):
```
wangch079, line 80: If this is supposed to be called externally, don't prefix it with `_`.

connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+
+        Yields:
+            list: list of keys in a database
+        """
+        cursor, keys = await redis_client.scan(
+            match=pattern, count=PAGE_SIZE, _type=_type
+        )
+        yield keys
+
+        while cursor != 0:
```
wangch079, line 96: You can consider using [scan_iter](https://redis.readthedocs.io/en/stable/commands.html#redis.commands.core.CoreCommands.scan_iter)

connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+                "type": "str",
+            },
+            "password": {
+                "label": "Password",
+                "order": 4,
+                "required": False,
+                "sensitive": True,
+                "type": "str",
+            },
+            "database": {
```
wangch079, line 171: Is this a product ask to have this field?

connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+            dictionary: Document of database content
+
+        Yields:
+            dictionary: Document from Redis.
+        """
+        async with self.redis_client.client() as redis_client:
+            db_list = await self.redis_client.get_databases()
+            for db in db_list:
+                async for document in self.get_db_records(
+                    db=db, redis_client=redis_client
```
wangch079, line 281: It feels strange that you need to pass `redis_client` to almost every method of `self.redis_client`, while `redis_client` is from `self.redis_client`.

I think this kind of details should be hidden inside `self.redis_client`.

connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+        """Get documents from Redis
+
+        Returns:
+            dictionary: Document of database content
+
+        Yields:
+            dictionary: Document from Redis.
+        """
+        async with self.redis_client.client() as redis_client:
+            db_list = await self.redis_client.get_databases()
```
wangch079, line 278: Why not make `get_databases` an async generator?

connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+                return await redis_client.lrange(key, 0, -1)
+            elif key_type == KeyType.SET.value:
+                return await redis_client.smembers(key)
+            elif key_type == KeyType.ZSET.value:
+                return await redis_client.zrange(key, 0, -1, withscores=True)
+            elif key_type == KeyType.JSON.value:
+                value = await redis_client.execute_command("JSON.GET", key)
+                return json.loads(value)
+            else:
+                value = await redis_client.get(key)
```
wangch079, line 128: Can we simply use `get` here? If so, we can use `mget` as well

connectors/sources/redis.py
```diff
@@ -0,0 +1,283 @@
...
+        except Exception:
+            self._logger.exception("Error while connecting to Redis.")
+            raise
+
+    async def get_key_metadata(self, redis_client, key, db):
+        key_type = await redis_client.type(key)
+        key_value = await self.redis_client.get_key_value(
+            redis_client=redis_client, key=key, key_type=key_type
+        )
+        key_size = await redis_client.memory_usage(key)
```
wangch079, line 248: This concerns me. This means for each key-value pair, we need 3 operations.

After a quick search, we can use `mget` to get values in batch, it seems there's no easy way to get the type and size in batch.

connectors/sources/redis.py
akanshi-elastic: `redis_client` is a context manager if we store it in `self.redis_client` then it will get close everytime.

connectors/sources/redis.py
akanshi-elastic: we can use `get` or `mget` only for `string` type of keys. And if we'll switch to `mget` then we need to iterate over all the keys to find the string type and store it in a list to pass. I guess in that case iteration will be increase.

connectors/sources/redis.py
akanshi-elastic: yes, There is no method present which returns type and size in single method or batch.

connectors/sources/redis.py
wangch079: Do you want to keep the pool open or the redis client as well?

connectors/sources/redis.py
akanshi-elastic: I guess `databases` misunderstood here. Basically we are passing the database number to which user want to connect as for redis server we have default 16 index available that can be changed from redis.conf file.

connectors/sources/redis.py
akanshi-elastic: Yes, I guess we need to open the pool/client until all documents get index into elastic search. Otherwise we need to create it every time which degrade the performance of connector. We also faced the performance issue in S3 because of creating multiple clients.

connectors/sources/redis.py
wangch079: Yes, I am concerned about this because I see the database is just a number. Is it easy for a customer to get this number? It's not as intuitive as `tables` in MySQL data source. Is this a feature expected from customers?

connectors/sources/redis.py
wangch079: I also see this is used inside `RedisClient`:

https://github.com/elastic/connectors/blob/4147a65dac50e0b6611d52b3f1e7f1e703b1994f/connectors/sources/redis.py#L75

if you only want to have one instance of the client and keep it open until the end of the `RedisClient`, consider assigning it to an instance variable, and close it in a `close` method.

connectors/sources/redis.py
akanshi-elastic: In redis database is referred with number only. In below image you can see the edit option where we can change the database number.  For better understanding we can change the name from `databsse--->index`
image (1)

wangch079: If the term in Redis is called `Database`, then we can leave it as `Database`.

connectors/sources/redis.py
```diff
@@ -0,0 +1,280 @@
...
+            yield client
+        finally:
+            await client.aclose()  # pyright: ignore
+
+    async def validate_database(self, db, redis_client):
+        try:
+            await redis_client.execute_command("SELECT", db)
+            return await redis_client.ping()
+        except Exception as exception:
+            self._logger.warning(f"Database {db} not found. Error: {exception}")
```
wangch079, line 62: Should this method explicitly return `Ture/False`?

akanshi-elastic: yes we need to mention it explicitly because on the basis of that we are validating valid and invalid databases.

connectors/sources/redis.py
akanshi-elastic: As of now we didn't added code for ssl support. But with ssl we need to pass the path of file in query while creating client as we don't have a way to pass ssl context and In that case we'll face the same issue as mongodb.
Please refer this issue for better understanding: https://github.com/elastic/connectors/issues/1948

wangch079: From the code snippet you posted, It seems the ssl path was passed when creating the pool. So ideally, we should have a lazy instantiated pool in `RedisClient` instance, and we should close it with a `close` method. From the pool, you can create a redis client via `redis.Redis.from_pool(pool)`.

>and in the call of the `validate`, connector was never called `close`.

I think it's the responsibility of the users (of `RedisClient`) to `close` it after use.

wangch079: Just curious, must a Redis client instance be instantiated from a pool? I don't think we have the use case of multiple connections.

wangch079: I think from the comment, Artem mentioned 
>call of `validate` on connector never called `close` after validation is ended.

But it doesn't mean the temporary files cannot be removed from `close` method.

connectors/sources/redis.py
```diff
@@ -0,0 +1,269 @@
...
+                        return value
+        except UnicodeDecodeError as exception:
+            self._logger.warning(
+                f"Something went wrong while fetching value for key {key}. Error: {exception}",
+                exc_info=True,
+            )
+            return ""
+
+    async def ping(self):
+        await self.client()
```
wangch079, line 134: Every call to `ping` will create a new client

akanshi-elastic: I guess ping calls once in a whole running connector and if that is the case then client will also be called once only.

wangch079 (commented):
Make `client()` a lazy loaded method, and only use `self.client().<function>()` within `RedisClient`.

Also you can consider rename the method to `_client()` and `redis_client` to `_redis_client` as they are not supposed to be accessed from outside.

tests/sources/test_redis.py
```diff
@@ -0,0 +1,199 @@
...
+    async with create_redis_source() as source:
+        source.client.ping = AsyncMock()
+        await source.ping()
+
+
+@pytest.mark.asyncio
+async def test_ping_negative():
+    async with create_redis_source() as source:
+        with pytest.raises(Exception):
+            await source.ping()
```
wangch079, line 100: This tries to make a real ping to `localhost`?

tests/sources/test_redis.py
```diff
@@ -0,0 +1,199 @@
...
+        async for database in source.client.get_databases():
+            assert database == 0
+
+
+@pytest.mark.asyncio
+async def test_get_databases_negative():
+    async with create_redis_source() as source:
+        source.client.database = ["*"]
+        async for database in source.client.get_databases():
+            assert database == []
```
wangch079, line 157: It seems this passed because it can't get any result from non-existing `localhost`.

tests/sources/test_redis.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+            with pytest.raises(Exception):
+                await source.ping()
+
+
+@pytest.mark.asyncio
+async def test_validate_config_when_database_type():
+    async with create_redis_source() as source:
+        source.client.database = ["1", "db123", "123"]
+        source.ping = AsyncMock(return_value=True)
```
wangch079, line 113: it won't call `source.ping` in `source.validate_config()`

tests/sources/test_redis.py
```diff
@@ -0,0 +1,210 @@
...
+        with pytest.raises(ConfigurableFieldValueError):
+            await source.validate_config()
+
+
+@pytest.mark.asyncio
+async def test_validate_config_when_database_is_invalid():
+    async with create_redis_source() as source:
+        source.client.database = ["123"]
```
wangch079, line 121: mock `source.client.validate_database`

connectors/sources/redis.py
```diff
@@ -0,0 +1,274 @@
...
+        Raises:
+            ConfigurableFieldValueError: Unavailable services error.
+        """
+        invalid_type = []
+        invalid_db = []
+        msg = ""
+        if self.client.database != ["*"]:
+            for db in self.client.database:
+                if not db.isdigit() or int(db) < 0:
+                    invalid_type.append(db)
```
wangch079, line 205:
```suggestion
                    invalid_type.append(db)
                    continue
```

tests/sources/test_redis.py
```diff
@@ -0,0 +1,210 @@
...
+                await source.ping()
+
+
+@pytest.mark.asyncio
+async def test_validate_config_when_database_type():
+    async with create_redis_source() as source:
+        source.client.database = ["1", "db123", "123"]
+        source.ping = AsyncMock(return_value=True)
+        with pytest.raises(ConfigurableFieldValueError):
+            await source.validate_config()
```
wangch079, line 115:
```suggestion
        mocked_client = Mock()
        mocked_client.database = ["1", "db123", "123"]
        mocked_client.validate_database = AsyncMock(return_value=True)
        with mock.patch("redis.from_url", return_value=mocked_client):
            with pytest.raises(ConfigurableFieldValueError):
                await source.validate_config()
```

tests/sources/test_redis.py
```diff
@@ -0,0 +1,215 @@
...
+            with pytest.raises(Exception):
+                await source.ping()
+
+
+@pytest.mark.asyncio
+async def test_validate_config_when_database_type():
+    async with create_redis_source() as source:
+        source.client.database = ["1", "db123", "123"]
+        mocked_client = Mock()
+        mocked_client.client.validate_database = AsyncMock(return_value=True)
```
wangch079, line 114: `mocked_client` is the mocked object returned from `redis.from_url`, and assigned to `self._redis_client` of `RedisClient`.

`validate_database` is a method from `RedisClient`.

When you want to unit test `RedisDataSource`, you can mock `RedisClient` as it's a dependency of `RedisDataSource`.

When you want to unit test `RedisClient`, you can mock `redis.from_url` as it's a dependency of `RedisClient`.

wangch079 (approved):
Conditional approve. Tests need to be improved.

PR #2045, opened 2024-01-10
Mark orphaned idle sync jobs as error

wangch079:
### What's the change

Instead of deleting orphaned jobs, this PR tries to mark those orphaned (i.e. sync jobs with deleted connectors) idle (sync jobs seeing no update for more than 1 min) sync jobs as `error`.

### Why the change

When a customer deletes a connector in Kibana, it will not delete all the associated sync jobs. Those orphaned sync jobs are kept for audit purporse. If the orphaned sync jobs are idle, we will mark them as `error`

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -1043,8 +1043,22 @@ class SyncJobIndex(ESIndex):
         async for job in self.get_all_docs(query=query, sort=sort):
             yield job
 
-    async def orphaned_jobs(self, connector_ids):
-        query = {"bool": {"must_not": {"terms": {"connector.id": connector_ids}}}}
+    async def orphaned_idle_jobs(self, connector_ids):
+        query = {
+            "bool": {
+                "must_not": {"terms": {"connector.id": connector_ids}},
+                "filter": [
+                    {
+                        "terms": {
+                            "status": [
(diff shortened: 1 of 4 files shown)


sphilipse (commented):
I wonder whether we should keep the function but just have it cancel running orphaned jobs, rather than deleting them 🤔

wangch079:
>I wonder whether we should keep the function but just have it cancel running orphaned jobs, rather than deleting them

Good catch. That should be done by `_process_idle_jobs`, which fetches idle jobs by connector IDs. An orphaned job won't be processed. 

The function `_process_idle_jobs` was designed to process connectors which the instance is supposed to process:

https://github.com/elastic/connectors/blob/225592b0823a895fc8cff7a7f4738232720993d5/connectors/services/job_cleanup.py#L89-L92

If this is a connector running on Cloud, it will process all the native connectors.
If this is a self-managed instance, it will only process the connector clients configured.

I will revisit this tomorrow, and target this fix to 8.12.1

wangch079:
Hi @elastic/ingestion-team , updated the PR to mark orphaned idle sync jobs as `error`, instead of deleting orphaned sync jobs.

artem-shelkovnikov (approved):

PR #2051, opened 2024-01-11
[CLI] Improve Connectors CLI's service type picker during prompted config; also some small fixes to the CLI docs

acrewdson:
**(ad hoc change)**

* improve the UI for selecting your service type during Connectors CLI prompted configuration
* some misc fixes to the CLI docs
 
**Notes**:
* when reviewing, please grab the branch locally and run `bin/connectors connector create` to see the new service type selection menu
* if we like this, we could also consider using it for the 'Index language' prompt


artem-shelkovnikov:
Change looks good, but not sure - how will it work if the list grows?

Also, I see it broke some tests - no point in approving it now cause approval will be discarded with your changes :)

connectors/connectors_cli.py
```diff
@@ -147,6 +149,16 @@ def validate_language(ctx, param, value):
     return value
 
 
+# override click's default 'choices' prompt with something a bit nicer
+def interactive_service_type_prompt():
+    options = list(_default_config()["sources"].keys())
+    print(f"{Fore.GREEN}?{Style.RESET_ALL} Service type:")  # noqa: T201
+    result = TerminalMenu(
+        options, menu_cursor_style=("fg_green",), clear_menu_on_exit=False
```
vidok, line 157: I would add these options: 

```
show_search_hint=True
search_key=None
```
To let customers know that they can type to search.

vidok:
Nice contribution 👍🏻 

Is there a way to limit the list of connector types? It makes total sense to have search capabilities however if we can't limit the list (aka pagination/scrolling) the problem of displaying a huge list will not go away.

acrewdson:
@artem-shelkovnikov thanks for the review 🙏 

> Change looks good, but not sure - how will it work if the list grows?

good question, I'll look into that

> Also, I see it broke some tests - no point in approving it now cause approval will be discarded with your changes :)

yep, I noticed that also after I requested reviews, so I un-requested reviews at that point; I'll work on fixing those tests 🤦 

@vidok 

> Is there a way to limit the list of connector types? It makes total sense to have search capabilities however if we can't limit the list (aka pagination/scrolling) the problem of displaying a huge list will not go away.

thanks also for the review, it's a good question, will get back to you (fits with Artem mentions above I think)

acrewdson:
@artem-shelkovnikov @vidok 

Re the concern about 

> the problem of displaying a huge list will not go away.

and 

> how will it work if the list grows?

I'm not sure it's a concern, see this recording demonstrating how the menu scrolls when you have a small window:

https://github.com/elastic/connectors/assets/10353074/1a1a1ebb-d19c-4b4a-bf8c-f59a5a89f105

What do you think?

> show_search_hint=True

I added it, seems nice 👍 

> search_key=None

Unfortunately it doesn't seem to work (though I realize it's documented and should work); I get an exception when scrolling the list if I have `search_key=None` 🤷 At least we have `/` to search the list.

Also, tests are passing again, let me know if you feel the way I achieved this is problematic. Thanks again for considering this PR, and no problem at all if you feel it's too hacky; I'd of course much prefer that we keep the CLI robust and testable, even if it means not having a great service type selection experience :-)

artem-shelkovnikov (approved):
Looks great, thanks for doing this!

vidok:
What I mean by scrolling is that the list of service types doesn't take the entire screen. 

Here is an example: 
I used [gh](https://cli.github.com) to record this video (when I select labels you can see how the menu scrolls):

https://github.com/elastic/connectors/assets/314956/27d22189-7cf2-4fff-810d-c501c0db6e43. 

> Unfortunately it doesn't seem to work (though I realize it's documented and should work); I get an exception when scrolling the list if I have search_key=None 🤷 At least we have / to search the list.

It's a pity. I also tested and it also crashed on my laptop when scrolling the list.  If there is no easy fix for that I'm fine moving forward with these changes.

PR #2055, opened 2024-01-11
[Salesforce] Add custom object support with different API Names

moxarth-elastic:
## Part of https://github.com/elastic/connectors/issues/1859

This problem is faced while the API name (which is used in SOQL query to fetch data) is different in the Salesforce edition. I also noticed that the `/sobjects` endpoint provides the custom objects with the suffix `__c`, `__e` and might be some others, so we filter out the custom objects having just `__c` as a suffix in object API name.

#### Pre-Review Checklist


navarone-feekery (approved):
LGTM

PR #2063, opened 2024-01-15
Implement consistent retry policy when calling Elasticsearch

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/6412

This PR makes our retry policy when calling Elasticsearch consistent between all requests to Elasticsearch.

It's possible to specify the following settings in a config file:

```yml
elasticsearch.max_retries: 5
elasticsearch.retry_interval: 10
```

The above setup states, that we'll retry every call (wrapped with retryable abstraction) 5 times, starting from 10 seconds on first retry. Internally, retry policy is Linear, so first retry will be 10 seconds, second retry will be 20 seconds, and so on.

For now absolutely every call to Elasticsearch will be retried like this (I wrapped all calls with the retrier).

Errors that are retried are:

- Api Error with response code 429, 500, 502, 503, 504
- Connection Timeout

To do the work I've introduced a `TransientElasticsearchRetrier` class that handles the logic.

Retry method signature is:

```python
async def execute_with_retry(self, func):
```

To retry something, you need to do the following:

```python
license_response = await self._retrier.execute_with_retry(
            self.client.license.get
        )
```

In case the ES call requires arguments, `partial` should be used:

```python
return await self._retrier.execute_with_retry(
            partial(
                self.client.indices.create,
                index=search_index_name,
                mappings=mappings,
                settings=settings,
            )
        )
```

Separate class was created because:

1. I need to make this retry configurable, so I need to pass configuration values into it. It's not possible to do unless we expose config globally, but that's not we do, thus the class.
2. I'm able to cancel the retries this way - currently retries in `@retryable` wrapper are not cancellable, so stopping a running job that's inside retry loop will generate an error

#### Pre-Review Checklist

## Release Note

Improve connection resiliency when accessing elasticsearch - retry all failed calls that fail because of unhealthy Elasticsearch before terminating with an error. This logic is configurable with newly added `elasticsearch.max_retries` and `elasticsearch.retry_interval` configuration fields.

`elasticsearch.bulk.max_retries` was increased from 3 to 5 to match newly added `elasticsearch.max_retries` default value. Additionally `elasticsearch.bulk.retry_interval` was added with default value of 10 seconds (previously was hardcoded to 1 second) to match behaviour of newly added `elasticsearch.retry_interval`. As a result, bulk inserts will retry for 10 * (1 + 2 + 3 + 4 + 5) = 150 seconds, before this number was 1 * (1 + 2 + 3) = 6 seconds.


connectors/es/client.py
```diff
@@ -162,6 +159,80 @@ async def wait(self):
...
+                self._logger.debug(
+                    f"Attempt {retry}: api error with status {e.status_code}"
+                )
+
+                if e.status_code != 429:
+                    raise
+                if retry >= self._max_retries:
+                    raise
+
+            await self._sleep(retry)
```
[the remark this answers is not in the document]
artem-shelkovnikov: Elasticsearch does not send `Retry-After` headers, so no handling of it here

connectors/es/client.py
```diff
@@ -162,6 +159,80 @@ async def wait(self):
...
+        self._max_retries = max_retries
+        self._retry_timeout = retry_timeout
+        self._retry_strategy = retry_strategy
+
+    async def close(self):
+        self._sleeps.cancel()
+        self._keep_retrying = False
+
+    async def _sleep(self, retry):
+        time_to_sleep = time_to_sleep_between_retries(
```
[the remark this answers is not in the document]
artem-shelkovnikov: Can do that too, but without exposing the backoff strategy exponential backoff is much harder to control, e.g.:

Linear with retry_count=5 and retry_interval=10 will retry for max of 10 + 20 + 30 + 40 + 50 = 150 seconds.
Exponential with retry_count=5 and retry_interval=10 will retry for max of 10 + 100 + 1000 + 10000 + 100000 = 111110 seconds. So retry intervals more than 3-4 won't make sense.

I'd leave it like it is or expose retry strategy in config, what's better to you?

connectors/es/client.py
```diff
@@ -162,6 +159,80 @@ async def wait(self):
...
+                self._logger.debug(f"Attempt {retry}: connection timeout")
+
+                if retry >= self._max_retries:
+                    raise
+            except ApiError as e:
+                self._logger.debug(
+                    f"Attempt {retry}: api error with status {e.status_code}"
+                )
+
+                if e.status_code != 429:
```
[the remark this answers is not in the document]
artem-shelkovnikov: Makes sense, I think we can check for these ones and retry:

```
500 Internal Server Error
502 Bad Gateway
503 Service Unavailable
504 Gateway Timeout
```

connectors/config.py
```diff
@@ -59,8 +59,9 @@ def _default_config():
                 "max_concurrency": 5,
                 "chunk_max_mem_size": 5,
                 "concurrent_downloads": 10,
-                "max_retries": 3,
```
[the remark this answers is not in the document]
artem-shelkovnikov: I technically moved this `elasticsearch.bulk.max_retries` to the new `elasticsearch.max_retries` + increased it. So it won't work _exactly_ like before, rather retry harder.

I don't think we've done a lot of it before, should I write a warning that this setting is deprecated and the new one should be used?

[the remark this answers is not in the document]
vidok: Yeah, adding more information about this change would be useful. A sentence or two in the release notes would be great too.

connectors/es/client.py
```diff
@@ -41,6 +46,10 @@ def __init__(self, config):
             use_default_ports_for_scheme=True,
         )
         self._sleeps = CancellableSleeps()
+        self._retrier = TransientElasticsearchRetrier(
+            logger, config.get("max_retries", 5), config.get("retry_timeout", 10)
```
vidok, line 50: I prefer to use constants like `DEFAULT_MAX_RETRIES` instead of the raw number. It's just easier to find and change them.

connectors/config.py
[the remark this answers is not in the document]
artem-shelkovnikov: I've added info to the release notes in the PR + made a change in the commit https://github.com/elastic/connectors/pull/2063/commits/a4d4f8bd1b15113d12f119920ea5ecb0753b48a7 to check deprecated configs and report to terminal.

Example message in the terminal:

```
artemshelkovnikov-m2@Work-M2 connectors-py % make run
bin/elastic-ingest
[FMWK][14:09:29][INFO] Running connector service version 8.13.0.0
[FMWK][14:09:29][INFO] Loading config from /Users/artemshelkovnikov-m2/git_tree/connectors-py/connectors/../config.yml
[FMWK][14:09:29][WARNING] Configuration option 'elasticsearch.bulk.max_retries' is deprecated: Use elasticsearch.max_retries instead.
```

connectors/es/client.py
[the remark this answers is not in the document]
artem-shelkovnikov: Added in https://github.com/elastic/connectors/pull/2063/commits/3a111f14b1203fbdf2bbfe3407485cccd646e039

connectors/es/client.py
artem-shelkovnikov: Added constants in https://github.com/elastic/connectors/pull/2063/commits/a4d4f8bd1b15113d12f119920ea5ecb0753b48a7

artem-shelkovnikov:
@elastic/ingestion-team 

Thinking of config options additionally - we have `elasticsearch.initial_backoff`, `elasticsearch.backoff_multiplier` and `elasticsearch.max_wait_duration` that is specific to Elasticsearch client `wait` action that happens during startup.

I'm adding `elasticsearch.max_retries` and `elasticsearch.retry_interval` that are related to all elasticsearch calls.

Does it make it confusing that these options are in the root of `elasticsearch` config option? Should I put the newly added options into some nested key, like `elasticsearch.connection` or similar?

connectors/es/sink.py
```diff
@@ -130,20 +129,16 @@ def _bulk_op(self, doc, operation=OP_INDEX):
...
         # TODO: treat result to retry errors like in async_streaming_bulk
         task_num = len(self.bulk_tasks)
 
         if self._logger.isEnabledFor(logging.DEBUG):
             self._logger.debug(
                 f"Task {task_num} - Sending a batch of {len(operations)} ops -- {get_mb_size(operations)}MiB"
             )
-        res = await _bulk_api_call()
+
+        # TODO: retry 429s for individual items here
```
vidok, line 140: Could you please expand more on this comment? I don't fully understand how can we handle individual items errors.

artem-shelkovnikov: So now we have 500 items that we send, then if 1 fails we retry the whole batch.

What I wanna do in later PRs:

- See if error is retriable. If so, then put it back to the queue to re-ingest in a different batch
- If error is not retriable, raise following the resilience protocol we have (not raise if number of errors during ingestion does not exceed certain %, so if 0.5% of items are not ingested we don't raise, but if we're at 1% then we raise)
- Maybe have some sort of "dead letter queue" for these items? (that could happen cause of mappings blowing up with too many fields generated and such). This I probably won't do without extensive discussion and seeing if we even want that

connectors/es/client.py
```diff
@@ -162,6 +165,81 @@ async def wait(self):
         await self.close()
         return False
 
+    async def ping(self):
+        try:
+            await self.client.info()
```
wangch079, line 170: Why `ping` is not retried?

vidok:
> @elastic/ingestion-team
> 
> Thinking of config options additionally - we have `elasticsearch.initial_backoff`, `elasticsearch.backoff_multiplier` and `elasticsearch.max_wait_duration` that is specific to Elasticsearch client `wait` action that happens during startup.
> 
> I'm adding `elasticsearch.max_retries` and `elasticsearch.retry_interval` that are related to all elasticsearch calls.
> 
> Does it make it confusing that these options are in the root of `elasticsearch` config option? Should I put the newly added options into some nested key, like `elasticsearch.connection` or similar?

Reposting my comment for the record (repeating your ideas :) ): 

```
The new keys `elasticsearch.max_retries` and `elasticsearch.retry_interval` should be in the root because the other root keys already indicate they are used as Elasticsearch connection configuration. 

The old keys can be moved under another prefix because they impact a specific use case. 
```

connectors/config.py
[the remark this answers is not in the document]
artem-shelkovnikov: We currently don't have any special handling for bulk tbh, before it was just blindly retrying any error `{bulk.max_retries}` times. Now I introduced this logic in a slightly better manner to all ES calls, so I don't think we want it separate.

Re: retry only failed operations - we don't do this at all now, and this will be addressed later.

connectors/es/client.py
artem-shelkovnikov: It's not retried cause of `wait` method - `wait` calls `ping` and does retrying based on its own config options. See [method](https://github.com/elastic/connectors/blob/8e431d529c557e6a6bb3b6a95f2b20b92cb965dd/connectors/es/client.py#L148):

```python
    async def wait(self):
        backoff = self.initial_backoff_duration
        # ...
        while time.time() - start < self.max_wait_duration:
            if not self._keep_waiting:
                await self.close()
                return False

            # ...
            if await self.ping():
                return True
            await self._sleeps.sleep(backoff)
            backoff *= self.backoff_multiplier

        await self.close()
        return False
```

artem-shelkovnikov:
@seanstory @vidok @wangch079 I've addressed all the points that I've seen in your comments, are there any other things that should be addressed before this PR can be approved?

vidok (dismissed):
Fantastic work 🚀 !

connectors/config.py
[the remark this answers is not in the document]
wangch079: >We currently don't have any special handling for bulk tbh

I know. My point is, there will be special retry logic for `bulk` function, and this `max_retries` is originally for `bulk` function, so if you keep this `max_retries` unchanged (reserved for future `bulk` retry handling), and introduce new configs in `elasticsearch` level for all non-bulk functions, you can avoid breaking change and deprecation warning.

Just a suggestion, I'm not so worried about config deprecation.

wangch079 (dismissed):
Left some comments, but it's totally fine to merge it now.

connectors/es/client.py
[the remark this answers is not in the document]
seanstory: :shockedpikachu:

connectors/config.py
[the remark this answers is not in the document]
artem-shelkovnikov: Good call, I'll do it like this while we're waiting for response about our config deprecation strategy!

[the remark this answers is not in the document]
artem-shelkovnikov: What I've done now:

Increased default `elasticsearch.bulk.max_retries` to 5 + added an option `elasticsearch.bulk.retry_interval` set to 10 so that current behaviour of bulk insert would be absolutely same as other calls to Elasticsearch that are configured by `elasticsearch.max_retries` and `elasticsearch.retry_interval`.

We can deprecate `elasticsearch.bulk.max_retries|retry_interval` later, what's important to me now is that these two by default would be same.

connectors/config.py
```diff
@@ -60,9 +64,12 @@ def _default_config():
                 "chunk_size": 1000,
                 "max_concurrency": 5,
                 "chunk_max_mem_size": 5,
+                "max_retries": 5,
```
seanstory, line 67: Will need a release note that the default values changed for `max_retries` and `retry_interval`

connectors/config.py
```diff
@@ -60,9 +64,12 @@ def _default_config():
...
+                "retry_interval": 10,
```
seanstory, line 68: nit: These two could use the constants, too.

connectors/utils.py
```diff
@@ -35,7 +35,8 @@
 DEFAULT_CHUNK_MEM_SIZE = 25
 DEFAULT_MAX_CONCURRENCY = 5
 DEFAULT_CONCURRENT_DOWNLOADS = 10
-DEFAULT_BULK_MAX_RETRIES = 3
+DEFAULT_BULK_MAX_RETRIES = 5
+DEFAULT_BULK_RETRY_INTERVAL = 10
```
seanstory, line 39: Should these be imported from config.py?

seanstory (approved):

connectors/config.py
artem-shelkovnikov: Do you think current release note is okay?

> Release Note
> Improve connection resiliency when accessing elasticsearch - retry all failed calls that fail because of unhealthy Elasticsearch before terminating with an error. This logic is configurable with newly added elasticsearch.max_retries and elasticsearch.retry_interval configuration fields.
> 
> elasticsearch.bulk.max_retries was increased from 3 to 5 to match newly added elasticsearch.max_retries default value. Additionally elasticsearch.bulk.retry_interval was added with default value of 10 seconds (previously was hardcoded to 1 second) to match behaviour of newly added elasticsearch.retry_interval. As a result, bulk inserts will retry for 10 * (1 + 2 + 3 + 4 + 5) = 150 seconds, before this number was 1 * (1 + 2 + 3) = 6 seconds.

connectors/utils.py
artem-shelkovnikov: Ah these are left after me refactoring things, I'll delete these, thanks!

artem-shelkovnikov: Addressed in 244ce2e7e29534bf2c08507e5fa59f496524e228

connectors/config.py
artem-shelkovnikov: Addressed in 244ce2e7e29534bf2c08507e5fa59f496524e228

connectors/config.py
seanstory: Ah shit - I wrote this and thought I'd remember to check your release note before hitting submit. And I forgot 🤦

seanstory (dismissed):

vidok (approved):

PR #2076, opened 2024-01-19
Replace usages of 'elastic_server' with 'sync_orchestrator'

seanstory:
I've been wanting to do this forever. The class SyncOrchetrator used to be called just Server and was in the `es` directory, so instances of it were referred to as `elastic_server`. Which is a total misnomer of what this is and does.

This is just a global find/replace - no logic changed.

#### Pre-Review Checklist


wangch079 (approved):
Nice work

PR #2080, opened 2024-01-22
Add max_file_download_size config and expose to DataSource

seanstory:
## Closes https://github.com/elastic/enterprise-search-team/issues/4716

Adds a new config, `service.max_file_download_size` that defaults to 10MB.

Also, adds a new abstraction for specific framework configs that we want to expose to the DataSource instances, without exposing all of our YAML configs. This PR exposes only the new `max_file_size`, but can be easily updated to expose more.

#### Pre-Review Checklist

## Release Note

`service.max_file_download_size` can now be set in the `config.yml` to control how large of files the framework will attempt to download. This config is only used for connectors that are not using the Data Extraction Service.

### DIFF
--- a/connectors/config.py
+++ b/connectors/config.py
@@ -10,7 +10,7 @@ from envyaml import EnvYAML
 
 from connectors.logger import logger
 
-
+DEFAULT_MAX_FILE_SIZE = 10485760 # 10MB
 def load_config(config_file):
     logger.info(f"Loading config from {config_file}")
     yaml_config = EnvYAML(config_file, flatten=False).export()
@@ -77,6 +77,7 @@ def _default_config():
             "max_errors_span": 600,
             "max_concurrent_content_syncs": 1,
             "max_concurrent_access_control_syncs": 1,
+            "max_file_download_size": DEFAULT_MAX_FILE_SIZE,
             "job_cleanup_interval": 300,
             "log_level": "INFO",
         },
@@ -177,3 +178,29 @@ def _merge_dicts(hsh1, hsh2):
             yield (k, hsh1[k])
         else:
             yield (k, hsh2[k])
+
+class DataSourceFrameworkConfig:
+    """
+    The configs that will be exposed to DataSource instances.
+    This abstraction prevents DataSource instances from having access to all configuration, while also
+    preventing them from requiring substantial changes to access new configs that may be added.
+    """
+
+    def __init__(self, max_file_size):
+        """
+        Should not be called directly. Use the Builder.
+        """
+        self.max_file_size = max_file_size
+
+
+    class Builder:
+        def __init__(self):
+            self.max_file_size = DEFAULT_MAX_FILE_SIZE
+
+        def with_max_file_size(self, max_file_size):
+            self.max_file_size = max_file_size
+            return self
+
+        def build(self):
+            return DataSourceFrameworkConfig(self.max_file_size)
+
--- a/connectors/sources/box.py
+++ b/connectors/sources/box.py
@@ -37,7 +37,6 @@ ENDPOINTS = {
 }
 RETRIES = 3
 RETRY_INTERVAL = 2
-FILE_SIZE_LIMIT = 10485760
 CHUNK_SIZE = 1024
 FETCH_LIMIT = 1000
 QUEUE_MEM_SIZE = 5 * 1024 * 1024  # ~ 5 MB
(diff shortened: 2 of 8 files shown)


connectors/config.py
```diff
@@ -177,3 +180,28 @@ def _merge_dicts(hsh1, hsh2):
...
+    class Builder:
+        def __init__(self):
+            self.max_file_size = DEFAULT_MAX_FILE_SIZE
+
+        def with_max_file_size(self, max_file_size):
+            self.max_file_size = max_file_size
+            return self
+
+        def build(self):
+            return DataSourceFrameworkConfig(self.max_file_size)
```
artem-shelkovnikov, line 207: Curious - why use builder here?

For now it's only one property, and probably in future it'll be like 10.

Do you see partial initialisation of this config (e.g. with some properties initialised, some not), or multistep initialisation of config?

connectors/sources/box.py
```diff
@@ -391,9 +390,9 @@ def _pre_checks_for_get_content(
             )
             return False
 
-        if attachment_size > FILE_SIZE_LIMIT:
+        if attachment_size > self.framework_config.max_file_size:
             self._logger.warning(
-                f"File size {attachment_size} of file {attachment_name} is larger than {FILE_SIZE_LIMIT} bytes. Discarding file content"
+                f"File size {attachment_size} of file {attachment_name} is larger than {self.framework_config.max_file_size} bytes. Discarding file content"
```
artem-shelkovnikov, line 395: I was also wondering if it's possible to move this condition elsewhere, so that each connector would not call it?

Best I could think of was that rather than doing comparison manually, make the class "richer" and do something like:

```
if self.framework_config.file_size_supported(attachment_size):
    # download
```

How do you feel?

connectors/config.py
seanstory: I want to make it easy for this to be formed up from multiple places eventually. We had a recent slack thread where we discussed how some service configs might one day have a UI and be stored in ES. And we've got some in the YAML, and maybe would someday have some in RCFs. Passing around a half-complete builder is a good pattern to use for scenarios like this where multiple actors are all contributing parts to something being built.

connectors/sources/box.py
seanstory: I'm not going to tackle that as part of this - my focus for this PR is making the existing logic configurable. I believe there was debate previously about moving this kind of thing to the Extraction utils, but that ended up being rejected, and I don't want to re-open that discussion at the moment.

connectors/config.py
artem-shelkovnikov: Yeah I understand that, though I felt that purpose of Builder is often to have a mutable object until at some point you're ready to create an immutable object and use it somewhere. Since Python has different encapsulation concepts, it's much less necessary, and Builder becomes more of a "convenience" class that makes creation of objects more pleasant.

I'd think that just having `DataSourceFrameworkConfig` class with getters/setters should be fine.

This however should not prevent from merging change as is, it's rather a discussion to see your POV here :)

artem-shelkovnikov (approved):

connectors/config.py
wangch079: I think the benefit of Builder pattern is that it makes the object instantiation simple and readable when there are a lot of variables. Apparently this is implemented for future-proof.

Do you think we should define the constructor of `DataSourceFrameworkConfig` like this:
```python
def __init__(self, builder):
	self.max_file_size = builder.max_file_size
```
and the `build` method like this:
```python
def build(self):
	return DataSourceFrameworkConfig(self)
```

artem-shelkovnikov: I think the truth is there will be N sources that provide config options, e.g.:

- Configuration file
- Connector-specific config
- Deployment-wide settings

So builder will have 3 args/3 `with_x` methods and number is unlikely to grow IMO, I might be wrong here.

As to the syntax, I think original syntax in Sean's changeset is already good Builder syntax - only improvement (also absolutely unnecessary) could be is to make DataSourceFrameworkConfig have read-only properties with data stored in `__`-prefixed keys so that they could not be accessed, but I don't think it matters a lot.

wangch079: >I think the truth is there will be N sources that provide config options, e.g.:
> 
> - Configuration file
> - Connector-specific config
> - Deployment-wide settings
> 
> So builder will have 3 args/3 with_x methods


I think the builder should not care of where this config is coming from. Take `max_file_size` as example, it should only define one method `with_max_file_size`. If it's from configuration file, you can do `builder.with_max_file_size(loaded_config.max_file_size)`, and if it's from a connector, you can do `builder.with_max_file_size(connector.max_file_size)`

seanstory: Let's just keep these options in mind as this grows. I think we've agreed that the current state is "fine", and that it could be improved on depending on how we use it in the future. So let's just plan to refactor as necessary as its usage grows.

PR #2083, opened 2024-01-23
[Redis] Add ftests 

akanshi-elastic:
Part of https://github.com/elastic/connectors/issues/1941

#### Pre-Review Checklist


tests/sources/fixtures/redis/fixture.py
```diff
@@ -0,0 +1,47 @@
...
+# ruff: noqa: T201
+import os
+import random
+
+import redis.asyncio as redis
+
+from tests.commons import WeightedFakeProvider
+
+DATA_SIZE = os.environ.get("DATA_SIZE", "small").lower()
+_NUM_TABLES = {"small": 2, "medium": 4, "large": 6}
```
artem-shelkovnikov, line 15: How much does this affect time of the run? Ideally, large should take 20+ minutes, while small should take <1 minute

tests/sources/fixtures/redis/fixture.py
```diff
@@ -0,0 +1,47 @@
...
+    for row_id in range(lines):
+        row_id += start
+        key = f"user_{row_id}"
+        rows[key] = fake_provider.get_text()
+    await redis_client.mset(rows)
+
+
+async def load():
+    """N tables of 10001 rows each. each row is ~ 1024*20 bytes"""
+    for table in range(NUM_TABLES):
```
wangch079, line 33: So the `table` is the database? if so, please rename it.

tests/sources/fixtures/redis/fixture.py
```diff
@@ -0,0 +1,47 @@
...
+        print(f"Adding data in {table}...")
+        redis_client = await redis.from_url(f"{ENDPOINT}{table}")
```
wangch079, line 35: Not necessary to re-instantiate the client for every database.

tests/sources/fixtures/redis/fixture.py
```diff
@@ -0,0 +1,47 @@
...
+ENDPOINT = "redis://localhost:6379/"
+
+fake_provider = WeightedFakeProvider(weights=[0.65, 0.3, 0.05, 0])
+
+
+async def inject_lines(redis_client, table, start, lines):
+    rows = {}
+    for row_id in range(lines):
+        row_id += start
+        key = f"user_{row_id}"
```
wangch079, line 26: What's the purpose of `start` here? Even though you don't use `start`, you can guarantee the uniqueness of the key within this database. Or is it required that the key must be unique across databases?

If it's required, I think you can also code it like
```suggestion
        key = f"user_{row_id + start}"
```

Instead of assigning it to `row_id`.

tests/sources/fixtures/redis/docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+    environment:
+      ELASTICSEARCH_URL: http://host.docker.internal:9200
+      ELASTICSEARCH_HOSTS: http://host.docker.internal:9200
+      ENTERPRISESEARCH_HOST: http://host.docker.internal:3002
+      ELASTICSEARCH_USERNAME: kibana_system
+      ELASTICSEARCH_PASSWORD: changeme
+    networks:
+      - esnet
+
+  enterprise_search:
```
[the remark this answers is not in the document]
artem-shelkovnikov: Do we already not need it for index creation/migrations?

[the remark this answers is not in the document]
wangch079: Do we need `migration`? It will be a new instance.

I think the index creation is handled in ES as an index template?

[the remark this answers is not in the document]
artem-shelkovnikov: Migrations definitely not. If index is created in ES, then yes, we can remove ent-search everywhere in ftests

tests/sources/fixtures/redis/fixture.py
akanshi-elastic: Yes, it is taking less than a minute to complete this.

tests/sources/fixtures/redis/docker-compose.yml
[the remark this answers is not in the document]
vidok: +1 to remove `enterprice_search`.  We can use Connectors CLI to create content indices.

artem-shelkovnikov:
@akanshi-elastic can you share Perf8 stats running this on large dataset?

tests/sources/fixtures/redis/fixture.py
```diff
@@ -0,0 +1,52 @@
...
+ENDPOINT = "redis://localhost:6379/"
+
+fake_provider = WeightedFakeProvider(weights=[0.65, 0.3, 0.05, 0])
+
+
+async def inject_lines(redis_client, start, lines):
+    text = fake_provider.get_text()
+    rows = {}
+    for row_id in range(lines):
+        key = f"user_{row_id + start}"
```
wangch079, line 28: I still don't get the use of `start`. for db `0`, the key will be `user_0` to `user_499`, and for db `1`, the key will be `user_1` to `user_500`. Assuming the key can be the same in different databases, I don't see the point of using `start` here.

tests/sources/fixtures/redis/fixture.py
```diff
@@ -0,0 +1,52 @@
...
+
+async def remove():
+    """Removes 10 random items per db"""
+    redis_client = await redis.from_url(f"{ENDPOINT}")
+    for db in range(NUM_DB):
+        print(f"Working on db {db}...")
+        await redis_client.execute_command("SELECT", db)
+        keys = [
+            f"user_{row_id}"
+            for row_id in random.sample(range(1, 100), RECORDS_TO_DELETE)
```
wangch079, line 50: `user_1` is a possible key and it won't exist for db `2`. Will it fail the test if it tries to delete a non-existing key?

tests/sources/fixtures/redis/fixture.py
akanshi-elastic: Right, Actual point here is to make the key unique, which can be done by this: `key = f"user_{row_id}_{start}"` here `start` will make the key unique every time because start cannot be same.

tests/sources/fixtures/redis/fixture.py
akanshi-elastic: No this will not throw any error, as it is bulk delete method

tests/sources/fixtures/redis/fixture.py
wangch079: > If we remove `start` from key then key can be same in db0 and db1 which create same `_id` of a document i.e. duplicacy of IDs.

Interesting. So it's ok to have the same key in `db0` and `db1`? If so, we should remove `start` here, the same key in `db0` and `db1` should generate different `_id` in ES document.

artem-shelkovnikov:
@akanshi-elastic can you attach all the artefacts from Perf8 response? I see you added `report.csv` file into my comment, but it's not sufficient.

Also, please don't edit my comments, add new comments and quote my comment if you want to respond to it.

tests/sources/fixtures/redis/fixture.py
akanshi-elastic: No it's important to be unique _id and for that I have changed it to `key = f"user_{row_id}_{start}"`

akanshi-elastic:
> @akanshi-elastic can you attach all the artefacts from Perf8 response? I see you added `report.csv` file into my comment, but it's not sufficient.
> 
> Also, please don't edit my comments, add new comments and quote my comment if you want to respond to it.

[perf8-report-redis.zip](https://github.com/elastic/connectors/files/14107457/perf8-report-redis.zip)

tests/sources/fixtures/redis/fixture.py
wangch079: A simple test shows it can have the same key in different dbs:
```
127.0.0.1:6379> select 0
OK
127.0.0.1:6379> set hello "value in db 0"
OK
127.0.0.1:6379> get hello
"value in db 0"
127.0.0.1:6379> select 1
OK
127.0.0.1:6379[1]> get hello
(nil)
127.0.0.1:6379[1]> set hello "value in db 1"
OK
127.0.0.1:6379[1]> get hello
"value in db 1"
127.0.0.1:6379[1]> select 0
OK
127.0.0.1:6379> get hello
"value in db 0"
```

In this case, I'd advise you to remove `start` here:
```python
key = f"user_{row_id}"
```

As `user_0` in db0 and db1 should result in different document id in ES.

artem-shelkovnikov:
Looking at performance report, why does the connector write so much to disk?

image

akanshi-elastic:
> Looking at performance report, why does the connector write so much to disk?
> 
> image

@artem-shelkovnikov It looks like redis server is consuming the disk space as it is an in-memory database so it storing all the records in disk. I have verified from connector side, all the session are closing properly and we are not storing any data into list or dictionary which can consume the disk

artem-shelkovnikov (approved):

wangch079 (approved):

PR #2086, opened 2024-01-24
Fix packaging

vidok:
Fix connectors packaging in Enterprise Search. 

#### This step is failing: 
```
python3 -m pip install .
```

#### Stack trace:
```
Processing ~/projects/connectors-python
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error

  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [8 lines of output]
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "~/projects/connectors-python/setup.py", line 12, in <module>
        File "~/projects/connectors-python/connectors/logger.py", line 17, in <module>
          import ecs_logging
      ModuleNotFoundError: No module named 'ecs_logging'
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

We can't use modules listed in requirements.txt files during `setup.py` execution. 

#### Pre-Review Checklist

## Related Pull Requests
* https://github.com/elastic/connectors/pull/2071
* https://github.com/elastic/connectors/pull/2070

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,20 +4,24 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 import os
+import sys
 
 from setuptools import find_packages, setup
 from setuptools._vendor.packaging.markers import Marker
 
-from connectors.logger import logger
-from connectors.utils import ensure_python_3_10_or_higher
-
 try:
     ARCH = os.uname().machine
+    raise Exception("test")
 except Exception as e:
     ARCH = "x86_64"
-    logger.info(
+    print(
         f"Defaulting to architecture '{ARCH}'. Unable to determine machine architecture due to error: {e}"
     )
+
+def ensure_python_3_10_or_higher():
+    if sys.version_info < (3, 10):
+        msg = "Requires Python 3.10 or higher."
+        raise ValueError(msg)
 
 ensure_python_3_10_or_higher()
 
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,6 @@ from setuptools._vendor.packaging.markers import Marker
 
 try:
     ARCH = os.uname().machine
-    raise Exception("test")
 except Exception as e:
     ARCH = "x86_64"
     print(
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ try:
     ARCH = os.uname().machine
 except Exception as e:
     ARCH = "x86_64"
-    print(
+    print(  # noqa: T201
         f"Defaulting to architecture '{ARCH}'. Unable to determine machine architecture due to error: {e}"
     )
 
--- a/setup.py
+++ b/setup.py
@@ -18,12 +18,9 @@ except Exception as e:
         f"Defaulting to architecture '{ARCH}'. Unable to determine machine architecture due to error: {e}"
     )
 
-def ensure_python_3_10_or_higher():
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -4,21 +4,25 @@
...
 
 from setuptools import find_packages, setup
 from setuptools._vendor.packaging.markers import Marker
 
-from connectors.logger import logger
-from connectors.utils import ensure_python_3_10_or_higher
-
 try:
     ARCH = os.uname().machine
+    raise Exception("test")
```
artem-shelkovnikov, line 14: Should it still be here?

setup.py
```diff
@@ -4,21 +4,25 @@
...
 except Exception as e:
     ARCH = "x86_64"
-    logger.info(
+    print(
```
artem-shelkovnikov, line 17: Why remove logger?

Maybe raise here too?

setup.py
vidok: oops

setup.py
artem-shelkovnikov: Ah you are right, logger should not be here

artem-shelkovnikov (dismissed):

setup.py
```diff
@@ -4,22 +4,22 @@
...
+import sys
 
 from setuptools import find_packages, setup
 from setuptools._vendor.packaging.markers import Marker
 
-from connectors.logger import logger
-from connectors.utils import ensure_python_3_10_or_higher
-
 try:
     ARCH = os.uname().machine
```
markjhoy, line 13: @vidok - thanks for looking at this - I'm concerned a bit still as it looks like this may be the line that is actually causing the original exception to be thrown... defaulting to `x86_64` in most cases I think is safe, however, what will be the consequence of defaulting to this on an `arm64` architecture (as this process is used in both the x76 and aarch64 builds in our processes)?

vidok: This PR does not change anything but the usage of `connectors.logger`. We just can't use it in setup.py. 

> I'm concerned a bit still as it looks like this may be the line that is actually causing the original exception to be thrown. 

nope, it failed on line 12 because it loads `connectors.logger` from `connectors.utils`. In turn, `connectors.logger`  depends on `ecs-logging` that is not available during the setup.py execution.

timgrein (commented):
Good catch! I remember it was hard to test the python version enforcement in a new `test_setup.py` file. Did you also try it? I think tests are valuable here as they also express intent

setup.py
vidok: Yep, it's been there for more than 1 year.

vidok:
> Good catch and thank you for directly fixing it! I remember it was hard to test the python version enforcement in a new `test_setup.py` file. Did you also try it? I think tests are valuable here as they also express intent

There is no need for `test_setup.py` because setup.py is executed outside of the package runtime (technically nobody stops us doing that). AFAIK setup.py should contain as less dynamic code as possible so we need to add an extra step to the CI pipelines when we try to install the elastic-connectors package.

seanstory (approved):
:rubberstamp:

PR #2093, opened 2024-01-29
Validate configurable fields every loop of job scheduling service

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/5678

Adding logic to run validation for content source configuration with every loop of scheduler service.

Before, configurable fields would be validated only when a sync job is ran, but now we're gonna run validation every time `job_scheduling_service` runs (every 30 seconds). That will actually expose errors in configurations to the customer eariler.

#### Pre-Review Checklist

## Release Note

Now configurations of connectors are validated periodically, exposing validation errors without need to run a sync.

### DIFF
--- a/connectors/services/job_scheduling.py
+++ b/connectors/services/job_scheduling.py
@@ -88,9 +88,19 @@ class JobSchedulingService(BaseService):
             raise DataSourceError(msg)
 
         source_klass = get_source_klass(self.source_list[connector.service_type])
+        data_source = source_klass(connector.configuration)
+        data_source.set_logger(connector.logger)
+
+        connector.log_debug("Validating configuration")
+        try:
+            data_source.validate_config_fields()
+            await data_source.validate_config()
+        except Exception as e:
+            connector.log_error(e, exc_info=True)
+            await connector.error(e)
+            return
+
         if connector.features.sync_rules_enabled():
-            data_source = source_klass(connector.configuration)
-            data_source.set_logger(connector.logger)
             try:
                 await connector.validate_filtering(validator=data_source)
             finally:
--- a/tests/services/test_job_scheduling.py
+++ b/tests/services/test_job_scheduling.py
@@ -343,3 +343,32 @@ async def test_run_when_sync_fails_then_continues_service_execution(
     # assert that service did not crash and kept asking index for connectors
     # we don't have a good criteria of what a "crashed service is"
     assert connector_index_mock.supported_connectors.call_count > 1
+
+
+@pytest.mark.asyncio
+@patch("connectors.services.job_scheduling.get_source_klass")
+async def test_run_when_connector_fields_are_invalid(
+    get_source_klass_mock, connector_index_mock, set_env
+):
+    error_message = "Something invalid is in config!"
+    actual_error = Exception(error_message)
+
+    data_source_mock = Mock()
+
+    def _source_klass(config):
+        return data_source_mock
+
+    get_source_klass_mock.return_value = _source_klass
+
+    data_source_mock.validate_config_fields = Mock()
+    data_source_mock.validate_config = AsyncMock(side_effect=[actual_error])
+    data_source_mock.close = AsyncMock()
+
+    connector = mock_connector(next_sync=datetime.utcnow())
+    connector_index_mock.supported_connectors.return_value = AsyncIterator([connector])
+    await create_and_run_service(JobSchedulingService, stop_after=0.15)
+
+    data_source_mock.validate_config_fields.assert_called()
+    data_source_mock.validate_config.assert_awaited_once()
+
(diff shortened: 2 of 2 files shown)


connectors/services/job_scheduling.py
```diff
@@ -88,9 +88,19 @@ async def _schedule(self, connector):
             raise DataSourceError(msg)
 
         source_klass = get_source_klass(self.source_list[connector.service_type])
+        data_source = source_klass(connector.configuration)
+        data_source.set_logger(connector.logger)
+
+        connector.log_debug("Validating configuration")
+        try:
+            data_source.validate_config_fields()
+            await data_source.validate_config()
```
seanstory, line 97: should we go ahead and call `data_source.ping` while we're at it?

seanstory (approved):
LGTM, but one small suggestion.

connectors/services/job_scheduling.py
```diff
@@ -88,9 +88,19 @@ async def _schedule(self, connector):
             raise DataSourceError(msg)
 
         source_klass = get_source_klass(self.source_list[connector.service_type])
+        data_source = source_klass(connector.configuration)
+        data_source.set_logger(connector.logger)
+
+        connector.log_debug("Validating configuration")
+        try:
+            data_source.validate_config_fields()
```
wangch079, line 96: Isn't this validated by `prepare`?
https://github.com/elastic/connectors/blob/17a76e99c2b3fc0151ba5c65eceb541f84dc6b2f/connectors/services/job_scheduling.py#L51

connectors/services/job_scheduling.py
wangch079, line 97: This validates the config of each connector, we can validate it together with this code block:
https://github.com/elastic/connectors/blob/17a76e99c2b3fc0151ba5c65eceb541f84dc6b2f/connectors/services/job_scheduling.py#L91-L97

connectors/services/job_scheduling.py
artem-shelkovnikov: I haven't found any validation logic in prepare.

`connector.prepare` has no access to instance of data source with configuration, rather it's a wrapper on top of an index record that does limited "sanity" validation for the connector fields, not RCFs.

connectors/services/job_scheduling.py
artem-shelkovnikov: I wanted to do so, but decided to split it into 2 prs for simplicity of reviewing :D

artem-shelkovnikov: I'm sorry, I don't understand the comment, can you elaborate more?

connectors/services/job_scheduling.py
wangch079: The `data_source.validate_config_fields()` here validates whether there are missing configs in the connector's configuration. 

https://github.com/elastic/connectors/blob/8cae11f4d9e652c047a5a970e599e066645f29e7/connectors/source.py#L546-L555

In `prepare`, it tries to call `validated_doc`, which will simply update the connector's configs if there's missing configs

https://github.com/elastic/connectors/blob/8cae11f4d9e652c047a5a970e599e066645f29e7/connectors/protocol/connectors.py#L807

connectors/services/job_scheduling.py
wangch079: Ignore me, I didn't see you moved the below code out of the if branch
https://github.com/elastic/connectors/blob/17a76e99c2b3fc0151ba5c65eceb541f84dc6b2f/connectors/services/job_scheduling.py#L92-L93

connectors/services/job_scheduling.py
artem-shelkovnikov: Hmmmm, I copied this logic from another place that validates fields:

https://github.com/elastic/connectors/blob/main/connectors/sync_job_runner.py#L129-131

```python
            self.sync_job.log_debug("Validating configuration")
            self.data_provider.validate_config_fields()
            await self.data_provider.validate_config()
```

Should we remove it in both places?

I'd honestly keep it, cause we haven't fully addressed migrating fields back and forth (deleting/modifying), so it _could_ work as some fail-safe mechanism that will raise if we migrated something wrong. How do you feel?

wangch079: When a sync job is created, it copies the configuration from the connector, and when it's executed, it will use the configuration in the sync job.

The validation in sync_job_runner is to validate the config in the sync job and should be kept.

In Job scheduling, it tries to check for each connector whether there's any discranpancy between the connector's configuration and the data source's configuration, if so, it will try to update the connector's, instead of raising an error. This is happening in the `prepare` method.

I doubt that the validation you added will ever raise, because it happens after `prepare`.

PR #2097, opened 2024-01-30
[CI] Update 'connectors-nightly-dockers' to stop trying to install Docker

acrewdson:
(ad hoc change)

The `connectors-nightly-docker` build recently started failing when trying to install Docker: https://buildkite.com/elastic/connectors-nightly-dockers/builds/1033

This is likely related to a recent change in the base VM image we're inheriting from. That image already has Docker and Docker Compose installed (and the Docker daemon is already running by default), so I think it should be safe to remove the steps where we install those packages at build time and then start the systemd unit.


markjhoy (approved):
I think this looks good

PR #2102, opened 2024-01-30
fix acl query template

seanstory:
## Closes https://github.com/elastic/connectors/issues/1696

In a review of the [DLS + Internal Knowledge Search blog](https://www.elastic.co/search-labs/blog/articles/dls-internal-knowledge-search), @xeraa [had commented](https://github.com/elastic/search-labs-elastic-co/pull/292#discussion_r1467288194) that our DLS query looked overly complex. This sent me down a rabbit hole, first figuring out that our example DLS app _was_ using an overly-complex query, then wondering why we were having to create an explicit query at all, then finally realizing I'd had these thoughts before and filed https://github.com/elastic/connectors/issues/1696 to eventually address the issue.

The time has come. 

This PR removes the outer `bool: { filter: {` clause that's no longer necessary (since Workplace Search) as we do not have a `_deny_access_control` pattern for any of our connectors (yet). 

It also fixes the syntax of the output document so that the template contents can be directly taken and shoved into a role descriptor.

#### Pre-Review Checklist

- [x] Security-related changes

## Related Pull Requests

* related update to the example app: https://github.com/elastic/elasticsearch-labs/pull/18 (we could replace most of the role descriptor once this PR is merged)

## Release Note

Changes the output structure of the documents created by ACL Syncs. Now these docs are appropriately structured to create Role Descriptors. The structure to access the identity and its permissions have not changed. This structure can take up to half as much space to store as the previous structure.

### DIFF
--- a/connectors/access_control.py
+++ b/connectors/access_control.py
@@ -5,6 +5,26 @@
 #
 
 ACCESS_CONTROL = "_allow_access_control"
+DLS_QUERY = """{
+                    "bool": {
+                        "should": [
+                            {
+                                "bool": {
+                                    "must_not": {
+                                        "exists": {
+                                            "field": "_allow_access_control"
+                                        }
+                                    }
+                                }
+                            },
+                            {
+                                "terms": {
+                                    "_allow_access_control.enum": {{#toJson}}access_control{{/toJson}}
+                                }
+                            }
+                        ]
+                    }
+                }"""
 
 
 def prefix_identity(prefix, identity):
@@ -27,26 +47,7 @@ def es_access_control_query(access_control):
         "query": {
             "template": {
                 "params": {"access_control": filtered_access_control},
-                "source": """{
-                    "bool": {
-                        "should": [
-                            {
-                                "bool": {
-                                    "must_not": {
-                                        "exists": {
-                                            "field": "_allow_access_control"
-                                        }
-                                    }
-                                }
-                            },
-                            {
-                                "terms": {
-                                    "_allow_access_control.enum": {{#toJson}}access_control{{/toJson}}
-                                }
-                            }
-                        ]
-                    }
-                }""",
+                "source": DLS_QUERY,
             }
         }
     }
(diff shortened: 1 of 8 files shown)


tests/sources/test_google_drive.py
```diff
@@ -1474,11 +1455,24 @@ async def test_prepare_access_control_doc(user, groups, access_control_doc):
...
             with mock.patch.object(ServiceAccountManager, "refresh"):
-                assert (
-                    access_control_doc
-                    == await source.prepare_single_access_control_document(user=user)
+                result = await source.prepare_single_access_control_document(user=user)
+
+                # mustache templates are hard to verify, so let's strip the mustache
+                filtered_input_source = " ".join(
+                    access_control_doc["query"]["template"]["source"]
+                    .replace("{{toJson}}access_control{{/toJson}}", "[]")
```
navarone-feekery, line 1463: Does the `#` in the toJson template string need to be included in this replace as well?

Something like `.replace("{{#toJson}}access_control{{/toJson}}", "[]")`

connectors/access_control.py
```diff
@@ -5,6 +5,26 @@
...
+                                    "must_not": {
+                                        "exists": {
+                                            "field": "_allow_access_control"
+                                        }
+                                    }
+                                }
+                            },
+                            {
+                                "terms": {
+                                    "_allow_access_control.enum": {{#toJson}}access_control{{/toJson}}
```
navarone-feekery, line 22: TIL about the [toJson search template](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-template.html#search-template-convert-json). Nice implementation!

navarone-feekery (approved):
LGTM, just one question

tests/sources/test_google_drive.py
seanstory: Huh. Yes, the `#` should matter. This must not be doing what I thought it was.

Turns out I can remove all the complexity I added to these tests. 🙈

seanstory:
@navarone-feekery can I get a re-approval?

navarone-feekery (approved):

PR #2103, opened 2024-01-30
[CI] Update nightly CI pipelines to use the right branches

acrewdson:
* we no longer need to build pre-`8.12`
* we should be building the same branches across all nightly pipelines


markjhoy (approved):
This _looks_ correct to me, but I'd defer to someone on the E&T Team to be certain

artem-shelkovnikov (approved):

artem-shelkovnikov:
Looks good to me! Thank you for doing this :)

PR #2104, opened 2024-01-31
[OTel] Initial integration

timgrein:
This is an old POC around APM and OTel for connectors


seanstory:
Bump @timgrein can you clean this up or close it?

PR #2105, opened 2024-01-31
Add logic to send a ping to connector's backend every loop

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/5678

Additionally to running RCF validation (see https://github.com/elastic/connectors/pull/2093) every time we check connector for whether the job should be scheduled, now we send a ping to the backend of this connector as well to verify, that it's available.

Before this logic would happen only before sync, which is too late.

#### Pre-Review Checklist

## Related Pull Requests

- https://github.com/elastic/connectors/pull/2093

## Release Note

Now configurations checks health of 3rd-party system periodically and reports errors to Kibana. Previously it was only done before the sync is started.

### DIFF
--- a/connectors/services/job_scheduling.py
+++ b/connectors/services/job_scheduling.py
@@ -91,10 +91,13 @@ class JobSchedulingService(BaseService):
         data_source = source_klass(connector.configuration)
         data_source.set_logger(connector.logger)
 
-        connector.log_debug("Validating configuration")
         try:
+            connector.log_debug("Validating configuration")
             data_source.validate_config_fields()
             await data_source.validate_config()
+
+            connector.log_debug("Pinging the backend")
+            await data_source.ping()
         except Exception as e:
             connector.log_error(e, exc_info=True)
             await connector.error(e)
--- a/tests/services/test_job_scheduling.py
+++ b/tests/services/test_job_scheduling.py
@@ -362,6 +362,7 @@ async def test_run_when_connector_fields_are_invalid(
 
     data_source_mock.validate_config_fields = Mock()
     data_source_mock.validate_config = AsyncMock(side_effect=[actual_error])
+    data_source_mock.ping = AsyncMock()
     data_source_mock.close = AsyncMock()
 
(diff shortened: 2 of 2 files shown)


connectors/services/job_scheduling.py
```diff
@@ -91,10 +91,13 @@ async def _schedule(self, connector):
...
         data_source.set_logger(connector.logger)
 
-        connector.log_debug("Validating configuration")
         try:
+            connector.log_debug("Validating configuration")
             data_source.validate_config_fields()
             await data_source.validate_config()
+
+            connector.log_debug("Pinging the backend")
+            await data_source.ping()
```
seanstory, line 100: Should we try/catch for a NotImplementedError? And just skip past it if it's not implemented? https://github.com/elastic/connectors/blob/main/connectors/source.py#L557-L562

tests/services/test_job_scheduling.py
```diff
@@ -370,5 +371,37 @@ def _source_klass(config):
...
+
+    connector.error.assert_awaited_with(actual_error)
+
+
+@pytest.mark.asyncio
+@patch("connectors.services.job_scheduling.get_source_klass")
+async def test_run_when_connector_ping_fails(
+    get_source_klass_mock, connector_index_mock, set_env
+):
+    error_message = "Something invalid is in config!"
```
seanstory, line 384:
```suggestion
    error_message = "Something went wrong when trying to ping the data source!"
```

seanstory (commented):
two small nits

connectors/services/job_scheduling.py
timgrein: Do we actually have data sources, which do not implement `ping`? Otherwise I would either keep it like that or log a clear error that `ping` is missing, if we expect it from every data source.

artem-shelkovnikov: We currently have no sources that do not implement ping.

I copied the logic from https://github.com/elastic/connectors/blob/main/connectors/sync_job_runner.py#L129-L134

PR #2107, opened 2024-01-31
Switch to an official Oracle container image in functional tests

acrewdson:
## Closes https://github.com/elastic/connectors/issues/1853

This updates the Oracle connector's functional tests to use an officlal container image from Oracle, instead of the community/third-party Oracle Express Edition image we have been using. According to [this post](https://medium.com/version-1/oracle-database-express-edition-now-database-free-ed5aa444152), Oracle Free appears to supersede Express Edition, so it probably makes sense to switch to testing against it.

Note however that with this change we are still can't enable the Oracle functional tests [on arm64](https://github.com/elastic/connectors/blob/main/.buildkite/nightly_steps.yml#L41), since getting an arm64 Oracle container image is [a bit more complicated](https://github.com/elastic/connectors/issues/1579#issuecomment-1924745013).

### DIFF
--- a/tests/sources/fixtures/oracle/fixture.py
+++ b/tests/sources/fixtures/oracle/fixture.py
@@ -21,10 +21,10 @@ RETRY_INTERVAL = 4
 RETRIES = 1
 BATCH_SIZE = 100
 
-USER = "admin"
+USER = "c##admin"
 PASSWORD = "Password_123"
 ENCODING = "UTF-8"
-DSN = "(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=127.0.0.1)(PORT=9090))(CONNECT_DATA=(SID=xe)))"
+DSN = "localhost:1521/FREE"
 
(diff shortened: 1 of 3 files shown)


tests/sources/fixtures/oracle/fixture.py
```diff
@@ -21,10 +21,10 @@
 RETRIES = 1
 BATCH_SIZE = 100
 
-USER = "admin"
+USER = "c##admin"
```
vidok, line 24: Just curious what "c##" means.

artem-shelkovnikov: I think it's like C++, but from Microsoft

image

acrewdson: @vidok seems like in newer multitenant-based Oracle versions, when you want to create a 'common' user (available in all containers), you have to prefix it with `c##`: 
> The username for the common user must be prefixed with "C##" or "c##" and contain only ASCII or EBCDIC characters.

We have to do this now because we are switching to a more modern version of Oracle.

PR #2108, opened 2024-02-01
Add DLS for Jira Data Center

moxarth-elastic:
## Part of https://github.com/elastic/connectors/issues/1957

Adding a support of DLS in Jira Data Center connector. There are some changes in the API endpoint regarding fetching of Atlassian users and object permissions.

**Limitation:** The API endpoint - `/rest/latest/user/search` for fetching the Data Center users will only fetch 1000 users from the AD. So, the Jira Data Center connector's access control will only show 1000 users.

#### Pre-Review Checklist


connectors/sources/jira.py
```diff
@@ -406,7 +408,6 @@ def get_default_configuration(cls):
             },
             "use_document_level_security": {
                 "display": "toggle",
-                "depends_on": [{"field": "data_source", "value": JIRA_CLOUD}],
                 "label": "Enable document level security",
                 "order": 14,
                 "tooltip": "Document level security ensures identities and permissions set in Jira are maintained in Elasticsearch. This enables you to restrict and personalize read-access users and groups have to documents in this index. Access control syncs ensure this metadata is kept up to date in your Elasticsearch documents.",
```
timgrein, line 413: Could you please add the information that you can only fetch 1000 users here?

connectors/sources/jira.py
```diff
@@ -65,7 +66,8 @@
     ATTACHMENT_CLOUD: "/rest/api/2/attachment/content/{attachment_id}",
     ATTACHMENT_SERVER: "/secure/attachment/{attachment_id}/{attachment_name}",
     USERS: "/rest/api/3/users/search",
-    PROJECT_PERMISSIONS_BY_KEY: "/rest/api/2/user/permission/search?projectKey={project_key}&permissions=BROWSE_PROJECTS&maxResults={max_results}&startAt={start_at}",
+    USERS_FOR_DATA_CENTER: "/rest/api/latest/user/search?username=''&startAt={start_at}&maxResults={max_results}",  # we can fetch only 1000 users for jira data center
```
timgrein, line 69: Is there some public source for this limitation we can add here?

connectors/sources/jira.py
```diff
@@ -445,35 +446,48 @@ def _decorate_with_access_control(self, document, access_control):
...
         start_at = 0
         while True:
             async for users in self.jira_client.api_call(
-                url_name=PROJECT_PERMISSIONS_BY_KEY,
-                project_key=project["key"],
+                url_name=PERMISSIONS_BY_KEY,
+                key=key,
                 start_at=start_at,
-                max_results=FETCH_SIZE,
+                max_results=1000,
```
timgrein, line 456: I would keep the constant here and maybe rename it to `MAX_USER_FETCH_LIMIT`

connectors/sources/jira.py
moxarth-elastic: There is no concrete discussion over this limitation, but it includes in this community post - https://community.atlassian.com/t5/Jira-Software-questions/Get-a-list-of-all-users-using-API-for-data-center/qaq-p/2401657. Do you think we should add this in the code?

connectors/sources/jira.py
moxarth-elastic: Added 👍

moxarth-elastic:
@timgrein can you please take a look at the changes? I've updated the PR as per your comments.

timgrein (approved):

PR #2111, opened 2024-02-01
Add problem URL to error message

seanstory:
See https://elastic.slack.com/archives/C7LLL50CA/p1706800182262059?thread_ts=1705426484.605769&cid=C7LLL50CA

Users can get an error that tells them that the tenant name is wrong, but it's difficult to track down _what_ is causing this issue. Adding the URL will let us go track down which item is associated with a different tennant. 

#### Pre-Review Checklist


timgrein (approved):
Nice improvement!

PR #2114, opened 2024-02-02
Improve debug log for next sync message

artem-shelkovnikov:
Minor improvement: 

Log message that says when next sync happens was missing the sync type.

Before:

```
[FMWK][17:03:59][DEBUG] [Connector id: network_drive, index name: search-network_drive] Next sync is at 2024-02-02 16:04:01.648207
```

After:

```
[FMWK][17:03:59][DEBUG] [Connector id: network_drive, index name: search-network_drive] Next 'full' sync is at 2024-02-02 16:04:01.648207
```


jedrazb (approved):
🚢

PR #2115, opened 2024-02-02
Use API keys for native connector syncs

navarone-feekery:
## Summary

In 8.13 we will be using API keys for native connector content ingestion. This is applicable to both content and ACL syncs.
API keys are generated for native connectors upon creation ([Kibana changes here](https://github.com/elastic/kibana/pull/176046)). The API keys are stored in secrets storage, with a reference in the connector document under `api_key_secret_id`.

When performing a sync, a native connector will now fetch its API key from secrets storage and use that as the authorization header for requests. This is only for the duration of that sync. The connector will fetch the API key each time it syncs. This is so that if the user changes the API key, they don't need to restart everything to have it applied.

These changes do not impact connector clients, which will continue to use the authorization set in the `confg.yml` file.

Lastly, this is behind a feature flag called `native_connector_api_keys`. This is `False` by default for all connectors for now. I intend for this to be `True` when we release, but these changes will likely impact others' development (and also I still need to confirm they will work on cloud), so for now it is `False`.

### Changes

- Add method in `ESManagementClient` for fetching connector secrets
- Change authorization for native connector syncs to API key found in connector secrets doc
- Add feature flag `native_connector_api_keys` with default value `False`
- Add unit tests

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -467,6 +467,8 @@ class Features:
     BASIC_RULES_OLD = "basic_rules_old"
     ADVANCED_RULES_OLD = "advanced_rules_old"
 
+    NATIVE_CONNECTOR_API_KEYS = "native_connector_api_keys"
+
     def __init__(self, features=None):
         if features is None:
             features = {}
@@ -481,6 +483,11 @@ class Features:
     def document_level_security_enabled(self):
         return self._nested_feature_enabled(
             ["document_level_security", "enabled"], default=False
+        )
+
+    def native_connector_api_keys_enabled(self):
+        return self._nested_feature_enabled(
+            ["native_connector_api_keys", "enabled"], default=False
         )
 
     def sync_rules_enabled(self):
@@ -623,6 +630,10 @@ class Connector(ESDocument):
     @property
     def sync_cursor(self):
         return self.get("sync_cursor")
+
+    @property
+    def api_key_secret_id(self):
+        return self.get("api_key_secret_id")
 
     async def heartbeat(self, interval):
         if (
--- a/connectors/es/management_client.py
+++ b/connectors/es/management_client.py
@@ -183,3 +183,13 @@ class ESManagementClient(ESClient):
             timestamp = source.get(TIMESTAMP_FIELD)
 
             yield doc_id, timestamp
+
+    async def get_connector_secret(self, connector_secret_id):
+        secret = await self._retrier.execute_with_retry(
+            partial(
+                self.client.perform_request,
+                "GET",
+                f"/_connector/_secret/{connector_secret_id}",
+            )
+        )
+        return secret.get("value")
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -48,6 +48,9 @@ from connectors.utils import (
     iso_utc,
     retryable,
 )
+from elasticsearch import (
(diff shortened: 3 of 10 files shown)


connectors/protocol/connectors.py
```diff
@@ -483,6 +485,11 @@ def document_level_security_enabled(self):
             ["document_level_security", "enabled"], default=False
         )
 
+    def native_connector_api_keys_enabled(self):
```
vidok, line 488: Why is this a feature toggle? Should this be just enabled for all native connectors?

navarone-feekery: > Should this be just enabled for all native connectors?

Yes it should be, I've set it to `False` because these changes are pretty breaking until the other components (ES and Kibana) are ready as well. I intend to change the default to `True`.

> Why is this a feature toggle?

I thought it would be good if we had a way to easily rollback for one release. Perhaps I'm being overly cautious? WDYT?

vidok:
I have 2 questions/concerns: 
* Connectors CLI [doesn't create API](https://github.com/elastic/connectors/blob/main/connectors/cli/connector.py#L109-L117) keys for native connectors. I assume this needs to be aligned with Kibana's changes. 
* When customers use Elasticsearch API keys they can't generate new API keys. If a customer uses Connectors API with ES API key auth they won't be able to create new native connectors. Do you have a plan to address this issue?

vidok:
> I intend to update CLI as well, I have an issue to track it https://github.com/elastic/enterprise-search-team/issues/6419. It'll be part of the 8.13 release.

Thank you. 

> For native connectors we could possibly have an API key generated during the preflight check if the connector is missing an API key. The native connectors authenticate themselves using the Enterprise Search user rather than an API key. Then we can continue the sync with the new API key. Does that sound feasible?

The connector service can't generate API keys when an ES API key is used for the authorization. So I think it's not feasible.

navarone-feekery:
> The connector service can't generate API keys when an ES API key is used for the authorization. So I think it's not feasible.

Native connectors authenticate the using Enterprise Search system role. The changes in this PR only impact syncs, so the Enterprise Search system role auth is still used for connector management. I think it's possible to dynamically generate an API key if one is missing if it's a native connector.

In any case I'll try it first and include it in this PR.

connectors/protocol/connectors.py
vidok: > Yes it should be, I've set it to False because these changes are pretty breaking until the other components (ES and Kibana) are ready as well. I intend to change the default to True.

Yeah, I think it's a good idea. However, we should remove it once all the components are ready. Perhaps the next release?

My main concern is that the customers may get confused to see this feature in their connectors configuration. I think I'm being overly cautious too :). 

Let's keep this as it is 👍🏻

vidok:
> Native connectors authenticate the using Enterprise Search system role. The changes in this PR only impact syncs, so the Enterprise Search system role auth is still used for connector management. I think it's possible to dynamically generate an API key if one is missing if it's a native connector.
> 
> In any case I'll try it first and include it in this PR.

Get it. You are right. Then we are good to move forward 👍🏻

navarone-feekery:
@vidok I've confirmed that I can do this locally, but I'm worried the PR will get too big. If the changes here are okay, do you mind if I address the API key generation in a follow-up PR?

vidok (approved):

PR #2117, opened 2024-02-05
[Framework] Allow bulk operations result logging

timgrein:
We've an open tech improvement, which states that it's important for some customers for audit purposes to be able to log ids of documents, which were deleted. 

This PR introduces one new configuration setting `elasticsearch.bulk.enable_operations_logging` to enable this option. The logs will be logged on `DEBUG` level.

Sample logs for successful operations:

```
[...]
[FMWK][12:02:59][DEBUG] [Connector id: PHx4ZY4B3uUGqDXsM_KG, index name: mysql-connector, Sync job id: QnzTZY4B3uUGqDXs5PJS] Successfully executed 'index' on document with id 'employees_1702'. Result: created
[FMWK][12:02:59][DEBUG] [Connector id: PHx4ZY4B3uUGqDXsM_KG, index name: mysql-connector, Sync job id: QnzTZY4B3uUGqDXs5PJS] Successfully executed 'delete' on document with id 'customers_112'. Result: deleted
[...]
```

Sample log for failed operations:
```
[FMWK][12:02:59][DEBUG] [Connector id: PHx4ZY4B3uUGqDXsM_KG, index name: mysql-connector, Sync job id: QnzTZY4B3uUGqDXs5PJS] Failed to execute 'delete' on document with id '123'. Result: not found. Error: Doc with id 123 was not found

```

#### Pre-Review Checklist
  - Not yet, will open a separate PR

~- [ ] if you added or changed Rich Configurable Fields for a Native Connector, you made a corresponding PR in [Kibana](https://github.com/elastic/kibana/blob/main/packages/kbn-search-connectors/types/native_connectors.ts)~

## Release Note

Add configuration settings to be able to log ids of deleted documents.


connectors/es/sink.py
```diff
@@ -129,10 +133,22 @@ def _bulk_op(self, doc, operation=OP_INDEX):
...
 
+    def _log_deleted_doc_id_if_enabled(self, doc_id):
+        if self._log_deleted_doc_ids:
+            msg = f"Deleted document with id '{doc_id}'"
+
+            if self._log_deleted_doc_ids_log_level == "INFO":
+                self._logger.info(msg)
+
+            elif self._log_deleted_doc_ids_log_level == "DEBUG":
+                self._logger.debug(msg)
```
seanstory, line 150: Looks like the only options are `INFO` and `DEBUG`. Should probably state that in the comment in `config.yml` and raise an error or default to something if the config has a value other than one of these.

seanstory (commented):
One nit, but largely LGTM.

connectors/es/sink.py
artem-shelkovnikov: I'm not sure if change is needed ATM - it was an ask of a customer at some point and they found a way around it.

If it's needed, we can probably approach it a bit more generically by logging added/updated/deleted document ids in DEBUG log level.

But again, I'm not sure if it's needed any more.

timgrein: > Looks like the only options are INFO and DEBUG. Should probably state that in the comment in config.yml and raise an error or default to something if the config has a value other than one of these.

Good point, will address it!

> But again, I'm not sure if it's needed any more.
Checking it with product

artem-shelkovnikov (commented):
Is it possible to do it not here, but after bulk call was made?

It should return stats of bulk request and will give you much more precise stats (e.g. when you upsert it can tell you whether it created or updated the record)

connectors/es/sink.py
```diff
@@ -171,6 +180,43 @@ async def _bulk_api_call():
...
+                self._logger.error(
+                    f"Unknown action item returned from _bulk API for item {item}"
+                )
+                return
+
+            doc_id = item[action_item].get("_id")
+            if doc_id is None:
+                # Should only happen, if the _bulk API changes
+                # Unlikely, but as this functionality could be used for audits we want to detect changes fast
+                self._logger.error(f"Could not retrieve '_id' for item {item}")
```
vidok, line 205: Should this also need to be "document" instead of "item" in the log string? 
```suggestion
                self._logger.error(f"Could not retrieve '_id' for document {item}")
```

connectors/es/sink.py
```diff
@@ -171,6 +180,43 @@ async def _bulk_api_call():
...
+                return
+
+            result = item[action_item].get("result")
+            operation_failed = result not in SUCCESSFUL_RESULTS
```
wangch079, line 209: According to the [doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-response-body):

>result
>(string) Result of the operation. Successful values are created, deleted, and updated.
>
>This parameter is only returned for successful operations.

https://github.com/elastic/connectors/blob/f4d65e252bae32d228f0e776e517977f5d2ba5ab/connectors/es/sink.py#L176-L178

connectors/es/sink.py
```diff
@@ -171,6 +180,43 @@ async def _bulk_api_call():
...
+
+            if operation_failed:
```
wangch079, line 211: For failed operation, you should check 
```python
if "error" in item[action_item]
```

config.yml
```diff
@@ -195,6 +195,10 @@
 ##  Connector service log level.
 #service.log_level: INFO
 #
+##  Enable to log ids of created/indexed/deleted/updated documents during a sync.
+##    This will be logged on 'DEBUG' log level.
+#service.enable_doc_id_trace_logging: false
```
wangch079, line 200: When I first looked at the PR subject, I thought you would log in at the trace level. I think we can simplify this config to something like `enable_operation_logging`.

Another point, do you think this should be under `elasticsearch.bulk`? This config applies to the Bulk API request, and we have a similar config:
https://github.com/elastic/connectors/blob/7081964fd567f968d559e708836db7081c2f4b83/config.yml#L99-L100

connectors/es/sink.py
timgrein: Nice spot, thanks!

connectors/es/sink.py
timgrein: I think there's a difference between a successful operation and a successful value. A successful operation could also be a operation, where the result is `not_found` (which is a non-successful value), which we might look at as the operation failed as we didn't find an id the connector expected to be there for some reason. WDYT?

connectors/es/sink.py
timgrein: See comment above. Maybe it makes sense to name the condition `non_successful_result`?

config.yml
timgrein: Fair point, I like the idea of putting it under `elasticsearch.bulk`. Should we use plural? `enable_operations_logging`?

config.yml
```diff
@@ -147,6 +147,10 @@
 #elasticsearch.bulk.retry_interval: 10
 #
 #
+##  Enable to log ids of created/indexed/deleted/updated documents during a sync.
```
seanstory, line 150: help me make sure that this isn't lost with https://github.com/elastic/connectors/pull/2280 (depending on which merges first)

connectors/es/sink.py
```diff
@@ -170,6 +179,48 @@ async def _bulk_api_call():
...
+                action_item = "create"
+            elif "update" in item:
+                action_item = "update"
+            else:
+                # Should only happen, if the _bulk API changes
+                # Unlikely, but as this functionality could be used for audits we want to detect changes fast
+                self._logger.error(
+                    f"Unknown action item returned from _bulk API for item {item}"
+                )
+                return
```
seanstory, line 198: should this `continue` instead?

connectors/es/sink.py
```diff
@@ -170,6 +179,48 @@ async def _bulk_api_call():
...
+
+            doc_id = item[action_item].get("_id")
+            if doc_id is None:
+                # Should only happen, if the _bulk API changes
+                # Unlikely, but as this functionality could be used for audits we want to detect changes fast
+                self._logger.error(f"Could not retrieve '_id' for document {item}")
+                return
```
seanstory, line 205: edge case worth validating - a pipeline removes `_id` and then a drop processor drops the doc, so it isn't assigned a random ID by Elasticsearch. IDK what the response for that record would look like.

connectors/es/sink.py
```diff
@@ -170,6 +179,48 @@ async def _bulk_api_call():
...
+            successful_result = result in SUCCESSFUL_RESULTS
+
+            if not successful_result:
+                if "error" in item[action_item]:
+                    self._logger.debug(
+                        f"Failed to execute '{action_item}' on document with id '{doc_id}'. Error: {item[action_item].get('error')}"
+                    )
+                else:
+                    self._logger.debug(
+                        f"Successfully executed '{action_item}' on document with id '{doc_id}', but got non-successful result: {result}"
```
seanstory, line 217:
```suggestion
                        f"executed '{action_item}' on document with id '{doc_id}', but got non-successful result: {result}"
```

seanstory (commented):
few nits. But I like it!

connectors/es/sink.py
timgrein: Oh yes of course, good catch! Adjusted with [153460d](https://github.com/elastic/connectors/pull/2117/commits/153460d660bd13673edc6aef960652bb19a725af)

config.yml
seanstory: mine merged first. To resolve the conflict, this'll just need to be moved to `config.yml.example`

artem-shelkovnikov (approved):
Good stuff!

PR #2119, opened 2024-02-06
Revert "Specify port when calling all smbclient methods in network drive connector (#1898)"

artem-shelkovnikov:
This reverts commit 463689bfed5572302b876bc5ab48d93de35c19ef.

As reported by some customers, this change makes connector not usable against certain versions of SMB protocol, specifically ones running on Windows Server.

## Release Note

Fix Network Drive connector unable to connect to certain versions of SMB server running on Windows

### DIFF
--- a/connectors/sources/network_drive.py
+++ b/connectors/sources/network_drive.py
@@ -343,9 +343,7 @@ class NASDataSource(BaseDataSource):
 
     @cached_property
     def get_directory_details(self):
-        return list(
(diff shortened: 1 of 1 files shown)


seanstory:
CC @acrewdson 

We probably need a medium between having this change and reverting this change. If we just revert it, we end up in the same problem space that Andrew was trying to fix with https://github.com/elastic/connectors/pull/1898

acrewdson (approved):
thank you!

PR #2123, opened 2024-02-07
Code for Redis SSL

akanshi-elastic:
Part of https://github.com/elastic/connectors/issues/1941

#### Pre-Review Checklist


artem-shelkovnikov (commented):
Does redis support connection strings?

If so, will we have conflicts if connection string specifies SSL as true and RCF specifies as false?

tests/sources/test_redis.py
```diff
@@ -335,3 +335,34 @@ async def test_advanced_rules_validation(advanced_rules, expected_validation_res
             advanced_rules
         )
         assert validation_result == expected_validation_result
+
+
+@pytest.mark.asyncio
+async def test_ping_when_mutual_ssl_enabled():
```
wangch079, line 341: the test doesn't `ping`

tests/sources/test_redis.py
```diff
@@ -335,3 +335,34 @@ async def test_advanced_rules_validation(advanced_rules, expected_validation_res
...
+        source.client.tls_certfile = (
+            "-----BEGIN CERTIFICATE----- Invalid-Certificate -----END CERTIFICATE-----"
+        )
+        source.client.tls_keyfile = "-----BEGIN RSA PRIVATE KEY----- Invalid-Certificate -----END RSA PRIVATE KEY-----"
+        mocked_client = AsyncMock()
+        with mock.patch(
+            "redis.from_url",
+            return_value=mocked_client,
+        ):
+            await source.client._client
```
wangch079, line 355: Probably you want to assert 
```python
mocked_client.assert_called_once_with(<connection_string>, decode_responses=True)
```

BTW, you don't use `await` here as `_client` is not a coroutine

tests/sources/test_redis.py
```diff
@@ -335,3 +335,34 @@ async def test_advanced_rules_validation(advanced_rules, expected_validation_res
...
+async def test_client_when_ssl_enabled():
+    async with create_redis_source() as source:
+        source.client.database = ["*"]
+        source.client.tls_enabled = True
+        mocked_client = Mock()
+        with mock.patch(
+            "redis.from_url",
+            return_value=mocked_client,
+        ):
+            await source.client._client
```
wangch079, line 368: Same comment as above.

tests/sources/test_redis.py
akanshi-elastic: mocked_client.assert_called_once_with(<connection_string>, decode_responses=True)
This line will be called 0 times which fails the test case. instead of that I added `hasaatr` assert for testing.

Screenshot 2024-02-08 144800

wangch079: My bad, I think I gave the wrong suggestion.

You should assert `redis.from_url` is called with the expected arguments.

connectors/sources/redis.py
```diff
@@ -249,6 +295,39 @@ def get_default_configuration(cls):
...
+                "required": False,
+                "tooltip": "Specifies the client certificate from the Certificate Authority. The value of the certificate is used to validate the certificate presented by the Redis instance.",
+                "type": "str",
+            },
+            "tls_keyfile": {
+                "depends_on": [{"field": "mutual_tls_enabled", "value": True}],
+                "label": "client private key file for SSL/TLS",
+                "order": 9,
+                "required": False,
+                "tooltip": "Specifies the client private key from the Certificate Authority. The value of the key is used to validate the connection in Redis instance",
```
timgrein, line 328:
```suggestion
                "tooltip": "Specifies the client private key from the Certificate Authority. The value of the key is used to validate the connection in the Redis instance.",
```

tests/sources/test_redis.py
```diff
@@ -0,0 +1,370 @@
...
+            "-----BEGIN CERTIFICATE----- Invalid-Certificate -----END CERTIFICATE-----"
+        )
+        source.client.tls_keyfile = "-----BEGIN RSA PRIVATE KEY----- Invalid-Certificate -----END RSA PRIVATE KEY-----"
+        mocked_client = AsyncMock()
+        with mock.patch(
+            "redis.from_url",
+            return_value=mocked_client,
+        ):
+            redis_client = source.client._client
+            assert hasattr(redis_client, "ping")
```
wangch079, line 356: Will this pass? `source.client._client` (i.e. an `AsyncMock` instance) was assigned to `redis_client`, and I don't think it has the attribute `ping`.

wangch079: The expectation of this test case (when mutual ssl is enabled) is that:
1. You mock `redis.from_url`
2. You have the source configured with `mutual_tls_enabled = Ture`
3. The mocked `reids.from_url` should be called with the expected url, i.e. rediss + query

Same for the test case when ssl is enabled, it should expect a call with rediss

connectors/sources/redis.py
```diff
@@ -0,0 +1,432 @@
...
+
+    async def close(self):
+        if self._redis_client:
+            await self._client.aclose()  # pyright: ignore
+        self.remove_temp_files()
+
+    async def validate_database(self, db):
+        try:
+            await self._client.execute_command("SELECT", db)
+            await self._client.ping()
```
timgrein, line 110: Should this be the other way around? You first ping and then execute a dummy select?

tests/sources/test_redis.py
```diff
@@ -0,0 +1,370 @@
...
+        with mock.patch("redis.from_url", return_value=mocked_client):
+            mocked_client.ping = AsyncMock(
+                side_effect=redis.exceptions.AuthenticationError
+            )
+            with pytest.raises(Exception):
+                await source.ping()
+
+
+@pytest.mark.asyncio
+async def test_validate_config_when_database_type():
```
timgrein, line 118: This method name should probably have something after `database_type` to indicate what was invalid, right? Also why does it expect `ConfigurableFieldValueError`, if `validate_database` returns `True`? Maybe a comment would help explaining the logic behind this test

tests/sources/test_redis.py
```diff
@@ -0,0 +1,370 @@
...
+            with pytest.raises(ConfigurableFieldValueError):
+                await source.validate_config()
+
+
+@pytest.mark.asyncio
+async def test_validate_config_when_database_is_invalid():
+    async with create_redis_source() as source:
+        source.client.database = ["123"]
+        mocked_client = Mock()
+        mocked_client.validate_database = AsyncMock(return_value=True)
```
timgrein, line 133: Same here, I'm a bit confused, why we return `True`, which indicates a valid database, but the test states that the database is invalid? Could you elaborate here?

tests/sources/test_redis.py
```diff
@@ -0,0 +1,370 @@
...
+from connectors.sources.redis import (
+    RedisAdvancedRulesValidator,
+    RedisDataSource,
+)
+from tests.commons import AsyncIterator
+from tests.sources.support import create_source
+
+ADVANCED_SNIPPET = "advanced_snippet"
+
+DOCUMENT = [
```
timgrein, line 27:
```suggestion
DOCUMENTS = [
```

As this is an array I would use plural here

tests/sources/test_redis.py
```diff
@@ -0,0 +1,370 @@
...
+@pytest.mark.asyncio
+async def test_get_databases_for_multiple_db():
+    async with create_redis_source() as source:
+        source.client.database = [1, 2]
+        async for database in source.client.get_databases():
+            assert database in [1, 2]
+
+
+@pytest.mark.asyncio
+async def test_get_databases_with_astric():
```
timgrein, line 163:
```suggestion
async def test_get_databases_with_asterisk():
```

tests/sources/test_redis.py
```diff
@@ -0,0 +1,370 @@
...
+    async with create_redis_source() as source:
+        source.client.database = ["*"]
+        source.client._client = RedisClientMock()
+        async for database in source.client.get_databases():
+            assert database == 0
+
+
+@pytest.mark.asyncio
+async def test_get_databases_negative():
```
timgrein, line 172:
```suggestion
async def test_get_databases_expect_no_databases_on_auth_error():
```

timgrein (commented):
Left some additional comments around testing

connectors/sources/redis.py
akanshi-elastic: Actually this line is not making much sense here: ` await self._client.ping()` because if database is not valid then line `109` will throw exception else not.. so this line can be removed.

tests/sources/test_redis.py
akanshi-elastic: mocked_client.validate_database = AsyncMock(return_value=True)
This line is not mocking `validate_database` correctly. let me fix this

tests/sources/test_redis.py
wangch079: The patch of `redis.from_url` won't work as you used the alias here:
https://github.com/elastic/connectors/blob/814678802b88c70ca5a4c5e47dc9c2e931d7a056/connectors/sources/redis.py#L11


Try below test:
```python
@pytest.mark.asyncio
async def test_client_when_mutual_ssl_enabled():
    async with create_redis_source() as source:
        source.client.database = ["*"]
        source.client.ssl_enabled = True
        source.client.mutual_tls_enabled = True
        source.client.store_ssl_key = Mock(return_value = "/tmp/tmp4vn0jxy7.crt")
        with mock.patch(
                "redis.asyncio.from_url",
                return_value=AsyncMock(),
        ) as from_url_mock:
            redis_client = source.client._client
            connection_string = "rediss://username:password@localhost:6379?ssl_certfile=/tmp/tmp4vn0jxy7.crt&ssl_keyfile=/tmp/tmp4vn0jxy7.crt"
            from_url_mock.assert_called_with(connection_string, decode_responses=True)
```

tests/sources/test_redis.py
```diff
@@ -335,3 +331,47 @@ async def test_advanced_rules_validation(advanced_rules, expected_validation_res
...
+    async with create_redis_source() as source:
+        source.client.database = ["*"]
+        source.client.ssl_enabled = True
+        source.client.mutual_tls_enabled = True
+        with mock.patch(
+            "redis.asyncio.from_url",
+            return_value=AsyncMock(),
+        ) as from_url_mock:
+            await source.client.ping()
+            from_url_mock.assert_called_once()
```
wangch079, line 377: Not quite sure what it's trying to test?

`source.client.ping()` will delegate to `self._redis_client`, which is an `AsyncMock` instance. you probably should assert the `_redis_client` is called/awaited once.

tests/sources/test_redis.py
```diff
@@ -129,9 +126,8 @@ async def test_validate_config_when_database_type():
 async def test_validate_config_when_database_is_invalid():
     async with create_redis_source() as source:
         source.client.database = ["123"]
-        mocked_client = Mock()
-        mocked_client.validate_database = AsyncMock(return_value=True)
-        with mock.patch("redis.from_url", return_value=mocked_client):
+        source.client.validate_database = AsyncMock(return_value=False)
+        with mock.patch("redis.from_url", return_value=Mock()):
```
wangch079, line 130: This patch doesn't work. You should replace all occurrences of `redis.from_url` with `redis.asyncio.from_url` in the test.

tests/sources/test_redis.py
akanshi-elastic: I just want to test ping here when_mutual_ssl_enabled without mocking `store_ssl_key` so that these lines also get cover with the tests.

wangch079 (approved):

PR #2126, opened 2024-02-07
POC: Role creation for access control syncs

timgrein:
This POC demonstrates how the connectors framework can create or update roles instead of ingesting access control documents.

The POC is hardcoded to create roles for an index called `search-spo` with `all` privileges. Before creating roles you need to update the API key for the connector client to include either `manage_security` or `all` cluster privileges (you can do that via Kibana under `stack management` -> `Security` -> `Roles`).

By calling `GET /_security/role` in the dev tools you can verify that the roles were created successfully.

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -569,40 +569,29 @@ class Extractor:
                 self._log_progress()
 
             doc_id = doc["id"] = doc.pop("_id")
-            doc_exists = doc_id in existing_ids
 
-            if doc_exists:
-                last_update_timestamp = existing_ids.pop(doc_id)
-                doc_not_updated = (
-                    TIMESTAMP_FIELD in doc
-                    and last_update_timestamp == doc[TIMESTAMP_FIELD]
-                )
+            self.total_docs_created += 1
 
-                if doc_not_updated:
-                    continue
+            if TIMESTAMP_FIELD not in doc:
+                doc[TIMESTAMP_FIELD] = iso_utc()
 
-                self.total_docs_updated += 1
+            self._logger.debug(f"Creating or updating role with id '{doc_id}'")
+            role_creation_or_update_result = await self.client.client.security.put_role(
+                name=doc_id,
+                cluster=["all"],
+                indices=[
+                    {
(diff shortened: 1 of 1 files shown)


connectors/es/sink.py
```diff
@@ -569,40 +569,29 @@ async def get_access_control_docs(self, generator):
...
 
-                operation = OP_INDEX
+            self._logger.debug(f"Creating or updating role with id '{doc_id}'")
+            role_creation_or_update_result = await self.client.client.security.put_role(
+                name=doc_id,
+                cluster=["all"],
+                indices=[
+                    {
+                        "names": ["search-spo"],
+                        "privileges": ["all"],
```
seanstory, line 585: nit: `read` would be what we'd actually want, I think

connectors/es/sink.py
```diff
@@ -569,40 +569,29 @@ async def get_access_control_docs(self, generator):
...
+                doc[TIMESTAMP_FIELD] = iso_utc()
 
-                operation = OP_INDEX
+            self._logger.debug(f"Creating or updating role with id '{doc_id}'")
+            role_creation_or_update_result = await self.client.client.security.put_role(
+                name=doc_id,
+                cluster=["all"],
+                indices=[
+                    {
+                        "names": ["search-spo"],
```
seanstory, line 584: Shouldn't we be able to get the actual index name from `self.index`? This name should be the document index for our full/incremental syncs, so it should be accessible from the connector record's `index_name` if nothing else.

connectors/es/sink.py
```diff
@@ -569,40 +569,29 @@ async def get_access_control_docs(self, generator):
                 self._log_progress()
 
             doc_id = doc["id"] = doc.pop("_id")
-            doc_exists = doc_id in existing_ids
```
seanstory, line 572: This is where it would be good if our List Roles API could find roles specific to this connector. 🤔

seanstory:
@timgrein can you clean this up or close it?

timgrein:
Can be closed as this was only a POC

PR #2132, opened 2024-02-08
Add volumes to Salesforce Fixture

moxarth-elastic:
## Part of https://github.com/elastic/connectors/issues/1859

This PR includes changes in the Salesforce fixture to ingest large amount of data depending on `DATA_SIZE` argument.

Perf8 Report: https://drive.google.com/file/d/1L0TUpjiDyvYc3X1TFKzXC0j6NL1rv8ql/view?usp=drive_link

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -405,15 +405,22 @@ class SalesforceClient:
         async for records in self._yield_non_bulk_query_pages(query):
             case_feeds_by_case_id = {}
             if await self._is_queryable("CaseFeed") and records:
-                case_ids = [x.get("Id") for x in records]
-                case_feeds = await self.get_case_feeds(case_ids)
+                all_case_ids = [x.get("Id") for x in records]
+                case_ids_list = [
+                    all_case_ids[i : i + 800] for i in range(0, len(all_case_ids), 800)
+                ]
+
+                all_case_feeds = []
+                for case_ids in case_ids_list:
+                    case_feeds = await self.get_case_feeds(case_ids)
+                    all_case_feeds.extend(case_feeds)
 
                 # groupby requires pre-sorting apparently
-                case_feeds.sort(key=lambda x: x.get("ParentId", ""))
+                all_case_feeds.sort(key=lambda x: x.get("ParentId", ""))
                 case_feeds_by_case_id = {
                     k: list(feeds)
                     for k, feeds in groupby(
-                        case_feeds, key=lambda x: x.get("ParentId", "")
+                        all_case_feeds, key=lambda x: x.get("ParentId", "")
                     )
                 }
 
--- a/tests/sources/fixtures/salesforce/fixture.py
+++ b/tests/sources/fixtures/salesforce/fixture.py
@@ -22,11 +22,11 @@ DATA_SIZE = os.environ.get("DATA_SIZE", "medium").lower()
 
 match DATA_SIZE:
     case "small":
-        RECORD_COUNT = 5000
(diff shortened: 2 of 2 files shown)


connectors/sources/salesforce.py
```diff
@@ -405,15 +405,22 @@ async def get_cases(self):
         async for records in self._yield_non_bulk_query_pages(query):
             case_feeds_by_case_id = {}
             if await self._is_queryable("CaseFeed") and records:
-                case_ids = [x.get("Id") for x in records]
-                case_feeds = await self.get_case_feeds(case_ids)
+                all_case_ids = [x.get("Id") for x in records]
+                case_ids_list = [
+                    all_case_ids[i : i + 800] for i in range(0, len(all_case_ids), 800)
+                ]
```
navarone-feekery, line 411: Question: where does 800 come from? Is it the upper limit for query size?

tests/sources/fixtures/salesforce/fixture.py
```diff
@@ -170,7 +163,7 @@ def token():
 def query(version):
     query = request.args.get("q")
     table_name = re.findall(r"\bFROM\s+(\w+)", query)[-1]
-    done = random.choice([True, False])
+    done = False
 
     response = {"done": done, "records": generate_records(table_name)}
```
navarone-feekery, line 168: Does this handle pagination or do we just ignore pagination for ftests? Asking because the final response should be `done: True`, but I can see we handle pagination now by the presence of `nextRecordsUrl` in the response. 

Also nit: there's no point declaring `done` anymore if it's always `False`. Might as well just have
```
response = {"done": False, "records": generate_records(table_name)}
```

tests/sources/fixtures/salesforce/fixture.py
```diff
@@ -203,7 +196,7 @@ def describe_sobject(_version, _sobject):
 
 @app.route("/sfc/servlet.shepherd/version/download/<_download_id>", methods=["GET"])
 def download(_download_id):
-    return io.BytesIO(bytes(fake_provider.get_html(), encoding="utf-8"))
+    return io.BytesIO(bytes(fake_provider.get_text(), encoding="utf-8"))
```
navarone-feekery, line 199: Question: why the switch to `get_text`?

connectors/sources/salesforce.py
moxarth-elastic: `IN` clause in SOQL query can't exceed more than 1000 ids or 20k characters whatever is first met. So, for the safer side i kept 800. while testing the E2E, i faced an error so i've changed this.

tests/sources/fixtures/salesforce/fixture.py
moxarth-elastic: This handles pagination for each objects and the final response will surely have `done: True` to terminate the pagination. Updated the code as well 👍

tests/sources/fixtures/salesforce/fixture.py
moxarth-elastic: There wasn't any reason for that, i just updated it to add text instead of html content for checking. Do you want me to revert it?

artem-shelkovnikov: Yeah please keep `get_html` so that we're also testing text extraction in functional tests

moxarth-elastic:
buildkite test this

moxarth-elastic:
@artem-shelkovnikov @navarone-feekery can you please verify the changes i made as per your suggestions?

artem-shelkovnikov (approved):
Looks good from my perspective. Can you share Perf8 report for this PR too?

PR #2134, opened 2024-02-08
Add GraphQL configurations and indexing Implementation 

parthpuri-elastic:
## Part of https://github.com/elastic/connectors/issues/2068

This PR adds Indexing functionality to GraphQL connector.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/graphql.py
+++ b/connectors/sources/graphql.py
@@ -123,27 +123,27 @@ class GraphQLClient:
             dictionary/list: Documents from the response
         """
         for keys in self.graphql_object_list:
-            duplicate_data = data
+            current_level_data = data
             key_list = keys.split(".")
 
             for key in key_list:
-                if isinstance(duplicate_data, dict):
-                    if key not in duplicate_data.keys():
+                if isinstance(current_level_data, dict):
+                    if key not in current_level_data.keys():
                         msg = (
                             f"{key} not found in the response. Please check the query."
                         )
                         raise ConfigurableFieldValueError(msg)
-                    duplicate_data = duplicate_data.get(key, {})
+                    current_level_data = current_level_data.get(key, {})
                 else:
                     msg = "Found list while processing GraphQL Objects List. Please check the query and provide path upto dictionary."
                     raise ConfigurableFieldValueError(msg)
 
-            if isinstance(duplicate_data, dict):
-                if duplicate_data:
-                    yield duplicate_data
+            if isinstance(current_level_data, dict):
+                if current_level_data:
+                    yield current_level_data
 
-            if isinstance(duplicate_data, list):
-                for doc in duplicate_data:
+            if isinstance(current_level_data, list):
+                for doc in current_level_data:
                     yield doc
 
     def extract_pagination_info(self, data):
--- a/connectors/sources/graphql.py
+++ b/connectors/sources/graphql.py
@@ -179,7 +179,7 @@ class GraphQLClient:
             msg = "Pagination is enabled but the query is missing 'pageInfo'. Please include 'pageInfo { hasNextPage endCursor }' in the query to support pagination."
             raise ConfigurableFieldValueError(msg)
 
-    def validate_query(self, graphql_query, visitor):
+    def validate_paginated_query(self, graphql_query, visitor):
         graphql_object = self.pagination_key.split(".")[-1]
         self._logger.debug(f"Finding pageInfo field in {graphql_object}.")
         if not (
@@ -195,7 +195,7 @@ class GraphQLClient:
             visitor = FieldVisitor()
             visit(ast, visitor)  # pyright: ignore
 
-            self.validate_query(graphql_query, visitor)
+            self.validate_paginated_query(graphql_query, visitor)
             while True:
                 self._logger.debug(
                     f"Fetching document with variables: {self.variables}."
@@ -394,7 +394,7 @@ class GraphQLDataSource(BaseDataSource):
                     {"label": "Cursor-based pagination", "value": CURSOR_PAGINATION},
                 ],
                 "order": 11,
-                "tooltip": "Cursor based pagination requires 'pageInfo' field along with argument 'after' for objects mentioned in 'GraphQL Objects List'. It also requires variable for 'after' argument.",
+                "tooltip": "For cursor-based pagination, add 'pageInfo' and an 'after' argument variable in your query at the desired node (Pagination key). Use 'after' query argument with a variable to iterate through pages. Detailed examples and setup instructions are available in the docs.",
                 "type": "str",
                 "value": NO_PAGINATION,
             },
--- a/connectors/sources/graphql.py
+++ b/connectors/sources/graphql.py
@@ -165,20 +165,26 @@ class GraphQLClient:
                 for key, documents in self.extract_graphql_data_items(
                     key="data", data=data, same_level=True
                 ):
-                    if isinstance(documents, dict) and documents.get("pageInfo"):
+                    if (
+                        isinstance(documents, dict)
+                        and documents.get("pageInfo")
+                        and {"hasNextPage", "endCursor"}.issubset(
+                            set(documents.get("pageInfo"))
+                        )
+                    ):
                         pageInfo = documents.get("pageInfo")
                         hasNextPage = pageInfo.get("hasNextPage")
                         endCursor = pageInfo.get("endCursor")
 
                         yield documents
-                        if hasNextPage:
+                        if hasNextPage and endCursor:
                             self.variables[
                                 visitor.variables_dict[key]["after"]
                             ] = endCursor
                             hasNewPage = True
 
                     else:
-                        msg = "Pagination is enabled but pageInfo field is missing please add pageInfo field."
+                        msg = "Pagination is enabled but pageInfo field is missing with hasNextPage and endCursor in the query, please add pageInfo field and add hasNextPage and endCursor inside it."
                         raise ConfigurableFieldValueError(msg)
                 if not hasNewPage:
                     return
@@ -353,7 +359,7 @@ class GraphQLDataSource(BaseDataSource):
                     {"label": "Cursor-based pagination", "value": "cursor_pagination"},
                 ],
                 "order": 11,
-                "tooltip": "Cursor based pagination requires 'pageInfo' field along with argument 'after' for objects mentioned in 'GraphQL Objects List'. It also requires 'afterCursor' variable for 'after' argument.",
+                "tooltip": "Cursor based pagination requires 'pageInfo' field along with argument 'after' for objects mentioned in 'GraphQL Objects List'. It also requires variable for 'after' argument.",
                 "type": "str",
                 "value": "none",
             },
--- a/connectors/sources/graphql.py
+++ b/connectors/sources/graphql.py
@@ -418,23 +418,37 @@ class GraphQLDataSource(BaseDataSource):
             },
         }
 
-    def is_query(self, graphql_query):
-        try:
-            ast = parse(graphql_query)  # pyright: ignore
-            for definition in ast.definitions:  # pyright: ignore
-                if (
-                    hasattr(definition, "operation")
-                    and definition.operation.value != "query"
-                ):
-                    return False
-            return True
-        except Exception as e:
-            self._logger.error(f"Failed to parse GraphQL query. Error: {e}")
-            return False
+    def is_query(self, ast):
+        for definition in ast.definitions:  # pyright: ignore
+            if (
+                hasattr(definition, "operation")
+                and definition.operation.value != "query"
+            ):
+                return False
+        return True
 
     def validate_endpoints(self):
         if re.match(URL_REGEX, self.graphql_client.url):
             return True
+        return False
+
+    def check_field_existence(self, ast, field_path):
+        def traverse(selections, path):
+            for selection in selections:
+                if selection.name.value == path[0]:
+                    if len(path) == 1:
+                        return True
+                    if selection.selection_set:
+                        return traverse(
+                            selections=selection.selection_set.selections, path=path[1:]
+                        )
+            return False
+
+        field_path = field_path.split(".")
+        for definition in ast.definitions:
+            if definition.operation.value == "query":
+                return traverse(definition.selection_set.selections, field_path)
+
         return False
 
     async def validate_config(self):
@@ -447,7 +461,13 @@ class GraphQLDataSource(BaseDataSource):
             msg = "GraphQL HTTP endpoint is not a valid URL."
             raise ConfigurableFieldValueError(msg)
 
-        if not self.is_query(graphql_query=self.graphql_client.graphql_query):
+        try:
+            ast = parse(self.graphql_client.graphql_query)  # pyright: ignore
+        except Exception as exception:
+            msg = f"Failed to parse GraphQL query. Error: {exception}"
+            raise ConfigurableFieldValueError(msg) from exception
+
+        if not self.is_query(ast=ast):
             msg = "The configured query is either invalid or a mutation which is not supported by the connector."
             raise ConfigurableFieldValueError(msg)
 
@@ -466,6 +486,20 @@ class GraphQLDataSource(BaseDataSource):
             except Exception as exception:
                 msg = f"Error while processing configured GraphQL variables. Exception: {exception}"
                 raise ConfigurableFieldValueError(msg) from exception
+
+        for graphql_object in self.graphql_client.graphql_object_list:
+            if not self.check_field_existence(ast=ast, field_path=graphql_object):
+                msg = f"{graphql_object} is not present in the query."
+                raise ConfigurableFieldValueError(msg)
+
+        if self.graphql_client.pagination_model == CURSOR_PAGINATION:
+            if not self.check_field_existence(
+                ast=ast, field_path=self.graphql_client.pagination_key
+            ):
+                msg = (
+                    f"{self.graphql_client.pagination_key} is not present in the query."
+                )
+                raise ConfigurableFieldValueError(msg)
 
     async def close(self):
         await self.graphql_client.close()
--- a/connectors/sources/graphql.py
+++ b/connectors/sources/graphql.py
@@ -28,6 +28,13 @@ from connectors.utils import (
 RETRIES = 3
 RETRY_INTERVAL = 2
 
+BASIC = "basic"
+BEARER = "bearer"
+CURSOR_PAGINATION = "cursor_pagination"
+GET = "get"
+NO_PAGINATION = "no_pagination"
+POST = "post"
+
 # Regular expression to validate the Base URL
 URL_REGEX = "^https?:\\/\\/(?:www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b(?:[-a-zA-Z0-9()@:%_\\+.~#?&\\/=]*)$"
 
@@ -84,7 +91,7 @@ class GraphQLClient:
     @cached_property
     def session(self):
         timeout = aiohttp.ClientTimeout(total=self.configuration["connection_timeout"])
-        if self.authentication_method == "bearer":
+        if self.authentication_method == BEARER:
             self.headers.update(
                 {
                     "Authorization": f"Bearer {self.configuration['token']}",
@@ -95,7 +102,7 @@ class GraphQLClient:
                 timeout=timeout,
                 raise_for_status=True,
             )
-        elif self.authentication_method == "basic":
+        elif self.authentication_method == BASIC:
             basic_auth = aiohttp.BasicAuth(
                 login=self.configuration["username"],
                 password=self.configuration["password"],
@@ -183,7 +190,7 @@ class GraphQLClient:
             raise ConfigurableFieldValueError(msg)
 
     async def paginated_call(self, graphql_query):
-        if self.pagination_model == "cursor_pagination":
+        if self.pagination_model == CURSOR_PAGINATION:
             ast = parse(graphql_query)  # pyright: ignore
             visitor = FieldVisitor()
             visit(ast, visitor)  # pyright: ignore
@@ -191,7 +198,7 @@ class GraphQLClient:
             self.validate_query(graphql_query, visitor)
             while True:
                 self._logger.debug(
-                    f"Fetching document with variables: {self.variables} and query: {graphql_query}."
+                    f"Fetching document with variables: {self.variables}."
                 )
 
                 has_new_page = False
@@ -221,7 +228,7 @@ class GraphQLClient:
     )
     async def make_request(self, graphql_query):
         try:
-            if self.http_method == "get":
+            if self.http_method == GET:
                 return await self.get(graphql_query=graphql_query)
             else:
                 return await self.post(graphql_query=graphql_query)
@@ -314,40 +321,40 @@ class GraphQLDataSource(BaseDataSource):
                 "display": "dropdown",
                 "label": "GET/POST",
                 "options": [
-                    {"label": "GET", "value": "get"},
-                    {"label": "POST", "value": "post"},
+                    {"label": "GET", "value": GET},
+                    {"label": "POST", "value": POST},
                 ],
                 "order": 2,
                 "type": "str",
-                "value": "POST",
+                "value": POST,
             },
             "authentication_method": {
                 "display": "dropdown",
                 "label": "Authentication Method",
                 "options": [
                     {"label": "No Auth", "value": "none"},
-                    {"label": "Basic Auth", "value": "basic"},
-                    {"label": "Bearer Token", "value": "bearer"},
+                    {"label": "Basic Auth", "value": BASIC},
+                    {"label": "Bearer Token", "value": BEARER},
                 ],
                 "order": 3,
                 "type": "str",
                 "value": "none",
             },
             "username": {
-                "depends_on": [{"field": "authentication_method", "value": "basic"}],
+                "depends_on": [{"field": "authentication_method", "value": BASIC}],
                 "label": "Username",
                 "order": 4,
                 "type": "str",
             },
             "password": {
-                "depends_on": [{"field": "authentication_method", "value": "basic"}],
+                "depends_on": [{"field": "authentication_method", "value": BASIC}],
                 "label": "Password",
                 "order": 5,
                 "sensitive": True,
                 "type": "str",
             },
             "token": {
-                "depends_on": [{"field": "authentication_method", "value": "bearer"}],
+                "depends_on": [{"field": "authentication_method", "value": BEARER}],
                 "label": "Bearer Token",
                 "order": 6,
                 "sensitive": True,
@@ -360,7 +367,7 @@ class GraphQLDataSource(BaseDataSource):
                 "type": "str",
             },
             "graphql_variables": {
-                "depends_on": [{"field": "http_method", "value": "post"}],
+                "depends_on": [{"field": "http_method", "value": POST}],
                 "display": "textarea",
                 "label": "Graphql Variables",
                 "order": 8,
@@ -383,17 +390,17 @@ class GraphQLDataSource(BaseDataSource):
                 "display": "dropdown",
                 "label": "Pagination model",
                 "options": [
-                    {"label": "No pagination", "value": "no_pagination"},
-                    {"label": "Cursor-based pagination", "value": "cursor_pagination"},
+                    {"label": "No pagination", "value": NO_PAGINATION},
+                    {"label": "Cursor-based pagination", "value": CURSOR_PAGINATION},
                 ],
                 "order": 11,
                 "tooltip": "Cursor based pagination requires 'pageInfo' field along with argument 'after' for objects mentioned in 'GraphQL Objects List'. It also requires variable for 'after' argument.",
                 "type": "str",
-                "value": "none",
+                "value": NO_PAGINATION,
             },
             "pagination_key": {
                 "depends_on": [
-                    {"field": "pagination_model", "value": "cursor_pagination"}
+                    {"field": "pagination_model", "value": CURSOR_PAGINATION}
                 ],
                 "label": "Pagination key",
                 "order": 12,
@@ -441,7 +448,7 @@ class GraphQLDataSource(BaseDataSource):
             raise ConfigurableFieldValueError(msg)
 
         if not self.is_query(graphql_query=self.graphql_client.graphql_query):
-            msg = "Configured Query is not supported by the connector."
+            msg = "The configured query is either invalid or a mutation which is not supported by the connector."
             raise ConfigurableFieldValueError(msg)
 
         headers = self.graphql_client.configuration["headers"]
@@ -480,7 +487,7 @@ class GraphQLDataSource(BaseDataSource):
                     yield document
 
     async def fetch_data(self, graphql_query):
-        if self.graphql_client.pagination_model == "no_pagination":
+        if self.graphql_client.pagination_model == NO_PAGINATION:
             data = await self.graphql_client.make_request(graphql_query=graphql_query)
             for documents in self.graphql_client.extract_graphql_data_items(data=data):
                 for document in self.yield_dict(documents):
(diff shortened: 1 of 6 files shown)


connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+        self.graphql_objects = self.configuration["object_list"]
+        self.http_method = self.configuration["http_method"]
+        self.variables = {}
+        self.headers = {}
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    @cached_property
+    def _get_session(self):
```
jedrazb, line 59: This is cached property rather than a method, should we rename this to e.g. `session`?

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+
+    @classmethod
+    def get_default_configuration(cls):
+        """Get the default configuration for GraphQL.
+
+        Returns:
+            dict: Default configuration.
+        """
+        return {
+            "http_endpoint": {
```
jedrazb, line 188: Perhaps we ca add some validation, e.g. regex to match that the endpoint is structured correctly. 

Are we only supporting `http` protocol, or is it possible to connect to graphql over `https`?

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+            },
+            "graphql_variables": {
+                "depends_on": [{"field": "http_method", "value": "post"}],
+                "display": "textarea",
+                "label": "Graphql Variables",
+                "order": 8,
+                "type": "str",
+                "required": False,
+            },
+            "object_list": {
```
jedrazb, line 250: consider adding tooltips with explanations for configuration fields, I'm not sure what `object_list` mean in this context (tooltips are useful in our UI and also when inspecting the config in the codebase)

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+                raise_for_status=True,
+            )
+        else:
+            return aiohttp.ClientSession(
+                headers=self.headers,
+                timeout=timeout,
+                raise_for_status=True,
+            )
+
+    def boil_down_data(self, key, data):
```
jedrazb, line 90: This function name is a bit vague, can we use sth like `extract_graphql_data_items` or something similar. Can we add some comments about the logic and why is it needed?

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+            "headers": {
+                "label": "Headers",
+                "order": 10,
+                "type": "str",
+                "required": False,
+            },
+        }
+
+    def is_query(self, graphql_query):
+        ast = parse(graphql_query)  # pyright: ignore
```
jedrazb, line 264: This checks only first operation right? could we have e.g. 
```
query GetUserInfo {
  user(id: "1") {
    id
    name
    email
  }
}

mutation UpdateUserEmail {
  updateUser(id: "1", email: "newemail@example.com") {
    id
    email
  }
}
```
That would bypass the check and execute a mutation against graphql backend?

jedrazb: Maybe sth like?
```
def is_query_only(self, graphql_query):
    try:
        ast = parse(graphql_query)
        for definition in ast.definitions:
            # Check if the operation type is 'query'; skip non-operation definitions like fragments
            if hasattr(definition, 'operation') and definition.operation != 'query':
                return False
        return True
    except GraphQLSyntaxError as e:
        # Handle syntax errors in the GraphQL query
        self._logger.error(f"Failed to parse GraphQL query. Error: {e}")
        return False
 ```

File: connectors/sources/graphql.py [hunk 6]
```diff
@@ -0,0 +1,333 @@
...
+    async def ping(self):
+        try:
+            await self.graphql_client.ping()
+            self._logger.debug("Successfully connected to GraphQL Instance.")
+        except Exception:
+            self._logger.exception("Error while connecting to GraphQL Instance.")
+            raise
+
+    async def fetch_data(self, graphql_query):
+        data = await self.graphql_client.make_request(graphql_query=graphql_query)
```
jedrazb, line 307: What would happen if our graphql backend has e.g. 100000000 docs? Would `make_request` result in a huge response containing all docs? We should paginate/scroll through the graphql db to actually make use of async programming benefits

tests/sources/test_graphql.py
```diff
@@ -0,0 +1,225 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+"""Tests the GraphQL source class methods"""
```
jedrazb, line 6: the boil_down_data has some important logic inside, can we add unit tests here as well? 

I will review tests further once we have connector logic ready :)

jedrazb (changes requested):
Good start 👍 I left a couple of comments. My biggest concern is around that (from my understanding) the current approach appears to fetch all data in a single request and then yield documents one by one. This needs to be addressed. Correct me if I misunderstood sth here

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+class GraphQLClient:
+    def __init__(self, configuration):
+        self._sleeps = CancellableSleeps()
+        self.configuration = configuration
+        self._logger = logger
+        self.graphql_query = self.configuration["graphql_query"]
+        self.url = self.configuration["http_endpoint"]
+        self.graphql_objects = self.configuration["object_list"]
```
timgrein, line 50: I would be prefer to be consistent here in the configuration names. Either `graphql_query` and `graphql_object_list` or `query` and `object_list`

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+        self.http_method = self.configuration["http_method"]
+        self.variables = {}
+        self.headers = {}
+
+    def set_logger(self, logger_):
+        self._logger = logger_
+
+    @cached_property
+    def _get_session(self):
+        timeout = aiohttp.ClientTimeout(total=300)
```
timgrein, line 60: I think it would help to make the timeout configurable under an advanced RCF

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+        if self.configuration["authentication_method"] == "bearer":
```
timgrein, line 61: I think we should store `self.configuration["authentication_method"]` as a field inside the object as it's used at multiple places. This will then also be consistent with other fields like `self.url = self.configuration["http_endpoint"]` for example.

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+        if operation_type.value == "query":
+            return True
+        return False
+
+    async def validate_config(self):
+        """Validates whether user input is empty or not for configuration fields
+        Also validate, if user configured repositories are accessible or not and scope of the token
+        """
+        await super().validate_config()
+        if self.is_query(graphql_query=self.graphql_client.graphql_query) is False:
```
timgrein, line 275:
```suggestion
        if not self.is_query(graphql_query=self.graphql_client.graphql_query):
```

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+            msg = "Mutation is not supported."
```
timgrein, line 276: I think this error message can be more precise. Mutation of what exactly?

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+            raise ConfigurableFieldValueError(msg)
+
+        headers = self.graphql_client.configuration["headers"]
+        graphql_variables = self.graphql_client.configuration["graphql_variables"]
+        if headers:
+            try:
+                self.graphql_client.headers = json.loads(headers)
+            except Exception as exception:
+                msg = f"Error while processing headers. Exception: {exception}"
```
timgrein, line 285:
```suggestion
                msg = f"Error while processing configured GraphQL headers. Exception: {exception}"
```

connectors/sources/graphql.py
```diff
@@ -0,0 +1,333 @@
...
+                raise ConfigurableFieldValueError(msg) from exception
+
+        if graphql_variables:
+            try:
+                self.graphql_client.variables = json.loads(graphql_variables)
+            except Exception as exception:
+                msg = f"Error while processing variables. Exception: {exception}"
```
timgrein, line 292:
```suggestion
                msg = f"Error while processing GraphQL variables. Exception: {exception}"
```

tests/sources/test_graphql.py
```diff
@@ -0,0 +1,225 @@
...
+                history=None,
+            )
+        )
+        with pytest.raises(UnauthorizedException):
+            await source.graphql_client.make_request("QUERY")
+
+
+@pytest.mark.asyncio
+@patch("connectors.utils.time_to_sleep_between_retries", Mock(return_value=0))
+async def test_make_request_with_429_exception():
```
timgrein, line 130: I see we've a test for `429` errors, but we don't handle it anywhere in the code explicitly right?

timgrein (commented):
Added some additional comments, I strongly support @jedrazb point around paginating results from GraphQL endpoints instead of fetching everything at once ++

connectors/sources/graphql.py
parthpuri-elastic: Sure added validation for endpoints and yes connector supports both `http` and `https`.

connectors/sources/graphql.py
parthpuri-elastic: Oh yes! updated as per suggestion

connectors/sources/graphql.py
parthpuri-elastic: In the connector user is able to execute one query only and almost every source have some GraphQL limit to return objects (for example first: 100 object). So connector only fetches objects which return by single query. So i guess pagination is not required.

connectors/sources/graphql.py
parthpuri-elastic: Done

tests/sources/test_graphql.py
parthpuri-elastic: I have added to test `clienterror` other that `401`. Regarding `429` we do not know the source so how we can handle the `429` as all source have different header name or API to fetch retry-after.

connectors/sources/graphql.py
parthpuri-elastic: Can we modify the user configuration? Suppose user configure below query:
```
repositories (first: 100, after:"endCursor"){
                  name
                  stars
}
```
So, in above query user explicitly mentioned that he/she wants 100 objects only.

Also, I guess pagination might be different in different source. So, how we can modify the query?

jedrazb: > So, in above query user explicitly mentioned that he/she wants 100 objects only.

What if they want all the objects? We need to add a support to be able to synchronize _all_ data matching a user query between the graphql instance and the index. 

> Also, I guess pagination might be different in different source. So, how we can modify the query?
What problematic case do you have in mind? We need to have a strategy for adding pagination to graphql queries, perhaps there are some open-source examples how to do it?

parthpuri-elastic: > What problematic case do you have in mind?

For example, we are working with 2 sources, `https://api.github.com/graphql` and `https://countries.trevorblades.com/graphql`. So, GitHub GraphQL supports pagination using `pageInfo` and `endCursor`. 
query: 
```
query ($login: String!, $cursor: String) {{
    user(login: $login) {
        repositories(first: 100, after: $cursor) {
        pageInfo {
            hasNextPage
            endCursor
        }
        nodes {
            id
            updatedAt
            name
            nameWithOwner
        }
        }
    }
    }
```
So above query fetches 100 repositories in each page and using `endCursor` we can move forward and fetches all the repositories.

But for `countries.trevorblades.com` we do not have `pageInfo` available and also does not have keywords like `first` , `after` etc. Using below query we can fetch all the countries using single API call. So, in generic how we can add pagination for all the sources which supports GraphQL?

```
query Countries {
    countries {
        name
    }
}
```

jedrazb: Hey @parthpuri-elastic 

> So, in generic how we can add pagination for all the sources which supports GraphQL?

I played with `countries.trevorblades.com` and yeah this doesn't support pagination at all, so in that case there is nothing we can do. 

Loooking at docs it seems that usually servers support:
- Cursor-based pagination with variables:
  - `$afterCursor`: The cursor to fetch items after. Initially, this could be null to start from the beginning.
  - `$pageSize`: The number of items to fetch.
- Offset-based pagination with variables
  - `$offset`: The offset from the beginning of the dataset to fetch items from. Initially, this could be 0.
  - `$limit`: The maximum number of items to return. 

I think we should let the user define the query with pagination variables. The connector implementation should use those variables to paginate, so:
- after each page load set `afterCursor` to `endCursor`
- after each page load set `offset` to (`offset + limit`)

What do you think about this approach? Our aim is to keep the connector implementations simple while supporting paginating over data if users wants it, what do you think? 😄

parthpuri-elastic: Sure, we can move forward with this approach for now 👍

connectors/sources/graphql.py
```diff
@@ -0,0 +1,449 @@
...
+                "tooltip": "List of object names from the query that needs to be indexed.",
+                "type": "list",
+            },
+            "headers": {
+                "label": "Headers",
+                "order": 10,
+                "type": "str",
+                "required": False,
+            },
+            "pagination_enabled": {
```
jedrazb, line 330: I think we should be explicit what pagination strategy we use? Should we instead have a dropdown with multiple values:
- no pagination
- offset-based pagination
- cursor-based pagination

Right now it's not clear what does `pagination enabled` refers to. Also the tooltip should have clear explanation about the variables that we expect in the query

connectors/sources/graphql.py
```diff
@@ -0,0 +1,449 @@
...
+            if not (
+                (
+                    {"after"}.issubset(set(visitor.fields_dict.get(graphql_object, [])))
+                    and "pageInfo" in visitor.fields_dict.keys()
+                )
+                or {"offset", "limit"}.issubset(
+                    set(visitor.fields_dict.get(graphql_object, []))
+                )
+            ):
+                msg = "Pagination is Enabled. Please add fields and variables related to pagination."
```
jedrazb, line 143: Can we be explicit about what variables we are missing? 

If user would select the pagination type (offset or cursor) in the configuration it will be easier to communicate at this step what variables are missing from the query

connectors/sources/graphql.py
```diff
@@ -0,0 +1,449 @@
...
+                raise ConfigurableFieldValueError(msg)
+
+            if "afterCursor" not in visitor.variables_dict.get(graphql_object, []):
+                msg = "Pagination is Enabled. Please set afterCursor variable for after argument."
```
jedrazb, line 147: why `Enabled` starts with upper case?

connectors/sources/graphql.py
```diff
@@ -0,0 +1,449 @@
...
+        if key in self.graphql_object_list:
+            yield data
+        if isinstance(data, dict):
+            for key, item in data.items():
+                yield from self.extract_graphql_data_items(key=key, data=item)
+        if isinstance(data, list):
+            for doc in data:
+                yield from self.extract_graphql_data_items(key=None, data=doc)
+
+    async def paginated_call(self, graphql_query):
```
jedrazb, line 128: This function needs more debug logging to troubleshoot configuration issues. We need to be able to log what pagination type is being used, what are the values of pagination variables

jedrazb (commented):
Looks better now! I left some comments. Did you manage to run the connector with pagination enabled on sufficiently big data source? I

connectors/sources/graphql.py
```diff
@@ -0,0 +1,449 @@
...
+            data = await self.graphql_client.make_request(graphql_query=graphql_query)
+            for documents in self.graphql_client.extract_graphql_data_items(
+                key="data", data=data
+            ):
+                if isinstance(documents, dict):
+                    yield documents
+                elif isinstance(documents, list):
+                    for document in documents:
+                        if isinstance(document, dict):
+                            yield document
```
jedrazb, line 430: This is duplicated code, we have same logic in 414 to 419

connectors/sources/graphql.py
```diff
@@ -0,0 +1,449 @@
...
+
+        Yields:
+            dict: Documents from GraphQL.
+        """
+        doc_id = 1
+        async for doc in self.fetch_data(
+            graphql_query=self.graphql_client.graphql_query
+        ):
+            doc = deepcopy(doc)
+            doc["_id"] = hash_id(str(doc_id))
```
jedrazb, line 446: I don't think we should autogenerate doc ids. This doesn't make much sense to me. We can easily end up in a situation that a same doc has assigned different ids in different full syncs. 

This breaks assumption in the framework that docs have unique ids that don't change. Could we perhaps utilize [graphql global object ids](https://graphql.org/learn/global-object-identification/)?

connectors/sources/graphql.py
jedrazb: Also I think we should assume reasonable defaults but allow the user to configure the pagination variables for offset and curosr based pagination. In different implementations of grpahql server they can be named slightly differently (e.g. `offset` and `limit` and `skip` and `take` (keystoneJS), also they can be custom named). Should we introduce more RFCs fields (with depend_on pagaintion type relation) that allow user to define the pagination variable names used by their framework?. It should be fine to set reasonable default values e.g. offset/limit

parthpuri-elastic: Currently, I am not able to find source with offset-based pagination. So, I have pushed only cursor-based pagination. Will add offset-based pagination in follow-up PR

connectors/sources/graphql.py
parthpuri-elastic: add this piece of code into a function

connectors/sources/graphql.py
parthpuri-elastic: After telling users to add unique ID how can we identify that unique field ? As every source can have a unique name for ID (id, uid, sys_id etc.) different from the other source, we would not be able to tell which is the unique field. Also in this connector I guess it is fine if same id assigned to different document as we are indexing all documents in every sync so previous document will get new ID.

As most of the source have unique field as `id`. So, for now i am using that id in case if it present in the doc otherwise uses custom id

connectors/sources/graphql.py
```diff
@@ -0,0 +1,449 @@
...
+                "display": "textarea",
+                "label": "Graphql Variables",
+                "order": 8,
+                "type": "str",
+                "required": False,
+            },
+            "graphql_object_list": {
+                "label": "GraphQL Objects List",
+                "order": 9,
+                "tooltip": "List of object names from the query that needs to be indexed.",
```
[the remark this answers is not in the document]
parthpuri-elastic: changed as per slack discussion

connectors/sources/graphql.py
```diff
@@ -0,0 +1,501 @@
...
+
+        Yields:
+            dict: Documents from GraphQL.
+        """
+        doc_id = 1
+        async for doc in self.fetch_data(
+            graphql_query=self.graphql_client.graphql_query
+        ):
+            doc = deepcopy(doc)
+            doc["_id"] = doc.get("id") or hash_id(str(doc_id))
```
jedrazb, line 498: I'm still not a fan of the `doc_id` approach. Given that we know exactly which objects to index (we have a path to find the key) can we somehow try to represent id as: `{path}-{hash of the dict}`?

parthpuri-elastic: I think to convert whole dict into hash is not a better approach as some documents are too large. So, as dicussed on slack i guess we can move forward with above apporach for now and if i find anything helpful i will raise follow up PR.

connectors/sources/graphql.py
```diff
@@ -0,0 +1,501 @@
...
+                "type": "str",
+                "required": False,
+            },
+            "graphql_object_list": {
+                "label": "GraphQL Objects List",
+                "order": 9,
+                "tooltip": "Specifies which GraphQL objects should be indexed as individual documents. This allows finer control over indexing, ensuring only relevant data sections from the GraphQL response are stored as separate documents. Use '.' to provide full path of the object from the starting. For example 'organization.users.nodes'",
+                "type": "list",
+            },
+            "headers": {
```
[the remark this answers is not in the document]
parthpuri-elastic: [Here](https://github.com/elastic/connectors/pull/2134/files#diff-1ac50fbc6d32ff7dfa228d41ad1b8945bd951532d30582faf148690d642ba397R447) we are assigning headers

connectors/sources/graphql.py
```diff
@@ -0,0 +1,501 @@
...
+        self.graphql_query = self.configuration["graphql_query"]
+        self.url = self.configuration["http_endpoint"]
+        self.graphql_object_list = self.configuration["graphql_object_list"]
+        self.http_method = self.configuration["http_method"]
+        self.authentication_method = self.configuration["authentication_method"]
+        self.pagination_technique = self.configuration["pagination_technique"]
+        self.pagination_key = self.configuration["pagination_key"]
+        self.valid_objects = []
+        self.variables = {}
+        self.headers = {}
```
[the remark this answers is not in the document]
parthpuri-elastic: It will override [here](https://github.com/elastic/connectors/pull/2134/files#diff-1ac50fbc6d32ff7dfa228d41ad1b8945bd951532d30582faf148690d642ba397R451)

connectors/sources/graphql.py
```diff
@@ -0,0 +1,501 @@
...
+                if duplicate_data:
+                    yield duplicate_data
+
+            if isinstance(duplicate_data, list):
+                for doc in duplicate_data:
+                    yield doc
+
+    def extract_pagination_info(self, data):
+        pagination_key_list = self.pagination_key.split(".")
+        for key in pagination_key_list:
```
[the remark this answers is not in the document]
parthpuri-elastic: No, we currently we are not supporting. I have changed it to pagination_key_path

connectors/sources/graphql.py
```diff
@@ -0,0 +1,501 @@
...
+        self._logger = logger
+        self.graphql_query = self.configuration["graphql_query"]
+        self.url = self.configuration["http_endpoint"]
+        self.graphql_object_list = self.configuration["graphql_object_list"]
+        self.http_method = self.configuration["http_method"]
+        self.authentication_method = self.configuration["authentication_method"]
+        self.pagination_technique = self.configuration["pagination_technique"]
+        self.pagination_key = self.configuration["pagination_key"]
+        self.valid_objects = []
+        self.variables = {}
```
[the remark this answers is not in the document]
parthpuri-elastic: I am overriding it [here](https://github.com/elastic/connectors/pull/2134/files#diff-1ac50fbc6d32ff7dfa228d41ad1b8945bd951532d30582faf148690d642ba397R458)

connectors/sources/graphql.py
```diff
@@ -0,0 +1,501 @@
...
+        Args:
+            key (string): Key of data.
+            data (dict): data to extract
+            same_level (bool): check if object is at same level or not
+
+        Yields:
+            dictionary/list: Documents from the response
+        """
+        for keys in self.graphql_object_list:
+            duplicate_data = data
```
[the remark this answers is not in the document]
parthpuri-elastic: As we are not updating the dict so deep copy is not required but as we are iterating and assigning the value of nested object. So, for next iteration we can start from the top level of dict.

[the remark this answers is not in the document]
jedrazb: I see, in that case a more descriptive name could be `current_level_data` or `nested_data`, which more accurately describes that this variable refers to the current level of data being accessed within the nested structure. I was confused when I saw `duplicate_data` that did something that looked like copying the data structure/

connectors/sources/graphql.py
jedrazb: Given that we won't support incremental syncs I think it's acceptable to keep it that way for now - I can't think of good strategy given that graphQL doesn't have unified way of handling global IDs

connectors/sources/graphql.py
```diff
@@ -0,0 +1,512 @@
...
+            "http_method": {
+                "display": "dropdown",
+                "label": "GET/POST",
+                "options": [
+                    {"label": "GET", "value": "get"},
+                    {"label": "POST", "value": "post"},
+                ],
+                "order": 2,
+                "type": "str",
+                "value": "POST",
```
jedrazb, line 322:
```suggestion
                "value": "post",
```
We should match the value, not the label

connectors/sources/graphql.py
```diff
@@ -0,0 +1,512 @@
...
+                "display": "dropdown",
+                "label": "Pagination model",
+                "options": [
+                    {"label": "No pagination", "value": "no_pagination"},
+                    {"label": "Cursor-based pagination", "value": "cursor_pagination"},
+                ],
+                "order": 11,
+                "tooltip": "Cursor based pagination requires 'pageInfo' field along with argument 'after' for objects mentioned in 'GraphQL Objects List'. It also requires variable for 'after' argument.",
+                "type": "str",
+                "value": "none",
```
jedrazb, line 392:
```suggestion
                "value": "no_pagination",
```
if we want to default to `no pagination` we should set to `no_pagination`

connectors/sources/graphql.py
```diff
@@ -0,0 +1,512 @@
...
+    @classmethod
+    def get_default_configuration(cls):
+        """Get the default configuration for GraphQL.
+
+        Returns:
+            dict: Default configuration.
+        """
+        return {
+            "http_endpoint": {
+                "label": "HTTP URL & Endpoint",
```
jedrazb, line 309:
```suggestion
                "label": "GraphQL HTTP endpoint",
```

connectors/sources/graphql.py
```diff
@@ -0,0 +1,512 @@
...
+        return False
+
+    async def validate_config(self):
+        """Validates whether user input is empty or not for configuration fields
+        Also validate, if user configured repositories are accessible or not and scope of the token
+        """
+        await super().validate_config()
+
+        if not self.validate_endpoints():
+            msg = "HTTP URL & Endpoint are not structured."
```
jedrazb, line 440:
```suggestion
            msg = "GraphQL HTTP endpoint is not a valid URL."
```

connectors/sources/graphql.py
```diff
@@ -0,0 +1,512 @@
...
+            raise ConfigurableFieldValueError(msg)
+
+        if not self.is_query(graphql_query=self.graphql_client.graphql_query):
+            msg = "Configured Query is not supported by the connector."
```
jedrazb, line 444: Can you elaborate that the query is either invalid or is not a query e.g. it's mutation?

jedrazb (commented):
More stylistic issues, getting better! The docs would need to explain well how to setup the cursor based pagination as it's still not super clear

connectors/sources/graphql.py
```diff
@@ -0,0 +1,512 @@
...
+    async def paginated_call(self, graphql_query):
+        if self.pagination_model == "cursor_pagination":
+            ast = parse(graphql_query)  # pyright: ignore
+            visitor = FieldVisitor()
+            visit(ast, visitor)  # pyright: ignore
+
+            self.validate_query(graphql_query, visitor)
+            while True:
+                self._logger.debug(
+                    f"Fetching document with variables: {self.variables} and query: {graphql_query}."
```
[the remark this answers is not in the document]
parthpuri-elastic: Yes, it is the same in all calls, removed !

connectors/sources/graphql.py
```diff
@@ -0,0 +1,519 @@
...
+            },
+            "pagination_model": {
+                "display": "dropdown",
+                "label": "Pagination model",
+                "options": [
+                    {"label": "No pagination", "value": NO_PAGINATION},
+                    {"label": "Cursor-based pagination", "value": CURSOR_PAGINATION},
+                ],
+                "order": 11,
+                "tooltip": "Cursor based pagination requires 'pageInfo' field along with argument 'after' for objects mentioned in 'GraphQL Objects List'. It also requires variable for 'after' argument.",
```
jedrazb, line 397: `It also requires variable for 'after' argument.` - you mean we need variable within the query not a variable in the `Graphql variables` section, it got me slightly confused? I was able to perform a paginated sync without specifying any variables in `Graphql variables` - just added a variable in a query? 

We do need to add example queries in docs as configuration is pretty tricky due to the nature of graphql

Screenshot 2024-04-02 at 10 13 05

jedrazb: Maybe something like, what do you think? I think this tooltip refers to the state before pagination key
```
For cursor-based pagination, add 'pageInfo' and an 'after' argument variable in your query at the desired node (Pagination key). Use 'after' query argument with a variable to iterate through pages. Detailed examples and setup instructions are available in the docs.
```

connectors/sources/graphql.py
```diff
@@ -0,0 +1,519 @@
...
+            page_info = data.get("pageInfo")
+            has_next_page = page_info.get("hasNextPage")
+            end_cursor = page_info.get("endCursor")
+
+            return has_next_page, end_cursor, pagination_key_path[-1]
+        else:
+            msg = "Pagination is enabled but the query is missing 'pageInfo'. Please include 'pageInfo { hasNextPage endCursor }' in the query to support pagination."
+            raise ConfigurableFieldValueError(msg)
+
+    def validate_query(self, graphql_query, visitor):
```
jedrazb, line 182: This only applies to paginated query? Can we rename this to `validate_paginated_query`?

connectors/sources/graphql.py
```diff
@@ -0,0 +1,519 @@
...
+        except Exception as e:
+            self._logger.error(f"Failed to parse GraphQL query. Error: {e}")
+            return False
+
+    def validate_endpoints(self):
+        if re.match(URL_REGEX, self.graphql_client.url):
+            return True
+        return False
+
+    async def validate_config(self):
```
jedrazb, line 440: Small ask: since we can build query ast at this stage already, can we validate that:
1. graphql objects list (. separated) are defined in the graphql query?
2. if pagination is enabled: pagination key (. separated) is a valid path and contains a pageInfo reference?

Does it make sense? My idea is to fail fast and give the feedback quickly in case this is miscomfigured

jedrazb (commented):
Some last nits, I think once addressed this should be good to go! Also, nice job with 93% test coverage!

connectors/sources/graphql.py
jedrazb: Actually let me check with the team how to handle that case, I will get back to you with decision.

connectors/sources/graphql.py
parthpuri-elastic: I tried to think in that direction and attempted to implement it. However, it is not possible because in AST, all the nodes are considered fields. But for a dot-separated object, we need to check if the field is a dictionary as well, which we are doing here.(https://github.com/elastic/connectors/pull/2134/files#diff-1ac50fbc6d32ff7dfa228d41ad1b8945bd951532d30582faf148690d642ba397R159)

parthpuri-elastic:
buildkite test this

parthpuri-elastic:
buildkite test this

jedrazb (approved):
This is a good start! 

As we agreed let's create a followup PR to deal with IDs. After discussion with the team I think it makes sense to:
- Strictly enforce user to specify what property of indexed object(s) should be treated as ID as random ID generation can break user experience in the framework. An RFC should be introduced that forces user to provide field name that should be treated as object id
- Additional issue: since we supporting indexing multiple different objects with same field names, it can lead to mapping errors, take for example:
```
query something {
  user {
     fieldA (this is `int`)
  }
  system {
     fieldA (this is `string`)
  }
}
```
If we index both `user` and `system` objects from response we might encounter mapping exceptions as the types of `fieldA` will differ. Something to keep in mind and at least mention in the docs

parthpuri-elastic:
> This is a good start!
> 
> As we agreed let's create a followup PR to deal with IDs. After discussion with the team I think it makes sense to:
> 
> * Strictly enforce user to specify what property of indexed object(s) should be treated as ID as random ID generation can break user experience in the framework. An RFC should be introduced that forces user to provide field name that should be treated as object id
> * Additional issue: since we supporting indexing multiple different objects with same field names, it can lead to mapping errors, take for example:
> 
> ```
> query something {
>   user {
>      fieldA (this is `int`)
>   }
>   system {
>      fieldA (this is `string`)
>   }
> }
> ```
> 
> If we index both `user` and `system` objects from response we might encounter mapping exceptions as the types of `fieldA` will differ. Something to keep in mind and at least mention in the docs

Sure, I will raise followup PR soon to cover both issues.

PR #2136, opened 2024-02-08
Artem/improve mem queue error message

artem-shelkovnikov:
#### Pre-Review Checklist


connectors/es/sink.py
```diff
@@ -203,6 +209,9 @@ async def run(self):
         except asyncio.CancelledError:
             self._logger.info("Task is canceled, stop Sink...")
             raise
+        except asyncio.QueueFull as e:
+            msg = "Connector was unable to ingest data into overloaded Elasticsearch. Make sure Elasticsearch instance is healthy, has enough resources and content index is healthy."
+            raise ElasticsearchOverloadedError(msg) from e
```
seanstory, line 214: I am not a python expert, but I _think_ that if you `raise` inside an `except`, you automagically get the `__cause__` set for you without having to make a special constructor or use the `from e`

connectors/es/sink.py
```diff
@@ -365,6 +374,14 @@ async def run(self, generator, job_type):
         except asyncio.CancelledError:
             self._logger.info("Task is canceled, stop Extractor...")
             raise
+        except asyncio.QueueFull as e:
+            self._logger.critical("Throttled by Elasticsearch", exc_info=True)
```
seanstory, line 378:
```suggestion
            self._logger.error("Throttled by Elasticsearch", exc_info=True)
```

critical should be for something that is crashing the whole process. This is just failing the job.

connectors/es/sink.py
```diff
@@ -365,6 +374,14 @@ async def run(self, generator, job_type):
...
+            # We clear the queue as we could not actually ingest anything.
+            # After that we indicate that we've encountered an error
+            self.queue.clear()
+            await self.put_doc("FETCH_ERROR")
+            msg = "Connector was unable to ingest data into overloaded Elasticsearch. Make sure Elasticsearch instance is healthy, has enough resources and content index is healthy."
```
seanstory, line 383: worth still sharing how long the queue was full for? Like `"Connector was unable to ingest data into Elasticsearch for 10 minutes. Elasticsearch may be overloaded. Make sure ...` ?

connectors/es/sink.py
artem-shelkovnikov: They are chained automatically, but I personally like `from e` as being more explicit

connectors/es/sink.py
artem-shelkovnikov: Good catch, it's left there from development/debugging :)

connectors/es/sink.py
artem-shelkovnikov: There's no access to the number of minutes here, but it gave me the idea to mention the configuration setting that's responsible for this: 

> Connector was unable to ingest data into overloaded Elasticsearch. Make sure Elasticsearch instance is healthy, has enough resources, content index is healthy or increase \"elasticsearch.bulk.queue_refresh_timeout\" setting.

connectors/es/sink.py
```diff
@@ -78,6 +78,13 @@ class ContentIndexDoesNotExistError(Exception):
     pass
 
 
+class ElasticsearchOverloadedError(Exception):
+    def __init__(self, cause=None):
+        msg = "Connector was unable to ingest data into overloaded Elasticsearch. Make sure Elasticsearch instance is healthy, has enough resources and content index is healthy."
```
timgrein, line 83: Can we include the index name here? "[...] has enough resources and content index ({index_name}) is healthy"

connectors/es/sink.py
```diff
@@ -365,6 +374,13 @@ async def run(self, generator, job_type):
         except asyncio.CancelledError:
             self._logger.info("Task is canceled, stop Extractor...")
             raise
+        except asyncio.QueueFull as e:
+            self._logger.error("Sync was throttled by Elasticsearch")
+            # We clear the queue as we could not actually ingest anything.
+            # After that we indicate that we've encountered an error
+            self.queue.clear()
```
timgrein, line 381: Do you think it could be helpful to log how many documents were deleted? And maybe also log the ids. This could help, if you want to know, which documents were dropped.

artem-shelkovnikov: I'm not sure why - what's the benefit of logging them?

connectors/es/sink.py
artem-shelkovnikov: There's actually no way to do this with current design - Sink does not operate on an index, rather it operates on "operations" abstraction. I can try to get data from the first operation that's left there and get index name, but rather not.

I'm planning to refactor this code more in next release to allow better transparent retries, and this error message might get improved on the way.

connectors/es/sink.py
artem-shelkovnikov: To expand my thoughts: I'm not sure what information it will give me.

Imagine you're syncing 50.000 documents. Elasticsearch got overloaded once 4.500 documents were inserted + 500 are stuck in the queue. So I can log that we dropped 500 operations and their metadata (index + id), but I'm not sure if it helps with anything.

connectors/es/sink.py
timgrein: I think we pass the target index [here](https://github.com/elastic/connectors/blob/main/connectors/es/sink.py#L770) (`async_bulk`), right? So we could simply pass that to `Sink`, too or do we've cases where the target indices differ for bulk operations of the same `Sink` instance?

But ofc nothing blocking, if you think we need some refactoring first, I'm also happy with that.

connectors/es/sink.py
timgrein: I'm just thinking about support cases here, where want to know, why documents we would've expected are missing. Also we could create some telemetry on top of these log messages to get a feeling of how severe these overload errors are (which percentage of documents are dropped).

artem-shelkovnikov: > which percentage of documents are dropped

I don't think it's possible to make a useful metric out of it - if source has 100.000 documents, we've ingested 10.000 and dropped 500, how can this metric help? We know that we dropped 5% of whatever was ingested, but we don't know document total to tell how many were dropped. And since 90.000 documents were not ingested, this 500 is a drop in the ocean.

Maybe I misunderstand something?

connectors/es/sink.py
artem-shelkovnikov: Sink can operate on multiple indices, so passing a single index into it feels like something that breaks design.

We don't have cases when target indices are different, so I can infer which index is being used, but same can be done in UI for now by simply seeing an error message and name of index.

I'd love to put specific name into the message, but current class design is not allowing for that, I want to address it in later work for next release.

artem-shelkovnikov:
@timgrein can I merge this PR and address the comments later?

timgrein (approved):
LGTM 🚢 (we've agreed async to address some of the changes in the next release)

PR #2137, opened 2024-02-08
introduce limited syncs

seanstory:
## Closes https://github.com/elastic/connectors/issues/2129

This PR introduces an optional `parameters` object to the connector sync job record. This will allow us to trigger sync jobs based on the same connector configuration but with different runtime parameters. 

The first such parameter (also introduced here) is `limit`, which will function much like a SQL "limit clause" to constrain how many documents are sent to Elasticsearch.

This does not necessarily influence how many documents are fetched from the 3rd-party - it just prematurely terminates the sync job once the limited number of documents have been pulled from the generator.

On its own, this change is not usable. Neither Kibana, the connectors API, nor the CLI leverage the `parameters` object yet. Those will be addressed in separate PRs.

To test, you can follow a process like:
1. create a connector
2. configure your config.yml with the connector ID
3. start the connectors process
4. configure your connector
5. stop the connectors process
6. start a full sync through Kibana (no connector process is available to pick up this job yet)
7. look up the job record in `.elastic-connectors-sync-jobs` and modify it to have `"parameters": {"limit": 1}`
8. re-start the connectors process
9. the job should index just one document and "complete" successfully

#### Pre-Review Checklist
  - will follow up once these parameters are actually exposed

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -296,6 +296,7 @@ class Extractor:
         concurrent_downloads=DEFAULT_CONCURRENT_DOWNLOADS,
         logger_=None,
         skip_unchanged_documents=False,
+        custom_params=None,
     ):
         if filter_ is None:
             filter_ = Filter()
@@ -318,6 +319,7 @@ class Extractor:
         self._logger = logger_ or logger
         self._canceled = False
         self.skip_unchanged_documents = skip_unchanged_documents
+        self.custom_params = custom_params
 
     async def _deferred_index(self, lazy_download, doc_id, doc, operation):
         data = await lazy_download(doit=True, timestamp=doc[TIMESTAMP_FIELD])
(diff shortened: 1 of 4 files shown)


seanstory:
1. You get a consistent number of docs. 
2. There's no race condition of worrying about interrupting too early, or forgetting to trigger an interrupt.
3. you don't end up with stack traces in your logs (which can happen if the cancelation isn't respected fast enough)
4. you don't end up with red records in your job history 

Screenshot 2024-02-09 at 11 27 27 AM

artem-shelkovnikov:
> You get a consistent number of docs.

Yes, but docs might not be consistent because sources don't guarantee stable sorting often, plus data can change. Idk if it's that important.

> There's no race condition of worrying about interrupting too early, or forgetting to trigger an interrupt.

But also, does it matter that much? Though, I can see a great benefit here only for testing sources with very tight throttling limits (like Salesforce), so it makes a lot of sense.

> you don't end up with stack traces in your logs (which can happen if the cancelation isn't respected fast enough)

True, but I think it's fine

> you don't end up with red records in your job history

Fair enough.

seanstory:
This work has been deprioritized for 8.14. Closing this PR for now

PR #2138, opened 2024-02-08
Make Gmail connector's 'subject' field label match docs

acrewdson:
In the Gmail connector's config, the 'Subject' config field's label didn't match how that field is referred to in the [docs](https://www.elastic.co/guide/en/enterprise-search/8.12/connectors-gmail.html). It should be labeled 'Google Workspace admin email' instead.


navarone-feekery (approved):
Nice catch!

PR #2152, opened 2024-02-13
Pipeline Creation for Building and Pushing Connectors Docker Images

markjhoy:
Creation of a pipeline for scheduled releases of Docker images for Elastic Connectors. This builds, tests, and pushes Docker images for `arm64`, `amd64`, and a multi-architecture image containing both.

This adds a pipeline and scripts that:
* Builds docker images for each platform (and runs a quick test on each to verify they look correct)
* Pushes the images to the registry
* Builds a multi-arch image from the individual pushed images

This will create three new artifacts in the [elastic-connectors Docker registry](https://www.docker.elastic.co/r/enterprise-search/elastic-connectors):
* `docker.elastic.co/enterprise-search/elastic-connectors-amd64:<VERSION>`: The `amd64` platform version of the image
* `docker.elastic.co/enterprise-search/elastic-connectors-arm64:<VERSION>`: The `arm64` platform version of the image
* `docker.elastic.co/enterprise-search/elastic-connectors:<VERSION>`: A multi-architecture image that includes both the `amd64` and `arm64` versions

#### Pre-Review Checklist


.buildkite/publish/build-docker.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [[ "${ARCHITECTURE:-}" == "" ]]; then
+  echo "!! ARCHITECTURE is not set. Exiting."
+  exit 2
+fi
+
+VERSION_PATH="$PROJECT_ROOT/connectors/VERSION"
+VERSION=$(cat $VERSION_PATH)
+
+if [[ "${USE_SNAPSHOT:-}" == "true" ]]; then
```
seanstory, line 22: I don't see anything that ever sets this. Is this just for if you're running locally, instead of through buildkite?

.buildkite/publish/build-docker.sh
```diff
@@ -0,0 +1,37 @@
...
+TAG_NAME="docker.elastic.co/enterprise-search/elastic-connectors-${ARCHITECTURE}:${VERSION}"
+OUTPUT_PATH="$PROJECT_ROOT/.artifacts"
+OUTPUT_FILE="$OUTPUT_PATH/elastic-connectors-docker-${VERSION}-${ARCHITECTURE}.tar.gz"
+
+docker build -t $TAG_NAME .
+
+mkdir -p $OUTPUT_PATH
+docker save $TAG_NAME | gzip > $OUTPUT_FILE
+
+popd
```
seanstory, line 37: TIL `pushd` and `popd`. Did you know that if you're just going back and forth between 2 dirs though, you can just do `cd -` to go back to the last place?

.buildkite/publish/build-multiarch-docker.sh
```diff
@@ -0,0 +1,47 @@
...
+function realpath {
+  echo "$(cd "$(dirname "$1")"; pwd)"/"$(basename "$1")";
+}
+
+SCRIPT_DIR=$(realpath "$(dirname "$0")")
+BUILDKITE_DIR=$(realpath "$(dirname "$SCRIPT_DIR")")
+PROJECT_ROOT=$(realpath "$(dirname "$BUILDKITE_DIR")")
+
+VERSION_PATH="$PROJECT_ROOT/connectors/VERSION"
+VERSION=$(cat $VERSION_PATH)
```
seanstory, line 15: worth extracting this into one file that can be `source` d from multiple places, so that we don't have to maintain it in multiple places?

.buildkite/publish/build-multiarch-docker.sh
```diff
@@ -0,0 +1,47 @@
...
+# ensure +x is set to avoid writing any sensitive information to the console
+set +x
+
+VAULT_ADDR=${VAULT_ADDR:-https://vault-ci-prod.elastic.dev}
+VAULT_USER="docker-swiftypeadmin"
+DOCKER_USER=$(vault read -address "${VAULT_ADDR}" -field user_20230609 secret/ci/elastic-connectors/${VAULT_USER})
+DOCKER_PASSWORD=$(vault read -address "${VAULT_ADDR}" -field secret_20230609 secret/ci/elastic-connectors/${VAULT_USER})
+
+echo "Logging into docker..."
+buildah login --username="${DOCKER_USER}" --password="${DOCKER_PASSWORD}" docker.elastic.co
```
seanstory, line 36: This isn't safe, as it puts the `DOCKER_PASSWORD` in clear text in the buildkite STDOUT. Instead, we should do something like:
```
vault read -address "${VAULT_ADDR}" -field secret_20230609 secret/ci/elastic-connectors/${VAULT_USER} | \
  buildah login --username="${DOCKER_USER}" --password-stdin docker.elastic.co
```

.buildkite/publish/build-multiarch-docker.sh
```diff
@@ -0,0 +1,47 @@
...
+
+echo "Creating manifest..."
+buildah manifest create $TAG_NAME \
```
seanstory, line 39: oh this is cool. I didn't know about `buildah manifest`. Do you know what the tradeoffs are between this and using drivah with their `toml` syntax? Should we switch the data-extraction-service to use `buildah manifest` instead?

.buildkite/publish/push-docker.sh
```diff
@@ -0,0 +1,44 @@
...
+# ensure +x is set to avoid writing any sensitive information to the console
+set +x
+
+VAULT_ADDR=${VAULT_ADDR:-https://vault-ci-prod.elastic.dev}
+VAULT_USER="docker-swiftypeadmin"
+DOCKER_USER=$(vault read -address "${VAULT_ADDR}" -field user_20230609 secret/ci/elastic-connectors/${VAULT_USER})
+DOCKER_PASSWORD=$(vault read -address "${VAULT_ADDR}" -field secret_20230609 secret/ci/elastic-connectors/${VAULT_USER})
+
+echo "Logging into docker..."
+docker login -u $DOCKER_USER -p $DOCKER_PASSWORD docker.elastic.co
```
seanstory, line 41: Same comment on this being insecure and leading to the DOCKER_PASSWORD getting leaked in the buildkite log output. I don't have the syntax handy, but this should also have a way to pipe the password in from STDIN

.buildkite/publish/test-docker.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
```
seanstory, line 2: Could you add some comment blocks to these files to explain at a high level what's happening? I followed the logic of them until this one, which looks like it's using bash to define a run-time addition of an extra buildkite job? If I'm confused now, I expect I'll be extra confused if I have to come troubleshoot something here when a release fails - so any extra explanation/breadcrumbs would be awesome.

seanstory (commented):
This is awesome.

All looks good - have a few nits, and two quick security concern comments. But once those are addressed I think we should go ahead and merge this so that we can try actually running it to publish something.

.buildkite/publish/build-docker.sh
markjhoy: In Buildkite, you can add add environment variables when you launch a build - I figured it was easy enough to at least add the option in...

image

.buildkite/publish/build-multiarch-docker.sh
markjhoy: Yes! I was considering that...

.buildkite/publish/build-multiarch-docker.sh
markjhoy: That I do not... I will look into it though!

.buildkite/publish/push-docker.sh
markjhoy: I will look into it (and above as well)... I did at least make sure of the `set +x` and warning to not echo the commands, but if there's a safer way, I'm all for it...

.buildkite/publish/build-docker.sh
seanstory: Oh sweet! TIL. I was sure there was a way to create special build params, but didn't realize it was this easy to add ad-hoc env vars.

.buildkite/publish/build-multiarch-docker.sh
markjhoy: ... and I don't know if `drivah` can create multi-arch images... :/.

seanstory (approved):

seanstory:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.12|[](https://github.com/elastic/connectors/pull/2219)|

### Questions ?

PR #2155, opened 2024-02-14
Change log level for retry messages

artem-shelkovnikov:
## Part of https://github.com/elastic/enterprise-search-team/issues/6414

Changing log lines that indicate that elasticsearch client retries a call to Elasticsearch to be warnings. This way customers don't need to set log level to DEBUG to see them and we get these lines in logs for our telemetry.

#### Pre-Review Checklist


vidok (dismissed):
LGTM!

connectors/es/client.py
```diff
@@ -221,12 +221,12 @@ async def execute_with_retry(self, func):
 
                 return result
             except ConnectionTimeout:
-                self._logger.debug(f"Attempt {retry}: connection timeout")
+                self._logger.warning(f"Attempt {retry}: connection timeout")
```
seanstory, line 224: I don't love the idea of having a bunch of WARN logs that don't tell you _what_ is being retried. Is there a way we can get the name of `func` or anything so that this can give you a little more information?

artem-shelkovnikov: Good call, I've added this code in https://github.com/elastic/connectors/pull/2155/commits/e6dda428e005d7457ae7b2e39b2c0b9e0fbf4e29

artem-shelkovnikov: Example log line would look like:

```
[FMWK][15:15:29][WARNING][Connector id: RmdNrY0Bo-hBYCnOmSCv, index name: search-outlook, Sync job id: VWdTrY0Bo-hBYCnOLyBy] Client method 'search' retry 1: api error with status 429
```

seanstory: ❤️ love it!

seanstory (approved):

PR #2156, opened 2024-02-14
[Connector API] Update postman collection

jedrazb:
## Changes
Update the postman collection to include `8.13` additions to Connector API


artem-shelkovnikov (commented):
Curiosity question - why index names and ids are not parametrised?

jedrazb:
> Curiosity question - why index names and ids are not parametrised?

Good question, connector ID can be a global param so I adapt the collection. I don't think index_names should be global parameters, they are only used in 2 actions and I think collection user should make a decision what index they are using for testing.

artem-shelkovnikov (approved):

PR #2158, opened 2024-02-14
[8.13] Bump stack versions.

afoucret:
As required in RM 8.13.0 tasks: https://github.com/elastic/enterprise-search-team/issues/6436


seanstory (approved):
Thanks!

PR #2161, opened 2024-02-15
Add atlassian.py for Jira Data Center DLS

moxarth-elastic:
## Related PR https://github.com/elastic/connectors/pull/2108

This PR contain the changes in atlassian.py file for Jira Data Center DLS

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/atlassian.py
+++ b/connectors/sources/atlassian.py
@@ -17,6 +17,11 @@ from connectors.utils import RetryStrategy, iso_utc, retryable
 RETRIES = 3
 RETRY_INTERVAL = 2
 USER_BATCH = 50
+DATA_CENTER_USER_BATCH = 1000
+JIRA_CLOUD = "jira_cloud"
+JIRA_SERVER = "jira_server"
+CONFLUENCE_CLOUD = "confluence_cloud"
+CONFLUENCE_SERVER = "confluence_server"
 
 
 class AtlassianAdvancedRulesValidator(AdvancedRulesValidator):
@@ -99,12 +104,24 @@ class AtlassianAccessControl:
     async def fetch_all_users(self, url):
         start_at = 0
         while True:
-            async for users in self.client.api_call(url=f"{url}?startAt={start_at}"):
+            url_ = (
+                f"{url}?startAt={start_at}"
+                if self.source.configuration["data_source"]
+                in [JIRA_CLOUD, CONFLUENCE_CLOUD]
+                else url.format(start_at=start_at, max_results=DATA_CENTER_USER_BATCH)
+            )
+            async for users in self.client.api_call(url=url_):
                 response = await users.json()
                 if len(response) == 0:
                     return
                 yield response
-                start_at += USER_BATCH
+                if self.source.configuration["data_source"] not in [
+                    JIRA_CLOUD,
+                    CONFLUENCE_CLOUD,
+                ]:
+                    start_at += DATA_CENTER_USER_BATCH
+                else:
+                    start_at += USER_BATCH
 
     async def fetch_user(self, url):
         async for user in self.client.api_call(url=url):
@@ -135,7 +152,7 @@ class AtlassianAccessControl:
                     ACCESS_CONTROL: [<prefixed_account_id>, <prefixed_group_ids>, <prefixed_role_keys>]
                 }
         """
-        account_id = user.get("accountId")
+        account_id = user.get("accountId") or user.get("name")
         account_name = user.get("displayName")
         email = user.get("emailAddress")
         locale = user.get("locale")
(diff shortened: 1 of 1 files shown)


connectors/sources/atlassian.py
```diff
@@ -17,6 +17,11 @@
 RETRIES = 3
 RETRY_INTERVAL = 2
 USER_BATCH = 50
+DATA_CENTER_USER_BATCH = 1000
+JIRA_CLOUD = "jira_cloud"
+JIRA_SERVER = "jira_server"
+CONFLUENCE_CLOUD = "confluence_cloud"
+CONFLUENCE_SERVER = "confluence_server"
```
navarone-feekery, line 24: Same as above; can we remove these from confluence.py and jira.py and use these as imports? I think having a single source is better.

connectors/sources/atlassian.py
```diff
@@ -17,6 +17,11 @@
 RETRIES = 3
 RETRY_INTERVAL = 2
 USER_BATCH = 50
+DATA_CENTER_USER_BATCH = 1000
```
navarone-feekery, line 20: What does `DATA_CENTER_USER_BATCH` specify exactly? Is this the max amount of users we can fetch in one request? Is it the same value as in jira.py? https://github.com/elastic/connectors/blob/1d4fb789f5588d018907aef786b055ba605dc562/connectors/sources/jira.py#L43

If they're the same value, I suggest having it defined in this file (atlassian.py) and importing it into any sources that use it.

navarone-feekery (commented):
Nice, left some comments about importing / keeping it DRY.

connectors/sources/atlassian.py
seanstory: I'm also confused about `DATA_CENTER_USER_BATCH`, specifically because below it seems to be not for DataCenter, but for anything non-cloud:

```
                f"{url}?startAt={start_at}"
                if self.source.configuration["data_source"]
                in [JIRA_CLOUD, CONFLUENCE_CLOUD]
                else url.format(start_at=start_at, max_results=DATA_CENTER_USER_BATCH)
```

connectors/sources/atlassian.py
```diff
@@ -135,7 +152,7 @@ async def user_access_control_doc(self, user):
                     ACCESS_CONTROL: [<prefixed_account_id>, <prefixed_group_ids>, <prefixed_role_keys>]
                 }
         """
-        account_id = user.get("accountId")
+        account_id = user.get("accountId") or user.get("name")
```
seanstory, line 155: Is `user.get("name")` guaranteed to be unique? Looks like `account_id` ends up being used as the access control document's `_id`. 

Is this added because `accountId` is sometimes missing?

connectors/sources/atlassian.py
```diff
@@ -186,7 +203,10 @@ def is_active_atlassian_user(self, user_info):
             )
             return False
 
-        if user_info.get("accountType") != "atlassian":
+        if (
+            self.source.configuration["data_source"] in [JIRA_CLOUD, CONFLUENCE_CLOUD]
+            and user_info.get("accountType") != "atlassian"
+        ):
             self.source._logger.debug(
                 f"Skipping {user_name} because the account type is {user_info.get('accountType')}. Only 'atlassian' account type is supported."
```
seanstory, line 211: Should this log message get updated too? Since now the account type could be `atlassian`, but the data source config is non-cloud.

connectors/sources/atlassian.py
moxarth-elastic: > Is user.get("name") guaranteed to be unique?

Yes, it is unique for all users.

> Is this added because accountId is sometimes missing?

Exactly, `accountId` is missing while fetching users for non-cloud.

connectors/sources/atlassian.py
moxarth-elastic: This is designed to be logged for just cloud data source according to this condition
```
self.source.configuration["data_source"] in [JIRA_CLOUD, CONFLUENCE_CLOUD]
```

Previously, DLS was designed to just cloud platform so did not have this condition but now we've explicitly added it just to exclude non-cloud data source.

connectors/sources/atlassian.py
moxarth-elastic: I think there will be a circular import in that case. we can't import partially initiated jira.py. 
However, we can import it inside the method itself to avoid circular import error.

connectors/sources/atlassian.py
moxarth-elastic: > If they're the same value, I suggest having it defined in this file (atlassian.py) and importing it into any sources that use it.

The value is definitely same but the purpose is different at https://github.com/elastic/connectors/blob/1d4fb789f5588d018907aef786b055ba605dc562/connectors/sources/jira.py#L43 and it is used for both cloud and non-cloud data source. where in `atlassian.py`, the value is used for fetching just non-cloud users.

> I'm also confused about DATA_CENTER_USER_BATCH, specifically because below it seems to be not for DataCenter, but for anything non-cloud:

renaming the constant to get more clear idea.

navarone-feekery (approved):
LGTM, thanks 🚀

PR #2182, opened 2024-02-16
Fix Box and Network Drive ftests

praveen-elastic:
## Closes https://github.com/elastic/connectors-py/issues/2174, https://github.com/elastic/connectors-py/issues/2175, https://github.com/elastic/connectors/issues/2173

This PR aims to allow FTest of Network Drive and Box to work when framework calls ping method multiple times.

#### Pre-Review Checklist


connectors/sources/network_drive.py
```diff
@@ -255,8 +275,10 @@ def __init__(self, configuration):
         self.drive_path = self.configuration["drive_path"]
         self.drive_type = self.configuration["drive_type"]
         self.identity_mappings = self.configuration["identity_mappings"]
-        self.session = None
         self.security_info = SecurityInfo(self.username, self.password, self.server_ip)
+        self.smb_connection = SMBSession(
+            self.server_ip, self.username, self.password, self.port
+        )
```
artem-shelkovnikov, line 281: Can it raise an exception when it's created? If so, we cannot have it in the initialiser

connectors/sources/network_drive.py
```diff
@@ -382,13 +395,20 @@ def find_matching_paths(self, advanced_rules):
 
     async def ping(self):
         """Verify the connection with Network Drive"""
+        if "RUNNING_FTEST" in os.environ:
+            return
```
artem-shelkovnikov, line 399: We definitely should not add this in our connectors. Any ftest code should live outside unless it's really impossible to work around this (e.g. when we're using 3rd-party library most of the time that we need to change URL for).

Can you remove this code and approach it differently?

connectors/sources/network_drive.py
praveen-elastic: It should not raise error, but for safer side i'll update this and move it to a separate method and make it a cached_property like below, WDYT ?

```
    @cached_property
    def smb_connection(self):
        return SMBSession(
            self.server_ip, self.username, self.password, self.port
        )
```

artem-shelkovnikov: Yeah if it works, then it's a good idea. Does SMBSession attempt to access 3rd-party? Cause if so, we might need to make it async too, and then cached property will not work any more :/

connectors/sources/network_drive.py
```diff
@@ -232,6 +233,26 @@ def fetch_members(self, group_name):
...
+        self.session = None
+
+    def create_connection(self):
+        """Creates an SMB session to the shared drive."""
+        self.session = smbclient.register_session(
+            server=self.server_ip,
+            username=self.username,
+            password=self.password,
+            port=self.port,
+        )
```
artem-shelkovnikov, line 253: Do you need to close it somehow btw?

connectors/sources/network_drive.py
```diff
@@ -387,13 +402,18 @@ async def find_matching_paths(self, advanced_rules):
 
     async def ping(self):
         """Verify the connection with Network Drive"""
+
+        if self.smb_connection.session is not None:
+            return
```
artem-shelkovnikov, line 407: Why? We still need to be able to ping the backend.

connectors/sources/network_drive.py
```diff
@@ -387,13 +402,18 @@ async def find_matching_paths(self, advanced_rules):
...
         loop = asyncio.get_running_loop()
-        await loop.run_in_executor(executor=None, func=self.create_connection)
+        await loop.run_in_executor(
+            executor=None, func=self.smb_connection.create_connection
+        )
```
artem-shelkovnikov, line 411: Ok I don't think it's working well - we still rely on the fact that `ping` is called before we can call `get_doc`. Can you make it work such that there is no sequential dependency like this?

E.g. making a method `async def get_connection(self)` that creates a connection if it does not exist, otherwise returns existing connection. This way we won't rely on the sequence of method calls to make connector work.

connectors/sources/network_drive.py
```diff
@@ -706,7 +735,10 @@ async def get_docs(self, filtering=None):
         Yields:
             dictionary: Dictionary containing the Network Drive files and folders as documents
         """
-
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(
+            executor=None, func=self.smb_connection.create_connection
+        )
```
artem-shelkovnikov, line 741: So if ping is called before, then it will create a different connection?

connectors/sources/network_drive.py
```diff
@@ -232,6 +233,26 @@ def fetch_members(self, group_name):
         return self.parse_output(members)
 
 
+class SMBSession(metaclass=Singleton):
```
artem-shelkovnikov, line 236: Why make it singleton?

IMO it becomes problematic potentially if you have several network drive connectors in same deployment. Can you make sure that syncing 2 network drives at the same time works when doing these changes?

connectors/sources/network_drive.py
praveen-elastic: the connection is closed via `delete_session` method being called in close() method in NASDataSource class. We can move the actual logic in this class within a method `delete_connection()` and call that method inside close() of source class

connectors/sources/network_drive.py
praveen-elastic: Truly sorry, I was trying multiple approaches and somehow missed to remove the singleton approach.

connectors/sources/network_drive.py
praveen-elastic: I have also removed the singleton approach so now there is no dependency between ping and get_docs. the create_connection method itself handles the case of:  `create a connection if it does not exist, otherwise returns existing connection`,  and create_connection is called at the start of get_docs too. So both ping and get_docs works independently.

connectors/sources/network_drive.py
praveen-elastic: Apologies, I was trying multiple approaches and somehow missed to remove the singleton approach.

connectors/sources/network_drive.py
praveen-elastic: Added this to make ping and get_docs independent of each other. `create_connection` works as `get_connection` method you mentioned in earlier comment

artem-shelkovnikov (dismissed):
Great job @praveen-elastic !

connectors/sources/network_drive.py
```diff
@@ -485,6 +485,10 @@ async def fetch_file_content(self, path):
         """
         self._logger.debug(f"Fetching the contents of file on path: {path}")
         try:
+            loop = asyncio.get_running_loop()
+            await loop.run_in_executor(
+                executor=None, func=self.smb_connection.create_connection
+            )
```
artem-shelkovnikov, line 491: Let's be consistent and `await asyncio.to_thread` everywhere

artem-shelkovnikov: Additionally - the connection should have already be created here, right?

praveen-elastic: No, the pull from main, added merge changes from PR of @akanshi-elastic related to local content extraction feature for Network Drive, which introduced a change in the way files would be downloaded in the connector.

I took a call with @akanshi-elastic and we found that:
Previously, the connector would itself call fetch_file_content : https://github.com/elastic/connectors/blob/8ce8283abee1ca1ce70eec4bca026d77cda66771/connectors/sources/network_drive.py#L494, so the connection would always be existing.
However, after the changes of local extraction PR, the method `fetch_file_content` is now passed as download_func to `download_and_extract_file` and then it's called outside the connector: https://github.com/elastic/connectors/commit/9b9609da683ca429d8e19fd1beeff35182b6d8a6#diff-29e12700a1a4cf24af22d5bf0a870c9ebfede0ec41dc991406b951d0100393ffR501, so it is not guaranteed that the connection would be existing.

connectors/sources/network_drive.py
```diff
@@ -485,6 +485,10 @@ async def fetch_file_content(self, path):
...
             with smbclient.open_file(
```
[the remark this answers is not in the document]
praveen-elastic: Yes, we thought about it but there is no other way to find if the connection is existing, register_session method itself returns the same connection if it already exists: https://github.com/jborean93/smbprotocol/blob/459100718eb6f54e6b7f90df6bb94f1a139e9b33/src/smbclient/_pool.py#L397

connectors/sources/network_drive.py
artem-shelkovnikov: But this is something I do not get - download cannot happen without us listing the objects, so how can we download something without meeting it when listing files/directories?

praveen-elastic: updated it and removed the create connection from here 👍

artem-shelkovnikov (approved):

PR #2184, opened 2024-02-16
Remove unsafe dict call from ServiceNow

navarone-feekery:
## Closes https://github.com/elastic/connectors/issues/2164

Remove the unsafe dict call from ServiceNow.
There are circumstances where tables fetched from `sys_db_object` may be missing `name` or `label` fields. Even if the user is not trying to query those tables, they are checked during the filtering operation. If `name` or `label` is missing, the connector shouldn't crash with the unsafe dict call. It should log the issue as debug and continue syncing.

### DIFF
--- a/connectors/sources/servicenow.py
+++ b/connectors/sources/servicenow.py
@@ -324,9 +324,21 @@ class ServiceNowClient:
                 ]
                 async for table_data in self.get_data(batched_apis=batched_apis):
                     for mapping in table_data:  # pyright: ignore
-                        if mapping["label"] in invalid_services:
-                            servicenow_mapping[mapping["label"]] = mapping["name"]
-                            invalid_services.remove(mapping["label"])
+                        name = mapping.get("name")
+                        if not name:
+                            msg = f"Entry in sys_db_object without sysparm_field 'name' was found. This is a non-issue if no invalid services are flagged."
+                            self._logger.debug(msg)
+                            continue
+
+                        label = mapping.get("label")
+                        if not label:
+                            msg = f"Entry in sys_db_object without sysparm_field 'label' was found. This is a non-issue if no invalid services are flagged."
+                            self._logger.debug(msg)
+                            continue
+
+                        if label in invalid_services:
+                            servicenow_mapping[label] = name
+                            invalid_services.remove(label)
 
             return servicenow_mapping, invalid_services
 
--- a/connectors/sources/servicenow.py
+++ b/connectors/sources/servicenow.py
@@ -326,13 +326,13 @@ class ServiceNowClient:
                     for mapping in table_data:  # pyright: ignore
                         name = mapping.get("name")
                         if not name:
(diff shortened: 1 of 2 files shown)


connectors/sources/servicenow.py
```diff
@@ -324,9 +324,21 @@ async def filter_services(self, configured_service):
...
-                            invalid_services.remove(mapping["label"])
+                        name = mapping.get("name")
+                        if not name:
+                            msg = "Entry in sys_db_object without sysparm_field 'name' was found. This is a non-issue if no invalid services are flagged."
+                            self._logger.debug(msg)
+                            continue
+
+                        label = mapping.get("label")
+                        if not label:
+                            msg = "Entry in sys_db_object without sysparm_field 'label' was found. This is a non-issue if no invalid services are flagged."
```
jedrazb, line 335: Nit, can we avoid code (string) duplication and refactor it to a func?

connectors/sources/servicenow.py
```diff
@@ -324,9 +324,21 @@ async def filter_services(self, configured_service):
                 ]
                 async for table_data in self.get_data(batched_apis=batched_apis):
                     for mapping in table_data:  # pyright: ignore
-                        if mapping["label"] in invalid_services:
-                            servicenow_mapping[mapping["label"]] = mapping["name"]
-                            invalid_services.remove(mapping["label"])
+                        name = mapping.get("name")
+                        if not name:
+                            msg = "Entry in sys_db_object without sysparm_field 'name' was found. This is a non-issue if no invalid services are flagged."
+                            self._logger.debug(msg)
```
jedrazb, line 330: I'm not a servicenow expert but can we at least provide some "id" of a mapping that is missing name or label? Otherwise it's a bit hard to debug from the user perspective

connectors/sources/servicenow.py
```diff
@@ -324,9 +324,20 @@ async def filter_services(self, configured_service):
                 ]
                 async for table_data in self.get_data(batched_apis=batched_apis):
                     for mapping in table_data:  # pyright: ignore
-                        if mapping["label"] in invalid_services:
-                            servicenow_mapping[mapping["label"]] = mapping["name"]
-                            invalid_services.remove(mapping["label"])
+                        sys_id = mapping.get("sys_id")
```
jedrazb, line 327: Hopefully this can't be null 🤞 ?

jedrazb (approved):
LGTM

connectors/sources/servicenow.py
jedrazb: It shouldn't be https://www.servicenow.com/community/developer-articles/servicenow-sys-id-importance-of-servicenow-sys-id-servicenow-sys/ta-p/2314412

navarone-feekery: I'm 99.999% sure it can't be, if this is also null then something is very wrong with their table 😅

PR #2188, opened 2024-02-19
Generate API keys when running native connector syncs

navarone-feekery:
## Follow-up PR for https://github.com/elastic/connectors/pull/2115

If a native connector is missing an `api_key_secret_id` field or value, generate an API key and store it in Connector Secrets it before syncing.

These changes allow users upgrading to 8.13 to sync immediately without needing to configure API keys for their native connectors. It also allows Connectors CLI to create connectors without worrying about generating and storing API keys.

### Changelist

- Generate an API key if a native connector is missing `api_key_secret_id` when syncing
- Store the generated API key in connector's secrets storage
- Update the connector doc with the new `api_key_id` and `api_key_secret_id`
- Make feature `native_connector_api_keys_enabled` default to `True`
- Add unit tests


artem-shelkovnikov (commented):
This is a little scary - why does this happen here, not in Kibana that creates a native connector?

navarone-feekery:
@artem-shelkovnikov there are two scenarios where this is useful.

1. Connectors created by CLI while using API as authentication can't generate API keys, because users can't generate API keys if they're using an API key themselves to authenticate.
2. Connectors prior to 8.13 will be missing an API key, this allows the API key to be generated automatically rather than the user needing to manually configure anything.

I feel that manual configuration goes against the ethos of a native connector so it made sense to me for this to be an automated process. What about it is scary to you?

artem-shelkovnikov:
I was a bit confused, feeling like on-prem connectors would fetch api key from the secrets storage - as far as I understand now, it's a change only for native connectors.

Thanks for the explanation!

connectors/es/management_client.py
```diff
@@ -193,3 +198,49 @@ async def get_connector_secret(self, connector_secret_id):
...
+                        ],
+                        "privileges": ["all"],
+                    }
+                ],
+            }
+        }
+        api_key_result = await self._retrier.execute_with_retry(
+            partial(
+                self.client.security.create_api_key,
+                name=f"{index_name}-connector",
```
artem-shelkovnikov, line 241: Why `-connector`?

connectors/services/job_execution.py
```diff
@@ -80,6 +81,34 @@ async def _sync(self, sync_job):
         if not self.should_execute(connector, sync_job):
             return
 
+        if (
+            connector.native
+            and connector.features.native_connector_api_keys_enabled()
```
artem-shelkovnikov, line 86: I see it's a killswitch for the problems in future, nice!

connectors/services/job_execution.py
```diff
@@ -80,6 +81,34 @@ async def _sync(self, sync_job):
...
+            and not connector.api_key_secret_id
+        ):
+            sync_job.log_debug(
+                f"No secret found for [{connector_id}], generating and storing API key..."
+            )
+            es_management_client = ESManagementClient(
+                self._override_es_config(connector)
+            )
```
artem-shelkovnikov, line 94: I'd move this to initialiser

connectors/services/job_execution.py
```diff
@@ -80,6 +81,34 @@ async def _sync(self, sync_job):
...
+            try:
+                storage_result = await es_management_client.generate_and_store_api_key(
+                    connector.index_name
+                )
+                sync_job.log_debug(
+                    f"API key stored as secret: {storage_result}, updating connector doc..."
+                )
+                await self.connector_index.update(connector_id, storage_result)
+                connector = await self.connector_index.fetch_by_id(connector_id)
```
artem-shelkovnikov, line 103: You can do `await connector.reload()`

artem-shelkovnikov (approved):
Looks good, you can address my comments if you want in a different PR

connectors/es/management_client.py
navarone-feekery: The API key name is arbitrary and is only useful for seeing the API key in the Kibana UI. This is just identical implementation to Kibana.
I think the `-connector` suffix is because the API key will have access to the connector's index, ACL index, and the `.elastic-connectors` index. So it's a very connector-centric API key.

navarone-feekery:
@artem-shelkovnikov I had to rebase, could you take another look please?

artem-shelkovnikov:
Great work!

navarone-feekery:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.13|[](https://github.com/elastic/connectors/pull/2198)|

### Questions ?

PR #2196, opened 2024-02-20
Fix native API key authentication

navarone-feekery:
API key authentication for native connectors is currently broken, because the `ESManagementClient.client.options` update func doesn't change requests to use API key authentication.
This PR changes how the `ESManagementClient` is initialised:

1. Fetch API key from connector secrets
2. Attach API key to es_config
3. Initialise ESManagementClient (using API key from the start)

To do this, it's necessary to create a second ESManagementClient that authenticates with username and password. This is because the connector secrets API is only accessible by the Enterprise Search system role.

### DIFF
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -7,11 +7,15 @@ import asyncio
 import time
 
 import elasticsearch
+from elasticsearch import (
+    NotFoundError as ElasticNotFoundError,
+)
 
 from connectors.config import DataSourceFrameworkConfig
 from connectors.es.client import License, with_concurrency_control
 from connectors.es.index import DocumentNotFoundError
 from connectors.es.license import requires_platinum_license
+from connectors.es.management_client import ESManagementClient
 from connectors.es.sink import OP_INDEX, SyncOrchestrator, UnsupportedJobType
 from connectors.logger import logger
 from connectors.protocol import JobStatus, JobType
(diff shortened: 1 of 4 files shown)


connectors/sync_job_runner.py
```diff
@@ -186,6 +193,33 @@ async def execute(self):
...
+            self.sync_job.log_debug(
+                f"Checking secrets storage for API key for index [{self.connector.index_name}]..."
+            )
+            api_key = await es_management_client.get_connector_secret(
+                self.connector.api_key_secret_id
+            )
+            self.sync_job.log_debug(
+                f"API key found in secrets storage for index [{self.connector.index_name}], will use this for authorization."
+            )
+            self.es_config["api_key"] = api_key
```
artem-shelkovnikov, line 216: Just to check - I'm not sure if `es_config` here is a copy of original config or not. If it's a copy, then this is fine, if it's a reference to original object, then it'll break if multiple connectors are running at the same time.

Can you verify that running multiple connectors in this scenario won't break?

navarone-feekery: @artem-shelkovnikov the `self.es_config` in `SyncJobRunner` is initially deepcopied from the real config [here](https://github.com/elastic/connectors/blob/b155f2a42e8a9457d8ed463c08dca217d7f2846e/connectors/services/job_execution.py#L83-L89), by the [_override_es_config](https://github.com/elastic/connectors/blob/b155f2a42e8a9457d8ed463c08dca217d7f2846e/connectors/services/base.py#L146C1-L159C25) func.
The `es_config` is initially just the [`elasticsearch` section of the config](https://github.com/elastic/connectors/blob/b155f2a42e8a9457d8ed463c08dca217d7f2846e/connectors/services/base.py#L74). So this config update won't affect multiple connectors.
(For sanity's sake, I ran two connectors together with this setup and they ran without issue).

artem-shelkovnikov (approved):

PR #2206, opened 2024-02-22
Add more logs in Outlook connector

moxarth-elastic:
## Fixes https://github.com/elastic/connectors/issues/2168

Adding more logs to Outlook connectors.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/outlook.py
+++ b/connectors/sources/outlook.py
@@ -835,9 +835,6 @@ class OutlookDataSource(BaseDataSource):
 
     async def _fetch_attachments(self, attachment_type, outlook_object, timezone):
         for attachment in outlook_object.attachments:
-            self._logger.debug(
-                f"Fetching attachment '{attachment.name}' for {attachment_type} record."
-            )
             await self.queue.put(
                 (
                     self.doc_formatter.attachment_doc_formatter(
--- a/connectors/sources/outlook.py
+++ b/connectors/sources/outlook.py
@@ -834,8 +834,8 @@ class OutlookDataSource(BaseDataSource):
         yield content
 
     async def _fetch_attachments(self, attachment_type, outlook_object, timezone):
-        self._logger.debug(f"Fetching attachments for {attachment_type}")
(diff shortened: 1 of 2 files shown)


connectors/sources/outlook.py
```diff
@@ -831,6 +834,7 @@ async def download_func(self, content):
         yield content
 
     async def _fetch_attachments(self, attachment_type, outlook_object, timezone):
+        self._logger.debug(f"Fetching attachments for {attachment_type}")
```
artem-shelkovnikov, line 837: This line should specify what the attachments are fetched for

connectors/sources/outlook.py
```diff
@@ -831,6 +834,7 @@ async def download_func(self, content):
...
         for attachment in outlook_object.attachments:
```
artem-shelkovnikov, line 838: Does this line make a call to Outlook Server?

moxarth-elastic: No, we get all the attachments in `outlook_object` i.e. from mail object, task object and calendar object. It does not call an extra call to server.

connectors/sources/outlook.py
```diff
@@ -832,6 +835,7 @@ async def download_func(self, content):
 
     async def _fetch_attachments(self, attachment_type, outlook_object, timezone):
         for attachment in outlook_object.attachments:
+            self._logger.debug(f"Fetching attachment for {attachment.name}")
```
artem-shelkovnikov, line 838: It's good to log the attachment name, but we need to log what outlook_object we're fetching attachment too, since afaik the connector is fetching data for multiple hierarchies at the same time.

E.g.:

`Fetching attachment "test_plan.xls" for calendar record.` or something, ideally indicating what the parent object is, so it would be possible to correlate this line to other lines you added.

moxarth-elastic: Got it. I've added a parent object in the log message, here `attachment_type` indicates the parent object of the attachment.

artem-shelkovnikov (approved):

PR #2228, opened 2024-03-05
Fix problem with cancel jobs action using closed client

artem-shelkovnikov:
## Problem found during https://github.com/elastic/enterprise-search-team/issues/7054

This PR fixes a problem that CLI uses already closed client during "cancel job" action.

This problem started happening because new transient retrier relies on proper usage of `close` method and stops executing actions if it's closed. Other clients do not (yet) follow this pattern, so the problem did not manifest itself before.

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli/job.py
+++ b/connectors/cli/job.py
@@ -79,9 +79,15 @@ class Job:
 
     async def __async_cancel_jobs(self, connector_id, index_name, job_id):
         try:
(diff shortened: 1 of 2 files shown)


vidok (approved):
Thanks for fixing this 👍🏻

vidok (approved):

PR #2230, opened 2024-03-05
Fix table/db names not being escaped for MySQL connector

artem-shelkovnikov:
## Closes https://github.com/elastic/enterprise-search-team/issues/6495

It's possible to create DB and TABLE names in MySQL that will break current connector, for example:

```
mysql> create database `really@strange*database$name`;
mysql> use `really@strange*database$name`;
mysql> create table `really@strange*table$name` (id MEDIUMINT NOT NULL AUTO_INCREMENT, name CHAR(30) NOT NULL, PRIMARY KEY (id) );
mysql> INSERT INTO `really@strange*table$name`(name)  VALUES('dog');
```
If you try to select without escaping the table name, it will not work, the table name has to be escaped (see PR changes). This PR makes sure that all places that need to have table names escaped properly do so.

#### Pre-Review Checklist

## Release Note

MySQL Connector: Fix an issue when special characters in database/table names make it impossible to sync data.

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -51,7 +51,7 @@ class MySQLQueries(Queries):
         return f"SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = '{self.database}' AND TABLE_NAME = '{table}' AND COLUMN_KEY = 'PRI'"
 
     def table_data(self, table):
-        return f"SELECT * FROM {self.database}.{table}"
+        return f"SELECT * FROM `{self.database}`.`{table}`"
 
     def table_last_update_time(self, table):
         return f"SELECT UPDATE_TIME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = '{self.database}' AND TABLE_NAME = '{table}'"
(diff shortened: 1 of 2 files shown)


timgrein (commented):
Do we also need to use backticks at the other places like inside `table_last_update_time`, `table_primary_key` etc.? What's the difference between escaping via backticks and "normal" single quotes?

artem-shelkovnikov:
@timgrein so the difference is the following, quoting documentation:

> Certain objects within MySQL, including database, table, index, column, alias, view, stored procedure, partition, tablespace, resource group and other object names are known as identifiers. 
> (see [here](https://dev.mysql.com/doc/refman/8.3/en/identifiers.html)).
> ...
> An identifier may be quoted or unquoted. If an identifier contains special characters or is a reserved word, you must quote it whenever you refer to it. (Exception: A reserved word that follows a period in a qualified name must be an identifier, so it need not be quoted.) Reserved words are listed at [Section 11.3, “Keywords and Reserved Words”](https://dev.mysql.com/doc/refman/8.3/en/keywords.html).
> ...
> The identifier quote character is the backtick (`):
> ...
> If the [ANSI_QUOTES](https://dev.mysql.com/doc/refman/8.3/en/sql-mode.html#sqlmode_ansi_quotes) SQL mode is enabled, it is also permissible to quote identifiers within double quotation marks:

So in short, if we're querying something that we're not sure whether needs to be quoted or not, then we need to escape it with "`" character. If we're querying columns that we know are not gonna break the naming conventions (like mentioned `table_last_update_time` and `table_primary_key`).

> What's the difference between escaping via backticks and "normal" single quotes?
Single quotes are not escaping, rather indication that it's a string when running a query. E.g. same as:

```
int field; # <<< field here is an identifier in terms of MySQL

value = fields.where(x=>x.name == 'field'); # <<< field here is a string value
```

timgrein (approved):
Thank you for the explanation! LGTM

PR #2231, opened 2024-03-05
Add Advanced Sync Rules for Notion connector

praveen-elastic:
## Part of https://github.com/elastic/connectors-py/issues/1958 

This PR adds Advanced Sync Rules for Notion connector.

#### Pre-Review Checklist

## Release Note

Support added for advanced sync rules in Notion Connector.

### DIFF
--- a/connectors/sources/notion.py
+++ b/connectors/sources/notion.py
@@ -178,13 +178,12 @@ class NotionAdvancedRulesValidator(AdvancedRulesValidator):
         "items": {
             "type": "object",
             "properties": {
-                "database_id": {"type": "string", "minProperties": 1},
+                "database_id": {"type": "string", "minLength": 1},
                 "filter": {
                     "type": "object",
-                    "properties": {"property": {"type": "string", "minProperties": 1}},
+                    "properties": {"property": {"type": "string", "minLength": 1}},
                     "additionalProperties": {
                         "type": ["object", "array"],
-                        "minProperties": 1,
                     },
                 },
             },
@@ -224,6 +223,7 @@ class NotionAdvancedRulesValidator(AdvancedRulesValidator):
 
     def __init__(self, source):
         self.source = source
+        self._logger = logger
 
     async def validate(self, advanced_rules):
         if len(advanced_rules) == 0:
@@ -231,6 +231,7 @@ class NotionAdvancedRulesValidator(AdvancedRulesValidator):
                 SyncRuleValidationResult.ADVANCED_RULES
             )
 
+        self._logger.info("Remote validation started")
         return await self._remote_validation(advanced_rules)
 
     async def _remote_validation(self, advanced_rules):
@@ -253,6 +254,7 @@ class NotionAdvancedRulesValidator(AdvancedRulesValidator):
             databases.append(document.get("id").replace("-", ""))
         for rule in advanced_rules:
             if "database_query_filters" in rule:
+                self._logger.info("Validating database query filters")
                 for database in advanced_rules.get("database_query_filters"):
                     database_id = (
                         database.get("database_id").replace("-", "")
@@ -268,6 +270,7 @@ class NotionAdvancedRulesValidator(AdvancedRulesValidator):
                         validation_message=f"Invalid database id: {', '.join(invalid_database)}",
                     )
             if "searches" in rule:
+                self._logger.info("Validating search filters")
                 for database_page in advanced_rules.get("searches"):
                     if database_page.get("filter", {}).get("value") == "page":
                         page_title.append(database_page.get("query"))
@@ -284,6 +287,7 @@ class NotionAdvancedRulesValidator(AdvancedRulesValidator):
                         is_valid=False,
                         validation_message=str(error),
                     )
+            self._logger.info("Remote validation successful")
         return SyncRuleValidationResult.valid_result(
             SyncRuleValidationResult.ADVANCED_RULES
         )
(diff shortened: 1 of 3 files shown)


tests/sources/test_notion.py
```diff
@@ -548,3 +556,126 @@ async def test_fetch_children_recursively():
...
+                ]
+            },
+            SyncRuleValidationResult(
+                SyncRuleValidationResult.ADVANCED_RULES,
+                is_valid=False,
+                validation_message=ANY,
+            ),
+        ),
+    ],
+)
```
seanstory, line 668: can you add examples where `is_valid=True`? Please be sure to capture the edge cases like:
- multiple searches
- different filter types
- illegal checkbox values
- etc

tests/sources/test_notion.py
```diff
@@ -548,3 +556,126 @@ async def test_fetch_children_recursively():
...
+                "searches": [
+                    {
+                        "filter": {"value": "database"},
+                        "query": "Demo",
+                    }
+                ]
+            },
+            SyncRuleValidationResult(
+                SyncRuleValidationResult.ADVANCED_RULES,
+                is_valid=False,
```
seanstory, line 623: Can you add comments to each of these to explain why each is False? This will help us maintain/troubleshoot down the road.

docs/sync-rules/NOTION.md
```diff
@@ -0,0 +1,176 @@
...
+
+  "database_query_filters" : [
+    {
+    "database_id": "notion_database_id1",
+    "filter": ​​{
+		"property": "Task completed",
+	      "checkbox": {
+	        "equals": true
+	      }
+	    }
```
seanstory, line 164: formatting looks a little weird here

docs/sync-rules/NOTION.md
```diff
@@ -0,0 +1,176 @@
...
+    }
+  ]
+}
+
+]
+```
+**Note** 
+For `database_query_filters`:
+    - `property` contains a string that matches the database column name.
+    - `checkbox` this is not constant depends on the column name property type it can be `date`,`files`,`formula`,`multi_select`,`number`,`people`,`phone_number`,`relation`,`rich_text`,`select`,`status`,`timestamp` or `ID`
```
seanstory, line 174: I don't think I saw this restriction captured in the JSON Schema. Why not add it?

connectors/sources/notion.py
```diff
@@ -158,18 +163,138 @@ async def fetch_comments(self, block_id):
...
+        "additionalProperties": False,
+    }
+
+    SCHEMA = fastjsonschema.compile(definition=RULES_OBJECT_SCHEMA_DEFINITION)
+
+    def __init__(self, source):
+        self.source = source
+
+    async def validate(self, advanced_rules):
+        if len(advanced_rules) == 0:
```
seanstory, line 229: Please add some logging to this function, so that we'll be able to troubleshoot what went wrong if something goes wrong.

connectors/sources/notion.py
```diff
@@ -443,3 +555,53 @@ async def get_docs(self, filtering=None):
                     block_id=block_id
                 ):
                     yield self._format_doc(comment), None
+
+    async def get_docs(self, filtering=None):
```
seanstory, line 559: Please add some logging so that it's clear which flow we're going down, and what queries are being executed against the notion client.

seanstory (commented):
Largely looks good. I don't see any obvious bugs, but I would like to see some more observability and some tighter tests. Left comments to that effect.

Thanks!

praveen-elastic:
Buildkite test this

docs/sync-rules/NOTION.md
praveen-elastic: Fixed the formatting issues 👍

docs/sync-rules/NOTION.md
praveen-elastic: Since these are the list of filters already listed on Notion's documentation I have removed them now from our docs and added the link to the Notion Documentation highlighting it supports Notion filter syntax. Also, it would be quite complex for handling all of them in JSON schema when compounded with `and` `or`, therefore we can't add them all in schema.

seanstory: I was just figuring we could use a [json schema enum](https://json-schema.org/understanding-json-schema/reference/enum) for `checkbox`, no?

seanstory (approved):
Looks good, with the one lingering question about JSON schema enum. If you've got a good answer for that though, I think this is ready to go in.

docs/sync-rules/NOTION.md
praveen-elastic: [json schema enum](https://json-schema.org/understanding-json-schema/reference/enum) cannot be used here since it only works for defining allowed list of values for the keys known in advance, for example, in the below schema, enum is used to provide allowed list of values for `street_type`. However, in our case we want to define allowed list of keys for an object. 

Enum use-case:
```
{
  "type": "object",
  "properties": {
    "street_type": { "enum": ["Street", "Avenue", "Boulevard"] }
  },
}
```

seanstory: I thought that `checkbox` was always the key name, with an enum of potential values like [`date`,`files`,`formula`,`multi_select`,`number`,`people`,`phone_number`,`relation`,`rich_text`,`select`,`status`,`timestamp`, `ID`]

praveen-elastic:
> Looks good, with the one lingering question about JSON schema enum. If you've got a good answer for that though, I think this is ready to go in.

Merging the PR since we can't use JSON enum schema for our case as discussed above.

PR #2233, opened 2024-03-06
Revert "Generate API keys when running native connector syncs  #2188"

navarone-feekery:
Revert #2188

It was discovered through QA that API keys generated by Enterprise Search on cloud cannot provide access to indices created by the user in Kibana, so this feature doesn't work.
Instead, I will make follow-up PRs providing verbose errors + instructions on how to fix native connectors that fail API key auth.


artem-shelkovnikov (approved):
:crying:

PR #2240, opened 2024-03-07
Catch unauthorized errors during sync

navarone-feekery:
Catch errors when connector is unauthorized to access a content index during sync and provide a human-readable/actionable error message.

Previous to this the error output would be:

```
Sync failure: AuthorizationException: AuthorizationException(403, 'security_exception', 'action [indices:data/read/search] is unauthorized for API key id [<API key id>] of user [<username>] on indices [<index-name>], this action is granted by the index privileges [read,all]').
```

### DIFF
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -196,12 +196,7 @@ class SyncJobRunner:
         except ConnectorJobCanceledError:
             await self._sync_done(sync_status=JobStatus.CANCELED)
         except ElasticAuthorizationException as e:
-            index_name = (
-                self.connector.acl_index_name
-                if job_type == JobType.ACCESS_CONTROL
-                else self.connector.index_name
-            )
-            error_msg = f"Connector is not authorized to access index [{index_name}]. API key may need to be regenerated. Status code: [{e.status_code}]."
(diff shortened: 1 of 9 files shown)


connectors/sync_job_runner.py
```diff
@@ -192,6 +195,15 @@ async def execute(self):
             await self._sync_done(sync_status=JobStatus.SUSPENDED)
         except ConnectorJobCanceledError:
             await self._sync_done(sync_status=JobStatus.CANCELED)
+        except ElasticAuthorizationException as e:
+            index_name = (
+                self.connector.acl_index_name
+                if job_type == JobType.ACCESS_CONTROL
+                else self.connector.index_name
+            )
```
wangch079, line 203: You don't need this. You can always use `self.sync_job.index_name`:

https://github.com/elastic/connectors/blob/2d5df6a3ac11a6b6124886c980f406ed7ce1d052/connectors/protocol/connectors.py#L1002-L1007

wangch079 (dismissed):
LGTM, left one comment

connectors/sync_job_runner.py
navarone-feekery: Oh neat! I've changed it to use this. Thank you.

navarone-feekery:
@wangch079 can you re-review please? Adding changes set your review to stale.

timgrein (approved):
Nice change 🚀

PR #2242, opened 2024-03-07
[PostgreSQL] Add custom pagination to reduce memory consumption

parthpuri-elastic:
## Closes https://github.com/elastic/connectors-py/issues/2222

After some analysis we found that  even using the cursor, it fetches whole table data at once and because of that memory will increase. To fix this issue we added custom pagination on query using limit-offset.

Now for given test data (1M rows with a single JSON field) connectors requires `152.5` MB.

image

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/postgresql.py
+++ b/connectors/sources/postgresql.py
@@ -38,6 +38,8 @@ from connectors.utils import (
     retryable,
 )
 
+FETCH_LIMIT = 1000
+
 
 class PostgreSQLQueries(Queries):
     """Class contains methods which return query"""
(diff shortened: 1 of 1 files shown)


connectors/sources/postgresql.py
```diff
@@ -295,21 +299,50 @@ async def data_streamer(self, table=None, query=None):
...
-        ):
-            yield data
+        if query is None and row_count is not None and order_by_columns is not None:
+            order_by_columns_list = ",".join(order_by_columns)
+            offset = 0
+            fetch_columns = True
+            while True:
+                async for data in fetch(
+                    cursor_func=partial(
+                        self.get_cursor,
```
vidok, line 309: What's the role of the cursor here?

parthpuri-elastic: So, basically, this is a generic functionality for PostgreSQL, Oracle and MS-SQL. Currently, I do not remove the cursor-based approach. I have divided the query to fetch data into chunks.

parthpuri-elastic:
buildkite test this

PR #2252, opened 2024-03-12
[CLI] Create and store API keys for native connectors in CLI

navarone-feekery:
### Summary
We had to revert automated API key generation [here](https://github.com/elastic/connectors/pull/2233) because it didn't work in Cloud as we expected.
Previously, we would just allow native connectors to be created in CLI without an API key, and rely on it being generated before a sync. Now that we know this isn't possible, we need to explicitly generate and store API keys for native connectors when they are created in CLI.

### Changes

- Allow API keys to be created for native connectors
- Store native connector API keys to secrets
- Update `api_key_secret_id` field on connector doc with secret id

### DIFF
--- a/connectors/cli/connector.py
+++ b/connectors/cli/connector.py
@@ -117,24 +117,29 @@ class Connector:
                 )
 
             api_key_id = None
+            api_key_secret_id = None
             api_key_encoded = None
             api_key_error = None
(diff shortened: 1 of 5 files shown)


seanstory (approved):
lgtm

PR #2261, opened 2024-03-14
Clarify Elasticsearch Client Log Message While Waiting for Connection

markjhoy:
## Helps with https://github.com/elastic/connectors/issues/1826

The `self.host` variable for the Elasticsearch client is set up before the AsyncElasticsearch client is created, and thus does not have various options set (such as `ca_certs` in this case) - or, better stated, the defaults for the NodeConfig are used. Once the AsyncElasticsearch client is created, a copy of the `self.host` NodeConfig is created and the additional options are merged.

This PR logs only the URL from the `self.host` at the INFO level. If DEBUG logging is turned on, an additional log line with the seed node configuration (`self.client.transport.node_pool._seed_nodes`) will be logged that has additional information including the `ca_certs` populated amongst others.

As an example, with `service.log_level: DEBUG`, the following is emitted: 
```
[FMWK][06:30:41][INFO] Waiting for Elasticsearch at https://localhost:9200 (so far: 3 secs)
[FMWK][06:30:41][DEBUG] Seed node configuration: (NodeConfig(scheme='https', host='localhost', port=9200, path_prefix='', headers={'user-agent': 'elasticsearch-py/8.12.0 (Python/3.11.6; elastic-transport/8.12.0)'}, connections_per_node=10, request_timeout=10.0, http_compress=False, verify_certs=True, ca_certs='/Users/mhoy/Development/connectors/certificate.pem', client_cert=None, client_key=None, ssl_assert_hostname=None, ssl_assert_fingerprint=None, ssl_version=None, ssl_context=None, ssl_show_warn=True, _extras={}),)
```

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -159,7 +159,9 @@ class ESClient:
             logger.info(
                 f"Waiting for Elasticsearch at {self.configured_host} (so far: {int(time.time() - start)} secs)"
             )
-            logger.debug(f"Seed node configuration: {self.client.transport.node_pool._seed_nodes}")
+            logger.debug(
+                f"Seed node configuration: {self.client.transport.node_pool._seed_nodes}"
+            )
             if await self.ping():
                 return True
             await self._sleeps.sleep(backoff)
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -156,8 +156,9 @@ class ESClient:
                 return False
 
             logger.info(
-                f"Waiting for {self.host} (so far: {int(time.time() - start)} secs)"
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov (approved):
LGTM but left a question - main concern is that we don't leak something that we should not in the line.

markjhoy:
> main concern is that we don't leak something that we should not in the line

Good call. I re-ran it and added an example to the PR description comment... I'm not seeing anything that would be anything sensitive...

artem-shelkovnikov (approved):

PR #2268, opened 2024-03-19
[GitHub] Do not add ACL list for objects from public repos

timgrein:
Objects (PRs, issues, etc.) from public repositories should not contain the `_allow_access_control` field so they're returned for every search, if they match the corresponding criteria. Absence of the `_allow_access_control` field implies that a document is always accessible.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -1791,21 +1791,23 @@ class GitHubDataSource(BaseDataSource):
             async for repo in self._fetch_repos():
                 if self.is_previous_repo(repo["nameWithOwner"]):
                     continue
+
                 access_control = []
-                if self._dls_enabled():
-                    is_public_repo = repo.get("visibility").lower() == "public"
-                    if is_public_repo:
-                        async for user in self.github_client._fetch_all_members():
-                            access_control.append(
-                                _prefix_user_id(user_id=user.get("id"))
-                            )
-                    else:
-                        access_control = await self._fetch_access_control(
-                            repo_name=repo.get("nameWithOwner")
-                        )
-                yield self._decorate_with_access_control(
-                    document=repo, access_control=access_control
-                ), None
+                is_public_repo = repo.get("visibility").lower() == "public"
+                needs_access_control = self._dls_enabled() and not is_public_repo
+
+                if needs_access_control:
+                    access_control = await self._fetch_access_control(
+                        repo_name=repo.get("nameWithOwner")
+                    )
+
+                if needs_access_control:
+                    yield self._decorate_with_access_control(
+                        document=repo, access_control=access_control
+                    ), None
+                else:
+                    yield repo, None
+
                 repo_name = repo.get("nameWithOwner")
                 default_branch = (
                     repo.get("defaultBranchRef", {}).get("name")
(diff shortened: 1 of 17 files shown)


tests/sources/test_github.py
```diff
@@ -1023,14 +1042,73 @@ async def test_fetch_files():
...
-        REPOS,
+        PUBLIC_REPO,
         MOCK_RESPONSE_PULL,
         MOCK_RESPONSE_ISSUE,
         MOCK_RESPONSE_ATTACHMENTS[0],
     ]
     actual_response = []
     async with create_github_source() as source:
-        source._fetch_repos = Mock(return_value=AsyncIterator([REPOS]))
+        source._fetch_repos = Mock(return_value=AsyncIterator([deepcopy(PUBLIC_REPO)]))
```
timgrein, line 1052: using `deepcopy` otherwise the global objects get mutated, which influences other tests

connectors/sources/github.py
```diff
@@ -1791,21 +1791,23 @@ async def get_docs(self, filtering=None):
...
-                ), None
+                is_public_repo = repo.get("visibility").lower() == "public"
+                needs_access_control = self._dls_enabled() and not is_public_repo
+
+                if needs_access_control:
+                    access_control = await self._fetch_access_control(
+                        repo_name=repo.get("nameWithOwner")
+                    )
+
+                if needs_access_control:
```
wangch079, line 1804:
```suggestion
```

This line can be removed

wangch079 (commented):
WDYT if we move the logic to `_decorate_with_access_control`:

First get `access_control` (default is `[]`):

```python
if self._dls_enabled() and not is_public_repo:
	access_control = await self._fetch_access_control(
		repo_name=repo.get("nameWithOwner")
	)
```

Then always call
```python
yield self._decorate_with_access_control(
	document=repo/pr/issue, access_control=access_control
), None
```

Within `_decorate_with_access_control`:

```python
def _decorate_with_access_control(self, document, access_control):
	if self._dls_enabled() and not access_control:
		document[ACCESS_CONTROL] = list(
			set(document.get(ACCESS_CONTROL, []) + access_control)
		)
	return document
```

timgrein:
> WDYT if we move the logic to `_decorate_with_access_control`:
> 
> First get `access_control` (default is `[]`):
> 
> ```python
> if self._dls_enabled() and not is_public_repo:
> 	access_control = await self._fetch_access_control(
> 		repo_name=repo.get("nameWithOwner")
> 	)
> ```
> 
> Then always call
> 
> ```python
> yield self._decorate_with_access_control(
> 	document=repo/pr/issue, access_control=access_control
> ), None
> ```
> 
> Within `_decorate_with_access_control`:
> 
> ```python
> def _decorate_with_access_control(self, document, access_control):
> 	if self._dls_enabled() and not access_control:
> 		document[ACCESS_CONTROL] = list(
> 			set(document.get(ACCESS_CONTROL, []) + access_control)
> 		)
> 	return document
> ```

I don't fully understand the condition `if self._dls_enabled() and not access_control:`. So if `access_control` is falsy (empty list) we would an empty `_allow_access_control` field, which means no one has access to a document. Should it be `self._dls_enabled() and access_control` without `not`? But even then I think it could fail for some edge cases, where the correct access control is that no one has access to a certain object (for example temporarily) as we wouldn't add the field, which would lead to the contrary behavior that everyone has access.

wangch079:
> Should it be `self._dls_enabled() and access_control` without `not` instead?

Yes, that is what I meant

>But even then I think it could fail for some edge cases, where the correct access control is that no one has access to a certain object (empty ACL list).

Although I don't this will happen, this is a valid edge case.

In that case, let's be specific.

wangch079 (approved):

PR #2271, opened 2024-03-20
Don't fail the sync if there are failed operations in the bulk api request

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/5808

This PR makes 2 changes:
1. When there's field `_id` missing in a document, skip the document and log an error message
2. When one or more operations fail in a bulk request, don't fail the whole sync job, log the error message instead.

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -198,7 +198,6 @@ class Sink:
                 for op, data in item.items():
                     if "error" in data:
                         self._logger.error(f"operation {op} failed, {data['error']}")
-                        raise Exception(data["error"]["reason"])
 
         self._populate_stats(stats, res)
 
@@ -360,6 +359,9 @@ class Sink:
                     break
                 operation = doc["_op_type"]
                 doc_id = doc["_id"]
+                if not doc_id:
+                    self._logger.warning(f"Skip document {doc} as '_id' is missing.")
+                    continue
                 if operation == OP_DELETE:
                     stats[operation][doc_id] = 0
                 else:
--- a/tests/test_sink.py
+++ b/tests/test_sink.py
@@ -1208,7 +1208,7 @@ async def test_batch_bulk_with_errors(patch_logger):
             "items": [{"index": {"_id": "1", "error": error}}],
         }
         client.client.bulk = AsyncMock(return_value=mock_result)
-        await sink._batch_bulk([], {OP_INDEX: {"1": 20}, OP_UPSERT: {}, OP_DELETE: {}})
+        await sink._batch_bulk([], {OP_INDEX: {"1": 20}, OP_UPDATE: {}, OP_DELETE: {}})
         patch_logger.assert_present(f"operation index failed, {error}")
 
 
--- a/tests/test_sink.py
+++ b/tests/test_sink.py
@@ -1178,6 +1178,40 @@ async def test_batch_bulk_with_retry():
         assert client.client.bulk.await_count == 2
 
 
+@pytest.mark.asyncio
+async def test_batch_bulk_with_errors(patch_logger):
+    config = {
+        "username": "elastic",
+        "password": "changeme",
+        "host": "http://nowhere.com:9200",
+    }
+    client = ESManagementClient(config)
+    client.client = AsyncMock()
+    sink = Sink(
+        client=client,
+        queue=None,
+        chunk_size=0,
+        pipeline={"name": "pipeline"},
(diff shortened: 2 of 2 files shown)


connectors/es/sink.py
```diff
@@ -164,7 +164,6 @@ async def _bulk_api_call():
                 for op, data in item.items():
                     if "error" in data:
                         self._logger.error(f"operation {op} failed, {data['error']}")
-                        raise Exception(data["error"]["reason"])
```
timgrein, line 167: Should we adapt the log line above to something like this: `self._logger.error(f"Operation: {op} failed. Reason: {data['error']['reason']}. Full error: {data['error']}")`. Could be easier to spot the error reason faster instead of looking through the whole error object

artem-shelkovnikov (commented):
I'm not sure it's a way to do TBH - documents can fail due to a variety of reasons, such as:

1. Mapping is incompatible with document schema
2. Document too big
3. 429

429 should be retried, others not. Just skipping errors breaks our resilience, so we need to parse the reason and retry some, while skipping/logging others.

connectors/es/sink.py
```diff
@@ -164,7 +164,6 @@ async def _bulk_api_call():
                 for op, data in item.items():
                     if "error" in data:
                         self._logger.error(f"operation {op} failed, {data['error']}")
```
seanstory, line 166: I think we should count how often this happens, and at least log out the total at the end of the sync. The customer needs to know of 99% of their documents fail here, especially if we start calling such sync jobs "complete" instead of "error"

wangch079:
>I think we should count how often this happens, and at least log out the total at the end of the sync. The customer needs to know of 99% of their documents fail here, especially if we start calling such sync jobs "complete" instead of "error"

Hi @seanstory , this reminds me of the `monitor` object in the Workplace Search, which tracks the error rate and raises if it reaches a certain threshold.

>429 should be retried, others not. Just skipping errors breaks our resilience, so we need to parse the reason and retry some, while skipping/logging others.

The Bulk API request will be retried for [these response codes](https://github.com/elastic/connectors/blob/f4d65e252bae32d228f0e776e517977f5d2ba5ab/connectors/es/client.py#L204), we yet have the retry logic for a single operation in the Bulk API request.

From the [Response body schema](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-response-body), there is `status` field in each item. We can follow the same [retriable response codes](https://github.com/elastic/connectors/blob/f4d65e252bae32d228f0e776e517977f5d2ba5ab/connectors/es/client.py#L204).


So in conclusion, we need to
1) Add retry logic to operations in Bulk API requests. We need to add a counter to each operation, and only raise if it reaches a threshold or the error is not retriable
2) Add a monitor object (to both extractor and sink), and when an error happens, we call `monitor.note_error`, and let `monitor` decide when to raise an error.

This sounds beyond the scope of this issue, but I just want to make sure we agree on the approach.

seanstory:
> This sounds beyond the scope of this issue, but I just want to make sure we agree on the approach.

works for me.

wangch079:
Hi @artem-shelkovnikov , the PR introduces 2 changes (added in the PR description), and only the first one is related to issue https://github.com/elastic/enterprise-search-team/issues/5808. Let me know if you still have concerns for the second change, and I can exclude it from this PR

artem-shelkovnikov (approved):

artem-shelkovnikov:
> Hi @artem-shelkovnikov , the PR introduces 2 changes (added in the PR description), and only the first one is related to issue https://github.com/elastic/enterprise-search-team/issues/5808. Let me know if you still have concerns for the second change, and I can exclude it from this PR

Just to check, I see that `raise` was removed, but where does it get logged? What happens here, do we retry the batch if we have errors? Do we log all errors?

wangch079:
> Just to check, I see that `raise` was removed, but where does it get logged?

It's logged [here](https://github.com/elastic/connectors/pull/2271/files#diff-68c31d9379d91378705cf6d47c447185e8c6ad952876c7a5fb979e3c89646148R200)

>What happens here, do we retry the batch if we have errors?

No, the current retry is on Bulk API request, i.e. only if the Bulk API fails entirely. Retry on a single operation is not in scope of this PR, see discussion here: https://github.com/elastic/connectors/pull/2271#issuecomment-2011320554

PR #2273, opened 2024-03-20
Wait for sync jobs to finish before closing clients

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/2167

The problem started happening due to recent fix to actually close ESClient instances properly.

The problem in short happened because the instances of ESClient were closed while SyncJobRunner was attempting to suspend the sync - to suspend the sync it needed to load the job record from ES, then update it, then do some more small calls to Elasticsearch. It failed because the client was already closed by the time these calls were made.

The change now ensures, that all tasks in ConcurrentTasks finish before actually closing ESClients.

#### Pre-Review Checklist

## Release Note

Fixed connectors not properly suspending jobs on host being forcefully stopped.

### DIFF
--- a/connectors/services/job_execution.py
+++ b/connectors/services/job_execution.py
@@ -148,6 +148,7 @@ class JobExecutionService(BaseService):
                     break
                 await self._sleeps.sleep(self.idling)
         finally:
+            await self.sync_job_pool.join()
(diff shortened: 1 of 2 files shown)


vidok (approved):
Looks good to me!

PR #2276, opened 2024-03-20
Eliminate Duplicate Buildkite Builds on Branches and PRs

markjhoy:
A small CI related change to reduce the number of duplicated Buildkite jobs. This:

1) sets `"build_on_commit": false` in the `pull-requests.json` to allow the webhook settings to determine when to build
2) sets `skip_pull_request_builds_for_existing_commits: false` in the `catalog-info.yaml` file to ensure new builds are created only when there's a new commit (you can still run the comment commands to rebuild that will flow through the other webhook)

#### Pre-Review Checklist

### DIFF
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -8,7 +8,7 @@
       "allowed_list": ["praveen-elastic", "moxarth-elastic", "khusbu-crest", "akanshi-crest"],
       "set_commit_status": true,
       "commit_status_context": "buildkite/connectors",
-      "build_on_commit": true,
+      "build_on_commit": false,
       "build_on_comment": true,
       "trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:build|test)\\W+(?:this|it))",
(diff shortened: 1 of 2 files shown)


acrewdson (approved):
> sets `skip_pull_request_builds_for_existing_commits: false` in the catalog-info.yaml file to ensure new builds are created only when there's a new commit (you can still run the comment commands to rebuild that will flow through the other webhook)

maybe I'm misreading, but I would have thought we'd want that to be `true` to get the outcome you describe (also assuming [these](https://registry.terraform.io/providers/buildkite/buildkite/latest/docs/resources/pipeline#skip_pull_request_builds_for_existing_commits) are the official docs for that option) 

but if it's what fixed things platfrom, let's give it a shot 👍

markjhoy:
> but if it's what fixed things platform, let's give it a shot

I agree it seems backwards, but if anything it shouldn't break anything. Also - in other repos internally, it's set to false as well... going to merge this and give it a try. We should see by tomorrow morning if there's anything amiss or if it seems to be working. If not, we can certainly revert this.

PR #2280, opened 2024-03-21
gitignore config.yml

seanstory:
## fixes https://github.com/elastic/connectors/issues/2205

Moves `config.yml` to `config.yml.example`, and modifies our Makefile so that `make install` (a required setup step) will copy `config.yml.example` to `config.yml` if it doesn't exist yet.

Further, it adds `config.yml` to the gitignore.

Customers shouldn't even notice that this change took place, but it should help us avoid situations where users accidentally commit/push secrets.

#### Pre-Review Checklist
  - docs PR: https://github.com/elastic/enterprise-search-pubs/pull/4302


docs/CONFIG.md
```diff
@@ -1,6 +1,8 @@
 # Configuration
 
-Configuration lives in [config.yml](../config.yml).
+Configuration lives in `config.yml`.
+This file is created when you first run `make install`, and is "git ignored" to help prevent committing secrets.
+The default configs can all be see in the [`config.yml.example`]((../config.yml.example).), which is not intended to be modified.
```
seanstory, line 5:
```suggestion
The default configs can all be seen in the [`config.yml.example`]((../config.yml.example).), which is not intended to be modified.
```

wangch079 (approved):
Nice work

wangch079:
I think one of the tests and our ftest assume the existence of `config.yml`

seanstory:
Yep, found the same. I thought that `install` was a prereq for our tests, but looks like it's not. I'll fix it.

artem-shelkovnikov:
I think tests also should not try to use real live config.yml - we already have tests that use one in `tests/` directory, might be worth to just update them?

seanstory:
Testing locally, if I:
1. `git checkout main`
2. `git checkout -b test`
3. make and commit changes to `config.yml` (pretending to be a careless developer on a fork/clone)
4. `git pull origin seanstory/2205-gitignore-config-yml` (pretending to be upgrading my clone/fork to latest, which includes the changes from this PR)
5. I see my changes from `config.yml` are now in `config.yml.example`.
6. If I create a PR at this point, I've exposed whatever changes I made, potentially leaking secrets.

I _think_ this is desired behavior, because it doesn't lose changes that I made. But I wanted to make explicit that this PR doesn't prevent scenarios where people might still accidentally push up a PR that contains secrets - it just hopefully makes it less likely _going forward_.

leemthompo:
`Random thought`: Would this require any updates to Mark's docker compose scripts?

seanstory:
> Random thought: Would this require any updates to Mark's docker compose scripts?

I had the same thought, and from manual testing looks like "no". And that's because the scripts, under the hood, use `make` too. https://github.com/elastic/connectors/blob/main/scripts/stack/configure-connectors.sh#L47

wangch079 (approved):

Makefile
```diff
@@ -7,7 +7,10 @@ SLOW_TEST_THRESHOLD=1 # seconds
 VERSION=$(shell cat connectors/VERSION)
 
 
-bin/python:
+config:
```
artem-shelkovnikov, line 10: You might wanna change it to `config.yml` - this way the command will not run if `config.yml` is already in-place

seanstory: Oh interesting, is that a Makefile feature? I'll admit that I don't know a lot about `make`. 

It already doesn't overwrite the file if it exists. The command is:
```
- cp -n config.yml.example config.yml
```

the `-n` is "no clobber" which means that if `config.yml` exists, terminate. Unfortunately it terminates with an error, so the `-` before `cp` tells make to ignore errors in this step and proceed.

I've tested pretty extensively to ensure that the config file doesn't get overwritten.

artem-shelkovnikov: Yup, dependencies in Makefiles are actually dependencies on files, like here: https://www.tutorialspoint.com/makefile/makefile_dependencies.htm

It's not a biggie not to do, we can fix in subsequent prs :)

PR #2286, opened 2024-03-22
[Network Drive] Change traversal strategy to ingest as you go instead of pre-fetching the whole tree

akanshi-elastic:
## Closes https://github.com/elastic/connectors/issues/2265###
This PR contains the traversal of directories/subdirectory changes, now we are using scandir method to fetch data from network drive, which start ingestion data in seconds after connector starts syncing.

#### Pre-Review Checklist


artem-shelkovnikov:
Can you test it against a reasonably large network drive (large enough so that without this change, walking through it would take at least 5 minutes) and provide benchmarks for the test?

Some numbers that are interesting to me:
- Time to sync all the data
- Time until first document lands into the index
- Perf8 reports before/after

praveen-elastic:
> Can you test it against a reasonably large network drive (large enough so that without this change, walking through it would take at least 5 minutes) and provide benchmarks for the test?
> 
> Some numbers that are interesting to me:
> 
> * Time to sync all the data
> * Time until first document lands into the index
> * Perf8 reports before/after

Hi [@artem-shelkovnikov](https://github.com/artem-shelkovnikov) Please find the response as below:
Time taken to sync all the data - 1 hour 47 mins to Index 152416 docs
Time until first document lands into the index - 1 sec
Link to Perf8 reports:- https://drive.google.com/drive/folders/1xtT4TibPH-Y0rHegUBwD9rkAtlW6KsQD?usp=drive_link

artem-shelkovnikov:
RSS graph looks concerning - it seems to never recycle the memory, can you check if you're releasing memory during traversal?

See graph before the change - you can see that RSS decreases from time to tim: 

image

praveen-elastic:
> RSS graph looks concerning - it seems to never recycle the memory, can you check if you're releasing memory during traversal?
> 
> See graph before the change - you can see that RSS decreases from time to time:

You are right, this does seem concerning, I have found a resolution, currently testing that, will update that once done.

praveen-elastic:
I've now changed the code to use iterative approach which has improved the rss and also fixed the loop_lag issue, attaching the two zips 
[perf8-report-network_drive_recursive.zip](https://github.com/elastic/connectors/files/14773659/perf8-report-network_drive_recursive.zip)
[perf8-report-network_drive_iterative.zip](https://github.com/elastic/connectors/files/14773661/perf8-report-network_drive_iterative.zip)

artem-shelkovnikov:
@praveen-elastic are the results received from ftest?

Where did the previous result come from, with more data? Is it possible to also check there? The charts that you've attached look okay, but I'm not sure why the results are the way they are:

- RSS is recycled in both
- Why did event loop lag decrease?
- Why did run time decrease?

praveen-elastic:
yes, these are the stats of ftests, the previous one was with actual connector sync with a lot of data. 

I did run the connector with that data for the updated code but didn't capture perf8 report, the total time to index was reduced to around: `1h 42 mins`(earlier approach took 1 hr 47 minutes). I'll try to run the connector again with that data tomorrow to generate the perf8 stats.

By the way, I shared the ftest stats as the RSS fluctuations in new code were lot less and I thought loop lap reduction was a good thing. Also the disk usage was also reduced from 11 Gb to 3 Gb.

artem-shelkovnikov:
It indeed looks good, but disk usage reduction looks too good to be true - is it possible we're missing some data?

Would be great to compare the data (index size + number of records) for both old/new implementation for the real data source you used before in the report

praveen-elastic:
> It indeed looks good, but disk usage reduction looks too good to be true - is it possible we're missing some data?
> 
> Would be great to compare the data (index size + number of records) for both old/new implementation for the real data source you used before in the report

Hi @artem-shelkovnikov here are the perf8 stats for connector sync with the updated code and the same data 
[perf8.zip](https://github.com/elastic/connectors/files/14788224/perf8.zip), rss seems to be recycled in this one.

artem-shelkovnikov:
Still seems to grow: 
image


Just to compare, how does this graph looks for sync to same data source, but from `main`, not from this branch?

praveen-elastic:
Hmm, let me generate perf8 for the code on main currently.

praveen-elastic:
@artem-shelkovnikov here's the report of code from main branch with same data 
[perf8-main-branch-code.zip](https://github.com/elastic/connectors/files/14792899/perf8-main-branch-code.zip)

artem-shelkovnikov (approved):
Current branch RSS looks similar, thus it should be good to merge, great job!

sphilipse:
@artem-shelkovnikov do we want to backport this to 8.13 still?

artem-shelkovnikov:
@sphilipse yup I'm planning to do a backport after we release 8.13.1, so hopefully tomorrow

Issue #1958, opened 2023-12-07
Notion connector client (8.13 Beta)


danajuratoni:
## New Notion connector
New connector client to be released with 8.13 in Beta.
Preliminary date for **8.13 Feature Freeze**:  Tue, Feb 06 2024

### Acceptance criteria
- [x] TDD in [GDrive](https://drive.google.com/drive/folders/1DN2y4TRXIuybX8vawHljkWozoyPEOYaf)
- [x] QA plan in [GDrive](https://drive.google.com/drive/folders/1DN2y4TRXIuybX8vawHljkWozoyPEOYaf)
- [x] Advanced Sync rules
- [x] Code merged in main following the defined PR process
- [x] Functional tests performed
- [x] Performance test/ reports against the last stable release branch of the connector framework and in Elastic Cloud 
     - Performance Stats Report in [GDrive](https://docs.google.com/document/d/1mYxwCliFTmG0M-CUJjxsfLzr0Pf5-Pb_/edit?usp=sharing&ouid=101132896583575395288&rtpof=true&sd=true)
- [x] Connector client documentation in repo 
- [x] Demo of the connector in latest builds in Elastic Cloud in [GDrive](https://drive.google.com/drive/folders/1DN2y4TRXIuybX8vawHljkWozoyPEOYaf)
- [x]  Ingestion engineer added a comment to this ticket to sign off
     - [x] Code backported to the correct version 
     - [x] Automated tests to 90% coverage
- [x] Regression test suite performed for the release along with QA day for all supported OS in https://www.elastic.co/support/matrix 
- [x] Compatibility tests for running connector as native in Cloud

praveen-elastic:
Progress Summary:

- Started analysis of Notion as a Source.
- Prepared the Design Document and shared it with the Team.

praveen-elastic:
Progress Summary:
- Addressed Comments in the TDD 
- Started Ping, Configuration validation and Indexing implementation.

praveen-elastic:
Progress Summary:
- Raised a PR for indexing functionality

praveen-elastic:
Progress Summary:
- Resolved comments for indexing functionality and the PR is merged.
- Started working on E2E.

praveen-elastic:
Mistakenly closed due to merging of PR

praveen-elastic:
Progress Summary: 
- PR for advanced sync rules got merged. 
[...]

PR #2297, opened 2024-03-25
[Notion] Implement an ability to retry for failed paginated api call

praveen-elastic:
## Part of #1958 

During performance testing, it was reported that the [notion sdk module](https://github.com/ramnes/notion-sdk-py/blob/d5e5f5c98ae5578ce4b0130b65ad6e8f9e92e02b/notion_client/helpers.py#L58) does not provision any retry mechanism for a paginated api call made via `async_iterate_paginated_api`. 

This PR attempts to implement `async_iterate_paginated_api` with an ability to retry a failed page. 

Simply decorating the module's helper method [async_iterate_paginated_api](https://github.com/ramnes/notion-sdk-py/blob/d5e5f5c98ae5578ce4b0130b65ad6e8f9e92e02b/notion_client/helpers.py#L58) with retryable decorator won't work as that would mean any error thrown by page k (k >1) would result in retrying from page 1 instead of retry on page k.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/notion.py
+++ b/connectors/sources/notion.py
@@ -8,13 +8,13 @@ import asyncio
 import os
 from copy import copy
 from functools import cached_property, partial
+from typing import Any, Awaitable, Callable
 from urllib.parse import unquote
 
 import aiohttp
 import fastjsonschema
 from aiohttp.client_exceptions import ClientResponseError
 from notion_client import APIResponseError, AsyncClient
-from notion_client.helpers import async_iterate_paginated_api
 
 from connectors.filtering.validation import (
     AdvancedRulesValidator,
@@ -28,7 +28,7 @@ RETRIES = 3
 RETRY_INTERVAL = 2
 DEFAULT_RETRY_SECONDS = 30
 BASE_URL = "https://api.notion.com"
-MAX_CONCURRENT_CLIENT_SUPPORT = 3
+MAX_CONCURRENT_CLIENT_SUPPORT = 30
 
 
 if "OVERRIDE_URL" in os.environ:
@@ -96,6 +96,48 @@ class NotionClient:
             else:
                 raise
 
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+        skipped_exceptions=NotFound,
+    )
+    async def fetch_results(
+        self, function: Callable[..., Awaitable[Any]], next_cursor=None, **kwargs: Any
+    ):
+        try:
+            return await function(start_cursor=next_cursor, **kwargs)
+        except Exception as exception:
+            if exception.code == "rate_limited" and exception.status == 429:
+                retry_after = (
+                    exception.headers.get("retry-after") or DEFAULT_RETRY_SECONDS
+                )
+                self._logger.info(
+                    f"Connector will attempt to retry after {int(retry_after)} seconds."
+                )
+                await self._sleeps.sleep(int(retry_after))
+                msg = "Rate limit exceeded."
+                raise Exception(msg) from exception
+            else:
+                raise
+
+    async def async_iterate_paginated_api(
+        self, function: Callable[..., Awaitable[Any]], **kwargs: Any
+    ):
+        """Return an async iterator over the results of any paginated Notion API."""
+        next_cursor = kwargs.pop("start_cursor", None)
+        while True:
+            try:
+                response = await self.fetch_results(function, next_cursor, **kwargs)
+                if response:
+                    for result in response.get("results"):
+                        yield result
+                    next_cursor = response.get("next_cursor")
+                    if not response["has_more"] or next_cursor is None:
+                        return
+            except Exception as error:
+                raise error
+
     async def fetch_owner(self):
         """Fetch integration authorized owner"""
         await self._get_client.users.me()
@@ -111,7 +153,7 @@ class NotionClient:
         """Iterate over user information retrieved from the API.
         Yields:
         dict: User document information excluding bots."""
-        async for user_document in async_iterate_paginated_api(
+        async for user_document in self.async_iterate_paginated_api(
             self._get_client.users.list
         ):
             if user_document.get("type") != "bot":
@@ -126,7 +168,7 @@ class NotionClient:
 
         async def fetch_children_recursively(block):
             if block.get("has_children") is True:
-                async for child_block in async_iterate_paginated_api(
+                async for child_block in self.async_iterate_paginated_api(
                     self._get_client.blocks.children.list, block_id=block.get("id")
                 ):
                     yield child_block
@@ -136,7 +178,7 @@ class NotionClient:
                     ):  # pyright: ignore
                         yield grandchild
 
-        async for block in async_iterate_paginated_api(
+        async for block in self.async_iterate_paginated_api(
             self._get_client.blocks.children.list, block_id=block_id
         ):
             if block.get("type") not in ["child_database", "child_page", "unsupported"]:
@@ -149,7 +191,7 @@ class NotionClient:
                     yield record
 
     async def fetch_by_query(self, query):
-        async for document in async_iterate_paginated_api(
+        async for document in self.async_iterate_paginated_api(
             self._get_client.search, **query
         ):
             yield document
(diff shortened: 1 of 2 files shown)


connectors/sources/notion.py
```diff
@@ -96,6 +96,48 @@ async def get_via_session(self, url):
...
+        self, function: Callable[..., Awaitable[Any]], next_cursor=None, **kwargs: Any
+    ):
+        try:
+            return await function(start_cursor=next_cursor, **kwargs)
+        except Exception as exception:
+            if exception.code == "rate_limited" and exception.status == 429:
+                retry_after = (
+                    exception.headers.get("retry-after") or DEFAULT_RETRY_SECONDS
+                )
+                self._logger.info(
```
seanstory, line 115: let's include in the logs what the request was. I assume that's captured in the `next_cursor` or the `kwargs`, but capturing the `function`'s name might be good too.

connectors/sources/notion.py
```diff
@@ -96,6 +96,48 @@ async def get_via_session(self, url):
...
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+        skipped_exceptions=NotFound,
+    )
+    async def fetch_results(
+        self, function: Callable[..., Awaitable[Any]], next_cursor=None, **kwargs: Any
+    ):
+        try:
+            return await function(start_cursor=next_cursor, **kwargs)
+        except Exception as exception:
+            if exception.code == "rate_limited" and exception.status == 429:
```
seanstory, line 111:
```suggestion
            if exception.code == "rate_limited" or exception.status == 429:
```

connectors/sources/notion.py
```diff
@@ -96,6 +96,48 @@ async def get_via_session(self, url):
...
                 raise
 
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+        skipped_exceptions=NotFound,
+    )
+    async def fetch_results(
+        self, function: Callable[..., Awaitable[Any]], next_cursor=None, **kwargs: Any
```
seanstory, line 106: why is `next_cursor` an explicit arg, instead of being included in `kwargs`? Below, it looks like this starts as 
```
kwargs.pop("start_cursor", None)
```
so I don't see the point.

connectors/sources/notion.py
```diff
@@ -96,6 +96,48 @@ async def get_via_session(self, url):
...
+            try:
+                response = await self.fetch_results(function, next_cursor, **kwargs)
+                if response:
+                    for result in response.get("results"):
+                        yield result
+                    next_cursor = response.get("next_cursor")
+                    if not response["has_more"] or next_cursor is None:
+                        return
+            except Exception as error:
+                raise error
```
seanstory, line 139: This seems entirely redundant. Unless you're going to handle `error` or log something, why bother catching it just to re-raise?

tests/sources/test_notion.py
```diff
@@ -740,13 +759,37 @@ async def test_advanced_rules_validation(advanced_rules, expected_validation_res
...
         ):
             validation_result = await NotionAdvancedRulesValidator(source).validate(
                 advanced_rules
             )
-        assert validation_result == expected_validation_result
+            assert validation_result == expected_validation_result
+
+
+@pytest.mark.asyncio
+async def test_async_iterate_paginated_api():
```
seanstory, line 776: Can you add tests to ensure that 429's are retried, but other errors are not? Perhaps also add a test to make sure that the original `async_iterate_paginated_api` is not being called anywhere?

connectors/sources/notion.py
```diff
@@ -28,7 +28,7 @@
 RETRY_INTERVAL = 2
 DEFAULT_RETRY_SECONDS = 30
 BASE_URL = "https://api.notion.com"
-MAX_CONCURRENT_CLIENT_SUPPORT = 3
+MAX_CONCURRENT_CLIENT_SUPPORT = 30
```
seanstory, line 31: This seems like a large change and looks unrelated. What's the rationale?

connectors/sources/notion.py
```diff
@@ -423,7 +465,7 @@ def get_default_configuration(cls):
                 "value": False,
             },
             "concurrent_downloads": {
-                "default_value": 20,
+                "default_value": 30,
```
seanstory, line 468: Why's this being changed?

connectors/sources/notion.py
praveen-elastic: the `next_cursor` passed to `fetch_results` method is only equal to `start_cursor` during 1st iteration, hence that can be fetched via `kwargs.pop("start_cursor", None)`. 

But in the subsequent iterations of while loop, the `next_cursor` is available in the response of previous call. Hence, we cannot rely on `kwargs` for these iterations. Therefore we need to pass that explicitly along with `kwargs`

tests/sources/test_notion.py
praveen-elastic: Updated via commit id: [2a601f1](https://github.com/elastic/connectors/pull/2297/commits/2a601f1935fd829667106e11af1621eac2c54482)

connectors/sources/notion.py
praveen-elastic: The downloading of the attachments happens via `aiohttp` session using public urls which expire in 1 hour, also those aren't restricted by Notion's rate limit of 3, so at a time more connections can be done for downloading files.

Hence following changes were made for `MAX_CONCURRENT_CLIENT_SUPPORT` and `concurrent_downloads` that is specifically used for downloading of files only.

Should I raise a separate PR for this ?

connectors/sources/notion.py
praveen-elastic: same as the reason above, it was changed for file downloads.

connectors/sources/notion.py
seanstory: No, I'm fine with you including small fixes in larger PRs. In the future, just try to call them out either with code comments or in the PR description, so it's clear what's being changed and why. 👍

tests/sources/test_notion.py
```diff
@@ -740,13 +759,85 @@ async def test_advanced_rules_validation(advanced_rules, expected_validation_res
...
+        assert exc_info.value.code == "object_not_found"
+        assert exc_info.value.status == 404
+
+
+@pytest.mark.asyncio
+async def test_original_async_iterate_paginated_api_not_called():
+    with patch(
+        "notion_client.helpers.async_iterate_paginated_api"
+    ) as mock_async_iterate_paginated_api:
+        assert not mock_async_iterate_paginated_api.called
```
seanstory, line 843: This isn't testing anything. You'd need to have the sync actually run inside the `with patch`.

tests/sources/test_notion.py
```diff
@@ -740,13 +759,85 @@ async def test_advanced_rules_validation(advanced_rules, expected_validation_res
...
+            message="Rate limit exceeded.",
+            response=Response(status_code=429, text="Rate limit exceeded."),
+        )
+
+    async with create_source(
+        NotionDataSource, notion_secret_key="secert_key"
+    ) as source:
+        with patch("connectors.sources.notion.DEFAULT_RETRY_SECONDS", 0.3):
+            with pytest.raises(Exception) as exc_info:
+                await source.notion_client.fetch_results(mock_function_with_429)
```
seanstory, line 813: Can you adapt this to show that it raises if it keeps getting the 429, but explicitly _does not_ raise an error if it 429s once, but then responds successfully after the retry? For example see https://github.com/elastic/connectors/blob/main/tests/sources/test_sharepoint_online.py#L571

tests/sources/test_notion.py
```diff
@@ -793,3 +793,51 @@ async def mock_function(**kwargs):
...
+
+    async with create_source(
+        NotionDataSource, notion_secret_key="secert_key"
+    ) as source:
+        with pytest.raises(APIResponseError) as exc_info:
+            await source.notion_client.fetch_results(mock_function_with_other_error)
+
+        assert exc_info.value.code == "object_not_found"
+        assert exc_info.value.status == 404
+
```
seanstory, line 836: I don't think this is currently asserting a lack of retries. Can you assert how many times the request is actually made to show that a retry is not occurring?

seanstory (approved):

PR #2308, opened 2024-03-29
Fix pagination for confluence datacenter sync rules

akanshi-elastic:
## Closes https://github.com/elastic/connectors/issues/1957###

This PR adding approach for confluence datacenter pagination for sync rules

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -36,6 +36,7 @@ from connectors.utils import (
 )
 
 RETRY_INTERVAL = 2
+LIMIT = 100
 SPACE = "space"
 SPACE_PERMISSION = "space_permission"
 BLOGPOST = "blogpost"
@@ -230,9 +231,9 @@ class ConfluenceClient:
                     yield json_response
 
                     start = url_kwargs.get("start", 0)
-                    start += 100
+                    start += LIMIT
                     url_kwargs["start"] = start
-                if len(json_response.get("results", [])) < 100:
+                if len(json_response.get("results", [])) < LIMIT:
                     break
             except Exception as exception:
                 self._logger.warning(
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -46,6 +46,7 @@ DOWNLOAD = "download"
 SEARCH = "search"
 USER = "user"
 USERS_FOR_DATA_CENTER = "users_for_data_center"
+SEARCH_FOR_DATA_CENTER = "search_for_data_center"
 USERS_FOR_SERVER = "users_for_server"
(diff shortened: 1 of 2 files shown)


connectors/sources/confluence.py
```diff
@@ -210,6 +212,34 @@ async def paginated_api_call(self, url_name, **url_kwargs):
                 )
                 break
 
+    async def paginated_api_call_for_datacenter_syncrule(self, url_name, **url_kwargs):
```
artem-shelkovnikov, line 215: What does `_syncrule` suffix stand for here?

akanshi-elastic: I have kept it like this because method is specific to to datacenter sync rule only.

connectors/sources/confluence.py
```diff
@@ -210,6 +212,34 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+            json_response = {}
+            try:
+                async for response in self.api_call(
+                    url=url,
+                ):
+                    json_response = await response.json()
+                    yield json_response
+
+                    start = url_kwargs.get("start", 0)
+                    start += 100
```
artem-shelkovnikov, line 233: 100 should be a constant

connectors/sources/confluence.py
```diff
@@ -210,6 +212,34 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+                    url_kwargs["start"] = start
+                if len(json_response.get("results", [])) < 100:
+                    break
+            except Exception as exception:
+                self._logger.warning(
+                    f"Skipping data for type {url_name} from {url}. Exception: {exception}."
+                )
+                break
```
artem-shelkovnikov, line 241: Why not reraise here? It can make it look like sync is successful, but actually fail

akanshi-elastic: If I raise it here then sync will be stop here and not work for other queries, issues etc. We are raising the error after retrying in api_call method

artem-shelkovnikov: What if every resource raises an error, will it result in successful sync?

> We are raising the error after retrying in api_call method
But I see retry relies on the method raising an error, so if you don't raise you don't retry?

akanshi-elastic: I guess this is the worst case that every resource raise error, even in such condition we'll skip all the resources and show successful sync message. We are following same approach for every connector right now. 
Also if api call throw any error in that case only we'll retry 3 times and raise the error with failure message on the terminal.

artem-shelkovnikov (approved):
We'll address the problem with potentially ignoring problems if remote endpoints raise too much later

PR #2312, opened 2024-03-29
[Bugfix][Notion] Objects not shared with Integration or not found should not fail the sync

praveen-elastic:
## Closes https://github.com/elastic/connectors-py/issues/2307

If an object is not shared with integration or if the object is not found, the API response returns code `object_not_found`. This status code can be handled to avoid the sync to fail. 

image

#### Pre-Review Checklist


connectors/sources/notion.py
```diff
@@ -136,17 +136,28 @@ async def fetch_children_recursively(block):
...
+                ]:
+                    yield block
+                    if block.get("has_children") is True:
+                        async for child in fetch_children_recursively(block):
+                            yield child
+                if block.get("type") == "child_database":
+                    async for record in self.query_database(block.get("id")):
+                        yield record
+        except APIResponseError as error:
+            if error.code == "object_not_found":
```
seanstory, line 156: Do we know what these objects are? In the linked issue, Davey expressed confusion about the ID that wasn't found, because he thought that he'd shared access to everything in his account. The ID in the error didn't mean anything to him and he doesn't know what that database is.

Are we potentially trying to access internal objects that we shouldn't even be trying to fetch in the first place?

praveen-elastic: In case of Davey, it was a `linked database` (a copy of another database which can have its own views, filters, and sorts). Notion docs mentions that it is not supported by API and to fetch the information in a linked database, add integration only to the original source database and not the linked one https://developers.notion.com/docs/working-with-databases#linked-databases. 

In other cases, it can also be an object that was deleted while the connector service was trying fetch its children using Notion API.

> Are we potentially trying to access internal objects that we shouldn't even be trying to fetch in the first place?

We shouldn't try to query the details of linked database. However, from the API response, it cannot be inferred whether the type of database is a linked one or not, so we can't decide which database needs to be queried in the first place.

seanstory: >  it was a linked database (a copy of another database which can have its own views, filters, and sorts). Notion docs mentions that it is not supported by API 

Gotcha. @praveen-elastic can you work with @leemthompo to get this captured in a "known issues" for the notion connector? I could imagine other folks hitting this.

> However, from the API response, it cannot be inferred whether the type of database is a linked one or not, so we can't decide which database needs to be queried in the first place.

That's unfortunate. I am curious though if the linked_database record has `has_children: true`? I'm noticing that while `fetch_children_recursively` checks for that, we don't check here before calling 
```
            async for block in async_iterate_paginated_api(
                self._get_client.blocks.children.list, block_id=block_id
            ):
```

If it does indicate it has children and there's no other way to identify that this is a linked_database, can you file a bug with Notion and link it here? We should be able to defend against a mis-configured integration without having to just ignore all errors, I think.

praveen-elastic: > Gotcha. @praveen-elastic can you work with @leemthompo to get this captured in a "known issues" for the notion connector? I could imagine other folks hitting this.

Sure.

> I am curious though if the linked_database record has has_children: true? If it does indicate it has children and there's no other way to identify that this is a linked_database, can you file a bug with Notion and link it here?

The error is raised by the method `query_database` which is called for all the blocks that are of type `child_database` from the list of blocks returned by `blocks.children.list`. This `child_database` can either be an inline database or a linked one but there is no field that can differentiate the two. 
To submit an issue we need to mail them. Should I mail the Notion team and request them to add a field that can differentiate the database is linked or not ?

seanstory: Sounds good, yes, thanks for mailing their support. I'll watch the thread, but will assume you'll reply back-and-forth with them.

Let's leave this PR pending for a few days while that discussion happens. If it stalls, we can merge this as-is, but I'm hopeful that they'll suggest a workaround for us.

praveen-elastic: Hi @seanstory It's been a few days since we last heard back from the Notion Team. Can we please merge this PR? If they provide a solution in the future, we'll be happy to accommodate it.

leemthompo: FYI docs PR for original issues flagged is open: https://github.com/elastic/enterprise-search-pubs/pull/4345

seanstory: Can we narrow this case any? Anything in the error response that makes it clear that it's a linked_database and not some other sort of 404?

praveen-elastic: I'm afraid we can't narrow this case, the error message returned is: `Could not find database with ID: be907abe-510e-4116-a3d1-7ea71018c06f. Make sure the relevant pages and databases are shared with your integration.` The same error message can also be returned if any other type database (inline/full-page) is deleted during a sync.

seanstory: dang. Ok. I'm out of ideas.

seanstory (approved):

PR #2323, opened 2024-04-01
Log all counters

seanstory:
## Closes https://github.com/elastic/connectors/issues/2314

Adds some logging on sync completion that prints out every metric that we have in the orchestrator's `ingestion_stats` dict. This means that going forward, all we have to do is add something to `ingestion_stats` in order to expose it from the logs.

The new output looks like:
```
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] --- Counters ---
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'attachment_extracted' : 0
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'bulk_operations' : {'index': 21}
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'deleted_document_count' : 0
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'doc_created' : 21
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'doc_deleted' : 0
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'doc_updated' : 0
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'indexed_document_count' : 21
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] 'indexed_document_volume' : 0
[FMWK][10:30:03][INFO] [Connector id: eV5Fmo4BAQ4Y7gVESEBV, index name: slack, Sync job id: fF5Hmo4BAQ4Y7gVES0Dr] ----------------
```

#### Pre-Review Checklist

## Release Note

Connector sync jobs now log more counters at their conclusion.

### DIFF
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -335,31 +335,31 @@ class SyncJobRunner:
             except asyncio.CancelledError:
                 self.sync_job.log_debug("Job reporting task is stopped.")
 
-        result = (
+        ingestion_stats = (
             {}
             if self.sync_orchestrator is None
             else self.sync_orchestrator.ingestion_stats()
         )
-        ingestion_stats = {
-            "indexed_document_count": result.get("indexed_document_count", 0),
-            "indexed_document_volume": result.get("indexed_document_volume", 0),
-            "deleted_document_count": result.get("deleted_document_count", 0),
+        persisted_stats = {
+            "indexed_document_count": ingestion_stats.get("indexed_document_count", 0),
+            "indexed_document_volume": ingestion_stats.get("indexed_document_volume", 0),
+            "deleted_document_count": ingestion_stats.get("deleted_document_count", 0),
         }
 
         if await self.reload_sync_job():
             if await self.reload_connector():
-                ingestion_stats[
+                persisted_stats[
                     "total_document_count"
                 ] = await self.connector.document_count()
 
             if sync_status == JobStatus.ERROR:
-                await self.sync_job.fail(sync_error, ingestion_stats=ingestion_stats)
+                await self.sync_job.fail(sync_error, ingestion_stats=persisted_stats)
             elif sync_status == JobStatus.SUSPENDED:
-                await self.sync_job.suspend(ingestion_stats=ingestion_stats)
+                await self.sync_job.suspend(ingestion_stats=persisted_stats)
             elif sync_status == JobStatus.CANCELED:
-                await self.sync_job.cancel(ingestion_stats=ingestion_stats)
+                await self.sync_job.cancel(ingestion_stats=persisted_stats)
             else:
-                await self.sync_job.done(ingestion_stats=ingestion_stats)
+                await self.sync_job.done(ingestion_stats=persisted_stats)
 
         if await self.reload_connector():
             sync_cursor = (
@@ -374,11 +374,15 @@ class SyncJobRunner:
 
         self.sync_job.log_info(
             f"Sync ended with status {sync_status.value} -- "
-            f"created: {result.get('doc_created', 0)} | "
-            f"updated: {result.get('doc_updated', 0)} | "
-            f"deleted: {result.get('doc_deleted', 0)} "
+            f"created: {ingestion_stats.get('doc_created', 0)} | "
+            f"updated: {ingestion_stats.get('doc_updated', 0)} | "
+            f"deleted: {ingestion_stats.get('doc_deleted', 0)} "
             f"(took {int(time.time() - self._start_time)} seconds)"  # pyright: ignore
         )
+        self.sync_job.log_info("--- Counters ---")
+        for k, v in sorted(ingestion_stats.items()):
+            self.sync_job.log_info(f"'{k}' : {v}")
+        self.sync_job.log_info("----------------")
 
     @with_concurrency_control()
     async def sync_starts(self):
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -342,7 +342,9 @@ class SyncJobRunner:
         )
         persisted_stats = {
             "indexed_document_count": ingestion_stats.get("indexed_document_count", 0),
-            "indexed_document_volume": ingestion_stats.get("indexed_document_volume", 0),
+            "indexed_document_volume": ingestion_stats.get(
+                "indexed_document_volume", 0
+            ),
             "deleted_document_count": ingestion_stats.get("deleted_document_count", 0),
         }
 
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -381,9 +381,24 @@ class SyncJobRunner:
             f"deleted: {ingestion_stats.get('doc_deleted', 0)} "
             f"(took {int(time.time() - self._start_time)} seconds)"  # pyright: ignore
         )
+        self.log_counters(ingestion_stats)
+
+    def log_counters(self, counters):
+        """
+        Logs out a dump of everything in "counters"
+
+        This serves a dual-purpose:
+        1. Providing human-readable counts on separate lines, which can be helpful for
+            visualizing how a counter changes over multiple runs (like in Discover)
+        2. Providing a machine-readable hash of all counters, that can be easily copy-pasted
+           into a REPL or script for analysis.
+        See more of the discussion: https://github.com/elastic/connectors/pull/2323
+        :param counters: a dictionary of counter_name -> counter_value like: {"added": 2, "deleted": 1}
+        """
         self.sync_job.log_info("--- Counters ---")
-        for k, v in sorted(ingestion_stats.items()):
+        for k, v in sorted(counters.items()):
             self.sync_job.log_info(f"'{k}' : {v}")
+        self.sync_job.log_info(f"full counters dictionary: {counters}")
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov:
Might be worth to put many of the counters that are not too important under DEBUG.

I read example output and the numbers confuse me (e.g. `bulk_operations.index` VS `doc_created` VS `indexed_document_count `).

Also personally not sure about multiline output, previous output was consistent between intermediate/final results + did not occupy a lot of space in logs (yet already we received reports about problems).

IMO having INFO log on same verbosity as now + adding DEBUG logs with better explanation and logical grouping + highlighting disrepancies would be better.

connectors/sync_job_runner.py
```diff
@@ -374,11 +376,15 @@ async def _sync_done(self, sync_status, sync_error=None):
...
-            f"deleted: {result.get('doc_deleted', 0)} "
+            f"created: {ingestion_stats.get('doc_created', 0)} | "
+            f"updated: {ingestion_stats.get('doc_updated', 0)} | "
+            f"deleted: {ingestion_stats.get('doc_deleted', 0)} "
             f"(took {int(time.time() - self._start_time)} seconds)"  # pyright: ignore
         )
+        self.sync_job.log_info("--- Counters ---")
+        for k, v in sorted(ingestion_stats.items()):
+            self.sync_job.log_info(f"'{k}' : {v}")
+        self.sync_job.log_info("----------------")
```
wangch079, line 387: This contains duplicated counters as we already logged `created`, `updated` and `deleted`.

Considering remove these from the above log to only contain the status and time taken?

wangch079 (commented):
I also have concerns about multi-lines loggings. It looks good in console, but it will be shipped to Cloud as multiple documents.

connectors/sync_job_runner.py
seanstory: I think that kind of exclusion is how we ended up in our current state where not all counters are logged. I think it's probably better for us to over-expose than to under-expose. Nothing is hurt by having 3 counters logged twice.

As mentioned in the threads with Artem, this is distinct from the fact that we have several counters that effectively count the same things, and we should stop doing that. That can be part of follow-up work to remove duplicate metrics, or rename metrics to more clearly highlight the differentiators. This change is just helping to expose that separate issue.

seanstory:
> I also have concerns about multi-lines loggings. It looks good in console, but it will be shipped to Cloud as multiple documents.

This is intentional. This will let a support engineer more easily search for and understand a given counter. For instance, I can do a query in discover like:
```
deployment.id: <deploymentId> AND message: "attachment_extracted"
```
over a 2-month window to quickly get an idea for how my attachment processing has changed over time. If I had all these counters in one `message`, that would be a lot harder to see, as I'd have to expand each row in order to scroll past the first ~20 characters and get to the counter I care about.

However, Artem raised a similar concern. We discussed quickly over zoom, and came to the conclusion that we can have both logs that are valuable for human eyes (one count per line) and logs that are more easily analyzed by a machine (log a single line with the dict of all counters).

Will that address your concern, @wangch079 ?

connectors/sync_job_runner.py
```diff
@@ -374,11 +376,15 @@ async def _sync_done(self, sync_status, sync_error=None):
...
             f"Sync ended with status {sync_status.value} -- "
-            f"created: {result.get('doc_created', 0)} | "
-            f"updated: {result.get('doc_updated', 0)} | "
-            f"deleted: {result.get('doc_deleted', 0)} "
+            f"created: {ingestion_stats.get('doc_created', 0)} | "
+            f"updated: {ingestion_stats.get('doc_updated', 0)} | "
+            f"deleted: {ingestion_stats.get('doc_deleted', 0)} "
             f"(took {int(time.time() - self._start_time)} seconds)"  # pyright: ignore
         )
+        self.sync_job.log_info("--- Counters ---")
```
vidok, line 384: I think it's worth moving the logging logic ([L384-L387](https://github.com/elastic/connectors/pull/2323/files#diff-f4def1564167603a89d66578a9ac61c61f5d0c63298e1ceab10b22f3d949ff8eR384-R387)) out of the _sync_done method just to keep it simple.

seanstory: Just into its own function? Or are you wanting it out of the file entirely?

connectors/sync_job_runner.py
wangch079: Eventually, I expect all counters to be logged together as shown [here](https://github.com/elastic/connectors/pull/2323#issue-2218560564).

With proper renaming, the counters can be logged more clearly. Then we can simply the end-sync log to something like

```
Sync ended with status completed (took 200 seconds)
```

But it's fine if you address it as a follow-up work.

artem-shelkovnikov: Ideally actually we would send counters to some sort of metrics index so that the customer or us could make dashboards, run queries on them and so on, probably worth to do later?

connectors/sync_job_runner.py
vidok: I think sync_job_runner.py is a good place for this method.

connectors/sync_job_runner.py
seanstory: Yep, persisting more counters is definitely a "later" thing, since it involves changes to the protocol and to our APIs.

seanstory:
OK, now the final output looks like:

```
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] Sync ended with status completed -- created: 1 | updated: 0 | deleted: 0 (took 119 seconds)
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] --- Counters ---
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'attachment_extracted' : 0
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'bulk_operations' : {'index': 1}
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'deleted_document_count' : 0
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'doc_created' : 1
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'doc_deleted' : 0
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'doc_updated' : 0
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'indexed_document_count' : 1
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] 'indexed_document_volume' : 0
[FMWK][13:02:07][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: UCAdpY4BS_YGsI0K8edF] full counters dictionary: {'doc_created': 1, 'attachment_extracted': 0, 'doc_updated': 0, 'doc_deleted': 0, 'bulk_operations': {'index': 1}, 'indexed_document_count': 1, 'indexed_document_volume': 0, 'deleted_document_count': 0}
```

seanstory:
Bump - can I get a re-review or approval?

artem-shelkovnikov (approved):
🚢

PR #2329, opened 2024-04-03
Add logging

moxarth-elastic:
## Part of https://github.com/elastic/connectors/issues/2299

This PR includes logging for the following connectors:
1. Azure Blob Storage
2. Confluence
3. MS SQL
4. Oracle
5. PostgreSQL

Also, there are few changes in confluence as per this PR: https://github.com/elastic/connectors/pull/2326

Drive folder for log files with DEBUG log_level: https://drive.google.com/drive/folders/13-P99PjgjFdjI9hrmPsuhNst6Kdu5C0-

#### Pre-Review Checklist


connectors/sources/confluence.py
```diff
@@ -714,6 +726,7 @@ async def fetch_server_space_permission(self, space_key):
             return {}
 
         url = URLS[SPACE_PERMISSION].format(space_key=space_key)
+        self._logger.debug(f"Fetching permissions for '{space_key}' space")
```
timgrein, line 729: Should this be `info` log-level as it indicates progress?

connectors/sources/mssql.py
```diff
@@ -255,6 +255,9 @@ async def ping(self):
     async def get_tables_to_fetch(self, is_filtering=False):
         tables = configured_tables(self.tables)
         if is_wildcard(tables) or is_filtering:
+            self._logger.info(
+                "Fetching all tables as the configuration field 'tables' is set to '*' or advanced sync rule is enabled."
+            )
```
timgrein, line 260:
```suggestion
            self._logger.info(
                "Fetching all tables as the configuration field 'tables' is set to '*' or advanced sync rules are enabled."
            )
```

connectors/sources/mssql.py
```diff
@@ -332,15 +340,20 @@ async def data_streamer(self, table=None, query=None):
         Yields:
             list: It will first yield the column names, then data in each row
         """
+        self._logger.debug(
+            f"Streaming records from database for table: {table} or using query: {query}"
+        )
```
timgrein, line 345: I'm not the biggest fan of using `or` inside a log statement as it adds ambiguity. I think we should have two distinct log messages depending on the value of `query`. You could also move the assignment of the parameter for `get_cursor` also into this condition to not check it twice.

connectors/sources/mssql.py
```diff
@@ -553,6 +568,9 @@ async def fetch_documents_from_query(self, tables, query):
         Yields:
             Dict: Document to be indexed
         """
+        self._logger.debug(
```
timgrein, line 571: Should this also be info level as it indicates progress?

connectors/sources/postgresql.py
```diff
@@ -218,6 +219,9 @@ async def ping(self):
     async def get_tables_to_fetch(self, is_filtering=False):
         tables = configured_tables(self.tables)
         if is_wildcard(tables) or is_filtering:
+            self._logger.info(
```
timgrein, line 222: Same reasoning as in `mssql.py`: I think we should've two distinct log messages instead of using one containing `or`.

connectors/sources/mssql.py
timgrein: I think we should also split the log message without concatenating two possible reasons with `or`. This adds ambiguity IMO.

connectors/sources/postgresql.py
```diff
@@ -537,6 +546,9 @@ async def fetch_documents_from_query(self, tables, query):
         Yields:
             Dict: Document to be indexed
         """
+        self._logger.debug(
```
timgrein, line 549: Should this be `info` level?

timgrein (commented):
Nice improvement so far! Left some comments

artem-shelkovnikov:
Please do not combine multiple connectors in same PR.

As for the changes - can you include example logs for the sync before/after for each connector with INFO and separately with DEBUG log level? E.g.:

- ABS Before INFO
- ABS After INFO
- ABS Before DEBUG
- ABS After DEBUG
- Confluence Before INFO
- Confluence After INFO
- etc...

moxarth-elastic:
> Please do not combine multiple connectors in same PR.

@artem-shelkovnikov We're planning to update the logging functionality across roughly 20 connectors. 
During our weekly meeting, we discussed with @DianaJourdan that bundling the changes for every 5 connectors into a single pull request would streamline the process. 
Since the changes are confined to the logger functionality, reviewing the updates for 5 connectors at a time should not be overly time-consuming or complicated. This approach avoids the complexity of simultaneously managing 20 open pull requests per connector. 
Consequently, we will proceed by consolidating the updates into four separate pull requests, each encompassing 5 connectors.

Let me know if you have differing opinion for raising four PRs each containing 5 connector.

connectors/sources/confluence.py
moxarth-elastic: Yes, this should be info.

connectors/sources/mssql.py
moxarth-elastic: Sure, i've bifurcated this and made it dependant on the value of `query`.

moxarth-elastic:
> can you include example logs for the sync before/after for each connector with INFO and separately with DEBUG log level?

I think we can inspect the logs from E2E tests in Buildkite rather than creating four separate log files (including before/after logs for both INFO and DEBUG levels). Generating four files for approximately 20 connectors seems very time-consuming.

artem-shelkovnikov:
> I think we can inspect the logs from E2E tests in Buildkite rather than creating four separate log files (including before/after logs for both INFO and DEBUG levels). Generating four files for approximately 20 connectors seems very time-consuming.

That is my concern as well, but also reviewing logs for E2E tests in Buildkite side by side for each reviewer is time consuming, I think it's fair to at least attach the links to E2E test runs here for each connector before/after. Also E2E only sends DEBUG logs, which produces a lot more than INFO, manually filtering these outs is huge work as well.

Without actually looking at logs generated we cannot tell if logs are good now or not.

parth-elastic:
> Without actually looking at logs generated we cannot tell if logs are good now or not.

Indeed, your point sounds reasonable. How about a hybrid approach, where we generate a before/after and an Info/Debug log file for just one connector within each PR for all the 4 PRs. 
Since the logs for different connectors are likely to be similar, this method would enable us to check the log changes effectively with reduced effort. WDYT?

artem-shelkovnikov:
> Since the logs for different connectors are likely to be similar, this method would enable us to check the log changes effectively with reduced effort. WDYT?

But I think also the point is that for some connectors (like S3/GCS/ABS) they will likely be similar, so for such connectors we can do that, but for different style connectors (S3 vs Jira vs Github) expected logs are gonna be different. We can try to break connectors by log simplicity into grouped PRs, but I'm also not super sure about this. Let me bring this up with the team

parth-elastic:
> the point is that for some connectors (like S3/GCS/ABS) they will likely be similar, so for such connectors we can do that

We can organize our pull requests by grouping similar types of connectors together. For instance, this PR includes 3 database connectors, so we could produce before/after logs specifically for MySQL as a representative. In a subsequent PR, if we bundle together connectors like S3, GCS, Box, OneDrive, and Dropbox, which are all storage services, we can create a single set of logs for one of them as an example. We can apply the same strategy for future PRs, streamlining our logging efforts by focusing on a representative connector from each group.

artem-shelkovnikov:
I've opened the discussion with the team, we'll get back to you about it shortly

artem-shelkovnikov:
Hey @moxarth-elastic, as mentioned in Slack, can we have logs in DEBUG log_level for each connector changed here to continue the review?

moxarth-elastic:
> Hey @moxarth-elastic, as mentioned in Slack, can we have logs in DEBUG log_level for each connector changed here to continue the review?

@artem-shelkovnikov Let me upload the log files for the connectors of this PR, so you can continue your review

moxarth-elastic:
@artem-shelkovnikov You can check the log file from this drive folder: https://drive.google.com/drive/folders/13-P99PjgjFdjI9hrmPsuhNst6Kdu5C0-

artem-shelkovnikov:
Can you break this PR into smaller PRs per connector? I've checked the logs and will have feedback for each connector, addressing all those in one PR will be pretty tough IMO

artem-shelkovnikov:
Closing in favour of other PRs

PR #2328, opened 2024-04-02
Create a nested_get_from_dict function in utils

wangch079:
I found the method [get_data_by_keys](https://github.com/elastic/connectors/blob/01d30bbcbeb1c4151ceb44d46cc008e8df0dc2e6/connectors/sources/github.py#L779) in GitHub should exist in `utils`. And there's another similar implementation [_nested_feature_enabled](https://github.com/elastic/connectors/blob/01d30bbcbeb1c4151ceb44d46cc008e8df0dc2e6/connectors/protocol/connectors.py#L520) in `Features` class.

This PR move the function to `utils`.

#### Pre-Review Checklist
- [x] this PR does NOT contain credentials of any kind, such as API keys or username/passwords (double check `config.yml.example`)
- [x] this PR has a meaningful title
- [x] this PR has a thorough description
- [x] Covered the changes with automated tests
- [x] Tested the changes locally
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

PR #2330, opened 2024-04-03
Apply function nested_get_from_dict to more places

wangch079:
A follow-up PR on #2328, to apply this function to more places.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -655,13 +655,11 @@ class GitHubClient:
     async def _get_retry_after(self, resource_type):
         current_time = time.time()
         response = await self._get_client.getitem("/rate_limit")
-        reset = (
-            response.get("resources", {})
-            .get(resource_type, {})
-            .get("reset", current_time)
+        reset = nested_get_from_dict(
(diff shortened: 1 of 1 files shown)


connectors/sources/github.py
```diff
@@ -655,13 +655,11 @@ def get_rate_limit_encountered(self, status_code, message):
...
-        reset = (
-            response.get("resources", {})
-            .get(resource_type, {})
-            .get("reset", current_time)
+        reset = nested_get_from_dict(
+            response, ["resources", resource_type, "reset"], default=current_time
         )
         # Adding a 5 second delay to account for server delays
-        return (reset - current_time) + 5
+        return (reset - current_time) + 5  # pyright: ignore
```
timgrein, line 662: What do we need to ignore here?

timgrein (approved):
LGTM, added one small question

connectors/sources/github.py
wangch079: pyright will complain `reset` can be `None` even though I have provided `current_time` as default

PR #2336, opened 2024-04-03
[GraphQL] Update GraphQL Object List RCF for identifying unique fields

parthpuri-elastic:
## Part of https://github.com/elastic/connectors-py/issues/2068

Added new RCF for identification of unique field for IDs.

#### Pre-Review Checklist
- [x] Adde a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)
   version: 8.14.0

### DIFF
--- a/connectors/sources/graphql.py
+++ b/connectors/sources/graphql.py
@@ -376,7 +376,7 @@ class GraphQLDataSource(BaseDataSource):
                 "required": False,
             },
             "graphql_object_list": {
-                "label": "GraphQL Objects List",
+                "label": "GraphQL Objects to ID mapping",
                 "order": 9,
                 "tooltip": "Specifies which GraphQL objects should be indexed as individual documents. This allows finer control over indexing, ensuring only relevant data sections from the GraphQL response are stored as separate documents. Use a JSON with key as the GraphQL object name and value as string field within the document, with the requirement that each document must have a distinct value for this field. Use '.' to provide full path of the object from the root of the response. For example {'organization.users.nodes': 'id'}",
                 "type": "str",
@@ -504,6 +504,9 @@ class GraphQLDataSource(BaseDataSource):
             except Exception as exception:
                 msg = f"Error while processing configured GraphQL objects and IDs. Exception: {exception}"
                 raise ConfigurableFieldValueError(msg) from exception
+        else:
+            msg = "GraphQL Objects to ID mapping field is not configured."
+            raise ConfigurableFieldValueError(msg)
 
         for (
             graphql_object,
--- a/connectors/sources/graphql.py
+++ b/connectors/sources/graphql.py
@@ -20,7 +20,6 @@ from connectors.source import BaseDataSource, ConfigurableFieldValueError
 from connectors.utils import (
     CancellableSleeps,
     RetryStrategy,
-    hash_id,
     iso_utc,
     retryable,
 )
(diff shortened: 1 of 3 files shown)


parthpuri-elastic:
buildkite test this

connectors/sources/graphql.py
```diff
@@ -380,9 +380,15 @@ def get_default_configuration(cls):
                 "tooltip": "Specifies which GraphQL objects should be indexed as individual documents. This allows finer control over indexing, ensuring only relevant data sections from the GraphQL response are stored as separate documents. Use '.' to provide full path of the object from the root of the response. For example 'organization.users.nodes'",
                 "type": "list",
             },
+            "graphql_field_id": {
+                "label": "GraphQL field name for ID",
+                "order": 10,
+                "tooltip": "Specifies a string field within the document, with the requirement that each document must have a distinct value for this field.",
```
jedrazb, line 386: Generally looks good, but if we allow to define multiple graphQL objects to index, we should also allow to specify ID for each of those objects, so maybe some approach with comma-separated list of field names?

parthpuri-elastic: But mostly single source have same name of field to identify the document. Suppose GitHub have unique field `id` so we can found `id` in every object.

jedrazb: > But mostly single source have same name of field to identify the document.

Can we guarantee that it holds for all sources? It really depends on the implementation and the user schema what field denotes the unique ID for the object

> Suppose GitHub have unique field id so we can found id in every object.

I think we should be careful overfitting connector to public github graphql server. We need to test the connector with other implementations?

parthpuri-elastic: Okay in that case if we want to move further with multiple ID we need to take json object as an input to map IDs with GraphQL objects. So IMO we can add IDs json in the same configuration field (`graphql_object_list`).
```
{
      "repository.issue": "id",
      "repository.pullRequest": "id"
}
```
So, WDYT? do you want me to move further with json?

jedrazb: Hmm I'm thinking about the aspect of "user friendliness", typing a JSON by hand can be cumbersome, what about two comma-separated lists?

parthpuri-elastic: But in that case, how can we map IDs with objects? I suppose we need to match the indices of two different lists, and users would need to manage two different config fields, which I find more complex. Additionally, I'm not sure if the framework returns a list in the same order as the input or not. We have many JSON input fields such as variables and headers. In my opinion, most of the users who work with GraphQL can easily configure this.

jedrazb: Can we wait with that decision for @danajuratoni? It seems to be a UX decision and should be quick and easy to implement

danajuratoni: I expect GraphQL users to be comfortable with JSON and I think JSON would be more intuitive than explaining the 2 lists are related, ordered and we'd need to validate the number of values match.
Let's go with JSON!

connectors/sources/graphql.py
```diff
@@ -377,8 +378,8 @@ def get_default_configuration(cls):
             "graphql_object_list": {
                 "label": "GraphQL Objects List",
```
jedrazb, line 379: Should be renamed to
`GraphQL Objects to ID mapping`

connectors/sources/graphql.py
```diff
@@ -487,15 +496,37 @@ async def validate_config(self):
                 msg = f"Error while processing configured GraphQL variables. Exception: {exception}"
                 raise ConfigurableFieldValueError(msg) from exception
 
-        for graphql_object in self.graphql_client.graphql_object_list:
-            if not self.check_field_existence(ast=ast, field_path=graphql_object):
+        if graphql_object_list:
```
jedrazb, line 499: Else, should we throw an error since nothing will get indexed?

parthpuri-elastic:
buildkite test this

parthpuri-elastic:
buildkite test this

parthpuri-elastic:
buildkite test this

jedrazb (approved):
LGTM

PR #2337, opened 2024-04-03
[Notion] [Bugfix] Handle error on fetching children of external objects

praveen-elastic:
## Closes https://github.com/elastic/connectors-py/issues/2306

Following changes have been added in this PR:

Notion API returns `validation error` for fetching children of block type `external object instance page`. This PR handles this error, logs the error and prevent the sync to fail.

```
{
    "object": "error",
    "status": 400,
    "code": "validation_error",
    "message": "Block type external_object_instance_page is not supported via the API.",
    "developer_survey": "https://notionup.typeform.com/to/bllBsoI4?utm_source=postman",
    "request_id": "1fffe248-53f2-4e9b-a658-4550f2ab6332"
}
```

#### Pre-Review Checklist


seanstory:
Looking at [the issue](https://github.com/elastic/connectors/issues/2306), I think Davey's suggestion is a good one, that we should be checking `has_children` _before_ calling `self._get_client.blocks.children.list`. The block that caused this problem had `has_children: false`, which would prevent us from getting an error in the first place.

As with the other PR, I'd prefer that we not swallow errors, but defend against getting into an error state to start with.

praveen-elastic:
> Looking at [the issue](https://github.com/elastic/connectors/issues/2306), I think Davey's suggestion is a good one, that we should be checking `has_children` _before_ calling `self._get_client.blocks.children.list`. The block that caused this problem had `has_children: false`, which would prevent us from getting an error in the first place.
> 
> As with the other PR, I'd prefer that we not swallow errors, but defend against getting into an error state to start with.

Yes, you are right, but we get the initial list of blocks from /search query which does not have the field `has_children` so we can't decide beforehand. The /children endpoint has the field, so we utilised it for next set of recursive calls. 

As for the other blocks returned by /search, if they don't have any child, they return empty list instead of raising an error.

Attaching the response of /search for reference: 
[response_search.json](https://github.com/elastic/connectors/files/14856237/response_search.json)
Here the first one in the response is the block whose children when fetched returns `validation error`

seanstory:
ooh, I also see:
```
      "is_inline": false,
```
In 
<details><summary>this object</summary>

```
    {
      "object": "database",
      "id": "72ba2d00-eed1-4652-ae23-43e3d1df2e8c",
      "cover": null,
      "icon": null,
      "created_time": "2024-04-01T09:18:00.000Z",
      "created_by": {
        "object": "user",
        "id": "261363a8-9b19-46a4-83b3-1fbc0853417f"
      },
      "last_edited_by": {
        "object": "user",
        "id": "261363a8-9b19-46a4-83b3-1fbc0853417f"
      },
      "last_edited_time": "2024-04-01T15:37:00.000Z",
      "title": [
        {
          "type": "text",
          "text": {
            "content": "Google drive connected app",
            "link": null
          },
          "annotations": {
            "bold": false,
            "italic": false,
            "strikethrough": false,
            "underline": false,
            "code": false,
            "color": "default"
          },
          "plain_text": "Google drive connected app",
          "href": null
        }
      ],
      "description": [],
      "is_inline": false,
      "properties": {
        "Google Drive File": {
          "id": "hQZR",
          "name": "Google Drive File",
          "type": "relation",
          "relation": {
            "database_id": "f238da5a-4563-4aec-b1ea-661bba665f9d",
            "synced_property_name": "Related to Google drive connected app (Google Drive File)",
            "synced_property_id": "T<F@"
          }
        },
        "Tags": {
          "id": "js%5DQ",
          "name": "Tags",
          "type": "multi_select",
          "multi_select": {
            "options": []
          }
        },
        "Name": {
          "id": "title",
          "name": "Name",
          "type": "title",
          "title": {}
        }
      },
      "parent": {
        "type": "workspace",
        "workspace": true
      },
      "url": "https://www.notion.so/72ba2d00eed14652ae2343e3d1df2e8c",
      "public_url": null,
      "archived": false
    },
```

</details>

praveen-elastic:
> ooh, I also see:
> 
> ```
>       "is_inline": false,
> ```
> 
> In
> 
> this object

Yes, it could have been useful in the other bugfix, if it was present in `/blocks/id/children` endpoint 😄. For the present one, I guess we cannot filter out which block would throw an error.

praveen-elastic:
Also, after testing with some more connected apps, I found this error occurs when a connected property is added to a database: https://www.notion.so/help/link-previews-and-synced-databases#connected-properties. Only four external applications can be used as a connected property.
image

All four have a property field of template: `Related to <database_name> (Figma File)` /  `Related to <database_name> (Google Drive File)` / `Related to <database_name> (GitHub Pull Requests)` ... , so if the block has a property matching these, we skip fetching it's children. WDYT ?

seanstory:
👍 Sounds good.

connectors/sources/notion.py
```diff
@@ -176,17 +176,27 @@ async def fetch_children_recursively(block):
...
+                        async for child in fetch_children_recursively(block):
+                            yield child
+                if block.get("type") == "child_database":
+                    async for record in self.query_database(block.get("id")):
+                        yield record
+        except APIResponseError as error:
+            if error.code == "validation_error":
+                self._logger.warning(
+                    f"Skipping children of block with id: {block_id} as not supported by API : {error}"
+                )
```
seanstory, line 199: I really don't like this, because it doesn't "fail fast." If there's a bug and we're making invalid requests, we won't catch it unless the customer is watching their logs. Instead, I'd rather raise errors _unless_ we know exactly what caused this error and can justify ignoring it other than logging that it happened.

connectors/sources/notion.py
```diff
@@ -567,6 +577,21 @@ def generate_query(self):
...
+        properties = page_database.get("properties")
+        if properties is None:
+            return False
+        for field in properties.keys():
+            if field.startswith("Related to") and (
+                "(Figma File)" in field
+                or "(Google Drive File)" in field
+                or "(GitHub Pull Requests)" in field
+                or "(Zendesk Ticket)" in field
+            ):
```
seanstory, line 590: Do we need this part of the check? Is `field.startswith("Related to")` not sufficient?
What I'm concerned about is that they'll have some other type of integration added for something like OneDrive, and then `Related to (OneDrive)` will not match this expression and we'll see this error again.

connectors/sources/notion.py
```diff
@@ -577,6 +602,9 @@ async def retrieve_and_process_blocks(self, query):
             yield self._format_doc(page_database), None
             self._logger.info(f"Fetching child blocks for block {block_id}")
 
+            if self.is_connected_property_block(page_database):
```
seanstory, line 605: Let's leave a debug log here before the `continue`

connectors/sources/notion.py
seanstory: Or maybe we could make it a regex match for `Related to .* connected app .*`?

connectors/sources/notion.py
praveen-elastic: Updated with commit id: [a0e3241](https://github.com/elastic/connectors/pull/2337/commits/a0e324129dad81dcbf0b949bfafa46ac79f73f55)

connectors/sources/notion.py
seanstory: What was the API response for the block, before we tried to fetch its children? There's got to be ways to identify that we shouldn't try to fetch the children of these.

praveen-elastic: Hey, the file [block-response.json](https://github.com/elastic/connectors/files/14947924/block-response.json), is the response for the block before we try to fetch its children.

seanstory: 🤦 I misunderstood. 
Ok, yeah, there's not a lot of info there.

Is the error more informative? Maybe I'd be less hesitant to swallow this error if we could be really narrow in what we moved past. Like, is the error message informative at all, and something that we could match on only if it's this specific error case of an external object?

praveen-elastic: The error is not specific to external_objects. Following is the response we get:
```
{
    "object": "error",
    "status": 400,
    "code": "validation_error",
    "message": "Block type external_object_instance_page is not supported via the API.",
    "developer_survey": "https://notionup.typeform.com/to/bllBsoI4?utm_source=postman",
    "request_id": "9f987592-c194-4cfd-afa7-0397f22faee6"
}
```

seanstory: That looks very specific to me. Like we could do:

```
try:
  do_thing()
except APIResponseError as error:
  if error.code == "validation_error" and "external_object" in error.message:
    log.warn("Encountered external object. Skipping")
  else:
    raise
```

The `and "external_object" in error.message:` would hopefully ensure we only do this for this narrow subset of "external_object" issues that we can't anticipate, but don't hide other bugs that might surface as Validation Errors.

praveen-elastic: Got it, updated with commit id:  [b18f0d4](https://github.com/elastic/connectors/pull/2337/commits/b18f0d4c1ed5b2568ae32c162d183b9942889e1e)

connectors/sources/notion.py
```diff
@@ -176,17 +177,29 @@ async def fetch_children_recursively(block):
...
+                ]:
+                    yield block
+                    if block.get("has_children") is True:
+                        async for child in fetch_children_recursively(block):
+                            yield child
+                if block.get("type") == "child_database":
+                    async for record in self.query_database(block.get("id")):
+                        yield record
+        except APIResponseError as error:
+            if error.code == "validation_error" and "external_object" in str(error):
```
seanstory, line 197: I assume APIResponseError doesn't have a property you can look at to get at the `external_object`? I don't love having to use `str()`, as that widens the opportunity for false positives.

seanstory (commented):
Looking better. Let's add a test to protect against regression for handling the external_object APIResponseError, and get this in. :)

praveen-elastic:
Buildkite test this

connectors/sources/notion.py
praveen-elastic: Updated with commit id: [fe41c9a](https://github.com/elastic/connectors/pull/2337/commits/fe41c9a2a9dbaaad1dad9cdd700ac33631c90a19)

seanstory (approved):

seanstory:
@praveen-elastic this didn't automatically make it into the 8.14 branch, but only main, because of a delay in bumping main and its backportrc to 8.15. If this needs to be in 8.14.0, you'll need to create a manual backport

praveen-elastic:
> @praveen-elastic this didn't automatically make it into the 8.14 branch, but only main, because of a delay in bumping main and its backportrc to 8.15. If this needs to be in 8.14.0, you'll need to create a manual backport

Okay, got it, i'll raise a manual backport PR for this change.

PR #2342, opened 2024-04-05
Update README.md

artem-shelkovnikov:
We really don't run on 3.10+, updating readme


efegurkan (approved):

seanstory:
@artem-shelkovnikov should we go ahead and backport this all the way back?

artem-shelkovnikov:
@seanstory good idea!

PR #2347, opened 2024-04-05
count more things

seanstory:
## Closes https://github.com/elastic/connectors/issues/2317
relates to https://github.com/elastic/connectors/issues/1489

Adds a bunch of new counters, and renames a few of the existing counters. This should provide us a much more nuanced view into what kinds of operations are done by a given job, and give clues to any systemic problems (duplicate IDs, agrressive filtering/dropping, unhealthy ES, etc).

Example:
```
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] Sync ended with status completed -- created: 103 | updated: 1 | deleted: 0 (took 120 seconds)
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] --- Counters ---
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'bulk_item_responses._id_duplicates' : 2
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'bulk_item_responses.index' : 104
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'bulk_operations.index' : 104
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'deleted_document_count' : 0
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'doc_creates_queued' : 103
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'doc_updates_queued' : 1
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'docs_extracted' : 104
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'indexed_document_count' : 103
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'indexed_document_volume' : 0
[FMWK][13:53:25][INFO] [Connector id: TCAXpY4BS_YGsI0Kbufn, index name: slack, Sync job id: WiKZr44BK4iIzr6Z6MxN] 'result_successes' : 104
```

Screenshot 2024-04-05 at 2 07 10 PM

This shows that we fetched 104 docs from the source, determined that 1 of those was already in Elasticsearch, sent 104 `index` operations, but only 103 documents are in the index currently. This is because two of the documents were computed to have the same ID, and therefore ended up overwriting  one another. Further, we can be confident that we're not at 103 docs because of any Drop Processor interactions, because the value for `docs_dropped` is not present (0)

#### Pre-Review Checklist

## Release Note

More counters are now present in the logs after each sync, to facilitate debugging and troubleshooting.


connectors/es/sink.py
```diff
@@ -117,27 +141,24 @@ def __init__(
...
         self._enable_bulk_operations_logging = enable_bulk_operations_logging
+        self.counters = Counters()
 
     def _bulk_op(self, doc, operation=OP_INDEX):
         doc_id = doc["_id"]
         index = doc["_index"]
 
         if operation == OP_INDEX:
             return [{operation: {"_index": index, "_id": doc_id}}, doc["doc"]]
-        if operation == OP_UPSERT:
```
artem-shelkovnikov, line 140: What about this change? Did we handle anything wrong here?

artem-shelkovnikov, line 140: I'm not sure about getting rid of upsert - it's more resilient to weird errors or manual index manipulations. How will OP_UPDATE act if record's not present in the index?

connectors/es/sink.py
```diff
@@ -181,47 +203,86 @@ async def _bulk_api_call():
...
+            requested_op = ids_to_ops.get(doc_id, None)
+            if requested_op is None:
+                self.counters.increment(
+                    ID_CHANGED_AFTER_REQUEST, namespace=BULK_RESPONSES
+                )
+            elif action_item != OP_UPDATE and result == "updated":
+                self.counters.increment(ID_DUPLICATE, namespace=BULK_RESPONSES)
+
+            if result == "noop":
+                self.counters.increment(DOCS_DROPPED, namespace=BULK_RESPONSES)
```
artem-shelkovnikov, line 264: noop == "dropped by pipeline"?

IMO might be useful to put short code comments here about what's the potential cause of each of that.

connectors/es/sink.py
```diff
@@ -684,14 +748,16 @@ async def enqueue_docs_to_delete(self, existing_ids):
...
+        self,
+    ):
         self._logger.info(
             "Sync progress -- "
-            f"created: {self.total_docs_created} | "
-            f"updated: {self.total_docs_updated} | "
-            f"deleted: {self.total_docs_deleted}"
+            f"created: {self.counters.get(CREATES_QUEUED)} | "
+            f"updated: {self.counters.get(UPDATES_QUEUED)} | "
+            f"deleted: {self.counters.get(DELETES_QUEUED)}"
```
artem-shelkovnikov, line 760: So now it also counts queued operations, not executed, right?

connectors/es/sink.py
```diff
@@ -802,26 +868,12 @@ async def cancel(self):
...
-                    "bulk_operations": dict(self._sink.ops),
-                    "indexed_document_count": self._sink.indexed_document_count,
-                    # return indexed_document_volume in number of MiB
-                    "indexed_document_volume": round(
-                        self._sink.indexed_document_volume / (1024 * 1024)
-                    ),
-                    "deleted_document_count": self._sink.deleted_document_count,
-                }
-            )
+            stats.update(self._sink.counters.to_dict())
```
artem-shelkovnikov, line 873: So we extend the protocol to contain more stats? Might be worth to change README for the protocol in the repo too!

artem-shelkovnikov (commented):
Like the direction! I've left a couple comments/questions :)

connectors/es/sink.py
seanstory: So, fun fact, `OP_UPSERT="update"`. See it in main:

https://github.com/elastic/connectors/blob/4047c2c6edf97ae8cfc9c846b3ec62e0c4f88157/connectors/es/sink.py#L57

This made me angry, so I renamed the variable to represent what the value actually was. Logically, nothing has changed.

connectors/es/sink.py
seanstory: > noop == "dropped by pipeline"?

I guess so?

Try it out:

```
PUT test

PUT _ingest/pipeline/test-pipeline
{
  "processors": [
    {
      "drop": {
        "if": "true"
      }
    }
  ]
}

POST test/_bulk?pipeline=test-pipeline
{"index": {"_index": "test", "_id":"1"}}
{"foo":"bar"}

```

Results in:

```
{
  "errors": false,
  "took": 0,
  "ingest_took": 0,
  "items": [
    {
      "index": {
        "_index": "test",
        "_id": "1",
        "_version": -3,
        "result": "noop",
        "_shards": {
          "total": 0,
          "successful": 0,
          "failed": 0
        },
        "status": 200
      }
    }
  ]
}
```


> IMO might be useful to put short code comments here about what's the potential cause of each of that.

Can do. 👍

connectors/es/sink.py
```diff
@@ -377,7 +438,7 @@ async def _deferred_index(self, lazy_download, doc_id, doc, operation):
         data = await lazy_download(doit=True, timestamp=doc[TIMESTAMP_FIELD])
 
         if data is not None:
-            self.total_downloads += 1
+            self.counters.increment(ATTACHMENTS_EXTRACTED)
```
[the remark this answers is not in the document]
seanstory: "Extraction" is an overloaded term. I expect it's related to the Extractor/Sink class terminology, in that these are attachments that were output by the Extractor. I just named the constant what the value currently is. See main:

https://github.com/elastic/connectors/blob/eba0a313924f6dd443f5992a76bc1f164f7e1aea/connectors/es/sink.py#L808

However, I don't think this counter is currently used anywhere, so I'm not opposed to renaming it. What about "binary_documents_downloaded"? "Attachment" is misleading too, for sources like S3 or Google Drive.

connectors/es/sink.py
seanstory: It was always counting queued operations. Now it also counts executions in the final result. ;) 

Do you want me to look at changing the progress logs to be based on executed results, rather than queued operations?

connectors/es/sink.py
seanstory: Nope, the protocol has not changed here. The protocol just takes a subset of our counter keys, and that's defined here: 
https://github.com/elastic/connectors/blob/8b07cf3a268db9361062e0384aa67ddac4488647/connectors/protocol/connectors.py#L82-L87

I constantized this in my changes, but didn't actually change what we're storing. I think we eventually should persist more, but that's outside the scope of this current ticket.

connectors/es/sink.py
artem-shelkovnikov: > Now it also counts executions in the final result. ;)
I love it!

> Do you want me to look at changing the progress logs to be based on executed results, rather than queued operations?

Not in this PR definitely, but I think it might be worth to take a look at it - it's great that now with this change we have options!

connectors/es/sink.py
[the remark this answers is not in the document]
artem-shelkovnikov: Yeah I like "BINARY_DOCUMENTS_DOWNLOADED", maybe even "ATTACHMENTS_DOWNLOADED" is okay - I vote for shorted name possible 😅

connectors/es/sink.py
artem-shelkovnikov: I can see why it was called "UPSERT" - cause it's indeed upsert, not an update - if document does not exist it's created - that would catch me off-guard if I was not familiar with how Elasticsearch handles it.

I don't oppose the rename, though :)

seanstory: the same is true with `OP_INDEX`. In fact, that's the better candidate for upsert, because `index` operations are expected to provide the full document, but `update` operations only require the piece of the document being updated. We don't make use of that, so there's really no need for us to use `update` operations at all. We just _count_ how many docs we're indexing that we already have IDs for as "updates", but we're actually using `OP_INDEX` for them. :/ It's a bit of a cluster.

connectors/es/sink.py
seanstory: side note, thanks for this comment. I went to leave a code comment with a link to the docs, but found that the docs were actually outdated, and didnt' include `noop` or `not_found`. Tech writing team is gonna fix that, now.

connectors/es/sink.py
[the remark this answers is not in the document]
seanstory: changed to `BIN_DOCS_DOWNLOADED` for maximum clarity with minimal characters! :)

artem-shelkovnikov (approved):

PR #2355, opened 2024-04-08
[Network Drive] Change traversal strategy to ingest as you go instead of pre-fetching the whole tree for sync rule

akanshi-elastic:
## Closes https://github.com/elastic/connectors/issues/2373###
This PR contains 
1. traversal of directories/subdirectory changes for sync rule, now we are using scandir method to fetch data from network drive, which start ingestion data in seconds after connector starts syncing.
2. fix sync rule validation for network drive by adding remote validation where drive path will not start with `\`.(https://github.com/elastic/connectors/issues/2239)
3. For issue:2373 just fix try except properly

If this is an ad-hoc/trivial change and does not have a corresponding
issue, please describe your changes in enough details, so that reviewers
and other team members can understand the reasoning behind the pull request.-->

#### Pre-Review Checklist


connectors/sources/network_drive.py
```diff
@@ -112,16 +107,19 @@ async def _remote_validation(self, advanced_rules):
...
             )
+        start_with_slash = set()
 
-        await asyncio.to_thread(self.source.smb_connection.create_connection)
-
-        _, invalid_rules = await self.source.find_matching_paths(advanced_rules)
+        for rule in advanced_rules:
+            pattern = rule["pattern"]
+            if pattern.startswith("/"):
+                start_with_slash.add(pattern)
```
artem-shelkovnikov, line 115: Is this sufficient glob validation? What about special characters like "$" and stuff, do these need to be escaped?

connectors/sources/network_drive.py
```diff
@@ -419,49 +417,90 @@ async def traverse_diretory(self, path):
...
-                normalized_path = path.split("/", 1)[1].replace("\\", "/")
-                is_match = glob.globmatch(
-                    normalized_path, glob_pattern, flags=glob.GLOBSTAR
-                )
+            async for document in self.traverse_directory_for_syncrule(
+                path=rf"\\{self.server_ip}/{self.drive_path}",
+                glob_pattern=glob_pattern,
+            ):
+                yield document
+                rule_matched = True
```
artem-shelkovnikov, line 488: I don't think it works well when specifying multiple glob patterns - it will repeat traversal for each pattern. Can we do one traversal this way?

akanshi-elastic: yes, it will repeat the traversal for each pattern but if we do like this:
```
async for document in self.traverse_directory_for_syncrule(
                path=rf"\\{self.server_ip}/{self.drive_path}",
                glob_pattern=glob_pattern,
            ):
            for rule in rules:
                if match:
                    yield document
```
In this case there can be a time, when file/folder doesn't match with any of the rule and this can happen continuously for n number of documents. In that case connector will stop syncing until it finds the match.

connectors/sources/network_drive.py
akanshi-elastic: yes, we have verified with all these special characters: `C:\Users\Administrator\Documents\perf-data-1lac\LatestData\home\runner\!work#$%^&&&^$^&^&(&^&+_)` and it seems to be working fine with this. The only validation needs to be add is that path should not start with a slash.

connectors/sources/network_drive.py
artem-shelkovnikov: That's right, but I also see in code that you don't check if the folder you're traversing through matches the glob, e.g. imagine:

You have a glob "\something\**\*.txt" and you have a 20GB directory "\images" - why traverse it if it did not satisfy the glob already?

This optimisation should eliminate the problem and make things fast.

connectors/sources/network_drive.py
```diff
@@ -414,54 +416,120 @@ async def traverse_diretory(self, path):
...
+                    partial(
+                        smbclient.scandir,
+                        path=current_path,
+                        port=self.port,
+                    ),
+                )
+                for file in directory_info:
+                    match_with_previous_rules = False
+                    if file.is_dir():
+                        stack.append(file.path)
```
artem-shelkovnikov, line 461: Should we ignore folders here if they already don't match a part of glob as discussed?

connectors/sources/network_drive.py
```diff
@@ -414,54 +416,120 @@ async def traverse_diretory(self, path):
...
+                    file_path = file.path.split("/", 1)[1].replace("\\", "/")
+                    is_file_match = glob.globmatch(
+                        file_path, glob_pattern, flags=glob.GLOBSTAR
+                    )
+                    match_with_previous_rules = self.is_match_with_previous_rules(
+                        file_path, indexed_rules, match_with_previous_rules
+                    )
```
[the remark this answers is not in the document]
akanshi-elastic: This function is to not yield documents twice if multiple rules trigger for one object. It will fetch all the file folders for first pattern and from next pattern we'll match files from the previous rules and if it match with any of the previous rules then we'll not index it as it was already indexed while running previous patterns.

artem-shelkovnikov:
Once this PR is merged, we need to backport to 8.13 as well and release a docker image with this change, then hand it over to https://elastic.slack.com/archives/C7LLL50CA/p1710439113440969

artem-shelkovnikov (approved):

akanshi-elastic:
Buildkite test this

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.13|[](https://github.com/elastic/connectors/pull/2462)|

### Questions ?

PR #2357, opened 2024-04-08
Bump product version to 8.13.2.1

demjened:
Part of [8.13.2](https://github.com/elastic/enterprise-search-team/issues/7245) release.


connectors/VERSION
```diff
@@ -1 +1 @@
-8.13.1.1
+8.13.2.1
```
seanstory, line 1:
```suggestion
8.13.2.0
```

artem-shelkovnikov (approved):

PR #2363, opened 2024-04-08
Add X-elastic-product-origin header

seanstory:
Learned today about a header that is used by Beats, Kibana, and Fleet, to help breakdown telemetry and identify product traffic vs customer traffic.

#### Pre-Review Checklist


artem-shelkovnikov (approved):

seanstory:
> Should we backport it too?

Nah, then we'll get misleading version quantities. Ask ourselves a year from now "why is 8.13 usage so much lower than 8.14, but 8.15 is about the same?" Better to have a clear starting point.

PR #2364, opened 2024-04-08
Attempt to handle errors from Sink

seanstory:
related to https://github.com/elastic/connectors/issues/2348

We noticed that errors coming from Sink were not being surfaced properly. This PR works to ensure that any errors that occur in our Extractor or our Sink are surfaced through the SyncOrchestrator, and lead to quick job termination.

This may address the recent uptick in the prevalence of "Sync has not seen any update in some time" errors, as cases where the Sink crashes without terminating the Sync must eventually end in timeouts.

#### Pre-Review Checklist

## Release Note

Improved error handling when batching operations to send to Elasticsearch.


connectors/es/sink.py
```diff
@@ -149,6 +150,9 @@ def _bulk_op(self, doc, operation=OP_INDEX):
 
     @tracer.start_as_current_span("_bulk API call", slow_log=1.0)
     async def _batch_bulk(self, operations, stats):
+        msg = "I think this will reproduce"
+        raise Exception(msg)
```
seanstory, line 154: I need to remove this, but this is how I've been testing my changes. This error throws every time `_batch_bulk` is called, and IMO it should crash the sync the _first_ time it's called.

connectors/es/sink.py
```diff
@@ -309,18 +313,27 @@ async def _run(self):
                         self._batch_bulk,
                         copy.copy(batch),
                         copy.copy(stats),
-                    )
+                    ),
+                    result_callback=functools.partial(self.batch_bulk_callback),
```
seanstory, line 317: Alternatively, we could set `self.sink_error` inside a try/except in `_batch_bulk()`. That's how the Extractor works today.

connectors/utils.py
```diff
@@ -425,10 +425,10 @@ def _callback(self, task, result_callback=None):
                 exc_info=True,
             )
         if result_callback is not None:
-            result_callback(task.result())
+            result_callback(task)
         # global callback
         if self.results_callback is not None:
-            self.results_callback(task.result())
+            self.results_callback(task)
```
seanstory, line 431: This was broken. `task.result()` throws if the task has an exception. And since this exception doesn't have an error handling, any task errors just get silently swallowed into the void.

connectors/utils.py
```diff
@@ -465,9 +465,9 @@ def try_put(self, coroutine, result_callback=None):
             return self._add_task(coroutine, result_callback=result_callback)
         return None
 
-    async def join(self):
+    async def join(self, raise_on_error=False):
         """Wait for all tasks to finish."""
-        await asyncio.gather(*self.tasks, return_exceptions=True)
+        await asyncio.gather(*self.tasks, return_exceptions=(not raise_on_error))
```
seanstory, line 470: `return_exceptions=True` feels like a bad default, but at least something that should be controllable. `gather()`'s own default is `return_exceptions=False`, so IDK why we'd want to have the opposite.

I thought that making the parameter more explicit (`raise_on_errror`) might help clarify what this does.

connectors/sync_job_runner.py
```diff
@@ -326,14 +326,14 @@ async def _execute_content_sync_job(self, job_type, bulk_options):
...
-        if self.sync_orchestrator is not None and not self.sync_orchestrator.done():
-            await self.sync_orchestrator.cancel()
         if self.job_reporting_task is not None and not self.job_reporting_task.done():
             self.job_reporting_task.cancel()
             try:
                 await self.job_reporting_task
             except asyncio.CancelledError:
                 self.sync_job.log_debug("Job reporting task is stopped.")
+        if self.sync_orchestrator is not None and not self.sync_orchestrator.done():
+            await self.sync_orchestrator.cancel()
```
seanstory, line 336: @artem-shelkovnikov this is the change you suggested, and it does seem to get rid of the `task exception was never received` issue on ctrl+c

connectors/sync_job_runner.py
```diff
@@ -189,11 +189,11 @@ async def execute(self):
             while not self.sync_orchestrator.done():
                 await self.check_job()
                 await asyncio.sleep(JOB_CHECK_INTERVAL)
-            fetch_error = self.sync_orchestrator.fetch_error()
-            sync_status = (
-                JobStatus.COMPLETED if fetch_error is None else JobStatus.ERROR
-            )
-            await self._sync_done(sync_status=sync_status, sync_error=fetch_error)
+                if self.sync_orchestrator.get_error() is not None:
+                    break
```
seanstory, line 193: IMO, if the sync_orchestrator hits an error, we should terminate the job - not keep running until "done".

connectors/es/sink.py
```diff
@@ -913,3 +938,20 @@ async def async_bulk(
...
+            )
+            self._sink_error = task.exception()
+
+    def extractor_task_callback(self, task):
+        if task.exception():
+            self._logger.error(
+                f"Encountered an error in the sync's Extractor: {task.get_name()}: {task.exception()}",
+                exc_info=True,
+            )
+            self._extractor_error = task.exception()
```
seanstory, line 957: These feel kinda redundant, but since these tasks are called without ConcurrentTasks, there's nothing to log if there was an exception in these that wasn't part of their internal ConcurrentTasks workloads.

connectors/es/sink.py
```diff
@@ -824,8 +839,15 @@ def ingestion_stats(self):
...
-    def fetch_error(self):
-        return None if self._extractor is None else self._extractor.fetch_error
+    def get_error(self):
+        return (
+            None
+            if self._extractor is None
+            else self._extractor.extractor_error
+            or self._extractor_error
+            or self._sink.sink_error
+            or self._sink_error
```
seanstory, line 849: Alternatively, we could try to union multiple errors if more than one of this is non-None?

connectors/es/sink.py
```diff
@@ -361,7 +374,7 @@ def __init__(
         self.total_docs_updated = 0
         self.total_docs_created = 0
         self.total_docs_deleted = 0
-        self.fetch_error = None
+        self.extractor_error = None
```
seanstory, line 377: Extractor used to be called Fetcher, so I assume `fetch_error` was a vestigial name. It doesn't make sense, so I think we should rename it. I'd also support it just being `self.error` - WDYT?

connectors/es/sink.py
```diff
@@ -309,18 +313,27 @@ async def _run(self):
...
+                    result_callback=functools.partial(self.batch_bulk_callback),
                 )
                 batch.clear()
                 stats = {OP_INDEX: {}, OP_UPSERT: {}, OP_DELETE: {}}
                 bulk_size = 0
 
             await asyncio.sleep(0)
 
-        await self.bulk_tasks.join()
+        await self.bulk_tasks.join(raise_on_error=True)
```
seanstory, line 325: We shouldn't wait for all the `bulk_tasks` to complete before we decide if there was an error. If any of them error, the sync should fail immediately. Right?

timgrein: Fail fast makes sense to me 👍

connectors/es/sink.py
timgrein: `self.error` sounds good to me, we're already in `Extractor` class, so I don't see a need to prefix it with itself

connectors/es/sink.py
timgrein: Failing fast sounds good to me 👍

connectors/es/sink.py
timgrein: I think `self.error` makes more sense, we're already in the `Extractor` class, so I don't see a need to prefix the field with the class name

connectors/es/sink.py
```diff
@@ -824,8 +839,15 @@ def ingestion_stats(self):
...
 
-    def fetch_error(self):
-        return None if self._extractor is None else self._extractor.fetch_error
+    def get_error(self):
+        return (
+            None
+            if self._extractor is None
+            else self._extractor.extractor_error
+            or self._extractor_error
+            or self._sink.sink_error
```
timgrein, line 848: To keep things consistent that should probably be also `self._sink.error`, if we decide to go with `error` in the `Extractor`

connectors/es/sink.py
```diff
@@ -913,3 +938,20 @@ async def async_bulk(
             enable_bulk_operations_logging=enable_bulk_operations_logging,
         )
         self._sink_task = asyncio.create_task(self._sink.run())
+        self._sink_task.add_done_callback(functools.partial(self.sink_task_callback))
+
+    def sink_task_callback(self, task):
+        if task.exception():
+            self._logger.error(
+                f"Encountered an error in the sync's Sink: {task.get_name()}: {task.exception()}",
```
timgrein, line 946:
```suggestion
                f"Encountered an error in the sync's {Sink.__class__.__name__}: {task.get_name()}: {task.exception()}",
```

nit: makes renaming a bit safer

connectors/es/sink.py
```diff
@@ -913,3 +938,20 @@ async def async_bulk(
...
+                exc_info=True,
+            )
+            self._sink_error = task.exception()
+
+    def extractor_task_callback(self, task):
+        if task.exception():
+            self._logger.error(
+                f"Encountered an error in the sync's Extractor: {task.get_name()}: {task.exception()}",
```
timgrein, line 954:
```suggestion
                f"Encountered an error in the sync's {Extractor.__class__.__name__}: {task.get_name()}: {task.exception()}",
```

nit: same rationale as above

timgrein (commented):
I like the change, left some minor comments

connectors/sync_job_runner.py
artem-shelkovnikov: `done` here actually repeats "done" concept of task object: https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.done

> A Task is done when the wrapped coroutine either returned a value, raised an exception, or the Task was cancelled.

So this code will theoretically never execute.

connectors/es/sink.py
artem-shelkovnikov, line 317: I'd honestly just avoid callbacks and do it with try/except. I feel like callbacks add a lot of complexity while actually needed only for niche cases (such as events and all sorts of weak cohesion between unknown number of unknown actors)

connectors/es/sink.py
artem-shelkovnikov: +1 to that

connectors/es/sink.py
artem-shelkovnikov: I think it's nice to have different errors reported differently (speaking of error categorisation for me):

Sink errors are a category of "We were unable to upload data to Elasticsearch", so IngestionError.
Extractor errors are a category of "We failed to get data from 3rd-party for whatever reason", so ConnectorError or ThirdPartyError.

Would it be reasonable to split it here and later I add different handling into SyncJobRunner?

connectors/utils.py
artem-shelkovnikov: How does this practically impact us now? Will it change any behaviour?

connectors/utils.py
artem-shelkovnikov: I checked and did not find any usage of result callbacks in code, might be worth to just delete it?

artem-shelkovnikov (commented):
Great catch with finding the problem! Left a couple comments, IMO we can simplify this PR a bit and let's 🚢 then!

connectors/es/sink.py
seanstory: How do you mean splitting it? They're already split (`_extractor.error` vs `_sink.error`). Are you wanting to just wrap those errors to that the source is clear when the message is displayed in kibana?

connectors/sync_job_runner.py
seanstory: hmm. I'll have to try again, but I'm pretty sure that I'd put this here because I was seeing ERROR logs every 500 docs, but the sync didn't stop until the Extractor stopped extracting (for my slack test, after fetching 30k docs, but literally 0 of them being able to get to ES)

connectors/es/sink.py
artem-shelkovnikov: Just making two methods - `sink_error` and `extractor_error` in this class should be good enough, and do the "or" logic handling in the caller of orchestrator (SyncJobRunner)

connectors/utils.py
seanstory: nope. I should add that, good point.

connectors/utils.py
seanstory: Ok, I've removed the _result_callback stuff from ConcurrentTasks.

It was used heavily in the tests, so you'll still see callbacks there, but I figure that's preferable to totally rewriting all those tests.

connectors/utils.py
```diff
@@ -422,52 +419,51 @@ def _callback(self, task, result_callback=None):
...
+
+    def raise_any_exception(self):
+        for task in self.tasks:
+            if task.done() and not task.cancelled():
+                if task.exception():
+                    logger.error(
+                        f"Exception found for task {task.get_name()}: {task.exception()}",
+                    )
+                    raise task.exception()
```
artem-shelkovnikov, line 466: It looks good, but also thinking out loud - what if there are multiple errors?

We're _probably_ okay raising only the first one.

Alternatives:
- Collect and log all errors
- Raise the first one (mentioning that we're raising the first one)
- Wrap all errors in one big errors

I think we can just roll with what we have here and address this later

tests/test_utils.py
```diff
@@ -259,16 +259,17 @@ def test_try_acquire():
 async def test_concurrent_runner():
     results = []
 
-    def _results_callback(result):
-        results.append(result)
+    def _results_callback(task):
+        results.append(task.result())
```
artem-shelkovnikov, line 263: Should we drop all these tests that were relying on callbacks?

connectors/utils.py
seanstory: my _guess_ if we have multiple errors is that they'll all be the same. And since they should already be getting logged by the default callback defined in the ConcurrentTasks abstraction, I think we're also ok with this as is. But I agree that this would be a good spot to come back to if/when we have something like the WS Monitor that can hold multiple errors in state for a diagnostic bundle.

tests/test_utils.py
seanstory: no, because they're actually testing that functionality in ConcurrentTasks is working. We could re-write them to have the `coroutine` functions directly modify the `results` variables, but I thought this would be easier to review since it wasn't much rewriting of logic.

connectors/sync_job_runner.py
seanstory: Ok, I see how this is _supposed_ to repeat the async "done" concept, but it has a logic bug currently:

```
    def done(self):
        if self._extractor_task is not None and not self._extractor_task.done():
            return False
        if self._sink_task is not None and not self._sink_task.done():
            return False
        return True
```

This short-circuits and returns `false` if just the first condition is met. So if the Sink has an error, we still say "we're not done!" because the  the Extractor isn't done.

So I'll just move the `get_error()` check _here_, so that the "done" pattern can still be followed, but we avoid the logical issue. Sound good?

connectors/es/sink.py
seanstory: That's what we landed on

connectors/es/sink.py
seanstory: I no longer think these are redundant, and they're actually critical for making sure that some niche edge case in our framework code doesn't lead to silent failures.

connectors/es/sink.py
seanstory: We'll leave that idea for later.

connectors/es/sink.py
seanstory: Done.

connectors/es/sink.py
seanstory: done

connectors/es/sink.py
artem-shelkovnikov: Yeah let's have it in one place for now, makes sense!

seanstory:
This PR is ready for review

connectors/es/sink.py
```diff
@@ -854,6 +883,7 @@ async def cancel(self):
             self._sink_task.cancel()
         if self._extractor_task_running():
             self._extractor_task.cancel()
+        self.canceled = True
```
artem-shelkovnikov, line 886: Is it used for anything? I only see assignments in this PR, but not usage

artem-shelkovnikov (commented):
Intuitively it looks good, left 1 small question.

All in all it's a good contained change with potential to bring subtle problems - I'm voting to merge it asap for 8.15 branch to have time before release to test it :)

connectors/es/sink.py
seanstory: it's used here: https://github.com/elastic/connectors/pull/2364/files#diff-f4def1564167603a89d66578a9ac61c61f5d0c63298e1ceab10b22f3d949ff8eR347

seanstory:
> I'm voting to merge it asap for 8.15 branch to have time before release to test it :)

I'm torn between only putting it in 8.15 vs getting it in for 8.14 as a bugfix. This should help us surface the ~100 daily errors of "Sink has not progressed in some time" as the actual underlying problem. But I agree that there's a higher risk of crashing syncs.

artem-shelkovnikov:
> I'm torn between only putting it in 8.15 vs getting it in for 8.14 as a bugfix. This should help us surface the ~100 daily errors of "Sink has not progressed in some time" as the actual underlying problem. But I agree that there's a higher risk of crashing syncs.

I totally trust your judgement on this one

seanstory:
Backporting is wonky right now between creating the 8.14 branch and bumping versions for 8.15. Manual backport: https://github.com/elastic/connectors/pull/2402

PR #2365, opened 2024-04-08
Fixing connection issue when JIRA has a path in URL (like jira.company.com/jira)

Gleb-Gadyatskiy:
parse.urljoin treats the first argument as host name only. Which won't work when it is host name + path (when JIRA is installed not as Root Web App). For example - https://jira.company.com/jira. The old code ignore path '/jira' and, as a result , fails to connect to JIRA on https://jira.company.com 

The proposed change free from this defect and allows to communicate with JIRA in both cases.

### DIFF
--- a/connectors/sources/jira.py
+++ b/connectors/sources/jira.py
@@ -70,7 +70,8 @@ URLS = {
     ATTACHMENT_CLOUD: "/rest/api/2/attachment/content/{attachment_id}",
     ATTACHMENT_SERVER: "/secure/attachment/{attachment_id}/{attachment_name}",
     USERS: "/rest/api/3/users/search",
-    USERS_FOR_DATA_CENTER: "/rest/api/latest/user/search?username=''&startAt={start_at}&maxResults={max_results}",  # we can fetch only 1000 users for jira data center. Refer this doc see the limitations: https://auth0.com/docs/manage-users/user-search/retrieve-users-with-get-users-endpoint#limitations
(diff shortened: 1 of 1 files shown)


moxarth-elastic:
buildkite test this

moxarth-elastic:
@Gleb-Gadyatskiy You need to update the same at the following locations o/w the connector will produce the same error there
https://github.com/elastic/connectors/blob/838ce4b9c20b02a5e6fcca0f544d834727ce4553/connectors/sources/jira.py#L253-L259
https://github.com/elastic/connectors/blob/838ce4b9c20b02a5e6fcca0f544d834727ce4553/connectors/sources/jira.py#L648

artem-shelkovnikov:
Thanks for your contribution, @Gleb-Gadyatskiy!

After your fix is ready you can run `make autoformat` to fix linting errors automatically - now you can see CI reports failure at linting stage

seanstory:
related to https://github.com/elastic/connectors/issues/2343

Gleb-Gadyatskiy:
Changes applied. 
Thank you

seanstory:
@Gleb-Gadyatskiy https://github.com/elastic/connectors/pull/2377 has just merged, which I believe addresses your issue. Can you try latest, and see if it's sufficient, or if we need additional changes on top of those?

Gleb-Gadyatskiy:
Confirmed - works fine

PR #2370, opened 2024-04-09
[CLI] Use constant for default scheduling interval

timgrein:
Small change to use a constant `EVERYDAY_AT_MIDNIGHT` instead of raw strings `0 0 0 * * ?`, so you can immediately understand the cron quartz expression.

### DIFF
--- a/connectors/cli/connector.py
+++ b/connectors/cli/connector.py
@@ -12,6 +12,8 @@ from connectors.protocol import (
 from connectors.source import get_source_klass
 from connectors.utils import iso_utc
 
+EVERYDAY_AT_MIDNIGHT = "0 0 0 * * ?"
+
 
 class IndexAlreadyExists(Exception):
     pass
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov (commented):
I'm not sure about this, it's good on its own without constants IMO

timgrein:
> I'm not sure about this, it's good on its own without constants IMO

Hmm, don't you think it's easier to understand? I'm not sure, if everyone immediately understands `0 0 0 * * ?`.

artem-shelkovnikov:
> Hmm, don't you think it's easier to understand? I'm not sure, if everyone immediately understands 0 0 0 * * ?.

It's okay to not understand it too immediately IMO - it's somewhat (hopefully) obvious that it's a cron expression and from there crontab.guru can help.

I'm also not opposed to merging it either, I just personally don't like constants unless they are really needed.

artem-shelkovnikov (approved):

timgrein:
> > Hmm, don't you think it's easier to understand? I'm not sure, if everyone immediately understands 0 0 0 * * ?.
> 
> It's okay to not understand it too immediately IMO - it's somewhat (hopefully) obvious that it's a cron expression and from there crontab.guru can help.
> 
> I'm also not opposed to merging it either, I just personally don't like constants unless they are really needed.

Thanks for the explanation and also for approving 🤝  I still think readability is a good reason here to have it, so merging it.

PR #2371, opened 2024-04-09
[Service CLI] Log exception, if uvloop couldn't be enabled

timgrein:
From the string it looks like we would've wanted to log the exception `[...] {e} [...]`, but we actually didn't, so adjusting it.


artem-shelkovnikov (approved):
Nice catch!

PR #2372, opened 2024-04-09
Use graphql method of GitHubAPI for GraphQL requests

wangch079:
### What's the change

1. Remove the `headers` of `aiohttp.ClientSession` object as this is not necessary. `GitHubAPI` will build its own header
2. For GraphQL, instead of calling `self._get_session.post`, use the built-in graphql method of `GitHubAPI`
3. Create a new method `get_personal_access_token_scopes`, which calls the low-level `_request` method to issue a `HEAD` request to endpoint `/`. The Github connector needs to get the scopes of the personal access token from the header, and the built-in methods of `GitHubAPI` don't return header so we have to call `_request` instead.

#### Pre-Review Checklist


timgrein (approved):
Just two minor comments, rest LGTM

connectors/sources/github.py
```diff
@@ -714,35 +716,18 @@ async def post(self, query_data, need_headers=False):
         """
         url = f"{self.github_url}/graphql"
         self._logger.debug(
-            f"Sending POST to {url} with body: '{json.dumps(query_data)}'"
+            f"Sending POST to {url} with query: '{json.dumps(query)}' and variables: '{json.dumps(variables)}'"
```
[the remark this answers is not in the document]
wangch079: `query` itself won't contain any sensitive information.

Common `variables` are something like `org_name`, `user_login` for a personal account, `repo_name`, so I think they are safe to log as well.

connectors/sources/github.py
```diff
@@ -775,44 +760,55 @@ async def get_github_item(self, resource):
...
         while True:
-            query_data = {"query": query, "variables": variables}
-            response = await self.post(query_data=query_data)
+            response = await self.graphql(query=query, variables=variables)
             yield response
 
-            page_info = nested_get_from_dict(
-                response, ["data"] + keys + ["pageInfo"], default={}
-            )
+            page_info = nested_get_from_dict(response, keys + ["pageInfo"], default={})
```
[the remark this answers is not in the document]
wangch079: An exmaple of `pageInfo`:
```json
"pageInfo": {
  "hasNextPage": true,
  "endCursor": "Y3Vyc29yOnYyOpHOAfmiog=="
}
```

We can log something like 
```
Fetching next page with cursor 'Y3Vyc29yOnYyOpHOAfmiog=='
```

But I don't feel it's super useful

connectors/sources/github.py
```diff
@@ -775,44 +760,55 @@ async def get_github_item(self, resource):
...
+            self._get_client.requester,
+            accept=sansio.accept_format(),
+            oauth_token=self.github_token,
+        )
+        _, headers, _ = await self._get_client._request(
+            "HEAD", self.github_url, request_headers
+        )
+        scopes = headers.get("X-OAuth-Scopes")
+        if not scopes or not scopes.strip():
+            return set()
```
seanstory, line 797: Should this log a warning?

tests/sources/fixtures/github/connector.json
```diff
@@ -29,28 +29,14 @@
...
-                        "display": "dropdown",
-                        "label": "Authentication method",
-                        "options": [
-                            {"label": "Personal access token", "value": "personal_access_token"},
-                            {"label": "GitHub App", "value": "github_app"}
-                        ],
-                        "order": 3,
-                        "type": "str",
-                        "value": "personal_access_token"
-                },
```
seanstory, line 47: Why's this being removed? I still see it in the `get_default_configuration` for github.py

connectors/sources/github.py
```diff
@@ -775,44 +760,55 @@ async def get_github_item(self, resource):
...
+    async def get_personal_access_token_scopes(self):
+        request_headers = sansio.create_headers(
+            self._get_client.requester,
+            accept=sansio.accept_format(),
+            oauth_token=self.github_token,
+        )
+        _, headers, _ = await self._get_client._request(
+            "HEAD", self.github_url, request_headers
+        )
+        scopes = headers.get("X-OAuth-Scopes")
```
seanstory, line 795: headers are technically case insensitive. But `get` is not. It probably doesn't matter, because Github's unlikely to change this? But commenting because this has bit us before with Zendesk.

seanstory (approved):
Few minor comments, which can be addressed separately (if at all), since I know this is blocking your other work

connectors/sources/github.py
wangch079: Will add in the next PR

tests/sources/fixtures/github/connector.json
wangch079: Initally, I have [the change to connector.json](https://github.com/elastic/connectors/pull/2374/files) in this PR, then I extracted the change to a separate PR, and reverted [here](https://github.com/elastic/connectors/pull/2372/commits/7a6b8877c1f88e0a8f1e56fe19e7e4b966374d9f) in this PR.

I guess this is because of the merge of PR https://github.com/elastic/connectors/pull/2374? I will try to update branch and see if it can remove the change.

wangch079: :sweat_smile: No it's still there, I will merge this PR and address it in a separate PR.

wangch079: Follow-up PR https://github.com/elastic/connectors/pull/2389

connectors/sources/github.py
wangch079: Added in PR https://github.com/elastic/connectors/pull/2381

https://github.com/elastic/connectors/blob/75e8e51c610ba1da8e180ed486dafe89ebc7a5ca/connectors/sources/github.py#L859

PR #2375, opened 2024-04-10
[Service CLI] Migrate to click

timgrein:
## Part of https://github.com/elastic/enterprise-search-team/issues/6442

This PR migrates the `service_cli.py` to use the `click` library without changing any messages or behavior. I've also closed some testing gaps. Follow-ups, which will be kept separate from this PR to keep reviews focused:

- Improve help messages
- Consolidate duplicate logic between `connectors_cli.py` and `services_cli.py`

#### Pre-Review Checklist

### DIFF
--- a/connectors/service_cli.py
+++ b/connectors/service_cli.py
@@ -16,7 +16,9 @@ import json
 import logging
 import os
 import signal
-from argparse import ArgumentParser
+
+import click
+from click import ClickException, UsageError
 
 from connectors import __version__
 from connectors.config import load_config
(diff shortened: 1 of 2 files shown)


connectors/service_cli.py
```diff
@@ -180,44 +102,42 @@ def run(args):
...
     set_logger(
-        args.log_level or config["service"]["log_level"] or logging.INFO,
-        filebeat=args.filebeat,
+        log_level or config["service"]["log_level"] or logging.INFO,
+        filebeat=filebeat,
     )
 
     # just display the list of connectors
-    if args.action == ["list"]:
+    if action == ("list",):
```
artem-shelkovnikov, line 125: Why is it a tuple comparison? Action is also a tuple?

connectors/service_cli.py
```diff
@@ -226,16 +146,16 @@ def run(args):
         print("Bye")  # noqa: T201
         return 0
```
artem-shelkovnikov, line 147: Do we still need to return?

connectors/service_cli.py
timgrein: Yes it's a tuple:

image

connectors/service_cli.py
timgrein: For now yes, otherwise we run into `if "config" in action:` down below. I'll plan to address the flow of execution in a separate PR as some things don't make too much sense IMO.

artem-shelkovnikov (approved):

PR #2379, opened 2024-04-10
[CLI] Track CLI usage through user agent header

timgrein:
## Part of https://github.com/elastic/enterprise-search-team/issues/6442

This PR adds a new property `user_agent` to the `ESClient` class, which can be overridden, so we're able to track different origins. `CLIClient` overrides this property to `connectors-cli` so we're able to track CLI usage. Also added one missing test to `ESClient`.

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/client.py
+++ b/connectors/es/client.py
@@ -29,8 +29,6 @@ from connectors.utils import (
     time_to_sleep_between_retries,
 )
 
-X_ELASTIC_PRODUCT_ORIGIN_HEADER = "X-elastic-product-origin"
-
 
 class License(Enum):
     ENTERPRISE = "enterprise"
@@ -43,7 +41,7 @@ class License(Enum):
 
 
 class ESClient:
-    product_origin = "connectors"
+    user_agent = f"elastic-connectors-{__version__}"
 
     def __init__(self, config):
         # We don't have a way to ask the server, but it's planned
@@ -107,10 +105,8 @@ class ESClient:
         self.backoff_multiplier = config.get("backoff_multiplier", 2)
 
         options["headers"] = config.get("headers", {})
-        options["headers"]["user-agent"] = f"elastic-connectors-{__version__}"
-        options["headers"][
-            X_ELASTIC_PRODUCT_ORIGIN_HEADER
-        ] = self.__class__.product_origin
+        options["headers"]["user-agent"] = self.__class__.user_agent
+        options["headers"]["X-elastic-product-origin"] = "connectors"
 
         self.client = AsyncElasticsearch(**options)
         self._keep_waiting = True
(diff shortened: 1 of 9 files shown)


connectors/es/client.py
```diff
@@ -101,9 +103,11 @@ def __init__(self, config):
         self.max_wait_duration = config.get("max_wait_duration", 60)
         self.initial_backoff_duration = config.get("initial_backoff_duration", 5)
         self.backoff_multiplier = config.get("backoff_multiplier", 2)
+
         options["headers"] = config.get("headers", {})
-        options["headers"]["user-agent"] = f"elastic-connectors-{__version__}"
+        options["headers"]["user-agent"] = self.__class__.user_agent
```
seanstory, line 108: Should we establish a "base" with both usages providing a suffix? Or should only the CLI have a suffix?
Like should the normal agent be changed to have a suffix like: 
```
        options["headers"]["user-agent"] = f"{self.__class__.user_agent}/service"
```

tests/es/test_cli_client.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_overrides_user_agent_header():
+    config = {
+        "username": "elastic",
+        "password": "changeme",
+        "host": "http://nowhere.com:9200",
+    }
+    cli_client = CLIClient(config)
+
+    assert cli_client.client._headers["user-agent"] == CLIClient.user_agent
```
seanstory, line 12: Might want to assert against a string literal, since this test will pass even if the user_agent isn't what you want it to be

connectors/es/client.py
timgrein: I like having `/service` as an explicit suffix, does not leave room for interpretation 👍

seanstory (approved):
nice

connectors/es/client.py
```diff
@@ -40,7 +40,12 @@ class License(Enum):
     UNSET = None
 
 
+USER_AGENT_BASE = f"elastic-connectors-{__version__}"
+
+
 class ESClient:
```
wangch079, line 46: `ESManagementClient` shares the same user agent with `ESClient`, and there's no way to differentiate them.

`ESClient` is used here:

https://github.com/elastic/connectors/blob/34ff685d7d83d4ab251ff7bb275fcbfb6c7ef479/connectors/es/index.py#L20

timgrein: > ESManagementClient shares the same user agent with ESClient, and there's no way to differentiate them.

I'm not quite sure, if I understand it correctly. Could you elaborate what's the potential problem?

timgrein: Yeah, we don't need to distinguish between `ESManagementClient` and `ESClient`. We just want to know, whether the `service` or the `cli` made a call.

PR #2380, opened 2024-04-10
[Confluence] Ensure end signals are added to queue if exception occurs

praveen-elastic:
## Closes https://github.com/elastic/connectors-py/issues/2353

As mentioned in the bug, if [fetch_attachments](https://github.com/elastic/connectors/blob/4047c2c6edf97ae8cfc9c846b3ec62e0c4f88157/connectors/sources/confluence.py#L805) fails, then the corresponding END_SIGNAL is never put into the queue in [_attachment_coro](https://github.com/elastic/connectors/blob/4047c2c6edf97ae8cfc9c846b3ec62e0c4f88157/connectors/sources/confluence.py#L937).
As a consequence, the fetcher_count will never reach 0 causing an infinite loop. 

To avoid this, ensure end-signals are put to queue even if error is encountered, in all relevant places (space, page/blog, attachment).

#### Pre-Review Checklist


connectors/sources/confluence.py
```diff
@@ -941,50 +941,62 @@ async def _attachment_coro(self, document, access_control):
...
+                        partial(
+                            self.download_attachment,
+                            download_link[1:],
+                            copy(attachment),
+                        ),
+                    )
+                )
+        except Exception as exception:
+            self._logger.exception(
+                f"Error while fetching attachments of {document.get('title')}: {exception}"
```
seanstory, line 966: can we add the ID, space, and type the the log? That will give us the info we need to be able to reproduce the error after the fact.

connectors/sources/confluence.py
```diff
@@ -993,46 +1005,55 @@ async def _page_blog_coro(self, api_query, target_type):
...
+                if attachment_count > 0:
+                    await self.fetchers.put(
+                        partial(self._attachment_coro, copy(document), access_control)
                     )
-            document = self._decorate_with_access_control(
-                document=document, access_control=access_control
+                    self.fetcher_count += 1
+        except Exception as exception:
+            self._logger.exception(
+                f"Error while fetching pages and blogposts: {exception}"
```
seanstory, line 1052:
```suggestion
                f"Error while fetching pages and blogposts with query '{api_query}': {exception}"
```

adding more info so that we can better reproduce if there's an issue

seanstory (commented):
looks good. Some minor nits

seanstory (approved):

PR #2381, opened 2024-04-11
Update GitHubClient to support GitHub App authentication

wangch079:
#### How does GitHub App authentication work

1. A GitHub app can be installed on multiple organizations/personal accounts, each called an installation.
2. To list all installations of a GitHub app, you need to call [List installations for the authenticated app](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#list-installations-for-the-authenticated-app) API with a jwt token generated from the GitHub App ID and a private key
3. To access content from an installation (on either an org or a personal account), you have to [Create an installation access token for an app](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-an-installation-access-token-for-an-app), which is valid for one hour.
4. With the installation access token, you can access the repos/PRs/issues/content (based on the permissions granted) of the org/personal account (This is where the original connector logic can be reused)

#### What's the changes:

1. Made the constructor  method `__init__` of `GitHubClient` receive individual params instead of `configuration`, so that it's decoupled from `DataSourceConfiguration` class
5. Made `GitHubClient` not hold information like configured `repo_type`, `org_name` and `repositories` as those information should stay in `GitHubDataSource`
6. Made `GitHubClient` support both `GitHub App` and `Personal Access Token` authentication methods
7. Instead of providing an access token when initializing `GitHubAPI`, provide the access token when making a request. Because the installation access token expires in one hour, and this makes it easier to refresh
8. Made `GitHub App` authenticated `GitHubClient` to hold an installation id variable. When a user want to query content from an installation, he/she has to call `update_installation_id`, which will update the installation id variable, and regenerate an installation access token.
9. Added a method `get_installations` to list all the installations of a GitHub App. 

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -763,6 +763,7 @@ class GitHubClient:
             f"Sending POST to {url} with query: '{json.dumps(query)}' and variables: '{json.dumps(variables)}'"
         )
         try:
+            self._get_client.oauth_token = self._access_token()
             return await self._get_client.graphql(
                 query=query, endpoint=url, **variables or {}
             )
@@ -846,13 +847,14 @@ class GitHubClient:
         request_headers = sansio.create_headers(
             self._get_client.requester,
             accept=sansio.accept_format(),
-            oauth_token=self.github_token,
+            oauth_token=self._access_token(),
         )
         _, headers, _ = await self._get_client._request(
-            "HEAD", self.github_url, request_headers
+            "HEAD", self.base_url, request_headers
         )
         scopes = headers.get("X-OAuth-Scopes")
         if not scopes or not scopes.strip():
+            self._logger.warning(f"Couldn't find 'X-OAuth-Scopes' in headers {headers}")
             return set()
         return {scope.strip() for scope in scopes.split(",")}
 
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -12,7 +12,7 @@ from functools import cached_property, partial
 import aiohttp
 import fastjsonschema
 from aiohttp.client_exceptions import ClientResponseError
-from gidgethub import sansio
+from gidgethub import RateLimitExceeded, sansio
 from gidgethub.abc import (
     BadGraphQLRequest,
     GraphQLAuthorizationFailure,
@@ -668,7 +668,9 @@ class GitHubClient:
 
     async def _get_retry_after(self, resource_type):
         current_time = time.time()
-        response = await self._get_client.getitem("/rate_limit")
+        response = await self._get_client.getitem(
+            "/rate_limit", oauth_token=self._access_token()
+        )
         reset = nested_get_from_dict(
             response, ["resources", resource_type, "reset"], default=current_time
         )
@@ -813,10 +815,10 @@ class GitHubClient:
                     raise
                 msg = "Your Github token is either expired or revoked. Please check again."
                 raise UnauthorizedException(msg) from exception
-            elif self.get_rate_limit_encountered(exception.status, exception):
-                await self._put_to_sleep("core")
             else:
                 raise
+        except RateLimitExceeded:
+            await self._put_to_sleep("core")
         except Exception:
             raise
 
@@ -874,11 +876,9 @@ class GitHubClient:
                 sansio.accept_format(),
                 get_jwt(app_id=self.app_id, private_key=self.private_key),
             )
-        except ClientResponseError as exception:
-            if self.get_rate_limit_encountered(exception.status, exception):
-                await self._put_to_sleep("core")
-            else:
-                raise
+        # we don't expect any 401 error as the jwt is freshly generated
+        except RateLimitExceeded:
+            await self._put_to_sleep("core")
         except Exception:
             raise
 
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -693,6 +693,9 @@ class GitHubClient:
 
     # update the current installation id and re-generate access token
     async def update_installation_id(self, installation_id):
+        self._logger.debug(
+            f"Updating installation id - new ID: {installation_id}, original ID: {self._installation_id}"
+        )
         self._installation_id = installation_id
         await self._update_installation_access_token()
 
@@ -704,7 +707,7 @@ class GitHubClient:
     async def _update_installation_access_token(self):
         try:
             access_token_response = await get_installation_access_token(
-                gh=self._app_client,
+                gh=self._get_client,
                 installation_id=self._installation_id,
                 app_id=self.app_id,
                 private_key=self.private_key,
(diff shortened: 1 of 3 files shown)


connectors/sources/github.py
```diff
@@ -714,15 +760,22 @@ async def graphql(self, query, variables=None):
         Yields:
             dictionary: Client response
         """
-        url = f"{self.github_url}/graphql"
+        url = f"{self.base_url}/graphql"
         self._logger.debug(
             f"Sending POST to {url} with query: '{json.dumps(query)}' and variables: '{json.dumps(variables)}'"
         )
         try:
+            self._get_client.oauth_token = self._access_token()
```
wangch079, line 768: Most of the methods in [gidgethub](https://github.com/gidgethub/gidgethub/blob/main/gidgethub/abc.py) accept params like `oauth_token` to override the default one, e.g. [getitem](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/abc.py#L129-L138), but [graphql](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/abc.py#L306-L312) doesn't and it [uses the default one](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/abc.py#L323)

I update the default `oauth_token` every time we make a `graphql` call, but in the future, we can create a pull request in gidgethub

connectors/sources/github.py
```diff
@@ -659,7 +668,9 @@ def get_rate_limit_encountered(self, status_code, message):
 
     async def _get_retry_after(self, resource_type):
         current_time = time.time()
-        response = await self._get_client.getitem("/rate_limit")
+        response = await self._get_client.getitem(
+            "/rate_limit", oauth_token=self._access_token()
+        )
```
artem-shelkovnikov, line 673: This naming always confused me, `oauth_token` is just api key in essence?

connectors/sources/github.py
```diff
@@ -675,6 +686,42 @@ async def _put_to_sleep(self, resource_type):
...
+        self._installation_id = installation_id
+        await self._update_installation_access_token()
+
+    @retryable(
+        retries=RETRIES,
+        interval=RETRY_INTERVAL,
+        strategy=RetryStrategy.EXPONENTIAL_BACKOFF,
+    )
+    async def _update_installation_access_token(self):
```
artem-shelkovnikov, line 709: What's an installation access token? Links to docs would be really helpful

connectors/sources/github.py
artem-shelkovnikov: I think it's also nice to just wrap all this logic (client calls + auth) into a class to make it easier to understand. Not for this PR probably, but for future it would help IMO

connectors/sources/github.py
```diff
@@ -714,15 +760,22 @@ async def graphql(self, query, variables=None):
...
         )
         try:
+            self._get_client.oauth_token = self._access_token()
             return await self._get_client.graphql(
                 query=query, endpoint=url, **variables or {}
             )
         except GraphQLAuthorizationFailure as exception:
+            if self.auth_method == GITHUB_APP:
+                self._logger.debug(
+                    f"The access token for installation #{self._installation_id} expired, Regenerating a new token."
```
artem-shelkovnikov, line 775: Can `self._installation_id` be none?

connectors/sources/github.py
```diff
@@ -1207,7 +1332,7 @@ async def _remote_validation(self):
             msg = "Configured token does not have required rights to fetch the content. Required scopes are 'repo', 'user', and 'read:org'."
             raise ConfigurableFieldValueError(msg)
 
-        if self.github_client.repos != [WILDCARD]:
+        if WILDCARD not in self.configured_repos:
```
artem-shelkovnikov, line 1335: Ha good catch!

artem-shelkovnikov (commented):
Looks alright, I left a couple comments-questions.

Have we tested it with both flows already?

connectors/sources/github.py
wangch079: This is the term used by `gidgethub`, to differentiate from `jwt`. `gidgethub` [uses different headers](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/sansio. Bearer`. See [here](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28#about-authentication):
>**Note**: In most cases, you can use `Authorization: Bearer` or `Authorization: token` to pass a token. However, if you are passing a JSON web token (JWT), you must use `Authorization: Bearer`.

connectors/sources/github.py
wangch079: It's just a personal preference, to handle an edge case, when a customer configures the `repositories` like `elastic/kibana, *, elastic/connectors`.

In this case, we will just sync all repos from the org.

wangch079:
>Have we tested it with both flows already?

In this PR, I only make `GitHubClient` support the GitHub App authentication, and add methods needed by GitHub App flow, e.g. `get_installations`. I have added tests for those new methods and also tested them locally.

But the GitHub App flow (in `GitHubDataSource`) will be added in other PRs, e.g. #2384, #2386

connectors/sources/github.py
wangch079: Could you elaborate more?

The Client, `GitHubAPI` does accept `oauth_token` as parameter in `__init__`, as the default token, see [here](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/abc.py#L44). And it does accept `oauth_token` in most methods, to override the default one, e.g. [getitem](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/abc.py#L136), [getiter](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/abc.py#L180), but not [graphql](https://github.com/gidgethub/gidgethub/blob/7f8cec45922056a38c51ece276844a4bc028e6f5/gidgethub/abc.py#L306)

I think the better approach is to file a PR in `gidgethub`, but I can't wait for it to be merged.

artem-shelkovnikov: I meant to make a client similar to what we have in [SPO](https://github.com/elastic/connectors/blob/main/connectors/sources/sharepoint_online.py#L521) - under the hood it tries to handle all inconsistencies and provide you with simple interface (doing token refreshes, choosing proper API under the hood).

It's of course possible to make a PR in 3rd-party lib but it definitely will not be fast

artem-shelkovnikov (approved):

PR #2386, opened 2024-04-12
Update GitHub access control sync to support GitHub App

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/7199

This PR updates the `get_access_control` method to be able to sync user access control documents when GitHub App is selected as the authentication method.

#### Pre-Review Checklist


connectors/sources/github.py
```diff
@@ -1312,10 +1313,20 @@ async def get_access_control(self):
             self._logger.warning("DLS is not enabled. Skipping")
             return
 
-        async for user in self.github_client._fetch_all_members(
-            self.configuration["org_name"]
-        ):
-            yield await self._user_access_control_doc(user=user)
+        async for org_name in self._get_orgs():
+            async for user in self.github_client._fetch_all_members(org_name):
```
wangch079, line 1317: A user may be a member of multiple organizations, and the same user will be indexed multiple times. We can filter out duplicate users by having a set but I'd like to see if there is any reported issue first before we proceed.

wangch079:
CI failed because the test coverage didn't reach 92.0%, this is because I added some methods from #2384 without tests.

The CI should be green once #2384 is merged.

connectors/sources/github.py
artem-shelkovnikov: What's the id for the user? Will it be resolved naturally somehow (e.g. we upload user multiple times, but since ids collide it's okay)

artem-shelkovnikov (dismissed):

connectors/sources/github.py
wangch079: The user id from GitHub.

If the same user belongs to multiple organizations, it will be indexed multiple times but it will only result in one document in ES. Adding a set will eliminate the multiple indexing of the same user.

seanstory (approved):
I'm rubber-stamping to replace Artem's previous review. If you have new changes that actually need re-review, let me know.

PR #2399, opened 2024-04-17
Build and push Docker image based on Chainguard base image

oli-g:
For 8.14, we want to release two flavours of `connectors` Docker images:

* the [current one](https://github.com/elastic/connectors/blob/d6fdf8d/Dockerfile#L1), based on Debian
* a [new flavour](https://github.com/elastic/connectors/pull/2399/files#diff-813ef534be9bab5e7b9ea7504eecc3058e7762f08c32462d3a52e235a9ed7c36R1) based on [Wolfi](https://www.chainguard.dev/unchained/introducing-wolfi-the-first-linux-un-distro) from [Chainguard](https://edu.chainguard.dev/chainguard/chainguard-images/overview/)

From 8.15, the new one based on Wolfi will be the default one, and we will no longer publish the current flavour based on Debian.

Please note that we are going to preserve the current `Dockerfile`, so that users will still be able to build their own custom images based on Debian: this is needed because `docker.elastic.co/wolfi/python:3.10-dev` is not a public base image, so `docker build` would fail for unauthenticated users.

#### Pre-Review Checklist

## Release Note

TBD.


mgreau (approved):
LGTM 🚀

Dockerfile.ftest.wolfi
```diff
@@ -0,0 +1,7 @@
+FROM docker.elastic.co/wolfi/python:3.10-dev
+USER root
+COPY . /connectors
+WORKDIR /connectors
+RUN make clean install
+RUN bin/pip install -r requirements/ftest.txt
+ENTRYPOINT []
```
acrewdson, line 7: nit: missing EOF newline

docs/DOCKER.md
```diff
@@ -69,11 +69,17 @@ docker build -t <TAG_OF_THE_IMAGE> .
 For example, if you've created a custom version of MongoDB connector, you can tag it with the following command:
 
 ```
-docker build -t connector/custom-mongodb:1.0 -f .
+docker build -t connector/custom-mongodb:1.0 .
 ```
 
 You can later use `<TAG_OF_THE_IMAGE>` instead of `docker.elastic.co/enterprise-search/elastic-connectors:<VERSION>-SNAPSHOT` in the next step to run the Docker image.
 
+If you're an Elastic employee, you may want to build a Chainguard-based image using `Dockerfile.wolfi`:
```
acrewdson, line 77: I'm thinking we'll want to get feedback from @elastic/ingestion-team on whether we want to call this out here (or not), and if we do, whether we need to make any changes to the wording.

/cc @seanstory

.buildkite/pipeline.yml
```diff
@@ -4,6 +4,9 @@ agents:
   useVault: true
   image: family/enterprise-search-ubuntu-2204-connectors-py
 
+env:
+  DOCKERFILE_FTEST_PATH: "Dockerfile.ftest.wolfi"
```
acrewdson, line 8: any reason not to just replace the existing `Dockerfile.ftest` with the new Wolfi-based one at this point, assuming the functional tests all run and pass using the new Wolfi-based image there?

acrewdson (commented):
awesome work! looks great to me ~~just added one question about the ftest Dockerfiles~~ ah, sorry, I realize now why we need to keep both, I assume since people may want to run the ftests outside the Buildkite context – apologies 👍 

just one thing then:
* maybe because it has just been added, looks like we need to make the ftests `dockerfile:` change to `tests/sources/fixtures/opentext_documentum/docker-compose.yml`, too

oli-g:
> maybe because it has just been added, looks like we need to make the ftests dockerfile: change to tests/sources/fixtures/opentext_documentum/docker-compose.yml, too

@acrewdson Thank you for spotting that: after the rebase I fixed `.buildkite/pipeline.yml`, but forgot to also fix `tests/sources/fixtures/opentext_documentum/docker-compose.yml`.

acrewdson (dismissed):

mgreau (approved):
LGTM

docs/DOCKER.md
vidok: Just wonder if there is still a way to build an image using the old Dockerfile? If yes, should we leave it? Or using chaingguard is the only option we have?

acrewdson: yep, the old Dockerfile is being left in place in this PR; my concern here was more to get someone from your team to confirm that we want to call out the internal-only image in this README in this way

oli-g: @vidok Yep, it's still possible to build an image using the current `Dockerfile` (so I preserved the existing documentation, some lines above).

Please also note that the same can be achieved with `make docker-build` (no changes here), and it's now also possible to build a Wolfi image with `DOCKERFILE_PATH=Dockerfile.wolfi make docker-build`: I haven't documented it because those `docker-*` Make targets are not documented at the moment (and it's OK in my opinion).

vidok (approved):
LGTM. 

It would be also useful to add some docs/links on how to update the base Wolfi image.

PR #2422, opened 2024-04-23
[Connector API] Update postman collection for 8.14

jedrazb:
Reflect endpoint changes in 8.14 in our postman collection. 

I removed the action that overrides the whole filtering object to discourage its use.


vidok (approved):
Thanks for keeping it up to date 👍🏻

PR #2441, opened 2024-04-25
Always close data source

wangch079:
When connector config validation fails, it doesn't close the data source. 

This PR makes sure the data source is always properly closed.

#### Pre-Review Checklist

### DIFF
--- a/connectors/services/job_scheduling.py
+++ b/connectors/services/job_scheduling.py
@@ -97,16 +97,15 @@ class JobSchedulingService(BaseService):
 
             connector.log_debug("Pinging the backend")
             await data_source.ping()
+
+            if connector.features.sync_rules_enabled():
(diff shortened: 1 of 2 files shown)


vidok (approved):
LGTM. 

I wonder if the datasources that have `close` implementation also initialise connections/clients. What if a datasource does not expect "close" method to be called without "get_docs" called?

wangch079:
> I wonder if the datasources that have `close` implementation also initialise connections/clients. 

Do we have such a data source or do we have a plan? I browsed the subclasses of `BaseDataSource`, and I didn't find any multiple inheritance

> What if a datasource does not expect "close" method to be called without "get_docs" called?

Hmm, if a class inherits from `BaseDataSource`, it should follow the expectation. If it does not need to do anything in `close`, it can use the implementation of the parent class

PR #2445, opened 2024-04-25
Copy config.yml from config.yml.example

wangch079:
## Closes https://github.com/elastic/enterprise-search-team/issues/7386

This PR makes 2 changes:

1. Copy `config.yml` from `config.yml.example` instead of `config.yml` due to change introduced in #2280 
2. Run `copy-config.sh` before starting Elastic connectors

#### Pre-Review Checklist

### DIFF
--- a/scripts/stack/copy-config.sh
+++ b/scripts/stack/copy-config.sh
@@ -17,7 +17,7 @@ fi
 
 is_example_config=false
 if [[ "${CONFIG_PATH:-}" == "" ]]; then
-    CONFIG_PATH="$PROJECT_ROOT/config.yml"
+    CONFIG_PATH="$PROJECT_ROOT/config.yml.example"
     is_example_config=true
 fi
 
--- a/scripts/stack/copy-config.sh
+++ b/scripts/stack/copy-config.sh
@@ -17,7 +17,8 @@ fi
 
 is_example_config=false
 if [[ "${CONFIG_PATH:-}" == "" ]]; then
-    CONFIG_PATH="$PROJECT_ROOT/config.yml.example"
+    cp -n "$PROJECT_ROOT/config.yml.example" "$PROJECT_ROOT/config.yml"
+    CONFIG_PATH="$PROJECT_ROOT/config.yml"
     is_example_config=true
 fi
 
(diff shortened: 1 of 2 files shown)


scripts/stack/copy-config.sh
```diff
@@ -17,7 +17,7 @@ fi
 
 is_example_config=false
 if [[ "${CONFIG_PATH:-}" == "" ]]; then
-    CONFIG_PATH="$PROJECT_ROOT/config.yml"
+    CONFIG_PATH="$PROJECT_ROOT/config.yml.example"
```
seanstory, line 20: I'd rather not do this, because if the user has been running from source, they may have already made the config changes they want to `config.yml`, and it would be nice for them to still pick these up.

scripts/stack/run-stack.sh
```diff
@@ -58,6 +58,7 @@ fi
 if [ "${no_connectors:-}" == false ]; then
   echo "Starting Elastic Connectors..."
 
+  source ./copy-config.sh
```
seanstory, line 61: This is done a few lines up if `$run_configurator == "yes"`. What if instead we pull that out of the `if` block, and move this `source ./copy-config.sh` call up to line 41? 

Also, I think we should edit `copy-config.sh` to make a call to `make config` to ensure that `config.yml` exists.

seanstory (commented):
I'd dropped some comments in the issue before I saw you already had a PR: https://github.com/elastic/enterprise-search-team/issues/7386#issuecomment-2079400941

Thanks for looking into this - my bad for not catching this edge case previously.

scripts/stack/copy-config.sh
wangch079: Make sense.

scripts/stack/run-stack.sh
wangch079: > move this `source ./copy-config.sh` call up to line 41?

Good idea


> Also, I think we should edit `copy-config.sh` to make a call to `make config` to ensure that `config.yml` exists.

Great suggestion, will update.

wangch079: Updated in https://github.com/elastic/connectors/pull/2445/commits/4d6d88a4e34455b9171d0efca31d52833422bf8a

artem-shelkovnikov (approved):

PR #2454, opened 2024-04-26
[GitHub] Add author to issue and pull request docs

navarone-feekery:
## Closes https://github.com/elastic/connectors/issues/2451

The author is missing from docs of types `issue` and `pull_request`. This change adds `{ "login": "<name>" }` to the `author` fields for these docs.
I considered directly adding `author.login` to the `author` field as a text-type, but I thought that might add unnecessary complexity if we want to add more `author.x` keys in the future.

Also doing this change exposed some tests that didn't work as expected for this source, so I also fixed those up.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -1692,10 +1692,7 @@ class GitHubDataSource(BaseDataSource):
                             self._prepare_review_doc(review=review)
                         )
                 else:
-                    if response is not None:
-                        type_obj[sample_dict[field_type]["es_field"]].extend(response)
-                    else:
-                        self._logger.info("Response was None!")
+                    type_obj[sample_dict[field_type]["es_field"]].extend(response)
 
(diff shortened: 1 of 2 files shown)


seanstory (approved):
🚢

connectors/sources/github.py
```diff
@@ -204,6 +204,9 @@ class GithubQuery(Enum):
             body
             state
             mergedAt
+            author {{
```
praveen-elastic, line 207: @navarone-feekery, I think you might need to add the field in `_prepare_pull_request_doc` and `_prepare_issue_doc` too, otherwise it won't be reflected in the document indexed.

```
"author": pull_request.get("author", {}).get("login")
```
cc @seanstory

navarone-feekery: I'm going to merge as this is approved, but if this is something you think warrants further discussion, let's talk in slack. We still have time to change it before release.

PR #2455, opened 2024-04-26
Check repo teams as well as collaborators

navarone-feekery:
Alters the action that applies community labels to check against a repo's members as well as collaborators.

Previously API failures were resulting in labels being applied, so that should no longer be an issue also.

### DIFF
--- a/.github/workflows/scripts/label_community_issues.py
+++ b/.github/workflows/scripts/label_community_issues.py
@@ -11,11 +11,12 @@ NUMBER = os.getenv("NUMBER")
 REPO = os.getenv("REPO")
 GITHUB_TOKEN = os.getenv("GITHUB_TOKEN")
 
+BASE_URL = "https://api.github.com"
 LABELS = ["community-driven", "needs-triage"]
 
 async def main():
     async with aiohttp.ClientSession() as session:
-        gh = GitHubAPI(session, requester="", base_url="https://api.github.com", oauth_token=GITHUB_TOKEN)
+        gh = GitHubAPI(session, requester="", base_url=BASE_URL, oauth_token=GITHUB_TOKEN)
 
         print("********")
         print(f"ACTOR: {ACTOR}")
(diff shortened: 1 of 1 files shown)


wangch079:
The [[List repository collaborators](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators)](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators) API list team members as well:

>For organization-owned repositories, the list of collaborators includes outside collaborators, organization members that are direct collaborators, **organization members with access through team memberships, organization members with access through default organization permissions**, and organization owners.

And more effectively, I think you can use [Check if a user is a repository collaborator](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#check-if-a-user-is-a-repository-collaborator) API directly.

navarone-feekery:
@wangch079 I know the documentation states that those endpoints work, but these don't seem to work as expected.

For list repository collaborators, running the following;

```
gh api -XGET /repos/elastic/connectors/collaborators --jq '.[] | .login' >> file.json
```

Returns only 60 usernames. Mine for example is not included in the list. It also doesn't include outside contributors. I can only get those if I add `affiliation=outside` to the URL params.

For checking directly if a user is a collaborator;

```
gh api -i /repos/elastic/connectors/collaborators/navarone-feekery
```

Calling this locally from my machine returns a `204` (expected response). However, when I did this as a github action within connectors, [it failed](https://github.com/elastic/connectors/actions/runs/8847874104). This is despite the fact that [it works on my own repo I tested it on](https://github.com/navarone-feekery/crawler-test-static-site/actions/runs/8850285825/job/24304235664). It is weird that it works locally. Maybe the github token provided to actions didn't give enough permissions 🤔

artem-shelkovnikov:
@navarone-feekery should we close this PR?

PR #2484, opened 2024-05-01
Use 8.3 mysql docker image for ftest

artem-shelkovnikov:
Functional tests for MySQL started failing due to new image of mysql released. New image seems to have plugin changes and some of our settings that we use (namely, auth) are done differently.

For now let's just pin MySQL version we use for testing (does not affect connector, affects only functional tests).

### DIFF
--- a/tests/sources/fixtures/mysql/docker-compose.yml
+++ b/tests/sources/fixtures/mysql/docker-compose.yml
@@ -26,7 +26,7 @@ services:
 
   mysql:
     container_name: mysql
-    image: mysql:latest
+    image: mysql:8.3
     environment:
       MYSQL_ROOT_PASSWORD: changeme
     command: --default-authentication-plugin=mysql_native_password


wangch079 (commented):
From the [official doc](https://www.elastic.co/guide/en/enterprise-search/current/connectors-mysql.html#connectors-mysql-compatibility):
>This connector is compatible with **MySQL 5.6 and later**.

So this is supposed to be `latest`

artem-shelkovnikov:
> So this is supposed to be latest

We are testing one image of 5.6+ in the end anyway (marked as latest, so previous patch releases or new minors are skipped). So if breaking changes happen in these earlier versions then we will not see them anyway.

Problem with `latest` tag here is maintenance of docker images that can change and bring new fun stuff and break tests like it happened here, I don't think we should use latest for testing. Connectors do work with `latest` anyway.

vidok (approved):

PR #2489, opened 2024-05-02
Add debug logs for extraction configuration state

navarone-feekery:
Add debug logs that show:

1. Whether content extraction is enabled or not
2. Whether content extraction is performed by the local extraction service or pipelines

#### Pre-Review Checklist


connectors/sync_job_runner.py
```diff
@@ -554,3 +551,19 @@ async def reload_connector(self):
         except DocumentNotFoundError:
             self.connector.log_error("Couldn't reload connector")
             return False
+
+    def _content_extraction_enabled(self, sync_job_config, pipeline_config):
+        if sync_job_config.get("use_text_extraction_service"):
+            logger.debug(
+                "Binary content extraction via local extraction service is enabled for this connector."
```
artem-shelkovnikov, line 558: These lines are not using correlation ids, so you would need to manually write sync job id/connector id to this line :/

navarone-feekery: @artem-shelkovnikov do you mean something like this prefix? https://github.com/elastic/connectors/blob/1d1cae1685971f082a97836a326a8ec657412cab/connectors/protocol/connectors.py#L373-L374

artem-shelkovnikov: We can do something similar to what's in the file already:

https://github.com/elastic/connectors/blob/60fe2ef3f56d0878d019d247dbd0f94d77e08204/connectors/sync_job_runner.py#L422-L426

```
                logger.debug(
                    f"A content sync job is started for connector {self.connector.id} by another connector instance, skipping..."
                )
```

Prefixing can be done too, but let's not touch it yet :)

navarone-feekery: I've added both ids to the log now.
The logs look long and repetitive, but I kept them that way instead of dynamically building them so it's easier to search the codebase for logs.

artem-shelkovnikov (approved):
🚀

PR #2490, opened 2024-05-02
Fix id logging in basic filtering rules

artem-shelkovnikov:
Basic sync rules do not output document ids because they try to access it by `_id`, but documents already have their ids in `id` field.

This happens because of the [following code](https://github.com/elastic/connectors/blob/main/connectors/es/sink.py#L545-L551): 

```
                doc_id = doc["id"] = doc.pop("_id")

                if self.basic_rule_engine and not self.basic_rule_engine.should_ingest(
                    doc
                ):
                    self.counters.increment((DOCS_FILTERED))
                    continue
```

As you can see, before basic rule engine checks the document, it's `_id` is removed from the document and written into `id`.

Before the change:

```
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
[FMWK][13:18:38][DEBUG] Document (id: 'None') didn't match any basic rule. Document will be included
```

After:

```
[FMWK][13:21:01][DEBUG] Document (id: '4467174') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '457829') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '746088') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '7596078') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '7686940') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '7901546') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '7926712') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '32037744') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '32165831') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '32178403') didn't match any basic rule. Document will be included
[FMWK][13:21:01][DEBUG] Document (id: '32180516') didn't match any basic rule. Document will be included
```

#### Pre-Review Checklist

## Release Note

Fix basic rule debug logging not displaying document ids properly.

### DIFF
--- a/connectors/filtering/basic_rule.py
+++ b/connectors/filtering/basic_rule.py
@@ -138,7 +138,7 @@ class BasicRuleEngine:
 
             if rule.matches(document):
                 logger.debug(
-                    f"Document (id: '{document.get('_id')}') matched basic rule (id: '{rule.id_}'). Document will be {rule.policy.value}d"
+                    f"Document (id: '{document.get('id')}') matched basic rule (id: '{rule.id_}'). Document will be {rule.policy.value}d"
                 )
 
                 self.rules_match_stats.setdefault(
@@ -151,7 +151,7 @@ class BasicRuleEngine:
         # default behavior: ingest document, if no rule matches ("default rule")
         self.rules_match_stats[BasicRule.DEFAULT_RULE_ID] += 1
         logger.debug(
-            f"Document (id: '{document.get('_id')}') didn't match any basic rule. Document will be included"
+            f"Document (id: '{document.get('id')}') didn't match any basic rule. Document will be included"
         )
         return True
 


artem-shelkovnikov:
nvm

artem-shelkovnikov:
Actually it's a problem. Added more info about it

wangch079:
https://github.com/elastic/connectors/blob/6b274109a09cb0502137bbdc9b2eec2301803f06/connectors/es/sink.py#L545-L551

You can solve the problem by moving line 545 after line 551. 

I also think it's not necessary to have an `id` field, we can simply do:
```python
doc_id = doc.pop("_id") 
```

artem-shelkovnikov:
> You can solve the problem by moving line 545 after line 551.

What would be the benefit of it?

> I also think it's not necessary to have an id field

I think it's true, but I don't know if I wanna test it and create other regression bug if I break something with it, I'd rather leave it as is 😅

wangch079:
>What would be the benefit of it?

I made the suggestion because I want to remove the duplicated `id` field. If this is done, you have to log it with field `_id`.

We had a discussion to remove the duplicated `id` field and I think the only concern is what if some users are utilizing the `id` field, which I'm not so worried about because we never documented the `id` field.

I agree this is not so relevant and I will approve it and you can decide later

wangch079 (approved):

artem-shelkovnikov:
I see. Do you think we're gonna do the id deletion any time soon? If it's a concrete plan, I'd go your way (moving the code that does basic sync rules before the id movement)

wangch079:
>Do you think we're gonna do the id deletion any time soon?

It will be just one-line change and we can do it if we want to.

But I suggest you merge this first as I see this PR will be backported to 8.14. Then we can have a separate one for 8.15 so that we have more time for regression (Although I don't really see any risk)

artem-shelkovnikov:
The only risk is that this logic will silently fail (like it did before) and not log ids properly.

PR #2494, opened 2024-05-02
Fix SharePoint Server API response for Site Pages and IDs

akanshi-elastic:
## Closes https://github.com/elastic/connectors/issues/2493###
1. Prefix collection with document IDs(API is returning same IDs if we copy one collection data to another)
2. Add CanvasContent1 to url for fetching site pages content( API response is not returning site pages content in Wiki Field now)

#### Pre-Review Checklist


artem-shelkovnikov (commented):
Can you add a test that demonstrates the fix? (ingesting multiple items with overlapping ids)

tests/sources/test_sharepoint_server.py
```diff
@@ -1630,3 +1630,56 @@ async def test_site_list_item_has_unique_role_assignments():
...
+        ),
+    ],
+)
+async def test_get_docs_having_same_site_ids(site_doc, expected_response):
+    async with create_sps_source() as source:
+        source.sharepoint_client.site_collections_path = ["collection1", "collection2"]
+        source.sharepoint_client.get_sites = AsyncIterator([site_doc])
+        source.sharepoint_client.get_lists = AsyncIterator([])
+        async for item, _ in source.get_docs():
+            assert item == expected_response
```
artem-shelkovnikov, line 1685: This test does not seem to demonstrate the behaviour.

As far as I understand you would need to make a test that returns 2 list items from different lists that have same id, and verify that both are returned with unique ids.

akanshi-elastic: Actually 2list item from different list in same collection never be same. we have added this change because if there are 2 collection having same copied list then ids will be same.

artem-shelkovnikov (approved):

seanstory:
@akanshi-elastic please manually create a backport to resolve the conflicts ☝️

PR #2495, opened 2024-05-02
Add ancestor title in confluence

akanshi-elastic:
#### Pre-Review Checklist


artem-shelkovnikov (commented):
Can you add a test that fetches ancestors? Additionally, can you add this logic to functional test suite?

connectors/sources/confluence.py
```diff
@@ -308,6 +310,21 @@ async def search_by_query(self, query):
...
+        try:
+            async for ancestors_data in self.api_call(
+                url=os.path.join(
+                    self.host_url, URLS[FETCH_ANCESTORS].format(id=page_id)
+                )
+            ):
+                ancestors = await ancestors_data.json()
+                return ancestors
+        except Exception as exception:
+            self._logger.warning(
```
wangch079, line 323: Should log error?

connectors/sources/confluence.py
```diff
@@ -861,13 +878,18 @@ async def fetch_documents(self, api_query):
...
+            )
+
             yield {
                 "_id": document.get("id"),
                 "type": document.get("type", ""),
                 "_timestamp": document.get("history", {})
                 .get("lastUpdated", {})
                 .get("when", iso_utc()),
                 "title": document.get("title", ""),
+                "title_ancestors": ancestor_title if ancestor_title else "",
```
wangch079, line 892: `if...else...` is not necessary here. You didn't do this either in line 967

connectors/sources/confluence.py
```diff
@@ -929,10 +951,20 @@ async def search_by_query(self, query):
...
+                )
+                ancestor_title = " ".join(
+                    ancestor.get("title", "")
+                    for ancestor in ancestors.get("ancestors", [])
+                )
 
             document = {
                 "_id": entity_details.get("id"),
                 "title": entity.get("title"),
+                "title_ancestors": ancestor_title,
```
wangch079, line 967: Why not just name it `ancestors`?

akanshi-elastic:
> Can you add a test that fetches ancestors? Additionally, can you add this logic to functional test suite?

I have added a test case but for adding functional test suite, we usually don't write ftest for sync rules part and here `fetch_ancestors` has been used for sync rules only. for normal sync we are getting it from `rest/api/content/search` this alraedy used API.

artem-shelkovnikov:
Can you address this comment as well:

> Additionally, can you add this logic to functional test suite?

akanshi-elastic:
> > I have added a test case but for adding functional test suite, we usually don't write ftest for sync rules part and here fetch_ancestors has been used for sync rules only
> 
> Why is `fetch_ancestors` used only by sync rules? Should it not be fetched by connector always?

for normal full sync w are already getting ancestors `rest/api/content/search` by just using `expand=ancestors` but that is not working for searching API used in sync rule so for that we need to do this extra API call.

akanshi-elastic:
Buildkite test this

tests/sources/test_confluence.py
```diff
@@ -1477,3 +1498,36 @@ async def test_end_signal_is_added_to_queue_in_case_of_exception():
...
+    "search_by_query",
+    return_value=AsyncIterator([EXPECTED_QUERY_RESPONSE]),
+)
+async def test_search_by_query_with_ancestors(search_by_query_response):
+    async with create_confluence_source() as source:
+        source.confluence_client.fetch_ancestors = Mock(
+            return_value=create_fake_coroutine(
+                {"ancestors": [{"title": "parent1"}, {"title": "parent2"}]}
+            )
+        )
```
wangch079, line 1520: Won't this work?

```suggestion
        source.confluence_client.fetch_ancestors = AsyncMock(
            return_value={"ancestors": [{"title": "parent1"}, {"title": "parent2"}]}
        )
```

akanshi-elastic:
Buildkite test this

artem-shelkovnikov (commented):
Have you estimated the impact of this? How slower does the sync get with this change?

akanshi-elastic:
> Have you estimated the impact of this? How slower does the sync get with this change?

- Full Sync:
  Before fix: 10512 docs in 8 min 21 sec (rate: 1266 docs/min)
  After fix: 10512 docs in 9 min 3 sec (rate: 1150 docs/min)

- Sync rule:
  before: 1007 in 22 sec
  after: 1007 in 6 min
 As we are doing API call for fetching ancestor of every page/blogpost sync rule will be slower than before. Either we can add a toggle for enable/disable of adding ancestor and will mention in the tooltip that enabling this option will slower down the performance.

connectors/sources/confluence.py
```diff
@@ -861,13 +878,18 @@ async def fetch_documents(self, api_query):
                 self.confluence_client.host_url,
                 document.get("_links", {}).get("webui", "")[1:],
             )
+            ancestor_title = " ".join(
+                ancestor["title"] for ancestor in document.get("ancestors", [])
+            )
```
seanstory, line 883: Why not have `"ancestors"` return an array, since that's what we get from Confluence? Join then array items into a single string makes it much harder to consume these downstream.

connectors/sources/confluence.py
```diff
@@ -929,10 +951,20 @@ async def search_by_query(self, query):
...
                 continue
+            ancestor_title = ""
+            if entity.get("entityType") != "space":
+                ancestors = await self.confluence_client.fetch_ancestors(
+                    page_id=entity_details.get("id")
+                )
+                ancestor_title = " ".join(
+                    ancestor.get("title", "")
+                    for ancestor in ancestors.get("ancestors", [])
+                )
```
seanstory, line 962: This is for advanced sync rules, if the user defined a specific query, right? In that case, I don't think we should execute extra requests. If they want the ancestors, they can include that in the query. When forming up the output document, we can just look to see if there's an `"ancestors"` key in the response.

seanstory (commented):
I don't think this is the right approach. See also my comment here: https://github.com/elastic/connectors/issues/2378#issuecomment-2096313901

connectors/sources/confluence.py
akanshi-elastic: We don't get an ancestors title array directly from confluence API. it is list contains dictionary having ancestors. I have shared the output [here](https://github.com/elastic/connectors/issues/2378#issuecomment-2097501463).

connectors/sources/confluence.py
akanshi-elastic: @seanstory But still if user right cql query, it will not fetch ancestors in sync rule because the API endpoint which we are using in sync rule has no query param to fetch ancestors directly. we need to do an additional API call for fetching it. What we can do - we can add a new boolean field in sync rules i.e. `is_ancestors` and if it's true we'll fetch ancestor otherwise not

connectors/sources/confluence.py
seanstory: That's fine. So let's trim everything except for title for now, but leave the structure so that if another user requests some other ancestor metadata, we have an easy way to add it. So our output would be like:
```
{
  ...
  "ancestors": [
    {
      "title": "Foo"
    },
    {
      "title": "Bar"
    }
  ],
  ...
}
```

connectors/sources/confluence.py
seanstory: That surprised me, but I took a quick look, and couldn't find a way to get ancestors (in bulk) from the search API either.

In that case, I think we should just include in the docs and release notes that `"ancestors"` will not be present if you use advanced sync rules. I don't think the performance hit is worth it, and I don't really want to add the extra cost of another configuration to make sync rules or the RCFs more complex.

akanshi-elastic:
Buildkite test this

seanstory (approved):

akanshi-elastic:
Buildkite test this

akanshi-elastic:
Buildkite test this

akanshi-elastic:
Buildkite test this

PR #2503, opened 2024-05-03
Make CLI docs up-to-date

vidok:
This PR will adjust the CLI docs: 
1. Make sure customers understand they need to run connector service to perform sync jobs. 
2. Remove the Limitations block because the CLI supports ES API keys.


navarone-feekery (approved):
LGTM!

PR #2511, opened 2024-05-07
Support serverless in index list command

vidok:
## Closes https://github.com/elastic/enterprise-search-team/issues/7441

`GET _stats` API isn't available in Elasticsearch serverless so this PR will introduce a workaround for `./bin/connectors index list` command. 

#### Pre-Review Checklist

### DIFF
--- a/connectors/cli/index.py
+++ b/connectors/cli/index.py
@@ -17,7 +17,7 @@ class Index:
         self.connectors_index = ConnectorIndex(self.elastic_config)
 
     def list_indices(self):
-        return asyncio.run(self.__list_indices())["indices"]
+        return asyncio.run(self.__list_indices())
 
     def clean(self, index_name):
         return asyncio.run(self.__clean_index(index_name))
--- a/connectors/cli/index.py
+++ b/connectors/cli/index.py
@@ -36,6 +36,9 @@ class Index:
         try:
             return await self.cli_client.list_indices()
         except ApiError as e:
+            # If the API is not available, we are in serverless mode
+            if e.error == "api_not_available_exception":
+                return await self.cli_client.list_indices_serverless()
             raise e
         finally:
             await self.__close()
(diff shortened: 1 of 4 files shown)


connectors/cli/index.py
```diff
@@ -36,6 +36,9 @@ async def __list_indices(self):
         try:
             return await self.cli_client.list_indices()
         except ApiError as e:
+            # If the API is not available, we are in serverless mode
+            if e.error == "api_not_available_exception":
+                return await self.cli_client.list_indices_serverless()
```
artem-shelkovnikov, line 41: Should we log something about it?

connectors/es/management_client.py
```diff
@@ -185,11 +186,40 @@ async def clean_index(self, index_name):
...
+        """
+        List indices in a serverless environment. This method is a workaround to the fact that
+        the `indices.stats` API is not available in serverless environments.
+        """
+
+        indices = {}
+        try:
+            response = await self._retrier.execute_with_retry(
+                partial(self.client.indices.get, index=index)
+            )
```
artem-shelkovnikov, line 213: So there's absolutely no way to do so? What if we do N+1 query here, doing count on each index instead?

artem-shelkovnikov (approved):
Looks good, left a couple non-blocking questions

connectors/cli/index.py
vidok: I couldn't find a good reason why we should log this.

artem-shelkovnikov: Mainly as a way to explain why we cannot show number of documents in the indices

connectors/es/management_client.py
vidok: I was thinking about the N+1 approach here but prefer not to because of the performance issues. I assume that getting indices names quickly is better than getting the names and docs counts slowly.

PR #2529, opened 2024-05-08
[PostgreSQL] Add logs

moxarth-elastic:

Adding more logs in PostgreSQL connector.

Log file: https://drive.google.com/file/d/12Pkq94J3ymnVGhXBtNc8S61xLqzDFCSK/view?usp=drive_link

#### Pre-Review Checklist


connectors/sources/postgresql.py
```diff
@@ -218,6 +219,12 @@ async def ping(self):
     async def get_tables_to_fetch(self, is_filtering=False):
         tables = configured_tables(self.tables)
         if is_wildcard(tables) or is_filtering:
+            msg = (
+                "Fetching all tables as the configuration field 'tables' is set to '*'"
+                if not is_filtering
+                else "Fetching all tables as the advanced sync rules are enabled."
+            )
+            self._logger.info(msg)
```
seanstory, line 227:
```suggestion
            msg = "Fetching all tables"
            self._logger.info(msg)
```

connectors/sources/postgresql.py
```diff
@@ -266,9 +274,12 @@ async def get_table_primary_key(self, table):
             )
         ]
 
+        self._logger.debug(f"Found primary keys for '{table}' table")
+
         return primary_keys
 
     async def get_table_last_update_time(self, table):
+        self._logger.debug(f"Fetching last updated time for table: {table}")
```
seanstory, line 282: Let's log the time we fetched

connectors/sources/postgresql.py
```diff
@@ -266,9 +274,12 @@ async def get_table_primary_key(self, table):
             )
         ]
 
+        self._logger.debug(f"Found primary keys for '{table}' table")
```
seanstory, line 277: let's log what the primary keys were

connectors/sources/postgresql.py
```diff
@@ -300,6 +311,7 @@ async def data_streamer(
             list: It will first yield the column names, then data in each row
         """
         if query is None and row_count is not None and order_by_columns is not None:
+            self._logger.debug(f"Streaming records from database using query: {query}")
```
seanstory, line 314: This seems wrong. `query is None` was just asserted above this

connectors/sources/postgresql.py
```diff
@@ -328,6 +340,7 @@ async def data_streamer(
                 if row_count <= offset:
                     return
         else:
+            self._logger.debug(f"Streaming records from database for table: {table}")
```
seanstory, line 343: In the sample log file, table is `None`. That seems like an error.

```
[FMWK][18:58:23][DEBUG] [Connector id: 3S2VrY4BAqKLTyj6P7_3, index name: postgresql, Sync job id: KpLkvY4BBpk9ZgBA2rJn] Streaming records from database for table: None
```

artem-shelkovnikov:
Copying from another PR - that's applicable to all SQL connectors:

The most useful connector logs are - which queries were actually executed by the connectors. Can you add debug logs that state which exact queries were executed? A good addition to it would be to also debug log the number of records returned by the query.

connectors/sources/postgresql.py
moxarth-elastic: I think the logs are swapped for if-else conditions.

moxarth-elastic:
> which queries were actually executed by the connectors. Can you add debug logs that state which exact queries were executed?

@artem-shelkovnikov  This is already present here https://github.com/elastic/connectors/blob/43ee46ce2852c3803a85656ee37ab1f02f2f3b40/connectors/sources/postgresql.py#L348

Also, i've added a log for the number of records fetched from table/query.

artem-shelkovnikov (approved):

PR #2530, opened 2024-05-08
[Azure Blob Storage] Add logs

moxarth-rathod:

Adding more logs in Azure Blob Storage connector.

Log file: https://drive.google.com/file/d/1RMuieFYk-llbqZUj__eYJbnUY5W2wq7K/view?usp=drive_link
Updated file: https://drive.google.com/file/d/1dVEpn6ccoz5uKAgzSrahtDp1vyifJDIh/view?usp=drive_link 

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/azure_blob_storage.py
+++ b/connectors/sources/azure_blob_storage.py
@@ -124,14 +124,14 @@ class AzureBlobStorageDataSource(BaseDataSource):
 
     async def ping(self):
         """Verify the connection with Azure Blob Storage"""
-        self._logger.info("Generating connection string...")
+        self._logger.debug("Generating connection string...")
         self.connection_string = self._configure_connection_string()
         try:
             async with BlobServiceClient.from_connection_string(
                 conn_str=self.connection_string, retry_total=self.retry_count
             ) as azure_base_client:
                 await azure_base_client.get_account_information()
-                self._logger.info("Successfully connected to the Azure Blob Storage.")
+                self._logger.debug("Successfully connected to the Azure Blob Storage.")
         except Exception:
             self._logger.exception("Error while connecting to the Azure Blob Storage.")
             raise
@@ -185,7 +185,6 @@ class AzureBlobStorageDataSource(BaseDataSource):
         if not self.can_file_be_downloaded(file_extension, filename, file_size):
             return
 
-        self._logger.debug(f"Downloading content for file: {filename}")
         document = {"_id": blob["id"], "_timestamp": blob["_timestamp"]}
         return await self.download_and_extract_file(
             document,
@@ -216,7 +215,7 @@ class AzureBlobStorageDataSource(BaseDataSource):
         Yields:
             dictionary: Container document with name & metadata
         """
-        self._logger.debug("Fetching containers")
+        self._logger.info("Fetching containers")
         container_set = set(container_list)
         async with BlobServiceClient.from_connection_string(
             conn_str=self.connection_string, retry_total=self.retry_count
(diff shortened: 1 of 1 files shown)


connectors/sources/azure_blob_storage.py
```diff
@@ -212,6 +216,7 @@ async def get_container(self, container_list):
         Yields:
             dictionary: Container document with name & metadata
         """
+        self._logger.debug("Fetching containers")
```
jedrazb, line 219: in `get_blob` we have log level `INFO`, maybe worth changing log level `INFO` here as well to keep it consistent?

jedrazb (commented):
Looks good, one additional thing I noticed is that in `get_docs` we have 

```
 async def get_docs(self, filtering=None):
        ...
        async for container_data in self.get_container(container_list=self.containers):
            if container_data:
                ...
```

I don't have the context to know why we need `if container_data` but if it's needed I would log a `debug` statement that in `else` clause that we are skipping container because it's empty(?) / doesn't have data?

connectors/sources/azure_blob_storage.py
```diff
@@ -185,6 +185,7 @@ async def get_content(self, blob, timestamp=None, doit=None):
         if not self.can_file_be_downloaded(file_extension, filename, file_size):
             return
 
+        self._logger.debug(f"Downloading content for file: {filename}")
```
artem-shelkovnikov, line 188: Looks like a duplicate line - you already have "f"Downloading content for blob: {blob_name} from {container_name} container"" that happens right after

connectors/sources/azure_blob_storage.py
```diff
@@ -247,6 +252,7 @@ async def get_blob(self, container):
         Yields:
             dictionary: Formatted blob document
         """
+        self._logger.info(f"Fetching blobs for '{container['name']}' container")
```
artem-shelkovnikov, line 255:
```suggestion
        self._logger.info(f"Fetching blobs for container '{container['name']}'")
```

artem-shelkovnikov (commented):
1. Would be also cool to log how many blobs are found in the container. Also would be nice to report progress in format of "Extracted X out of Y blobs for container {container}" reported every 100 lines and in the end of sync.
2. Log lines "Generating connection string." should be DEBUG
3. Log lines "Successfully connected to the Azure Blob Storage" should be DEBUG
4. Need a DEBUG log line saying which containers were found, then which ones will be synced in the end after all validations for it

moxarth-elastic:
> I don't have the context to know why we need `if container_data` but if it's needed I would log a `debug` statement that in `else` clause that we are skipping container because it's empty(?) / doesn't have data?

This condition was added due this `if` statement. So, at the end of the execution of this method https://github.com/elastic/connectors/blob/22bfb35cbfa412b845aeb55b19f080594235af3e/connectors/sources/azure_blob_storage.py#L211 we return None so added a condition in get_docs to handle it.

jedrazb:
@moxarth-elastic could you share updated example DEBUG output, before we hit ✅

moxarth-elastic:
> @moxarth-elastic could you share updated example DEBUG output, before we hit ✅

Here is the updated log file: https://drive.google.com/file/d/1HJ36WDN87P3q4fCeQGu8FiQUrh9texD-/view?usp=drive_link

praveen-elastic:
Buildkite test this

connectors/sources/azure_blob_storage.py
```diff
@@ -280,12 +282,18 @@ async def get_blob(self, container):
...
+        blob_count = 0
         try:
             async for blob in container_client.list_blobs(include=["metadata"]):
+                blob_count += 1
                 yield self.prepare_blob_doc(
                     blob=blob, container_metadata=container["metadata"]
                 )
+            self._logger.info(
+                f"Fetched {blob_count} blobs from '{container['name']}' container"
+            )
```
artem-shelkovnikov, line 296: This should go into a `finally` statement so that it's always shown.

artem-shelkovnikov (commented):
Any chance we can have logs of actual requests to the ABS?

moxarth-elastic:
@artem-shelkovnikov added logs here: https://drive.google.com/file/d/1dVEpn6ccoz5uKAgzSrahtDp1vyifJDIh/view?usp=drive_link

PR #2543, opened 2024-05-10
[Azure Blob Storage] Optimize memory consumption

moxarth-elastic:
## Closes https://github.com/elastic/connectors/issues/2465

Improve memory consumption in ABS connector. Added perf8 reports for the comparision:

Before: https://drive.google.com/file/d/12fcQWT8NKUr9CMgdEZsqVscXLEiFNJw-/view?usp=drive_link
After: https://drive.google.com/file/d/1IOKhm2s8hk0abjGarFy75d64AmKeANYE/view?usp=drive_link

## Fix:

The previous version of the connector instantiated a BlobClient for each blob it downloaded. Where, the new approach utilizes a single ContainerClient to download data for all blobs within that container. 

For instance, if a container contains 100 blobs, the old code was creating 100 separate BlobClient instances to download those blobs, but now the new code leverages just one ContainerClient for all blobs in the container. This optimization was inspired by a suggestion from[ this GitHub issue](https://github.com/Azure/azure-sdk-for-python/issues/27023#issuecomment-1314296101).

Additionally, the older code used an asynciterator with the data.chunks() method. The drawback of asynciterator is that it does not release memory until it has completed all iterations.

#### Pre-Review Checklist

- Updated the version for `azure-storage-blob` in the requirements.

### DIFF
--- a/connectors/sources/azure_blob_storage.py
+++ b/connectors/sources/azure_blob_storage.py
@@ -20,6 +20,7 @@ DEFAULT_RETRY_COUNT = 3
 MAX_CONCURRENT_DOWNLOADS = (
     100  # Max concurrent download supported by Azure Blob Storage
 )
+INITIAL_DOWNLOAD_SIZE = 100000
 
 
 class AzureBlobStorageDataSource(BaseDataSource):
@@ -137,6 +138,13 @@ class AzureBlobStorageDataSource(BaseDataSource):
             self._logger.exception("Error while connecting to the Azure Blob Storage.")
             raise
 
+    async def close(self):
+        if not self.container_clients:
+            return
+        for container_client in self.container_clients.values():
+            await container_client.close()
+        self.container_clients = {}
+
     def prepare_blob_doc(self, blob, container_metadata):
         """Prepare key mappings to blob document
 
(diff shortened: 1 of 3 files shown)


artem-shelkovnikov (commented):
Can you add description to the PR why this PR improves RAM usage? What were your findings?

moxarth-elastic:
> Can you add description to the PR why this PR improves RAM usage? What were your findings?

Added a fix in the description.

moxarth-elastic:
Buildkite test this

connectors/sources/azure_blob_storage.py
```diff
@@ -190,19 +191,38 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
+                retry_total=self.retry_count,
+            )
+            return self.container_clients[container_name]
+        else:
+            return self.container_clients[container_name]
+
+    async def blob_download_func(self, blob_name, container_name, file_size):
+        container_client = self._get_container_client(container_name=container_name)
+        offset = 0
+        length = 100000
```
artem-shelkovnikov, line 213: What is this number? Why 100000? Can we put it into a meaningful constant or even a setting?

connectors/sources/azure_blob_storage.py
```diff
@@ -190,19 +191,38 @@ async def get_content(self, blob, timestamp=None, doit=None):
...
-        async with BlobClient.from_connection_string(
-            conn_str=self.connection_string,
-            container_name=container_name,
-            blob_name=blob_name,
-            retry_total=self.retry_count,
-        ) as blob_client:
-            data = await blob_client.download_blob()
-            async for content in data.chunks():
-                yield content
+    def _get_container_client(self, container_name):
```
artem-shelkovnikov, line 197: Will you need to close these clients manually?

moxarth-elastic:
buildkite test this

connectors/sources/azure_blob_storage.py
moxarth-elastic: Yes, I've added a `close` method for the same.

moxarth-elastic:
buildkite test this

connectors/sources/azure_blob_storage.py
moxarth-elastic: It's used to fetch the file content in the chunks of 100 Kb. I've put it in a constant.

artem-shelkovnikov (approved):

praveen-elastic:
Buildkite test this

praveen-elastic:
Buildkite test this

PR #2546, opened 2024-05-13
[Servicenow] Fix end signal not enqueued

praveen-elastic:
## Closes https://github.com/elastic/connectors/issues/2545

Following changes are included in this PR:

1. Ensure end-signals are put to queue even if error is encountered, in all relevant places. 
2. Set advanced_rules_enabled to True 

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/servicenow.py
+++ b/connectors/sources/servicenow.py
@@ -425,6 +425,7 @@ class ServiceNowDataSource(BaseDataSource):
 
     name = "ServiceNow"
     service_type = "servicenow"
+    advanced_rules_enabled = True
     dls_enabled = True
     incremental_sync_enabled = True
 
@@ -483,7 +484,6 @@ class ServiceNowDataSource(BaseDataSource):
                 "order": 4,
                 "tooltip": "List of services is ignored when Advanced Sync Rules are used.",
                 "type": "list",
-                "value": "*",
             },
             "retry_count": {
                 "default_value": RETRIES,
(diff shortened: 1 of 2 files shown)


connectors/sources/servicenow.py
```diff
@@ -671,28 +672,38 @@ async def _fetch_attachment_metadata(self, batched_apis, table_access_control):
...
+                ]
+                await self.fetchers.put(
+                    partial(
+                        self._fetch_attachment_metadata,
+                        batched_apis,
+                        table_access_control,
+                    )
                 )
+                self.task_count += 1
+        except Exception as exception:
```
seanstory, line 701: This feels like it should still fail the sync, no?

seanstory (commented):
Can you add unit tests too?

connectors/sources/servicenow.py
praveen-elastic: Hey, I didn't understand what you are trying to convey. 

These changes are similar to the ones we implemented in confluence https://github.com/elastic/connectors/pull/2380. 

Using these changes we make sure that END_SIGNALS for a given task are added to queue even if error occurs inside that task, this prevents _consumer method from running in a forever while loop.

seanstory: In that linked PR, the caught exceptions are logged with `_logger.exception` and then re-`raise`'d. Here, you're just swallowing the error, so the sync won't fail after this is merged, but the same issue _would_ fail the sync today.

Unless you've got a good reason (which you might!), I suggest re-raising the error, and logging these as error/exception instead of warn.

praveen-elastic: Got it, i'll update the code to re-raise the error

praveen-elastic: @seanstory updated the PR with the changes.

seanstory (approved):

PR #2555, opened 2024-05-14
[Dropbox ] Add logs

moxarth-rathod:

Adding more logs in Dropbox connector.

Log file: https://drive.google.com/file/d/1njpVpExhM5Fw_efUgvl9SRxSGOtr-UAA/view?usp=drive_link

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/dropbox.py
+++ b/connectors/sources/dropbox.py
@@ -189,9 +189,6 @@ class DropboxClient:
         if self.token_expiration_time and (
             not isinstance(self.token_expiration_time, datetime)
         ):
-            self._logger.debug(
-                f"Token expiration time '{self.token_expiration_time}' is not in the correct format. Converting it into ISO format"
-            )
             self.token_expiration_time = datetime.fromisoformat(
                 self.token_expiration_time
             )
@@ -1000,7 +997,6 @@ class DropboxDataSource(BaseDataSource):
     async def get_permission(self, permission, account_id):
         permissions = []
         if identities := permission.get("users"):
-            self._logger.debug("Fetching users")
             for identity in identities:
                 permissions.append(
                     _prefix_user_id(identity.get("user", {}).get("account_id"))
@@ -1045,7 +1041,7 @@ class DropboxDataSource(BaseDataSource):
             )
 
     async def get_account_details(self):
-        self._logger.debug("Retrieving account details")
+        self._logger.debug("Fetching account details")
         response = await anext(
             self.dropbox_client.api_call(
                 base_url=BASE_URLS["FILES_FOLDERS_BASE_URL"],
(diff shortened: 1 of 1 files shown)


moxarth-elastic:
buildkite test this

navarone-feekery:
@moxarth-elastic can you provide the debug log output for this during a standard run [as requested on the issue](https://github.com/elastic/connectors/issues/2299#issuecomment-2020285728)?

moxarth-elastic:
> @moxarth-elastic can you provide the debug log output for this during a standard run [as requested on the issue](https://github.com/elastic/connectors/issues/2299#issuecomment-2020285728)?

Sure, attached a log file here: https://drive.google.com/file/d/1njpVpExhM5Fw_efUgvl9SRxSGOtr-UAA/view?usp=drive_link

connectors/sources/dropbox.py
```diff
@@ -189,6 +189,9 @@ async def _set_access_token(self):
         if self.token_expiration_time and (
             not isinstance(self.token_expiration_time, datetime)
         ):
+            self._logger.debug(
+                f"Token expiration time '{self.token_expiration_time}' is not in the correct format. Converting it into ISO format"
+            )
```
navarone-feekery, line 194: I don't know that this log provides much value. It isn't a significant code fork path and it isn't changing data being synced, it's just fixing a var type. I think it can be removed.

connectors/sources/dropbox.py
```diff
@@ -987,6 +1000,7 @@ def get_email(self, permission, identity):
     async def get_permission(self, permission, account_id):
         permissions = []
         if identities := permission.get("users"):
+            self._logger.debug("Fetching users")
```
navarone-feekery, line 1003: I'm unclear on what exactly is being logged here. There are no API calls happening to fetch users. Also there are no logs for `invitees` or `groups`. I think this log is unnecessary.

navarone-feekery (commented):
Nice! I have some change requests.

- I think a log should also be added to `map_permission_with_document` (single debug log with keys of docs that will have permissions mapped to them)
- The debug logs have inconsistent wording (they use `Fetching ...` or `Retrieving ...`). Can you choose one or the other and apply it to all the logs? It would be clearer to the user that these logs are all related to API requests.
- The logfile you provided doesn't contain any logs from this PR. Was it on the correct branch when you ran it?
- Also half of the new debug logs are for ACL, but the provided logfile was for a regular sync. Can you also link a logfile for an ACL sync?

navarone-feekery:
I've also just been made aware that a similar PR is out, so I think we should include the missing logs from that PR into this one: https://github.com/elastic/connectors/pull/2324

PR #2561, opened 2024-05-15
Implement feature flag for Elasticsearch API + use for connector heartbeat

artem-shelkovnikov:
### Closes https://github.com/elastic/enterprise-search-team/issues/7469

This PR attempt to do the following things:
- Add `elasticsearch.feature_use_connectors_api` config option that turns of usage of connectors API if on. Turned off by default
- Use this feature flag to call Connectors API for connector health check function as a demonstration of implementation.

Some notes:

I went for the path of least resistance here keeping in mind that current design of classes and code is more ORM-style with `ESIndex` abstraction actually encapsulating calls to specific indices. With Connectors API it changes - instead of indices it _might_ work in future on the route basis, E.G. one class for calls to Connector endpoints, one class for calls to Connector Sync job endpoints and so on.

Decisions for the above can be done later, especially when Connectors API will be natively available in Python Elasticsearch client. When it happens, we can do design of the API classes better.

Additionally, with feature flag we'll need to maintain both versions of code - calling indices and API until we're ready to make indices hidden. The path forward is:

1. Make Connectors API production-ready
2. Mark feature as deprecated
3. In appropriate time deprecate feature and make indices not accessible publicly based on Elastic policies of deprecation
4. Clean up the code

This means that abstractions will have to least for one more year, so I opted in for simplest way forward that's also easy to clean up - one class `ESApi` with all methods to call APIs that are wrapped in retry policies + use this class in `ESIndex` class - seemed like the cleanest way ahead. All of that might change as we switch more endpoints to the API, but for now it seems like simplest way forward.

#### Pre-Review Checklist

## Release Note

Add a preview feature for connectors to switch to Elasticsearch Connectors API

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -37,6 +37,7 @@ from connectors.utils import (
     iso_utc,
     nested_get_from_dict,
     next_run,
+    parse_datetime_string
 )
 
 __all__ = [
@@ -146,9 +147,13 @@ class ConnectorIndex(ESIndex):
         logger.debug(f"ConnectorIndex connecting to {elastic_config['host']}")
         # initialize ESIndex instance
         super().__init__(index_name=CONNECTORS_INDEX, elastic_config=elastic_config)
+        self.feature_use_connectors_api = elastic_config.get("feature_use_connectors_api")
 
     async def heartbeat(self, doc_id):
-        await self.update(doc_id=doc_id, doc={"last_seen": iso_utc()})
+        if self.feature_use_connectors_api:
+            await self.api.connector_check_in(doc_id)
+        else:
+            await self.update(doc_id=doc_id, doc={"last_seen": iso_utc()})
 
     async def supported_connectors(self, native_service_types=None, connector_ids=None):
         if native_service_types is None:
@@ -535,10 +540,7 @@ class Connector(ESDocument):
 
     @property
     def last_seen(self):
-        last_seen = self.get("last_seen")
-        if last_seen is not None:
-            last_seen = datetime.fromisoformat(last_seen)  # pyright: ignore
-        return last_seen
+        return self._property_as_datetime("last_seen")
 
     @property
     def native(self):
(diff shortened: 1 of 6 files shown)


connectors/protocol/connectors.py
```diff
@@ -591,7 +595,7 @@ def last_access_control_sync_status(self):
     def _property_as_datetime(self, key):
         value = self.get(key)
         if value is not None:
-            value = datetime.fromisoformat(value)  # pyright: ignore
+            value = parse_datetime_string(value)  # pyright: ignore
```
artem-shelkovnikov, line 598: Had to do this because `datetime.fromisoformat` expects milliseconds, but connectors API actually returns nanoseconds. That causes errors like this:

```
[FMWK][14:40:30][CRITICAL] [job_scheduling_service] Invalid isoformat string: '2024-05-15T12:37:52.429924009Z'
  File "/Users/artemshelkovnikov-m2/git_tree/connectors-py/connectors/services/job_scheduling.py", line 162, in _run
    await self._schedule(connector)
  File "/Users/artemshelkovnikov-m2/git_tree/connectors-py/connectors/services/job_scheduling.py", line 69, in _schedule
    await connector.heartbeat(self.heartbeat_interval)
  File "/Users/artemshelkovnikov-m2/git_tree/connectors-py/connectors/protocol/connectors.py", line 637, in heartbeat
    self.last_seen is None
  File "/Users/artemshelkovnikov-m2/git_tree/connectors-py/connectors/protocol/connectors.py", line 546, in last_seen
    last_seen = datetime.fromisoformat(last_seen)  # pyright: ignore
ValueError: Invalid isoformat string: '2024-05-15T12:37:52.429924009Z'
```

Had a short discussion with @jedrazb and confirmed, that we want to keep nanoseconds format, so switching to parsing of dates to `dateutil.parser.parse` that allows for this datetime format.

jedrazb: Sounds good! One question, is this function backward compatible? Would it be able to handle old version of timestamps expressed with millisecond level precision? I'm thinking about some scenarios where connector service would be upgraded with some old pending jobs in the index

jedrazb (commented):
LGTM! Just one question regarding backward compatibility of timestamps

connectors/protocol/connectors.py
artem-shelkovnikov: It should, but I'll add a couple tests for this

jedrazb (approved):
LGTM!

PR #2571, opened 2024-05-20
added support to read proxy settings from env vars

chris-gunawardena:
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/box.py
+++ b/connectors/sources/box.py
@@ -138,6 +138,7 @@ class BoxClient:
         self._logger = logger
         self.is_enterprise = configuration["is_enterprise"]
         self._http_session = aiohttp.ClientSession(
+            trust_env=True,
             base_url=BASE_URL, raise_for_status=True
         )
         self.token = AccessToken(
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -156,6 +156,7 @@ class ConfluenceClient:
         basic_auth = aiohttp.BasicAuth(login=auth[0], password=auth[1])
         timeout = aiohttp.ClientTimeout(total=None)  # pyright: ignore
         self.session = aiohttp.ClientSession(
+            trust_env=True,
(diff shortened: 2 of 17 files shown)


navarone-feekery:
@chris-gunawardena thanks for the contribution! I think this is a good change to add to Connectors. There are a few things we need to work on before merging first.

1. I think it will be important that it is a configurable value. Ideally this would be configurable from the [config.yml file](https://github.com/elastic/connectors/blob/main/config.yml.example). I'll try and find an example of how this could be done.
2. This doesn't completely solve the proxy issue linked, because ENV vars can't be set on cloud. We still think it's worth merging for users running on self-managed infrastructure. Can you change the PR link to the issue to `Relates to` instead of `Closes`?
3. Also, do you have any links to documentation that explains which ENV vars can be used like this?

MatheusGelinskiPires:
> @chris-gunawardena thanks for the contribution! I think this is a good change to add to Connectors. There are a few things we need to work on before merging first.
> 
> 1. I think it will be important that it is a configurable value. Ideally this would be configurable from the [config.yml file](https://github.com/elastic/connectors/blob/main/config.yml.example). I'll try and find an example of how this could be done.
> 2. This doesn't completely solve the proxy issue linked, because ENV vars can't be set on cloud. We still think it's worth merging for users running on self-managed infrastructure. Can you change the PR link to the issue to `Relates to` instead of `Closes`?
> 3. Also, do you have any links to documentation that explains which ENV vars can be used like this?

We also need to add proxy support between the Connector and the Elasticsearch endpoint.

seanstory:
> We also need to add proxy support between the Connector and the Elasticsearch endpoint.

I think that's why @navarone-feekery asked to change the PR description from `closes` to `relates to`. This change will have value on its own, even without also adding proxy support for Elasticsearch, which can be added in a separate PR.

seanstory:
buildkite test this

PR #2575, opened 2024-05-23
Use same aiomysql version on different platforms

oli-g:
This resolves a `PyMySQL` CVE which was shipped only with the version installed on aarch64.


sphilipse:
Related bug: https://github.com/elastic/connectors/issues/2584

sphilipse (approved):

Issue #2067, opened 2024-01-16
Unify datetime format used in connectors framework


jedrazb:
### Problem Description

We use different timestamp datetime formats for different fields. For programatic control perspective it's a bad UX to enforce different formats for different fields:

Example: `.elastic-connector` index (created in 8.12 deployment via Kibana, no use of Connector APIs). Different datetime formats in `filtering` and `last_seen` stats - first is populated by Kibana, last sync data is populated by framework

```
{

   "filtering":[
      {
         "domain":"DEFAULT",
         "draft":{
            "advanced_snippet":{
               "updated_at":"2024-01-16T11:02:39.245Z",
               "created_at":"2024-01-16T11:02:39.245Z",
               "value":{
                  
               }
            },
            ... 
   ],
   "index_name":"search-mysql",
   "last_seen":"2024-01-16T11:06:40.233382+00:00",
   "last_sync_status":"completed",
   "last_synced":"2024-01-16T11:00:06.809586+00:00"

}
```

Note: setting `"2024-01-16T11:02:39.245Z"` in last seen will break the framework in https://github.com/elastic/connectors/blob/main/connectors/protocol/connectors.py#L596

```
Python 3.10.13 (main, Aug 24 2023, 22:36:46) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> datetime.fromisoformat("2024-01-16T11:02:39.245Z")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: '2024-01-16T11:02:39.245Z'
```

It's even more interesting with sync jobs, note different formats between `created_at` and `started_at`.

```
{
  "created_at": "2024-01-16T10:59:35.334Z",
   ...
  "job_type": "full",
  "last_seen": "2024-01-16T11:00:06.544795+00:00",
  "metadata": {},
  "started_at": "2024-01-16T11:00:04.397788+00:00",
  "status": "completed",
  ...
}
```

### Proposed Solution

Have framework-level method to handle common datetime formats parsing,

### Alternatives

For now I'm just hardcoding serializing different datetime formats on the Connectors API level, but this is error-prone and hard to properly QA.

jedrazb:
I realised that the service is actually writing two different timestamps formats:
[...]

PR #2587, opened 2024-05-28
Migration for timezone aware timestamps

jedrazb:
## Relates to https://github.com/elastic/connectors/issues/2067

Followup to #2067. I realised that migrating from naive to offset aware timestamps for job scheduling might break existing connectors that had scheduling enabled and were using naive datatime (no tz info) representation. 

I tried this scenario, as expected after migrating to new offset aware logic the scheduling was broken ... I got a following error from a scheduler

```
  File "/Users/jedr/connectors-python/connectors/services/job_scheduling.py", line 207, in _should_schedule
    and last_sync_scheduled_at > last_wake_up_time
TypeError: can't compare offset-naive and offset-aware datetimes
```

The fix is to try to cast naive datetimes, to offset aware datetimes when reading `last_access_control_sync_scheduled_at`, `last_incremental_sync_scheduled_at` and  `last_sync_scheduled_at` from ES index. This essentially takes care of the migration. 

I modified `_property_as_datetime` function to try to set datetime to UTC timezone, with the `with_utc_tz` function. 

The `_property_as_datetime` is used by following properties in protocol logic:
- `last_incremental_sync_scheduled_at` this migration
- `last_access_control_sync_scheduled_at` this migration
- `last_sync_scheduled_at` this migration
- `last_seen` - already [indexed as offset aware timezone](https://github.com/elastic/connectors/blob/main/connectors/protocol/connectors.py#L158) (no effect on logic)

### Verification
- Tested migration manually (enable all sync scheduling, execute some syncs, upgrade connectors, scheduling still works fine)
- Unit tests

#### Pre-Review Checklist

### DIFF
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -597,7 +597,9 @@ class Connector(ESDocument):
         value = self.get(key)
         if value is not None:
             value = parse_datetime_string(value)  # pyright: ignore
-            # ensure that the datetime is in UTC timezone
+            # Ensure the datetime is in UTC. This handles historically naive timestamps
+            # that might be present in the index, ensuring the job scheduling logic with
+            # offset-aware timestamps works correctly.
             value = with_utc_tz(value)
         return value
 
--- a/connectors/protocol/connectors.py
+++ b/connectors/protocol/connectors.py
@@ -38,6 +38,7 @@ from connectors.utils import (
     nested_get_from_dict,
     next_run,
     parse_datetime_string,
+    with_utc_tz,
 )
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


connectors/protocol/connectors.py
```diff
@@ -596,6 +597,8 @@ def _property_as_datetime(self, key):
         value = self.get(key)
         if value is not None:
             value = parse_datetime_string(value)  # pyright: ignore
+            # ensure that the datetime is in UTC timezone
```
artem-shelkovnikov, line 600: I think it's also worth to add a comment here mentioning the problem found and saying that it's done here for backwards-compatibility

jedrazb: Okok, added the comment!

artem-shelkovnikov (approved):

artem-shelkovnikov:
@jedrazb any objections to backporting it back up to 8.13?

PR #2598, opened 2024-05-29
[Sharepoint Online] Replace deep dictionary gets with safer util function

bean710:
## Closes https://github.com/elastic/connectors/issues/2590

In some cases (like the one referenced in https://github.com/elastic/connectors/issues/2590), the Graph API will respond with a field that resolves to `None`. Because `get`'s default option is only used if the key is not found, the `get` will resolve as `None` (instead of the intended empty dict) and cause subsequent `get`s to raise an error.

This PR replaces deep `get`s with the util function `nested_get_from_dict` which will fall back to the default value in more cases:

https://github.com/elastic/connectors/blob/b8b66aa31dec2d48396c7e79e604e9290695c82f/connectors/utils.py#L963-L977

#### Pre-Review Checklist


artem-shelkovnikov:
buildkite test this

artem-shelkovnikov:
@bean710 the change looks good, but I see that the linters complained about it.

You can fix it by doing `make autoformat` and pushing the change into this branch, after CI is green I will approve it and backport to 8.14 and 8.13 versions.

artem-shelkovnikov:
buildkite test this

bean710:
Looks like the `(Any | None)` return type of the util function is a problem for linting.

Not sure what best practice here is, but this works:
```users = nested_get_from_dict(role_assignment, ["Member", "Users"], []) or []```
But seems redundant

artem-shelkovnikov:
Feel free to put `# pyright: ignore` for this line - it's done for other occurrences. We'll look into it sometime later maybe, ignoring is fine :)

artem-shelkovnikov:
buildkite test this

seanstory (approved):
Good stuff. Thanks for the contribution!

seanstory:
nested_get_from_dict already has tests, so I'm fine with us skipping adding additional specs here.

PR #2601, opened 2024-05-30
Change version after 8.11.4.1 release

seanstory:


navarone-feekery:
For clarification, when we have a patch release, should we be bumping to the next revision instead of the next patch?
e.g. version file is already `8.15.1.0` and `8.15.1.0` is released. Do we bump to `8.15.1.1` or `8.15.2.0` now?

navarone-feekery (approved):

seanstory:
I know we discussed this in the project sync today, but for breadcrumbs for later...

>  should we be bumping to the next revision instead of the next patch?

yeah probably - unless it's an emergency patch or something, where we know that there's another patch coming imminently that we'd rather target. 
In the long run, I'd like our release build to automate this version bumping. But for now, we can try to just catch this during PR reviews, and if we don't catch it, there's usually not much harm in changing the version back.

PR #2606, opened 2024-05-31
Catch job claim failures, and result in job termination

seanstory:
## Part of https://github.com/elastic/connectors/issues/2604

If a connector thinks a job is running for it, we'd better ensure that a job either starts or fails. Before this PR, if `sync_job.claim()` failed, the connector would not be informed that there was nothing running the job, and the job it expected to start running would never transition out of the pending state. This PR does a few things:

1. Ensures that after the connector record has been informed that the sync is starting, that everything past that is in a `try/except` block to handle any errors and fail the job.
2. wraps all SyncJob methods individually in `try/except` to ensure that any surfaced error messages are as actionable as possible
3. Adds tests

This PR does not yet address the root issue of the linked ticket, which is _why_ we're getting to a place where `claim()` is resulting in ES throwing a 413 error.

#### Pre-Review Checklist

## Release Note

Fixes a bug where sync jobs can be permanently  stuck in "pending" (until manually canceled) because of an unfortunately timed Elasticsearch exception. Now such exceptions will cause jobs to fail and provide actionable error messaging.

### DIFF
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -132,18 +132,19 @@ class SyncJobRunner:
 
         self.sync_job.log_debug(f"Starting execution of {job_type} sync job.")
 
-        await self.sync_starts()
-        sync_cursor = (
-            self.connector.sync_cursor
-            if self.sync_job.job_type == JobType.INCREMENTAL
-            else None
-        )
-        await self.sync_job.claim(sync_cursor=sync_cursor)
-        self._start_time = time.time()
-
-        self.sync_job.log_debug("Successfully claimed the sync job.")
+        await self.sync_starts()  # This MUST be the last line before the `try` begins
 
         try:
+            sync_cursor = (
+                self.connector.sync_cursor
+                if self.sync_job.job_type == JobType.INCREMENTAL
+                else None
+            )
+            self._start_time = time.time()
+            await self.sync_job.claim(sync_cursor=sync_cursor)
+
+            self.sync_job.log_debug("Successfully claimed the sync job.")
+
             self.data_provider = self.source_klass(
                 configuration=self.sync_job.configuration
             )
@@ -216,6 +217,7 @@ class SyncJobRunner:
             self.sync_job.log_error(e, exc_info=True)
             await self._sync_done(sync_status=JobStatus.ERROR, sync_error=e)
         finally:
+            self.sync_job.log_info("Terminating the job in error, and cleaning up")
             self.running = False
             if self.sync_orchestrator is not None:
                 await self.sync_orchestrator.close()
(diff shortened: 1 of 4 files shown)


connectors/sync_job_runner.py
```diff
@@ -216,6 +217,7 @@ async def execute(self):
             self.sync_job.log_error(e, exc_info=True)
             await self._sync_done(sync_status=JobStatus.ERROR, sync_error=e)
         finally:
+            self.sync_job.log_info("Terminating the job in error, and cleaning up")
```
jedrazb, line 220: Afaik the `finally` block is always executed, regardless of whether an exception was raised or not. So this log might be misleading. We are just cleaning up here

Out of scope: We are already terminating sync in `except` clauses above this line with `self._sync_done(sync_status=JobStatus.ERROR ....` perhaps we could add logs in each of the clause (e.g. we don't log anything if job is suspended / canceled)

connectors/sync_job_runner.py
```diff
@@ -374,7 +376,9 @@ async def _sync_done(self, sync_status, sync_error=None):
 
         if await self.reload_connector():
             sync_cursor = (
-                self.data_provider.sync_cursor()
+                None
+                if not self.data_provider  # If we failed before initializing the data provider, we don't need to change the cursor
```
jedrazb, line 380: Could the `None` sync_cursor cause problems down the line if there are intermittent issues with data provider initialization. 

I'm thinking of:
- we have connector, we run incremental syncs fine
- we once run into `if not self.data_provider` is `true` case
- we set this `None` cursor in connector doc `self.connector.sync_done(...`

How to recover for further incremental syncs (e.g. scheduled)? Can our logic communicate that user needs to run a full sync first given that the cursor is now `None`?

jedrazb (commented):
Great improvement to the framework! 👍  I left some comments

connectors/sync_job_runner.py
seanstory: omg you're totally right. I'd originally just added this as a "made it here" debugging log. I think I'll still leave something here for that purpose, but you're right, it shouldn't assume error state.

I can add extra logs too, that's easy.

connectors/sync_job_runner.py
seanstory: Sync cursor only gets set if if the job completed successfully: 
https://github.com/elastic/connectors/blob/main/connectors/protocol/connectors.py#L763-L764 

so we shouldn't be at risk here. If `self.data_provider` is not set, we didn't complete successfully.

jedrazb (approved):
LGTM!

seanstory:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.13|[](https://github.com/elastic/connectors/pull/2620)|

### Questions ?

PR #2610, opened 2024-06-03
[Confluence] Improve error handling in Confluence pagination

parth-elastic:
## Closes https://github.com/elastic/connectors-py/issues/2394

- Enhanced error handling mechanism to raise the error if exception is more than 30%. 
- The api_call function utilizes retry and re-raises errors if they are not handled.

#### Pre-Review Checklist

## Release Note

- api_call method utilizes retry and re-raises errors if they are not handled.
- Implemented a constant API_FAILURE_THRESHOLD set to 10% of the total number of indexed documents. An error will be raised if the exception rate surpasses this threshold.


connectors/sources/confluence.py
```diff
@@ -1178,6 +1268,16 @@ async def _consumer(self):
             else:
                 yield item
 
+    def check_api_exceptions_and_raise(self):
+        failed_percentage = (
+            self.confluence_client.api_failed_count
+            / self.confluence_client.api_total_count
+        ) * 100
+        if failed_percentage >= 30:
```
seanstory, line 1276: One day, I'd like to have logic similar to this at the framework level. But I have a few concerns about using error thresholds at the connector level like this.
1. first, this isn't currently configurable. IDK where you got the "30%" as a default, but I think you'd agree that no matter what default you pick, customers are going to want to change it. You didn't even do this as a constant that's easy to find and edit.
2. Even if it were configurable, it would need to be configurable specific to this connector. That means adding another RCF, and I don't think we want to keep polluting those.
3. Having this configurable for one connector makes all the other connectors look like they're missing the same feature.
4. Having this available is great, but it's only available for _some_ of the calls in this connector.

seanstory (changes requested):
I wish you'd started a discussion about how you'd go about solving this, before making a whole PR. I feel bad that you put in this work, but I don't agree with the approach taken here. I'd prefer that you revisit the Acceptance Criteria on the linked ticket, and take the approach suggested there.

parth-elastic:
We appreciate your perspective and understand your preference for discussing the solution before implementing it. According to the Acceptance Criteria, current approach involves utilizing retryable for error handling and re-throwing errors if they are not resolved. Furthermore, a mechanism is implemented to manage errors in a manner that prevents a large synchronization from failing due to minor data issues. This is achieved by establishing a threshold of 30% for the number of indexed documents before triggering an error, striking a balance between the imperative of quick error detection ("fail fast") and the necessity to prevent catastrophic failures in large synchronization processes.

connectors/sources/confluence.py
parth-elastic: > first, this isn't currently configurable. IDK where you got the "30%" as a default, but I think you'd agree that no matter what default you pick, customers are going to want to change it. You didn't even do this as a constant that's easy to find and edit.

Thank you for highlighting the lack of configurability and transparency in the default threshold setting. We acknowledge the need for flexibility in this aspect. If you believe it would be beneficial, we are open to discussing the possibility of removing the 30% logic to better align with user preferences.Please let us know your thoughts on this potential adjustment.

> Having this available is great, but it's only available for some of the calls in this connector.


The functionality mentioned is implemented for the api_call, it is also indirectly applicable to the paginated_api_call. This is because the paginated_api_call internally calls the api_call method to make API requests.

seanstory: Ok, after talking this over some with the team, I've changed my mind. We'll leave the threshold logic in this PR, and plan to revisit (potentially removing) when we're able to do this for the whole framework.

For now, can you pull the threshold default out as a constant, and let's set its value to 10%. For now, we won't expose as an RCF or via YAML config, but would require a code change in order to adjust the threshold.

Please add a release note section to this PR's description.

parth-elastic: Updated PR & Added release note, as you mentioned.

connectors/sources/confluence.py
```diff
@@ -221,17 +232,27 @@ async def api_call(self, url):
...
-            ) as response:
-                yield response
-        except ServerDisconnectedError:
-            await self.close_session()
-            raise
-        except ClientResponseError as exception:
-            await self._handle_client_errors(url=url, exception=exception)
+        if url != os.path.join(self.host_url, PING_URL):
+            self.api_total_count += 1
+        while True:
```
seanstory, line 237: Is this necessary? In the success case, we hit the `break` statement, and in all the `except` clauses we will end up re-raising (had to double check for `self._handle_client_errors()`, but I don't see a scenario where a `raise` does not happen). So in what case do we ever loop?

connectors/sources/confluence.py
```diff
@@ -221,17 +232,27 @@ async def api_call(self, url):
...
+                    yield response
+                    break
+            except ServerDisconnectedError:
+                await self.close_session()
+                raise
+            except ClientResponseError as exception:
+                await self._handle_client_errors(url=url, exception=exception)
+            except ClientPayloadError:
+                self.api_failed_count += 1
+                raise
```
seanstory, line 252: Is this redundant? Won't the `except Exception` below accomplish the same?

connectors/sources/confluence.py
```diff
@@ -221,17 +232,27 @@ async def api_call(self, url):
...
+            except Exception:
+                self.api_failed_count += 1
```
seanstory, line 254: Let's add debug logs every time `self.api_filed_count` is incremented

connectors/sources/confluence.py
```diff
@@ -258,9 +279,18 @@ async def paginated_api_call(self, url_name, **url_kwargs):
                         self.host_url,
                         links.get("next")[1:],
                     )
+            except ServerDisconnectedError:
+                raise
+            except ClientResponseError:
+                raise
```
seanstory, line 285:
```suggestion
            except (ServerDisconnectedError, ClientResponseError) as e:
                raise
```

If two exceptions are being handled the exact same way, no need to define the handling twice.

connectors/sources/confluence.py
```diff
@@ -258,9 +279,18 @@ async def paginated_api_call(self, url_name, **url_kwargs):
                         self.host_url,
                         links.get("next")[1:],
                     )
+            except ServerDisconnectedError:
+                raise
```
seanstory, line 283: Can you add a comment (or maybe even a debug log) to help future readers understand why these errors should be re-raised, instead of treated like the exceptions in the below blocks?

connectors/sources/confluence.py
```diff
@@ -258,9 +279,18 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+            except ClientResponseError:
+                raise
+            except ClientPayloadError as exception:
+                self._logger.error(
+                    f"Payload error occurred while fetching {url_name} from {url}. Exception: {exception}."
+                )
+                break
             except Exception as exception:
-                self._logger.warning(
-                    f"Skipping data for type {url_name} from {url}. Exception: {exception}."
```
seanstory, line 263: the "skipping" part of the message feels important to indicate that pagination terminated early. Can you make sure that this is still clear from the logs, that some requests may not have been executed because we prematurely exited pagination?

connectors/sources/confluence.py
```diff
@@ -258,9 +279,18 @@ async def paginated_api_call(self, url_name, **url_kwargs):
                         self.host_url,
                         links.get("next")[1:],
                     )
+            except ServerDisconnectedError:
+                raise
+            except ClientResponseError:
+                raise
+            except ClientPayloadError as exception:
```
seanstory, line 286: Can you add a comment to explain why this shouldn't re-raise?

connectors/sources/confluence.py
```diff
@@ -258,9 +279,18 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+                self._logger.error(
```
seanstory, line 287: Why ERROR and not WARN?

connectors/sources/confluence.py
```diff
@@ -258,9 +279,18 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+                    f"Payload error occurred while fetching {url_name} from {url}. Exception: {exception}."
+                )
+                break
             except Exception as exception:
-                self._logger.warning(
```
seanstory, line 262: Why did this change from a warning to an exception? Exception logging will print out the full stack trace and at ERROR level. But we're not crashing/re-raising here, so is this an error, or do we need to warn the user that there _might_ be an issue?

connectors/sources/confluence.py
```diff
@@ -258,9 +279,18 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+            except ServerDisconnectedError:
+                raise
+            except ClientResponseError:
+                raise
+            except ClientPayloadError as exception:
+                self._logger.error(
+                    f"Payload error occurred while fetching {url_name} from {url}. Exception: {exception}."
+                )
+                break
             except Exception as exception:
```
seanstory, line 291: This looks like we'll still be swallowing _most_ exceptions that occur while paginating. Can you explain why not switch to something like:
```
except ErrorWeCanSafelyIgnore:
  self._logger.warn("skipping paginating the rest, because of {good reason}")
except Exception:
  self._logger.error("We hit an unexpected error while paginating, and that could be a problem you want to handle")
  raise
```

connectors/sources/confluence.py
```diff
@@ -289,9 +319,18 @@ async def paginated_api_call_for_datacenter_syncrule(self, url_name, **url_kwarg
...
-                self._logger.warning(
-                    f"Skipping data for type {url_name} from {url}. Exception: {exception}."
+                self._logger.exception(
+                    f"Something went wrong while fetching {url_name} from {url}. Exception: {exception}."
                 )
                 break
```
seanstory, line 335: Let's DRY all this up, it looks identical to the error handling above.

connectors/sources/confluence.py
```diff
@@ -1178,6 +1269,16 @@ async def _consumer(self):
             else:
                 yield item
 
+    def check_api_exceptions_and_raise(self):
+        failed_percentage = (
+            self.confluence_client.api_failed_count
+            / self.confluence_client.api_total_count
+        ) * 100
+        if failed_percentage >= API_FAILURE_THRESHOLD * 100:
+            msg = f"High percentage of API exceptions: {failed_percentage}%. Please review the logs for more information."
```
seanstory, line 1278: Let's add the threshold to this error message too, and make it clear that this is calculated at the end of the sync.

connectors/sources/confluence.py
```diff
@@ -1231,3 +1332,6 @@ async def get_docs(self, filtering=None):
             async for item in self._consumer():
                 yield item
             await self.fetchers.join()
+
+        if self.confluence_client.api_total_count > 0:
```
seanstory, line 1336: should this be an even higher number? Like IDK if we really should call it a failure if there was 1 error out of 10 total API calls. That doesn't feel statistically significant. 

Or maybe we should just make this number another constant that we can let folks edit if they want to adjust the minimum threshold?

connectors/sources/confluence.py
```diff
@@ -333,6 +372,16 @@ async def fetch_server_space_permission(self, url):
...
+        except ServerDisconnectedError:
+            raise
+        except ClientPayloadError as exception:
+            self._logger.error(
+                f"Payload error occurred while fetching {url}. Exception: {exception}."
+            )
+        except Exception as exception:
+            self._logger.exception(
+                f"Something went wrong while fetching from {url}. Exception: {exception}."
+            )
```
seanstory, line 384: You've added this block in a lot of places. This is just swallowing _more_ errors that would currently have bubbled up. Why?

connectors/sources/confluence.py
```diff
@@ -1178,6 +1269,16 @@ async def _consumer(self):
             else:
                 yield item
 
+    def check_api_exceptions_and_raise(self):
+        failed_percentage = (
+            self.confluence_client.api_failed_count
+            / self.confluence_client.api_total_count
+        ) * 100
```
seanstory, line 1276:
```suggestion
        failed_percentage = round((
            self.confluence_client.api_failed_count
            / self.confluence_client.api_total_count
        ) * 100, 2)
```
let's limit the number of digits here so that the error message isn't obscene

connectors/sources/confluence.py
parth-elastic: > should this be an even higher number? Like IDK if we really should call it a failure if there was 1 error out of 10 total API calls. That doesn't feel statistically significant.


Adding this conditional check to just serves as a protective measure to prevent division by zero errors in the calculation process.

parth-elastic: > Or maybe we should just make this number another constant that we can let folks edit if they want to adjust the minimum threshold?

Already assigning the constant `API_FAILURE_THRESHOLD` a predefined value, folks have the option to easily adjust this threshold by editing the constant if needed to tailor the failure criteria to their requirements.

seanstory: Yes, but `N%` is not always significant. For example, if >10% of 1000000 requests fail, that's a LOT of failures. That probably means there's a significant issue. But if >10% of 2 requests fail, that might just be a fluke. You can address this by having two thresholds - one for a minimum sample size, and one for a percentage.

In Workplace Search, we had those two as well as a two others, (a window size, and a window percentage). So that you could fail if 30% of a window of 100 failed, or if 20% overall failed, but only if you had more than 1000 total. I don't think we need to get that complex, but I'm trying to illustrate the nuance that we've found important in the past.

connectors/sources/confluence.py
seanstory: I think you're misunderstanding my question. The issue this PR is attached to was filed because the connector wasn't raising enough errors. Instead of surfacing them, this _swallows_ more errors. The jobs would complete successfully, and you'd need to be monitoring the logs to see that there were issues. That's not a good user experience if there are significant problems occurring.

parth-elastic: In the current implementation, we raise exceptions specifically for ServerDisconnectError and ClientResponseError if not handled, while for any other encountered exceptions, we log the error and increment the `api_failed_count`.

seanstory: Yes, I understand that's what the code does in this branch. I'm saying that's probably not solving the issue raised in the linked ticket. 
If there's some error that isn't a ServerDisconnectError and isn't a ClientResponseError, by not re-raising it you don't let the caller determine how problematic the failure is. And if a critical request fails, but just that _one_ critical request, where a lot of less important requests succeed, the error threshold may not be high enough for the sync to register as a failure.

moxarth-elastic:
buildkite test this

moxarth-elastic:
buildkite test this

parth-elastic:
@seanstory The PR is ready for review

moxarth-elastic:
buildkite test this

moxarth-elastic:
@artem-shelkovnikov @seanstory we have implemented the changes, can we have another round of review?

connectors/sources/confluence.py
```diff
@@ -195,6 +215,17 @@ async def _handle_client_errors(self, url, exception):
 
             await self._sleeps.sleep(retry_seconds)
             raise ThrottledError
+        elif exception.status == 400:
+            self._logger.error(f"Bad Request Error (400). Exception: {exception}.")
+            raise UnauthorizedException
```
seanstory, line 220: Should this be "Unauthorized" here? For a 400?

connectors/sources/confluence.py
```diff
@@ -204,6 +235,39 @@ async def _handle_client_errors(self, url, exception):
...
+                    f"Error while reading value of retry-after header {exception}. Using default retry time: {DEFAULT_RETRY_SECONDS} seconds"
+                )
+        await self._sleeps.sleep(retry_seconds)
+        raise
+
+    async def _handle_api_call_error(self, url, exception):
+        known_errors = {
+            ServerConnectionError,
+            ClientResponseError,
+            ClientPayloadError,
```
seanstory, line 255: Why is this in the list if there's a `_handle_client_payload_errror()` function?

connectors/sources/confluence.py
```diff
@@ -204,6 +235,39 @@ async def _handle_client_errors(self, url, exception):
         else:
             raise
 
+    async def _handle_client_payload_error(self, exception):
+        retry_seconds = DEFAULT_RETRY_SECONDS
+        response_headers = exception.headers or {}
+        if "Retry-After" in response_headers:
```
seanstory, line 241: What about client payload errors makes it special and the only thing that should look at a retry-after header?

connectors/sources/confluence.py
```diff
@@ -377,19 +473,47 @@ async def fetch_confluence_server_users(self):
...
+                ServerConnectionError,
+                ClientResponseError,
+                ClientPayloadError,
+                Forbidden,
+                UnauthorizedException,
+                ThrottledError,
+                NotFound,
+                InternalServerError,
+                Exception,
+            ) as exception:
```
seanstory, line 494: What's the purpose in having this whole list, but including `Exception`? Isn't this the same as
```
```
?

seanstory (changes requested):
Sorry for the delay, for some reason your comments did not show up in my github notifications.

This PR is still not in a mergable state. There are still numerous comments from the last review that were not responded to or closed, and it's unclear if they were considered  at all.

This implementation raises more exceptions in some spots, swallows exceptions in other spots, but other than one place does not _handle_ the errors at all. It is unclear to me how this benefits the end user or solves the ticket it is linked to.

If it is still unclear how to move forward with this, we will close this PR and implement it internally.

parth-elastic:
> This PR is still not in a mergable state. There are still numerous comments from the last review that were not responded to or closed, and it's unclear if they were considered at all.

I've reviewed and addressed all previous comments. If I've missed anything, please let me know.

> This implementation raises more exceptions in some spots, swallows exceptions in other spots, but other than one place does not _handle_ the errors at all. It is unclear to me how this benefits the end user or solves the ticket it is linked to.

Ideally we are raising the exception in api_call method and the unknown exception is being swallowed if threshold is within the criteria and handling the exception where the api is being called. If you have any confusion or doubt redarding this please let us know

> If it is still unclear how to move forward with this, we will close this PR and implement it internally.

As per the certiera
We have tried to handled the exception in api_call whichever can we handled like ServerConnectionError,ClientResponseError,ClientPayloadError
If those exception are not handled we are re-raising them whereever the `api_call()` is called
And to balance the need to "fail fast" with the need to not crash a large sync because of a small data issue we are using the `API_FAILURE_THRESHOLD` and `MIN_API_CALL` variables which raise the exception if error percentage is more than 10%
If we are missing anything else please let us know

connectors/sources/confluence.py
parth-elastic: We are already raising `UnauthorizedException` for 401 updated BadRequestError for 400

connectors/sources/confluence.py
parth-elastic: To re-raised the exception if error is not handled by `_handle_client_payload_errror()`

connectors/sources/confluence.py
parth-elastic: Client payload errors often include a Retry-After header to manage rate limits and server load, indicating how long to wait before retrying, similar to how we handle it for OneDrive and SharePoint Online.

seanstory: But why not other error types too? Like 429 response codes should pretty much always include a `Retry-After` header

parth-elastic: You're right, 429 response codes should also include a Retry-After header. However, we already handle this logic in the `_handle_client_errors()`, which checks for the Retry-After header for 429 response code

seanstory:
I'm going to close this PR in favor of https://github.com/elastic/connectors/pull/2903. @moxarth-elastic if you'd like to contribute to that branch (it still needs unit tests), please go ahead. Otherwise, when I get to it, I'll re-assign the parent issue internally, and close it.

PR #2622, opened 2024-06-05
Fix multi-arch container image builds

acrewdson:
`buildah` is failing to build a multi-arch image when the image is Wolfi-based:

> ERRO[0010] While applying layer: ApplyLayer stdout:  stderr: operation not permitted exit status 1

We've seen in the past that switching to `docker` in these situations makes sense as a work-around.


acrewdson:
closing for now in favor of https://github.com/elastic/connectors/pull/2623

we can re-open if that approach doesn't work out

PR #2626, opened 2024-06-07
[Connector API] Validate and activate filtering with API call

jedrazb:
## Closes https://github.com/elastic/search-team/issues/7701

Next step after #2561

If the feature flag `feature_use_connectors_api` we call Connector APIs to update the filtering validation info and activate the filtering if valid.

- `PUT _connector/{connector_id}/_filtering/_validation` updates the draft validation info
- `PUT _connector/{connector_id}/_filtering/_activate` activates the draft


seanstory (approved):
nice

PR #2641, opened 2024-06-17
[Connector APIs] Connector update last sync info, status, error

jedrazb:
## Closes https://github.com/elastic/search-team/issues/7792

Use [update error api](https://www.elastic.co/guide/en/elasticsearch/reference/current/update-connector-error-api.html), [update status API](https://www.elastic.co/guide/en/elasticsearch/reference/current/update-connector-status-api.html) and [update last sync stats api](https://www.elastic.co/guide/en/elasticsearch/reference/current/update-connector-last-sync-api.html) to manage connector lifecycle during syncs.

This feature is behind a feature flag that is disabled by default.

### Validation
- added unit tests
- tested e2e on all sync types

#### Pre-Review Checklist


connectors/protocol/connectors.py
```diff
@@ -738,24 +743,43 @@ async def sync_starts(self, job_type):
...
+        if self.index.feature_use_connectors_api:
+            await self.index.api.connector_update_last_sync_info(
+                connector_id=self.id, last_sync_info=last_sync_information
+            )
+            await self.index.api.connector_update_status(
+                connector_id=self.id, status=Status.CONNECTED.value
+            )
+            await self.index.api.connector_update_error(
+                connector_id=self.id, error=None
+            )
```
artem-shelkovnikov, line 755: That's a little concerning - we do 3 calls to do single thing? Should we merge them into one call?

connectors/protocol/connectors.py
```diff
@@ -738,24 +743,43 @@ async def sync_starts(self, job_type):
...
-            "error": str(error),
-        }
-        await self.index.update(doc_id=self.id, doc=doc)
+        if self.index.feature_use_connectors_api:
+            await self.index.api.connector_update_error(
+                connector_id=self.id, error=error
+            )
+            await self.index.api.connector_update_status(
+                connector_id=self.id, status=Status.ERROR.value
+            )
```
artem-shelkovnikov, line 776: Same here - this is an atomic action "mark as error" that updates status and writes error, should it be single action?

connectors/protocol/connectors.py
```diff
@@ -806,7 +828,19 @@ async def sync_done(self, job, cursor=None):
...
+        if self.index.feature_use_connectors_api:
+            await self.index.api.connector_update_status(
+                connector_id=self.id, status=connector_status.value
+            )
+            await self.index.api.connector_update_error(
+                connector_id=self.id, error=job_error
+            )
+            await self.index.api.connector_update_last_sync_info(
+                connector_id=self.id, last_sync_info=last_sync_information
+            )
```
artem-shelkovnikov, line 840: And same thing here - is there any chance we unite these three into one?

artem-shelkovnikov (commented):
Left some questions - currently some of the calls are updating several fields at once in a way that does not make connector enter invalid state.

With new api changes I see that it's possible that only partial updates are applied to the records in connectors index (e.g. error is populated, but status is not changed) if something goes wrong - CTRL+C, network blip, Elasticsearch crashing, etc

connectors/protocol/connectors.py
jedrazb: Agreed that we could unify this in a following way:
- we just call `_error` endpoint, the logic in ES could set the status depending if `error` is null or not

connectors/protocol/connectors.py
jedrazb: See answer above about other 3 calls

jedrazb:
> With new api changes I see that it's possible that only partial updates are applied to the records in connectors index (e.g. error is populated, but status is not changed) if something goes wrong - CTRL+C, network blip, Elasticsearch crashing, etc

@artem-shelkovnikov Thank you for your review. While I agree that making several calls in a non-atomic block is not ideal, if we pursue the path of creating a single call that updates the connector document in the given scenario, we could end up with numerous endpoints or go back to what the OG `_update` endpoint was doing (single endpoint many fields in payload). 

I propose we do a slight improvement to error endpoint: if we set `error` to `null` it means that connector is in `connected` state, for non-null errors it means we ended up in `error` state. This can eliminate one call from that block.

artem-shelkovnikov:
@jedrazb my main concern is not performance, but changing the system into invalid state with API - even if it happens for 1-2 seconds.

Optimisation of calls on the other hand I think is not as important - as you mentioned, we will end up with just `_update` endpoint in the end.

connectors/protocol/connectors.py
artem-shelkovnikov: That would be perfect IMO!

connectors/protocol/connectors.py
artem-shelkovnikov: Agreed, uniting status update + error update in one endpoint and leaving `update_last_sync_info` endpoint separately is a good way forward

jedrazb:
> my main concern is not performance, but changing the system into invalid state with API - even if it happens for 1-2 seconds.

Tbh as long as we can update current error string and status we should be fine (so let's optimise this into a single request, update error with status side-effect).

The `last_sync_*` stuff could in theory fail as this is purely for informational purposes in Kibana, see: 
Our operational logic doesn't depend on this in `elastic/connectors` repo or Kibana ([reference](https://github.com/elastic/kibana/blob/4468ce52c373eac076d3f25e71d7697fab467431/x-pack/plugins/enterprise_search/public/applications/enterprise_search_content/utils/connector_status_helpers.ts#L49), usage)

PR #2642, opened 2024-06-18
try out pushed perf8 branch

seanstory:
Over the weekend, perf8 started causing issues as it was pinning a `matplotlib` version that was incompatible with a version of `numpy` that another dep was pulling in.

This PR aims to make use of the changes made here: https://github.com/elastic/perf8/pull/17

We won't actually merge this. Once this confirms that perf8 is working, we can close this PR, merge the perf8 PR, and CI should go back to working 🤞


seanstory:
CI is green! Looks like that worked. Closing this PR now.

PR #2647, opened 2024-06-20
[MySQL] Add new field id_columns in sync rules to generate id and adding support for views

parthpuri-elastic:
## Closes https://github.com/elastic/connectors-py/issues/467

Added new field `id_columns` in advanced sync rules to ingest tables which do not have a primary key. So, now table uses id_columns to generate unique ids. Also using id_columns users is able to fetch data of views.

#### Pre-Review Checklist

## Release Note

Added a new optional field in MySQL advanced sync rules `id_columns`. The user needs to pass the name of unique fields of the table in this field, and it will be used to generate a unique ID for documents.


parthpuri-elastic:
buildkite test this

parthpuri-elastic:
buildkite test this

connectors/sources/mysql.py
```diff
@@ -486,12 +487,12 @@ async def get_docs(self, filtering=None):
             for query_info in advanced_rules:
                 tables = query_info.get("tables", [])
                 query = query_info.get("query", "")
-
+                id_columns = query_info.get("id_columns", [])
                 self._logger.debug(
```
seanstory, line 491: let's update the log message too, to include `id_columns`

connectors/sources/mysql.py
```diff
@@ -569,6 +573,14 @@ async def _yield_docs_custom_query(self, client, tables, query):
         )
         column_names = await client.get_column_names_for_query(query=query)
 
+        if not set(primary_key_columns) - set(column_names):
+            pass
+        else:
+            self._logger.warning(
+                f"Skipping query {query} for tables {', '.join(tables)} as primary key column name/id_column is not present in query."
+            )
+            return
```
seanstory, line 582:
```suggestion
        if set(primary_key_columns) - set(column_names):
            self._logger.warning(
                f"Skipping query {query} for tables {', '.join(tables)} as primary key column name/id_column is not present in query."
            )
            return
```

this feels less convoluted.

tests/sources/test_mysql.py
```diff
@@ -667,11 +677,12 @@ async def test_validate_config_when_host_empty_then_raise_error():
 
 
 @pytest.mark.parametrize(
-    "tables_present_in_source, advanced_rules, expected_validation_result",
+    "tables_present_in_source, advanced_rules, id_in_source, expected_validation_result",
```
seanstory, line 680: Can you include tests to make sure that things still work if `id_in_source` is not provided? It looks like you modified all the pre-existing tests to specify column ids, but this change needs to remain backwards compatible.

seanstory (commented):
Please add a release note to the pr description, and the `release-note` label. Also, have you coordinated making sure this feature gets documented?

parthpuri-elastic:
> Please add a release note to the pr description, and the `release-note` label. Also, have you coordinated making sure this feature gets documented?

Sure, @leemthompo can you please add these example for MySQL Sync Rules

## For Tables without primary keys

If there are tables without primary keys, we can index those tables by providing at least one unique field called `id_columns` with advanced sync rule.

#### Single unique field

```json
[
  {
    "tables": ["table1"],
    "query": "SELECT * FROM table1"
    "id_columns": ["column1"]
  }
]
```

#### Multiple Unique fields

```json
[
  {
    "tables": [
      "table1",
      "table2"
    ],
    "query": "SELECT table1.id AS id, table2.order_id AS order_id, table1.name AS user_name, table2.product_name AS product_name, table2.order_date AS order_date FROM table1 JOIN table2 ON table1.id = table2.user_id",
    "id_columns": [
      "id",
      "order_id"
    ]
  }
]
```

tests/sources/test_mysql.py
```diff
@@ -666,6 +676,214 @@ async def test_validate_config_when_host_empty_then_raise_error():
...
+        (
+            # invalid: id_columns field missing
+            [TABLE_ONE],
+            [ID_ONE],
+            [{"tables": [TABLE_ONE], "query": "SELECT * FROM *"}],
+            SyncRuleValidationResult(
+                SyncRuleValidationResult.ADVANCED_RULES,
+                is_valid=False,
+                validation_message=ANY,
+            ),
```
seanstory, line 784: This means this is a backwards incompatible change, right? I don't think that's ok. `id_columns` needs to be optional.

seanstory:
buildkite test this

parthpuri-elastic:
buildkite test this

tests/sources/test_mysql.py
parthpuri-elastic: Yes, this field is optional and this will work without `id_columns`. it fails because of wrong arguments. I have updated it.

parthpuri-elastic:
@seanstory, I have updated the PR with suggested changes. Can you please re-review it.

artem-shelkovnikov:
@seanstory any concerns about this PR?

artem-shelkovnikov (approved):

praveen-elastic:
Buildkite test this

praveen-elastic:
Buildkite test this

seanstory:
Sorry y'all. I was very behind on my notifications.  🙈 
Thanks for wrapping it up

PR #2649, opened 2024-06-21
Add set_env to tests that init elasticsearch client

artem-shelkovnikov:
Example failures:

```

FAILED tests/protocol/test_connectors.py::test_create_job_with_connector_api[JobTriggerMethod.SCHEDULED-JobType.INCREMENTAL] - ValueError: Strict mode enabled, variables $elasticsearch.password are not defined!
--
  | FAILED tests/protocol/test_connectors.py::test_create_job_with_connector_api[JobTriggerMethod.ON_DEMAND-JobType.FULL] - ValueError: Strict mode enabled, variables $elasticsearch.password are not defined!
```

It happens because the tests are not using `set_env` fixture - original configuration file that's used for these tests does not have actual elasticsearch configuration definitions, it uses variables. These variables are populated from ENV.

Sometimes these tests pass because previous tests already defined these environment variables. To avoid these tests being flaky I just use `set_env` fixture in more tests.

### DIFF
--- a/tests/protocol/test_connectors.py
+++ b/tests/protocol/test_connectors.py
@@ -797,7 +797,7 @@ async def test_sync_job_claim():
 
 
 @pytest.mark.asyncio
-async def test_sync_job_claim_with_connector_api():
+async def test_sync_job_claim_with_connector_api(set_env):
     source = {"_id": "1"}
     index = Mock()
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov:
cc @jedrazb, that's some arcane magic here :D

seanstory (approved):
Nice find. I can see this biting us again, but I don't have a suggestion on a better approach.

PR #2650, opened 2024-06-21
Check ES version compatibility and build flavor on startup

navarone-feekery:
## Closes https://github.com/elastic/connectors/issues/2305

Add a preflight check to confirm the Connector is compatible with the configured ES server.
This check also determines if the server is Serverless or not, and injects that state into the ES config so it can be referenced later without repeated calls to the ES server.
If the server is Serverless, Connectors does not perform the version compatibility check.

See the issue for the version compatibility matrix.

I'm unsure if injecting the `elasticsearch.serverless` bool into the config is the best approach. If it feels too invasive I'd like to discuss other options (e.g. checking build flavor whenever needed rather than once during preflight check).

#### Pre-Review Checklist


connectors/preflight_check.py
```diff
@@ -40,22 +41,98 @@ async def run(self):
...
+
+        # major
+        if es_version_parts[0] != connector_version_parts[0]:
+            logger.critical(
+                f"Elasticsearch {es_version} and Connectors {self.version} are incompatible: major versions are different"
+            )
+            return False
+
+        # minor
+        if es_version_parts[1] > connector_version_parts[1]:
```
seanstory, line 112: Should this check for non-equality, rather than only if ES is greater?

Like if connectors is a higher version, and needs an ES Connector API that doesn't exist in the targeted ES deployment, we'll have issues.

seanstory: Actually, I see below that you check for `>` separately. I think we could just check both together, and error that they're non-equal. Could probably even do that with the major version check.

Also, the docstring says you only check that the ES minor isn't greater.

connectors/es/client.py
```diff
@@ -47,9 +46,7 @@ class ESClient:
     user_agent = f"{USER_AGENT_BASE}/service"
 
     def __init__(self, config):
-        # We don't have a way to ask the server, but it's planned
-        # for now we just use an env flag
-        self.serverless = "SERVERLESS" in os.environ
```
seanstory, line 52: Do we have docs anywhere for this SERVERLESS env var that we need to edit/remove?

connectors/service_cli.py
```diff
@@ -40,12 +40,16 @@ async def _start_service(actions, config, loop):
...
     for sig in (signal.SIGINT, signal.SIGTERM):
         loop.add_signal_handler(sig, functools.partial(preflight.shutdown, sig))
     try:
-        if not await preflight.run():
+        success, is_serverless = await preflight.run()
+        if not success:
             return -1
+
+        # injecting this value into the config allows us to avoid checking the server again before requests
+        config["elasticsearch"]["serverless"] = is_serverless
```
seanstory, line 52: by putting this in the service cli, will this break if you use Dmitrii's CLI?

connectors/es/client.py
```diff
@@ -162,36 +159,40 @@ async def wait(self):
...
-        return False
+        return None
 
     async def ping(self):
         try:
-            await self.client.info()
+            response = await self.client.info()
+            logger.info(
+                f"Connected to Elasticsearch server with version '{response['version']['number']}' and build flavor '{response['version']['build_flavor']}'"
+            )
```
seanstory, line 184: Is this going to be logged frequently? If so, we may want to make it `debug`

connectors/preflight_check.py
```diff
@@ -40,22 +41,98 @@ async def run(self):
...
+        Returns two values:
+
+        1. Whether or not the ES server is compatible with the running Connectors version
+        2. Whether or not the ES server is serverless
+        """
+
+        es_info = await self.es_management_client.wait()
+        if es_info is None:
+            logger.critical(
+                f"{self.elastic_config['host']} seems to be unreachable. Bye!"
```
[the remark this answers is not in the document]
seanstory: ah, I see this just came from above.

seanstory (commented):
I don't love putting `serverless` in the config, but I see why you landed there. I can't come up with an objectively better approach. We could set it in an ENV var, or in a global variable. But I don't feel any better about those. I think it's fine to do for now, especially if we don't document it, so we can reserve the right to change it if necessary.

jedrazb (commented):
Great stuff! Should include the compatibility matrix in docs as a part of this PR? you can copy paste from https://github.com/elastic/connectors/issues/2305

connectors/preflight_check.py
navarone-feekery: > Like if connectors is a higher version, and needs an ES Connector API that doesn't exist in the targeted ES deployment, we'll have issues.

This is a good point. I think we could have it a bit different from the Kibana compatibility matrix as we have this dependency to consider.

navarone-feekery: > Also, the docstring says you only check that the ES minor isn't greater.

I had a line at the end stating `If the above is satisfied but the versions are not exactly the same we log a warning` which I thought could be a catch-all for other inconsistencies, but I can see it just caused confusion.

connectors/es/client.py
navarone-feekery: It's not referenced in our docs for ingesting into serverless: https://www.elastic.co/docs/current/serverless/elasticsearch/ingest-data-through-integrations-connector-client
Kind of surprised at that!

connectors/es/client.py
navarone-feekery: It should only happen on startup but I made it debug anyway, as the version compatibility check will output something afterwards anyway.

connectors/service_cli.py
navarone-feekery: I tested all the CLI commands and nothing breaks from this.

navarone-feekery: Tested:

- `bin/connectors connector`
- `bin/connectors job`
- `bin/connectors index`

connectors/preflight_check.py
navarone-feekery: Actually I had implemented this wrong.

If the ES minor is greater than Connectors it should terminate. If the ES minor is lower than Connectors it should output a warning. I had these the wrong way around.

I'd prefer to keep them separate so we have the same compatibility as Kibana does with ES.

navarone-feekery:
@seanstory @jedrazb thanks for the reviews!
I applied some changes, but also I realised that `verify.py` from ftests was still referencing the env var so I updated that.

connectors/es/client.py
seanstory: I bet it was originally for the serverless CI pipeline we had.

connectors/service_cli.py
seanstory: you tested those commands against a serverless project? I'm surprised it worked. I must not fully understand how the CLIs interact with one another. I thought those commands would skip this code. 👀

connectors/preflight_check.py
seanstory: 👍 I suppose that makes sense. It assumes that Elasticsearch will be backwards-compatible on a minor upgrade, which is a reasonable assumption.

connectors/service_cli.py
navarone-feekery: > you tested those commands against a serverless project?

Haha.. ha.... one second.

navarone-feekery: I confirmed this works with serverless. I was able to login, create a connector, start a sync job, and delete the index docs without issue.

navarone-feekery: > I thought those commands would skip this code

This code is only run on startup (so `bin/elastic-ingest`). The CLI doesn't really care if something is serverless, it just wants a host and some credentials.

README.md
```diff
@@ -76,3 +76,25 @@ The framework serves two distinct, but related use cases:
...
+> Version compatibility will not be checked if Elasticsearch is serverless.
+
+The Connector will perform a version compatibility check with the configured Elasticsearch server on startup.
+If the versions are incompatible, the Connector will terminate and output the incompatible versions in the shell.
+If the versions are different but otherwise compatible, the Connector will output a warning in the shell but will continue operating.
+
+We recommend running on the same version as Elasticsearch.
+However, if you want to hold back upgrading one or the other for any reason, use this table to determine if your versions will be compatible.
+
+| Situation                 | Example Conenctor Framework version | Example ES version | Outcome |
```
seanstory, line 92:
```suggestion
| Situation                 | Example Connector Framework version | Example ES version | Outcome |
```

README.md
```diff
@@ -76,3 +76,25 @@ The framework serves two distinct, but related use cases:
...
+| ------------------------- |-------------------------------------|--------------------| ------- |
+| Versions are the same.    | 8.15.1                              | 8.15.1             | 💚 OK      |
```
seanstory, line 94:
```suggestion
| Versions are the same.    | 8.15.1.0                              | 8.15.1             | 💚 OK      |
```

Doesn't really matter, but might be good to include BUILD versions in these examples

connectors/preflight_check.py
seanstory: wait, no, that's the opposite of what I said. 😕 
Where does Kibana's compatibility get defined? I'm surprised that an older kibana can't talk to a newer Elasticsearch.

navarone-feekery: @jedrazb on the issue is a version compatibility matrix and reference to it being similar to Kibana but there's no linnk, do you remember where the information came from?

README.md
navarone-feekery: Added, also added a row specifying that build version doesn't matter.

connectors/service_cli.py
seanstory: ok, looking closer, I think I get it. I was "right", that `bin/connectors` won't set `config["elasticsearch"]["serverless"]`. But since it _also_ doesn't do the preflight check, and that config isn't used anywhere else, it doesn't matter.

Eventually we may need to make sure that we're always aware of if our destination is a Serverless project, but we can cross that bridge when we come to it.

🙈 Sorry for the noise

connectors/preflight_check.py
navarone-feekery: @seanstory the version compatibility table is on Kibana's main README: https://github.com/elastic/kibana?tab=readme-ov-file#version-compatibility-with-elasticsearch

seanstory: > If the ES minor is greater than Connectors it should terminate. If the ES minor is lower than Connectors it should output a warning

This is the _opposite_ of what's in that table. 

Screenshot 2024-06-26 at 1 14 31 PM

navarone-feekery: I think in my mind I was treating "newer" == "lower number" like its an age 🤦🏻

navarone-feekery: I fixed it and changed the wording to use `older` and `newer` so it's easier to mentally parse compared to the table.

seanstory (approved):

PR #2653, opened 2024-06-25
Add service.idling: 1 to all ftests + decrease size of ftest for gcs

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/2638

Functional tests became a bit slower after change in https://github.com/elastic/connectors/commit/b8b66aa31dec2d48396c7e79e604e9290695c82f#diff-5f4b320c98a5619e1197864adbca34517cf6ce4aafdb3d17ca5079dc859998a1: now cron does not respect seconds at all (which is impossible to change in Kibana anyway), so tests started to run a bit slower - full sync is scheduled 30/60 seconds later.

This PR slightly speeds up functional tests by making `service.idling: 1` making service poll for jobs and try to schedule jobs every second.

Additionally this PR decreases the GCS functional test payload by around 10% so that it would always pass. Now test takes around 500-550 seconds and with scheduling taking up to 90 seconds it could exceed the threshold of 600 seconds that ftests has to run the first sync.

One small change is just standartisation of all ftest config.yml files: deleting unnecessary lines.

#### Pre-Review Checklist


seanstory (approved):
LGTM

PR #2661, opened 2024-06-26
Ensure orchestrator tasks are always canceled

seanstory:
## relates to https://github.com/elastic/connectors/issues/2638

We've been having an issue with the GCS ftest taking > 10 minutes. Digging with @artem-shelkovnikov today, we found that there was a sync that received a SIGINT, and acted like it was stopping/cleaning up (printed job counters and everything) but the extractor clearly kept going. Digging further, we found a logical error, where the SyncOrchestrator's tasks were only being canceled if there was an "error" in the job.

This change instead makes sure that `orchestrator.close()` always cancels the tasks if they are still running. It also makes sure that the SyncJobRunner's `_sync_done()` function always calls `orchestrator.cancel()`. There is no risk in calling `cancel()` multiple times.

#### Pre-Review Checklist

## Release Note

Fixes a bug where the framework might receive an interrupt, but not terminate all running sync job tasks correctly.

### DIFF
--- a/connectors/connectors_cli.py
+++ b/connectors/connectors_cli.py
@@ -633,7 +633,7 @@ cli.add_command(job)
 
 
 def main(args=None):
-    cli()
+    cli()  # pyright: ignore
 
 
 if __name__ == "__main__":
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -817,6 +817,7 @@ class SyncOrchestrator:
 
     async def close(self):
         await self.es_management_client.close()
+        await self.cancel()
 
     async def has_active_license_enabled(self, license_):
         # TODO: think how to make it not a proxy method to the client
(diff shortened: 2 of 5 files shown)


connectors/connectors_cli.py
```diff
@@ -633,7 +633,7 @@ def view_job(obj, job_id, output_format):
 
 
 def main(args=None):
-    cli()
+    cli()  # pyright: ignore
```
seanstory, line 636: This is unrelated to my changes, but `make lint` was failing locally with:
```
/Users/seanstory/Desktop/Dev/connectors-python/connectors/connectors_cli.py
  /Users/seanstory/Desktop/Dev/connectors-python/connectors/connectors_cli.py:636:5 - error: Arguments missing for parameters "ctx", "config" (reportGeneralTypeIssues)
```

artem-shelkovnikov (approved):
Great catch! 🚀 🌔

PR #2663, opened 2024-06-27
Fix flakey interrupt test

seanstory:
We've seen a few tests fail in buildkite with:
```
Exited with status 255 (after intercepting the agent's termination signal, sent because the job was canceled)
```

Looking deeper, those seem to have been coming because all the tests were taking > 5 min.
What's the long pole in this case? Looks like:

```
2024-06-27 14:29:01 UTC | tests/test_service_cli.py::test_main_exits_on_sigterm PASSED
2024-06-27 14:32:13 UTC | tests/test_service_cli.py::test_shutdown_called_on_shutdown_signal[Signals.SIGINT] # Received cancellation signal, interrupting
2024-06-27 14:32:13 UTC | FAILED
```

Notice the 3 minute gap.

Two problems here:
1. probably there is a race condition in the `test_shutdown_called_on_shutdown_signal` test, causing the CLI to not get the near-immediate interrupt signal, and instead waiting to shut down naturally when a connection to ES isn't established (takes a few minutes)
2. we have no logs to use for investigating. The test uses `patch_logger`, so we need to keep that capability, while also dumping the logs for debugging.

This PR attempts to fix both. It increases the `asyncio.sleep` to avoid the race condition (and increases the slow test threshold accordingly) and makes changes to PatchLogger to make `silent=False` actually work.

Finally, it also increases the buildkite timeout, so that we can get the full test output in these scenarios.

#### Pre-Review Checklist

### DIFF
--- a/tests/test_service_cli.py
+++ b/tests/test_service_cli.py
@@ -56,7 +56,7 @@ def test_version_action(option):
     assert __version__ in result.output
 
 
-@pytest.mark.parametrize("sig", [signal.SIGINT, signal.SIGTERM])
+@pytest.mark.parametrize(["sig", 'patch_logger'], [[signal.SIGINT, False], [signal.SIGTERM, False]], indirect=["patch_logger"])
 @patch("connectors.service_cli.PreflightCheck")
 def test_shutdown_called_on_shutdown_signal(
     patch_preflight_check, sig, patch_logger, mock_responses, set_env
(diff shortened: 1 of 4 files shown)


tests/test_service_cli.py
```diff
@@ -56,21 +56,20 @@ def test_version_action(option):
     assert __version__ in result.output
 
 
-@pytest.mark.parametrize("sig", [signal.SIGINT, signal.SIGTERM])
+@pytest.mark.parametrize(["sig", 'patch_logger'], [[signal.SIGINT, False], [signal.SIGTERM, False]], indirect=["patch_logger"])
```
navarone-feekery, line 59: Just curious, what does `indirect` do here?

navarone-feekery (dismissed):
LGTM

tests/test_service_cli.py
seanstory: TIL about it! It's how you pass arguments to pytest fixtures. So this is saying, "For the parametrized `patch_logger` values, don't actually pass them directly as parameters, but as parameters of that parameter". Check out: https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

seanstory:
Whelp, the failing test tells me that this didn't actually help anything 💀 
https://buildkite.com/elastic/connectors/builds/13321#01905abe-19d3-4f85-b847-f5057eed044f

Closing.

PR #2664, opened 2024-06-28
Attempt to fix the flaky test

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/2517

Problem seems to happen because the test is running interrupt in the wrong moment - not when the service has started. Therefore when SIGINT/SIGTERM is issued, service is not there yet and cannot start the shutdown procedure.

This PR attempts to fix it by changing the code that issues SIGTERM/SIGINT: now it reads the logger and issues the signal only after a log line saying that service has started is present in the logs.

#### Pre-Review Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,6 +43,16 @@ class Logger:
                 if isinstance(log, str) and msg in log:
                     raise AssertionError(f"'{msg}' found in {self.logs}")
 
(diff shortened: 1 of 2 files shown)


navarone-feekery (approved):
Nice one 🚀

artem-shelkovnikov:
Does not work

PR #2667, opened 2024-07-01
Bump requests

seanstory:
Snyk flagged an issue with `requests<2.32`:
```
When making requests through a Requests Session, if the first request is made with verify=False to disable cert verification, all subsequent requests to the same origin will continue to ignore cert verification regardless of changes to the value of verify. This behavior will continue for the lifecycle of the connection in the connection pool.
```
bumps `requests` to the latest version.

#### Pre-Review Checklist


artem-shelkovnikov (commented):
@seanstory you accidentally committed some files, can you revert them before approval? :)

seanstory:
woops! Thanks. 🙈 
Fixed

artem-shelkovnikov (approved):

PR #2671, opened 2024-07-02
Port ErrorMonitor class from ruby connectors

artem-shelkovnikov:
Porting the ErrorMonitor class from https://github.com/elastic/connectors-ruby/blob/main/lib/utility/error_monitor.rb

ErrorMonitor class is there to add transient error handling and ignore errors until certain threshold.

For example, class can be used with the following code:
```
async def fetch_some_info(self, id):
    with self.with_error_monitoring():
        yield self.client.fetch_some_not_critical_entity(id)
```

Alternative way of usage:

```
try:
    entity = self.client.query(id=id)
    self.error_monitor.track_success()
except Exception as ex:
    self.error_monitor.track_error(ex)
```

What it will do:

- Log the error
- Ignore it UNLESS too many errors happened.

What does "too many errors" mean?
1. Too many errors happened over the sync - default number is 1000
2. Too many errors happened consecutively within the blocks `with self.with_error_monitoring()` - default number is 10
3. Too many errors happened in the window of X blocks `with self.with_error_monitoring()` - by default it's 15%

This PR makes use of ErrorMonitor for generic framework operations.

For now error monitor is created and passed only to Sink, because this change is simple enough and easy to understand.

Sink uses error monitor to track errors while ingesting data - each bulk operation is checked and each document that was not ingested into Elasticsearch triggers an "error" tracking, each successfully ingested document produces "success" tracking. In practice with default settings it means, for example, that if 100 documents are sent to Elasticsearch and 16 of them fail, sync will fail.

If sync is terminated due to this error, Kibana UI will show reason (too many errors in total, too many consequent errors, error rate is too high) along with the last error that happened.

So, as an example, if during a sync 200 documents fail to be downloaded and 801 documents fail to be ingested to Elasticsearch, sync will fail.


connectors/utils.py
```diff
@@ -994,3 +994,101 @@ def get(self, key) -> int:
...
+
+
+class ErrorMonitor:
+    def __init__(
+        self,
+        max_total_errors=1000,
+        max_consecutive_errors=10,
+        max_error_rate=0.15,
+        error_window_size=100,
+        error_queue_size=20,
```
seanstory, line 1012: Can we get these configurable through YAML?

seanstory (approved):
Love it. 
I'm fine to have the configuration be a follow-up. Just as long as it isn't used before it's configurable.

connectors/utils.py
artem-shelkovnikov: Yes, absolutely!

config.yml.example
```diff
@@ -203,6 +203,37 @@
...
 #
 #
+## ------------------------------- Service  Error Monitor ----------------------------------
+#
+##  Configirations related to Error Monitor functionality of the syncs.
+##    Each running sync has an error monitor attached to it. Error monitor is taking care of
+##    ignoring transient errors while ingesting the data. For example, failing to download or ingest a single
+##    document should not stop the sync. Failing to ingest some meaningful number of documents, however,
+##    should stop the sync.
+##    Parameters in this section are responsible for how many transient errors the error monitor will swallow
```
seanstory, line 213:
```suggestion
##    Parameters in this section are responsible for how many transient errors the error monitor will move past
```

Don't want to alarm anyone

config.yml.example
```diff
@@ -203,6 +203,37 @@
...
+##    before marking the sync as failed because too many transient errors happened.
```
seanstory, line 214:
```suggestion
##    before marking the sync as failed because too many transient/minor errors happened.
```

config.yml.example
```diff
@@ -203,6 +203,37 @@
 #service.job_cleanup_interval: 300
 #
 #
+## ------------------------------- Service  Error Monitor ----------------------------------
```
seanstory, line 206: Should we have a `service.error_monitor.enabled: true`? So that this doesn't immediately start getting used in Native Connectors without us having a way to expose the configs there? (fine if you'd rather that be in a follow-up)

connectors/es/sink.py
```diff
@@ -449,27 +457,37 @@ def __init__(
...
+                {
+                    "_op_type": operation,
+                    "_index": self.index,
+                    "_id": doc_id,
+                    "doc": doc,
+                }
+            )
+            self.error_monitor.track_success()
+        except ForceCanceledError:
+            raise
```
seanstory, line 485: Should we also re-raise aio CanceledError?

connectors/es/sink.py
```diff
@@ -599,9 +617,17 @@ async def get_docs(self, generator, skip_unchanged_documents=False):
...
+                lazy_downloads.raise_any_exception()
+
                 await asyncio.sleep(0)
+
+            # Sit and wait until an error happens
+            await lazy_downloads.join(raise_on_error=True)
         finally:
             # wait for all downloads to be finished
+            # even if we errored out
             await lazy_downloads.join()
```
seanstory, line 631: Since we call `await lazy_downloads.join(raise_on_error=True)` above in the success case, we'd only get down here without having already called `join` if an error was raised. In that case, should we do `lazy_downloads.cancel()`?

connectors/es/sink.py
```diff
@@ -884,7 +911,7 @@ def _extractor_task_running(self):
     async def cancel(self):
         if self._sink_task_running():
             self._logger.info(
-                f"Cancling the Sink task: {self._sink_task.name}"  # pyright: ignore
+                f"Canceling the Sink task: {self._sink_task.get_name()}"  # pyright: ignore
```
seanstory, line 914: 🤦 thanks

connectors/source.py
```diff
@@ -767,12 +780,14 @@ async def download_and_extract_file(
                 doc = await self.handle_file_content_extraction(
                     doc, source_filename, temp_filename
                 )
+            self.error_monitor.track_success()
```
seanstory, line 783: won't this end up double-counting? Since `sink.py` also uses the monitor for the lazy downloads?

connectors/source.py
```diff
@@ -582,6 +586,15 @@ async def get_access_control(self):
...
         raise NotImplementedError
 
+    @contextmanager
+    def with_error_monitoring(self):
+        try:
+            yield
+            self.error_monitor.track_success()
+        except Exception as ex:
+            self._logger.error(ex)
+            self.error_monitor.track_error(ex)
```
seanstory, line 596: Something we had in ent-search was the concept of "fatal exceptions". See: https://github.com/elastic/ent-search/blob/a1ddc884ad1543f61a91da7b9511a01090adcac5/connectors/lib/connectors/content_sources/base/extractor.rb#L132-L139

it might be worth adding that here too, for things like aio CanceledError, OOME, etc, to make sure we fail-fast when necessary.

config.yml.example
artem-shelkovnikov: Good point, will add this!

connectors/es/sink.py
artem-shelkovnikov: I'm not 100% sure it's needed. I'll double check and add if it is

connectors/source.py
artem-shelkovnikov: True, will fix!

connectors/source.py
artem-shelkovnikov: Yeah good call. OOME can't get caught, but stuff like CancelledError + having a custom fatal exception class should help

connectors/es/sink.py
artem-shelkovnikov: This logic is here to actually wait for currently running downloads to finish. Alternatively we could cancel,  but even cancellation requires joining in the end

seanstory: So should we cancel and then join?

seanstory:
Looks like this may need a rebase - I think you've pulled in other commits you may not have meant to.

connectors/source.py
artem-shelkovnikov: Okay, since we don't have code that uses this I'd address it later once we actually use this code somewhere, for now this doesn't make any difference.

I'd also like to separately add handlers for terminal exceptions - for example our download function and some other do `catch Exception` that can catch and "handle" cancelled errors in the weird way

connectors/source.py
artem-shelkovnikov: I don't really know what to do with that, we have several potential points of failure/success:

Extraction:
- File was successfully downloaded
- File was not successfully downloaded

Ingestion:
- File was successfully ingested
- File was not successfully ingested

It's possible to have all the permutations!

For example, file was not downloaded successfully, but we still choose to upload its metadata to Elasticsearch and everything worked out - is it a failure?

I'll open this up to Slack

seanstory:
Related: https://github.com/elastic/connectors/issues/1582 (maybe even "fixes"?)

connectors/config.py
```diff
@@ -102,6 +110,16 @@ def _default_config():
             "max_file_download_size": DEFAULT_MAX_FILE_SIZE,
             "job_cleanup_interval": 300,
             "log_level": "INFO",
+            "extraction": {
```
seanstory: You've only got `config.yml.example` diff for bulk, but not for extraction. Do we need both?

connectors/sync_job_runner.py
```diff
@@ -109,6 +109,10 @@ def __init__(
         self.es_config = es_config
         self.service_config = service_config
         self.sync_orchestrator = None
+        error_monitor_config = service_config.get("extraction", {}).get(
+            "error_monitor", {}
+        )
+        self.error_monitor = ErrorMonitor(**error_monitor_config)
```
seanstory: Is this used anywhere?

seanstory (approved):

connectors/sync_job_runner.py
artem-shelkovnikov: Nope, cleaned up!

connectors/config.py
artem-shelkovnikov: Nope, cleaned up!

seanstory (approved):

PR #2681, opened 2024-07-08
[PostgreSQL] Add new field id_columns in sync rules to generate id and adding support for views

parthpuri-elastic:
## Part of https://github.com/elastic/connectors-py/issues/1761

Added new field `id_columns` in advanced sync rules to ingest tables which do not have a primary key. So, now connector uses `id_columns` to generate unique ids. Also using `id_columns` users is able to fetch data of views.

#### Pre-Review Checklist

## Release Note

Added a new optional field in PostgreSQL advanced sync rules `id_columns`. The user needs to pass the name of unique fields of the table in this field, and it will be used to generate a unique ID for documents. Users are able to index data of views by using advanced sync rules and `id_columns`.


parthpuri-elastic:
buildkite test this

connectors/sources/postgresql.py
```diff
@@ -83,6 +83,7 @@ class PostgreSQLAdvancedRulesValidator(AdvancedRulesValidator):
         "properties": {
             "tables": {"type": "array", "minItems": 1},
             "query": {"type": "string", "minLength": 1},
+            "id_columns": {"type": "array", "minItems": 0},
```
artem-shelkovnikov, line 86: Why `minItems: 0`? So it's not required, right?

artem-shelkovnikov (approved):

connectors/sources/postgresql.py
parthpuri-elastic: No, we need to keep `minItems: 1` for `id_columns` as its optional field, but if the user added that, we need at least one column to generate IDs. I have updated the code. Thanks.

parthpuri-elastic:
buildkite test this

artem-shelkovnikov (approved):

praveen-elastic:
Buildkite test this

PR #2690, opened 2024-07-09
Replace task.name with task.get_name()

seanstory:
## Closes https://github.com/elastic/connectors/issues/2689

Fixes a bug where `task.name` is not a thing, but my previous usage of mocks in the spec covered up the AttributeError

#### Pre-Review Checklist

## Related Pull Requests

* first fixed in https://github.com/elastic/connectors/pull/2671/files#diff-68c31d9379d91378705cf6d47c447185e8c6ad952876c7a5fb979e3c89646148R914, but that might be a minute before merged, so putting this up separately.

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -884,7 +884,7 @@ class SyncOrchestrator:
     async def cancel(self):
         if self._sink_task_running():
             self._logger.info(
-                f"Cancling the Sink task: {self._sink_task.name}"  # pyright: ignore
+                f"Canceling the Sink task: {self._sink_task.get_name()}"  # pyright: ignore
             )
             self._sink_task.cancel()
         else:
@@ -894,7 +894,7 @@ class SyncOrchestrator:
 
         if self._extractor_task_running():
             self._logger.info(
-                f"Canceling the Extractor task: {self._extractor_task.name}"  # pyright: ignore
+                f"Canceling the Extractor task: {self._extractor_task.get_name()}"  # pyright: ignore
             )
             self._extractor_task.cancel()
         else:
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -883,9 +883,7 @@ class SyncOrchestrator:
 
     async def cancel(self):
         if self._sink_task_running():
-            self._logger.info(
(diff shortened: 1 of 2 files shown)


tests/test_sink.py
```diff
@@ -1419,11 +1419,11 @@ async def test_cancel_sync(extractor_task_done, sink_task_done, force_cancel):
     es._sink = Mock()
     es._sink.force_cancel = Mock()
 
-    es._extractor_task = Mock()
+    es._extractor_task = mock.create_autospec(asyncio.Task)
```
navarone-feekery, line 1422: TIL about `create_autospec`, neat!

artem-shelkovnikov (approved):

artem-shelkovnikov:
It was accidental in my PR - I wanted to make it a separate one, thanks for taking care of it!

connectors/es/sink.py
```diff
@@ -884,7 +884,7 @@ def _extractor_task_running(self):
     async def cancel(self):
         if self._sink_task_running():
             self._logger.info(
-                f"Cancling the Sink task: {self._sink_task.name}"  # pyright: ignore
+                f"Canceling the Sink task: {self._sink_task.get_name()}"  # pyright: ignore
```
navarone-feekery, line 887: Just curious, what was the purpose of `# pyright: ignore`? Maybe just to retain the line breaks? Because removing it doesn't anger the linter.

navarone-feekery (approved):
LGTM

tests/test_sink.py
seanstory: Same. I had to google for it, because I figured there had to be a way to use mocks but still be protected against AttributeError issues like this. I'll have to use these more going forward.

connectors/es/sink.py
seanstory: huh does it not for you? My local linter is messed up, but I'd added these originally because it was telling me that `_sink_task` might be `None`. I know that `if self._sink_task_running()` already checked that, but the linter didn't seem to.

I'll try removing them and see if CI will allow it, though.

navarone-feekery: It seemed fine for my linter, it just warned me that it would autoformat that section to be one line. But yes let's let CI decide.

artem-shelkovnikov (approved):

PR #2691, opened 2024-07-09
[Github] Add Path Filter in Advanced Rules

parth-elastic:
## Closes https://github.com/elastic/connectors-py/issues/2637

Added path filter for advanced sync rule to index supported files from a specific directory.

Limitation:
The path filter API does not support wildcard patterns. 
The GitHub API has a limitation of 1000 files per directory. [Reference](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content)

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -42,6 +42,7 @@ from connectors.utils import (
 
 WILDCARD = "*"
 BLOB = "blob"
+FILE = "file"
 GITHUB_CLOUD = "github_cloud"
 GITHUB_SERVER = "github_server"
 PERSONAL_ACCESS_TOKEN = "personal_access_token"  # noqa: S105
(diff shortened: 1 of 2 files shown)


connectors/sources/github.py
```diff
@@ -1834,44 +1846,54 @@ async def _fetch_last_commit_timestamp(self, repo_name, path):
         )
         return commit["commit"]["committer"]["date"]
 
-    async def _fetch_files(self, repo_name, default_branch):
+    async def _fetch_files(self, repo_name, path_branch, path=False):
```
artem-shelkovnikov, line 1849: Can we break this method into two: `fetch_files` with old signature/logic and `fetch_path` with new logic to fetch filtering by path?

connectors/sources/github.py
```diff
@@ -1834,6 +1846,31 @@ async def _fetch_last_commit_timestamp(self, repo_name, path):
...
+                    "extension": file_extension,
+                }
+            )
+
+            document = self.adapt_gh_doc_to_es_doc(
+                github_document=repo_object, schema=schema
+            )
+
+            document["_id"] = f"{repo_name}/{repo_object['path']}"
+            yield document, repo_object
```
artem-shelkovnikov, line 1872: You only return once from this method, can you replace `yield` with `return`?

connectors/sources/github.py
```diff
@@ -1847,31 +1884,37 @@ async def _fetch_files(self, repo_name, default_branch):
...
-                        document["_id"] = f"{repo_name}/{repo_object['path']}"
-                        yield document, repo_object
+    async def _fetch_files_by_path(self, repo_name, path):
+        self._logger.info(f"Fetching files from repo: '{repo_name}' (path: '{path}')")
+        try:
+            files = await self.github_client.get_github_item(
+                resource=self.github_client.endpoints["PATH"].format(
+                    repo_name=repo_name, path=path
+                )
+            )
```
artem-shelkovnikov, line 1908: Can you do it without fetching all files metadata into memory?

artem-shelkovnikov (approved):

PR #2695, opened 2024-07-09
log in utc

seanstory:
I noticed log lines like:
```
[FMWK][14:25:44][DEBUG] [job_scheduling_service] Scheduler woke up at 2024-07-09 19:25:44.085357+00:00.
```

notice that the logging framework timestamp is `14:25:44` (my local time) but the connectors logic understands that the time should be reasoned about in UTC (`19:25:44`). 

This discrepancy makes it hard to get scheduling right. If I just look at the _majority_ of timestamps that I see in the logs, I'd assume that the scheduling logic should use my local time, and not UTC time.

This PR fixes the issue by overwriting `logging.Formatter.format_time()` for our Formatter class.

It also simplifies the logic for changing the style depending on log level so that function overwrites work. 

#### Pre-Review Checklist

## Release Note

Connector logs now use UTC timestamps, instead of machine-local timestamps. This only impacts logging output.

### DIFF
--- a/connectors/logger.py
+++ b/connectors/logger.py
@@ -29,15 +29,23 @@ class ColorFormatter(logging.Formatter):
     RED = "\x1b[31;20m"
     BOLD_RED = "\x1b[31;1m"
     RESET = "\x1b[0m"
+    COLORS = {
+        logging.DEBUG: GREY,
+        logging.INFO: GREEN,
+        logging.WARNING: YELLOW,
+        logging.ERROR: RED,
+        logging.CRITICAL: BOLD_RED,
+    }
 
     DATE_FMT = "%H:%M:%S"
 
     def __init__(self, prefix):
         self.custom_format = "[" + prefix + "][%(asctime)s][%(levelname)s] %(message)s"
         super().__init__(datefmt=self.DATE_FMT)
+        self.local_tz = tzlocal()
 
     def converter(self, timestamp):
-        dt = datetime.fromtimestamp(timestamp, tzlocal())
+        dt = datetime.fromtimestamp(timestamp, self.local_tz)
         return dt.astimezone(timezone.utc)
 
     # override logging.Formatter to use an aware datetime object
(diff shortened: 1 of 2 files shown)


connectors/logger.py
```diff
@@ -32,41 +35,38 @@ class ColorFormatter(logging.Formatter):
...
-    def critical_formatter(self):
-        return logging.Formatter(
-            fmt=self.BOLD_RED + self.custom_format + self.RESET, datefmt=self.DATE_FMT
-        )
+        super().__init__(datefmt=self.DATE_FMT)
+
+    def converter(self, timestamp):
+        tzinfo = tzlocal.get_localzone()
+        dt = datetime.datetime.fromtimestamp(timestamp, tzinfo)
+        return dt.astimezone(pytz.UTC)
```
jedrazb, line 43: I think we can avoid adding `pytz` to the mix, `datetime` lib supports all we need, see example:
https://github.com/elastic/connectors/blob/644c4e018043e24e529b8dadce7665e7da47f5a1/connectors/utils.py#L97-L102

connectors/logger.py
```diff
@@ -32,41 +35,38 @@ class ColorFormatter(logging.Formatter):
...
+            try:
+                s = dt.isoformat(timespec="milliseconds")
+            except TypeError:
+                s = dt.isoformat()
+        return s
 
     def format(self, record):  # noqa: A003
-        formatter = getattr(self, f"{record.levelname.lower()}_formatter")
-        return formatter.format(record)
+        # record.asctime = self.formatTime(record, self.DATE_FMT)
```
jedrazb, line 58: Should this line be removed?

connectors/logger.py
```diff
@@ -32,41 +35,38 @@ class ColorFormatter(logging.Formatter):
...
+        if record.levelno == logging.DEBUG:
```
jedrazb, line 59: Nice! Less is more! ✨

connectors/logger.py
seanstory: thanks! done.

jedrazb (approved):

PR #2697, opened 2024-07-10
Log error when sync ends

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/2645

Once syncs end with an error we just say:

```
Sync ended with status error -- created: 0 | updated: 4000 | deleted: 0 (took 423 seconds)
```

Actual error is not logged anywhere - you need to go to UI or connectors index to figure out what's the error. This is problematic when doing support of connectors via reading logs.

This PR adds a log line for situation when connector terminates with status error, this line contains the error message that will be shown in UI.

#### Pre-Review Checklist


artem-shelkovnikov:
NVM there's no need

PR #2703, opened 2024-07-12
Bump VERSION to 8.14.3.1

artem-shelkovnikov:
## Part of https://github.com/elastic/search-team/issues/7867

Bumping connectors VERSION to next build version. Don't mind the branch name.


navarone-feekery (dismissed):

navarone-feekery:
> Don't mind the branch name.
@artem-shelkovnikov Wait wait, the branch name and the change are `8.14.4.0` but the PR name is `8.15.3.1`. Just checking you have the right version here?

artem-shelkovnikov:
> @artem-shelkovnikov Wait wait, the branch name and the change are 8.14.4.0 but the PR name is 8.14.3.1. Just checking you have the right version here?

I think so, let's just double check this together:

I'm following Release Issue flow here: https://github.com/elastic/search-team/issues/7867

We're on patch release 8.14.2 -> 8.14.3. By this step I've released docker image 8.14.3.0.

We have only one step to bump the VERSION of connectors, so at some times we bump the PATCH version, in some we bump BUILD version.

We don't have planned 8.14.4, so it kinda makes sense to not bump the VERSION to 8.14.4 because of this. We _might_ however release new docker version.

navarone-feekery:
I'd also expect the next version to be `8.14.4.0` if we're following the release stack logic. I think we're opening ourselves up to more confusion if we have a different rule based on whether there's another PATCH stack release planned.

But ultimately I think either is fine, my confusion was more that the PR name is different from the actual change.

PR name: `8.14.3.1`
Branch name: `8.14.4.0`
Actual change: `8.14.4.0`

artem-shelkovnikov:
I actually did not push my change, I was supposed to have VERSION changed to 8.14.3.1 too

PR #2718, opened 2024-07-23
[Dropbox] Fix fetching documents from team shared folders

moxarth-elastic:
## Closes https://github.com/elastic/connectors/issues/2704

This PR now fetches all the files present at the team shared folders. 
Earlier we pass folder_id for the header - `Dropbox-API-Path-Root`, which was causing this issue of missing root folder content. Now, we're passing `root_namespace_id` for the header which includes all team shared folders content.

#### Pre-Review Checklist

## Release Note

Dropbox connector now fetches the files from the team shared folders.


connectors/sources/dropbox.py
```diff
@@ -844,11 +854,41 @@ async def close(self):
...
+    async def set_user_info(self):
+        try:
             endpoint = EndpointName.AUTHENTICATED_ADMIN.value
-        await self.dropbox_client.ping(endpoint=endpoint)
-        self._logger.info("Successfully connected to Dropbox")
+            response = await self.dropbox_client.ping(endpoint=endpoint)
+        except ClientResponseError as exception:
+            if (
+                str(exception.request_info.url)
+                == "https://api.dropboxapi.com/2/team/token/get_authenticated_admin"
```
navarone-feekery, line 879: I feel like this URL should be a const

connectors/sources/dropbox.py
```diff
@@ -1043,9 +1085,7 @@ async def get_account_details(self):
 
     async def get_permission_list(self, item_type, item, account_id):
         if item_type == FOLDER:
-            shared_folder_id = item.get("shared_folder_id") or item.get(
-                "parent_shared_folder_id"
-            )
```
navarone-feekery, line 1048: Why was this removed?

connectors/sources/dropbox.py
```diff
@@ -1146,20 +1192,19 @@ async def add_document_to_list(self, func, account_id, folder_id, is_shared=Fals
...
-                func=self._fetch_shared_files,
-                account_id=account_id,
-                folder_id=folder_id,
-                is_shared=True,
-            ):
-                yield mapped_document
+        async for mapped_document in self.add_document_to_list(
+            func=self._fetch_files_folders,
+            account_id=account_id,
+            folder_id=self.dropbox_client.member_id,
```
navarone-feekery, line 1198: Why are we using the `self.dropbox_client.member_id` here instead of the original `folder_id`?

navarone-feekery (commented):
I've left a few questions because I'm quite unfamiliar with dropbox API. It'd be a great help if you expand the PR description to explain why you've made many of these changes.

connectors/sources/dropbox.py
moxarth-elastic: Seems like these changes does not change anything, i tried to revert the changes and tested the connector; it also works properly. So, reverting this changes to reduce confusion :)

moxarth-elastic:
> It'd be a great help if you expand the PR description to explain why you've made many of these changes.

I've mentioned the root cause of the issue in the description; and some of the changes are of code auto-formatting (make autoformat).

connectors/sources/dropbox.py
```diff
@@ -844,11 +857,35 @@ async def close(self):
...
             endpoint = EndpointName.AUTHENTICATED_ADMIN.value
-        await self.dropbox_client.ping(endpoint=endpoint)
-        self._logger.info("Successfully connected to Dropbox")
+            await self.dropbox_client.ping(endpoint=endpoint)
+            self._logger.info("Successfully connected to Dropbox")
+        except ClientResponseError as exception:
+            if str(exception.request_info.url) == AUTHENTICATED_ADMIN_URL:
+                endpoint = EndpointName.PING.value
+                await self.dropbox_client.ping(endpoint=endpoint)
+                self._logger.info("Successfully connected to Dropbox")
```
navarone-feekery, line 868: Why do we attempt to ping a different endpoint if the first one fails?

Also, `endpoint` has already been set to `endpoint = EndpointName.AUTHENTICATED_ADMIN.value` on L861, but if there is a client response error the code checks `if str(exception.request_info.url) == AUTHENTICATED_ADMIN_URL:`. Can the value from `exception.request_info.url` ever _not_ be the authenticated admin URL used in the request?

tests/sources/test_dropbox.py
```diff
@@ -1215,9 +1220,14 @@ async def test_advanced_rules_validation_with_invalid_repos(
         [JSONAsyncMock(MOCK_SEARCH_FILE_3, 200)],
     ),
 )
+@patch.object(
+    DropboxClient,
+    "ping",
+    return_value=JSONAsyncMock(MOCK_AUTHENTICATED_ADMIN, 200),
```
navarone-feekery, line 1226: Can you add tests covering the following for the methods `ping` and `set_user_info`:

1. When the authenticated admin ping returns a `ClientResponseError`
2. When the authenticated admin ping returns a different error (e.g. `TimeoutError`)
3. When the authenticated admin ping returns a `ClientResponseError` and the `exception.request_info.url` was an unexpected value

connectors/sources/dropbox.py
```diff
@@ -844,11 +857,35 @@ async def close(self):
...
+    async def set_user_info(self):
+        try:
+            endpoint = EndpointName.AUTHENTICATED_ADMIN.value
+            response = await self.dropbox_client.ping(endpoint=endpoint)
+        except ClientResponseError as exception:
+            if str(exception.request_info.url) == AUTHENTICATED_ADMIN_URL:
+                endpoint = EndpointName.PING.value
+                response = await self.dropbox_client.ping(endpoint=endpoint)
+            else:
+                raise
```
navarone-feekery, line 881: Same as previous question.

navarone-feekery: Also, if the first request returns a `ClientResponseError`, `response` will become the result from the ping endpoint instead of the authenticated admin endpoint. Will that not cause issues when trying to extract values for `self.dropbox_client.member_id` and `self.dropbox_client.root_namespace_id`?

connectors/sources/dropbox.py
```diff
@@ -844,11 +857,35 @@ async def close(self):
...
+        _json = await response.json()
+        self.dropbox_client.member_id = _json.get("admin_profile", {}).get(
+            "team_member_id"
+        )
+        self.dropbox_client.root_namespace_id = _json.get("root_info", {}).get(
+            "root_namespace_id"
+        ) or _json.get("admin_profile", {}).get("root_folder_id")
```
navarone-feekery, line 888: The following two dict calls can raise an raise an error:

1. `_json.get("admin_profile", {}).get("team_member_id")` if `{ admin_profile: None }`
2. `_json.get("admin_profile", {}).get("root_folder_id")` if `{ admin_profile: None }`
3. `_json.get("root_info", {}).get("root_namespace_id")` if `{ root_info: None }`

Can you null-safe these?

Also, if the json has `{ admin_profile: None }`, does that effectively mean Dropbox cannot sync documents? If so I think we need the following changes, otherwise the stack trace will not provide enough information to a user about what went wrong.

1. A detailed log explaining what is wrong with the response format
2. Raise a custom error (e.g. `DropboxResponseMalformed` error or something)

navarone-feekery (commented):
Thanks for the extra summary on the PR description!
I've done another review now that I have more context. I can't test this against a dropbox instance myself, so can you also provide a log output example?

connectors/sources/dropbox.py
moxarth-elastic: https://github.com/elastic/connectors/blob/76ef0c1618e1d67d92c1bd066ea820871fc30008/connectors/sources/dropbox.py#L846-L851

The previous ping method was checking if the DLS is enabled; we use `EndpointName.AUTHENTICATED_ADMIN.value` endpoint else `EndpointName.PING.value` endpoint. 

So, the problem arrives when the business account is used and DLS is not enabled. FYI, to fetch content from business account, the token should have team permission and so it will not work for the [PING](https://www.dropbox.com/developers/documentation/http/documentation#users-get_current_account) endpoint ; even though the connection can be established, the connector throws ClientRequestError for `EndpointName.PING.value` endpoint in that case. 

To fix this, we have two options:
1. Add an RCF for the Dropbox account type; either Basic or Business 
2. (current approach) We can check the `EndpointName.AUTHENTICATED_ADMIN.value` endpoint, if we get ClientResponseError then we move forward with the another endpoint - `PING`.

navarone-feekery: @moxarth-elastic thanks for the detailed explanation. Sorry, I didn't realise from the diff that this was how it already was, I thought the addition of two different ping endpoints was from this PR.

navarone-feekery (approved):
LGTM

PR #2723, opened 2024-07-29
Change docker-compose network in stack scripts to `elastic`

navarone-feekery:
## Summary

The stack scripts `script/stack/run-stack.sh` was using a generated network called `connectors_stack_net` (which would finalize as `docker_connectors_stack_net`) whenever the script was run.
In [Elasticsearch](https://github.com/elastic/elasticsearch) and [Connectors](https://github.com/elastic/connectors/blob/34efbe579e65eba89f70b3e3f1228e6ade0cb26f/docs/DOCKER.md#5-run-the-docker-image) documentation, we always specify to use the docker network `elastic`. This change updates the stack scripts docker compose to use the `elastic` network.

Because users may be following manual steps from other documentation, we can't rely on creating the network in the docker-compose file. Instead, we must assume the network was already created (the flag `external: true`). Without this flag, the docker compose file will crash if the network was already manually created.
To stop users from _needing_ to manually create the `elastic` network if they're using the stack scripts exclusively, I added a step that checks if the network exists and creates it if it's missing. This network can then be handled as "external" in the docker compose file.

## Changes

1. Manually create `elastic` network when running stack scripts (if it doesn't exist)
2. Use the "externally" created `elastic` network in the docker-compose file

#### Pre-Review Checklist

### DIFF
--- a/scripts/stack/docker/docker-compose.yml
+++ b/scripts/stack/docker/docker-compose.yml
@@ -1,7 +1,9 @@
 version: "3"
 
 networks:
-  connectors_stack_net:
+  elastic:
+    name: 'elastic'
+    external: true
 
 services:
   elasticsearch:
@@ -19,7 +21,7 @@ services:
         soft: -1
         hard: -1
     networks:
-      - connectors_stack_net
+      - elastic
     ports:
       - 9200:9200
     volumes:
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov (approved):
🔥

PR #2727, opened 2024-07-29
Make sure event loops are closed

seanstory:
One more try to make this test stop flaking out. Hopefully avoiding https://github.com/elastic/connectors/pull/2726


seanstory:
Nope still flakey. 😡

artem-shelkovnikov:
We definitely need to close the loops though

PR #2734, opened 2024-08-05
Make connectors work with python 3.10 and 3.11

artem-shelkovnikov:
## Part of https://github.com/elastic/connectors/issues/2672

This PR makes connectors work with 3.10 and 3.11 version of Python. 3.12 is not supported yet because of some deprecations for `datetime` package that we need to handle: https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated. Once it's handled, adding 3.12 to matrix is gonna be trivial.

CI image was updated to make it work, here's the CI image script: https://github.com/elastic/ci-agent-images/blob/main/vm-images/enterprise-search/scripts/connectors-python/install-deps.sh

Link to nightlies using this branch: https://buildkite.com/elastic/connectors-nightly/builds/2402

#### Pre-Review Checklist

## Release Note

Now it's possible to run connectors with Python 3.11 as well as Python 3.10


.buildkite/pipeline.yml
```diff
@@ -9,22 +9,33 @@ env:
...
+      CONNECTOR: "dir"
+      DATA_SIZE: "small"
     timeout_in_minutes: 45
     agents:
       machineType: "n1-standard-8"
 
   - label: ":sweating: Checking for changes in connectors"
     plugins:
-      chronotc/monorepo-diff#v2.4.0:
+      monorepo-diff#v1.0.1:
```
artem-shelkovnikov, line 38: Changed to official plugin

.buildkite/run_functional_test.sh
```diff
@@ -10,21 +10,22 @@ if [ "$MACHINE_TYPE" != "x86_64" ] && [ -v SKIP_AARCH64 ]; then
   exit
 fi
 
+source ~/.bash_profile
+
+pyenv global $PYTHON_VERSION
+echo "Python version:"
+pyenv global
```
artem-shelkovnikov, line 17: This should help figure out which python version was used for nightly

.buildkite/run_functional_test.sh
```diff
@@ -10,21 +10,22 @@ if [ "$MACHINE_TYPE" != "x86_64" ] && [ -v SKIP_AARCH64 ]; then
...
 
 BASEDIR=$(realpath $(dirname $0))
 ROOT=$(realpath $BASEDIR/../)
 
-# TODO to be moved in the image at https://github.com/elastic/ci-agent-images/blob/main/vm-images/enterprise-search/scripts/connectors-python/install-deps.sh#L6
-sudo apt-get -y install liblz4-dev libunwind-dev
```
artem-shelkovnikov, line 18: This was inlined into the VM image here: https://github.com/elastic/ci-agent-images/pull/828/files#diff-6942f445af31cff91a64513380d85c7ae18e5973d4d91ec6171cc0d9e831b7b0R11-R12

.buildkite/run_functional_test.sh
```diff
@@ -42,4 +43,4 @@ if [ -v BUILDKITE ]; then
   sudo sysctl -w vm.max_map_count=262144
 fi
 
-PERF8=yes NAME=$1 DATA_SIZE=$DATA_SIZE make ftest
+PERF8=yes NAME=$CONNECTOR DATA_SIZE=$DATA_SIZE make ftest
```
artem-shelkovnikov, line 46: Now all that is passed as environmental variables instead of script args

Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM python:3.10
+FROM python:3.12-slim-bookworm
+# RUN apt update && apt install make
+RUN apt update && apt upgrade && apt install make -y
```
artem-shelkovnikov, line 3: Slim image has no make

Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: test lint autoformat run ftest install release docker-build docker-run docker-push
 
-PYTHON=python3.10
+PYTHON=python
```
artem-shelkovnikov, line 3: Not sure whether having it `python` or `python3` here is better, open to discussion here

requirements/framework.txt
```diff
@@ -5,7 +5,8 @@ httpx==0.27.0
 httpx-ntlm==1.4.0
```
artem-shelkovnikov, line 5: A lot of these updates are to make it actually build on all systems

tests/sources/fixtures/oracle/fixture.py
```diff
@@ -66,17 +65,13 @@ def inject_lines(table, cursor, lines):
 async def load():
     """Generate tables and loads table data in the oracle server."""
     """N tables of RECORD_COUNT rows each"""
-    connection = oracledb.connect(
-        user="system", password=PASSWORD, dsn=DSN, encoding=ENCODING
-    )
+    connection = oracledb.connect(user="system", password=PASSWORD, dsn=DSN)
```
artem-shelkovnikov, line 68: Encoding was deprecated: https://python-oracledb.readthedocs.io/en/latest/api_manual/deprecations.html

seanstory:
Holding off on review while the build is red, but holler when this is ready 🤞

artem-shelkovnikov:
> Holding off on review while the build is red, but holler when this is ready 🤞

It's just a random blip from CI not being able to start Elasticsearch, feel free to review :)

.buildkite/nightly_steps.yml
```diff
@@ -2,133 +2,290 @@ env:
...
+    env:
+      PYTHON_VERSION: "{{ matrix }}"
+      CONNECTOR: "mysql"
+      DATA_SIZE: "small"
     artifact_paths:
       - "perf8-report-*/**/*"
-  - label: "🔨 Network Drive"
-    command:
-      - ".buildkite/run_nigthly.sh network_drive"
+    matrix:
```
seanstory, line 13: TIL about "matrix"
would it be worth it to just redo this whole file as a single step, using `matrix.setup` to define python version, connector service type, and data size? Looking at https://buildkite.com/docs/pipelines/build-matrix

.buildkite/pipeline.yml
```diff
@@ -9,22 +9,33 @@ env:
 
 steps:
   - label: ":face_with_peeking_eye: Lint"
-    command: "make lint"
+    command: ".buildkite/run_linter.sh"
+    env:
+      PYTHON_VERSION: "3.10" # We don't really care about linter on multiple versions for now
```
seanstory, line 14:
```suggestion
      PYTHON_VERSION: "3.11" # We don't really care about linter on multiple versions for now
```
I'd vote we actually move to whatever the latest is that we support.

.buildkite/pipeline.yml
```diff
@@ -9,22 +9,33 @@ env:
...
     timeout_in_minutes: 5
+    matrix:
+      - "3.10"
+      - "3.11"
 
   - label: ":shipit: Smoke test"
-    command: "make ftest NAME=dir DATA_SIZE=small"
+    command: ".buildkite/run_functional_test.sh"
+    env:
+      PYTHON_VERSION: "3.10" # TODO: can actually do a matrix too
```
seanstory, line 29:
```suggestion
      PYTHON_VERSION: "3.11" # TODO: can actually do a matrix too
```

seanstory: Though, I might be interested in doing the full matrix for the smoketest, since they'll run in parallel, right?

.buildkite/pipeline.yml
```diff
@@ -36,8 +47,12 @@ steps:
             - "requirements/**"
             config:
               label: "🔨 MySQL"
+              env:
+                - PYTHON_VERSION=3.10
```
seanstory, line 51: I assume the plugin config doesn't automagically work with `matrix` and that's why this is all staying single version for now?

.buildkite/run_functional_test.sh
seanstory: no objection, but that should be obvious from the step name, right?

Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM python:3.10
+FROM python:3.12-slim-bookworm
+# RUN apt update && apt install make
```
seanstory, line 2: lingering comment?

Dockerfile
```diff
@@ -1,4 +1,7 @@
...
+RUN apt update && apt upgrade && apt install make -y
 COPY . /app
 WORKDIR /app
 RUN make clean install
+ENTRYPOINT []
```
seanstory, line 7: why an empty entrypoint?

Makefile
seanstory: Do we even need this anymore? Now we should just be providing guidance to make sure that you you've installed a compatible system python version, or set up `pyenv` to enforce a compatible version. Right?

seanstory (commented):
few comments/questions, but looks pretty straightforward, honestly. Nice work!

I kicked off a nightly, just for a double check: https://buildkite.com/elastic/connectors-nightly/builds/2404

Dockerfile.wolfi
```diff
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/wolfi/python:3.10-dev@sha256:2b7e585c59b5f7b5800de99aa2dc0e576bc1f7358e0d1acfcc14dcd90bb3925f
+FROM docker.elastic.co/wolfi/python:3.12-dev
```
seanstory, line 1:
```suggestion
FROM docker.elastic.co/wolfi/python:3.11-dev
```

Dockerfile.ftest.wolfi
```diff
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/wolfi/python:3.10-dev@sha256:2b7e585c59b5f7b5800de99aa2dc0e576bc1f7358e0d1acfcc14dcd90bb3925f
+FROM docker.elastic.co/wolfi/python:3.12-dev
```
seanstory, line 1:
```suggestion
FROM docker.elastic.co/wolfi/python:3.11-dev
```

Dockerfile.ftest
```diff
@@ -1,5 +1,7 @@
-FROM python:3.10
-COPY . /connectors
-WORKDIR /connectors
+FROM python:3.12-slim-bookworm
```
seanstory, line 1:
```suggestion
FROM python:3.11-slim-bookworm
```

Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM python:3.10
+FROM python:3.12-slim-bookworm
```
seanstory, line 1:
```suggestion
FROM python:3.11-slim-bookworm
```

.buildkite/pipeline.yml
seanstory: And if it is gonna be one version, should it be 3.11, to align with the version we actually publish from our dockerfiles?

seanstory (commented):
also, realized that 3.12 had slipped through in a few places

.buildkite/nightly_steps.yml
artem-shelkovnikov: Yup good idea!

.buildkite/run_functional_test.sh
artem-shelkovnikov: It _should_, but if something broke in CI/image/elsewhere, it can bring surprises

.buildkite/pipeline.yml
artem-shelkovnikov: Official plugin docs don't mention matrix support, unfortunately, so I skipped it here. Updating to 3.11 makes sense!

Dockerfile
artem-shelkovnikov: Good catch!

Dockerfile
artem-shelkovnikov: Actually just for consistency - `.wolfi` file has empty entry point, so I just added it here. In practice it does not change anything

.buildkite/nightly_steps.yml
seanstory: Ah, service type being > 20? Boo. 
Well thanks for looking. :sigh:

artem-shelkovnikov: We can actually make it work too, just breaking into matrices by 20 service types x 2 versions, or 3 versions x 16 service types too, but can do later :)

Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: test lint autoformat run ftest install release docker-build docker-run docker-push
 
-PYTHON=python3.10
```
efegurkan, line 3: this change requires your setup to have an alias for python, in my machine I didn't had that and to make it work I have to do `PYTHON=python3 make install`

Is this an issue ? If no do we want to mention it in somewhere?

artem-shelkovnikov: Opened discussion in https://elastic.slack.com/archives/C01795T48LQ/p1723648672552739?thread_ts=1723465470.280719&cid=C01795T48LQ

PR #2749, opened 2024-08-14
add dra support

seanstory:
WIP

#### Pre-Review Checklist

### DIFF
--- a/.buildkite/publish/dra/init_dra_publishing.sh
+++ b/.buildkite/publish/dra/init_dra_publishing.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+########
+# publishes a DRA artifact
+########
+
+set -exu
+set -o pipefail
+
+# Load our common environment variables for publishing
+export CURDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source $CURDIR/../publish-common.sh
+
+# Mock out creating an artifact
+echo "Hello, world"
+echo "The version is: ${VERSION}"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,3 +1,10 @@
+definitions:
+  steps:
+    - step: &docker-login
+        plugins:
+          - elastic/vault-docker-login#v0.5.2:
+              secret_path: 'kv/ci-shared/serverless/external-docker-registry'
+
 agents:
   provider: "gcp"
   machineType: "n1-standard-2"
@@ -9,12 +16,14 @@ env:
 
 steps:
   - label: ":face_with_peeking_eye: Lint"
+    key: "lint"
     command: ".buildkite/run_linter.sh"
     env:
       PYTHON_VERSION: "3.11" # We don't really care about linter on multiple versions for now
     timeout_in_minutes: 5
 
   - label: ":pytest: Test with Python {{ matrix }}"
+    key: "unit_tests"
     command: ".buildkite/run_tests.sh"
     env:
       PYTHON_VERSION: "{{ matrix }}"
@@ -24,6 +33,7 @@ steps:
       - "3.11"
 
   - label: ":shipit: Smoke test"
+    key: "smoke_test"
     command: ".buildkite/run_functional_test.sh"
     env:
       PYTHON_VERSION: "3.11" # TODO: can actually do a matrix too
@@ -34,6 +44,7 @@ steps:
       machineType: "n1-standard-8"
 
   - label: ":sweating: Checking for changes in connectors"
+    key: "relevant_ftests"
     plugins:
       monorepo-diff#v1.0.1:
         diff: ".buildkite/diff ${BUILDKITE_COMMIT}"
@@ -482,3 +493,23 @@ steps:
                 - ".buildkite/run_functional_test.sh"
               artifact_paths:
                 - "perf8-report-*/**/*"
+
+  # ----
+  # DRA publishing
+  # ----
+  - group: ":truck: DRA"
+    key: "mbp_dra_group"
+    if: "(build.branch == \"main\")" # Add new maintenance branches here
+    depends_on:
+      - "lint"
+      - "unit_tests"
+      - "smoke_test"
+      - "relevant_ftests"
+    steps:
+      - label: ":truck: Publish DRA Items"
+        agents:
+          provider: gcp
+        key: "publish-dra-artifacts"
+        timeout_in_minutes: 45
+        command: ".buildkite/publish/dra/init_dra_publishing.sh"
+        <<: *docker-login
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -499,7 +499,7 @@ steps:
   # ----
   - group: ":truck: DRA"
     key: "mbp_dra_group"
-    if: "(build.branch == \"main\")" # Add new maintenance branches here
+    if: "(build.branch == \"main\" || build.branch == \"seanstory/add-DRA-support\")" # Add new maintenance branches here
     depends_on:
       - "lint"
       - "unit_tests"


seanstory:
This is ready for review!

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+  echo "-------- Publishing SNAPSHOT DRA Artifacts"
+  cp $RELEASE_DIR/dist/elasticsearch_connectors-${VERSION}.zip $RELEASE_DIR/dist/dra-artifacts/connectors-${VERSION}-SNAPSHOT.zip
+  setDraVaultCredentials
+  export WORKFLOW="snapshot"
+
+  # TODO - when ready, remove the gate and ungate test vars
+  if [[ "${DRA_PUBLISH_GATE:-}" == "" || "${DRA_UNGATE_TEST:-}" == "true" ]]; then
+    source "${PROJECT_ROOT}/.buildkite/publish/dra/publish-daily-release-artifact.sh"
+  else
+    echo "DRA_PUBLISH_GATE is locked. This is where we would call 'source ${PROJECT_ROOT}/.buildkite/publish/dra/publish-daily-release-artifact.sh'"
```
artem-shelkovnikov, line 128: What does the gate actually mean? Can we change the message to be actionable (e.g. pass DRA_PUBLISH_GATE to unlock the publish or something)?

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+  dependencyReportName="dependencies-${VERSION}.csv";
+  echo "-------- Generating STAGING dependency report: ${dependencyReportName}"
+  generateDependencyReport $DEPENDENCIES_REPORTS_DIR/$dependencyReportName
+
+  echo "-------- Publishing STAGING DRA Artifacts"
+  cp $RELEASE_DIR/dist/elasticsearch_connectors-${VERSION}.zip $RELEASE_DIR/dist/dra-artifacts/connectors-${VERSION}.zip
+  setDraVaultCredentials
+  export WORKFLOW="staging"
+
+  # TODO - when ready, remove the gate var
```
artem-shelkovnikov, line 145: When is the ready moment? :)

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+  if [[ "${DRA_PUBLISH_GATE:-}" == "" ]]; then
+    source "${PROJECT_ROOT}/.buildkite/publish/dra/publish-daily-release-artifact.sh"
+  else
+    echo "DRA_PUBLISH_GATE is locked. This is where we would call 'source ${PROJECT_ROOT}/.buildkite/publish/dra/publish-daily-release-artifact.sh'"
+  fi
+  unsetDraVaultCredentials
+  rm -rf "${DEPENDENCIES_REPORTS_DIR}/*"
```
artem-shelkovnikov, line 152: You can probably always run it after script exits like [here](https://stackoverflow.com/questions/15656492/writing-try-catch-finally-in-shell)?

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
artem-shelkovnikov, line 1: Are we allowed to run such scripts in Python?

Bash is sooo hard to read/write :/

artem-shelkovnikov (commented):
Looks good, left a couple questions :)

Makefile
```diff
@@ -85,3 +85,6 @@ docker-run:
 
 docker-push:
 	docker push $(DOCKER_IMAGE_NAME):$(VERSION)-SNAPSHOT
+
+sdist: bin/python
```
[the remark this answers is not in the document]
jedrazb: source distribution in python terminology, https://docs.python.org/3.10/distutils/sourcedist.html

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+    exit 2
+  fi
+
+  export VAULT_SECRET_ID="$(echo $DRA_CREDS | jq -r '.secret_id')"
+  if [[ "${VAULT_SECRET_ID:-}" == "" ]]; then
+    echo "Could not get VAULT_ADDR from DRA credentials"
+    exit 2
+  fi
+
+  unset DRA_CREDS
```
jedrazb, line 91: should we use `trap` to unset this if process exits or is interrupted?

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+  setDraVaultCredentials
+  export WORKFLOW="staging"
+
+  # TODO - when ready, remove the gate var
+  if [[ "${DRA_PUBLISH_GATE:-}" == "" ]]; then
+    source "${PROJECT_ROOT}/.buildkite/publish/dra/publish-daily-release-artifact.sh"
+  else
+    echo "DRA_PUBLISH_GATE is locked. This is where we would call 'source ${PROJECT_ROOT}/.buildkite/publish/dra/publish-daily-release-artifact.sh'"
+  fi
+  unsetDraVaultCredentials
```
jedrazb, line 151: we are unsetting DraVaultCreds in multiple places 

should we instead use sth like trap? https://www.linuxjournal.com/content/bash-trap-command

```
function cleanup()
{
    # ...
}

trap cleanup EXIT
```

And just put it at begging of bash script?

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+
+########
+# publishes a DRA artifact
+########
+
+set -exu
+set -o pipefail
+
+# Load our common environment variables for publishing
+export DRA_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
jedrazb, line 11: sth like that should work better but please test!! 🙏 
```suggestion
export DRA_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
```

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+  # and reset our original vault address
+  export VAULT_ADDR="${ORIGINAL_VAULT_ADDR:-}"
+
+  set -x
+}
+
+# function to generate dependency report.
+function generateDependencyReport() {
+  echo "name,version,url,license,sourceURL" > $1 # TODO, do this right
+}
```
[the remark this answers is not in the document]
seanstory: That's implemented in https://github.com/elastic/connectors/pull/2757. I made a second PR to try to keep the diff more manageable.

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -0,0 +1,153 @@
...
+
+# Sanity check in the logs to list the downloaded artifacts
+chmod -R a+rw "${RELEASE_DIR}/dist"
+echo "---- listing artifact directory: ${RELEASE_DIR}/dist:"
+ls -al $RELEASE_DIR/dist
+
+# ensure we have a directory and permissions for our dependency report to be created
+# this path is set in the actual config file in
+# https://github.com/elastic/infra/tree/master/cd/release/release-manager/project-configs
+export DEPENDENCIES_REPORTS_DIR="${RELEASE_DIR}/dist/cd/output" # TODO, change this path?
```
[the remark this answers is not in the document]
seanstory: This path is controlled by https://github.com/elastic/infra/blob/073fc27ba44f932c32ec35c2cb523a24cd2f1723/cd/release/release-manager/project-configs/master/connectors.gradle#L21. So in order to change it in our repo, I have to first change it there. Figured that it was better to merge it as it is now (working) and negotiate separately on paths that make more sense, as opposed to ones I copied from [the setup guide](https://docs.elastic.dev/release/stack/dra-getting-started#-dsl---the-contract) before I understood what their significance was 🙈

.buildkite/publish/dra/publish-daily-release-artifact.sh
```diff
@@ -0,0 +1,88 @@
...
+fi
+
+# Branch Name. Required for the cli collect artifact publish
+if [[ "${BRANCH_NAME:-}" == "" ]]; then
+  echo "ERROR: BRANCH_NAME required!"
+  exit 2
+fi
+
+#---------------------------------------------------------------------------------------------------
+echo "Made it to 'docker run'" # TODO, replace this with actual docker run
```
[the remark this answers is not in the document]
seanstory: yep, good call. 👀

.buildkite/publish/dra/publish-daily-release-artifact.sh
```diff
@@ -0,0 +1,88 @@
...
+echo "Conents of 'src':"
+ls -lah ${RELEASE_DIR}/dist
+BRANCH_NAME="main" #TODO, hacking around dev branch iterations
```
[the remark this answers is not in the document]
seanstory: Ah, missed this TODO, thanks. `BRANCH_NAME` is set in the `init_dra_publish.sh` to be the branch currently being used, but I hardcoded it here to `main` so that I could publish images from _this_ branch during development. I will remove this line. 👀

.buildkite/publish/dra/publish-daily-release-artifact.sh
```diff
@@ -0,0 +1,88 @@
...
+# Branch Name. Required for the cli collect artifact publish
+if [[ "${BRANCH_NAME:-}" == "" ]]; then
+  echo "ERROR: BRANCH_NAME required!"
+  exit 2
+fi
+
+#---------------------------------------------------------------------------------------------------
+echo "Made it to 'docker run'" # TODO, replace this with actual docker run
+echo "Conents of 'src':"
+ls -lah ${RELEASE_DIR}/dist
```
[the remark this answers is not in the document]
seanstory: this I think we should leave, since the `ls` doesn't hurt anything, and it was useful for debugging what was/wasn't there when the DRA command was saying files were missing. WDYT?

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: good question. Probably will also get covered in https://github.com/elastic/connectors/pull/2749#discussion_r1725529770

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: Yeah, we definitely could. I just copied this from prior DRA art for ent-search though. And I'd rather use a script that I know works already, than write a new one from scratch. 🙈

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: I don't think we need to worry, because if the process exits, nothing's going to log this env-var. I think this is here to make sure that nothing after this accidentally dumps the creds to buildkite logs.

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: discussed here: https://github.com/elastic/connectors/pull/2749#discussion_r1725533515

.buildkite/publish/dra/init_dra_publishing.sh
kderusso: I _suspect_ this can be removed but would like to see if @markjhoy confirms this

markjhoy: Yeah - it can be removed... this was to ensure it could be run without the final publish as a test... having the gate there is good for testing (as it can be set in the environment during a manual Buildkite run) - but if you don't want this testability without publishing - remove it...

.buildkite/publish/dra/publish-daily-release-artifact.sh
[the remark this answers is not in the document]
markjhoy: Having a logged list of the files that it is ready to use for DRA has come in handy so many times during debugging on `ent-search`... keep it, it'll help ;)

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: `realpath` is a function that folks often define in the first lines, like:
```
function realpath {
  echo "$(cd "$(dirname "$1")"; pwd)"/"$(basename "$1")";
}
export BUILDKITE_DIR=$(realpath "$(dirname "$SCRIPT_DIR")")
```

So it's doing the same thing, just as a one-liner (but I think you got dirname/realpath order backwards)

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: TIL about `trap`. However, reading about it, I don't think this is what we want. We need to clean up the SNAPSHOT dependency report file before we move on to the staging dependency reporting. We can't wait until exit.

And since we have `set -e`, there's no risk of something going wrong in producing the SNAPSHOT, leaving that dependency CSV, but progressing to the staging logic.

markjhoy: If you want to be extra safe, adding this (and the `unset`) into a `trap` command on exit is best as discussed, but since the Buildkite container will be removed after completion, I think it's fine and not a huge concern...

markjhoy (approved):
Other than the cleanups discussed, I think it 👍 looks good

.buildkite/publish/dra/init_dra_publishing.sh
markjhoy: > We need to clean up the SNAPSHOT dependency report file before we move on to the staging dependency reporting. We can't wait until exit.

Ah - my bad - I didn't notice that... carry on :)

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -142,12 +140,8 @@ if [[ "${PUBLISH_STAGING:-}" == "true" ]]; then
   setDraVaultCredentials
   export WORKFLOW="staging"
```
markjhoy, line 141: Just realized this - do you want to publish both on `staging` and `snapshot` for the `main` branch? IIRC - `staging` should only be for numbered branches...

seanstory: You're correct, and I believe that's handled correctly here? https://github.com/elastic/connectors/pull/2749/files#diff-4ec243366bbf09e6e1da80e47528b1fab33c1245a8e9c993c5d7a65e5464f927R44-R46

markjhoy: Correct - and I didn't see you already have the check for `PUBLISH_STAGING` above the block... 👍  all good

.buildkite/publish/dra/init_dra_publishing.sh
jedrazb: I think realpath is [UNIX cmnd](https://man7.org/linux/man-pages/man3/realpath.3.html)
```
(connectors-python) ➜  connectors-python git:(main) ✗ pwd
/Users/jedr/connectors-python
(connectors-python) ➜  connectors-python git:(main) ✗ export DRA_DIR="$(dirname "$(realpath connectors/sources)")"
(connectors-python) ➜  connectors-python git:(main) ✗ echo $DRA_DIR                                               
/Users/jedr/connectors-python/connectors
```

And if I understand correctly you want to get absolute path of the dirname of relative BASH_SOURCE[0] ? In that case I think it should get the order correctly

Nothing blocking though

jedrazb (approved):

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: TIL. I've always seen `realpath` defined explicitly. Like: https://github.com/elastic/ent-search/blob/aa06d741001a1e3a5a7a25d7356d1ec49e1a9555/dist/enterprise-search/package_assets/bin/enterprise-search#L8-L10

> Nothing blocking though

In that case, I'm going to leave it as-is, since this script is copied from elsewhere, it's pretty stable/mature. I'd rather make as few changes to it as possible, and avoid letting my own (obvious!) bash inexperience lead to sneaking in a bug.

Issue #201, opened 2022-11-22
add the list of deps licenses


tarekziade:
Add a `make` target that does:

```
pip install pip-licenses
pip-licenses --format=markdown > deps-licenses.md
```

and commit it

tarekziade:
Also, add the step of checking the licences every time we add a new connector

tarekziade:
pipdeptree run (anectodical):

```
aioboto3==10.1.0
  aiobotocore==2.4.0
    aiohttp==3.8.3
      aiosignal==1.3.1
        frozenlist==1.3.3
      async-timeout==4.0.2
      attrs==22.1.0
      charset-normalizer==2.1.1
      frozenlist==1.3.3
      multidict==6.0.2
      yarl==1.8.1
        idna==3.4
        multidict==6.0.2
    aioitertools==0.11.0
    botocore==1.27.59
      jmespath==1.0.1
      python-dateutil==2.8.2
        six==1.16.0
      urllib3==1.26.13
    wrapt==1.14.1
aiomysql==0.1.1
  PyMySQL==1.0.2
boto3==1.24.59
  botocore==1.27.59
    jmespath==1.0.1
    python-dateutil==2.8.2
      six==1.16.0
    urllib3==1.26.13
  jmespath==1.0.1
  s3transfer==0.6.0
    botocore==1.27.59
      jmespath==1.0.1
      python-dateutil==2.8.2
        six==1.16.0
      urllib3==1.26.13
cron-schedule-triggers==0.0.11
  pytz==2019.3
ecs-logging==2.0.0
elasticsearch==8.5.0
  elastic-transport==8.4.0
    certifi==2022.9.24
    urllib3==1.26.13
envyaml==1.10.211231
  PyYAML==6.0
guppy3==3.1.2
motor==3.0.0
  pymongo==4.3.3
    dnspython==2.2.1
pip==22.2.2
pip-licenses==4.0.1
  prettytable==3.5.0
    wcwidth==0.2.5
pipdeptree==2.3.3
Pympler==1.0.1
setuptools==63.2.0
smbprotocol==1.10.1
  cryptography==38.0.3
    cffi==1.15.1
      pycparser==2.21
  pyspnego==0.6.3
    cryptography==38.0.3
      cffi==1.15.1
        pycparser==2.21
```

seanstory:
Why do we need this?

tarekziade:
> Why do we need this?

[...]

PR #2757, opened 2024-08-20
csv dependency report

seanstory:
## Part of https://github.com/elastic/search-team/issues/8047
Relates to https://github.com/elastic/connectors/issues/201

For our DRA process, we need to produce a CSV of dependencies. This does that, building off a [prior PR ](https://github.com/elastic/connectors/pull/205/files) that had tried to do something similar.

Since I was introducing `pip-licenses`, I figured I'd go ahead and add a NOTICE.txt file while I was at it. This isn't done, which is why #201 will remain open. We still do not have:
- a CI check to make sure all the license are in the green list
- automation to provide any license that `pip-licenses` cannot find
- a CI check to make sure that any new deps added were also added to the NOTICE file

However, these are outside the scope of the current project, which is to expeditiously provide a CSV. We can iterate later.

#### Pre-Review Checklist


Makefile
```diff
@@ -18,7 +18,11 @@ bin/python: config.yml
 	bin/pip install --upgrade pip
 	bin/pip install --upgrade setuptools
 
-install: bin/python bin/elastic-ingest
+bin/pip-licenses: bin/python
+	bin/pip install pip-licenses
```
artem-shelkovnikov, line 22: This should go to our requirements files. Out of current requirements files best one is `tests.txt`, but we might wanna already rename it to `dev.txt` or something similar, cause we also have our linters there

artem-shelkovnikov: Let's keep it the way you've implemented it, good enough for now definitely

artem-shelkovnikov (approved):

PR #2759, opened 2024-08-20
[s3] Allow S3-compatible endpoints, expose more options

### LINKED ISSUE #2758: [s3] Allow usage of S3-Compatible endpoints
### Problem Description

Many cloud providers and self-hosted object storage software utilize S3-compatible APIs which are compatible with boto3 with some extra configuration. Currently, those configuration options are not exposed to the users.

### Proposed Solution
Pass-through options commonly utilized by S3-compatible endpoints:
* An `endpoint-url` override, users may set this to `https://ceph-rgw.example.com/` for example.
* An option to select [path-style bucket addressing](https://boto3.amazonaws.com/v1/documentation/api/1.9.42/guide/s3.html) as opposed to virtual-host

### Alternatives
N/A

### Additional Context
Rclone's documentation provides a good list of S3-compatible providers: https://rclone.org/s3/

nanderson94:
## Closes https://github.com/elastic/connectors/issues/2758

This change exposes many more configuration options of boto3 to the user. This has the primary intent of allowing S3-Compatible services to be used, but I also carried-through FIP and Dualstack endpoint selection.

#### Pre-Review Checklist

## Release Note

The S3 Connector now exposes options to support S3-Compatible object storage providers.

### DIFF
--- a/connectors/sources/s3.py
+++ b/connectors/sources/s3.py
@@ -47,7 +47,20 @@ class S3Client:
             read_timeout=self.configuration["read_timeout"],
             connect_timeout=self.configuration["connect_timeout"],
             retries={"max_attempts": self.configuration["max_attempts"]},
+            s3={
+                "addressing_style": self.configuration["s3_addressing_style"],
+            },
+            use_fips_endpoint=self.configuration["aws_use_fips_endpoint"],
+            use_dualstack_endpoint=self.configuration["aws_use_dualstack_endpoint"],
         )
+
+        if self.configuration["s3_endpoint_url"]:
+            self.s3_endpoint_url = self.configuration["s3_endpoint_url"]
+        elif AWS_ENDPOINT is not None:
+            self.s3_endpoint_url = AWS_ENDPOINT
+        else:
+            self.s3_endpoint_url = None
+
         self.clients = {}
         self.client_context = []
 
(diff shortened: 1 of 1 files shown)


connectors/sources/s3.py
```diff
@@ -362,6 +375,47 @@ def get_default_configuration(cls):
...
+                "order": 6,
+                "tooltip": "Not available in all regions.",
+                "type": "bool",
+                "ui_restrictions": ["advanced"],
+                "value": False,
+            },
+            "aws_use_dualstack_endpoint": {
+                "display": "toggle",
+                "label": "Use Dualstack Endpoints",
+                "order": 7,
```
navarone-feekery, line 413: Can you go through the `order` values for the configurations after this and update them?

navarone-feekery (commented):
Thanks for the contribution!
Could you add some unit tests to cover the new configurations to the [test file](https://github.com/elastic/connectors/blob/d9175248d89ba9c99ddc0801407b0362b44ae1ea/tests/sources/test_s3.py)?
I think some good tests would be:

1. ensuring `self.s3_endpoint_url` is assigned the right value in all circumstances
2. ensuring `self.config` initializes for all the options for new configs `s3_addressing_style`, `aws_use_fips_endpoint`, and `aws_use_dualstack_endpoint`.

Marty1703:
This would be a great addition for us. We are running on-premise and can't use AWS, but do have Ceph OBS. Could someone have a look at the PR and include the missing items from the checklist?

danai18:
This feature could help us a lot as well, as we are also running an on-premise instance.
Is there any way to contribute and make sure it'll be added?

PR #2762, opened 2024-08-21
[SPO] Add support of `sites.selected` permission

moxarth-elastic:
## Closes https://github.com/elastic/connectors/issues/2639

Makes the SPO connector compatible with the `Sites.Selected` permission. The current code fetches the sites and its metadata only with the `Sites.Read.All` permission but this PR makes it run with narrow permission `Sites.Selected`.

To run the connector with `Sites.Selected` permission, we have to provide a `read` access to the specific sites and then use them in the configuration field. To add a permission to specific site - refer this [documentation](https://techcommunity.microsoft.com/t5/microsoft-sharepoint-blog/develop-applications-that-use-sites-selected-permissions-for-spo/ba-p/3790476).

**Note:** If `Comma-separated list of sites` is set to `*` or user enables a toggle button - `Enumerate all sites`, the connector requires `Sites.Read.All` permission.

#### Pre-Review Checklist
- [x]  Covered the changes with automated tests

## Release Note

Sharepoint Online connector now works with `Sites.Selected` permission instead of broader permission `Sites.Read.All`.


connectors/sources/sharepoint_online.py
```diff
@@ -589,14 +589,7 @@ async def group_sites(self, group_id):
...
     async def site_collections(self):
-        filter_ = url_encode("siteCollection/root ne null")
-        select = "siteCollection,webUrl"
-
-        async for page in self._graph_api_client.scroll(
-            f"{GRAPH_API_URL}/sites/?$filter={filter_}&$select={select}"
-        ):
-            for site_collection in page:
-                yield site_collection
+        yield await self._graph_api_client.fetch(url=f"{GRAPH_API_URL}/sites/root")
```
seanstory, line 592: https://learn.microsoft.com/en-us/graph/api/resources/site?view=graph-rest-1.0 says:

> /sites/root: The tenant root site.

and https://learn.microsoft.com/en-us/graph/api/site-list?view=graph-rest-1.0&tabs=http says:

> `siteCollection/root ne null` Lists all root-level site collections in the organization. Useful for discovering the home site for each geography.

Does the change proposed here mean that we'll lose the ability to handle geography-specific site roots? What happens for an organization that has multiple roots?

connectors/sources/sharepoint_online.py
```diff
@@ -718,16 +711,23 @@ async def sites(
...
+                    if [WILDCARD] != allowed_root_sites and site[
+                        "name"
+                    ] not in allowed_root_sites:
+                        continue
+                    yield site
+        except PermissionsMissing as exception:
+            if allowed_root_sites == [WILDCARD]:
+                msg = "The configuration field 'Comma-separated list of sites' with '*' value is only compatible with 'Sites.Read.All' permission."
+            else:
+                msg = "To enumerate all sites, the connector requires 'Sites.Read.All' permission"
```
seanstory, line 729: how would we end up here, if `allowed_root_sites != [WILDCARD]`?

seanstory (commented):
Some clarifying questions:

connectors/sources/sharepoint_online.py
moxarth-elastic: That will fall under this block of code https://github.com/elastic/connectors/blob/efaa2d48b281110004d765390b8d9d218866b7e5/connectors/sources/sharepoint_online.py#L695-L710

`_all_sites` method is only responsible for this `if` condition - https://github.com/elastic/connectors/blob/efaa2d48b281110004d765390b8d9d218866b7e5/connectors/sources/sharepoint_online.py#L691

connectors/sources/sharepoint_online.py
moxarth-elastic: Yes, the updated code fetches the root site of the tenant. Since, the only endpoint `/sites/root` works with the `Sites.Selected` permission, and other endpoint like `/sites/getAllSites` or `/sites?filter=siteCollections/root ne null` requires `Sites.Read.All` permission. 
If we don't have to change the endpoint, we can work around it as follows:
```
call /sites?filter=siteCollections/root ne null
if we get AccessDenied/Forbidden (403 status code)
then call /sites/root
``` 
This can prevent the connector to miss the geo-specific root sites. The connector fetches the sites according to the permission given. 
WDYT?

seanstory: 👍 yeah, I like that approach better. This will protect us from any current users losing functionality that they rely on today. Thanks!

connectors/sources/sharepoint_online.py
seanstory: `or enumerate_all_sites:`

That was the piece I was missing. Thanks!

seanstory (approved):
LGTM. Thanks for the quick turn around!

seanstory:
I removed the v8.15 label, because this is not a bugfix, so we shouldn't backport to any patch releases

PR #2763, opened 2024-08-22
Replace python setup.py develop with pip install --editable .

artem-shelkovnikov:
## Part of https://github.com/elastic/search-team/issues/8096

The far-reaching goal of current work is to be able to have libraries separate from executables - for simplicity and maintenance, but also to be able to distribute executables separately. Unlikely that running connectors natively would need CLIs, or running connectors on agent would need a native connectors service.

This PR just starts the process by removing deprecated call to `python setup.py develop` and replacing it with `pip install --editable .`. This breaks tests as they do not know where to get the actual package code, so a change to `pytest.ini` was done.

#### Pre-Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ install: bin/python bin/elastic-ingest
 
 bin/elastic-ingest: bin/python
 	bin/pip install -r requirements/$(ARCH).txt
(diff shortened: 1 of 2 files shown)


seanstory (approved):
LGTM

PR #2765, opened 2024-08-22
add docker images to dra

seanstory:
## Part of https://github.com/elastic/search-team/issues/8047

This will add docker images to our DRA src dir. When combined with a change to the infra definition of our dra gradle file (https://github.com/elastic/infra/pull/41437), it will allow us to get rid of our own `release.yml` pipeline, and instead rely on the Unified Release and the Agent Release for all our needs going forward.

- snapshot: https://artifacts-snapshot.elastic.co/connectors/8.16.0-a895a968/summary-8.16.0-SNAPSHOT.html#artifacts
- staging: https://artifacts-staging.elastic.co/connectors/8.16.0-6d04a3e0/summary-8.16.0.html#artifacts

#### Pre-Review Checklist

## Related Pull Requests

* Infra PR: https://github.com/elastic/infra/pull/41437


.buildkite/nightly_docker.yml
seanstory, line 1: I don't think we need nightly dockers, if we can have a nightly `main` build that makes DRAs?

.buildkite/pipeline.yml
```diff
@@ -499,9 +508,73 @@ steps:
       - "smoke_test"
       - "relevant_ftests"
     steps:
-      - label: ":truck: Publish DRA Items"
+      # ----
+      # Docker builds for amd64
+      # ----
+      - label: "Building amd64 Docker image"
```
seanstory, line 514: These steps were just copied from `release-pipeline.yml`

.buildkite/publish/build-multiarch-docker.sh
seanstory, line 1: I'm told that the multi-arch bits are handled by DRA. Example for prodfiler: https://github.com/elastic/unified-release/blob/bc780387b5854e30e62f4f61273373df91d77c90/projects-config/prodfiler.yaml#L2-L9

.buildkite/publish/publish-common.sh
```diff
@@ -24,6 +24,3 @@ fi
 export BASE_TAG_NAME=${DOCKER_IMAGE_NAME:-docker.elastic.co/enterprise-search/elastic-connectors}
 export DOCKERFILE_PATH=${DOCKERFILE_PATH:-Dockerfile}
 export DOCKER_ARTIFACT_KEY=${DOCKER_ARTIFACT_KEY:-elastic-connectors-docker}
-
-export VAULT_ADDR=${VAULT_ADDR:-https://vault-ci-prod.elastic.dev}
-export VAULT_USER="docker-swiftypeadmin"
```
seanstory, line 29: Nice benefit is that we don't need to look up docker creds anymore, so no generic usage of vault secrets

.buildkite/run_functional_test.sh
```diff
@@ -28,18 +28,8 @@ export PIP=$VENV_ROOT/bin/pip
...
-# !!! WARNING be cautious about the following lines to avoid leaking the secrets in the CI logs
-set +x  # Do not remove so we don't leak passwords
-if [ -v BUILDKITE ]; then
-  echo "Connecting to Vault"
-  VAULT_ADDR=${VAULT_ADDR:-https://vault-ci-prod.elastic.dev}
-  VAULT_USER="docker-swiftypeadmin"
-  echo "Fetching Docker credentials for '$VAULT_USER' from Vault..."
-  DOCKER_USER=$(vault read -address "${VAULT_ADDR}" -field user_20230609 secret/ci/elastic-connectors/${VAULT_USER})
-  DOCKER_PASSWORD=$(vault read -address "${VAULT_ADDR}" -field secret_20230609 secret/ci/elastic-connectors/${VAULT_USER})
-  echo "Done!"
```
seanstory, line 41: I'm not sure that our ftests ever needed docker creds. These env vars don't seem to be used anywhere in our ftests.

catalog-info.yaml
```diff
@@ -48,6 +48,11 @@ spec:
       branch_configuration: "main"
       pipeline_file: ".buildkite/pipeline.yml"
       repository: "elastic/connectors"
+      schedules:
+        Daily main:
+          branch: main
+          cronline: '@daily'
+          message: "Builds, tests, and pushes daily `main` DRA artifacts"
```
seanstory, line 55: The goal is to replace the nightly dockers pipeline

catalog-info.yaml
```diff
@@ -150,81 +155,3 @@ spec:
...
-      pipeline_file: ".buildkite/nightly_docker.yml"
-      provider_settings:
-        trigger_mode: "none"
-      repository: "elastic/connectors"
-      schedules:
-        Daily 8_14:
-          branch: '8.14'
-          cronline: '@daily'
-          message: "Builds and pushes daily `8.14` docker images"
-        Daily 8_15:
```
seanstory, line 186: Is it safe to remove this on `main`? Will Backstage still be able to run these daily jobs for 8.14 and 8.15 as long as this exists in those branches?

catalog-info.yaml
```diff
@@ -150,81 +155,3 @@ spec:
...
-        search-productivity-team: {}
-
-########
-# Docker image build and publish - manual release
-########
----
-apiVersion: "backstage.io/v1alpha1"
-kind: "Resource"
-metadata:
-  name: "connectors-docker-build-publish"
```
seanstory, line 207: Similar question, is this safe to remove, or will it delete this pipeline for all branches if backstage doesn't see it in `main`?

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -19,7 +19,18 @@ make clean sdist
...
-echo "The artifacts are: $(ls $RELEASE_DIR/dist)"
+# Download previous step artifacts
+buildkite-agent artifact download '.artifacts/*.tar.gz*' $RELEASE_DIR/dist/ --step build_docker_image_amd64
+buildkite-agent artifact download '.artifacts/*.tar.gz*' $RELEASE_DIR/dist/ --step build_docker_image_arm64
+cp $RELEASE_DIR/dist/.artifacts/* $RELEASE_DIR/dist/dra-artifacts
+
+# Rename to match DRA expectations (<name>-<version>-<classifier>-<os>-<arch>) (also, x86_64 instead of amd64)
+cd $RELEASE_DIR/dist/dra-artifacts
+mv elastic-connectors-docker-$VERSION-amd64.tar.gz elastic-connectors-$VERSION-docker-image-linux-x86_64.tar.gz
+mv elastic-connectors-docker-$VERSION-arm64.tar.gz elastic-connectors-$VERSION-docker-image-linux-arm64.tar.gz
```
oli-g, line 30: We could replace `elastic-connectors` with a reference to a new variable (maybe `$PROJECT_NAME`?).

catalog-info.yaml
oli-g: > Will Backstage still be able to run these daily jobs

I think you mean Buildkite here: I'm really not sure unfortunately. If we register this change in Backstage, those scheduled jobs could be deleted from Buildkite?

catalog-info.yaml
oli-g: I've looked at the following docs:
* https://docs.elastic.dev/ci/catalog/catalog-info-setup
* https://docs.elastic.dev/ci/getting-started-with-buildkite-at-elastic#2-adjust-your-pipeline-settings

I think we have to preserve these `catalog-info.yaml` bits in `main`, while we could remove the unneeded pipelines definitions and scripts from `main`. @acrewdson Do you also see the same?

.buildkite/pipeline.yml
```diff
@@ -35,8 +40,12 @@ steps:
     timeout_in_minutes: 45
     agents:
       machineType: "n1-standard-8"
+      provider: "gcp"
+      useVault: true
+      image: family/enterprise-search-ubuntu-2204-connectors-py
```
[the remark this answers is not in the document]
oli-g: I don't think it's possible with plain YAML unfortunately.

catalog-info.yaml
acrewdson: yea my guess would have been 'removal from a yaml file in `main` == removal from buildkite'

catalog-info.yaml
acrewdson: yep, that would also be my guess

.buildkite/pipeline.yml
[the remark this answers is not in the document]
seanstory: Then I'll leave it as is. It's not too bad.

catalog-info.yaml
seanstory: Ok, so I should just change these to not run main, but leave them _in_ main for now? And I can add a YAML comment to explain that these will eventually phase out.

acrewdson: I think that sounds like the safest option, yep 👍

seanstory:
Ready for re-review!

oli-g (approved):
Do we need to verify that the Docker build steps work as expected, before merging this? 

I see those steps are only executed in `main` (see [here](https://github.com/elastic/connectors/blob/edaadb1b9631737c063250628e7491ccdcc89c03/.buildkite/pipeline.yml#L504C11-L504C23)): we could change that to `if: "(build.branch == \"main\" || build.pull_request.labels includes \"ci:packaging\")"`, as we do for [ent-search](https://github.com/elastic/ent-search/blob/main/.buildkite/pipeline.yml#L77). In that case, we should split the `:truck: DRA` group in two (one for packaging, one for publishing), again similarly to what we have for [ent-search](https://github.com/elastic/ent-search/blob/main/.buildkite/pipeline.yml#L72-L94).

seanstory:
> Do we need to verify that the Docker build steps work as expected, before merging this?

I did that before I opened this for review. I used to have `if: "(build.branch == \"main\" || build.branch == \"seanstory/add-docker-images-to-DRA\")` :) 
That's how I got the artifacts linked in the description.

> `build.pull_request.labels includes \"ci:packaging\")"`, as we do for [ent-search](https://github.com/elastic/ent-search/blob/main/.buildkite/pipeline.yml#L77). In that case, we should split the :truck: DRA group in two (one for packaging, one for publishing), again similarly to what we have for [ent-search](https://github.com/elastic/ent-search/blob/main/.buildkite/pipeline.yml#L72-L94).

I'm not opposed, but I didn't want to add too much scope and I think ent-search has magic to decide when to add that label. And I don't think we typically want maintenance branches to build these? But we can add this later if we want/need it.

seanstory:
> Do we need to verify that the Docker build steps work as expected, before merging this?

Whelp, I should have listened. 🤦 
https://buildkite.com/elastic/connectors/builds/13782#01919421-2208-42d5-8bb1-7e80813f3507 didn't pass. 

Follow-up PR incoming 💀

PR #2766, opened 2024-08-23
Revert "Replace python setup.py develop with pip install --editable ."

artem-shelkovnikov:
Reverts elastic/connectors#2763


navarone-feekery (approved):
:sad:

PR #2768, opened 2024-08-23
Move virtual environment to a separate directory

artem-shelkovnikov:
Part of project restructure:

For simplicity/cleanliness sake moving virtual environment to `.venv` directory in project root. 

All the files created by virtual environment are:

```
artemshelkovnikov-m2@Work-M2 connectors-py % ls .venv
CHANGELOG.rst LICENSE       README.rst    bin           include       lib           pyvenv.cfg    share
```

So after this change, all these files will be in `.venv` instead of project root - it will also solve the problem of some packages overwriting our LICENSE and adding unnecessary files to the project on installation.


Makefile
```diff
@@ -1,5 +1,3 @@
-.PHONY: test lint autoformat run ftest install release docker-build docker-run docker-push
```
navarone-feekery, line 1: Question: does removing all of these not break anything?

navarone-feekery (dismissed):
:rubberstamp: 
though I see CI is failing, happy to approve again if needed.

Makefile
artem-shelkovnikov: Reference: https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile

`.PHONY` indicates that the build target is not a file, but a command. In essence `.PHONY test` says "even if there's a file called test", it should not be taken into account, but a command `test` should be executed.

navarone-feekery (approved):

.buildkite/run_functional_test.sh
```diff
@@ -18,12 +18,13 @@ pyenv global
 
 BASEDIR=$(realpath $(dirname $0))
 ROOT=$(realpath $BASEDIR/../)
+VENV_ROOT=$ROOT/.venv
```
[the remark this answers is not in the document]
artem-shelkovnikov: Good catch, renaming in https://github.com/elastic/connectors/pull/2781

PR #2770, opened 2024-08-23
Swap to pyproject.toml and use it for making builds

artem-shelkovnikov:
## Part of https://github.com/elastic/search-team/issues/8165

Removing `setup.py` and moving to `pyproject.toml`.

`setup.py` was an executable, but `pyproject.toml` is a markup file that describes "how to build a package". Actual build system used is hatch: https://hatch.pypa.io/latest/ (took it from other Elastic python projects).

This allows a smoother path forward when splitting the code into library code and applications, for now everything should just work as is.

#### Pre-Review Checklist


Makefile
```diff
@@ -24,9 +24,12 @@ config.yml:
 install: .venv/bin/python .venv/bin/pip-licenses .venv/bin/elastic-ingest
 	.venv/bin/pip-licenses --format=plain-vertical --with-license-file --no-license-path > NOTICE.txt
 
+.venv/bin/hatch: .venv/bin/python
+	.venv/bin/pip install -r requirements/$(ARCH).txt
+
 .venv/bin/elastic-ingest: .venv/bin/python
 	.venv/bin/pip install -r requirements/$(ARCH).txt
-	.venv/bin/python setup.py develop
+	.venv/bin/pip install --editable .
```
jedrazb, line 32: Why is the `--editable` ? 

`make run` uses this command to run the app in prod use cases, in such scenario we probably don't want to include `--editable` flag?

pyproject.toml
```diff
@@ -0,0 +1,106 @@
+[project]
```
jedrazb, line 1: Should we also add `license` - it can point to file? 

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

artem-shelkovnikov:
@jedrazb thanks for the review, I'll shelve this PR for now to focus on agent first, then I'll start another one keeping your comments in mind

PR #2772, opened 2024-08-23
Comment out release-manager

seanstory:
Stop-gap to fix main build until https://github.com/elastic/crawler/pull/103 is ready


artem-shelkovnikov (approved):
Approving a temporary fix, we can fix it next week to keep our CI green!

PR #2787, opened 2024-08-28
Add code for running connectors in Agent

artem-shelkovnikov:
## Part of https://github.com/elastic/search-team/issues/8096

This PR adds a way to run Connectors Service in Agent.

Entry point is `connectors/agent/cli.py` - if it's ran from an Agent component via the python in `.venv/bin/python` it will enable Connector Service running within an Agent (docker or not).

This PR adds bare bone logic: positive cases work, so expect tricky bugs to still be out there if we haven't accounted for anything.

I've tried to keep it very small, and I think there's no way to make it simpler/smaller :)

Additionally - I had to exclude tests for the component from `make test` because it would break `make test` for people who have no access to the python agent client repository - this one will be opened later so we can revert this part of the changeset. If you need to run tests there is a make target `make test-agent` - I kept high coverage for future.

#### Pre-Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,9 @@ config.yml:
 install: .venv/bin/python .venv/bin/pip-licenses .venv/bin/elastic-ingest
 	.venv/bin/pip-licenses --format=plain-vertical --with-license-file --no-license-path > NOTICE.txt
 
+install-agent: install
+	.venv/bin/pip install -r requirements/agent.txt
+
 .venv/bin/elastic-ingest: .venv/bin/python
 	.venv/bin/pip install -r requirements/$(ARCH).txt
 	.venv/bin/python setup.py develop
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,10 @@ autoformat: .venv/bin/python .venv/bin/black .venv/bin/elastic-ingest
 	.venv/bin/ruff scripts --fix
 
 test: .venv/bin/pytest .venv/bin/elastic-ingest
-	.venv/bin/pytest --cov-report term-missing --cov-fail-under 92 --cov-report html --cov=connectors --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests
+	.venv/bin/pytest --cov-report term-missing --cov-fail-under 92 --cov-report html --cov=connectors --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests --ignore tests/agent
+
+test-agent: .venv/bin/pytest .venv/bin/elastic-ingest install-agent
+	.venv/bin/pytest --cov-report term-missing --cov-fail-under 92 --cov-report html --cov=connectors/agent --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests/agent
 
 release: install
 	.venv/bin/python setup.py sdist
--- a/connectors/agent/cli.py
+++ b/connectors/agent/cli.py
@@ -11,6 +11,11 @@ from connectors.agent.component import ConnectorsAgentComponent
 
 
 def main(args=None):
+    """Script entry point into running Connectors Service on Agent.
+
+    It initialises an event loop, creates a component and runs the component.
+    Additionally, signals are handled for graceful termination of the component.
+    """
     loop = asyncio.get_event_loop()
     logger.info("Running agent")
     component = ConnectorsAgentComponent()
--- a/connectors/agent/cli.py
+++ b/connectors/agent/cli.py
@@ -2,10 +2,10 @@ import asyncio
 import functools
 import signal
 
-from es_agent_client.util.async_tools import (
+from elastic_agent_client.util.async_tools import (
     sleeps_for_retryable,
 )
-from es_agent_client.util.logger import logger
+from elastic_agent_client.util.logger import logger
 
 from connectors.agent.component import ConnectorsAgentComponent
 
--- a/connectors/agent/cli.py
+++ b/connectors/agent/cli.py
@@ -1,0 +1,32 @@
+import asyncio
+import functools
+import signal
+
+from es_agent_client.util.async_tools import (
+    sleeps_for_retryable,
+)
+from es_agent_client.util.logger import logger
+
+from connectors.agent.component import ConnectorsAgentComponent
+
+
+def main(args=None):
+    loop = asyncio.get_event_loop()
+    logger.info("Running agent")
+    component = ConnectorsAgentComponent()
+
+    def _shutdown(signal_name):
+        sleeps_for_retryable.cancel(signal_name)
+        component.stop(signal_name)
+
+    for sig in (signal.SIGINT, signal.SIGTERM):
+        loop.add_signal_handler(sig, functools.partial(_shutdown, sig.name))
+
+    return loop.run_until_complete(component.run())
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    finally:
+        logger.info("Bye")
--- a/connectors/agent/cli.py
+++ b/connectors/agent/cli.py
@@ -1,3 +1,8 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
 import asyncio
 import functools
 import signal
--- a/connectors/agent/component.py
+++ b/connectors/agent/component.py
@@ -5,7 +5,7 @@ from elastic_agent_client.reader import new_v2_from_reader
 from elastic_agent_client.service.actions import ActionsService
 from elastic_agent_client.service.checkin import CheckinV2Service
 
-from connectors.agent.config import ConnectorsAgentConfiguration
+from connectors.agent.config import ConnectorsAgentConfigurationWrapper
 from connectors.agent.protocol import ConnectorActionHandler, ConnectorCheckinHandler
 from connectors.agent.service_manager import ConnectorServiceManager
 from connectors.services.base import MultiService
@@ -14,20 +14,43 @@ CONNECTOR_SERVICE = "connector-service"
 
 
 class ConnectorsAgentComponent:
+    """Entry point into running connectors service in Agent.
+
+    This class provides a simple abstraction over Agent components and Connectors Service manager.
+
+    It instantiates everything needed to read from Agent protocol, creates a wrapper around Connectors Service
+    and provides applied interface to be able to run it in 2 simple methods: run and stop.
+    """
+
     def __init__(self):
+        """Inits the class.
+
+        Init should be safe to call without expectations of side effects (connections to Agent, blocking or anything).
+        """
         self.ver = VersionInfo(
             name=CONNECTOR_SERVICE, meta={"input": CONNECTOR_SERVICE}
         )
         self.opts = V2Options()
         self.buffer = sys.stdin.buffer
-        self.agent_config = ConnectorsAgentConfiguration()
+        self.config_wrapper = ConnectorsAgentConfigurationWrapper()
 
     async def run(self):
+        """Start reading from Agent protocol and run Connectors Service with settings reported by agent.
+
+        This method can block if it's not running from Agent - it expects the client to be able to read messages
+        on initialisation. These messages are a handshake and it's a sync handshake.
+        If no messages are sent, this method can and will hang.
+
+        However, if ran under agent, this method will read configuration from Agent and attempt to start an
+        instance of Connectors Service with this configuration.
+
+        Additionally services for handling Check-in and Actions will be started to implement the protocol correctly.
+        """
         client = new_v2_from_reader(self.buffer, self.ver, self.opts)
         action_handler = ConnectorActionHandler()
-        self.connector_service_manager = ConnectorServiceManager(self.agent_config)
+        self.connector_service_manager = ConnectorServiceManager(self.config_wrapper)
         checkin_handler = ConnectorCheckinHandler(
-            client, self.agent_config, self.connector_service_manager
+            client, self.config_wrapper, self.connector_service_manager
         )
 
         self.multi_service = MultiService(
@@ -39,4 +62,8 @@ class ConnectorsAgentComponent:
         await self.multi_service.run()
 
     def stop(self, sig):
+        """Shutdown everything running in the component.
+
+        Attempts to gracefully shutdown the services that are running under the component.
+        """
         self.multi_service.shutdown(sig)
--- a/connectors/agent/component.py
+++ b/connectors/agent/component.py
@@ -40,4 +40,3 @@ class ConnectorsAgentComponent:
 
     def stop(self, sig):
         self.multi_service.shutdown(sig)
-        self.connector_service_manager.shutdown_multi_service(sig)
--- a/connectors/agent/component.py
+++ b/connectors/agent/component.py
@@ -1,0 +1,43 @@
+import sys
+
+from es_agent_client.client import V2Options, VersionInfo
+from es_agent_client.reader import new_v2_from_reader
+from es_agent_client.service.actions import ActionsService
+from es_agent_client.service.checkin import CheckinV2Service
+
+from connectors.agent.config import ConnectorsAgentConfiguration
+from connectors.agent.protocol import ConnectorActionHandler, ConnectorCheckinHandler
+from connectors.agent.service_manager import ConnectorServiceManager
+from connectors.services.base import MultiService
+
+CONNECTOR_SERVICE = "connector-service"
+
+
+class ConnectorsAgentComponent:
+    def __init__(self):
+        self.ver = VersionInfo(
+            name=CONNECTOR_SERVICE, meta={"input": CONNECTOR_SERVICE}
+        )
+        self.opts = V2Options()
+        self.buffer = sys.stdin.buffer
+        self.agent_config = ConnectorsAgentConfiguration()
+
+    async def run(self):
+        client = new_v2_from_reader(self.buffer, self.ver, self.opts)
+        action_handler = ConnectorActionHandler()
+        self.connector_service_manager = ConnectorServiceManager(self.agent_config)
+        checkin_handler = ConnectorCheckinHandler(
+            client, self.agent_config, self.connector_service_manager
+        )
+
+        self.multi_service = MultiService(
+            CheckinV2Service(client, checkin_handler),
+            ActionsService(client, action_handler),
+            self.connector_service_manager,
+        )
+
+        await self.multi_service.run()
+
+    def stop(self, sig):
+        self.multi_service.shutdown(sig)
+        self.connector_service_manager.shutdown_multi_service(sig)
--- a/connectors/agent/component.py
+++ b/connectors/agent/component.py
@@ -1,9 +1,9 @@
 import sys
 
-from es_agent_client.client import V2Options, VersionInfo
-from es_agent_client.reader import new_v2_from_reader
-from es_agent_client.service.actions import ActionsService
-from es_agent_client.service.checkin import CheckinV2Service
+from elastic_agent_client.client import V2Options, VersionInfo
+from elastic_agent_client.reader import new_v2_from_reader
+from elastic_agent_client.service.actions import ActionsService
+from elastic_agent_client.service.checkin import CheckinV2Service
 
 from connectors.agent.config import ConnectorsAgentConfiguration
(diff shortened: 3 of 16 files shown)


requirements/ftest.txt
```diff
@@ -1,7 +1,7 @@
 flask==2.3.2
 flask_limiter==3.3.1
 faker==18.11.2
-mysql-connector-python==8.1.0
+mysql-connector-python==9.0.0
```
seanstory, line 4: why? Major jumps scare me. :)

connectors/services/job_cleanup.py
```diff
@@ -34,6 +34,8 @@ async def _run(self):
             while self.running:
                 await self._process_orphaned_idle_jobs()
                 await self._process_idle_jobs()
+                if not self.running:
+                    break
```
seanstory, line 38: needs a rebase?

seanstory (commented):
Haven't really read the logic yet (since this is still in draft) but I like the structure/direction. :feelsgood:

connectors/agent/service_manager.py
```diff
@@ -0,0 +1,53 @@
...
+    def stop(self):
+        logger.info("Service manager is stopping connector services")
+        self.running = False
+        if self._multi_service:
+            self._multi_service.shutdown(None)
+
+    def restart(self):
+        logger.info("Service manager is restarting connector services")
+        if self._multi_service:
+            self._multi_service.shutdown(None)
```
jedrazb, line 53: I get the reasoning behind using `None` as a signal, but it will produce info logs like `Caught None. Graceful shutdown.`

I looked at shutdown def and the `sig` argument can be anything, should we just use a constant String e.g. `AGENT_SERVICE_RESTART` so that logs are:

```
Caught AGENT_SERVICE_RESTART. Graceful shutdown.
```

jedrazb: Now I see that we are logging `info` one line above `Service manager is restarting connector services` so I guess it's fine to keep this at it is

connectors/agent/service_manager.py
```diff
@@ -0,0 +1,53 @@
...
+                        ["schedule", "sync_content", "sync_access_control", "cleanup"],
+                        self.agent_config.get(),
+                    )
+                    await self._multi_service.run()
+                except Exception as e:
+                    logger.exception(
+                        f"Error while running services in ConnectorServiceManager: {e}"
+                    )
+                    raise
+        finally:
```
jedrazb, line 41: should we care about stopping running services here?

jedrazb (approved):
LGTM, couple of nits

jedrazb (commented):
I meant to "comment", did you manage to run this with es_agent_client as dependency?

connectors/agent/component.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class ConnectorsAgentComponent:
+    def __init__(self):
+        self.ver = VersionInfo(
+            name=CONNECTOR_SERVICE, meta={"input": CONNECTOR_SERVICE}
+        )
+        self.opts = V2Options()
+        self.buffer = sys.stdin.buffer
+        self.agent_config = ConnectorsAgentConfiguration()
```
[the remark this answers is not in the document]
artem-shelkovnikov: Yeah I'll try renaming it, but I'm also not sure about this class - maybe we can even pass agent configuration inside, parse it and emit connectors configuration to just separate this out of the checkin handler.

connectors/agent/protocol.py
```diff
@@ -0,0 +1,67 @@
...
+                        "password"
+                    ):
+                        es_creds["username"] = source["username"]
+                        es_creds["password"] = source["password"]
+                    else:
+                        msg = "Invalid Elasticsearch credentials"
+                        raise ValueError(msg)
+
+                    new_config = {
+                        "elasticsearch": es_creds,
```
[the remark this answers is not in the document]
artem-shelkovnikov: Oh absolutely!

connectors/agent/service_manager.py
```diff
@@ -0,0 +1,53 @@
...
+            msg = f"{self.__class__.__name__} is already running."
+            raise ServiceAlreadyRunningError(msg)
+
+        self.running = True
+        try:
+            while self.running:
+                try:
+                    logger.info("Starting connector services")
+                    self._multi_service = get_services(
+                        ["schedule", "sync_content", "sync_access_control", "cleanup"],
```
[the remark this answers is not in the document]
artem-shelkovnikov: We can restructure even here :)

connectors/agent/service_manager.py
artem-shelkovnikov: It should already be stopped - it's not possible to reach here without stopping. I can add additional step of verifying that everything stopped, but not sure it's needed

connectors/agent/service_manager.py
artem-shelkovnikov: Yeah `None` here is arbitrary. I'll look into making it a default None and handle None in multiservice by producing a different log line

connectors/services/job_cleanup.py
artem-shelkovnikov: Yup!

requirements/ftest.txt
artem-shelkovnikov: Transitive dependency bump - `mysql-connector-python` needed protobuf of unsupported version:

```
mysql-connector-python==8.1.0
└── protobuf [required: >=4.21.1,<=4.21.12, installed: 4.21.12]
```

Elastic Python Agent repo has 5.27+ protobuf declared as a dependency.

This package is also only used by functional tests, so should not affect production code.

artem-shelkovnikov:
> I meant to "comment", did you manage to run this with es_agent_client as dependency?

Oh yes, this branch is 100% functioning

connectors/agent/component.py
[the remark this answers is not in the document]
artem-shelkovnikov: I renamed it to `ConnectorsAgentConfigurationWrapper` and changed logic a bit + added code comments. I really could not make good names for stuff that I wrote here, open to suggestions

connectors/agent/service_manager.py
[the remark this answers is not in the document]
artem-shelkovnikov: Though let's do later? I like keeping this changeset reasonably small to be able to finish faster and allow for the packaging story to continue

connectors/agent/service_manager.py
artem-shelkovnikov: Let's address in subsequent PRs?

connectors/agent/service_manager.py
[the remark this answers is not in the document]
jedrazb: Yeah we can do it later!

Makefile
```diff
@@ -64,7 +67,10 @@ autoformat: .venv/bin/python .venv/bin/black .venv/bin/elastic-ingest
 	.venv/bin/ruff scripts --fix
 
 test: .venv/bin/pytest .venv/bin/elastic-ingest
-	.venv/bin/pytest --cov-report term-missing --cov-fail-under 92 --cov-report html --cov=connectors --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests
+	.venv/bin/pytest --cov-report term-missing --cov-fail-under 92 --cov-report html --cov=connectors --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests --ignore tests/agent
```
artem-shelkovnikov, line 70: Ignoring `tests/agent` here cause repo that we build package from is private and it'll fail in CI. We can set it up in CI to pull the repo too, but occasional user of main can run into problems and I'd rather keep `main` clean and functioning

connectors/config.py
```diff
@@ -27,6 +27,11 @@ def load_config(config_file):
     return configuration
 
 
+def add_defaults(config):
```
artem-shelkovnikov, line 30: Not sure about method name 🤷

requirements/agent.txt
```diff
@@ -0,0 +1 @@
+elastic-agent-client@git+https://github.com/elastic/python-elastic-agent-client@main
```
artem-shelkovnikov, line 1: For now we just pick `main`, later we'll have more solid plan, hopefully

tests/agent/test_component.py
```diff
@@ -0,0 +1,40 @@
+import asyncio
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from connectors.agent.component import ConnectorsAgentComponent
+
+
+class StubMultiService:
```
artem-shelkovnikov, line 9: Stub makes the test code much cleaner IMO so I went for it instead of a mock

Makefile
jedrazb: > I'd challenge that we don't care about 2xx, in operational review discussions we've discussed that bumps in traffic

++

connectors/agent/config.py
```diff
@@ -0,0 +1,105 @@
...
+    def try_update(self, source):
+        """Try update the configuration and see if it changed.
+
+        This method takes the check-in event coming from Agent and checks if config needs an update.
+
+        If update is needed, configuration is updated and method returns True. If no update is needed
+        the method returns False.
+        """
+
+        # TODO: find a good link to what this object is.
```
jedrazb, line 62: Trying to be helpful, so digged into proto def, https://github.com/elastic/elastic-agent-client/blob/main/elastic-agent-client.proto#L168 

But it seems like `source` is a an arbitrary struct hehe (so we have to look somewhere else for the actual definition) and take a leap of faith now

connectors/agent/protocol.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    async def handle_action(self, action: proto.ActionRequest):
+        """Implementation of BaseActionHandler.handle_action
+
+        Right now does nothing as Connectors Service has no actions to respond to.
+        """
+        msg = (
+            f"This connector component can't handle action requests. Received: {action}"
+        )
```
jedrazb, line 20: should we include the error log? I don't see any other "wrapper" logic that would surface this to the user / logs

connectors/agent/protocol.py
```diff
@@ -0,0 +1,68 @@
...
+        raise NotImplementedError(msg)
+
+
+class ConnectorCheckinHandler(BaseCheckinHandler):
+    """Class handling to Agent check-in events.
+
+    Agent sends check-in events from time to time that might contain
```
jedrazb, line 27: from time to time -> "at least every 30s"  https://github.com/elastic/elastic-agent-client/blob/main/elastic-agent-client.proto#L29

So if apply_from_client blocks for more than 30s the component is marked as failed

connectors/agent/service_manager.py
```diff
@@ -0,0 +1,91 @@
...
+
+        Once started, the service attempts to run all needed connector subservices
+        in parallel via MultiService.
+
+        Service can be restarted - it will keep running but with refreshed config,
+        or it can be stopped - it will just gracefully shut down.
+        """
+        if self._running:
+            msg = f"{self.__class__.__name__} is already running."
+            raise ServiceAlreadyRunningError(msg)
```
jedrazb, line 50: should we add the error log, I'm not sure we catch-and-log the exception anywhere

requirements/agent.txt
jedrazb: "Always releasable" enforced here 🤙

jedrazb (dismissed):
LGTM. Couple of questions about error logging and doc improvements suggestion, nothing blocking I think

Makefile
artem-shelkovnikov: Sorry what's that quote? :D

connectors/agent/service_manager.py
artem-shelkovnikov: This one is not catched, it'll just crash the caller. Supposedly this case is "never gonna happen", so I'd just keep it like this.

Makefile
seanstory: lol looks like a paste error from a thread in https://github.com/elastic/cloud/pull/130868#pullrequestreview-2271519957

Makefile
```diff
@@ -64,7 +67,10 @@ autoformat: .venv/bin/python .venv/bin/black .venv/bin/elastic-ingest
...
+
+test-agent: .venv/bin/pytest .venv/bin/elastic-ingest install-agent
```
seanstory, line 72: I think we should add this to CI now, to avoid forgetting or letting it get too behind in coverage. I _think_ that ElasticMachine should be able to clone the private repo. Hit me up if it's not working.

connectors/agent/README.md
seanstory, line 1: If we don't have anything to say here yet, let's just remove the file for now.

connectors/agent/cli.py
seanstory, line 1: Missing license header.
We should find a way to add this to our linter, I bet we have a lot of files missing this. Only noticed because it's top-of-mind.

connectors/agent/component.py
```diff
@@ -0,0 +1,69 @@
...
+
+from elastic_agent_client.client import V2Options, VersionInfo
+from elastic_agent_client.reader import new_v2_from_reader
+from elastic_agent_client.service.actions import ActionsService
+from elastic_agent_client.service.checkin import CheckinV2Service
+
+from connectors.agent.config import ConnectorsAgentConfigurationWrapper
+from connectors.agent.protocol import ConnectorActionHandler, ConnectorCheckinHandler
+from connectors.agent.service_manager import ConnectorServiceManager
+from connectors.services.base import MultiService
```
seanstory, line 11: This is using the MulitService from Connectors instead of from the elastic_agent_client. Which I _think_ is ok (and probably for the best) but just wanted to flag in case it wasn't intentional. I hadn't tested using ActionsService or CheckinV2Service with the Connectors MultiService.

connectors/agent/config.py
```diff
@@ -0,0 +1,105 @@
...
+                "network_drive",
+                "postgresql",
+                "s3",
+                "salesforce",
+                "servicenow",
+                "sharepoint_online",
+                "slack",
+                "microsoft_teams",
+                "zoom",
+            ],
```
seanstory, line 48: For starters, should we limit this just to github?

connectors/agent/config.py
seanstory: The way I interrogated this during my POC was to just add logging. I'd get a checkin event, deserialize it, re-serialize it as json, then dump it in the logs.

connectors/agent/config.py
```diff
@@ -0,0 +1,105 @@
...
+                msg = "Invalid Elasticsearch credentials"
+                raise ValueError(msg)
+
+            new_config = {
+                "elasticsearch": es_creds,
+            }
+            self.specific_config = new_config
+            return True
+
+        return False
```
seanstory, line 86: Should probably add an error log here or something to leave a breadcrumb if we can't find basic auth or an api key

connectors/agent/protocol.py
seanstory: It should log the error. See: https://github.com/elastic/python-elastic-agent-client/blob/ba2525d23ed424c5740d9bbf87baff80a90533e7/elastic_agent_client/service/actions.py#L51-L61

It won't terminate, no. It just sends a message back through the protocol that the requested action failed. See this slack thread: https://elastic.slack.com/archives/C01QQ449KE1/p1719235054818189?thread_ts=1718987763.249969&cid=C01QQ449KE1

connectors/agent/protocol.py
```diff
@@ -0,0 +1,72 @@
+from elastic_agent_client.generated import elastic_agent_client_pb2 as proto
+from elastic_agent_client.handler.action import BaseActionHandler
+from elastic_agent_client.handler.checkin import BaseCheckinHandler
+from elastic_agent_client.util.logger import logger
+
+
+class ConnectorActionHandler(BaseActionHandler):
```
seanstory, line 7: Super nit/opinion - I don't like having multiple classes in one file like this, since I can never remember file names. I'd rather see each of these in their own file.

connectors/agent/protocol.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, client, agent_connectors_config_wrapper, service_manager):
+        """Inits the class.
+
+        Initing this class should not produce side-effects.
+        """
+        super().__init__(client)
+        self.agent_connectors_config_wrapper = agent_connectors_config_wrapper
+        self.service_manager = service_manager
+
+    async def apply_from_client(self):
```
seanstory, line 43: This looks a lot like the "Fake example", but I'm worried we'll need to implement more here. This seems to only look at the first unit's `config`, but doesn't consider `log_level`, `features`, or anything else from the checkin service's `sync_component` or `sync_units` method.

connectors/agent/service_manager.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class ConnectorServiceManager:
+    """Class responsible for properly configuring and running Connectors Service in Elastic Agent
+
+    ConnectorServiceManager is a middle man between Elastic Agent and ConnectorsService.
+
+    This class is taking care of starting Connectors Service subservices with correct configuration.
+    If configuration changes, as reported by Agent, then this Manager class gracefully shuts down the
+    subservices and starts them again with new configuration.
```
seanstory, line 16: I'm loving these docstrings. Makes this PR so much easier to grok. :)

connectors/config.py
seanstory: I like it. ;)

tests/agent/test_agent_config.py
```diff
@@ -0,0 +1,50 @@
...
+    assert config_wrapper.try_update(source_mock) is True
+    assert config_wrapper.get()["elasticsearch"]["host"] == hosts[0]
+    assert config_wrapper.get()["elasticsearch"]["api_key"] == api_key
+
+
+@pytest.mark.asyncio
+async def test_try_update_with_basic_auth_auth_data():
+    hosts = ["https://localhost:9200"]
+    username = "elastic"
+    password = "hold the door"
```
seanstory, line 38: anti-password? 😂

tests/agent/test_cli.py
```diff
@@ -0,0 +1,20 @@
...
+    async def kill():
+        await asyncio.sleep(0.2)
+        os.kill(os.getpid(), signal.SIGTERM)
+
+    loop = asyncio.new_event_loop()
+    loop.create_task(kill())
+
+    main()
+
+    loop.close()
```
seanstory, line 20: nothing's asserted here. Intended?

seanstory (approved):
Bunch of little nits, but I'm ok with addressing them in follow-ups rather than blocking this.

Awesome stuff. I want to see a demo an make some toasts!

connectors/agent/config.py
```diff
@@ -0,0 +1,105 @@
...
+        has_hosts = source.fields.get("hosts")
+        has_api_key = source.fields.get("api_key")
+        has_basic_auth = source.fields.get("username") and source.fields.get("password")
+        if has_hosts and (has_api_key or has_basic_auth):
+            es_creds = {
+                "host": source["hosts"][0],
+            }
+
+            if source.fields.get("api_key"):
+                es_creds["api_key"] = source["api_key"]
```
jedrazb, line 72: @artem-shelkovnikov commenting so that it doesn't get lost, by default we would get Fleet-specific api key format, this should do the trick to turn it to our version:

```
    if source.fields.get("api_key"):
        api_key = source["api_key"]
        # if beats_logstash_format
        if ":" in api_key:
            api_key = base64.b64encode(api_key.encode()).decode()
```

connectors/agent/config.py
jedrazb: This will coming from the agent protocol, for now I guess it's fine to leave this as it is?

jedrazb (dismissed):

connectors/agent/config.py
jedrazb: Verified that this works

artem-shelkovnikov: Thanks for the investigation! I'm gonna add this code as well :)

connectors/agent/component.py
artem-shelkovnikov: It's technically same code - we need to clean up the python-agent repo from some classes like this or move them to `example` namespace to not distribute these classes.

tests/agent/test_cli.py
artem-shelkovnikov: Yup. `main` blocks so if the test fails the test will be killed by a timeout. I can add some asserts for log messages too, but not sure it's needed

seanstory: Maybe just a comment then, with what you've written in this thread. 👍

PR #2795, opened 2024-09-03
Bump ruff and add license check

artem-shelkovnikov:
This PR:
- Bumps ruff to freshest version (quite a huge bump)
- Runs `autoformat`
- Adds license rule `CPY001` - https://docs.astral.sh/ruff/rules/missing-copyright-notice/. It's in preview, so added preview flag too
- Minor other changes to make linter work (see comments in the PR)


.ruff.toml
```diff
@@ -13,14 +17,16 @@ exclude = [
...
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
-target-version = "py310"
 
-[per-file-ignores]
+[lint.per-file-ignores]
 "connectors/*" = ["S608"]
 "scripts/verify.py" = [ "EM" ]
 "setup.py" = [ "EM" ]
-"tests/*" = ["B017", "S101", "S"]
+"tests/*" = ["B017", "S101", "S", "ASYNC110"]
```
artem-shelkovnikov, line 25: Complained about insignificent changes in tests, so decided to ignore

.ruff.toml
```diff
@@ -13,14 +17,16 @@ exclude = [
...
 "tests/conftest.py" = [ "EM" ]
 
-[isort]
+[lint.isort]
 known-first-party=["connectors", "tests"]
+
+[lint.flake8-copyright]
+notice-rgx = "#\n# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one\n# or more contributor license agreements. Licensed under the Elastic License 2.0;\n# you may not use this file except in compliance with the Elastic License 2.0.\n#"
```
artem-shelkovnikov, line 32: That's pretty much our header, this config verifies that first 4096 bytes of the file start with this regex

Makefile
```diff
@@ -41,27 +41,27 @@ install: .venv/bin/python .venv/bin/pip-licenses .venv/bin/elastic-ingest
...
-	.venv/bin/black --check tests
-	.venv/bin/black --check setup.py
-	.venv/bin/black --check scripts
-	.venv/bin/ruff connectors
-	.venv/bin/ruff tests
-	.venv/bin/ruff setup.py
-	.venv/bin/ruff scripts
+lint: .venv/bin/python .venv/bin/ruff .venv/bin/elastic-ingest
+	.venv/bin/ruff check connectors
+	.venv/bin/ruff format connectors --check
```
artem-shelkovnikov, line 46: Fully identical to running `.venv/bin/black --check tests`

connectors/sources/__init__.py
```diff
@@ -1,5 +0,0 @@
-#
```
artem-shelkovnikov, line 1: This file does not seem needed and it triggers weird linter error, so I decided to delete

connectors/utils.py
```diff
@@ -13,7 +13,7 @@
 import re
 import shutil
 import ssl
-import subprocess
+import subprocess  # noqa S404
```
artem-shelkovnikov, line 16: This one is the most dangerous, please take a look at it too to make sure I did not ignore something that is super problematic

scripts/deps-csv.py
```diff
@@ -1,4 +1,8 @@
-#!/usr/bin/python3
```
artem-shelkovnikov, line 1: I took liberty to drop the shebang here cause we don't run python2 anywhere afaik

connectors/utils.py
navarone-feekery: I think it's fine and not a blocker. Looking at the code, it's only used for converting to base 64:
https://github.com/elastic/connectors/blob/68add520d325bdd000ff372dab401260b7770dad/connectors/utils.py#L256-L268
Checking git blame, it seems it has been that way since implementation, and was due to speed + less memory cost: https://github.com/elastic/connectors/pull/276

artem-shelkovnikov: Yeah that's what I feel too - we don't use `subprocess` for anything else.

navarone-feekery (approved):
LGTM!

.ruff.toml
seanstory: omg this is amazing

seanstory (approved):
🥇

PR #2796, opened 2024-09-03
Allow connectors to sync content into alias 

artem-shelkovnikov:
## Part of https://github.com/elastic/search-team/issues/8228

We need to make connectors work with aliases. This PR takes a step forward this without backwards compatibility problems.

What it does is: if you attach connector to an alias of an index, connector will be able to sync data through the alias.

For now it's not possible to do it in UI at all - it needs to be done via Dev Tools or CLI to make it function, so for end user nothing changes yet.

#### Pre-Review Checklist


connectors/es/utils.py
```diff
@@ -0,0 +1,24 @@
...
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+
+
+def extract_index_or_alias(response, name):
+    index = response.get(name, None)
+
+    if index:
+        return index
```
jedrazb, line 12: A bit more concise, since we don't reuse `index` anywhere
```suggestion
    if name in response:
        return response[name]
```

connectors/es/utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+    for idx_name in response.keys():
+        index_definition = response.get(idx_name)
+
+        if not index_definition.get("aliases", None):
+            continue
+
+        for alias in index_definition.get("aliases"):
+            if name == alias:
+                return index_definition
```
jedrazb, line 22: I think this can be more concise
```suggestion
 for index_name, index_definition in response.items():
       # make it shorter without local vars, get aliases just once
      aliases = index_definition.get("aliases", {})
      if name in aliases:
            return index_definition
```

connectors/es/utils.py
```diff
@@ -0,0 +1,24 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+
+
+def extract_index_or_alias(response, name):
```
jedrazb, line 8: can we add docstrings what s the purpose of this function, it took me a while to digest

jedrazb: maybe rename `response` to `get_index_response` just to make clear what we are dealing with here?

connectors/es/management_client.py
```diff
@@ -63,30 +63,29 @@ async def create_content_index(self, search_index_name, language_code):
             )
         )
 
-    async def ensure_content_index_mappings(self, index, mappings):
+    async def ensure_content_index_mappings(self, index_name, index, desired_mappings):
```
jedrazb, line 66: we have both `index` and `index_name` in args? 

can we add docstrings that `index_name` is `str` (and it can be an alias?) 
and `index` is `object` representaiton of index

jedrazb (commented):
LGTM. My main concern is that the fact we support both aliases and indexes is somewhat hidden in the fine print of our codebase.

Could we add explicit logs to indicate when we’re accessing an index via an alias and when we’re not? I’m just worried about a situation where the index-alias relationship gets messed up, and it’s hard to tell just by looking at the logs.

Should we log something in `extract_index_or_alias` that we detected alias relationship?

connectors/es/management_client.py
artem-shelkovnikov: `index_name` is just a name and `index` its definition (mappings/settings).

It's not the best, but it follows the same logic another method in the class has:

```
    async def ensure_content_index_settings(
        self, index_name, index, language_code=None
    ):
```

We can refactor/rename, but in a different PR. I'd lazily keep it as is - methods are small and stable

connectors/es/utils.py
artem-shelkovnikov: I've done a full refactoring of this logic instead :)

connectors/es/management_client.py
```diff
@@ -225,15 +224,43 @@ async def index_exists(self, index_name):
...
+
+        for (
+            existing_index_name,
+            existing_index_definition,
+        ) in get_index_response.items():
+            if "aliases" not in existing_index_definition:
+                continue
+
+            if index_name not in existing_index_definition["aliases"]:
+                continue
```
jedrazb, line 259: Nit: it feels not ideal that we explicitly `continue` if not index name not found in aliases, instead we could check if index_name is in aliases and immediately return the definition? 

That would be clearer when and what is returned IMO but nothing blocking

jedrazb (approved):

connectors/es/management_client.py
artem-shelkovnikov: Sorry I don't understand the difference - it's same approach but slightly moving pieces around without any real difference?

jedrazb: Yeah, it’s mainly about readability, but that’s pretty subjective. Nothing blocking

artem-shelkovnikov: It's a simple change, so I just went ahead and did it :)

jedrazb (approved):

PR #2804, opened 2024-09-05
add back release pipeline

seanstory:
## Closes https://github.com/elastic/search-team/issues/8226

This PR does a few things.
1. Adds back the scripts that used to exist for triggering a manual release
2. Introduces a `MANUAL_RELEASE` env var (through the pipeline) that gates some special versioning logic
3. That logic (in `publish-common.sh`) will re-compute the VERSION by going to the most recent patch version, and adding a `+build<timestamp>` suffix (this matches the pattern that Agent is establishing)
4. The pipeline adds a setup and a cleanup group, which actually commit the VERSION file change and create a git tag on that commit. This ensures that the docker tag, git tag, and `bin/elastic-ingest --version` all are in agreement.
5. every part of this pipeline is gated, and can only be run off of `X.Y` branches (not `main`, not other dev branches)

#### Pre-Review Checklist

- [ ] added automated committing and tagging to a buildkite pipeline

## Related Pull Requests

* https://github.com/elastic/ci/pull/3327 gave the elastic CI bot push access

### DIFF
--- a/.buildkite/publish/publish-common.sh
+++ b/.buildkite/publish/publish-common.sh
@@ -24,6 +24,9 @@ if [[ "${USE_SNAPSHOT:-}" == "true" ]]; then
 fi
 
 if [[ "${MANUAL_RELEASE:-}" == "true" ]]; then
+  # This block is for out-of-band releases, triggered by the release-pipeline
+  # See discussion in https://github.com/elastic/connectors/pull/2804/commits/d27e4c18650bc2dfd099018080fe16ad307eb18b#r1758850508
+  # See also RELEASING.md
   export ORIG_VERSION=$(buildkite-agent meta-data get orig_version)
   IFS='.' read -ra VERSION_PARTS <<< "$ORIG_VERSION"
   PATCH_PART=${VERSION_PARTS[2]}
--- a/.buildkite/publish/publish-common.sh
+++ b/.buildkite/publish/publish-common.sh
@@ -25,3 +25,5 @@ export BASE_TAG_NAME=${DOCKER_IMAGE_NAME:-docker.elastic.co/enterprise-search/el
 export DOCKERFILE_PATH=${DOCKERFILE_PATH:-Dockerfile}
 export PROJECT_NAME=${PROJECT_NAME:-elastic-connectors}
 export DOCKER_ARTIFACT_KEY=${DOCKER_ARTIFACT_KEY:-${PROJECT_NAME}-docker}
+export VAULT_ADDR=${VAULT_ADDR:-https://vault-ci-prod.elastic.dev}
+export VAULT_USER="docker-swiftypeadmin"
--- a/.buildkite/publish/publish-common.sh
+++ b/.buildkite/publish/publish-common.sh
@@ -13,12 +13,29 @@ export SCRIPT_DIR="$CURDIR"
 export BUILDKITE_DIR=$(realpath "$(dirname "$SCRIPT_DIR")")
 export PROJECT_ROOT=$(realpath "$(dirname "$BUILDKITE_DIR")")
 
+source $SCRIPT_DIR/git-setup.sh
+
 VERSION_PATH="$PROJECT_ROOT/connectors/VERSION"
 export VERSION=$(cat $VERSION_PATH)
 
 if [[ "${USE_SNAPSHOT:-}" == "true" ]]; then
   echo "Adding SNAPSHOT labeling"
   export VERSION="${VERSION}-SNAPSHOT"
+fi
+
+if [[ "${MANUAL_RELEASE:-}" == "true" ]]; then
+  export ORIG_VERSION=$(buildkite-agent meta-data get orig_version)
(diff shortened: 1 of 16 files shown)


seanstory:
Tada got it working! https://buildkite.com/elastic/connectors-docker-build-publish/builds/48

seanstory:
Adding Productivity team as optional reviewers, since this is so much buildkite and bash, and y'all are the subject matter experts here.

seanstory:
This is ready for review

.buildkite/publish/publish-common.sh
```diff
@@ -21,7 +23,24 @@ if [[ "${USE_SNAPSHOT:-}" == "true" ]]; then
...
+else
+  export DOCKER_TAG_VERSION=${VERSION}
+fi
+
 export BASE_TAG_NAME=${DOCKER_IMAGE_NAME:-docker.elastic.co/enterprise-search/elastic-connectors}
 export DOCKERFILE_PATH=${DOCKERFILE_PATH:-Dockerfile}
 export PROJECT_NAME=${PROJECT_NAME:-elastic-connectors}
 export DOCKER_ARTIFACT_KEY=${DOCKER_ARTIFACT_KEY:-${PROJECT_NAME}-docker}
+export VAULT_ADDR=${VAULT_ADDR:-https://vault-ci-prod.elastic.dev}
+export VAULT_USER="docker-swiftypeadmin"
```
oli-g, line 46: Swiftype is all over the place! 😅

oli-g (commented):
Reviewing this PR would be tremendously easier for rubber-stampers that didn't fully follow all the sage here, if we just have one commit that adds back the original files that were deleted (as-is, with no changes), and then add a commit with just the required changes to those files.

Would that be possible to have? If not, I think any rubber-stamps here would not have big value: and it does not make sense to you, to wait for low-value reviews, and be blocked to merge this until you get a rubber-stamp.

What do you think @seanstory? I'm also happy to just leave this PR as is, and give you my green rubber-stamp, if needed.

.buildkite/publish/publish-common.sh
```diff
@@ -21,7 +23,24 @@ if [[ "${USE_SNAPSHOT:-}" == "true" ]]; then
   export VERSION="${VERSION}-SNAPSHOT"
 fi
 
+if [[ "${MANUAL_RELEASE:-}" == "true" ]]; then
+  export ORIG_VERSION=$(buildkite-agent meta-data get orig_version)
+  IFS='.' read -ra VERSION_PARTS <<< "$ORIG_VERSION"
+  PATCH_PART=${VERSION_PARTS[2]}
+  LAST_PATCH="$((PATCH_PART - 1))"
+  LAST_VERSION="${VERSION_PARTS[0]}.${VERSION_PARTS[1]}.${LAST_PATCH}"
```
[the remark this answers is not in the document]
seanstory: > What if we need to manually publish 8.16.3, but the VERSION file has been already bumped to 8.17.0? 

This shouldn't ever happen. As soon as we Feature Freeze for an upcoming `.0` release, that minor gets its own branch. So as we prep to release 8.16.0 next month (FF mid Oct), we'll create an `8.16` branch. And we wouldn't ever be bumping the `8.16` branch's `VERSION` file to `8.17.0` - that will happen _usually_ on `main`, but for `8.17.0` specifically it'll happen on `8.x` (because of the prep for 9.0.0 major bump).

Also, this manual release job will be ONLY for out-of-band releases. The Unified Release build will handle `8.16.3`. This would be if we were preparing for an 8.16.3 Elastic-wide, but had a sudden, urgent need to release a bugfix a week before the planned release date. So if our `VERSION` file was already at `8.16.3`, we'd calculate `LAST_VERSION` as `8.16.2`, and then append the `build<timestamp>` suffix.

I'll take a note to add some more comments in here, so that future reviewers can at least look back to this PR discussion.

.buildkite/release-pipeline.yml
```diff
@@ -68,6 +93,20 @@ steps:
           - "mkdir -p .artifacts"
           - buildkite-agent artifact download '.artifacts/*.tar.gz*' .artifacts/ --step build_docker_image_arm64
           - ".buildkite/publish/test-docker.sh"
+
+  - group: "Release repo cleanup"
```
oli-g, line 97: Do you mean `Release branch cleanup` here? Or just `Release cleanup` (as we have `Release setup`)?

oli-g (approved):
Thank you for the explanation, I guess I just forgot to consider release branches, when I asked my original question. And also, thank you for cleaning up the Git history in this PR, it helped the review a lot.

.buildkite/release-pipeline.yml
seanstory: `Release cleanup` would be more consistent. The intent is to make sure that the repo/branch isn't left in a weird state, even if the tests or packaging fail. I don't want to leave the VERSION file as `8.16.0-build2024111012` on the HEAD of the `8.16` branch 😅

oli-g (approved):

PR #2808, opened 2024-09-06
Config option to disable connector secrets for native connectors

jedrazb:
## Closes https://github.com/elastic/search-team/issues/8229

For Elastic-managed connectors in agentless we need to disable connector secrets on a service-level. 

I added new "internal" option `_use_native_connector_api_keys` that **is not documented** in `config.py` , similar to how we handle the `_force_allow_native` flag in: https://github.com/elastic/connectors/blob/4e97330803087151952debbdfa35408ff240666d/connectors/preflight_check.py#L192

Let me know if this approach work for you!

### Verification

I ran the setup e2e and verified that this allows native connectors to run without connector secrets. If flag is not provided to service config the logic defaults it to `True` which is backward compatible option.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sync_job_runner.py
+++ b/connectors/sync_job_runner.py
@@ -10,9 +10,7 @@ import elasticsearch
 from elasticsearch import (
     AuthorizationException as ElasticAuthorizationException,
 )
-from elasticsearch import (
-    NotFoundError as ElasticNotFoundError,
-)
+from elasticsearch import NotFoundError as ElasticNotFoundError
 
 from connectors.config import DataSourceFrameworkConfig
 from connectors.es.client import License, with_concurrency_control
@@ -178,6 +176,7 @@ class SyncJobRunner:
             if (
                 self.connector.native
                 and self.connector.features.native_connector_api_keys_enabled()
+                and self.service_config.get("_use_native_connector_api_keys", True)
             ):
                 # Update the config so native connectors can use API key authentication during sync
                 await self._update_native_connector_authentication()
@@ -380,9 +379,11 @@ class SyncJobRunner:
             sync_cursor = (
                 None
                 if not self.data_provider  # If we failed before initializing the data provider, we don't need to change the cursor
-                else self.data_provider.sync_cursor()
-                if self.sync_job.is_content_sync()
-                else None
+                else (
+                    self.data_provider.sync_cursor()
+                    if self.sync_job.is_content_sync()
+                    else None
+                )
             )
             await self.connector.sync_done(
                 self.sync_job if await self.reload_sync_job() else None,
--- a/tests/test_sync_job_runner.py
+++ b/tests/test_sync_job_runner.py
@@ -1005,7 +1005,7 @@ async def test_native_connector_sync_fails_when_api_key_secret_missing(
     ],
 )
 @pytest.mark.asyncio
-async def test_native_sync_runs_with_secrets_disabled(
+async def test_native_sync_runs_with_secrets_disabled_when_no_permissions(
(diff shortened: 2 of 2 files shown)


tests/test_sync_job_runner.py
```diff
@@ -88,6 +88,7 @@ def create_runner(
     index_name=SEARCH_INDEX_NAME,
     sync_cursor=SYNC_CURSOR,
     connector=None,
+    service_config=None,
```
navarone-feekery, line 91:
```suggestion
    service_config={},
```

Does this work? If so, you could remove the assigning on L117.

artem-shelkovnikov (commented):
Which api key will be used in this case? The one that's specified in the config, right?

Can we have a test that verifies which api key will be used for the sync?

navarone-feekery (commented):
I like this approach, nice! Just left one comment

tests/test_sync_job_runner.py
jedrazb: Ohohohoh, this is a very bad practice :D

jedrazb: Linking an example: https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/

navarone-feekery: I didn't know D:

connectors/sync_job_runner.py
```diff
@@ -380,9 +379,11 @@ async def _sync_done(self, sync_status, sync_error=None):
...
                 None
                 if not self.data_provider  # If we failed before initializing the data provider, we don't need to change the cursor
-                else self.data_provider.sync_cursor()
-                if self.sync_job.is_content_sync()
-                else None
+                else (
+                    self.data_provider.sync_cursor()
+                    if self.sync_job.is_content_sync()
+                    else None
+                )
```
gkousouris, line 386: can simplify this to:
```
sync_cursor = (
    self.data_provider.sync_cursor()
    if self.data_provider and self.sync_job.is_content_sync()
    else None
)
```

jedrazb:
> Which api key will be used in this case? The one that's specified in the config, right?

@artem-shelkovnikov yeah the default authentication method (whatever comes from the config) will be used. 

This is the result of the fact that we don't call the `_update_native_connector_authentication` function in the if statement

https://github.com/elastic/connectors/blob/4e97330803087151952debbdfa35408ff240666d/connectors/sync_job_runner.py#L183

Which in turns overrides es config api_key used for authentication. 

https://github.com/elastic/connectors/blob/4e97330803087151952debbdfa35408ff240666d/connectors/sync_job_runner.py#L248-L251

I think it should be ok just to add extra assertion that `_update_native_connector_authentication` never gets called?

seanstory:
Any reason why we shouldn't just remove all vestiges of connector secrets in 9.x? Why make it configurable instead of just removing all that code.

jedrazb:
> Any reason why we shouldn't just remove all vestiges of connector secrets in 9.x? Why make it configurable instead of just removing all that code?

@seanstory This isa a minimal change that unblocks us short-tem.

Connector secrets are already available in the connector service, CLI, and even through ES APIs for managing them. We should plan for a dedicated initiative to deprecate our current "secrets" for 9.x

jedrazb:
@artem-shelkovnikov I addressed your comment to verify if es_config api key is changed in https://github.com/elastic/connectors/pull/2808/commits/897e692ccf480accd127bbfee472256b5ece694d

artem-shelkovnikov (approved):
Looks good, since you've tested it end-to-end and confirmed it works

PR #2809, opened 2024-09-06
Update approach to logging in connector component

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/2792

Multiple logging-related changes and a small bugfix bundled together:

1. Now Agent Python component and Connectors hosted by it will log in ECS format. This will make agent/fleet parse logs better and provide better UX (see screenshots later)
2. Logging is actually consistently implemented in the agent component. **There's a root logger and each file creates a child logger. Changes for root logger are automatically bubbled to child loggers which is nice**
3. Added logging into all agent python component files
4. Python Agent Component now reacts to changes in log_level and properly bubbles log level changes to connectors as well
5. Fixed a bug when you change only log_level for fleet policy for the connector and it corrupts the configuration

Screenshots of the change in the Fleet UI:

image

image

#### Pre-Review Checklist

### DIFF
--- a/connectors/agent/config.py
+++ b/connectors/agent/config.py
@@ -3,6 +3,8 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+import base64
+
 from connectors.config import add_defaults
 
 
@@ -24,6 +26,9 @@ class ConnectorsAgentConfigurationWrapper:
         """
         self._default_config = {
             "_force_allow_native": True,
+            "service": {
+                "_use_native_connector_api_keys": False,
+            },
             "native_service_types": [
                 "azure_blob_storage",
                 "box",
@@ -74,7 +79,12 @@ class ConnectorsAgentConfigurationWrapper:
             }
 
             if source.fields.get("api_key"):
-                es_creds["api_key"] = source["api_key"]
+                api_key = source["api_key"]
+                # if beats_logstash_format we need to base64 the key
+                if ":" in api_key:
+                    api_key = base64.b64encode(api_key.encode()).decode()
+
+                es_creds["api_key"] = api_key
             elif source.fields.get("username") and source.fields.get("password"):
                 es_creds["username"] = source["username"]
                 es_creds["password"] = source["password"]
--- a/connectors/agent/config.py
+++ b/connectors/agent/config.py
@@ -76,13 +76,23 @@ class ConnectorsAgentConfigurationWrapper:
         the method returns False.
         """
 
+        source = unit.config.source
+
         # TODO: find a good link to what this object is.
         has_hosts = source.fields.get("hosts")
         has_api_key = source.fields.get("api_key")
         has_basic_auth = source.fields.get("username") and source.fields.get("password")
+
+        assumed_configuration = {}
+
+        # Log-related
+        assumed_configuration["service"] = {}
+        assumed_configuration["service"]["log_level"] = unit.log_level
+
+        # Auth-related
         if has_hosts and (has_api_key or has_basic_auth):
             es_creds = {
-                "host": source["hosts"][0],
+                "host": source["hosts"][0]
             }
 
             if source.fields.get("api_key"):
--- a/connectors/agent/config.py
+++ b/connectors/agent/config.py
@@ -31,11 +31,9 @@ class ConnectorsAgentConfigurationWrapper:
         self._default_config = {
             "service": {
                 "log_level": "INFO",
-            },
-            "_force_allow_native": True,
-            "service": {
                 "_use_native_connector_api_keys": False,
             },
+            "_force_allow_native": True,
             "native_service_types": [
                 "azure_blob_storage",
                 "box",
@@ -178,8 +176,8 @@ class ConnectorsAgentConfigurationWrapper:
         """
         # First take "default config"
         config = self._default_config.copy()
-        # Then override with what we get from Agent
-        config.update(self.specific_config)
+        # Then merge with what we get from Agent
+        config = dict(add_defaults(self.specific_config, default_config=config))
         # Then merge with default connectors config
         configuration = dict(add_defaults(config))
 
--- a/connectors/agent/config.py
+++ b/connectors/agent/config.py
@@ -91,9 +91,7 @@ class ConnectorsAgentConfigurationWrapper:
 
         # Auth-related
         if has_hosts and (has_api_key or has_basic_auth):
-            es_creds = {
-                "host": source["hosts"][0]
-            }
+            es_creds = {"host": source["hosts"][0]}
 
             if source.fields.get("api_key"):
                 logger.debug("Found api_key")
--- a/connectors/agent/config.py
+++ b/connectors/agent/config.py
@@ -125,6 +125,13 @@ class ConnectorsAgentConfigurationWrapper:
         return False
 
     def config_changed(self, new_config):
+        """See if configuration passed in new_config will update currently stored configuration
+
+        This method takes the new configuration received from the agent and see if there are any changes
+        to existing configuration.
+
+        If new_config contains new values for relevant fields, then True is returned, otherwise it returns False.
+        """
         # TODO: For now manually check, need to think of a better way?
         # Not super proud of this function, but hey it's tested
         logger.debug("Checking if config changed")
@@ -150,9 +157,11 @@ class ConnectorsAgentConfigurationWrapper:
             )
 
         if _log_level_changed():
+            logger.debug("log_level changed")
             return True
 
         if _elasticsearch_config_changed():
+            logger.debug("elasticsearch changed")
             return True
 
         return False
--- a/connectors/agent/config.py
+++ b/connectors/agent/config.py
@@ -7,6 +7,7 @@ import base64
 
 from connectors.agent.logger import get_logger
 from connectors.config import add_defaults
+from connectors.utils import nested_get_from_dict
 
 logger = get_logger("config")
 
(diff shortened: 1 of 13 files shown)


connectors/agent/config.py
```diff
@@ -64,28 +71,78 @@ def try_update(self, source):
...
         """
 
+        source = unit.config.source
+
         # TODO: find a good link to what this object is.
         has_hosts = source.fields.get("hosts")
         has_api_key = source.fields.get("api_key")
         has_basic_auth = source.fields.get("username") and source.fields.get("password")
+
+        assumed_configuration = {}
```
artem-shelkovnikov, line 81: Could not get a better name for "temporary config storage that collects config changes that we'll try to apply"

seanstory: `tmp_cfg` 😝 
I think "assumed" works great here

seanstory (dismissed):
One question and two nits. Don't feel strongly about them.

Nice stuff.

connectors/agent/config.py
```diff
@@ -64,28 +71,78 @@ def try_update(self, source):
...
+            # were received
+            self.specific_config.update(assumed_configuration)
+            return True
+
+        logger.debug("No changes detected for connectors-relevant configurations")
+        return False
+
+    def config_changed(self, new_config):
+        # TODO: For now manually check, need to think of a better way?
+        # Not super proud of this function, but hey it's tested
```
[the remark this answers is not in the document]
artem-shelkovnikov: I have a feeling there's a better way, but recursive dict traversals are so hard to read and debug 🤷 

I guess we'll just have it like this until it bites us haha

connectors/agent/logger.py
```diff
@@ -0,0 +1,29 @@
...
+handler.setFormatter(ecs_logging.StdlibFormatter())
+root_logger.addHandler(handler)
+root_logger.setLevel(logging.INFO)
+
+
+def get_logger(module):
+    logger = root_logger.getChild(module)
+
+    if logger.hasHandlers():
+        return logger
```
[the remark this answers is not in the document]
artem-shelkovnikov: It's actually a bit awkward.

This function operates on the assumption that only it can create logger and add handlers.

There's no way to get child logger and learn if it was just created, or it's already been created and we get the cached one back. 

One way people suggest is to check for some factor that can distinguish whether it was just created or not, and having handlers in our case is such a factor - if it was just created, there are no handlers and we should add them. If it has handlers, this file has created it and correct handlers should already be assigned.

I'm not super happy with it, I'm up to follow up with #python channel about it.

connectors/agent/service_manager.py
```diff
@@ -3,13 +3,19 @@
...
+import connectors.agent.logger
+import connectors.logger
+from connectors.agent.logger import get_logger
 from connectors.services.base import (
     ServiceAlreadyRunningError,
     get_services,
 )
 from connectors.utils import CancellableSleeps
 
+logger = get_logger("component")
```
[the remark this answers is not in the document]
artem-shelkovnikov: Oh whoops it's actually a mistake - this should be different in all files 😓 Every logger should have a unique name, great catch!

tests/agent/test_agent_config.py
```diff
@@ -3,53 +3,118 @@
...
-
-import pytest
+from unittest.mock import MagicMock, Mock
 
 from connectors.agent.config import ConnectorsAgentConfigurationWrapper
 
 
-@pytest.mark.asyncio
-async def test_try_update_without_auth_data():
+def prepare_unit_mock(fields, log_level):
```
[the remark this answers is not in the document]
artem-shelkovnikov: Python confuses me with this one - called it like that cause an object is of actual `Mock` type :D

connectors/agent/logger.py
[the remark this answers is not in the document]
artem-shelkovnikov: https://stackoverflow.com/questions/53129716/how-to-check-if-a-logger-exists

jedrazb:
For some reason python elastic agent logs have `error` level

Screenshot 2024-09-20 at 13 59 07

artem-shelkovnikov:
@jedrazb this gets fixed with this change: https://github.com/elastic/python-elastic-agent-client/pull/62

You can update `requirements.txt` to point to my branch artem/update-component-logger for the python elastic agent and see it yourself :)

connectors/agent/config.py
```diff
@@ -25,6 +29,9 @@ def __init__(self):
         Service config and specific config coming from Agent.
         """
         self._default_config = {
+            "service": {
```
jedrazb, line 32: This adds a duplicate entry `service` to `_default_config`, I will merge those

connectors/agent/config.py
```diff
@@ -69,33 +76,92 @@ def try_update(self, source):
...
+
         # TODO: find a good link to what this object is.
         has_hosts = source.fields.get("hosts")
         has_api_key = source.fields.get("api_key")
         has_basic_auth = source.fields.get("username") and source.fields.get("password")
+
+        assumed_configuration = {}
+
+        # Log-related
+        assumed_configuration["service"] = {}
```
jedrazb, line 89: I think it clears the `service._use_native_connector_api_keys` setting

jedrazb: The root cause is how we were merging the two dicts with default python `update` https://github.com/elastic/connectors/blob/4f6ca9bb4d9d05f8c400252d717088a7734f4b99/connectors/agent/config.py#L116

This would override the whole `service` object from the default config stored in this the `agent/config.py` class (so this default flag `service._use_native_connector_api_keys` would be cleared)

I'm fixing this in this PR

connectors/agent/service_manager.py
```diff
@@ -59,26 +59,34 @@ async def run(self):
...
                         ["schedule", "sync_content", "sync_access_control", "cleanup"],
-                        self._agent_config.get(),
+                        config,
                     )
+                    log_level = config.get("service", {}).get(
+                        "log_level", logging.INFO
+                    )  # Log Level for connectors is managed like this
+                    connectors.logger.set_logger(log_level, filebeat=True)
+                    # Log Level for agent connectors component itself
+                    connectors.agent.logger.update_logger_level(log_level)
```
seanstory, line 73: It looks like we're only actually changing our log level based on what we're sent at startup, but I don't see anything in `protocol.py` or `config.py` that actually respects a changed log level later. Am I missing it? I would expect this logic here to be part of the checkin handler.

tests/agent/test_agent_config.py
```diff
@@ -3,71 +3,150 @@
...
+    assert config_wrapper.config_changed(new_config) is True
+
+
+def test_config_changed_when_log_level_config_changed():
+    config_wrapper = ConnectorsAgentConfigurationWrapper()
+    config_wrapper.try_update(prepare_unit_mock({}, "INFO"))
+
+    new_config = {"service": {"log_level": "DEBUG"}}
+
+    assert config_wrapper.config_changed(new_config) is True
```
seanstory, line 143: Might be interesting to modify this test or add another to actually assert that the log level does change. That you get new log statements after the new config is passed.

connectors/agent/service_manager.py
jedrazb: > It looks like we're only actually changing our log level based on what we're sent at startup, but I don't see anything in protocol.py or config.py that actually respects a changed log level later. 

The service manager allows the connector service to be restarted (see the `outer while True` loop) - the connector service will restart and it will take whatever the config was updated to (including the updated log level). This is the same "loop" in which we are getting the updated config with ES credentials upon fleet enrolment. 

See the place we actually trigger the restart (this PR changes the `try_update` logic to also trigger for logging change) https://github.com/elastic/connectors/blob/4f6ca9bb4d9d05f8c400252d717088a7734f4b99/connectors/agent/protocol.py#L72

The try_update takes the checkin payload from [UnitExpected](https://github.com/elastic/elastic-agent-client/blob/86f524ddc40d8aae5a01244ae223fb87f8b33265/elastic-agent-client.proto#L186) protobuf object (that contains both config and log_level)

jedrazb: You can modify the agent logging level in the logs tab at the very bottom:
Screenshot 2024-09-24 at 11 39 52



I performed 2 syncs: one with debug one with info level, the log level changes work as expected:
Screenshot 2024-09-24 at 11 58 23

jedrazb: It somehow worked before (was agent process setting some env variable for us?) .... I detect the log level changes explicitly as a part of this PR: https://github.com/elastic/python-elastic-agent-client/pull/73

jedrazb: I performed a couple of syncs, the python component logs with expected level (see variance in info, debug logs). In the last interval I set level to `error` and we see no logs:

Screenshot 2024-09-24 at 12 37 24

tests/agent/test_agent_config.py
jedrazb: Added more comprehensive test in https://github.com/elastic/connectors/pull/2809/commits/9d14acd7f470f203cf49d0ebeed96e9ea3d0b2b2 it tests if log level from unit (from agent protocol) correctly gets propagated to the resulting configuration

connectors/agent/service_manager.py
seanstory: > the connector service will restart and it will take whatever the config was updated to (including the updated log level)

I wouldn't expect a restart to be required to pick up a log level change, when running through Agent. I know that's needed with connectors today, because we only read from config.yml once on startup, but for this, I'd expect a checkin to change the log level, but not interrupt anything in the process or syncs. Am I misunderstanding? Regardless, that's out of the scope of this PR, so not blocking, but let's be sure to revisit.

> I performed a couple of syncs, the python component logs with expected level (see variance in info, debug logs). In the last interval I set level to error and we see no logs

Fantastic! Thanks for verifying.

seanstory (approved):

PR #2812, opened 2024-09-10
Mass backport into 8.x branch

seanstory:
Main is still on 8.16, as https://github.com/elastic/connectors/pull/2800 has not yet been able to build. But DRAs for 8.16 are being built now from 8.x branches, not main branches. We need to backport everything. 

I'm going to mass-backport this, then set up main to auto-backport to 8.x until we can merge the 9.0 version bump PR.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,2 @@
 [report]
-omit = connectors/quartz.py,connectors/conftest.py,tests/*
+omit = connectors/quartz.py,connectors/conftest.py,tests/*,connectors/agent/*,connectors/cli/*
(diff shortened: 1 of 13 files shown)


jedrazb (approved):
You just rebased on main rather than backporting manually right?

artem-shelkovnikov (approved):

seanstory:
> You just rebased on main rather than backporting manually right?

I went to the "pull requests" tab, and compared `main` going into `8.x` and then clicked "create PR" 😅 
Seemed easier

artem-shelkovnikov:
We'll lose some history, unfortunately, but might not be too important

seanstory:
We shouldn't have lost any history. All the commits should still be there, and still link to their PRs. Right?

PR #2815, opened 2024-09-11
Add everything needed to have a dockerfile for agent testing

artem-shelkovnikov:
## Part of https://github.com/elastic/search-team/issues/8211

This change provides a way to run docker image of agent with connectors component installed.

It's used purely for dev/testing purposes, should not be used for anything else.

When we finish the work with Agent, this docker image might change significantly or will be deleted.

#### Pre-Review Checklist

### DIFF
--- a/Dockerfile.agent
+++ b/Dockerfile.agent
@@ -1,0 +1,33 @@
+FROM docker.elastic.co/elastic-agent/elastic-agent:sha256-fff74ac535403af9307ed6f93aad71ec95866bf6f1ec2da1bd7966e6bc17ae72
+USER root
+# Install dependencies
+RUN apt update
+RUN apt install software-properties-common -y
+RUN add-apt-repository ppa:deadsnakes/ppa
+RUN apt install python3.11 python3.11-venv make -y
+
+# TEMPORARY STUFF
+# I need vim to edit some fields
+# Git is needed to pull connectors repo
+# yq is needed to append our input to elastic-agent.yml
+RUN add-apt-repository ppa:rmescandon/yq
+RUN apt install vim git yq -y
+
+# Copy and install python agent client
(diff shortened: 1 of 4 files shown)


Dockerfile.agent
```diff
@@ -0,0 +1,33 @@
+FROM docker.elastic.co/elastic-agent/elastic-agent:sha256-fff74ac535403af9307ed6f93aad71ec95866bf6f1ec2da1bd7966e6bc17ae72
```
seanstory, line 1: Let's add a comment or something at the top of this, so that it's clear to anyone looking at it that this is for internal/experimental purposes only.

seanstory (dismissed):
LGTM, but in the long run, would be interesting to brainstorm with Platform folks to see if we can somehow build the agentless image on-demand. Seems like that's nearly identical to what this is doing, right? Like maybe steps of:
1. make local connector changes
2. `git commit && git push origin/my-branch`
3. clone elastic-agent
4. `FETCH_CONNECTORS_FROM_BRANCH=my-branch mage build-agentless-image`
5. `docker run new-local-agentless-image-tag`

CC @blakerouse

Dockerfile.agent
blakerouse: The real `elastic-agent-service` container is based off of Wolfi, best to use `elastic-agent-wolfi` here.

artem-shelkovnikov: `elastic-agent-wolfi` does not seem to have `apt` and a couple other things installed.

Since this file is just for local development (which is also probably a temporary file, we'll find something more suitable hopefully), I think it's fine to use non-wolfi image.

I'm gonna update this line to point to latest 9.0.0-SNAPSHOT docker image

seanstory (approved):

PR #2825, opened 2024-09-13
Add retry for on-commit and nightly pipelines

artem-shelkovnikov:
Adding retries to our buildkite pipelines:
- commit/main
- Nightly

For now we'll retry 2 times for signals -1, 255 and 2.

See https://buildkite.com/blog/retrying-ci-cd-steps-when-spot-instances-terminate for docs


seanstory (approved):
lgtm

PR #2839, opened 2024-09-18
Minor fixes for agent component

artem-shelkovnikov:
## Closes https://github.com/elastic/search-team/issues/8211

Bundles two changes:

- Makes it possible to run the component with fleet. This is achieved by base64 encoding the api key if it's received in a format "a:b"
- Set `service._use_native_connector_api_keys` to `False` to make the service run without using native api keys and use agent/fleet api key instead

#### Pre-Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ docker-push:
 
 ## Agent Docker Zone
 # Only use it for local testing, that's it
-AGENT_ES_HOSTS?=http://127.0.0.1:9200
+AGENT_ES_HOSTS?=[http://127.0.0.1:9200]
 AGENT_ES_USERNAME?=elastic
 AGENT_ES_PASSWORD?=changeme
 AGENT_DOCKERFILE_NAME?=Dockerfile.agent
--- a/connectors/agent/config.py
+++ b/connectors/agent/config.py
@@ -3,6 +3,8 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+import base64
+
 from connectors.config import add_defaults
 
 
@@ -24,6 +26,9 @@ class ConnectorsAgentConfigurationWrapper:
         """
         self._default_config = {
             "_force_allow_native": True,
+            "service": {
+                "_use_native_connector_api_keys": False,
+            },
             "native_service_types": [
                 "azure_blob_storage",
                 "box",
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -96,7 +96,7 @@ docker-push:
 
 ## Agent Docker Zone
 # Only use it for local testing, that's it
-AGENT_ES_HOSTS?=http://127.0.0.1:9200
+AGENT_ES_HOSTS?=[http://127.0.0.1:9200]
```
seanstory, line 99: my "make" knowledge is weak - is this making it an array? I usually assume that make syntax is the same as bash (I know it's not), but this would just be a string that starts with `[` and ends with `]`

connectors/agent/config.py
```diff
@@ -74,7 +79,12 @@ def try_update(self, source):
             }
 
             if source.fields.get("api_key"):
-                es_creds["api_key"] = source["api_key"]
+                api_key = source["api_key"]
+                # if beats_logstash_format we need to base64 the key
+                if ":" in api_key:
+                    api_key = base64.b64encode(api_key.encode()).decode()
```
seanstory, line 85: I thought that @jedrazb had discovered that encode/decode isn't necessary - you actually want to just grab the bit after the colon?

seanstory (approved):
rubber stamping, in case both my comments are obviously nonsense and not concerning

Makefile
artem-shelkovnikov: It's still just a string, but when it's passed inside the docker image it is embedded into a yaml file as is, and yaml recognises this as an array :)

So multi-level magic!

connectors/agent/config.py
artem-shelkovnikov: Checked with Jedr:

encode/decode are for string operations here.
`base64.b64encode` wants a byte array, so we `encode` and returns a byte array, so we `decode` to get a string out of it

jedrazb: > I thought that @jedrazb had discovered that encode/decode isn't necessary - you actually want to just grab the bit after the colon?

The colon indicates that the API key we’re receiving is in an unencoded `beats_logstash_format`, whereas the ES client expects it to be base64-encoded. Therefore, we need to handle the encoding ourselves to authenticate with the ES client.

IMO checking for the presence of a `:` is an "acceptable" way to detect the API key format since `:` isn’t a character used in base64-encoded strings.

PR #2844, opened 2024-09-23
[Outlook] Feature: Office365 multi-user support

### LINKED ISSUE #2843: Enhancement: Allow Outlook Cloud connector to specify client emails instead of defaulting to top 999 users
### Problem Description
The current Outlook Cloud connector attempts to fetch the top 999 users by default. This can lead to performance issues and unnecessary load, especially in environments where only a few client emails are relevant for the integration.

### Proposed Solution
Introduce a configuration option in the Outlook Cloud connector to specify client emails (as a comma-separated list) to fetch only the relevant user data. This change will enhance performance by reducing the number of users fetched and improve the connector's efficiency.

ilyasabdellaoui:
Introduced BaseOffice365User abstract base class to standardize Office 365 user handling. Added MultiOffice365Users to manage multiple emails from config. Added client_emails (comma-separated) in OutlookDataSource config. Resolved issue with fetching too many users causing SMTP server not found error.

#### Pre-Review Checklist


connectors/sources/outlook.py
```diff
@@ -735,6 +823,13 @@ def get_default_configuration(cls):
                 "sensitive": True,
                 "type": "str",
             },
+            "client_emails": {
+                "depends_on": [{"field": "data_source", "value": OUTLOOK_CLOUD}],
+                "label": "Client Email Addresses (comma-separated)",
+                "order": 5,
```
seanstory, line 829: Because `exchange_server` is currently `"order": 5`, you'll need to bump that to 6, and `active_director_server` to 7, etc.

connectors/sources/outlook.py
```diff
@@ -735,6 +823,13 @@ def get_default_configuration(cls):
                 "sensitive": True,
                 "type": "str",
             },
+            "client_emails": {
+                "depends_on": [{"field": "data_source", "value": OUTLOOK_CLOUD}],
+                "label": "Client Email Addresses (comma-separated)",
```
seanstory, line 828: We'd probably want a tooltip, to explain why you'd want to set this.

tests/sources/test_outlook.py
```diff
@@ -597,13 +624,49 @@ async def test_get_users_for_cloud():
...
-                side_effect=side_effect_function,
+                side_effect=side_effect_function(),
             ):
                 async for response in source.client._get_user_instance.get_users():
                     user_mails = [user["mail"] for user in response["value"]]
                     users.extend(user_mails)
                 assert users == ["test.user@gmail.com", "dummy.user@gmail.com"]
 
+    client_emails = "one.user@gmail.com"
+    async with create_outlook_source(client_emails=client_emails) as source:
```
seanstory, line 635: I suggest making these their own test functions. That way, any failures are more easily identifyable, and they can have more focused test names to help the reader understand what's being tested.

connectors/sources/outlook.py
```diff
@@ -456,6 +473,57 @@ async def get_user_accounts(self):
...
+    """Fetch multiple Office365 users based on a list of email addresses."""
+
+    def __init__(self, client_id, client_secret, tenant_id, client_emails: List[str]):
+        super().__init__(client_id, client_secret, tenant_id)
+        self.client_emails = client_emails
+
+    async def get_users(self):
+        access_token = await self._fetch_token()
+        for email in self.client_emails:
+            url = f"https://graph.microsoft.com/v1.0/users/{email}"
```
seanstory, line 486: I imagine that this is not particularly efficient if you have a large list of client emails. Did you look to see if there's a way to fetch them by email address in bulk?

connectors/sources/outlook.py
```diff
@@ -735,6 +823,13 @@ def get_default_configuration(cls):
                 "sensitive": True,
                 "type": "str",
             },
+            "client_emails": {
+                "depends_on": [{"field": "data_source", "value": OUTLOOK_CLOUD}],
+                "label": "Client Email Addresses (comma-separated)",
+                "order": 5,
+                "required": False,
+                "type": "str",
```
seanstory, line 831: I'd expect this to be a list. The framework will then do the "split on comma" work for you. See:
https://github.com/elastic/connectors/blob/6e74c3a8023f3fd80de240e5ffb446321610225f/connectors/source.py#L142-L144

connectors/sources/outlook.py
```diff
@@ -583,6 +651,27 @@ def attachment_doc_formatter(self, attachment, attachment_type, timezone):
         }
 
 
+class UserFactory:
+    """Factory class for creating Office365 user instances"""
+
+    @staticmethod
+    def create_user(configuration: dict) -> BaseOffice365User:
+        if configuration.get("client_emails"):
+            client_emails = [email.strip() for email in configuration["client_emails"].split(",")]
```
seanstory, line 660: See other comment - I think you can drop the `split(",")` logic here if you use the `list` type in the configuration field

seanstory (commented):
This is great! Thanks for submitting a patch! I've left a few comments, but overall I think this is a pretty clean approach and solves a clear ask.

Outside of the below comments, I'd also request that you change the 

> Closes https://github.com/elastic/connectors/issues/2843

in your PR description to use "Relates to" or "Part of", since  another part of the issue you filed is that the current solution is limited to 999 users. I expect that's something we can better control with the graph API and another separate configuration, for folks who want "all users", and have more than 999.

seanstory:
Screenshot 2024-09-24 at 3 45 50 PM

The tests pass, but the linter is currently failing. You can fix this by running `make autoformat` when you're done making changes.

ilyasabdellaoui:
> in your PR description to use "Relates to" or "Part of", since another part of the issue you filed is that the current solution is limited to 999 users. I expect that's something we can better control with the graph API and another separate configuration, for folks who want "all users", and have more than 999.

@seanstory After rechecking the code, I realized I was mistaken in my initial description. The code isn't limited to fetching the top 999 users—it handles pagination using the @odata.nextLink property from the Microsoft Graph API, which retrieves all users. I missed that detail. I'll update my issue description to reflect the correct behavior.
Snippet of code from line 428 to 446:
```python
async def get_users(self):
    access_token = await self._fetch_token()
    url = f"https://graph.microsoft.com/v1.0/users?$top={TOP}"
    while True:
        try:
            async with self._get_session.get(
                url=url,
                headers={
                    "Authorization": f"Bearer {access_token}",
                    "Content-Type": "application/json",
                },
            ) as response:
                json_response = await response.json()
                yield json_response
                url = json_response.get("@odata.nextLink")
                if url is None:
                    break
        except Exception:
            raise
```

connectors/sources/outlook.py
ilyasabdellaoui: 104e299
I updated the implementation to use JSON batching which processes client emails in batches of up to 20 (max supported by Graph API). I also handled exceptions and collected errors which are raised after processing all batches.

[Graph API JSON batching Documentation](https://learn.microsoft.com/en-us/graph/json-batching)

seanstory:
buildkite test this

artem-shelkovnikov:
@ilyasabdellaoui do you want to make this PR ready for review, or is there anything we can help with?

tests/sources/test_outlook.py
ilyasabdellaoui: I've added a test function to cover the batch POST request handling for fetching user data with client emails

ilyasabdellaoui:
buildkite test this

ilyasabdellaoui:
@artem-shelkovnikov Thanks for checking in! I encountered some linting issues, which I've now resolved in my latest commit, particularly related to async iteration. Everything is fixed and ready for review.

artem-shelkovnikov:
buildkite test this

connectors/sources/outlook.py
```diff
@@ -735,45 +857,53 @@ def get_default_configuration(cls):
...
                 "type": "str",
             },
+            "client_emails": {
+                "depends_on": [{"field": "data_source", "value": OUTLOOK_CLOUD}],
+                "label": "Client Email Addresses (comma-separated)",
+                "order": 5,
+                "tooltip": "Specify the email addresses to limit data fetching to specific clients. If left empty, data will be fetched for all users.",
+                "required": False,
+                "type": "list",
+            },
```
artem-shelkovnikov, line 867: We normally handle lists like this with `*` being default value and `*` meaning that everything should be exported.

See this as an example: https://github.com/elastic/connectors/blob/main/connectors/sources/sharepoint_online.py#L1192-L1199

connectors/sources/outlook.py
```diff
@@ -456,6 +485,78 @@ async def get_user_accounts(self):
...
+    """Fetch multiple Office365 users based on a list of email addresses."""
+
+    def __init__(self, client_id, client_secret, tenant_id, client_emails: List[str]):
+        super().__init__(client_id, client_secret, tenant_id)
+        self.client_emails = client_emails
+
+    async def get_users(self):
+        access_token = await self._fetch_token()
+        errors = []
+        for i in range(0, len(self.client_emails), 20):
```
artem-shelkovnikov, line 498: Would be great to make it a constant at least (20). Also why was 20 chosen as batch size?

artem-shelkovnikov (commented):
Looks good! Left a comment about configurable field to make it consistent with other connectors, but otherwise great change.

Additional question - have you done any benchmarks on how faster the connector gets?

I know that you're saving a lot of time on network, but throttling-wise I believe this change makes not impact, is it true?

connectors/sources/outlook.py
ilyasabdellaoui: I've updated the code to use the `GRAPH_API_BATCH_SIZE` constant instead of hardcoding the value `20`, see commit 
e4923b5.

The batch size of `20` is due to the max limit set by the Graph API. For ref you can check the [Graph API documentation on JSON batching](https://learn.microsoft.com/en-us/graph/json-batching#:~:text=JSON%20batch%20requests%20are%20currently%20limited%20to%2020%20individual%20requests)

connectors/sources/outlook.py
ilyasabdellaoui: I've applied the suggested changes and updated the logic to the WILDCARD `*` when client_emails is not specified 
920697a

ilyasabdellaoui:
> Looks good! Left a comment about configurable field to make it consistent with other connectors, but otherwise great change.
> 
> Additional question - have you done any benchmarks on how faster the connector gets?
> 
> I know that you're saving a lot of time on network, but throttling-wise I believe this change makes not impact, is it true?

I haven't run any benchmarks yet, but I expect this change will improve performance by reducing unnecessary user fetches.

As for throttling, the client_emails configuration is most effective at avoiding it when the list of emails is small, by limiting the API calls to only the relevant users, it reduces the load on the API. For larger email lists, though, the impact on throttling might not be as important!

artem-shelkovnikov:
buildkite test this

artem-shelkovnikov (approved):

artem-shelkovnikov:
Change looks good, but I'll not merge immediately, will address a bit later - we need to do same change in Kibana to make it properly work for native connectors. I'll follow-up and merge both at the same time

artem-shelkovnikov:
Hi @ilyasabdellaoui, it's on my radar - we currently don't have a running subscription with. Outlook to be able to manually test it, it's ongoing work and we'll get back to this PR once it's been properly tested.

ilyasabdellaoui:
Thanks for the update @artem-shelkovnikov! Let me know if I can assist with anything to facilitate testing.

PR #2850, opened 2024-09-25
[Agent] Improvements to agents protocol handler

jedrazb:
### Changes
Mostly improvements around handling output units, this PR explicitly assumes that:
- We are only dealing with `elasticsearch` output types, see supported output types, see docs https://www.elastic.co/guide/en/fleet/current/elastic-agent-output-configuration.html
- We are only sending data to the first output unit of elasticsearch type, if there are multiple ES outputs in the agent policy a warning is logged

### Verfication
- Works locally 🤌 
- Unit tests

#### Pre-Review Checklist

### DIFF
--- a/connectors/agent/protocol.py
+++ b/connectors/agent/protocol.py
@@ -84,7 +84,7 @@ class ConnectorCheckinHandler(BaseCheckinHandler):
                         "Multiple Elasticsearch outputs detected. The first ES output defined in the agent policy will be used."
                     )
 
-                logger.debug("Elasticse arch outputs found.")
+                logger.debug("Elasticsearch outputs found.")
 
                 configuration_changed = self.agent_connectors_config_wrapper.try_update(
                     elasticsearch_outputs[0]
--- a/connectors/agent/protocol.py
+++ b/connectors/agent/protocol.py
@@ -60,20 +60,39 @@ class ConnectorCheckinHandler(BaseCheckinHandler):
         agent expects the components to respond within 30 seconds.
         See comment in https://github.com/elastic/elastic-agent-client/blob/main/elastic-agent-client.proto#L29
         """
-        logger.info("There's new information for the components/units!")
+        logger.info("New information available for components/units")
+
         if self.client.units:
             logger.debug("Client reported units")
+
             outputs = [
                 unit
                 for unit in self.client.units
                 if unit.unit_type == proto.UnitType.OUTPUT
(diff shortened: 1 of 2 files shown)


connectors/agent/protocol.py
```diff
@@ -60,20 +60,39 @@ async def apply_from_client(self):
...
-                logger.debug("Outputs were found")
 
-                changed = self.agent_connectors_config_wrapper.try_update(outputs[0])
-                if changed:
-                    logger.info("Updating connector service manager config")
+            # Filter Elasticsearch outputs from the available outputs
+            elasticsearch_outputs = [
+                output
+                for output in outputs
+                if output.config and output.config.type == "elasticsearch"
```
jedrazb, line 78: We can assume type exists on config object, see proto: https://github.com/elastic/elastic-agent-client/blob/86f524ddc40d8aae5a01244ae223fb87f8b33265/elastic-agent-client.proto#L172

tests/agent/test_protocol.py
```diff
@@ -105,7 +106,84 @@ async def test_apply_from_client_when_units_with_output_and_updating_config(self
...
+
+        checkin_handler = ConnectorCheckinHandler(
+            client_mock, config_wrapper_mock, service_manager_mock
+        )
+
+        await checkin_handler.apply_from_client()
+
+        # Only ES output from the policy should be used by connectors component
+        assert config_wrapper_mock.try_update.called_once()
+        called_unit = config_wrapper_mock.try_update.call_args[0][0]
```
jedrazb, line 143: We need to explicitly check the call arguments to compare IDs. The called_once_with assertion is more lenient and doesn’t verify object properties like IDs.

connectors/agent/protocol.py
```diff
@@ -60,20 +60,39 @@ async def apply_from_client(self):
...
+            ]
+
+            if elasticsearch_outputs:
+                if len(elasticsearch_outputs) > 1:
+                    logger.warning(
+                        "Multiple Elasticsearch outputs detected. The first ES output defined in the agent policy will be used."
+                    )
+
+                logger.debug("Elasticse arch outputs found.")
```
seanstory, line 87:
```suggestion
                logger.debug("Elasticsearch outputs found.")
```

seanstory (approved):

seanstory (approved):

PR #2859, opened 2024-10-01
Update docker namespace to integrations

wangch079:
## Part of https://github.com/elastic/search-team/issues/7481

This PR updates the docker namespace to `integrations`.

#### Pre-Review Checklist


seanstory:
Figured it was easier to discuss in slack, so started a thread there, but I think we’ll want to ship to both repos for a while, not make a hard switch.

jedrazb (approved):

wangch079:
Tested with different namespaces:
integrations - https://buildkite.com/elastic/connectors-docker-build-publish/builds/52
enterprise-search - https://buildkite.com/elastic/connectors-docker-build-publish/builds/53

seanstory:
> Figured it was easier to discuss in slack, so started a thread there, but I think we’ll want to ship to both repos for a while, not make a hard switch.

It seems that this is still the case for a single job? Is the expectation then that you have to manually trigger two jobs, with different env vars?

Also, how does this work now with the connector DRAs? Since those run automatically, I'm worried this will cause them to just hard-switch to integrations, since I don't see any changes that would re-trigger them on the old namespace too.

wangch079:
>It seems that this is still the case for a single job? Is the expectation then that you have to manually trigger two jobs, with different env vars?

Yes, it's a manual process, so I tend to keep it manual, especially considering we will drop the use of namespace `enterprise-search` soon

>Also, how does this work now with the connector DRAs?

In PR https://github.com/elastic/infra/pull/41576, I couldn't add additional entry but update the existing one. From Julien:

>Images copies on different namespace is handled in the unified-release repo.

And this is tracked in https://elasticco.atlassian.net/browse/ENGPRD-874.

seanstory:
👍 thanks Chenhui - I hadn't seen that infra PR or the follow-up issue for them. In that case, all makes sense and LGTM

PR #2879, opened 2024-10-11
Introduce sync job checkpoints and implement simple checkpointing in GitHub connector

artem-shelkovnikov:
## Closes https://github.com/elastic/search-team/issues/8289

This PR is WIP: it's here to demonstrate the approach to be able to discuss it.

This PR adds support for sync checkpoints: connector reports that it handled some atomic piece of work and if sync restarts this piece of work can be skipped. GitHub connector is updated to use this approach on repo-per-repo basis: *if a repo was synced and sync gets suspended, then the sync will skip the synced repo after it's restarted.*

### Protocol

Protocol update is:
- Introduce "checkpoint" field into `.elastic-connectors-sync-jobs` index of type `object` that stores free-form checkpoints.
- (not done) Update connectors API to add checkpoint as an argument for `update_job_metadata` method OR add a separate method.

I've purposely named it different than cursor to be able to distinguish these two easier - *cursors are for transferring state between syncs, checkpoints are for keeping state within the sync.*

Checkpoints are stored in the job. When a connector is created, a new field `checkpoint` is populated with content of the field from the index and is accessible from connector via `_checkpoint` field similar to how cursors work. So during `get_docs` it's possible to call `_checkpoint`

❓ Might be worth to call it `sync_state` or `job_state` rather than a checkpoint.

### GitHub connector

The logic update is very primitive - during `get_docs` every time a repository is processed save state to the `_checkpoint`. Before a repository is processed check if it was already processed, and if so just skip to the next one

### Sink / Extractor arrangement

Fun starts here. So once a document is yielded from the connector, it's added to the queue, not sent directly to Elasticsearch. So to save a checkpoint we need to verify that the data that was extracted was already saved - otherwise we would save checkpoint *before the data is saved*. *This could cause inconsistency issues when connector is suspended before the batch is saved.*

To avoid this I've introduced the following logic:

1. *Checkpoints are updated every time job metadata is updated*, not immediately. This does not have direct impact on the inconsistency, just something to keep in mind.
2. Once the job stats tracking mechanism sees that a checkpoint is updated it calls to the `SyncOrchestrator` and says: "*forcefully ingest the current batch without waiting for batch to be generated*"
3. *Once the batch is ingested, the checkpoint is saved*

Some more technical details/things to consider:
- While data is ingested, *connector keeps extracting the data from 3rd-party*. This means that while batch is being sent, another checkpoint might be queued up for saving. We need to first save the first one, and then correctly flush data/save the second one. Current mechanism _enforces_ it, but please double check my logic here
- The way the above is implemented is SyncOrchestrator adding a new abstraction - `ForceFlushSignal` into the queue. *Once queue meets this signal it stops reading and immediately sends data to Elasticsearch.* In the meantime the owner of `ForceFlushSignal` awaits for it to be processed (see `ForceFlushSignal.trigger`). While this is happening, the logic that does job metadata/checkpoint processing is not executed! *This prevents from having concurrent flushes.*
- The above means that in certain cases *it will slow down the connector* due to some inefficiency of the process, but I believe *the slowdown will be insignificant*

#### Pre-Review Checklist

## Release Note

Framework now supports sync state: if connector implements it and sync for this connector gets suspended, then when sync is resumed it'll resume from the last checkpoint saved by a connector.

Github connector implements this logic: stores checkpoints per repository.


connectors/es/sink.py
```diff
@@ -356,33 +385,48 @@ async def _run(self):
             batch_num = 0
 
             while True:
+                force_flush = False
                 batch_num += 1
                 doc_size, doc = await self.fetch_doc()
-                if doc in (END_DOCS, EXTRACTOR_ERROR):
+                if isinstance(doc, ForceFlushSignal):
+                    self._logger.debug("Found ForceFlushSignal")
```
seanstory, line 392:
```suggestion
                    self._logger.debug("Received flush signal")
```
slightly less alarming, IMO. Just a suggestion.

connectors/es/sink.py
```diff
@@ -356,33 +385,48 @@ async def _run(self):
...
+
+                    bulk_size += doc_size
+
+                if (
+                    len(batch) >= self.chunk_size
+                    or bulk_size > self.chunk_mem_size
+                    or force_flush
+                ):
+                    if force_flush:
+                        self._logger.debug("Flushing batch forcefully")
```
seanstory, line 428:
```suggestion
                        self._logger.debug("Flushing batch due to flush signal")
```

connectors/es/sink.py
```diff
@@ -391,6 +435,12 @@ async def _run(self):
                         ),
                         name=f"Elasticsearch Sink: _bulk batch #{batch_num}",
                     )
+
+                    if force_flush:
+                        # Release whoever was waiting for flush to happen
+                        self._logger.debug("Releasing the flush operation awaiter")
+                        doc.trigger()
```
seanstory, line 442: this call to `doc.trigger()` is happening outside the block where you checked `isinstance(doc, ForceFlushSignal)`. Which is safe because you're using `force_flush` to ensure the same condition is true. But I wonder if this is just an indicator that we need a `def flush()` function that can be called from various places, so that the code can look like:
```
if isinstance(doc, ForceFlushSignal):
  flush("from signal")
else:
  add_to_batch(doc)
  if batch_full():
    flush("from capacity")
```

connectors/protocol/connectors.py
```diff
@@ -290,6 +290,10 @@ def sync_cursor(self):
     def terminated(self):
         return self.status in (JobStatus.ERROR, JobStatus.COMPLETED, JobStatus.CANCELED)
 
+    @property
+    def checkpoint(self):
+        return self.get("checkpoint")
```
seanstory, line 295: And remind me why you don't want to to use `cursor`?

connectors/sources/github.py
```diff
@@ -2075,6 +2076,24 @@ async def _fetch_access_control(self, repo_name):
...
 
+    def _is_repo_processed(self, repo_name):
+        if self._checkpoint is None:
+            self._logger.debug(f"{repo_name} was not processed by this job before.")
+            return False
+
+        self._logger.info(
+            f"Skipping repository {repo_name} - has already been processed."
+        )
+        return self._checkpoint.get(repo_name)
```
seanstory, line 2087: How do we know it's been processed, before checking  the checkpoint?

connectors/sync_job_runner.py
```diff
@@ -516,19 +523,45 @@ async def generator(self):
...
+
         while True:
             await asyncio.sleep(interval)
 
             if not await self.reload_sync_job():
                 break
 
+            checkpoint = (
+                None
+                if not self.data_provider  # If we failed before initializing the data provider, we don't need to change the cursor
```
seanstory, line 536:
```suggestion
                if not self.data_provider  # If we failed before initializing the data provider, we don't need to change the checkpoint
```

seanstory (commented):
Few nits/questions, but overall I think this is solid. Really extensible too, where the next service type needs only add minimal logic to add checkpointing.

connectors/es/sink.py
```diff
@@ -356,33 +385,48 @@ async def _run(self):
             batch_num = 0
 
             while True:
+                force_flush = False
```
jedrazb, line 388: I think we don't beed to set `force_flush = True` here, instead we want to only set this after `if isinstance(doc, ForceFlushSignal):` evaluates to True?

I just noticed we set `force_flush = False` in two places one after another.

connectors/es/sink.py
```diff
@@ -356,33 +385,48 @@ async def _run(self):
...
+                        stats[operation][doc_id] = max(doc_size - overhead_size, 0)
+                    self.counters.increment(operation, namespace=BULK_OPERATIONS)
+                    batch.extend(self._bulk_op(doc, operation))
+
+                    bulk_size += doc_size
+
+                if (
+                    len(batch) >= self.chunk_size
+                    or bulk_size > self.chunk_mem_size
+                    or force_flush
```
jedrazb, line 425: Why do we need to add check for `force_flush` in here? The `force_flush = True` is set to True at every iteration of the loop, so the debug log will always get triggered if I understand correctly?

connectors/protocol/connectors.py
```diff
@@ -393,29 +409,50 @@ async def fail(self, error, ingestion_stats=None, connector_metadata=None):
...
             self._wrap_errors("terminate as suspended", e)
 
     async def _terminate(
-        self, status, error=None, ingestion_stats=None, connector_metadata=None
+        self,
+        status,
+        error=None,
+        ingestion_stats=None,
+        connector_metadata=None,
+        checkpoint=None,
```
jedrazb, line 455: Why do we need to update checkpoint in the `_terminate` operation? 

We are updating job metadata (docs ingested etc) but I don't see any changes w.r.t. to done/suspend/error functions in the `sync_job_runner` (those methods don't care about the current checkpoint object)

connectors/sync_job_runner.py
```diff
@@ -516,19 +523,45 @@ async def generator(self):
...
             result = self.sync_orchestrator.ingestion_stats()
             ingestion_stats = {
                 INDEXED_DOCUMENT_COUNT: result.get(INDEXED_DOCUMENT_COUNT, 0),
                 INDEXED_DOCUMENT_VOLUME: result.get(INDEXED_DOCUMENT_VOLUME, 0),
                 DELETED_DOCUMENT_COUNT: result.get(DELETED_DOCUMENT_COUNT, 0),
             }
-            await self.sync_job.update_metadata(ingestion_stats=ingestion_stats)
+            await self.sync_job.update_metadata(
+                ingestion_stats=ingestion_stats, checkpoint=checkpoint
+            )
```
jedrazb, line 564: Can we add debug log that checkpoint was updated? Especially that we have 

`Data was successfully flushed before saving the checkpoint` above, so I would expect to see the logs "saving checkpoint" operation somewhere

jedrazb (commented):
Good job! Couple of nits and questions.

> (not done) Update connectors API to add checkpoint as an argument for update_job_metadata method OR add a separate method.

This also need to include updating the client specification. Adding separate method might be better for o11y if we ever want to track total checkpoint data transfer as checpoints can easily grow (especially in agentless) but adding additional body param to `update_job_metadata` seems natural to me.

connectors/es/sink.py
artem-shelkovnikov: Sorry I don't get it - it's False here and it's only in one place?

connectors/es/sink.py
artem-shelkovnikov: `force_flush` is `True` only if a signal was received, which will happen really rarely - only if checkpoint is saved. Sorry I don't understand your comment fully :/

connectors/es/sink.py
artem-shelkovnikov: Might be a good idea indeed, for readability and future maintenance!

connectors/protocol/connectors.py
artem-shelkovnikov: Mostly to have them different cognitively: `cursor` is to pass state between sync jobs. `checkpoint` is to pass state within the job. I'm thinking of changing this variable to `sync_state`, how do you feel?

connectors/es/sink.py
jedrazb: Yeah ... sorry it might have been late when I reviewed and I misread this as `force_flush = True`  🤦

connectors/sources/github.py
artem-shelkovnikov: It's a mistake haha, good catch

connectors/protocol/connectors.py
navarone-feekery: I would personally prefer `checkpoint` over `sync_state`. I would worry about sync state being conflated with the health of a sync.
I think `pointer` could be a good word as well.

connectors/sources/github.py
```diff
@@ -2142,24 +2168,14 @@ async def get_docs(self, filtering=None):
...
-                            )
-                        else:
-                            yield file_document, None
+                self._mark_repo_as_processed(repo_name)
         else:
             async for repo in self._fetch_repos():
-                if self.is_previous_repo(repo["nameWithOwner"]):
+                repo_name = repo.get("nameWithOwner")
+                if self.is_previous_repo(repo_name):
+                    continue
```
navarone-feekery, line 2176: I wonder if this is obsolete if we are already tracking the processed repos?

connectors/sources/github.py
```diff
@@ -2142,24 +2168,14 @@ async def get_docs(self, filtering=None):
...
-                        if file_document["type"] == FILE:
-                            attachment_metadata["url"] = attachment_metadata["git_url"]
-                            yield (
-                                file_document,
-                                partial(
-                                    self.get_content, attachment=attachment_metadata
-                                ),
-                            )
-                        else:
-                            yield file_document, None
```
navarone-feekery, line 2159: Why was this block removed?

navarone-feekery:
This looks really good! I really love the comments, makes it easy to understand the complex parts.

connectors/protocol/connectors.py
seanstory: I also like the original of `checkpoint`, if we're adding a new field. 

>  cursor is to pass state between sync jobs. checkpoint is to pass state within the job

I don't think I agree with this distinction, though. Checkpoint is really for in case the job fails or is interrupted, right? In which case, it _is_ for passing state between two different sync jobs (or, the same job, but two different runs of it). IMO, this feature is blurring the lines between full and incremental syncs, which I think is a good thing. I'm trying to come up with a situation where you wouldn't want this checkpoint capability to overwrite your `cursor`, and I'm struggling to come up with one. Which is why I'm thinking they should use the same field. Can you describe a counter-example?

connectors/sources/github.py
artem-shelkovnikov: Omg it was a mistake, deleted accidentally. I'll return it back!

connectors/sources/github.py
artem-shelkovnikov: Hmmm you are right, good catch.

I really don't know where this is coming from, so anxious to delete it cause I'm not sure how to test it. I'll give it a look and see if I can test, if not I'll just add a #TODO and leave it be

connectors/sources/github.py
```diff
@@ -2075,6 +2076,28 @@ async def _fetch_access_control(self, repo_name):
...
+
+        return processed
+
+    def _mark_repo_as_processed(self, repo_name):
+        self._logger.debug(f"Setting a checkpoint for {repo_name}.")
+        if self._checkpoint is None:
+            self._checkpoint = {}
+
+        when = datetime.now(timezone.utc)
+        self._checkpoint[repo_name] = when
```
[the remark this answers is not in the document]
artem-shelkovnikov: Mostly it's for debugging/tracing, there's no real need to have a timestamp here.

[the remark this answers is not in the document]
navarone-feekery: I'm okay with the timestamp, I just thought if it's used for comparison at some point it might run into issues with timezones, but maybe I'm overthinking things anyway.

connectors/sync_job_runner.py
artem-shelkovnikov: Added a log line in https://github.com/elastic/connectors/pull/2879/commits/366aba4296d74f1cf5a5ba22c4d9686fa107f0a4: I chose not to mention checkpoints cause logic here always updates the checkpoint anyway and sequence of log lines should be sufficient to understand what happened:

```
Connector reported a new checkpoint, triggering batch flush before saving
Data was successfully flushed before saving the checkpoint
Sync job metadata was updated successfully
```

Do you think we should still be more specific with this?

connectors/protocol/connectors.py
artem-shelkovnikov: I tried to look at it from the following perspective:

1. Immutable "cursor" that gets never updated on the job record after it's initialised is a great audit/tracing mechanism: you can look back in time to see what point the job started from
2. The cursor gets updated only after successful full sync, meaning that if an incremental job is throttled/cancelled mid-way, the next one will try from same checkpoint. This is IMO pretty good, because the data get stale over time. If we choose to combine cursor with checkpoint, it can potentially bring bigger data inconsistencies: e.g. connector synced 50% of data from target system, hang up for a day, restarted and synced the rest 50%, whatever it means. First 50% of data will be a day older than the rest. Not sure if it's a large problem, because I don't know use-cases of our target audience.

So all in all having two fields is a bit more problematic in a short run - I'll also need to write a migration to remove the checkpoint field from mapping, so that there would be no field explosion/mapping conflicts. I don't know if all of the above make it worth to have both fields - theoretically the outcome is very similar.

I also like the blurred line between full/incremental sync - in UI they can be the same, but connector will have a single method that changes behaviour based on cursor provided.

jedrazb: This gets pretty confusing as there a [connector sync_cursor](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/entsearch/connector/elastic-connectors-mappings.json#L300) and [sync job cursor](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/entsearch/connector/elastic-connectors-sync-jobs-mappings.json#L112) that references connector cursor. 
 
> I'm trying to come up with a situation where you wouldn't want this checkpoint capability to overwrite your cursor, and I'm struggling to come up with one. Which is why I'm thinking they should use the same field. Can you describe a counter-example?

Resonating with what Artem mentioned, the `sync_cursor` acts as an immutable entry point state indicating where sync jobs should start (e.g., the last sync timestamp in `google_drive`). On the other hand, checkpoints track an arbitrarily defined unit of work within the same sync operation (e.g., synced drives in gdrive). These checkpoints do not necessarily need to align with the `sync_cursor` abstraction.

This distinction becomes significant when running incremental syncs (Note: I'm aware that Artem’s changes only impact full syncs). Still, I think it's worth discussing a potential scenario with incremental syncs, as we are aware that incremental syncs are necessity for some large connector users. 

For instance, in `google_drive` incremental sync, we cannot overwrite the sync_cursor (which is a timestamp used with a delta API to fetch changes since the last sync) with a periodic checkpoint because:
- The checkpoint cannot be updated with the current timestamp since we cannot confirm that all changes up to that time have been fully synced (e.g., if the last sync was 24 hours ago and there are 10,000 new documents to process).
- In google_drive, we might need to track specific units of work such as folders, shared drives, etc., for checkpointing purposes. This is not the same as `timestamp` sync cursor
- If an incremental sync is interrupted (e.g., due to an agentless pod restart), both the last successful sync_cursor and the checkpoint are essential to correctly resume the incremental job.

connectors/sources/sharepoint_online.py
```diff
@@ -1147,6 +1147,7 @@ def __init__(self, configuration):
 
         self._client = None
         self.site_group_cache = {}
+        self._temporary_cursor = None
```
seanstory, line 1150: I don't think I understand what's changed here.
This doesn't implement checkpointing for SPO, AFAICT. But this is just storing the cursor info we used to have in `self._sync_cursor` instead in `self._temporary_cursor`, and then overwriting `self._sync_cursor` with `self._temporary_cursor` at the end. Why keep it in to variables? As long as the intermediate `self._sync_cursor` isn't persisted to ES, it's effectively "temporary" because it's just in-memory for this instance of the DataSource, right?

connectors/sync_job_runner.py
```diff
@@ -151,6 +151,13 @@ async def execute(self):
                 self._data_source_framework_config()
             )
 
+            # Only full syncs restart from a checkpoint that's stored in a sync job
+            if self.sync_job.sync_cursor and self.sync_job.job_type == JobType.FULL:
+                self.sync_job.log_debug(
```
seanstory, line 156:
```suggestion
                self.sync_job.log_info(
```
loosely-held suggestion. Seems low-noise, but valuable.

connectors/sync_job_runner.py
```diff
@@ -516,19 +522,46 @@ async def generator(self):
...
+                None
+                if not self.data_provider  # If we failed before initializing the data provider, we don't need to change the checkpoint
+                else (
+                    self.data_provider.sync_cursor()
+                    if self.sync_job.is_content_sync()
+                    else None
+                )
+            )
+
+            if cursor != last_cursor:
```
seanstory, line 542:
```suggestion
            if cursor != last_cursor and cursor != None:
```

I don't think we want to overwrite an old cursor if this was an access-control sync or if the data_provider failed to initialize

connectors/sync_job_runner.py
```diff
@@ -516,19 +522,46 @@ async def generator(self):
...
+
             result = self.sync_orchestrator.ingestion_stats()
             ingestion_stats = {
                 INDEXED_DOCUMENT_COUNT: result.get(INDEXED_DOCUMENT_COUNT, 0),
                 INDEXED_DOCUMENT_VOLUME: result.get(INDEXED_DOCUMENT_VOLUME, 0),
                 DELETED_DOCUMENT_COUNT: result.get(DELETED_DOCUMENT_COUNT, 0),
             }
-            await self.sync_job.update_metadata(ingestion_stats=ingestion_stats)
+            await self.sync_job.update_metadata(
+                ingestion_stats=ingestion_stats, cursor=cursor
```
seanstory, line 563: Do we also need to check here that the cursor was updated but non-null?

tests/test_sync_job_runner.py
```diff
@@ -34,6 +34,7 @@
 ACCESS_CONTROL_INDEX_NAME = ".search-acl-filter-search-mysql"
 TOTAL_DOCUMENT_COUNT = 100
 SYNC_CURSOR = {"foo": "bar"}
+NEXT_SYNC_CURSOR = {"foo": "bar", "hehe": "not_hehe"}
```
seanstory, line 37: ho ho :)

connectors/sources/sharepoint_online.py
artem-shelkovnikov: That is the thing: to make checkpoints work, connector needs to be able to write its state to some variable. Since we use the same `connector.sync_cursor`, connector will store checkpoints in `self._sync_cursor` variable (see github implementation).

Then framework checks this variable from time to time, and if it changed, it's saved into the job state (not connector state).

Now thinking of it, this change is not necessary because the connector state is updated at the end of successful sync, but all of it is just highly confusing.

This is the main reason I wanted checkpoints separated - our current way of arranging things (different job types, incremental cursors) just amplify the complexity of checkpointing a lot. To have something solid we'd need a good revamp of this logic, now it feels it works but is very fragile

connectors/sync_job_runner.py
artem-shelkovnikov: `update_metadata` internally checks [that](https://github.com/elastic/connectors/pull/2879/files#diff-c2f73fe85d1652beb4af8939b73360a399a3496c1863e34b900e12d694420571R402-R403): 

```
                if cursor:
                    doc["connector"] = {"sync_cursor": cursor}
```

Idk if it would cause confusion too, maybe there should be a way for the connector to indicate that cursor needs a reset, e.g. if it became corrupted?

seanstory:
Should we close this? We can always re-open or cherry-pick later when we come back to this.

artem-shelkovnikov:
Good call, closing

PR #2884, opened 2024-10-16
Set `feature_use_connectors_api` to true

navarone-feekery:
Set `feature_use_connectors_api` to true so we can have Connectors API usage as default.

### Release Note

Directly accessing connector and sync job state through `.elastic-connectors*` indices is deprecated, and will be disallowed entirely in a future release. Instead, the Elasticsearch Connector APIs should be used. Connectors framework code now uses the Elasticsearch Connector APIs by default.


seanstory:
heh jinx: https://github.com/elastic/connectors/pull/2898

This is what I get for not keeping up on my github notifications. :)

seanstory (approved):

navarone-feekery:
Thanks for the much better improved note @seanstory!

navarone-feekery:
Closing in favour of https://github.com/elastic/connectors/pull/2902 which resolves the bugs that flipping this flag surfaced

PR #2887, opened 2024-10-16
Docker: make a symlink to maintain backwards-compatibility after venv change, update makefile to build with --no-cache

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/2886

This PR makes a symlink from `/app/.venv/bin` to `/app/bin` to preserve old logic of running the app from `/app/bin/elastic-ingest`. Otherwise they'd have to run it with `/app/.venv/bin/elastic-ingest` and it feels getting out of hand. Documentation still points to `/app/bin/elatic-ingest`, so let's just symlink it!

#### Pre-Review Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM python:3.11-slim-bookworm
-RUN apt update && apt upgrade && apt install make -y
+RUN apt update && apt upgrade && apt install make git -y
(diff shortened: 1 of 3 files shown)


jedrazb (approved):
🚢

PR #2890, opened 2024-10-16
Bump version to 8.17 on 8.x

miguelgrinberg:
Part of the 8.16 release (see https://github.com/elastic/search-team/issues/8045)

#### Pre-Review Checklist


catalog-info.yaml
```diff
@@ -194,6 +202,10 @@ spec:
         trigger_mode: "none"
       repository: "elastic/connectors"
       schedules:
+        Daily 8_16:
```
miguelgrinberg, line 205: @seanstory there is a comment at the top of this cron schedule that says it applies to versions < 8.16. Let me know if adding 8.16 here is not needed.

seanstory: Correct, no need to add this block. Once we ship 8.15.3 in a few days, we'll be able to delete this whole job spec.

tests/ftest.sh
```diff
@@ -20,7 +20,7 @@ export PERF8_TRACE=${PERF8_TRACE:-False}
 export REFRESH_RATE="${REFRESH_RATE:-5}"
 export DATA_SIZE="${DATA_SIZE:-medium}"
 export RUNNING_FTEST=True
-export VERSION='8.16.0-SNAPSHOT'
+export VERSION='8.17.0-SNAPSHOT'
```
seanstory, line 23: I expect this is going to cause things to fail, until Elasticsearch produces an 8.17.0-SNAPSHOT. We need to find a way around this.

miguelgrinberg:
@seanstory Is this still good to go with a failure on the smoke tests?

tests/ftest.sh
miguelgrinberg: Should we add the `skip-tests` label to be able to merge and avoid the chicken and egg problem?

.backportrc.json
```diff
@@ -2,6 +2,7 @@
   "targetBranchChoices": [
```
artem-shelkovnikov, line 2: These changes need to go to `main`

catalog-info.yaml
```diff
@@ -97,6 +97,10 @@ spec:
         trigger_mode: "none"
```
artem-shelkovnikov, line 97: Same - these need to go to main

.backportrc.json
miguelgrinberg: @artem-shelkovnikov I have zero familiarity with this project so I need better directions. Does the entire PR need to be merged against main, or do I need to split this PR into two with some changes going to 8.x and some going to main? And in that case, which changes go to each branch?

miguelgrinberg: I've added this change for main on a separate PR https://github.com/elastic/connectors/pull/2893

catalog-info.yaml
miguelgrinberg: I've added this change for main in https://github.com/elastic/connectors/pull/2893

.backportrc.json
artem-shelkovnikov: Sorry just seen your response - CI-related things normally stay on `main` - so `catalog-info.yaml` changes just go there.

Not sure about this file - `.backportrc.json` TBH cause we might need to do backports `8.x` -> `8.something`. So merging this file in `8.x` would not hurt, but it might not affect anything

seanstory (approved):
I thought that CI would fail with es for fests, but it's not...
```
[FMWK][10:52:19][INFO] Elasticsearch 8.17.0-SNAPSHOT and Connectors 8.17.0 are compatible
``` 

Weird. But ok, let's get this in.

PR #2894, opened 2024-10-17
Update pipeline.yml to do packaging on 8.16

seanstory:
part of https://github.com/elastic/search-team/issues/8045

We need to make the new maintenance branch produce packaging artifacts

#### Pre-Review Checklist


artem-shelkovnikov (approved):

artem-shelkovnikov:
Do you need to backport it tho?

seanstory:
> Do you need to backport it tho?

yes, because the pipeline itself is determined by what's on the branch. _which_ pipeline file is run is picked by `main`, but the contents of the file are picked up locally.

PR #2898, opened 2024-10-17
Switch to API access, over direct index access, by default

seanstory:
For 8.16.0, we want to keep moving forward with deprecating direct index access for `.elastic-connectors*`, and a good way to dogfood that is to toggle this feature flag by default. We can still leave the feature flag as an escape hatch for now, in case any bugs surface.

#### Pre-Review Checklist

## Release Note

Directly accessing connector and sync job state through `.elastic-connectors*` indices is deprecated, and will be disallowed entirely in a future release. Instead, the Elasticsearch Connector APIs should be used.

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -157,7 +157,8 @@
 #
 #
 ##  Enable usage of Connectors API instead of calling connectors indices directly
-#elasticsearch.feature_use_connectors_api: false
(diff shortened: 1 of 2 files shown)


seanstory:
closing in favor of: https://github.com/elastic/connectors/pull/2884

PR #2903, opened 2024-10-18
Stop swallowing errors when paginating

seanstory:
## Closes https://github.com/elastic/connectors/issues/2394

Previously, `paginated_api_call` would swallow all exceptions (after logging a warning). Now, it raises, them, and let's the caller decide whether to swallow or re-raise them.

- `_remote_validation()`: re-raises (seems important to fail validation if invalid)
- `fetch_attachments()`: logs warnings, but does not raise
- `fetch_page_blog_documents()`: logs warnings, but does not raise
- `fetch_spaces()`: re-raises (seems important to not skip whole spaces - that should fail the sync)
- `search_by_query()`: re-raises (seems important to let you know if your advanced sync rules are incorrect)

Also, I took a pass through the file to make sure that we were using `exc_info=exception` everywhere we were logging about an exception, instead of stuff like `f"Exception: {exception}"`

#### Pre-Review Checklist

## Related Pull Requests

* this PR replaces https://github.com/elastic/connectors/pull/2610

## Release Note

Improved the Confluence connector by ensuring that Confluence pagination API request errors bubble up when appropriate. This may result in more syncs ending in an error state, but these would not be new errors/issues, they will just be more visible and actionable.

### DIFF
--- a/.buildkite/nightly_steps.yml
+++ b/.buildkite/nightly_steps.yml
@@ -1,19 +1,23 @@
-retry:
-  automatic:
-    - exit_status: -1  # Connection to the Agent was lost
-      signal_reason: none
-      limit: 2
-    - exit_status: 255  # Timeout
-      signal_reason: none
-      limit: 2
-    - signal_reason: 2  # Flaky test
-      limit: 2
+definitions:
+  retries: &retries
+    retry:
+      automatic:
+        - exit_status: -1  # Connection to the Agent was lost
+          signal_reason: none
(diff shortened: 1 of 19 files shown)


seanstory:
CI failed. Separate fix is: https://github.com/elastic/connectors/pull/2905
Will rebuild after that merges

seanstory:
I didn't write tests. I'd asked Crest to, but now I think it's back on us.

PR #2902, opened 2024-10-18
[Connector APIs] Enable Connector APIs framework-wide + update functional tests

jedrazb:
Closes https://github.com/elastic/search-team/issues/8476

### Release Note

Directly accessing connector and sync job state through `.elastic-connectors*` indices is deprecated, and will be disallowed entirely in a future release. Instead, the Elasticsearch Connector APIs should be used. Connectors framework code now uses the Elasticsearch Connector APIs by default.

### Changes

Major change: toggle `feature_use_connectors_api` flag to True

Additionally, use Connector APIs in functional test runner for:
- creating new connector doc
- updating scheduling 
- setting connector configuration 

Cleanup unnecessary boilerplate code for anything other than `configuration` in existing connector.json files. Also format the files with fixed indentation. This was scripted operation so no human errors here (hopefully). 

Update logic where necessary. Update documentation. 

### Bonus

Required Connector APIs endpoints added with newly released ES client 🥳 

### Validation
- Tested locally
- Let's see what buildkite says about this PR

PR #2914, opened 2024-10-23
Update example config yml with connectors api feature flag

jedrazb:
Followup on: #2902

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -157,6 +157,7 @@
 #
 #
 ##  Enable usage of Connectors API instead of calling connectors indices directly
+##  Using direct index access is deprecated, and will be disallowed entirely in a future version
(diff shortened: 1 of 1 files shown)


config.yml.example
```diff
@@ -157,7 +157,7 @@
 #
 #
 ##  Enable usage of Connectors API instead of calling connectors indices directly
```
navarone-feekery, line 159: I really liked @seanstory's addition to the flag description in [this closed PR](https://github.com/elastic/connectors/pull/2898/files). I think we should also add this.

```suggestion
##  Enable usage of Connectors API instead of calling connectors indices directly
##    Using direct index access is deprecated, and will be disallowed entirely in a future version
```

PR #2921, opened 2024-10-24
Removes gratuitous 'critical' logging

seanstory:
Related to https://github.com/elastic/connectors/pull/2916, which reminded me that this has been annoying for a while. We should only use CRITICAL when the process is imminently crashing. Sync errors are not "CRITICAL".


seanstory:
Bump. Can I get a review?

artem-shelkovnikov (approved):

PR #2923, opened 2024-10-25
Followup on #2861

jedrazb:
## Follow up on unaddressed comments from #2861 

Addressed items:
- added debug logs in connector record manager
- moved `connector_exists` to ConnectorIndex class
- moved `generate_random_id` to utils class
- adapted tests

### DIFF
--- a/connectors/agent/connector_record_manager.py
+++ b/connectors/agent/connector_record_manager.py
@@ -77,7 +77,7 @@ class ConnectorRecordManager:
             return False, "No 'connectors' key found in the service configuration."
 
         if len(connectors) == 0:
-            return False, "Empty connectors array found in the service configuration."
+            return False, "Empty 'connectors' array found in the service configuration."
 
         for connector in connectors:
             if "connector_id" not in connector:
--- a/connectors/agent/connector_record_manager.py
+++ b/connectors/agent/connector_record_manager.py
@@ -3,12 +3,10 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
-import secrets
-import string
 
 from connectors.agent.logger import get_logger
-from connectors.es.index import DocumentNotFoundError
 from connectors.protocol import ConnectorIndex
+from connectors.utils import generate_random_id
 
 logger = get_logger("agent_connector_record_manager")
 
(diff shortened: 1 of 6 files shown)


connectors/agent/connector_record_manager.py
```diff
@@ -30,25 +28,27 @@ async def ensure_connector_records_exist(self, agent_config, connector_name=None
         """
 
         if not self._agent_config_ready(agent_config):
+            logger.debug(
+                "Agent configuration is not ready to create a connector record."
+            )
```
artem-shelkovnikov, line 33: Can we update the copy to provide more specific detail on what's going on? How can configuration be not ready? Is it gonna be happening a lot?

connectors/protocol/connectors.py
```diff
@@ -172,6 +173,18 @@ async def connector_put(
             index_name=index_name,
         )
 
+    async def connector_exists(self, connector_id):
+        try:
+            doc = await self.fetch_by_id(connector_id)
+            return doc is not None
```
artem-shelkovnikov, line 179: It's probably fine now, but is there a way to do this with as little traffic as possible? We can have connector records with cursors that'll weigh up to 5MB, personally I expect a method like "exists" to be veeeery lightweight

jedrazb: Agreed, it should be lightweight, there is not much we can do in this PR as we would need to adapt the API to allow to query only for selected fields.

I filed an issue: https://github.com/elastic/connectors/issues/2937

connectors/agent/connector_record_manager.py
jedrazb: > Is it gonna be happening a lot?

It should not be happening a lot, perhaps only once the service starts up with default values.  As soon as we receive first checkin the configuration should be "ready"  ... but this is extra check to ensure that the expected key/value pairs are actually there.

artem-shelkovnikov (approved):

connectors/agent/connector_record_manager.py
```diff
@@ -62,43 +64,42 @@ async def ensure_connector_records_exist(self, agent_config, connector_name=None
...
         """
+
         connectors = agent_config.get("connectors")
-        if connectors is None or len(connectors) == 0:
-            return False
+        if connectors is None:
+            return False, "No 'connectors' key found in the service configuration."
+
+        if len(connectors) == 0:
+            return False, "Empty connectors array found in the service configuration."
```
artem-shelkovnikov, line 80:
```suggestion
            return False, "Empty `connectors` array found in the service configuration."
```

artem-shelkovnikov (approved):

PR #2928, opened 2024-10-30
Retry buildkite pipelines for nightly tests + on-commit pipelines except the ones in the diff plugin

artem-shelkovnikov:
Properly implement retries for failing buildkite pipelines: before they did not work at all.

All commits you see here were to test what works.

What works:
- Retrying all steps in nightly pipelines
- Retrying steps that are not a part of `diff` plugin

This change should reduce flakyness of our pipelines

### DIFF
--- a/.buildkite/nightly_steps.yml
+++ b/.buildkite/nightly_steps.yml
@@ -1,19 +1,23 @@
-retry:
-  automatic:
-    - exit_status: -1  # Connection to the Agent was lost
-      signal_reason: none
-      limit: 2
-    - exit_status: 255  # Timeout
-      signal_reason: none
-      limit: 2
-    - signal_reason: 2  # Flaky test
-      limit: 2
+definitions:
+  retries: &retries
+    retry:
+      automatic:
+        - exit_status: -1  # Connection to the Agent was lost
+          signal_reason: none
+          limit: 2
+        - exit_status: 255  # Timeout
+          signal_reason: none
+          limit: 2
+        - exit_status: 2  # Flaky test
+          signal_reason: none
+          limit: 2
 
 env:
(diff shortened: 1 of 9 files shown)


.buildkite/pipeline.yml
```diff
@@ -522,7 +526,6 @@ steps:
             - "requirements/**"
             config:
               label: "🔨 GraphQL"
-              <<: *test-agents
```
navarone-feekery, line 525: Did you intend to have `*retries` added here too?

artem-shelkovnikov: Oh wow I also did not intend to delete the test-agents here, good catch!

navarone-feekery (approved):
LGTM!

.buildkite/nightly_steps.yml
```diff
@@ -1,19 +1,23 @@
...
+    retry:
+      automatic:
+        - exit_status: -1  # Connection to the Agent was lost
+          signal_reason: none
+          limit: 2
+        - exit_status: 255  # Timeout
+          signal_reason: none
+          limit: 2
+        - exit_status: 2  # Flaky test
+          signal_reason: none
```
navarone-feekery, line 12: Late non-blocking question: `signal_reason: none` wasn't originally included for the flaky test retry. What changes by adding this?

artem-shelkovnikov: It's here for consistency.

`signal_reason` itself can be [one of these options](https://buildkite.com/docs/pipelines/command-step): 

- "*"
- none
- cancel
- agent_stop

For now it seems like retrying on "cancel" or "agent_stop" does not make sense, so retrying only when there's no reason.

PR #2943, opened 2024-11-07
Bump pyOpenSSL lib version

jedrazb:
## Changes 

Bump `pyOpenSSL` to latest version, this in turns updates `cryptography` (our transitive dependency) as Snyk was complaining about our `cryptography` version. 

+ regenerate `Notice.txt` file, seems like it was not updated in a while

Relates to [CVE-2024-6119](https://www.cve.org/CVERecord?id=CVE-2024-6119)

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -978,7 +978,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
 aiohappyeyeballs
-2.4.0
+2.4.3
 Other/Proprietary License; Python Software Foundation License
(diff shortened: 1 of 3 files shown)


NOTICE.txt
```diff
@@ -3212,6 +3213,233 @@ Apache Software License
 
 
 
+elastic_agent_client
+0.0.1
+Other/Proprietary License
+ELASTIC LICENSE AGREEMENT
```
seanstory, line 3219: :D this feels so legit

seanstory (approved):

seanstory:
> regenerate Notice.txt file, seems like it was not updated in a while

Yeah, need to figure out how to cover this in CI. One day.

PR #2957, opened 2024-11-12
use quotes on 8.16 version in catalog-info.yml

seanstory:
I broke terrazo 🙈 see: https://github.com/elastic/connectors/pull/2951#issuecomment-2471635528


tpgilmore (approved):
Looks good, thank you!

PR #2979, opened 2024-11-19
[Publishing] Load ES image for ftest from DRA artifact

jedrazb:
### Changes

This PR updates the functional test pipeline to download and load es docker images from DRA artifacts instead of relying on the snapshot image in the registry. This ensures compatibility during the release process when the ES snapshot image is not yet available in the elastic docker registry.

Changes:
- Resolves the appropriate ES tarball URL based on the architecture (see available [arch](https://artifacts-snapshot.elastic.co/elasticsearch/9.0.0-0124d28c/summary-9.0.0-SNAPSHOT.html)) using the Elasticsearch DRA manifest.
- Removed Kibana and Enterprise Search from `docker-compose.yml` fixtures as they are not required for our ftests.

Validation:
- Tested manually different versions 
- How you can validate? Go to buildkite build and check e.g. smoke test steps to confirm that the image is coming from DRA artifact

Error handling
- Added a lot of logs to make sure that we know why we failed, e.g. when we need to produce an artifact and ES still not produced one we will see message like:

```
+++ curl -sS -f https://artifacts-snapshot.elastic.co/elasticsearch/latest/9.0.1-SNAPSHOT.json
++ RESPONSE='curl: (22) The requested URL returned error: 404'
++ CURL_EXIT_CODE=22
++ '[' 22 -ne 0 ']'
++ echo 'Error: Failed to fetch DRA manifest for artifact elasticsearch and version 9.0.1-SNAPSHOT.'
```

### DIFF
--- a/Dockerfile.ftest.wolfi
+++ b/Dockerfile.ftest.wolfi
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:b70abdb46f6a1ed5057fb50018924eea03345044b5224c260a3d0334a748eb96
+FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:62936d0d2db57c9bd195b31c1fcafff1413040ea94d37c88435487ebe041fb54
 USER root
 COPY . /connectors
 WORKDIR /connectors
--- a/Dockerfile.wolfi
+++ b/Dockerfile.wolfi
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:b70abdb46f6a1ed5057fb50018924eea03345044b5224c260a3d0334a748eb96
+FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:62936d0d2db57c9bd195b31c1fcafff1413040ea94d37c88435487ebe041fb54
 USER root
 COPY . /app
 WORKDIR /app
--- a/connectors/sources/outlook.py
+++ b/connectors/sources/outlook.py
@@ -44,6 +44,7 @@ from connectors.utils import (
     html_to_text,
     iso_utc,
     retryable,
+    url_encode,
 )
 
 RETRIES = 3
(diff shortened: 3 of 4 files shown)


tests/ftest.sh
```diff
@@ -14,13 +14,115 @@ VIRTUAL_ENV="$ROOT_DIR/.venv"
...
+  echo "Error: Docker tarball URL not found in the manifest."
+  exit 1
+fi
+
+# Execute the functions to download and load the image
+download_docker_tarball "$DOCKER_TARBALL_URL" "$DOCKER_TARBALL_NAME"
+load_docker_image "$DOCKER_TARBALL_NAME"
+
+# Export image name following DRA conventions
+export ELASTICSEARCH_DRA_DOCKER_IMAGE="elasticsearch:$ARCH"
```
artem-shelkovnikov, line 120: How well would that work if you run ftests locally on different branches (9.0 and 8.x) for example?

Since there's no version it can have weird conflicts, right? Or is it resolved somewhere else?

jedrazb: It works well, I tried it with e.g. 8.16 

We are fetching the manifest for the right version that depends on the connector version (in the VERSION file) see this line https://github.com/elastic/connectors/pull/2979/files#diff-2c88fda54d210cfb99f46dd547063eb3b88f8f41296ce60881535313cd7b281cR105 and you can backtrack where the version variable comes from. 

Unfortunately the `elasticsearch` images loaded from tarball do not have the "stack version" tag, instead they are just in format  `elasticsearch:{aarch60|x86_64}`, so the arch is the tag.

But given that we are fetching it from the right manifest URL it is the correct version.

jedrazb: E.g. this is for SPO ftests in 9.0.0
Screenshot 2024-11-19 at 13 07 13

jedrazb: If you run first on `main` branch, then checkout to 8.16 and run ftest again, the docker load will override the the image for your arch and the logs will show
```
[FMWK][12:10:39][INFO] Elasticsearch 8.16.0-SNAPSHOT and Connectors 8.16.0 are compatible
```
( I tried this sequence of steps locally by tweaking the version in the file)

artem-shelkovnikov (approved):
🚀 🌕

jedrazb:
@seanstory 👋 you might want to take a look as well as you have been involved with onboarding connectors to DRA

jedrazb:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.x|[](https://github.com/elastic/connectors/pull/2981)|

### Questions ?

PR #2989, opened 2024-11-20
[Agent] Respect connector_id defined in agentless

jedrazb:
## Closes https://github.com/elastic/search-team/issues/8669

Changes:
- if `connector_id` is provided in fleet policy, create a connector with such id
- create all agentless connectors with `is_native` flag
- remove secret api key usage for native connectors
- support discovering native connectors by id (the framework will no longer be running in "force native" mode, hence the modified query works for both managed and self-managed connectors)

Other changes:
- optimize agent dockerfile and fix issue with `yq` installation (I think in 9.0.0 "correct" yq was removed in agent image and apt-get was fetching different tool, confirmed it was coming from different repo) 

See example in action:

Screenshot 2024-11-20 at 11 21 53

Screenshot 2024-11-20 at 11 20 46

#### Pre-Review Checklist

#### Followup work
I am aware that there is some more work to do, this PR is just what we need now, followup tickets are filed:
- https://github.com/elastic/connectors/issues/2990
- https://github.com/elastic/connectors/issues/2991
- https://github.com/elastic/connectors/issues/2992


connectors/protocol/connectors.py
```diff
@@ -231,7 +237,6 @@ async def supported_connectors(self, native_service_types=None, connector_ids=No
         custom_connectors_query = {
             "bool": {
                 "filter": [
-                    {"term": {"is_native": False}},
```
jedrazb, line 234: since we treat all connectors the same (discover by ID), remove this filter

connectors/agent/connector_record_manager.py
```diff
@@ -56,6 +56,7 @@ async def ensure_connector_records_exist(self, agent_config, connector_name=None
                         connector_id=connector_id,
                         service_type=service_type,
                         connector_name=connector_name,
+                        is_native=True,
```
artem-shelkovnikov, line 59: So we've settled on `is_native` flag reusage in the end?

connectors/source.py
```diff
@@ -384,7 +384,7 @@ class BaseDataSource:
     advanced_rules_enabled = False
     dls_enabled = False
     incremental_sync_enabled = False
-    native_connector_api_keys_enabled = True
+    native_connector_api_keys_enabled = False
```
artem-shelkovnikov, line 387: What happens here?

jedrazb: This is old setting to enable `native connector secrets` for native connectors. Since we no longer rely on secrets for elastic managed connectors I just set this flag to false without thinking worrying about proper cleanup, I've filed followup ticket:
https://github.com/elastic/connectors/issues/2990

connectors/agent/connector_record_manager.py
jedrazb: This is what I understood from the "mega-thread" we've had last week, will DM you the link

connectors/source.py
artem-shelkovnikov: So it'll break connectors running in native mode in traditional setup in 9.x, but we don't care since it's something nobody's supposed to do, right?

jedrazb: The native connectors would be broken anyway since only service account that manages the ent-search node has  access to read those https://github.com/elastic/elasticsearch/blob/8c20ac5884158b88fdd598e422db632e1734aabb/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/service/ElasticServiceAccounts.java#L48 

After migration to 9.0 those secrets are useless because noone has permissions to read them

artem-shelkovnikov (approved):

PR #3015, opened 2024-12-05
Stringify doc[id] field

jedrazb:
## Changes

Always stringify `doc["id"]` field. The default index mappings were removed in https://github.com/elastic/connectors/pull/3013. 

This limits the scenario of running into weird issues with ES mappings for numeric-like ID fields (overflows, etc). ID should be keyword never numeric ideally.

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -661,7 +661,8 @@ class Extractor:
                 if count % self.display_every == 0:
                     self._log_progress()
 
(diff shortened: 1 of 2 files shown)


seanstory:
One of the network drive ftests passed. The other failed. I'm gonna merge this anyway, as it's clearly an improvement.

seanstory (approved):

PR #3044, opened 2024-12-17
Fix mysql advanced sync rule execute error

JunJie-GitHub:
BUG: When configuring the advanced sync rule for the MySQL connector, the sync action throws an error: 'MySQLClient.yield_rows_for_query() missing 1 required positional argument: 'primary_key_columns''.

Root Cause: The yield_rows_for_query method is called without the required primary_key_columns parameter.

Fix: Add the missing primary_key_columns parameter when calling the yield_rows_for_query method.

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -638,7 +638,7 @@ class MySqlDataSource(BaseDataSource):
             )
             return
 
-        async for row in client.yield_rows_for_query(query):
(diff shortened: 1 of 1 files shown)


seanstory:
buildkite test this

seanstory:
buildkite test this

seanstory (approved):
LGTM

seanstory:
Thanks for the contribution!

PR #3057, opened 2024-12-24
[New connector] Onelake connector

Delacrobix:
TITLE: [New connector] Onelake connector

## Closes https://github.com/elastic/connectors/issues/3051

Added Onelake connector files, the connector’s code, test, requirements and the reference to the connector in the sources list (connectors/config.py)

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/onelake.py
+++ b/connectors/sources/onelake.py
@@ -113,7 +113,7 @@ class OneLakeDataSource(BaseDataSource):
             self._logger.error(f"Error while getting token credentials: {e}")
             raise
 
-    def _get_service_client(self):
+    async def _get_service_client(self):
         """Get the service client for OneLake
 
         Returns:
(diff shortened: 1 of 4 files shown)


connectors/sources/onelake.py
```diff
@@ -0,0 +1,326 @@
...
+        Args:
+            doc_paths (list): List of paths extracted from OneLake
+
+        Yields:
+            tuple: File document and partial function to get content
+        """
+
+        for path in doc_paths:
+            file_name = path.name.split("/")[-1]
+            field_client = await self._get_file_client(file_name)
```
artem-shelkovnikov, line 306: Looking at the code, for each file multiple clients are created. Is there a way to reuse the clients between calls? I can see how this can become a problem with RAM

connectors/sources/onelake.py
```diff
@@ -0,0 +1,326 @@
...
+            raise
+
+    def _get_account_url(self):
+        """Get the account URL for OneLake
+
+        Returns:
+            str: Account URL
+        """
+
+        return f"https://{self.configuration['account_name']}.dfs.fabric.microsoft.com"
```
artem-shelkovnikov, line 99: This can just become an field of the class that's set during __init__

connectors/sources/onelake.py
Delacrobix: Hello Artem, thank you so much for your feedback.

Regarding this comment: each file client represents a specific file and is initialized with the file name, so it’s not possible to reuse the client. I believe the garbage collector should remove unused clients, but that’s just an assumption.

connectors/sources/onelake.py
Delacrobix: I moved it to the constructor method:

https://github.com/Delacrobix/connectors/blob/78627c81e17ed666bdf1d30a9fd8cc24740893df/connectors/sources/onelake.py#L33-L35

PR #3059, opened 2024-12-25
Set connector status to CONNECTED after successful validation and ping

artem-shelkovnikov:
## Closes https://github.com/elastic/search-team/issues/8535

Currently connectors have no way to report they're connected.

This PR makes it so that if a data source succeeds in validation and ping, then its status is set to CONNECTED.

In practice it means, that a user who had problems with their setup will see an error in UI. If they fix the credentials, before the error would go away after a sync starts. After this change the error will disappear within the service poll interval from user's action.

#### Pre-Review Checklist

## Release Note

Fix a UI issue with connector reporting problems with connection even if wrong configuration was fixed. Before the error would go away only when a sync is executed.

### DIFF
--- a/tests/commons.py
+++ b/tests/commons.py
@@ -15,18 +15,28 @@ class AsyncIterator:
     Async documents generator fake class, which records the args and kwargs it was called with.
     """
 
-    def __init__(self, items):
+    def __init__(self, items, reusable=False):
+        """
+        AsyncIterator is a test-only abstraction to mock async iterables.
+        By default it's usable only once: once iterated over, he iterator will not
+        iterate again any more.
+        If reusable is True, then iterator can be re-used, but only if it's used by a single coroutine.
+        If AsyncIterator is used in several coroutines, it'll not work correctly at all
+        """
         self.items = items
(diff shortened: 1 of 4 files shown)


tests/commons.py
```diff
@@ -15,18 +15,28 @@ class AsyncIterator:
...
         self.call_count = 0
+        self.reusable = reusable
 
     def __aiter__(self):
         return self
 
     async def __anext__(self):
         if self.i >= len(self.items):
+            if self.reusable:
+                self.i = 0
```
navarone-feekery, line 39: I can see it's only for tests so not a big deal but does this open a risk for infinite looping anywhere?

navarone-feekery (approved):
LGTM, just have a question

tests/commons.py
artem-shelkovnikov: It should not be, no. Here there's no return statement, so next line that raises an error will be triggered and iterations will stop. Next time this iterator is used it's going to be iterating over same original collection, before it would immediately stop iteration.

PR #3063, opened 2024-12-27
Switch Dockerfile image to wolfi and add pipeline for vulnerability scanning

kostasb:
## Description

The goal of this PR is to:

1) move "extensible Dockerfiles" to a distroless base image. Dockerfile and Dockerfile.ftest are not used in Elastic's docker image releases but are rather provided as a stepping stone for users to build custom images on.
2) add a pipeline for vulnerability scanning of built images

This is based on @acrewdson 's [draft](https://github.com/elastic/connectors/pull/3041). It uses `wolfi-base` images and adds buildkite pipelines to build, test and scan the resulting docker images using trivy.

A notification method for vulnerability reports from Trivy is TBD.

#### Pre-Review Checklist

## Release Note

The OSS Dockerfile provided in the connectors repo has been updated to use a different base image for improved security. Users building custom docker images based on this Dockerfile may have to review their configuration for compatibility with the new base image.


Dockerfile
```diff
@@ -1,7 +1,13 @@
...
+RUN apk update && apk add python3=~${python_version} make git
+
+COPY --chown=nonroot:nonroot . /app
+
+USER nonroot
 WORKDIR /app
-RUN make clean install
-RUN ln -s .venv/bin /app/bin
+RUN make clean install && ln -s .venv/bin /app/bin
+
```
acrewdson, line 12: I had [this line](https://github.com/elastic/connectors/pull/3041/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R15)
> `RUN apk del make git`

so that we could remove now-unnecessary packages from the final images, with the assumption that we do not need `make` or `git` at runtime, but we could confirm with someone from @elastic/search-extract-and-transform that it makes sense

catalog-info.yaml
```diff
@@ -177,6 +177,39 @@ spec:
         search-extract-and-transform: {}
         search-productivity-team: {}
 
+# Dockerfiles nightly images build pipeline
+---
+apiVersion: "backstage.io/v1alpha1"
+kind: "Resource"
+metadata:
+  name: "connectors-dockerfiles-nightly"
+  description: "Nightly Connectors Dockerfiles Build"
```
acrewdson, line 186:
```suggestion
  description: "Nightly build and scan of the public/extensible connectors Dockerfiles"
```

describing how these Dockerfiles are different to the 'official' (`.wolfi`) ones is sort of difficult

maybe there is a better descriptor than 'public/extensible'? but that's my best idea at this point

catalog-info.yaml
```diff
@@ -177,6 +177,39 @@ spec:
...
+spec:
+  type: "buildkite-pipeline"
+  owner: "group:search-extract-and-transform"
+  system: "buildkite"
+  implementation:
+    apiVersion: "buildkite.elastic.dev/v1"
+    kind: "Pipeline"
+    metadata:
+      name: "connectors-dockerfiles-nightly"
+      description: "Nightly Connectors Dockerfiles Build"
```
acrewdson, line 196: see comment/suggestion above about `description:`

.buildkite/dockerfiles-pipeline.yml
```diff
@@ -0,0 +1,124 @@
...
+
+steps:
+  - group: ":truck: Building, Testing and Scanning Dockerfile and Dockerfile.ftest"
+    key: "build_test_scan_group"
+    if: "(build.branch == \"main\")"
+    steps:
+      # ----
+      # Dockerfile build and tests on amd64
+      # ----
+      - label: "Build amd64 image from Dockerfile"
```
acrewdson, line 30: instead of only building `amd64` images, we could consider building a multi-arch image using `buildah` or `drivah`, like [we are doing here](https://github.com/elastic/connectors/blob/main/.buildkite/release-pipeline.yml#L142-L150), and that way we could also potentially avoid using these hefty ec2 instances

catalog-info.yaml
```diff
@@ -177,6 +177,39 @@ spec:
...
+    spec:
+      pipeline_file: ".buildkite/dockerfiles-pipeline.yml"
+      provider_settings:
+        trigger_mode: "none"
+      repository: "elastic/connectors"
+      schedules:
+        Daily main:
+          branch: main
+          cronline: '@daily'
```
acrewdson, line 205: wondering if we should consider running this less frequently? I think it would depend on what @elastic/search-extract-and-transform prefer, but the scenario I picture is something like:
* this job runs once a week, or maybe even once every two weeks
* if Trivy identifies any vulnerabilities, a message is sent to a Slack channel, or a GitHub issue is created

let's see what folks think 👍

.buildkite/dockerfiles-pipeline.yml
```diff
@@ -0,0 +1,124 @@
...
+        key: "trivy-scan-dockerfile-image"
+        agents:
+          provider: k8s
+          image: "docker.elastic.co/ci-agent-images/trivy:latest"
+        command: |-
+          mkdir -p .artifacts
+          buildkite-agent artifact download '.artifacts/*.tar.gz*' .artifacts/ --step build_dockerfile_image_amd64
+          trivy --version
+          env | grep TRIVY
+          find .artifacts -type f -name '*.tar.gz*' -exec trivy image --quiet --input {} \;
```
acrewdson, line 110: nice 👍 

I'm thinking it could also be nice to add one additional stage to this pipeline, if it's not too difficult:

a smoke-test/acceptance-test that verifies that connectors actually starts up in the resultant image, just as a way of gaining confidence that the built image does what we expect it to do

.buildkite/dockerfiles-pipeline.yml
```diff
@@ -0,0 +1,124 @@
+definitions:
+  steps:
+    - step: &test-agents
+        agents:
+          provider: "gcp"
+          machineType: "n1-standard-8"
+          useVault: true
+          image: family/enterprise-search-ubuntu-2204-connectors-py
```
acrewdson, line 8: we don't need this here, right?

.buildkite/dockerfiles-pipeline.yml
```diff
@@ -0,0 +1,124 @@
...
+  retries: &retries
```
acrewdson, line 9: good to have retries, but this block is not being included in the `steps` below where it would have an effect, I don't think?

Dockerfile
```diff
@@ -1,7 +1,13 @@
-FROM python:3.11-slim-bookworm
-RUN apt -y update && apt -y upgrade && apt -y install make git
-COPY . /app
+FROM cgr.dev/chainguard/wolfi-base
+ARG python_version=3.11
+
+USER root
+RUN apk update && apk add python3=~${python_version} make git
```
acrewdson, line 5: TIL `=~` in apk, nice

acrewdson (commented):
great start! I added a few comments

Dockerfile
kostasb: @acrewdson My thought process for omitting that line was: 

- since we're going to scan the built image we'd want to be able to detect vulnerabilities on these packages, even if they're not required at runtime. 
- they've been included in the Dockerfile so removing them now can cause a "breaking" change for users who already build on top of this dockerfile. Changing the base image itself can be a much more substantial breaking change though so perhaps that shouldn't be factored in here.

Let me know what you think?

.buildkite/dockerfiles-pipeline.yml
kostasb: Thank you for the suggestion about a multi-arch image, reading in more detail how buildah/drivah [multiarch push](https://github.com/elastic/connectors/blob/main/.buildkite/release-pipeline.yml#L142C17-L142C54) works, it [pulls](https://github.com/elastic/connectors/blob/main/.buildkite/publish/build-multiarch-docker.sh#L20) the already built amd64/arm64 images from the previous steps and combines them into a common manifest which references both builds. Unfortunately doesn't save us any build resources. 

Still, a good to have - but it means that we'd build the image on both architectures. Let me know of your thoughts.

catalog-info.yaml
oli-g: > * if Trivy identifies any vulnerabilities, a message is sent to a Slack channel, or a GitHub issue is created

I think these notifications could potentially be a duplicate of existing Snyk issues detected on the official container image. What's the value added by these new notifications?

.buildkite/dockerfiles-pipeline.yml
kostasb: Correct, this was copied over from the main pipeline and is now removed - thanks for pointing it out.

.buildkite/dockerfiles-pipeline.yml
kostasb: Correct, this was copied over from the main pipeline and is now removed - thanks for pointing it out.

kostasb:
Since in this PR we are building a Dockerfile on a different base image (`cgr.dev/chainguard/wolfi-base`) there are package discrepancies comparing to the base image used in the [official built images ](https://github.com/elastic/connectors/blob/main/Dockerfile.wolfi#L1) `wolfi/python:3.11-dev`.

For reference, [here](https://github.com/wolfi-dev/os/blob/main/python-3.11.yaml#L17) is a list of packages which, among others (like bash and curl), are available in the wolfi python-3.11 image use for the production image builds.

Do we want this list of packages added to the wolfi-base image? Are connectors requiring some, or all, of them? @elastic/search-extract-and-transform

.buildkite/dockerfiles-pipeline.yml
kostasb: Good point, this pipeline includes the same test step that is used for the production docker images built from Dockerfile.wolfi [here](https://github.com/elastic/connectors/blob/kostasb/switch-dockerfile-image/.buildkite/pipeline.yml#L637) so that the testing approach is aligned on both built images. 

The pipeline step `test_extensible_dockerfile_image_amd64` runs the [.buildkite/publish/test-docker.sh](https://github.com/elastic/connectors/blob/kostasb/switch-dockerfile-image/.buildkite/extensible-dockerfiles-pipeline.yml#L37C14-L37C47) which performs basic sanity checks and verifies the output of `elastic-ingest --version` [here](https://github.com/elastic/connectors/blob/kostasb/switch-dockerfile-image/.buildkite/publish/test-docker.sh#L79).

Since adding functional tests at the connector level is a more involved task which could benefit both the production and the "extensible" image I'd suggest we create a dedicated issue for that, what do you think?

artem-shelkovnikov:
I'm not sure if we need the packages mentioned, we really only need dependencies for 3.11 python + `base64` available.

kostasb:
> I'm not sure if we need the packages mentioned, we really only need dependencies for 3.11 python + `base64` available.

Thanks, then I'll stick with the bare minimum packages and we can iterate if needed.

I added both amd64 and arm64 image builds, mainly for the purpose of smoke testing (with `.buildkite/publish/test-docker.sh`) on both architectures. I don't expect deviations in the vulnerability testing between architectures.

Do we need some sort of release note for the new base image? It could be a "breaking" change for those users who build their images on top of our Dockerfiles.

kostasb:
Update re. the packages included with the base image vs the python one that we do docker image builds on: @oli-g suggested to check whether installing `python-3.11` on the `wolfi-base` image installes those as dependencies. It does install most of them as part of the Dockerfile step that adds python (`RUN apk add --no-cache python3=~${python_version} make git`), with some exceptions (`linux-headers`) which are unlikely to be relevant for connectors. That said, there shouldn'be any concerns regarding the base image diff.

.buildkite/dockerfiles-pipeline.yml
kostasb: I added pipeline steps to build and test both amd64 and arm64 architectures. Given the above analysis I don't see a use for a multiarch image step since we aren't publishing that artifact either way, let me know if you have any concerns.

artem-shelkovnikov:
>  It could be a "breaking" change for those users who build their images on top of our Dockerfiles.

Just to check - `cgr.dev/chainguard/wolfi-base` is publicly available, right?

kostasb:
> > It could be a "breaking" change for those users who build their images on top of our Dockerfiles.
> 
> Just to check - `cgr.dev/chainguard/wolfi-base` is publicly available, right?

Yes, we've confirmed that this is a publicly available image that can be downloaded by unauthenticated (not logged in to any registry) docker instances.

artem-shelkovnikov:
> Do we need some sort of release note for the new base image? It could be a "breaking" change for those users who build their images on top of our Dockerfiles.

We can mention it in our release notes indeed. I don't expect _a lot_ of customers to build customer docker images and likely it's just gonna work for them, but it would not hurt to mention that we've updated our docker images recently to make them more secure with potential incompatible changes that the customers will have to fix

Dockerfile
seanstory: I'm ok with removing make and git. I expect very little being extended on the _image_, and more likely that the _connectors code_ is what is seeing extension. We shouldn't need make or git for that.

artem-shelkovnikov: +1. `git` is needed at build-time, but for run-time we need neither git nor make

catalog-info.yaml
seanstory: is Snyk going to scan the images produced by these CI jobs?
I think this PR is doing 3 things, but maybe doesn't need to.
1. switches our `Dockerfile` to use chainguard (has customer value)
2. Adds CI to validate the `Dockerfile` _works_ (has customer value)
3. Adds CI to fail if a vuln is found (internal value, with indirect customer value)

I think (3) could probably be done separately, and reuse whatever machinery we use to scan other artifacts, especially since (1) and (2) will significantly cut down on false-positives that we'd have if we tried to do (3) on its own today.

kostasb:
> > Do we need some sort of release note for the new base image? It could be a "breaking" change for those users who build their images on top of our Dockerfiles.
> 
> We can mention it in our release notes indeed. I don't expect _a lot_ of customers to build customer docker images and likely it's just gonna work for them, but it would not hurt to mention that we've updated our docker images recently to make them more secure with potential incompatible changes that the customers will have to fix

Thank you, I added a release note in the issue description.

catalog-info.yaml
acrewdson: > What's the value added by these new notifications?

... 

> which is technically the exact same base OS build.

These are really good points that I hadn't considered. Given that with this PR we are:
* switching the `Dockerfile` to become distroless, and making the resultant image considerably more secure
* making the `Dockerfile` essentially the same as the `Dockerfile.wofli`, the latter of which produces the images that are pushed and scanned by Snyk
* adding nothing in the Dockerfile except the bare minimum that we need to run connectors (assuming we get rid of `git` and `make` 👍)

... then I actually don't see much value in adding the Trivy step (let alone adding notifications based on its results). I wish I'd had these realizations sooner, apologies @kostasb. 

What do you all think?

.buildkite/dockerfiles-pipeline.yml
acrewdson: seems good to me 👍

.buildkite/dockerfiles-pipeline.yml
acrewdson: > The pipeline step test_extensible_dockerfile_image_amd64 runs the [.buildkite/publish/test-docker.sh](https://github.com/elastic/connectors/blob/kostasb/switch-dockerfile-image/.buildkite/extensible-dockerfiles-pipeline.yml#L37C14-L37C47)

ah I'd missed that, thank you for pointing it out 👍 

>  I'd suggest we create a dedicated issue for that, what do you think?

I'd say no need for another issue at this point; adding more rigorous image acceptance testing sounds like work that would probably need to come from E&T folks if they saw the necessity for it 

so I think we are well covered with respect to the scope of this PR

Dockerfile
kostasb: Thank you for your inputs! I take from this conversation that it's best to remove these packages as unnecessary during runtime. The `RUN apk del make git` step has been added back.

kostasb: Followed the same approach for `Dockerfile.ftest` which also shouldn't need `git` and `make`.

catalog-info.yaml
acrewdson: @kostasb thanks, I like the idea of leaving the Trivy step there, but not having it send any notifications. But I'll defer to E&T folks (@seanstory @artem-shelkovnikov) to say what they prefer 👍

seanstory: I have no objection to leaving it (without notifications) but also I don't think a human will be manually looking at the outputs often. So if we're ok as treating it essentially as a source of debug logs that may or may not be examined, great.

kostasb: Merging the extensible Dockerfile build/test/scan steps into the main [pipeline](https://github.com/elastic/connectors/blob/main/.buildkite/pipeline.yml) makes sense to me, we can proceed this way if the team agrees.

kostasb: I pushed a commit that merges the extensible pipeline into the main pipeline (easy to revert if the team @seanstory @artem-shelkovnikov has a different opinion after all).

As a bonus point, this also enables test the steps as part of this PR (the dedicated pipeline wouldn't be picked up until merged).

kostasb:
The pipeline steps have been [tested successfully](https://buildkite.com/elastic/connectors/builds/15698).

The extensible Dockerfile and Dockerfile.ftest steps are added as dependencies for the main pipeline, so the build will fail if they don't work, making them part of the release process.

As [discussed](https://github.com/elastic/connectors/pull/3063#discussion_r1905846389) Trivy scans are included for informative and logging purposes but won't cause the pipeline to fail, since we're getting alerts on equivalent images from Snyk.

I am planning to tag this against 8.18.0 without backporting , because I consider it to be too big of a change for a patch release and we want to include a clear release note with it.

Let me know if you have any objections or comments.

acrewdson (dismissed):
great work 🚀 

I'll let E&T folks review and give the go-ahead on merging

thank you for doing this!

.buildkite/pipeline.yml
```diff
@@ -647,11 +647,211 @@ steps:
...
+        key: "trivy-scan-amd64-extensible-dockerfile-ftest-image"
+        agents:
+          provider: k8s
+          image: "docker.elastic.co/ci-agent-images/trivy:latest"
+        command: |-
+          mkdir -p .artifacts
+          buildkite-agent artifact download '.artifacts/*.tar.gz*' .artifacts/ --step build_extensible_dockerfile_ftest_image_amd64
+          trivy --version
+          env | grep TRIVY
+          find .artifacts -type f -name '*.tar.gz*' -exec trivy image --quiet --input {} \;
```
oli-g, line 742: I think we are not going to get a lot of value from these steps, given that `Dockerfile.ftest` is almost equivalent to `Dockerfile`.

So I'm wondering: what if instead (maybe in a different PR, as a follow-up) we delete the two `.ftest` Dockerfiles and we try to live with only two Dockerfiles instead of four (`Dockerfile` and `Dockerfile.wolfi`), and we move the `RUN .venv/bin/pip install -r requirements/ftest.txt` command out of the Dockerfiles, where we actually run those specs?

.buildkite/pipeline.yml
```diff
@@ -647,11 +647,211 @@ steps:
...
+        key: "trivy-scan-arm64-extensible-dockerfile-ftest-image"
+        agents:
+          provider: k8s
+          image: "docker.elastic.co/ci-agent-images/trivy:latest"
+        command: |-
+          mkdir -p .artifacts
+          buildkite-agent artifact download '.artifacts/*.tar.gz*' .artifacts/ --step build_extensible_dockerfile_ftest_image_arm64
+          trivy --version
+          env | grep TRIVY
+          find .artifacts -type f -name '*.tar.gz*' -exec trivy image --quiet --input {} \;
```
oli-g, line 844: Similarly, I don't think we are going to get a lot of value from building and testing and scanning the extensible `Dockerfile.ftest` artifact, since it's totally similar to the extensible `Dockerfile` artifact.

.buildkite/pipeline.yml
kostasb: That makes sense, I'll remove the ftest builds/tests/scans from this pipeline and follow up with an issue about replacing/removing the ftest dockerfiles.

kostasb: Created an issue for it: https://github.com/elastic/search-developer-productivity/issues/3611

kostasb:
> I think as a user it would be clearer to read `oss` instead of `extensible`: what do you all think? I see other Elastic products doing something similar, see [here](https://github.com/elastic/logstash/tree/ff44b7cc2073aab5a03bc721644b8c7768c49d7f/docker/data/logstash/config) for example.
> 
> I also left a couple of comments related to the following: I'm proposing to not build, test and scan the artifacts produced from `Dockerfile.ftest`, as they're too similar to the artifacts produced from `Dockerfile`.

I agree and adapted the PR to remove ftest from the pipeline and refer to the Dockerfile as OSS instead of extensible.
I share the same thoughts regarding ftest dockerfiles so we can follow up to replace them under a dedicated [issue](https://github.com/elastic/search-developer-productivity/issues/3611).

oli-g (approved):
Thank you for this work!

kostasb:
Let me know if any objections with merging this PR towards 8.18.0 without backporting to previous minors so that it doesn't show up in a patch release. I consider this a potential breaking change for some users, as reflected by the release note in issue description.

If no concerns until then, I plan to merge on Monday, Jan 20.

PR #3072, opened 2025-01-06
Make a CI step to verify and update NOTICE.txt file

artem-shelkovnikov:


jedrazb:
@artem-shelkovnikov LGTM but  the bot's commit has ~7k deletion with only +800 additions to Notice file. It seems like we lost something from that file, is this expected?

Screenshot 2025-01-07 at 14 16 13

artem-shelkovnikov:
@jedrazb I think it's correct, because notice file was always pushed manually.

The way NOTICE.txt is generated is that we run notice generator against our whole distribution in .venv/bin. So if you, for example, ran tests with `make test`, it'll also put test deps there, but we should not. Our NOTICE should only rely on the files needed to run connectors.

For example if you look at the changeset, you can see that `Faker` was deleted from it, and it's correct - it's only there for tests. Also it removed `azure.identity` which I had locally installed on a branch, but since I did not do clean install it stayed there after I switched away from the branch. So I think this diff is correct

jedrazb (approved):
LGTM

PR #3075, opened 2025-01-06
Feature: Google Drive - Presentation to Markdown / Audio|Video Transcribed (10MB Limit)

that-dom:
## Closes https://github.com/elastic/connectors/issues/3073 & https://github.com/elastic/connectors/issues/3074 ###

In the Google Drive source this adds the functionality to convert Presentation to Markdown using Azure's OpenAI models and using Whisper service to transcribe audio or video to text with the limit of 10MB. 

#### Pre-Review Checklist

- [x] New external service dependencies added. 

This adds external services for Azure hosted OpenAI GPT4-mini and Whisper API's


seanstory (changes requested):
Looks like you may need to rebase and force-push - you're pulling in a lot of diff that is not yours. 

Also, it looks like you're tightly coupling the Google Drive connector to an Azure AI tool. Which doesn't make a lot of sense to me.

I don't see this being mergable. If you'd like to have a chat with the team in #connectors-feedback, we can discuss how we can work to better integrate with LLM tools, but this diff as I understand it does not align with our vision for the architecture.

artem-shelkovnikov (changes requested):
+1 to Sean's points.

If we want similar functionallity, specific connectors are not the place to do it. Should be either supported framework-wise, or on later stages such as ingest pipelines.

that-dom:
Thanks for the review, and I agree with all points. I am glad to see that this is kicking a more considerable discussion because markdown and text are the future for any GenAI use cases. 😄

PR #3076, opened 2025-01-07
Make lint step run autoformat and commit changes for you + minor refactoring

artem-shelkovnikov:
This PR attempts to run autoformat if possible on linting stage

Actual changes:
- If on a branch with a PR, linting stage will run `make autoformat` and commit changes if it fixes linting failure
- Notice check now runs first and all other commands wait for it to finish. It's done because both linting and notice check stages can do autofix + push, and it'll cause conflicts and weird job restarts. The `wait` step ensures we don't run both steps at the same time. This makes CI a bit slower though
- Cleans up shell scripts a bit to re-use as much code as possible + delete broken/unused code


navarone-feekery (approved):
Very cool!

navarone-feekery (approved):

PR #3077, opened 2025-01-07
[Agent]  Don't auto-create connector doc if connector was deleted

jedrazb:
## Closes https://github.com/elastic/search-team/issues/9010

Soft deletes supported as of ES 9 ✅ https://github.com/elastic/elasticsearch/pull/118669

### Changes
If connector has been soft deleted, do not autocreate the connector record by the agent component logic. 

Fix agentdocker file, as agent image is now based of UBI. 

### Validation
- Add integration
- ✅ connector is autocreated
- Soft delete connector,  restart the docker img 
- ✅ Connector is not auto created 
- Hard delete connector, restart the docker img
- ✅ connector is autocreated

I also found this bug while testing: https://github.com/elastic/connectors/issues/3118 this should be addressed separately

#### Pre-Review Checklist


connectors/es/index.py
```diff
@@ -27,6 +27,16 @@ class TemporaryConnectorApiWrapper(ESClient):
     def __init__(self, elastic_config):
         super().__init__(elastic_config)
 
+    async def connector_get(self, connector_id, include_deleted):
+        include_deleted_param = (
+            f"?include_deleted={'true' if include_deleted else 'false'}"
```
jedrazb, line 32: It took me way too long to realise python's bools are uppercased ....

Dockerfile.agent
```diff
@@ -5,16 +5,17 @@ FROM docker.elastic.co/elastic-agent/elastic-agent:9.0.0-SNAPSHOT
...
+# Install basic dependencies
+RUN microdnf update && microdnf install -y \
   vim \
   wget \
   git \
   make \
-  && add-apt-repository ppa:deadsnakes/ppa \
-  && apt-get update && apt-get install -y python3.11 python3.11-venv \
-  && apt-get clean && rm -rf /var/lib/apt/lists/*
+  gcc \
```
jedrazb, line 14: lemme check if we need it

connectors/protocol/connectors.py
```diff
@@ -179,18 +183,23 @@ async def connector_put(
...
             logger.error(
                 f"Error while checking existence of connector '{connector_id}': {e}"
             )
             raise e
 
+    async def get_connector(self, connector_id, include_deleted=False):
+        return await self.api.connector_get(
+            connector_id=connector_id, include_deleted=include_deleted
+        )
+
```
artem-shelkovnikov, line 202: This method should be inlined

connectors/protocol/connectors.py
```diff
@@ -179,18 +183,23 @@ async def connector_put(
...
         )
 
-    async def connector_exists(self, connector_id):
+    async def connector_exists(self, connector_id, include_deleted=False):
         try:
-            doc = await self.fetch_by_id(connector_id)
+            doc = await self.get_connector(connector_id, include_deleted)
             return doc is not None
-        except DocumentNotFoundError:
+        except ElasticNotFoundError:
```
artem-shelkovnikov, line 190: What's the difference with `DocumentNotFoundError`?

connectors/agent/connector_record_manager.py
```diff
@@ -64,6 +66,10 @@ async def ensure_connector_records_exist(self, agent_config, connector_name=None
                         f"Failed to create connector record for {connector_id}: {e}"
                     )
                     raise e
+            else:
+                logger.debug(
+                    f"Skipping connector creation. Connector record for {connector_id} already exists."
+                )
```
artem-shelkovnikov, line 72: Skipping connector creation is one thing, but should we also terminate early if we find out that connector was deleted?

What would happen in case connector is deleted mid-way through a sync?
What would happen in case connector has been deleted and host starts again and sees it's deleted?

connectors/es/index.py
artem-shelkovnikov: Is it possible to pass params as hash and do generic encoding in `perform_request`?

jedrazb: Good callout! Yes it's [possible](https://github.com/elastic/elasticsearch-py/blob/14f22ae202d548caf9f6b620fb6fc73f1e224667/elasticsearch/_sync/client/_base.py#L312) and it will take care of lowercasing the bool ([reference](https://github.com/elastic/elasticsearch-py/blob/14f22ae202d548caf9f6b620fb6fc73f1e224667/elasticsearch/_sync/client/utils.py#L279))

Dockerfile.agent
jedrazb: no need for it, removing!

connectors/protocol/connectors.py
jedrazb: For now it can be inlined 👍 at some point we will need to revisit this to be able to do `connector_index.get` (same as `connector_index.fetch_by_id` to replace fetch_by_id with get API call

connectors/protocol/connectors.py
jedrazb: `ElasticNotFoundError` is a `NotFound` error from `elasticsearch` package. We do `from elasticsearch import ( NotFoundError as ElasticNotFoundError)` to make it explicit that it comes from es client.

Framework implements it's own [DocumentNotFoundError](https://github.com/elastic/connectors/blob/b1d4f221a6afa1c7eb1095d451dfeb015f4975a4/connectors/es/index.py#L16) but IMO we should stay consistent with ES client error defs, wdyt? Otherwise I would need to catch not found just to use our own error class. They are not compatible.

 For now it's just used for this endpoint

connectors/agent/connector_record_manager.py
jedrazb: > Skipping connector creation is one thing, but should we also terminate early if we find out that connector was deleted?

If we terminate the process agent will restart it anyway (from what I observed), we should just stay in this state waiting for the agentless resource to be cleaned up

> What would happen in case connector is deleted mid-way through a sync?

I don’t think this changes anything compared to the current (elastic-managed/self-managed) behavior:
- If the connector is deleted but sync jobs are not (default behavior), the sync will likely continue since each sync job stores a copy of the connector document. In this case, the sync will run until the agentless resource is terminated by the connector manager’s background task.
- If the connector is deleted and the sync jobs are also removed (via [API](https://www.elastic.co/guide/en/elasticsearch/reference/current/delete-connector-api.html#delete-connector-api-path-params)), the sync will crash immediately.

> What would happen in case connector has been deleted and host starts again and sees it's deleted?

The service would start and identify the single connector in the config.yml file that it no longer exists (once we fix https://github.com/elastic/connectors/issues/3118 lol). This would likely throw some warning/error logs, but eventually, we assume the connector manager’s background task should handle the cleanup of the service.

artem-shelkovnikov (approved):
Should be okay if everything's manually tested

artem-shelkovnikov (approved):

PR #3089, opened 2025-01-09
Minor CI lint/notice steps improvement

artem-shelkovnikov:
1. Fix `is_pr` function to properly tell if job is running on top of PR or not
2. Add `git diff` to notice/lint stages for debugging

### DIFF
--- a/.buildkite/shared.sh
+++ b/.buildkite/shared.sh
@@ -29,8 +29,8 @@ retry() {
 
 is_pr() {
   if [ "${BUILDKITE_PULL_REQUEST-}" ] || [ "${BUILDKITE_PULL_REQUEST}" = "false"]; then
-    return false
+    return 1 # false
   else
-    return true
+    return 0 # true
   fi
 }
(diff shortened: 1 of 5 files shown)


.buildkite/shared.sh
```diff
@@ -28,8 +28,11 @@ retry() {
 }
 
 is_pr() {
-  [[ "${BUILDKITE_PULL_REQUEST-}" ]] && return
-  false
+  if [ "${BUILDKITE_PULL_REQUEST-}" ] || [ "${BUILDKITE_PULL_REQUEST}" = "false"]; then
+    return true
+  else
+    return false
```
seanstory, line 34: I think you've got these backwards?

artem-shelkovnikov: Yup, that actually helped me test it on the PR, just in case

seanstory (dismissed):
LGTM. We'll at least have more info if main keeps failing

seanstory (approved):

PR #3090, opened 2025-01-09
Remove ' ' as a configured Confluence space when there's a trailing comma

meghanmurphy1:
## Closes https://github.com/elastic/connectors/issues/2643

A fix for when a trailing comma is used as part of 'Confluence space keys' when configuring a Confluence Connector.

#### Pre-Review Checklist

### DIFF
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -146,7 +146,7 @@ class Field:
                     for item in value.split(","):
                         item = item.strip()
                         if not item:
-                            logger.warning(
+                            logger.debug(
                                 "Empty string detected in the comma-separated list. It will be skipped."
                             )
                         else:
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -141,7 +141,7 @@ class Field:
         # list requires special type casting
         if cast_type is list:
             if isinstance(value, str):
-                return [item.strip() for item in value.split(",")] if value else []
+                return [item.strip() for item in value.rstrip(",").split(",")] if value else []
             elif isinstance(value, int):
                 return [value]
             elif isinstance(value, set):
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -142,7 +142,11 @@ class Field:
         if cast_type is list:
             if isinstance(value, str):
                 return (
-                    [item.strip() for item in value.rstrip(",").split(",")]
+                    [
+                        item.strip()
+                        for item in value.rstrip(",").split(",")
+                        if item.strip()
+                    ]
                     if value
                     else []
                 )
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -141,7 +141,11 @@ class Field:
         # list requires special type casting
         if cast_type is list:
             if isinstance(value, str):
-                return [item.strip() for item in value.rstrip(",").split(",")] if value else []
+                return (
+                    [item.strip() for item in value.rstrip(",").split(",")]
+                    if value
+                    else []
+                )
             elif isinstance(value, int):
                 return [value]
             elif isinstance(value, set):
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -141,15 +141,15 @@ class Field:
         # list requires special type casting
         if cast_type is list:
             if isinstance(value, str):
-                return (
-                    [
-                        item.strip()
-                        for item in value.rstrip(",").split(",")
-                        if item.strip()
-                    ]
-                    if value
-                    else []
-                )
+                items=[]
+                if value:
+                    for item in value.rstrip(",").split(","):
+                        item = item.strip()
+                        if not item:
+                            logger.warning("Empty string detected in the comma-separated list. It will be skipped.")
+                        else:
+                            items.append(item)
+                return items
             elif isinstance(value, int):
                 return [value]
             elif isinstance(value, set):
--- a/connectors/source.py
+++ b/connectors/source.py
@@ -141,12 +141,14 @@ class Field:
         # list requires special type casting
         if cast_type is list:
             if isinstance(value, str):
-                items=[]
+                items = []
                 if value:
                     for item in value.rstrip(",").split(","):
                         item = item.strip()
                         if not item:
-                            logger.warning("Empty string detected in the comma-separated list. It will be skipped.")
+                            logger.warning(
+                                "Empty string detected in the comma-separated list. It will be skipped."
+                            )
                         else:
                             items.append(item)
                 return items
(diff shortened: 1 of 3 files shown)


seanstory:
Great find! I agree that this will address the problem reported in the issue.

Let's add a new test though. While we don't need the test to prove this works (it's simple enough a change that it's pretty obvious it'll work), adding a test will protect us against a regression in case this line/logic is ever broken in the future.

You should be able to add just one pytest function to https://github.com/elastic/connectors/blob/main/tests/sources/test_confluence.py like:
```
@pytest.mark.asyncio
async def test_fetch_spaces_with_configured_trailing_comma():
  # setup here
  # assertions here
```

However, I'd bet that this bug goes deeper. Since any `"type": "list"` configured filed like `spaces` (see https://github.com/elastic/connectors/blob/c664243a030cd14638e5ac9925b5fb97c28eb591/connectors/sources/confluence.py#L517-L522) is represented by comma-separated strings, this probably means that this bug is lying in wait for other connectors too. I suggest looking into the `Field` class (see https://github.com/elastic/connectors/blob/main/connectors/source.py#L77) to see if we can eliminate this issue for list-type fields.

meghanmurphy1:
ahh ok I see now where the string gets parsed. Thanks!

jedrazb (commented):
LGTM, but would it would be good to have unit tests to ensure this keeps working correctly, I can see `_convert` unit tests in https://github.com/elastic/connectors/blob/main/tests/test_source.py#L77

artem-shelkovnikov:
buildkite test this

artem-shelkovnikov:
☝️ Do we need to do changes in backstage so that Meghan/Matt would trigger CI too?

artem-shelkovnikov:
Ah sorry, I see you're forking the repository @meghanmurphy1. To make CI run automatically you need to make PRs in this repository, not in forks. I'll fix CI breaking on the above in the meantime

meghanmurphy1:
> Ah sorry, I see you're forking the repository @meghanmurphy1. To make CI run automatically you need to make PRs in this repository, not in forks. I'll fix CI breaking on the above in the meantime

Ohhh ok. Thanks for the fix!

meghanmurphy1:
> LGTM, but would it would be good to have unit tests to ensure this keeps working correctly, I can see `_convert` unit tests in https://github.com/elastic/connectors/blob/main/tests/test_source.py#L77

Yes! Currently going through how the connectors unit tests work RN. Should these tests be in `test_source.py` or `test_confluence.py` you think?

artem-shelkovnikov:
buildkite test this

jedrazb (approved):
LGTM

jedrazb:
(this ^ is magic cmnd to trigger checks in forks)

seanstory:
CI is mad because it's not up-to-date with `main`. If you click the "update branch" button, that _should_ fix it 🤞 
(you'll then need to tell it to "test this" again)

meghanmurphy1:
buildkite test this

meghanmurphy1:
buildkite test this

connectors/source.py
```diff
@@ -141,7 +141,11 @@ def _convert(self, value, field_type_):
         # list requires special type casting
         if cast_type is list:
             if isinstance(value, str):
-                return [item.strip() for item in value.split(",")] if value else []
+                return (
+                    [item.strip() for item in value.rstrip(",").split(",")]
+                    if value
+                    else []
+                )
```
artem-shelkovnikov, line 148: Would it make sense to move this function definition to utils module as we have comma-separated values in many connectors?

This way we can:
1. Re-use it
2. Have unit-tests for this logic outside of other tests.

connectors/source.py
```diff
@@ -141,7 +141,11 @@ def _convert(self, value, field_type_):
         # list requires special type casting
         if cast_type is list:
             if isinstance(value, str):
-                return [item.strip() for item in value.split(",")] if value else []
+                return (
+                    [item.strip() for item in value.rstrip(",").split(",")]
```
artem-shelkovnikov, line 145: Would it work for cases like these ones?

- `1, 2, , , , ,5`
- `,,,,3`
- `,1,,2,`
- `1,2,3,    `

I known, they are nonsensical, but possible and it's great to give user a warning if we were unable to properly parse it and had to truncate their input.

E.g.:

> Field "site_collections" contained blank strings as some of comma-separated values, skipping them.

meghanmurphy1: I think adding a warning that they're being skipped sounds good to me! 

I see this happening more `1,2,3,    ` so I can add an extra `.strip()` to remove the whitespace, but that would just help with this case. May be more cohesive to just not fix the million way a blank string could end up in the list, and somehow skip them. That also begs the question though would there be an instance where the blank string is intended?

meghanmurphy1:
buildkite test this

connectors/source.py
artem-shelkovnikov: We don't really have good input validation in the frontend for this, so it's just easy to put something invalid into the input - they are likely never intended, but can just appear, especially if you're entering a list with 10+ items, it's just easy to make a mistake.

> I think adding a warning would be good, but unsure where that would go? 

Yes, warning is good. It should happen in this method IMO - you'll need to change the code to `for` loop for this likely, so that you can warn. But it helps readability IMO, now the list comprehension is quite confusing already 😅 

> And would it be a logger.warning?

Yes!

meghanmurphy1: Sounds good! - I agree the one-liner list comprehension has gotten pretty mind-boggling 😵‍💫

meghanmurphy1:
buildkite test this

meghanmurphy1:
buildkite test this

connectors/source.py
```diff
@@ -141,7 +141,17 @@ def _convert(self, value, field_type_):
         # list requires special type casting
         if cast_type is list:
             if isinstance(value, str):
-                return [item.strip() for item in value.split(",")] if value else []
+                items = []
+                if value:
+                    for item in value.rstrip(",").split(","):
+                        item = item.strip()
+                        if not item:
+                            logger.warning(
```
seanstory, line 149: Nit: let's make this `debug`. Validation occurs every ~30 seconds, so this could lead to a lot of logs by default.

Alternatively, if we want to really indicate that this isn't acceptable, we should fail validation if we detect null/empty elements in the lists.

meghanmurphy1: Oh true- it does seem to populate the logs alot. IMO I think just skipping over them would be ok? Seems like these blank strings are usually just a slip of the hand.

meghanmurphy1:
buildkite test this

seanstory (approved):
LGTM. Nice work!

PR #3094, opened 2025-01-10
Don't run CI github actions that commit/push on forks

artem-shelkovnikov:
Subj:

Our machine might not have access to forks, so let's not do git commands on it and just fail steps if notice/autoformat fails.


artem-shelkovnikov:
Gonna merge from a fork: https://github.com/elastic/connectors/pull/3095

PR #3095, opened 2025-01-10
Avoid git commands when running against a fork

artem-shelkovnikov:
See failure of CI in PR https://github.com/elastic/connectors/pull/3090: https://buildkite.com/elastic/connectors/builds/15589#01945040-19d0-4239-915a-173479f8aee8

```
++ git switch -
--
  | Warning: you are leaving 2 commits behind, not connected to
  | any of your branches:
  |  
  | 824f57b6 use rstrip for lists to remove whitespace
  | cf502ec0 remove '' as a space with trailing comma
  |  
  | If you want to keep them by creating a new branch, this may be a good time
  | to do so with:
  |  
  | git branch <new-branch-name> 824f57b6
  |  
  | Switched to branch 'main'
  | Your branch is up to date with 'origin/main'.
  | ++ git checkout meghanmurphy1:confluence-trailing-comma
  | error: pathspec 'meghanmurphy1:confluence-trailing-comma' did not match any file(s) known to git
  | 🚨 Error: The command exited with status 1
  | user command error: exit status 1
```

I happens because forks are not known to Buildkite agent. Forks will not by default give permissions to Elasitc agents, so we cannot really push anything into them. This PR just makes forks treated same way we treat `main` branch: just report failures, don't try to fix them.

### DIFF
--- a/.buildkite/shared.sh
+++ b/.buildkite/shared.sh
@@ -34,3 +34,11 @@ is_pr() {
     return 0 # true
   fi
 }
+
+is_fork() {
+  if [ "BUILDKITE_PULL_REQUEST_REPO" = "https://github.com/elastic/connectors.git"]; then
+    return 1 # false
+  else
+    return 0 # true
+  fi
+}
--- a/.buildkite/run_linter.sh
+++ b/.buildkite/run_linter.sh
@@ -30,10 +30,9 @@ if is_pr && ! is_fork; then
     exit 1
   fi
 else
-  echo "We're not on PR, running only linter"
+  echo "We're not on PR or running against a fork, running only linter"
   # On non-PR branches the bot has no permissions to open PRs.
   # Theoretically this would never fail because we always ask
   # linter to succeed to merge. It can fail intermittently?
   make lint
-  return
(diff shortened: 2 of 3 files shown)


artem-shelkovnikov:
buildkite test this

artem-shelkovnikov:
buildkite test this

jedrazb (approved):
🚢

.buildkite/shared.sh
```diff
@@ -34,3 +34,11 @@ is_pr() {
     return 0 # true
   fi
 }
+
+is_fork() {
+  if [ "BUILDKITE_PULL_REQUEST_REPO" = "https://github.com/elastic/connectors.git"]; then
```
erikcurrin-elastic, line 39: 1 is false?

artem-shelkovnikov: Fun unix stuff :D Functions actually return exit codes, not regular variables:

https://stackoverflow.com/questions/17336915/return-value-in-a-bash-function

So 0 = "all good", non-0 are "something is wrong"

PR #3110, opened 2025-01-15
Updating ESApi class methods to use ES client calls instead of temporary wrapper methods

mattnowzari:
## Closes https://github.com/elastic/connectors/issues/2964

This PR updates the few methods of `class ESApi` to utilize ES client methods instead of `TemporaryConnectorAPIWrapper` methods. This does not fully replace all of the methods of `TemporaryConnectorAPIWrapper`, as we are waiting on ES client v8.17.1 to support `sync_job_claim` and `sync_job_update_stats` API calls. This will be attended to post-8.17.1 release.

#### Pre-Review Checklist


connectors/es/index.py
```diff
@@ -35,66 +35,6 @@ async def connector_get(self, connector_id, include_deleted):
             params={"include_deleted": include_deleted},
```
mattnowzari, line 35: Leaving connector_get in for now because 8.17.1 version of Python client does not yet support include_deleted field

tests/es/test_index.py
```diff
@@ -255,3 +255,166 @@ async def test_get_all_docs(mock_responses):
     assert doc_count == total
 
     await index.close()
+
+
```
erikcurrin-elastic, line 259: I see that we mock the calls to the API, call the API, and then check that the mock was called with the params we specified.  It does not feel this tests anything other than the signature of the method.  Is this typical in our tests?

mattnowzari: I saw a few tests that behave this way ([here](https://github.com/elastic/connectors/blob/5c29055fb8eb416120abfbb4301712072be7c212/tests/protocol/test_connectors.py#L780) is an example of one), and I figured that because ESApi is simply wrapping the ES client, we could simply have coverage to make sure we are wrapping those classes correctly.

I can totally go back to the drawing board with these if we wanted to make them do more, though!

connectors/es/index.py
artem-shelkovnikov: Can you create a follow-up issue so that we don't forget?

tests/es/test_index.py
artem-shelkovnikov: We do test like this sometimes, alternative is testing with mocking HTTP requests: https://github.com/elastic/connectors/blob/main/tests/test_sink.py#L95

connectors/es/index.py
mattnowzari: Follow-up issue --> https://github.com/elastic/search-team/issues/9172

mattnowzari:
buildkite test this

tests/es/test_index.py
erikcurrin-elastic: I feel like the later we can mock, the better.  Otherwise, we are testing things which are not likely to reveal a bug.  The questions is why do we write the test?  It should be to do sanity checking and to protect ourselves against an incompatible change, not just for the sake of tests.

mattnowzari:
buildkite test this

tests/es/test_index.py
erikcurrin-elastic: OK.  If something changed in the underlying call and this test fails, it is worthy to keep.  Your call

jedrazb: I agree with Matt, these tests follow the pattern found in this repo and ensure that we pass arguments correctly to the client from our wrapper methods, which add the retry logic.

For the end-to-end scenario, we have a suite of functional tests that spin up an ES cluster and use these client calls to communicate with ES. 

So, this code will ultimately be tested during the functional test step anyway.

jedrazb (approved):
LGTM, all of those client calls should be tested by the smoke test and functional tests in our CI pipeline. 

The only exception is the code responsible for the filtering validation:
Can you follow docs and try setting up [advanced sync rules](https://www.elastic.co/guide/en/elasticsearch/reference/current/es-sync-rules.html#:~:text=Advanced%20sync%20rules,-edit&text=If%20an%20advanced%20sync%20rule,queries%20to%20the%20data%20source.) for a connector that is managed by connector service from this branch, you can find some examples in this [tutorial](https://www.elastic.co/search-labs/blog/elasticsearch-connector-apis-sync-data-to-serverless)

If you manage add sync rule and see that framework is able to validate it or detect errors than feel free to merge

mattnowzari:
## 💔 Some backports could not be created

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.x|[](https://github.com/elastic/connectors/pull/3159)|
|❌|8.17|Conflict resolution was aborted by the user|

### Manual backport
To create the backport manually run:
```
backport --pr 3110
```

### Questions ?

PR #3134, opened 2025-01-22
Add pagerduty service link

seanstory:
See the email: `[ACTION REQUIRED] Backstage Linking to PagerDuty Service`.


NOTICE.txt
```diff
@@ -4511,7 +4511,7 @@ THE SOFTWARE.
 
 
 grpcio
-1.69.0
```
artem-shelkovnikov, line 4514: This doesn't look right

artem-shelkovnikov: I checked it and we need to re-release python-elastic-agent-client, will do today

artem-shelkovnikov: Should be fixed by https://github.com/elastic/connectors/pull/3150

artem-shelkovnikov (approved):

navarone-feekery (approved):

PR #3154, opened 2025-01-27
Create build.properties file during build

navarone-feekery:
#### Pre-Review Checklist

### DIFF
--- a/.buildkite/publish/publish-common.sh
+++ b/.buildkite/publish/publish-common.sh
@@ -19,9 +19,29 @@ VERSION_PATH="$PROJECT_ROOT/connectors/VERSION"
 export VERSION=$(cat $VERSION_PATH)
 
 if [[ "${USE_SNAPSHOT:-}" == "true" ]]; then
-  echo "Adding SNAPSHOT labeling"
+  echo "Adding √SNAPSHOT labeling"
   export VERSION="${VERSION}-SNAPSHOT"
+  version_qualifier=""
+elif [[ -n "${VERSION_QUALIFIER:-}" ]]; then
+  echo "Adding version qualifier labeling"
+  # SNAPSHOT workflows ignore version qualifier
+  export VERSION="${VERSION}-${VERSION_QUALIFIER}"
+  version_qualifier="${VERSION_QUALIFIER}"
 fi
+
+REVISION="$(git rev-parse HEAD)"
+REPOSITORY="$(git config --get remote.origin.url)"
+
+# Create a build.properties file for reference during build process
+cat <<EOL > build.properties
+version=$VERSION
+qualifier=$version_qualifier
+revision=$REVISION
+repository=$REPOSITORY
+EOL
+
+echo "Created build.properties file:"
+cat build.properties
 
 if [[ "${MANUAL_RELEASE:-}" == "true" ]]; then
   # This block is for out-of-band releases, triggered by the release-pipeline


.buildkite/publish/publish-common.sh
```diff
@@ -18,11 +18,34 @@ source $SCRIPT_DIR/git-setup.sh
...
+version="${VERSION}"
+version_qualifier="${VERSION_QUALIFIER:-}"
+revision="$(git rev-parse HEAD)"
+repository="$(git config --get remote.origin.url)"
+
 if [[ "${USE_SNAPSHOT:-}" == "true" ]]; then
+  # SNAPSHOT workflows ignore version qualifier
   echo "Adding SNAPSHOT labeling"
   export VERSION="${VERSION}-SNAPSHOT"
+  version="${VERSION}"
```
seanstory, line 31: oh this is nice, now we'll get `*-SNAPSHOT` output for snapshot artifacts too! :feelsgood:

.buildkite/publish/publish-common.sh
```diff
@@ -18,11 +18,34 @@ source $SCRIPT_DIR/git-setup.sh
...
+elif [[ -n "${VERSION_QUALIFIER:-}" ]]; then
+  # version qualifier should not append version in build.yaml
+  echo "Adding version qualifier labeling"
+  version="${VERSION}-${VERSION_QUALIFIER}"
 fi
 
+# Create a build.yaml file for reference after build process
+cat <<EOL > connectors/build.yaml
```
seanstory, line 39: This file will end up being written every time a script calls `source publish-common.sh`. I think that's ok, but wanted to call it out in case that wasn't intended, and it might have negative side effects. Fore example, this file is sourced during the DRA publish steps, not just during the image building steps

.buildkite/publish/publish-common.sh
```diff
@@ -18,11 +18,34 @@ source $SCRIPT_DIR/git-setup.sh
...
+version: "$version"
+qualifier: "$version_qualifier"
+revision: "$revision"
+repository: "$repository"
```
seanstory, line 43: Does this result in literal quote chars in the YAML file? If not, I suggest you add escaped quotes too, so that these do get quoted. Just to avoid any nasty bugs where a repository name has YAML special characters, or some shit.

connectors/__init__.py
```diff
@@ -5,5 +5,17 @@
 #
 import os
 
+import yaml
+
 with open(os.path.join(os.path.dirname(__file__), "VERSION")) as f:
     __version__ = f.read().strip()
+
+yaml_path = os.path.join(os.path.dirname(__file__), "build.yaml")
```
seanstory, line 13: For our sanity later, can you add a comment here with an explanation of where this file comes from? Maybe even a link to this PR?

seanstory (commented):
Awesome stuff. The example outputs from the PR description look perfect. 👨‍🍳 👌

.buildkite/publish/publish-common.sh
navarone-feekery: It results in literal quote chars. The file that gets build looks like this:

```
version: "9.0.0-beta1"
qualifier: "beta1"
revision: "27721f59f93e21980b81d5322248d2e6238b5aa2"
repository: "git@github.com:elastic/connectors-python.git"
```

navarone-feekery: It also means empty values will look like this:

```
version: "9.0.0"
qualifier: ""
revision: "27721f59f93e21980b81d5322248d2e6238b5aa2"
repository: "git@github.com:elastic/connectors-python.git"
```

.buildkite/publish/publish-common.sh
navarone-feekery: I hadn't considered this, so that's actually unintentional. If we decided that we do care about that, do you think this should be moved to another file, so we can control its importing better?

seanstory: I think this logic _could_ be moved to `build-docker.sh`, but I don't see any obvious risks with where it is. If you don't see it hurting anything either, I'm find to leave it here for now.

.buildkite/publish/publish-common.sh
seanstory: fantastic

seanstory (approved):

jedrazb (approved):
LGTM

PR #3157, opened 2025-01-28
Don’t auto-open PRs to update cgr.dev images

oli-g:
As a follow-up to https://github.com/elastic/connectors/pull/3063, we want to unpin digest in Dockerfiles for external users, and un-manage those Dockerfiles in Renovate. We want to avoid PRs like https://github.com/elastic/connectors/pull/3151.


kostasb (approved):
Thank you for fixing this.

kostasb:
As for the target release, should that be the `8.18.0` tag without backporting? That's how we merged the OSS Dockerfile updates: https://github.com/elastic/connectors/pull/3063

PR #3162, opened 2025-01-30
Add hardcoded qualifier

navarone-feekery:
Add hard-coded qualifier `beta1` so nightly builds have a qualifier. This will be removed when we release 9.0.

### DIFF
--- a/.buildkite/publish/publish-common.sh
+++ b/.buildkite/publish/publish-common.sh
@@ -13,6 +13,9 @@ export SCRIPT_DIR="$CURDIR"
 export BUILDKITE_DIR=$(realpath "$(dirname "$SCRIPT_DIR")")
 export PROJECT_ROOT=$(realpath "$(dirname "$BUILDKITE_DIR")")
 
+# temporary hard-coded build qualifier, this should be removed once we've released 9.0
+export VERSION_QUALIFIER="${VERSION_QUALIFIER:-beta1}"
+
 source $SCRIPT_DIR/git-setup.sh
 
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov (commented):
We run nightlies for recent branches, would that also affect them?

navarone-feekery:
@artem-shelkovnikov according to @seanstory:

> Nightlies are snapshots AND staging DRAs.
Yes, that change would work. I was thinking putting it in the pipeline.yml, but this is totally fine.
Snapshots are exempt from qualifiers.
Relevant logic block:
https://github.com/elastic/connectors/blob/d4c86f451a0649a6a2b0b41970e011e30c72cc79/.buildkite/publish/dra/init_dra_publishing.sh#L54-L63
[init_dra_publishing.sh](https://github.com/elastic/connectors/blob/d4c86f451a0649a6a2b0b41970e011e30c72cc79/.buildkite/publish/dra/init_dra_publishing.sh)
So after we cut a 9.0 branch, the nightlies will start producing staging and snapshot builds for 9.0.0. Right now, they're only snapshot, because main isn't on the staging branch pattern, and is only included if you specify a qualifier. (edited)

navarone-feekery:
> We run nightlies for recent branches, would that also affect them?

Or do you mean branches that aren't `main`?
If we don't backport it should be fine, right?

seanstory:
correct, as long as we dont' backport, it's fine

seanstory (approved):

PR #3164, opened 2025-01-30
bump for 8.18

joemcelroy:
Bump for 8.18 release


.backportrc.json
```diff
@@ -2,6 +2,7 @@
   "targetBranchChoices": [
     { "name": "main", "checked": true },
     "8.x",
+    "8.18",
```
seanstory, line 5: this file's changes need to go into `main`, not `8.x`

seanstory (commented):
Let's hold off on merging this until we fix the 8.18 branch, so that it can be forked from 8.x

PR #3178, opened 2025-02-05
[Framework] Correctly handle soft deleted connectors

jedrazb:
## Closes https://github.com/elastic/connectors/issues/3118

Since we introduced soft-deletes in 9.0, make sure framework respects deleted flag. 

### Changes

Exclude any connectors with `deleted: true` when fetching supported connectors in a query (note, I added this to native connectors query for conssitency but this code should be deleted anyway https://github.com/elastic/connectors/issues/2991)

This is in line with how API handles soft-deleted connectors https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/ent-search/src/main/java/org/elasticsearch/xpack/application/connector/ConnectorIndexService.java#L410C12-L410C119

#### Pre-Review Checklist

### Verification

Ran this locally with debug log enabled with ES 9.0. Soft-deleted connector, and that was resulting output of the framework:
```
[content_sync_job_execution_service] There's no supported connectors found with native service types [] or connector ids [Fq4v1pQBjAGa8X_ervuC]         
```
`Fq4v1pQBjAGa8X_ervuC` was soft deleted so behaviour is correct


jedrazb:
buildkite test this

PR #3189, opened 2025-02-06
Adding warning filter to suppress GA warning from ES Python client

mattnowzari:
## Closes https://github.com/elastic/search-team/issues/9234
This PR adds a warning filter to suppress the following warning from output:

`
GeneralAvailabilityWarning: This API is in technical preview and may be changed or removed in a future release. Elastic will work to fix any issues, but features in technical preview are not subject to the support SLA of official GA features. result = await func()`

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/__init__.py
+++ b/connectors/es/__init__.py
@@ -5,11 +5,11 @@
 #
 import warnings
 
+from elasticsearch.exceptions import GeneralAvailabilityWarning
+
 from connectors.es.client import ESClient  # NOQA
 from connectors.es.document import ESDocument, InvalidDocumentSourceError  # NOQA
 from connectors.es.index import ESIndex  # NOQA
-
-from elasticsearch.exceptions import GeneralAvailabilityWarning
 
 warnings.filterwarnings("ignore", category=GeneralAvailabilityWarning)
 
--- a/connectors/es/__init__.py
+++ b/connectors/es/__init__.py
@@ -3,9 +3,15 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
+import warnings
+
 from connectors.es.client import ESClient  # NOQA
 from connectors.es.document import ESDocument, InvalidDocumentSourceError  # NOQA
 from connectors.es.index import ESIndex  # NOQA
+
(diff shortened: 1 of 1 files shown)


mattnowzari:
Just a general note - I made a best guess as to where to place this warning suppressor, if there is a more ideal location for it that I missed I'd love to know! 😄

seanstory:
I think this makes good sense. Since it's in the `__init__.py`, I'd expect this would get picked up by tests too. So does this mean it can be removed from our pytest config?

mattnowzari:
@seanstory that is a great point, let me do some testing around that and if we can safely remove it from the pytest config I will roll that diff into this PR as well

mattnowzari:
Update: `pytest` won't pick this up for some reason :( I'm not 100% sure why this is yet. From what I can tell, the only test that needs the filter in `pytest.ini` is `test_kibana.py`.

jedrazb:
Yup, hmm since we put this filter to `connector/es/client.py` will help? I see similar issues reported on stack overflow https://stackoverflow.com/questions/75345190/pytest-overrides-existing-warning-filters , I think we can live with those warning filtered in 2 places

artem-shelkovnikov:
> From what I can tell, the only test that needs the filter in pytest.ini is test_kibana.py.

You can go ahead and add the filter there too - it's a standalone code that's used for functional tests and not used in production.

mattnowzari:
Thanks for the input guys!

So final accounting:

- `pytest.ini` is going to keep its current warning filter [here](https://github.com/elastic/connectors/blob/7dcc99121c453bb45050eaacd677185e7c57a632/pytest.ini#L20) as test_kibana.py needs it (I tried removing that line and putting `warnings.filterwarnings("ignore", category=GeneralAvailabilityWarning)` _directly_ intside test_kibana.py and it doesn't work, most likely for the reason Jedr found in the Stackoverflow thread)

- The diffs in this PR will stay inside es/__init__.py as it is a good general location that filters out the warnings we need

Once buildkite passes I'll merge this in

mattnowzari:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.18|[](https://github.com/elastic/connectors/pull/3192)|

### Questions ?

PR #3190, opened 2025-02-07
[Makefile] Improvements to python selection and requirement changes

jedrazb:
#### 1. Fix Makefile target

Fix Makefile `elastic-ingest` target to force reinstall when requirements change. The `requirements/$(ARCH).txt` references  inside`requirements/framework.txt` the `framework.txt` was not listed as dependency of makefile target. Hence we were not reinstalling when framework-level dependency was updated.

#### 2. Fix python3 selection 

That was triggering me for a while. With every new python version update the global alias of python3 is bumped (on MacOS at least with homebrew python), e.g. now `python3` has version 3.13 for me. 

This caused running `make clean && make run` always fail as connectors are not compatible with python 3.12+. As requirements will fail to install.

In order to address this issue, I added logic to:
- use `PYTHON=python3` as before (backward compatible for CI setup)
  - check CI step that we use correct version 
- check python3 runtime version 
- if version > 3.11 try using `python3.11` or `3.10` alias (in that order)
- if those python versions are not available on the system fail with error
- print info which python version we are using
- works pretty well for me :) 

No backport required.

#### Pre-Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,22 @@
 .PHONY: install clean lint autoformat test release ftest ftrace install run default-config release docker-build docker-run docker-push
 
-PYTHON?=python3
+# Default to python3, but if its version is >3.11, try using python3.11 or python3.10.
+PYTHON ?= python3
+PYTHON_VERSION := $(shell $(PYTHON) -c "import sys; print(f'{sys.version_info[0]}.{sys.version_info[1]}')" 2>/dev/null)
+CHECK_UNSUPPORTED_PYTHON_VERSION := $(shell $(PYTHON) -c "import sys; print(1 if sys.version_info[:2] > (3,11) else 0)" 2>/dev/null)
+ifeq ($(CHECK_UNSUPPORTED_PYTHON_VERSION),1)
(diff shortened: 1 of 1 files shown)


navarone-feekery (approved):

artem-shelkovnikov:
Btw highly recommend using `pyenv` to avoid awkward stuff and simplify working with connectors. I use it and never had any issue with version updates in OS + pyenv is used for our CI so that it could run tests for all supported versions of python

jedrazb:
@artem-shelkovnikov in that case should we add section about pyenv in `DEVELOPING.md`? I was not aware we should be using this with that repo

Still I think this PR is helpful because now we can fail fast if there is no matching python version

artem-shelkovnikov:
`pyenv` is generally good for any python project development supporting multiple version, indeed would be great to add this to `DEVELOPING.md`, but also to main README.md too, let's create an issue and address later?

jedrazb:
Created https://github.com/elastic/connectors/issues/3193

PR #3321, opened 2025-03-22
Improve pagination handling in Confluence API client

WildDogOne:
## Closes https://github.com/elastic/connectors-py/issues/3320

The code is here to enhance the pagination handling in case the /rest/content/search API is used since there is no next link in that api.
I would be happy if someone could check if my logic is somewhat acceptable.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -259,23 +259,31 @@ class ConfluenceClient:
         Yields:
             response: JSON response.
         """
-        url = os.path.join(self.host_url, URLS[url_name].format(**url_kwargs))
+        base_url = os.path.join(self.host_url, URLS[url_name].format(**url_kwargs))
+        start = 0
+
         while True:
             try:
+                url = f"{base_url}&start={start}"
                 self._logger.debug(f"Starting pagination for API endpoint {url}")
                 response = await self.api_call(url=url)
                 json_response = await response.json()
+
                 links = json_response.get("_links")
                 yield json_response
-                if links.get("next") is None:
+                if links.get("next"):
+                    url = os.path.join(
+                        self.host_url,
+                        links.get("next")[1:],
+                    )
+                elif json_response.get("start") + json_response.get("size") < json_response.get("totalSize"):
+                    start = json_response.get("start") + json_response.get("size")
+                    url = f"{base_url}&start={start}"
+                else:
                     return
-                url = os.path.join(
-                    self.host_url,
-                    links.get("next")[1:],
-                )
             except Exception as exception:
                 self._logger.warning(
-                    f"Skipping data for type {url_name} from {url}. Exception: {exception}."
+                    f"Skipping data for type {url_name} from {base_url}. Exception: {exception}."
                 )
                 break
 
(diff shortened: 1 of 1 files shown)


seanstory:
buildkite test this

connectors/sources/confluence.py
```diff
@@ -259,23 +259,31 @@ async def paginated_api_call(self, url_name, **url_kwargs):
...
+                else:
                     return
-                url = os.path.join(
-                    self.host_url,
-                    links.get("next")[1:],
-                )
             except Exception as exception:
                 self._logger.warning(
-                    f"Skipping data for type {url_name} from {url}. Exception: {exception}."
+                    f"Skipping data for type {url_name} from {base_url}. Exception: {exception}."
```
seanstory, line 286: I think we'd actually still want to log out `url`, since it would let us know which "page" of results had an issue.

connectors/sources/confluence.py
```diff
@@ -984,8 +992,9 @@ async def search_by_query(self, query):
             # entity can be space or content
             entity_details = entity.get(SPACE) or entity.get(CONTENT)
 
-            if (
-                entity_details.get("type", "") == "attachment"
+            if not entity_details:
+                continue
+            if (entity_details.get("type", "") == "attachment"
```
seanstory, line 997:
```suggestion
            if not entity_details or (entity_details.get("type", "") == "attachment"
```

might be a bit cleaner. But what's the scenario you found where `entity_details` does not exist?

seanstory (commented):
Thanks for submitting these! Would you be willing to add some unit tests as well, to exercise your changes and demonstrate the situations where you hit them?

seanstory:
Also, the linter is failing. You can autoformat your code with `make autoformat`

artem-shelkovnikov:
@WildDogOne when is the `/rest/content/search` API used? What's the best way to test your change against a real Confluence?

connectors/sources/confluence.py
WildDogOne: I never debugged that into depth, I just noticed that the connector sometimes crashes because of entity_details not being around. I can look into this closer

seanstory:
buildkite test this

connectors/sources/confluence.py
```diff
@@ -259,23 +259,31 @@ async def paginated_api_call(self, url_name, **url_kwargs):
         Yields:
             response: JSON response.
         """
-        url = os.path.join(self.host_url, URLS[url_name].format(**url_kwargs))
+        base_url = os.path.join(self.host_url, URLS[url_name].format(**url_kwargs))
+        start = 0
+
         while True:
             try:
+                url = f"{base_url}&start={start}"
```
seanstory, line 267: This line probably needs to be outside of the `while True`, otherwise it'll overwrite the last iteration's attempts to set the URL to the "next" link, if that strategy was used.

connectors/sources/confluence.py
seanstory: Ah! Ok, hadn't caught this, but I think `url` needs to be defined outside of the loop to start. That way we can reference it here, and we also can be sure that it's not re-set on each iteration (see my most recent comment)

PR #3337, opened 2025-04-04
Parse invalid OneDrive permissions without crashing

navarone-feekery:
This is in response to an SDH where Connectors were crashing because the user had a misconfigured permission in OneDrive. Ideally, the Connector shouldn't crash in this case and instead should log a useful warning about the permission missing.

- Handle cases where the user id from `grantedToV2` permissions API response is empty
- Add warning logs for when permissions are unexpectedly not present

### DIFF
--- a/connectors/sources/onedrive.py
+++ b/connectors/sources/onedrive.py
@@ -661,7 +661,7 @@ class OneDriveDataSource(BaseDataSource):
                     permissions.append(_prefix_user_id(identity_id))
                 else:
                     log = f"No user id found for user {user_id} for file {file_id} in `grantedToV2`"
-                    logger.warning(log)
+                    self._logger.warning(log)
 
             if identities := permission.get("grantedToIdentitiesV2"):
                 for identity in identities:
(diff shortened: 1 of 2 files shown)


connectors/sources/onedrive.py
```diff
@@ -656,7 +656,12 @@ async def get_entity_permission(self, user_id, file_id):
             user_id=user_id, file_id=file_id
         ):
             if identity := permission.get("grantedToV2"):
-                permissions.append(_prefix_user_id(identity.get("user").get("id")))
+                identity_id = identity.get("user", {}).get("id")
+                if identity_id:
+                    permissions.append(_prefix_user_id(identity_id))
+                else:
+                    log = f"No user id found for user {user_id} for file {file_id} in `grantedToV2`"
+                    self._logger.warning(log)
```
artem-shelkovnikov, line 664: So user won't have access to this file?

artem-shelkovnikov (approved):

connectors/sources/onedrive.py
navarone-feekery: Good point, perhaps this log should be re-thought.
Instead maybe log when the fields `user` or `group` fields are present but they don't have an `id`. WDYT?

navarone-feekery: I've changed it to only log a warning if the response has `user` or `group` and no nested `id`

artem-shelkovnikov:
I see the tests fail consistently, ping me once they're fixed and I'll hit approve asap :)

artem-shelkovnikov (approved):

PR #3358, opened 2025-04-14
Attempt to retry 403 errors when they are actually secondary throttling limit

artem-shelkovnikov:
## Closes https://github.com/elastic/search-team/issues/8624

This PR attempts to fix a problem with 403 returned from GitHub being interpreted as `UNAUTHORIZED` when in some cases it can mean that the secondary rate limit quota has been exceeded, see https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit and https://github.com/changesets/action/issues/192

We use `gidgethub` package that allows to interact with Github with a chosen HTTP client. We use `aiohttp` so we pass the aiohttp client into `gitgethub` library:

```python
    @cached_property
    def _get_client(self):
        return GitHubAPI( # this is `gidgethub`
            session=self._get_session, # this is aiohttp session
            requester="",
            base_url=self.base_url,
        )

    @cached_property
    def _get_session(self):
        connector = aiohttp.TCPConnector(ssl=self.ssl_ctx)
        timeout = aiohttp.ClientTimeout(total=None)
        return aiohttp.ClientSession(
            timeout=timeout,
            raise_for_status=True, # <<<<< this got changed to False
            connector=connector,
        )
```

Before this PR `aiohttp.ClientSession` was created with `raise_for_status=True`. This led to the problem that `gidgethub` would not be able to handle exceptions - their status handling is build on reading the requests/responses coming from Github, if `raise_for_status=True` is set then `aiohttp` would just bubble up.

Because of this, we've never actually received exceptions from `gidgethub` library at all. In this PR I've changed `raise_for_status` to `False` and that allowed `gidgethub` handle and wrap status errors. Therefore I had to change the code around to catch `gidgethub.HTTPException` instead of `aiohttp.ClientResponseError`.

`gidgethub.RateLimitExceeded` is a subclass of `gidgethub.HTTPException` so I had to move its `except` statements up.

I've checked the code for throwing `gidgethub.RateLimitExceeded` and [it handles 403s](https://github.com/gidgethub/gidgethub/blob/ab4fadb3b7c62f09bb9271a15091abef3db81a89/gidgethub/sansio.py#L339): 

```
            if status_code == 403:
                rate_limit = RateLimit.from_http(headers)
                if rate_limit and not rate_limit.remaining:
                    raise RateLimitExceeded(rate_limit, message)
```

Why we cannot handle it on our side:

`aiohttp` with `raise_for_status=True` will raise an instance `aiohttp.ClientResponseError`. This will cause the body of response to be lost - `aiohttp.ClientResponseError` does not store it and just returns a message that represents the status. See [this SO question](https://stackoverflow.com/questions/55142824/how-do-i-get-the-body-of-an-http-403-response-using-pythonaiohttp):

> The newer versions unfortunately recycle the connection once raise_for_status() is called hence you can't fetch the error body later.

We could work around it, but to do so we need to make an HTTP call ourselves, which we cannot do - `gidgethub` does it.

So TLDR: I've swapped our exception handling fro handling `aiohttp` exceptions to handling `gidgethub` exceptions to be more precise and allow properly catching 403 that is a secondary rate-limit.

#### Pre-Review Checklist


seanstory (approved):
lgtm

seanstory (approved):

artem-shelkovnikov:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.0|[](https://github.com/elastic/connectors/pull/3380)|
|✅|8.x|[](https://github.com/elastic/connectors/pull/3381)|
|✅|8.18|[](https://github.com/elastic/connectors/pull/3382)|
|✅|8.17|[](https://github.com/elastic/connectors/pull/3383)|

### Questions ?

PR #3375, opened 2025-04-21
New configurable service.idle_jobs_threshold

mattnowzari:
## Closes https://github.com/elastic/connectors/issues/2298

This PR creates a new configurable field

`service.idle_jobs_threshold`

which is the amount of time (in seconds) before a sync job is considered 'idle'.

We make this field configurable by grabbing it from the `service` configurations and passing the value to `SyncJobIndex::idle_jobs()` as an new argument - previously this was a hardcoded global variable.

**TODO**
- [ ] Setting this value to zero should disable 'idle' jobs fully (?)
- [ ] Possibly provide a log message as to why this timeout occurred

#### Pre-Review Checklist

### DIFF
--- a/config.yml.example
+++ b/config.yml.example
@@ -217,6 +217,10 @@
 #service.log_level: INFO
 #
 #
+##  The threshold (in seconds) before a sync job is considered idle.
+#service.idle_jobs_threshold: 300
+#
+#
 ## ------------------------------- Extraction Service ----------------------------------
 #
 ##  Local extraction service-related configurations.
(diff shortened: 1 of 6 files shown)


artem-shelkovnikov:
There's one thing to keep in mind here - Kibana has its own threshold where it'll also mark jobs as "Idle" in UI. Don't remember where exactly, but somewhere in our module.

Connectors fail the job once idle threshold is reached, while Kibana just signalises that it haven't seen any updates for a bit and job _might_ be stuck - might be worth to re-visit this part too.

PR #3384, opened 2025-04-23
Fix Confluence connector swallowing all errors

mattnowzari:
## Closes https://github.com/elastic/connectors/issues/2394

The goal of this PR is to update the following `paginated_api_call` function to re-raise and/or handle specific Exceptions instead of blanket-swallowing any Exceptions that arise.

Current exception-handling code:
https://github.com/elastic/connectors/blob/8c8d4a3f7775c6edbb598a81a2b67348a7244f3a/connectors/sources/confluence.py#L194-L198

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -108,6 +108,9 @@ class NotFound(Exception):
     pass
 
 
+class BadRequest(Exception):
+    pass
+
 class Unauthorized(Exception):
     pass
 
@@ -214,18 +217,21 @@ class ConfluenceClient:
 
             await self._sleeps.sleep(retry_seconds)
             raise ThrottledError
+        elif exception.status == 400:
+            self._logger.error(f"Received Bad Request error for URL: {url}")
+            raise BadRequest from exception
         elif exception.status == 401:
             self._logger.error(f"Received Unauthorized error for URL: {url}")
-            raise Unauthorized
+            raise Unauthorized from exception
         elif exception.status == 403:
             self._logger.error(f"Received Forbidden error for URL: {url}")
-            raise Forbidden
+            raise Forbidden from exception
         elif exception.status == 404:
             self._logger.error(f"Received Not Found error for URL: {url}")
-            raise NotFound
+            raise NotFound from exception
         elif exception.status == 500:
             self._logger.error(f"Internal Server Error occurred for URL: {url}")
-            raise InternalServerError
+            raise InternalServerError from exception
         else:
             self._logger.error(
                 f"Error while making a GET call for URL: {url}. Error details: {exception}"
--- a/connectors/sources/confluence.py
+++ b/connectors/sources/confluence.py
@@ -107,6 +107,12 @@ class InternalServerError(Exception):
 class NotFound(Exception):
     pass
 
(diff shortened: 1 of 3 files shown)


mattnowzari:
> Agreed, this looks like the problem. We try to balance between the ideologies of:
       - "don't just fail the sync if there's some minor error, move on"
       - "fail fast if there's an 

I was reading through [this](https://github.com/elastic/connectors/issues/2393) GH issue and came across the above quote from @seanstory 

I don't have as much context or hands-on experience with Connectors (yet) - what constitutes a 'minor' or 'major' error in the case of connector syncs (for Confluence or any data source for that matter)

For example, in this PR I've added exception catching for 401s and 403s - are these worth re-raising so they get treated as a major error, or are they minor enough that we don't want to fail a sync because of them?

I noticed a lot of data sources don't have this fine-grain control over Exceptions, so if this is treading new-ish ground I might take a beat to think harder about this and propose a heuristic that we can apply to other data sources, too.

cc @artem-shelkovnikov @erikcurrin-elastic as well

mattnowzari:
Marking this as ready-to-review, with the asterisk that this may need more work - see the conversation starter above!

connectors/sources/confluence.py
```diff
@@ -206,6 +214,12 @@ async def _handle_client_errors(self, url, exception):
 
             await self._sleeps.sleep(retry_seconds)
             raise ThrottledError
+        elif exception.status == 401:
```
jedrazb, line 217: nit, can we add handling `400` bad reqest here for completness?  every now and then we do see client request errors for some edge cases so it would be good to log some more details here

jedrazb (dismissed):
lgtm, can we also add handling for 400s?

connectors/sources/confluence.py
```diff
@@ -206,6 +214,12 @@ async def _handle_client_errors(self, url, exception):
...
+            self._logger.error(f"Received Unauthorized error for URL: {url}")
+            raise Unauthorized
```
artem-shelkovnikov, line 219: We're losing exception messages and other metadata here.

I think all places need to `raise X from exception` and make sure that the message gets propagated. If not, it should be added to logs too.

mattnowzari: Isn't `raise x from exception` used for chaining exceptions? In the `_handle_client_errors()` method, we aren't catching exceptions, but rather checking status codes and raising exceptions for the first time - in other words, there isn't an exception to raise _from_.

As an alternative, I could implement these custom exceptions (they currently just have `pass`) so they have messaging and metadata. The `raise` keywords in `paginated_api_call` will simply re-raise the specific exceptions anyways.

artem-shelkovnikov (approved):

seanstory:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.19|[](https://github.com/elastic/connectors/pull/3460)|
|✅|8.18|[](https://github.com/elastic/connectors/pull/3461)|

### Questions ?

jedrazb:
Thank you Sean for manual backporting! It was late in the day and I was sjut checking whether automation would do it for me automatically :D

PR #3424, opened 2025-05-19
Extend config to only sync specific standard objects

meghanmurphy1:
## Closes https://github.com/elastic/search-team/issues/9961

Adds a text area to allow users to only sync specific Salesforce Object Types.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/salesforce.py
+++ b/connectors/sources/salesforce.py
@@ -56,6 +56,16 @@ OFFSET = 200
 OBJECT_READ_PERMISSION_USERS = "SELECT AssigneeId FROM PermissionSetAssignment WHERE PermissionSetId IN (SELECT ParentId FROM ObjectPermissions WHERE PermissionsRead = true AND SObjectType = '{sobject}')"
 USERNAME_FROM_IDS = "SELECT Name, Email FROM User WHERE Id IN {user_list}"
 FILE_ACCESS = "SELECT ContentDocumentId, LinkedEntityId, LinkedEntity.Name FROM ContentDocumentLink WHERE ContentDocumentId = '{document_id}'"
+WILDCARD = "*"
+
+STANDARD_SOBJECTS = [
+    "Account",
+    "Opportunity",
+    "Contact",
+    "Lead",
+    "Campaign",
+    "Case",
+]
 
 RELEVANT_SOBJECTS = [
     "Account",
@@ -186,6 +196,7 @@ class SalesforceClient:
         self._queryable_sobject_fields = {}
         self._sobjects_cache_by_type = None
         self._content_document_links_join = None
+        self.standard_objects_to_sync = configuration["standard_objects_to_sync"]
 
         self.base_url = base_url
         self.api_token = SalesforceAPIToken(
@@ -1313,6 +1324,7 @@ class SalesforceDataSource(BaseDataSource):
         )
         self.doc_mapper = SalesforceDocMapper(base_url)
         self.permissions = {}
+        self.standard_objects_to_sync = configuration["standard_objects_to_sync"]
 
     def _set_internal_logger(self):
         self.salesforce_client.set_logger(self._logger)
@@ -1340,10 +1352,17 @@ class SalesforceDataSource(BaseDataSource):
                 "tooltip": "The client secret for your OAuth2-enabled connected app. Also called 'consumer secret'",
                 "type": "str",
             },
+             "standard_objects_to_sync": {
+                "display": "textarea",
+                "label": "Standard Objects to Sync",
+                "order": 4,
+                "tooltip": "The list of Salesforce Standard Objects to sync",
+                "type": "list",
+            },
             "use_text_extraction_service": {
                 "display": "toggle",
                 "label": "Use text extraction service",
-                "order": 4,
+                "order": 5,
                 "tooltip": "Requires a separate deployment of the Elastic Text Extraction Service. Requires that pipeline settings disable text extraction.",
                 "type": "bool",
                 "ui_restrictions": ["advanced"],
@@ -1352,7 +1371,7 @@ class SalesforceDataSource(BaseDataSource):
             "use_document_level_security": {
                 "display": "toggle",
                 "label": "Enable document level security",
-                "order": 5,
+                "order": 6,
                 "tooltip": "Document level security ensures identities and permissions set in Salesforce are maintained in Elasticsearch. This enables you to restrict and personalize read-access users and groups have to documents in this index. Access control syncs ensure this metadata is kept up to date in your Elasticsearch documents.",
                 "type": "bool",
                 "value": False,
@@ -1484,7 +1503,14 @@ class SalesforceDataSource(BaseDataSource):
     async def get_docs(self, filtering=None):
         # We collect all content documents and de-duplicate them before downloading and yielding
         content_docs = []
+        logger.info(f"Fetching Standard Objects: {self.standard_objects_to_sync}")
 
+        objects_to_sync = (
+            STANDARD_SOBJECTS
+            if [WILDCARD] == self.standard_objects_to_sync 
+            else self.standard_objects_to_sync
+        )
+        
         if filtering and filtering.has_advanced_rules():
             advanced_rules = filtering.get_advanced_rules()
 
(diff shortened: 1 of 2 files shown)


jedrazb (commented):
Good start! There are couple of cases we should think about when it comes to syncing standard and custom objects in Salesforce

connectors/sources/salesforce.py
```diff
@@ -1502,92 +1527,99 @@ async def get_docs(self, filtering=None):
...
         else:
-            for sobject in [
-                "Account",
-                "Opportunity",
-                "Contact",
-                "Lead",
-                "Campaign",
-                "Case",
-            ]:
+            for sobject in objects_to_sync:
```
artem-shelkovnikov, line 1530: I think it'd be nice to make this a hash instead and avoid `for` loop, it just makes it harder to perceive, e.g.:

```
objects_to_sync = { 'Account': 1, 'Opportunity': 0}

if 'Account' in objects_to_sync:
    ...

if 'Case' in objects_to_sync:
   ...
```

connectors/sources/salesforce.py
```diff
@@ -1502,92 +1527,99 @@ async def get_docs(self, filtering=None):
...
-                )
-                yield (
-                    self.doc_mapper.map_salesforce_objects(
-                        self._decorate_with_access_control(
-                            custom_object, access_control
-                        )
-                    ),
-                    None,
-                )
+            if [WILDCARD] == self.standard_objects_to_sync or any(object in self.standard_objects_to_sync for object in custom_objects):
```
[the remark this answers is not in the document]
artem-shelkovnikov: +1 on separating the two.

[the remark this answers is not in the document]
meghanmurphy1: ah ok! - I'll update so the Wildcard only pertains to list of standard objects. 
I like the toggle idea- so when the toggle is true a dropdown appears with the list of custom objects and they should be all checked at first? Then the user can uncheck ones they don't want synced?

connectors/sources/salesforce.py
```diff
@@ -1340,10 +1351,17 @@ def get_default_configuration(cls):
                 "tooltip": "The client secret for your OAuth2-enabled connected app. Also called 'consumer secret'",
                 "type": "str",
             },
+            "standard_objects_to_sync": {
+                "display": "textarea",
+                "label": "Standard Objects to Sync",
+                "order": 4,
+                "tooltip": "The list of Salesforce Standard Objects to sync",
+                "type": "list",
```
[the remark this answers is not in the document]
meghanmurphy1: I tried to go off of how the Confluence connector does spaces so it's more consistent

connectors/sources/salesforce.py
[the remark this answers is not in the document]
meghanmurphy1: I added just a toggle, but interested to know if there's a way to allow users to check which custom objects to sync?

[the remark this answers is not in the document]
artem-shelkovnikov: There's really no good way to check what custom objects are available, so comma-separated list that they enter and we validate at the backend is the only real way to go IMO

connectors/sources/salesforce.py
```diff
@@ -1340,10 +1352,26 @@ def get_default_configuration(cls):
                 "tooltip": "The client secret for your OAuth2-enabled connected app. Also called 'consumer secret'",
                 "type": "str",
             },
+            "standard_objects_to_sync": {
+                "display": "textarea",
+                "label": "Standard Objects to Sync",
+                "order": 4,
+                "tooltip": "The list of Salesforce Standard Objects to sync",
```
jedrazb, line 1359: We should explain wildcard somewhere

```suggestion
                "tooltip": "The list of Salesforce Standard Objects to sync, use '*' to include all",
```

connectors/sources/salesforce.py
```diff
@@ -1340,10 +1352,26 @@ def get_default_configuration(cls):
...
+                "type": "list",
+                "value": WILDCARD,
```
jedrazb, line 1361: I know I suggested wildcard before :D .... but maybe we should be more explicit about this ? Given that the list is short maybe we should just prepopulate with everything. It's easier to delete one entry than to build the list yourself
```suggestion
                "value": ",".join(STANDARD_SOBJECTS), 
```

connectors/sources/salesforce.py
```diff
@@ -1340,10 +1352,26 @@ def get_default_configuration(cls):
...
+            },
+            "sync_custom_objects": {
+                "display": "toggle",
+                "label": "Sync Custom Objects",
+                "order": 5,
+                "tooltip": "The list of Salesforce Custom Objects to sync",
```
jedrazb, line 1367: the tooltip is wrong, this is just a toggle

connectors/sources/salesforce.py
```diff
@@ -1340,10 +1352,26 @@ def get_default_configuration(cls):
...
             },
+            "standard_objects_to_sync": {
+                "display": "textarea",
+                "label": "Standard Objects to Sync",
+                "order": 4,
+                "tooltip": "The list of Salesforce Standard Objects to sync",
+                "type": "list",
+                "value": WILDCARD,
+            },
+            "sync_custom_objects": {
```
jedrazb, line 1363: I think we should add a `custom_objects_to_sync` config field.

When the `sync_custom_objects`  toggle is true, show the list to the user so they can choose which custom objects to sync.

Right now, we sync all custom by default, right? It’d be better to give users more control.

connectors/sources/salesforce.py
```diff
@@ -1485,6 +1513,12 @@ async def get_docs(self, filtering=None):
         # We collect all content documents and de-duplicate them before downloading and yielding
         content_docs = []
 
+        objects_to_sync = (
```
jedrazb, line 1516: this should also go to `else `statement? I don' think we need it to be top-level

connectors/sources/salesforce.py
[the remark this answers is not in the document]
jedrazb: ++ agreed with Artem. You don't need to know in advance what custom objects are available. If a user is advanced enough to create custom sobjects then they should know how to look them up.

connectors/sources/salesforce.py
meghanmurphy1: Yeah that makes sense especially since it's such a short list 🫡

connectors/sources/salesforce.py
meghanmurphy1: Yes, they all sync by default right now. IIUC, I can only display the list of custom objects at runtime so would have to do something where the user would need to know the list of custom objects and if one is wrong or errors out then show the list and say something like "Object x isn't available. Available custom objects are ..."

meghanmurphy1: I also am thinking if we do a list then probably don't need the toggle?

jedrazb: > I also am thinking if we do a list then probably don't need the toggle?

Yet another textarea that is list might make the configuration overwhelming. My take is to be explicit: "do you want to sync custom objects? no, you are good. If yes, we only then show the textarea with custom objects list"

>  I can only display the list of custom objects at runtime

We don't need to worry about displaying all custom objects. Keep the list empty. Say sth like: "please enter comma separated custom objects you want to sync". If "*" is used we will attempt to fetch all custom objects (as with current logic)

connectors/sources/salesforce.py
```diff
@@ -1340,10 +1356,36 @@ def get_default_configuration(cls):
...
+                "display": "toggle",
+                "label": "Sync Custom Objects",
+                "order": 5,
+                "tooltip": "Whether or not to sync custom objects",
+                "type": "bool",
+                "value": False,
+            },
+            "custom_objects_to_sync": {
+                "display": "textarea",
+                "label": "Sync Custom Objects",
```
jedrazb, line 1378: To be consistent with `"label": "Standard Objects to Sync",`
```suggestion
                "label": "Custom Objects to Sync",
```

jedrazb (approved):
one nit but otherwise LGTM

tests/sources/test_salesforce.py
```diff
@@ -746,7 +750,12 @@ def generate_account_doc(identifier):
 
 def test_get_default_configuration():
     config = DataSourceConfiguration(SalesforceDataSource.get_default_configuration())
-    expected_fields = ["client_id", "client_secret", "domain"]
+    expected_fields = [
```
jedrazb, line 753: this should be extended?

jedrazb (approved):

PR #3425, opened 2025-05-20
Leverage slack integration

seanstory:
See the email: `[PSA] Slack Workspace Integration for Buildkite`

after this merges, I'll check to make sure this is working as expected. I'm mildly worried that it'll send notifications for _every_ success, not just the first. We'll see. 🤞

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -5655,7 +5655,7 @@ MIT License
 
 
 multidict
-6.4.3
+6.4.4
 Apache Software License
    Copyright 2016 Andrew Svetlov and aio-libs contributors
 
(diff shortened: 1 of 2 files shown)


mattnowzari (approved):

seanstory:
> We already have solid notifications to #search-et-alerts

We do, but those have to be manually maintained, and there have been a few times when we realize that a given pipeline isn't set up. This would make it easier to maintain in code, rather than in buildkite UIs, and less chance that new repos get left out (since we'll likely copy-paste setup from elsewhere)

> I've skimmed through the docs and I'm not sure what these changes would do, what would it practically change?

Hopefully nothing. If this works the way we want, we can delete our setup from the buildkite UI and replace it with this.

> Similar setting in our deployable services report on each success, not just the first consecutive, I think it will spam

I'm hoping it won't do this - docs aren't clear. But if needs must, there's a more complex variant we can put in the pipelines themselves. Totally open to reverting/refactoring if this goes poorly.

seanstory:
You were right, didn't work. Trying again: https://github.com/elastic/connectors/pull/3426

PR #3426, opened 2025-05-20
Add slack settings directly in pipelines

seanstory:
undoes https://github.com/elastic/connectors/pull/3425/files, and puts them instead directly into the pipeline definitions.

This should explicitly let us only notify on failure and first-pass.

Also uses a regex to keep us from needing to update the branches we care about. `catalog-info.yml` will still control which branches/pipelines run on a schedule.


jedrazb (approved):
fingers crossed

PR #3459, opened 2025-06-10
Add support for default UUID type 4 serialization

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/3299

MongoDB is pretty special about UUID handling.

Before a certain version of MongoDB, there was UUID type 3 - driver-specific UUID. Here's example from docs:

```
An example GUID created and displayed in a C# application would be shown as follows:

00010203-0405-0607-0809-0A0B0C0D0E0F

The MongoDB C# driver (in the default configuration) will convert it to a Binary database representation and store it with the following byte order:

03 02 01 00 05 04 07 06 08 09 0A 0B 0C 0D 0E 0F

However, the Java driver used to access the same field will return and display the following:

06070405-0001-0203-0F0E-0D0C0B0A0908

The Python driver will display something different again:

03020100-0504-0706-0809-0A0B0C0D0E0F
```

At some point type 4 was introduced, which is serialized in the same way across all the drivers.

Now, second part is that it's possible to choose `uuidRepresentation` (see https://pymongo.readthedocs.io/en/stable/api/bson/binary.html#bson.binary.UuidRepresentation) when initalising the connection - for example through connection string: `mongodb://127.0.0.1:27021?uuidRepresentation=standard` - this will use the modern serialisation format that's cross-compatible.

If the application uses old Java client that writes legacy format, it can be done with `mongodb://127.0.0.1:27021?uuidRepresentation=javaLegacy`.

Without this `uuidRepresentation` field set the MongoDB client returns instances of Binary type - it has a value (UUID) + type (3 or 4 for UUID).

When `uuidRepresentation` is set, though, the driver decodes the field and returns regular python UUID objects that can be serialised without any addition.

Default `uuidRepresentation` is `unspecified` and by default MongoDB driver we use returns Binary instances.

What this PR does is: it parses Binary fields it receive with type=4 and treats them as UUID. If type=3 is received, a warning is emitted and field is skipped.

If a user still wants to sync MongoDB instance with type=3 UUIDs, they can modify their connection string to set `uuidRepresentation` with correct value and it will naturally make the connector work.

Additionally to the above, I've moved some stuff around and fixed minor problems.

#### Pre-Review Checklist

## Release Note

Better support for UUID serialisation - modern UUIDs are properly handled out of the box.

### DIFF
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -177,8 +177,6 @@ class MongoDataSource(BaseDataSource):
         try:
             client_params = {}
 
-            client_params["uuidRepresentation"] = "standard"
-
             if self.configuration["direct_connection"]:
                 client_params["directConnection"] = True
 
--- a/connectors/sources/mongo.py
+++ b/connectors/sources/mongo.py
@@ -198,16 +198,14 @@ class MongoDataSource(BaseDataSource):
             else:
                 client_params["tls"] = False
 
-            yield AsyncIOMotorClient(self.host, **client_params)
+            client = AsyncIOMotorClient(self.host, **client_params)
+
+            db = client[self.configuration["database"]]
+            self.collection = db[self.configuration["collection"]]
+
+            yield client
         finally:
-            if os.path.exists(certfile):
-                try:
-                    os.remove(certfile)
-                except Exception as exception:
-                    self._logger.warning(
-                        f"Something went wrong while removing temporary certificate file. Exception: {exception}",
-                        exc_info=True,
-                    )
+            self.remove_temp_file(certfile)
 
     def advanced_rules_validators(self):
         return [MongoAdvancedRulesValidator()]
@@ -215,6 +213,16 @@ class MongoDataSource(BaseDataSource):
     async def ping(self):
         with self.get_client() as client:
             await client.admin.command("ping")
+
+    def remove_temp_file(self, temp_file):
+        if os.path.exists(temp_file):
+            try:
+                os.remove(temp_file)
+            except Exception as exception:
+                self._logger.warning(
+                    f"Something went wrong while removing temporary certificate file. Exception: {exception}",
+                    exc_info=True,
+                )
 
     # TODO: That's a lot of work. Find a better way
     def serialize(self, doc):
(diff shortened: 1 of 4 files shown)


connectors/sources/mongo.py
```diff
@@ -196,14 +198,16 @@ def get_client(self):
...
         finally:
-            self.remove_temp_file(certfile)
+            if os.path.exists(certfile):
+                try:
+                    os.remove(certfile)
+                except Exception as exception:
+                    self._logger.warning(
+                        f"Something went wrong while removing temporary certificate file. Exception: {exception}",
+                        exc_info=True,
+                    )
```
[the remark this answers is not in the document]
artem-shelkovnikov: Removed on purpose

connectors/sources/mongo.py
```diff
@@ -196,14 +198,16 @@ def get_client(self):
             else:
                 client_params["tls"] = False
 
-            client = AsyncIOMotorClient(self.host, **client_params)
-
-            db = client[self.configuration["database"]]
-            self.collection = db[self.configuration["collection"]]
```
artem-shelkovnikov: Fun problem from design perspective:

This is a context manager, but it not only yields the `client`, but also sets up instance variable `self.collection` - looks like a logical leak that should not be there

connectors/sources/mongo.py
```diff
@@ -196,14 +198,16 @@ def get_client(self):
...
         finally:
-            self.remove_temp_file(certfile)
+            if os.path.exists(certfile):
+                try:
+                    os.remove(certfile)
+                except Exception as exception:
+                    self._logger.warning(
+                        f"Something went wrong while removing temporary certificate file. Exception: {exception}",
+                        exc_info=True,
+                    )
```
artem-shelkovnikov: Removed the helper method as it's very small and is pure.

Ideally I'd even delegate the file disposal to the function opening file instead of having our own custom code here instead.

connectors/sources/mongo.py
erikcurrin-elastic: It feels like this and some of the other files are mixing two separate issues.  What in this PR is related to UUID serialization?

connectors/sources/mongo.py
```diff
@@ -238,6 +232,16 @@ def _serialize(value):
...
             elif isinstance(value, DBRef):
                 value = _serialize(value.as_doc().to_dict())
+            elif isinstance(value, Binary):
+                # UUID_SUBTYPE is guaranteed to properly be serialized cross-platform and cross-driver
+                if value.subtype == UUID_SUBTYPE:
+                    value = value.as_uuid()
+                # OLD_UUID_SUBTYPE is platform-specific. If Java writes old UUID and Python reads it they won't match
+                elif value.subtype == OLD_UUID_SUBTYPE:
+                    self._logger.warning(
+                        f"Unexpected uuid subtype, skipping serialization of a field {value}. Please provide uuidRepresentation=VALUE with correct value in connection string"
```
jedrazb, line 250: Should we suggest setting `uuidRepresentation` to `standard`? Is there any other valid option? Other representations are legacy ones? https://pymongo.readthedocs.io/en/stable/api/bson/binary.html#bson.binary.UuidRepresentation

jedrazb (approved):
This LGTM, just a question about payload in the warning log?

connectors/sources/mongo.py
jedrazb: Also, is it possible that this only affects a subset of IDs in the MongoDB collection? If so, should we include the ID of the offending document in the payload? Though, if it’s not serialized correctly, we can’t read it… right?

jedrazb:
> Should we add documentation record about handling UUIDs?

++

connectors/sources/mongo.py
artem-shelkovnikov: I _think_ it's possible that the user has UUIDs in different formats, but it's a large mistake rather, so no need to account for it.

`uuidRepresentation` set to `standard` will not change anything in this case - these UUIDs are already stored in driver-specific format, so they won't really be converted to regular UUIDs, user needs to use the same uuidRepresentation as was used for the storage.

connectors/sources/mongo.py
artem-shelkovnikov: Yeah I'll move it to a different PR

jedrazb (approved):

connectors/sources/mongo.py
jedrazb: I misunderstood, and thought that `uuidRepresentation=standard` will do the conversion on the backend. Makes sense !

jedrazb (approved):

PR #3497, opened 2025-06-30
Remove MySQLQueries class and inline its methods

artem-shelkovnikov:
Small cleanup:

Before all SQL connectors were using absolutely same logic that was later split up.

The split did not clean up everything, so this PR just cleans up one thing - unnecessary `MySQLQueries` class.

This PR removes it and inlines queries into the places that use these queries - all queries are used exactly once (or even never).

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -17,7 +17,6 @@ from connectors.filtering.validation import (
 )
 from connectors.source import BaseDataSource, ConfigurableFieldValueError
 from connectors.sources.generic_database import (
-    Queries,
     configured_tables,
     is_wildcard,
 )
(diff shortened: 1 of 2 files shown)


connectors/sources/mysql.py
```diff
@@ -221,7 +189,9 @@ async def get_column_names_for_query(self, query):
             return [f"{column[0]}" for column in cursor.description]
 
     async def get_column_names_for_table(self, table):
-        return await self.get_column_names_for_query(self.queries.table_data(table))
+        return await self.get_column_names_for_query(
+            f"SELECT * FROM `{self.database}`.`{table}`"
```
jedrazb, line 193: isn't it a textbook case for sql injection scenario?

```
self.database = "main_db`; DROP DATABASE main_db; --"
```

I know it's the user who is responsible for providing `database` in config but I wonder if we can add extra validation as a followup

jedrazb (approved):
Refactor lgtm. What are your thoughts about my concern?

connectors/sources/mysql.py
artem-shelkovnikov: True, but it's not the biggest problem: you can get username/password for the MySQL instance from connector configuration.

artem-shelkovnikov: Issue: https://github.com/elastic/connectors/issues/3503

jedrazb (approved):

jedrazb (approved):

PR #3516, opened 2025-07-03
Improve Slack API error handling

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/1410

Improves error handling for Slack API calls.

Before failed API call could be swallowed by a connector resulting in data loss.

Imagine, you're fetching 10.000 messages. Then there is an error fetching data somewhere around message #1.000.

Before this PR, the sync would continue, and rest of 9.000 messages would have been deleted from the index.

With this PR, the sync will, firstly, actually retry (it did not work properly before for cases when Slack returned "ok": False), but then fail the sync - this will not result in data loss compared to previous implementation.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/slack.py
+++ b/connectors/sources/slack.py
@@ -39,6 +39,17 @@ class ThrottledError(Exception):
     pass
 
 
+class SlackAPIError(Exception):
+    """Internal exception class that wraps all non-ok responses from Slack"""
+
+    def __init__(self, reason):
+        self.__reason = reason
+
+    @property
+    def reason(self):
+        return self.__reason
+
+
 class SlackClient:
     def __init__(self, configuration):
         self.token = configuration["token"]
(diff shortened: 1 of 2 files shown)


connectors/sources/slack.py
```diff
@@ -53,8 +64,13 @@ def __init__(self, configuration):
...
 
-    def ping(self):
-        return self.test_auth()
+    async def ping(self):
+        url = f"{BASE_URL}/auth.test"
+        try:
+            await self._get_json(url)
+            return True
+        except SlackAPIError:
+            return False
```
navarone-feekery, line 73: `SlackApiError` is inheriting from `Exception`, so this should raise for any exception that happens during the ping call, right?
If so, is it worth logging the specific error that occurred?

navarone-feekery (approved):
LGTM, left an optional nit/question

connectors/sources/slack.py
navarone-feekery: Wait nevermind, I see that SlackApiError is raised manually. But still curious how the error gets surfaced to the user.

navarone-feekery: Actually `ping` just returns a boolean based on if connection could be made. So it's probably fine. So this thread could be ignored :D

artem-shelkovnikov: Yup that was in the original design of the connector so I kept it as is - client returns True/False, but DataSource class raises an error taking this as a result

PR #3546, opened 2025-07-23
test: Simplify functional test Dockerfile setup

quinn-harper:
## Context

Currently, we have two Dockerfiles for functional tests and they both build from the same base image as production images. That means we duplicate build behaviour and don't inherit any changes to the Docker build process by default.

This change introduces a single Dockerfile for functional testing which is based on the production image. It also deletes the extra functional testing Dockerfile as it was not validated in CI so I expect it is not necessary. This way, we maintain only one functional test image and it will be based on the production image, keeping up with any changes to it. This is based on a [previous discussion around the unnecessary Dockerfiles](https://github.com/elastic/connectors/pull/3063#discussion_r1918083228).

To futher make it a bit cleaner the Dockerfile has been moved to `tests/` as it should not require context outside of it (other than the base image). 

### Alternatives

The `Dockerfile.ftest` is technically not required. The required line could be added to all respective `docker-compose.yml` files:

```
build:
  dockerfile: Dockerfile.wolfi
  command: sh -c ".venv/bin/python -m pip install -r requirements/ftest.txt && <actual command>"
```

However, due to the duplication and somewhat hacky nature of it, I elected to keep one Dockerfile around.

#### Pre-Review Checklist

## Related Pull Requests

First discussed in:

- https://github.com/elastic/connectors/pull/3063#discussion_r1918083228


artem-shelkovnikov:
buildkite test this

quinn-harper:
Ah, it used to be set in the env variables but the source for which image is used has changed so that no longer works. I will update it to the same image as before.

artem-shelkovnikov (approved):
LGTM! Shall we also backport as far as we can? (up to 8.17, I guess)

quinn-harper:
If you think it's worth it. I will add the labels.

PR #3559, opened 2025-07-29
[ServiceNow] Paginate API calls when we have advanced sync rules

lorenabalan:
## Closes https://github.com/elastic/connectors/issues/3509

* Updated list of API calls in `get_filter_apis()` with appropriate offset instead of 0, similarly to how we do it in `get_record_apis()`
* Turned the function `async` because of the call to `get_table_length`

Tested it locally and we get more than 50 documents upserted, as per screenshot below:
Screenshot 2025-07-29 at 11 58 00

#### Pre-Review Checklist


connectors/sources/servicenow.py
```diff
@@ -203,18 +203,20 @@ def get_filter_apis(self, rules, mapping):
...
-                    "method": "GET",
-                    "url": self._prepare_url(
-                        url=ENDPOINTS["TABLE"].format(table=table_name),
-                        params=params.copy(),
-                        offset=0,
-                    ),
-                }
-            )
+            total_count = await self.get_table_length(table_name)
+            for page in range(int(total_count / TABLE_FETCH_SIZE) + 1):
```
meghanmurphy1: `int()` truncates then the + 1 will then round upwards so could we do something that would just round upwards in one go? Seems like `math.ceil()` might work? 🤔

lorenabalan: Good point! I used the same existing pattern we had in `get_record_apis()` but I've just updated both in [7976f24](https://github.com/elastic/connectors/pull/3559/commits/7976f24e33ae1a51b274ba15986ec556bc4c5cb4). 😊

meghanmurphy1 (dismissed):
LGTM! I think we just need to add backport labels to 8.17, 8.18, 9.0, and 9.1 😵‍💫

connectors/sources/servicenow.py
```diff
@@ -194,7 +195,7 @@ def _prepare_url(self, url, params, offset):
             full_url = f"{url}?{params_string}"
         return full_url
 
-    def get_filter_apis(self, rules, mapping):
+    async def get_filter_apis(self, rules, mapping):
```
artem-shelkovnikov, line 198: Functions `get_record_apis` and `get_filter_apis` are very similar, do you think it's possible to leave only one?

lorenabalan: Yessss good push, could totally see it now with fresh eyes! Fixed in [0e5f4e7](https://github.com/elastic/connectors/pull/3559/commits/0e5f4e79baa9f979554bdb57b4686633e6b57c86)

PR #3562, opened 2025-07-29
bump version to 9.1.1

mattnowzari:
Bump `9.1` branch to `9.1.1`


artem-shelkovnikov:
You'll have to wait until 9.1.1-SNAPSHOT Elasticsearch appears in the DRA, then re-trigger CI and once it gets green we can merge it.

In reality that could take days to happen, unfortunately

jedrazb:
@artem-shelkovnikov @mattnowzari I wonder if we should relax the CI smoke test for patch releases? E.g. have a fallback to 9.1.(n-1) in case 9.1.(n) is not available? 

In theory patch releases should never introduce breaking chanes

artem-shelkovnikov:
Even simpler, we can probably always use `9.X.0` for patch releases and only have this problem happen for minor bumps.

Additionally, would be great to modify code that uses DRA artefacts not to download the artefacts if they're present (with a flag is fine)

meghanmurphy1 (approved):

jedrazb:
@artem-shelkovnikov 

> Additionally, would be great to modify code that uses DRA artefacts not to download the artefacts if they're present (with a flag is fine)

Do you mean using docker images instead?

PR #3613, opened 2025-08-15
Salesforce and Google Drive mappings

mattnowzari:
### Part of https://github.com/elastic/search-team/issues/10779 and  https://github.com/elastic/search-team/issues/10780

This PR adds two explicit mappings for Salesforce and Google Drive connectors under the `agent` folder.
This is preliminary work to eventually have agentless `connectors` create fresh indices using these mappings.

#### Pre-Review Checklist


.python-version
```diff
@@ -0,0 +1 @@
+3.11.11
```
mattnowzari: Ahh! I will delete this.

connectors/agent/mappings/salesforce.json
```diff
@@ -0,0 +1,744 @@
+{
+  "mappings": {
+    "_meta": {
```
erikcurrin-elastic, line 3: Do we want to expound on this at all?  Something like: it can contain support cases, sales opportunities...

connectors/agent/mappings/salesforce.json
```diff
@@ -0,0 +1,744 @@
...
+      "briefing": "This index contains data from Salesforce."
+    },
+    "properties": {
+      "Account": {
+        "properties": {
+          "Id": {
+            "type": "keyword"
+          },
```
erikcurrin-elastic, line 11: Weird that we use different casing for these.  But, I imagine you inherited this schema

connectors/agent/mappings/salesforce.json
```diff
@@ -0,0 +1,744 @@
...
+        "properties": {
+          "type": {
+            "type": "keyword"
+          },
+          "url": {
+            "type": "text"
+          }
+        }
+      },
+      "body": {
```
erikcurrin-elastic, line 640: Can you explain why we use sparse embedding here and semantic text in google drive?

mattnowzari: AFAIK, both are `semantic_text`.

`model_settings.task_type : sparse_embedding` is the a default setting for `semantic_text` that gets set when you create an index and have not set it to something else manually. So even if we didn't specify this, I think Kibana will automatically populate it to this unless we manually specify something else.

So, a mapping that is written as...
```json
{
  "mappings": {
    "properties": {
      "field1": {
        "type": "semantic_text"
      }
    }
  }
}
```

...gets turned into the following when the index is created:
```json
{
  "mappings": {
    "properties": {
      "field1": {
        "inference_id": ".elser-2-elasticsearch",
        "model_settings": {
          "service": "elasticsearch",
          "task_type": "sparse_embedding"
        },
        "type": "semantic_text"
      }
    }
  }
}
```
This is the behavior I've seen at any rate. It would be worth making sure the mappings are consistent in how `semantic_text` field are defined, so it's a good callout regardless.

As always, it might also be worth researching to see if there is any fine-tuning we can do with the settings!

connectors/agent/mappings/salesforce.json
mattnowzari: Yep, this is the data shape that comes from our Salesforce connector

connectors/agent/mappings/salesforce.json
mattnowzari: Short answer - yes.

Long answer:
I initially thought `_meta` fields were beholden to the 50 char limit that field-level metadata has, but turns out they're not! Abhi's experiments have shown that long-form mapping descriptions do have benefit, so expanding on top-level `_meta` fields is going to be easy.

The per-field char limit should probably be increased to match - Sean has a draft PR [here](https://github.com/elastic/elasticsearch/pull/131634) that has been on my radar to possibly pick up and push through.

.python-version
seanstory: I just got this added to the `.gitignore` a few days ago

connectors/agent/mappings/google_drive.json
```diff
@@ -0,0 +1,81 @@
...
+      "id": {
+        "type": "keyword"
+      },
+      "last_updated": {
+        "type": "date"
+      },
+      "mime_type": {
+        "type": "keyword"
+      },
+      "name": {
```
seanstory, line 40: I might suggest `title`, as I think that might be more common across data sources. We don't _have_ to use a shared schema, but if we see opportunities to create overlap, I don't think it'll hurt.

connectors/agent/mappings/google_drive.json
```diff
@@ -0,0 +1,81 @@
...
+        "type": "text",
+        "meta": {
+          "description": "This is the name of a document."
```
seanstory: grey area, but I'd suggest `semantic_text` for file names/titles. Like a file might be "Earnings Report.pdf", but it would be great for it to show up in a search for "Quarterly profits"

seanstory: Though - maybe we want both? I could see someone also expecting exact-match to work.

connectors/agent/mappings/google_drive.json
```diff
@@ -0,0 +1,81 @@
...
+      "shared_drive": {
+        "fields": {
+          "keyword": {
+            "ignore_above": 256,
+            "type": "keyword"
+          }
+        },
+        "type": "text",
+        "meta": {
+          "description": "A document with this field is from a shared drive"
```
seanstory: This makes it sound like it should be a boolean. Is this the _name_ of its shared drive?

connectors/agent/mappings/google_drive.json
seanstory, line 1: we're missing meta.description values for a lot of the properties.

connectors/agent/mappings/salesforce.json
seanstory: This is an instance where we've previously wanted to maintain the output of a given source so that it feels familiar to the customer (Maybe they know `Account.Id`), but we can afford to be more opinionated now.

I'm not saying these shouldn't be capitalized - just don't feel beholden to what comes out of our Salesforce connector. We can rename fields with Ingest Pipelines.

connectors/agent/mappings/salesforce.json
```diff
@@ -0,0 +1,744 @@
...
+                        }
+                      }
+                    }
+                  },
+                  "Title": {
+                    "inference_id": ".elser-2-elasticsearch",
+                    "model_settings": {
+                      "service": "elasticsearch",
+                      "task_type": "sparse_embedding"
+                    },
```
seanstory: I'd scrap all this. If we leave just `"type": "semantic_text"`, then when the relevance team changes the default model/settings, new indices will pick up the new defaults. Less maintenance for us.

connectors/agent/mappings/salesforce.json
```diff
@@ -0,0 +1,744 @@
...
+      },
+      "ConvertedAccount": {
+        "type": "object"
+      },
+      "ConvertedContact": {
+        "type": "object"
+      },
+      "ConvertedOpportunity": {
+        "type": "object"
+      },
```
seanstory, line 233: Looks like we don't really know what these are. If we don't know what they are and don't know why we'd want to search them, let's not even index them. We can drop this stuff in an ingest pipeline.

connectors/agent/mappings/salesforce.json
```diff
@@ -0,0 +1,744 @@
...
+            "type": "boolean"
+          },
+          "records": {
+            "properties": {
+              "BccAddress": {
+                "type": "text"
+              },
+              "CcAddress": {
+                "type": "text"
+              },
```
seanstory, line 278: I'd think email addresses should be `keyword`, since we probably only care about exact match when searching. WDYT?

connectors/agent/mappings/salesforce.json
seanstory, line 1: This mapping is huge. I'd suggest that we try to pair down how much we keep for WorkChat. We want it to have plenty of context, and plenty of ways to filter its search results, but if we can't imagine a realistic way that we might leverage a field for search or for valuable result insights - let's remove it for now. Adding fields later will be much easier than deleting fields later.

connectors/agent/mappings/google_drive.json
mattnowzari: Acknowledged. I think I took Abhi's guidance that it's "easy to overdo" a little too seriously, esp given how small the GDrive mapping is.

For the Salesforce mapping, I do agree with your other comment that we should focus on the fields that will make searching easier and remove the ones that seem less relevant currently.

connectors/agent/mappings/google_drive.json
mattnowzari: For sure, I can add this to a list of fields we can manipulate with a ingest pipeline

connectors/agent/mappings/google_drive.json
mattnowzari: This is true. I think I was in the headspace of trying to 'direct' the LLM to make an infereance as what it 'means' when this field is populated, but that may have been thinking too hard.

I'll reword it

connectors/agent/mappings/google_drive.json
mattnowzari: It may be worth having it both ways. I'll try it out

connectors/agent/mappings/salesforce.json
mattnowzari: > I would suggest using just text (not semantic_text) for human names and email addresses. We may want to refine this later. But my guess is that searching for "Bab" to get results of "Bob" is important (typo-tolerant text search, not just exact-match keyword), but searching "Bobby" to get results of "Rob" (semantically understand 'Robert' variants) is less important.

I made the email fields `text` based on a previous recommendation you made [here](https://github.com/elastic/search-team/issues/10465#issuecomment-3165432513). For the record, I agree with your initial argument for emails being `text`.  

I worry that someone searching for `mattnowzari@email.com` when they mean `matt.nowzari@email.com` or some close variant may lock the LLM out of getting results because emails will be exact-match only.

All this to say - if you think `keyword` search is a better baseline, I could be convinced to switch it. I just think the typo-tolerance is going to be important, especially because typing emails wrong is SUCH a common mistake.

seanstory: Yeah ok - my old argument convinces me. :) 
Really, I'd hope the LLM would craft a query that uses both, so an exact-match gets a higher score than a partial match. Not sure how we tokenize email addresses. But we can worry about analyzers and query crafting/weighting/boosting later.

connectors/agent/mappings/google_drive.json
```diff
@@ -0,0 +1,92 @@
+{
```
erikcurrin-elastic, line 1: minor... but indent mappings?

erikcurrin-elastic (dismissed):
Just one nit, but ready to go

connectors/agent/mappings/google_drive.json
mattnowzari: Fixed!

seanstory:
>💔 Failed to create backport PR(s)

This is a quirk of the connectors repo and our backport automation that no one has ever bothered to address. We automatically add the `auto-backport` label, which then errors like this if there's no `vX.Y.Z` label that matches to a non-target branch. We're very used to seeing this error, don't worry. :)

PR #3614, opened 2025-08-15
Add .python-version file to gitignore

seanstory:
Made the mistake of setting up python12 as my system python. So now am using pyenv to set the local version lower, and didn't want to commit the generated `.python-version` file.


artem-shelkovnikov (approved):

artem-shelkovnikov:
Approved, but when I used pyenv I don't remember it happening 🤔

seanstory:
> Approved, but when I used pyenv I don't remember it happening 🤔

if you've done it the other way - making the connector version your "global", then it wouldn't.

PR #3629, opened 2025-08-19
Handle KeyError in the ES sink.py

mattnowzari:
## Closes https://github.com/elastic/connectors/issues/3628

This PR adds a `try/except` block around a dictionary access in the `_populate_stats` method in `es/sink.py`.

While rare, this method can occasionally raise a `KeyError` when the sink hits a lot of indexing errors - this will gracefully catch and log the exception at the DEBUG log level.

#### Pre-Review Checklist

### DIFF
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -307,11 +307,11 @@ class Sink:
             for op, data in item.items():
                 # "result" is only present in successful operations
                 if "result" not in data:
-                    try:
+                    if data["_id"] in stats[op]:
                         del stats[op][data["_id"]]
-                    except KeyError:
-                        self._logger.info(
-                            f"KeyError in _populate_stats: {op}, {data['_id']}"
+                    else:
+                        self._logger.debug(
+                            f"Document {data['_id']} not in stats for operation: {op}"
                         )
 
         self.counters.increment(
--- a/connectors/es/sink.py
+++ b/connectors/es/sink.py
@@ -307,7 +307,10 @@ class Sink:
             for op, data in item.items():
                 # "result" is only present in successful operations
                 if "result" not in data:
-                    del stats[op][data["_id"]]
+                    try:
+                        del stats[op][data["_id"]]
+                    except KeyError:
+                        self._logger.info(f"KeyError in _populate_stats: {op}, {data['_id']}")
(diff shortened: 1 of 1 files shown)


connectors/es/sink.py
```diff
@@ -367,6 +370,7 @@ async def _run(self):
             while True:
                 batch_num += 1
                 doc_size, doc = await self.fetch_doc()
+                print(f"Sink: fetched doc {doc}")
```
mattnowzari: Ahh! Getting rid of this

artem-shelkovnikov:
Do we know why this error happens?

connectors/es/sink.py
```diff
@@ -307,7 +307,12 @@ def _populate_stats(self, stats, res):
             for op, data in item.items():
                 # "result" is only present in successful operations
                 if "result" not in data:
-                    del stats[op][data["_id"]]
+                    try:
+                        del stats[op][data["_id"]]
```
artem-shelkovnikov, line 311: Instead of try/except I would just add explicit handling of the error - checking for the struct to have a key we're trying to delete and emit logs there.

mattnowzari: For sure, I can do it that way too

mattnowzari:
> Do we know why this error happens?

@artem-shelkovnikov This `KeyError` exception occurs when a connector gets flooded with a ton of index-time errors. Looking at the code that handles the `stats` object, it's not immediately obvious to me why the stats object would not have a document ID in it, even if the index attempt failed.

My line of thinking is: we got this far without this being an issue, and based on a customer's logs, this exception comes up maybe 1-3 times for every several _thousand_  indexing errors.

Given the rarity of this issue, I figured we can capture it as a log message for now. If we notice this issue becoming more prevalent in the future, we can do a deeper dive investigation.

artem-shelkovnikov (approved):
It could later be nice to wrap stats into a class that makes it easier to understand what's going on, but it's outside of scope of this PR

mattnowzari:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.17|[](https://github.com/elastic/connectors/pull/3635)|

### Questions ?

PR #3630, opened 2025-08-20
fix: Exclude /contentstorage/ URLs from Sharepoint Online Connector

maxesse:
## Closes https://github.com/elastic/connectors/issues/3603

The change excludes specific /contentstorage/ urls from the sync in all API calls to Sharepoint. These URLs should not be attempted to be accessed as they're created internally by Sharepoint for Teams private channels, loop components, etc. (it's fairly undocumented what they're used for to be honest), and have a different permission model that will cause 401 errors and the connector to stop syncing.

#### Pre-Review Checklist

## Release Note

Fixes an issue where a Sharepoint Online sync configured to crawl the entire tenant by selecting * in the site list, might stop with 401 errors when trying to access URLs containing /contentstorage/.

### DIFF
--- a/Dockerfile.wolfi
+++ b/Dockerfile.wolfi
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:2b6277ffc3e044d1457517491642409dbc1f6b35c5c417464cd3e96cc1d26025
+FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:c85a849c182a812950c3036b3965426c796813b221b0dd35ae254d6ae6d3f259
 USER root
 COPY . /app
 WORKDIR /app
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1854,7 +1854,7 @@ Copyright (C) 2016-present the asyncpg authors and contributors.
 
 
 attrs
-25.3.0
+25.4.0
 UNKNOWN
 The MIT License (MIT)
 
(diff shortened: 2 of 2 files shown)


artem-shelkovnikov:
buildkite test this

maxesse:
Ok I signed the contributor agreement too, hopefully the cla-checker-service will do another pass.

maxesse:
> CLA checked is happy, but test coverage fell below 92% now - you can verify it by running `make ftest`.
> 
> I also see that you've added the check in a lot of places - is it needed?
> 
> Theoretically, if you just ignore the site itself, you won't need to propagate the change down to all entities, or is it not the case?

Hi Artem - sorry for the long silence, between holidays and lots of other stuff I didn't have time to fix this. You're right, it should be enough to keep the check in the core places. I updated it.

artem-shelkovnikov:
Thanks for the writeout!

For me it was important to check what API returns this entity - and it seems like the `sites` API returns the entity that we later try to crawl and fail.

Then your code, that was also added only on the level of the sites, would exclude them and path `/contentstorage/` would not appear anywhere else.

I'm double checking because we've had similar experience with special Lists that Sharepoint has - they are available via `lists` API, but it's impossible to enumerate them as their permission model is different seemingly.

artem-shelkovnikov:
Alternatively, if you can provide the details about how to do the same setup on our side, we'll be happy to reproduce and test ourselves

artem-shelkovnikov:
That's a good writeup, thanks! I'll give it a test and get back to you

maxesse:
That's awesome Artem, happy to help!

PR #3637, opened 2025-08-22
Add support for configurable maximum queue memory size

mattnowzari:
## Part of https://github.com/elastic/connectors/issues/3584

This PR allows the maximum size of the in-memory queue for data sources to be configurable.

Currently, the following data sources have their maximum queue sizes hardcoded:
- Confluence (25 MB)
- ServiceNow (25 MB)
- Box (5 MB)
- Jira (5 MB)
- Microsoft Teams (5 MB)

### Proposed changes
The user can add the following to their `config.yml`
```yaml
connectors:
-
  connector_id: "connector_id"
  service_type: "confluence"
  api_key: "..."
elasticsearch:
  host: "https://elasticsearch-host.com:443"
  api_key: "..."
service.queue_max_mem_size: 31_457_280 # new configurable value
```

If the configured queue size is _greater_ than what is specified inside the data source itself, the maximum memory queue size will be updated. This allows us to keep the current hardcoded values for backwards compatibility while enabling users to increase them if needed.

`service.queue_max_mem_size` has a default value of `5 MB`.

`config::DataSourceFrameworkConfig` is initialized at this default value. Upon the execution of a sync job, this configuration is updated and exposed to the data source via `self.framework_config` and is updated inside each relevant data source's `get_docs` method.

This PR also unifies how default values are set to be more consistent and maintainable.

### Why do it this way?
Reading the code, it's apparent that minimal exposure of framework-level config to the data sources was desired. We **could** refactor `BaseDataSource` to attempt to pick up framework-level configuration on startup, but this requires fundamentally changing its visibility on all framework-level configurations. Initializing framework configs with default values and reading the user-set values on sync appears to be the intended way of exposing these configs to the data sources.

#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/box.py
+++ b/connectors/sources/box.py
@@ -475,7 +475,21 @@ class BoxDataSource(BaseDataSource):
             else:
                 yield item
 
-    async def get_docs(self, filtering=None):
+    def _update_queue_max_mem_size(self):
+        """
+        Helper function to update the queue's max memory size
+        if the configured value exceeds the constant defined in the data source
+        """
+        max_mem_size_from_config = self.framework_config.max_queue_mem_size
+
+        if max_mem_size_from_config > QUEUE_MEM_SIZE:
+            self.queue.update_maxmemsize(max_mem_size_from_config)
+            logger.debug(f"MemQueue max memory size updated from {QUEUE_MEM_SIZE} to {max_mem_size_from_config}")
+
+    async def get_docs(self, filtering=None): # type: ignore
+        # update queue max memory size if needed
+        self._update_queue_max_mem_size()
+
         seen_ids = set()
         root_folder = "0"
         if self.is_enterprise == BOX_ENTERPRISE:
--- a/connectors/config.py
+++ b/connectors/config.py
@@ -9,12 +9,12 @@ import os
 from envyaml import EnvYAML
 
 from connectors.logger import logger
-
-DEFAULT_ELASTICSEARCH_MAX_RETRIES = 5
-DEFAULT_ELASTICSEARCH_RETRY_INTERVAL = 10
-
-DEFAULT_MAX_FILE_SIZE = 10485760  # 10MB
-
+from connectors.utils import (
+    DEFAULT_MAX_FILE_SIZE,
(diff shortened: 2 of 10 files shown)


connectors/sources/box.py
```diff
@@ -475,7 +475,23 @@ async def _consumer(self):
...
+
+        if max_mem_size_from_config > QUEUE_MEM_SIZE:
+            self.queue.update_maxmemsize(max_mem_size_from_config)
+            logger.debug(
+                f"MemQueue max memory size updated from {QUEUE_MEM_SIZE} to {max_mem_size_from_config}"
+            )
+
+    async def get_docs(self, filtering=None):  # type: ignore
+        # update queue max memory size if needed
+        self._update_queue_max_mem_size()
```
mattnowzari, line 493: Note to self - we should should probably call this `self._update_queue_max_mem_size()` in `get_docs_incrementally()` as well, or think of a different way to update the queue max mem size on sync at the datasource level

artem-shelkovnikov (commented):
I'm not sure that this change is good for several reasons:

- It adds a configuration setting that might override datasource-specific setting in non-obvious way
- It is likely not going to address the issue

The problem with memory queue in data sources is fundamental, and increasing size of the queue can have a chance to delay when it's happening, but for large amount of data this delay might be too small to notice.

Fundamental problem is that _more documents get into the queue than the documents that get out of the queue_. In a nutshell it means that we need to either:
- remove the memory queue
- make sure that more documents are ingested than extracted on average over the course of timeout

Current error about MemQueue being full for 120 seconds means that:
not a single document has been extracted from the queue over 120 seconds. So if no documents have been extracted for 120 seconds, then increasing the queue size will not fix anything.

Likely there is a problem that we need to address that connector fills up the document queues, but documents do not get ingested to Elasticsearch - either due to a bug, or due to Elasticsearch being severely overloaded.

TLDR: we're trying to address a symptom, while we need to figure out what is causing it. Ideally, we need a reproducible way to trigger the issue locally or have extensive debug level logs that we can analyse to see why the queues do not get emptied.

PR #3692, opened 2025-09-16
Add fixture wait action and implement it in Oracle

artem-shelkovnikov:
Our CI is broken because OracleDB functional test fails.

Failure happens because the app inside OracleDB docker image takes more time to start than Elasticsearch.

In this PR I add new `wait` action that can be implemented in any fixture of functional test. If implemented, it will be executed before loading the data into 3rd-party system.

I implemented the action in only OracleDB to make the functional tests work again.

## Alternative

Alternatively, we can make the `wait` a part of `load` method, but I like that semantically they are different.

### DIFF
--- a/tests/sources/fixtures/oracle/docker-compose.yml
+++ b/tests/sources/fixtures/oracle/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       - esnet
 
   oracle:
-    image: container-registry.oracle.com/database/free:latest
+    image: container-registry.oracle.com/database/free:23.8.0.0
     ports:
       - 1521:1521
     environment:


tests/sources/fixtures/oracle/fixture.py
```diff
@@ -62,6 +63,14 @@ def inject_lines(table, cursor, lines):
         print(f"Inserted batch #{batch} of {batch_size} documents.")
 
 
+@retryable(retries=7, interval=2, strategy=RetryStrategy.EXPONENTIAL_BACKOFF)
+async def wait():
```
lorenabalan, line 67: Where is this being called though? Are we not passing it as a Pytest fixture, just rely on fixtures/fixture:main?

artem-shelkovnikov: It's called via `$PYTHON fixture.py --name $NAME --action wait` - this in turn calls this function by name in fixture.py

artem-shelkovnikov: This code is not called by `pytest` at all - functional tests are facilitated with ftesh.sh file and fixture.py only

jedrazb (approved):
lol just throw more retries

artem-shelkovnikov:
Yeah I feel like something's wrong with the container - current policy is 2 minutes of retries after Elasticsearch starts. So container has around 3 mins to start. Locally it starts reasonably fast, but in CI it cannot, and CI machine is 8-core 30GB RAM, it should not have trouble with a small OracleDB

PR #3699, opened 2025-09-17
Add https-only validation to github connector

Apmats:
## Closes https://github.com/elastic/connectors-py/issues/3683

As per the above issue, using github URLs with a plain http protocol results in unauthorized errors. 
In this PR we're going to be adding a validation step for github connector configuration, to only allow https URLs.

Since configuration for connectors is already tied to a validation mechanism, we're simply going to use that to validate that URLs fit a pattern.

In the

#### Pre-Review Checklist

### DIFF
--- a/tests/sources/test_github.py
+++ b/tests/sources/test_github.py
@@ -5,6 +5,7 @@
 #
 """Tests the Github source class methods"""
 
+from aioresponses import aioresponses
 from contextlib import asynccontextmanager
 from copy import deepcopy
 from http import HTTPStatus
@@ -815,6 +816,7 @@ async def create_github_source(
     repo_type="other",
     org_name="",
     repos="*",
+    host="github.example.com",
     use_document_level_security=False,
     use_text_extraction_service=False,
 ):
@@ -828,6 +830,7 @@ async def create_github_source(
         repositories=repos,
         repo_type=repo_type,
         org_name=org_name,
+        host=host,
         ssl_enabled=False,
         use_document_level_security=use_document_level_security,
         use_text_extraction_service=use_text_extraction_service,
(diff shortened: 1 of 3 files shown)


tests/sources/test_github.py
```diff
@@ -863,6 +866,34 @@ async def test_validate_config_missing_fields_then_raise(field):
...
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "host_url,should_raise",
+    [
+        pytest.param("http://github.example.com", True, id="HTTP URL should be rejected"),
+        pytest.param("https://github.example.com", False, id="HTTPS URL should be accepted"),
+    ]
+)
+async def test_validate_config_https_url_validation(host_url, should_raise):
```
Apmats, line 881: The other tests in this test module mock methods like `get_personal_access_token_scopes` - but that method seems like an implementation detail (maybe it's even public just to allow mocking which seems like a code smell).

Instead of doing the same here I mock a response from the remote server - I guess this is a bit less brittle to refactoring but also assumes things about the connector internals. 

Whoever is reviewing, if you have an opinion on doing things one or the other way, let me know.

artem-shelkovnikov: I like using `aioresponses` here

artem-shelkovnikov (commented):
Looks good - I see functional test is failing because it points to `http` host - you'll need to update it in https://github.com/elastic/connectors/blob/main/tests/sources/fixtures/github/connector.json

I also see that the issue description is not finished - can you finish it and check the checkboxes in it?

tests/sources/fixtures/github/fixture.py
```diff
@@ -345,4 +345,4 @@ def get_scopes(self):
 
 
 if __name__ == "__main__":
-    GitHubAPI().app.run(host="0.0.0.0", port=9091)
+    GitHubAPI().app.run(host="0.0.0.0", port=9091, ssl_context="adhoc")
```
Apmats, line 348: Had no idea what was going wrong with tests here but with a little genai help...

artem-shelkovnikov (approved):

PR #3701, opened 2025-09-17
Define pyproject.toml

lorenabalan:
## Closes https://github.com/elastic/search-team/issues/10998

Add a `pyproject.toml` as per [PEP-621](https://peps.python.org/pep-0621/) and [PEP-518](https://peps.python.org/pep-0518/). Packaging still uses `setuptools` as build backend, as it was before, but more modern Pythonic interface. This will more easily enable us to build Python packages, swap build backends, declare dependencies and set up tools configuration.

Most of the changes were due to `python setup.py develop` --> `pip install -e .` change and to our current folder structure.

Notes:

* [PyPA recommends the `src` layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) - which I tried in https://github.com/elastic/connectors/pull/3693 - however  there are quite a few changes that need to happen to enable it, so I left it out of the scope of this PR. 
  * Moreover, when we break up the main package into a set of separate mini-packages, we might want to consider swapping the build-backed to `uv` and use its [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/) feature to manage one common `pyproject.toml` (e.g. for common linting settings, etc) _as well as_ a `pyproject.toml` per package. AFAIK Python doesn't support that on its own, it would force us to have duplicate code across all `pyproject.toml` files. We can cross that bridge when we get to it, see https://docs.astral.sh/uv/concepts/projects/workspaces/#when-not-to-use-workspaces.

### References
https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
https://chatgpt.com/share/68cd3746-cc24-8005-8718-fc271e6f6b8b

## Testing

CI passes. Also locally:
```
make sdist
```

```
make install
```

```
pip install -e .

elastic-ingest --help
fake-kibana --help
connectors --help
test-connectors --help
```
all work without throwing errors.

#### Pre-Review Checklist


Makefile
```diff
@@ -48,7 +47,7 @@ install-agent: .venv/bin/elastic-ingest
 .venv/bin/elastic-ingest: .venv/bin/python requirements/framework.txt requirements/$(ARCH).txt requirements/agent.txt
 	.venv/bin/pip install -r requirements/$(ARCH).txt
 	.venv/bin/pip install -r requirements/agent.txt
-	.venv/bin/python setup.py develop
+	.venv/bin/pip install -e .
```
lorenabalan, line 50: Calling `python setup.py` is deprecated, using recommended alternative for this as per [table](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead)

pyproject.toml
```diff
@@ -0,0 +1,120 @@
...
+]
+
+[project.scripts]
+elastic-ingest = "connectors.service_cli:main"
+fake-kibana = "connectors.kibana:main"
+connectors = "connectors.connectors_cli:main"
+test-connectors = "scripts.testing.cli:main"
+
+[tool.setuptools]
+packages = ["connectors"]
```
lorenabalan: Without this, the automatic package discovery gets confused

Image

We don't add `tests` because they shouldn't be packaged up with the .whl file. 
I've tried the second suggestion in a [separate PR](https://github.com/elastic/connectors/pull/3693) but that requires a lot more changes that blows up the scope a bit, so keeping this PR to a minimum.

tests/ftest.sh
```diff
@@ -246,6 +246,9 @@ fi
 
 cd $ROOT_DIR/tests/sources/fixtures
 
+# Add project root to PYTHONPATH so tests module can be imported
+export PYTHONPATH="$ROOT_DIR${PYTHONPATH:+:$PYTHONPATH}"
```
lorenabalan, line 250: This is to address failures like [this one](https://buildkite.com/elastic/connectors/builds/19554#01995830-19b9-44c1-b000-23b976eb8e8a)

```

+ /opt/buildkite-agent/builds/bk-agent-prod-gcp-1758121194693719806/elastic/connectors/tests/../.venv/bin/python fixture.py --name github --action setup
--
  | [FTEST][15:04:36][INFO] Executing action: 'setup'.
  | Traceback (most recent call last):
  | File "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1758121194693719806/elastic/connectors/tests/sources/fixtures/fixture.py", line 234, in <module>
  | loop.run_until_complete(main())
  | File "/opt/buildkite-agent/.pyenv/versions/3.11.13/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
  | return future.result()
  | ^^^^^^^^^^^^^^^
  | File "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1758121194693719806/elastic/connectors/tests/sources/fixtures/fixture.py", line 203, in main
  | spec.loader.exec_module(module)
  | File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  | File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  | File "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1758121194693719806/elastic/connectors/tests/sources/fixtures/github/fixture.py", line 14, in <module>
  | from tests.commons import WeightedFakeProvider
  | ModuleNotFoundError: No module named 'tests'
  | make: *** [Makefile:90: ftest] Error 1


```

Because of the switch to `pip install -e .` it doesn't find `tests` as importable - so we need to either update the PYTHONPATH or run `python -m pytest` instead of `pytest`. https://docs.pytest.org/en/stable/explanation/pythonpath.html?utm_source=chatgpt.com#invoking-pytest-versus-python-m-pytest

pyproject.toml
artem-shelkovnikov: Makes a lot of sense, let's start like this and maybe try `src` layout during the package separation

tests/ftest.sh
artem-shelkovnikov: Overall `ftest` stucture is super awkward, something that we'll need to address later. It's been bugging me for a while

tests/sources/fixtures/box/docker-compose.yml
```diff
@@ -28,7 +28,7 @@ services:
     build:
       context: ../../../../
       dockerfile: ${DOCKERFILE_FTEST_PATH}
-    command: .venv/bin/python tests/sources/fixtures/box/fixture.py
+    command: .venv/bin/python -m tests.sources.fixtures.box.fixture
```
lorenabalan, line 31: https://chatgpt.com/share/68cd3746-cc24-8005-8718-fc271e6f6b8b

Dockerfile
```diff
@@ -8,7 +8,9 @@ COPY --chown=nonroot:nonroot . /app
 
 USER nonroot
 WORKDIR /app
+RUN echo "PYTHONPATH from base: $PYTHONPATH"
```
artem-shelkovnikov:
```suggestion
```

Dockerfile
```diff
@@ -8,7 +8,9 @@ COPY --chown=nonroot:nonroot . /app
...
 RUN make clean install
+RUN echo "PYTHONPATH from base after make clean install: $PYTHONPATH"
```
artem-shelkovnikov:
```suggestion
```

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -138,7 +138,7 @@ if [[ "${PUBLISH_SNAPSHOT:-}" == "true" ]]; then
   generateDependencyReport $DEPENDENCIES_REPORTS_DIR/$dependencyReportName
 
   echo "-------- Publishing SNAPSHOT DRA Artifacts"
-  cp $RELEASE_DIR/dist/elasticsearch_connectors-${VERSION}.zip $DRA_ARTIFACTS_DIR/connectors-${VERSION}-SNAPSHOT.zip
+  cp $RELEASE_DIR/dist/elasticsearch_connectors-${VERSION}.tar.gz $DRA_ARTIFACTS_DIR/connectors-${VERSION}-SNAPSHOT.tar.gz
```
artem-shelkovnikov: Would be great to confirm that it doesn't break stuff (zip -> tar.gz move). We might need to edit this somewhere, @seanstory might remember this as we've discussed this

tests/Dockerfile.ftest
```diff
@@ -1,3 +1,5 @@
 FROM connectors-base
 
+RUN echo "PYTHONPATH: $PYTHONPATH"
+
```
artem-shelkovnikov:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,120 @@
...
+]
+classifiers = [
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3 :: Only",
+]
+requires-python = ">=3.10,<3.12"
+dependencies = [
```
mattnowzari, line 21: Asking bc I don't know, not bc this is right/wrong - what is the relation between this dependency list and `requirements.txt`? Does having this technically negate needing a `requirements.txt`? Obviously we are still using them in our Makefile, but does this mean we can eventually move away from having them at all?

Also, can we do architecture-specific dependency lists (aarch64, arm64 etc) in this .toml file?

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: oh, yes this will break agentless I believe. The agent image currently consumes our zip.
I don't think it's the end of the world, we'd just need to decide if we definitely want tar.gz instead, and if so, we have to update the DRA manifest, and update the agent image build to pick up the new artifact and install it correctly.

Might be better to leave this for later

pyproject.toml
artem-shelkovnikov: Oh yeah, I know that with pyproject.toml we can eventually drop requirements.txt and just have this file as a source of versions. Can we do it as a part of this work (not necessarily this PR, just the ticket), or can do later?

lorenabalan: We would eventually move away to having them only here (source of truth), and if we need to install them anywhere else we'd generate the `requirements.txt` on the fly based on `pyproject.toml`. But actually let me check, I think I might be able to get rid of some of the `pip install -r` in the `Makefile` as `pip install .` should install the dependencies too. 

> Also, can we do architecture-specific dependency lists (aarch64, arm64 etc) in this .toml file?

Yes, we can specify any of the [environment markers](https://peps.python.org/pep-0508/#environment-markers) we could have in requirements.txt - in this case 

```
dependencies = [
    ...
    'SQLAlchemy[asyncio]==2.0.1; platform_machine == "arm64"',
]
```

.buildkite/publish/dra/init_dra_publishing.sh
artem-shelkovnikov: I had trouble generating zips with `hatch`, but `setuptools` might allow for this.

Then would be nice to keep ZIPs and when we can package connectors as a PyPi package swap agentless to consume the package rather than archives

artem-shelkovnikov:
I dismissed my PR as it'll break agentless due to tar.gz change

pyproject.toml
artem-shelkovnikov: We've had env markers because we've started making connectors at a time when Apple just released M1s. I think at this point most packages support both - might be nice to check it and just drop separate requirements files.

.buildkite/publish/dra/init_dra_publishing.sh
lorenabalan: I wasn't able to generate zips with `setuptools` either, even if I pin it to the old version. 😬 But I can update Makefile so that `make sdist` converts the tar.gz to zip that that should give us the same result.

lorenabalan: Yeah apparently `zip` format was only possible with `setup.py`, it's not been ported to the [PEP-517](https://peps.python.org/pep-0517/)-compatible backend, i.e. not possible to specify zip format with `pyproject.toml`. I've added `make zip` in [eec0558](https://github.com/elastic/connectors/pull/3701/commits/eec0558448ad71d76506eff4f808a5cec8e7d24d) instead. 😊

pyproject.toml
lorenabalan: I removed a bunch of references in [47b1171](https://github.com/elastic/connectors/pull/3701/commits/47b117199464fa411a3db415411cc735d679bfe4) and deleted `agents.txt`. There are only 3 references to the requirements files (2 in Makefile, under `.venv/bin/ruff` and 1 in `Dockerfile.ftest`). I'll address those in a follow-up PR (and then we can delete `requirements/` folder), I need to think a bit more about them.

artem-shelkovnikov (approved):
🙌

PR #3708, opened 2025-09-19
Am/excessive gh config validation reqs

Apmats:
#### Pre-Review Checklist

### DIFF
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -1516,8 +1516,7 @@ class GitHubDataSource(BaseDataSource):
         # Run remote validation with caching to avoid excessive API calls
         await _validation_cache.run_validation(
             configuration=self.configuration,
-            validation_func=self._remote_validation,
-            logger_instance=self._logger
+            validation_func=self._remote_validation
         )
 
     async def close(self):
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -5,10 +5,8 @@
 #
 """GitHub source module responsible to fetch documents from GitHub Cloud and Server."""
 
-from datetime import datetime, timedelta
 import json
 import time
-from dataclasses import dataclass
 from enum import Enum
 from functools import cached_property, partial
 
@@ -36,7 +34,7 @@ from connectors.logger import logger
 from connectors.source import BaseDataSource, ConfigurableFieldValueError
 from connectors.utils import (
     CancellableSleeps,
-    HTTPS_URL_PATTERN,
+    ValidationCache,
     RetryStrategy,
     decode_base64_value,
     nested_get_from_dict,
@@ -61,15 +59,8 @@ UNAUTHORIZED = 401
 NODE_SIZE = 100
 REVIEWS_COUNT = 45
 
-@dataclass
-class ConfigValidationRun:
-    """Tracks validation run time and any error that occurred."""
-    validation_time: datetime
-    error: Exception | None = None
-
-
-_CONFIG_VALIDATION_RUNS = {}
-CONFIG_VALIDATION_EXPIRY_SECONDS = 60
+# Validation cache with 300 second expiry for remote validations
+_validation_cache = ValidationCache(expiry_seconds=300)
 
 SUPPORTED_EXTENSION = [".markdown", ".md", ".rst"]
 
@@ -1521,45 +1512,13 @@ class GitHubDataSource(BaseDataSource):
         Also validate, if user configured repositories are accessible or not and scope of the token
         """
         await super().validate_config()
-        
-        # only run remote validation if config hasn't been encountered yet or
-        # we think the validation is stale - this is to avoid hitting rate limits
-        config_hash = self.configuration.hash_digest()
-        last_validation_run = _CONFIG_VALIDATION_RUNS.get(config_hash)
 
-        #clean up old entries
-        for key in list(_CONFIG_VALIDATION_RUNS.keys()):
-            if key == config_hash:
-                continue
-            if (
-                datetime.now() - _CONFIG_VALIDATION_RUNS[key].validation_time
-            ) > timedelta(seconds=CONFIG_VALIDATION_EXPIRY_SECONDS):
-                del _CONFIG_VALIDATION_RUNS[key]
-
-    
-        if last_validation_run and (
-            datetime.now() - last_validation_run.validation_time
-        ) < timedelta(seconds=CONFIG_VALIDATION_EXPIRY_SECONDS):
-            self._logger.error(
-                "Skipping remote validation of config as it was validated recently"
-            )
-            # Re-raise the cached error if one occurred during the last validation
-            if last_validation_run.error:
-                raise last_validation_run.error
-        else:
-            # Run remote validation and cache the result
-            validation_error = None
-            try:
-                await self._remote_validation()
-            except Exception as e:
-                validation_error = e
-                raise
-            finally:
-                # Always update the validation run info, whether successful or not
-                _CONFIG_VALIDATION_RUNS[config_hash] = ConfigValidationRun(
-                    validation_time=datetime.now(),
-                    error=validation_error
-                )
+        # Run remote validation with caching to avoid excessive API calls
+        await _validation_cache.run_validation(
+            configuration=self.configuration,
+            validation_func=self._remote_validation,
+            logger_instance=self._logger
+        )
 
     async def close(self):
         await self.github_client.close()
--- a/connectors/sources/github.py
+++ b/connectors/sources/github.py
@@ -5,8 +5,10 @@
 #
 """GitHub source module responsible to fetch documents from GitHub Cloud and Server."""
 
+from datetime import datetime, timedelta
 import json
 import time
+from dataclasses import dataclass
 from enum import Enum
 from functools import cached_property, partial
 
@@ -34,6 +36,7 @@ from connectors.logger import logger
 from connectors.source import BaseDataSource, ConfigurableFieldValueError
 from connectors.utils import (
     CancellableSleeps,
+    HTTPS_URL_PATTERN,
     RetryStrategy,
     decode_base64_value,
     nested_get_from_dict,
@@ -57,6 +60,16 @@ FORBIDDEN = 403
 UNAUTHORIZED = 401
 NODE_SIZE = 100
 REVIEWS_COUNT = 45
+
+@dataclass
+class ConfigValidationRun:
+    """Tracks validation run time and any error that occurred."""
+    validation_time: datetime
+    error: Exception | None = None
+
+
+_CONFIG_VALIDATION_RUNS = {}
+CONFIG_VALIDATION_EXPIRY_SECONDS = 60
 
 SUPPORTED_EXTENSION = [".markdown", ".md", ".rst"]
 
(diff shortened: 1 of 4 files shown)


connectors/sources/github.py
```diff
@@ -1002,6 +1036,98 @@ async def get_foreign_repo(self, repo_name):
...
+                }}
+                forkCount
+                createdAt
+                isArchived
+            }}"""
+            query_parts.append(query_part)
+
+        # Build variable declarations for the query
+        variable_declarations = []
+        for var_name in variables.keys():
```
lorenabalan: nit: iterating through a dict iterates through the keys by default anw
```suggestion
        for var_name in variables:
```

lorenabalan: actually we can simplify this into a list comprehension 
```python
variable_declarations = [f"..." for var_name in variables]
```

connectors/sources/github.py
```diff
@@ -1002,6 +1036,98 @@ async def get_foreign_repo(self, repo_name):
...
+
+        data = await self.graphql(
+            query=query, variables=variables, ignore_errors=["NOT_FOUND"]
+        )
+
+        # Map results back to repo names
+        results = {}
+        for i, repo_name in enumerate(repo_names):
+            alias = f"repo{i}"
+            results[repo_name] = data.get(alias)
```
lorenabalan: Similar here, we can use a dict comprehension

connectors/sources/github.py
```diff
@@ -807,12 +813,40 @@ async def graphql(self, query, variables=None):
...
+                        if error.get("type") not in ignore_errors
+                    ]
+
+                    if other_errors:
+                        # If there are other types of errors, still raise an exception
+                        msg = f"Error while executing query. Exception: {exception.response.get('errors')}"
+                        raise Exception(msg) from exception
+
+                    # All errors are ignored, return just the data part without errors
+                    return exception.response.get("data", {})
```
lorenabalan: There are a few ifs here that make a bit harder to reason about, could this effectively be simplified to the below?

```python
ignore_errors = ignore_errors or []
other_errors = [error for error in errors if error.get("type") not in ignore_errors]

if other_errors:
    raise ...

return exception.response.get("data", {})
```

connectors/sources/github.py
```diff
@@ -1327,18 +1441,62 @@ async def _get_invalid_repos_for_github_app(self):
...
+            if owner not in self._installations:
+                self._logger.debug(
+                    f"Invalid repo {full_repo_name} as the github app is not installed on {owner}"
+                )
                 invalid_repos.add(full_repo_name)
+                continue
+
+            if owner not in repos_by_owner:
+                repos_by_owner[owner] = []
+            repos_by_owner[owner].append(full_repo_name)
```
lorenabalan: We can use a `defaultdict(list)` (or even `defaultdict(set)`?) for `repos_by_owner` so we handle this in one line.

connectors/sources/github.py
```diff
@@ -807,13 +813,36 @@ async def graphql(self, query, variables=None):
...
+
+            errors_to_report = exception.response.get("errors")
+
+            if ignore_errors and all(
+                error.get("type") in ignore_errors for error in errors
+            ):
+                # All errors are ignored, return just the data part without errors
+                return exception.response.get("data", {})
+
+            if ignore_errors:
```
lorenabalan: Optional, we could get rid of the `if ignore_errors` condition in both places, and handle the default case at the top of the function `ignore_errors = ignore_errors or []`. That way we don't need L829, errors_to_report will always be non_ignored_errors. To me it's a little easier to read, but up to you. 

```python
if all(...):
    return exception.response.get()

non_ignored_errors = [error for error in errors if error.get("type") not in ignore_errors]
raise Exception(...non_ignored_errors)
```

connectors/sources/github.py
lorenabalan: Outdated, ignore this one

connectors/sources/github.py
```diff
@@ -197,6 +197,11 @@ class GithubQuery(Enum):
     }
     }
     """
+    BATCH_REPO_QUERY_TEMPLATE = """
```
lorenabalan, line 201: I don't see this being used though. Should it be in L1103?

connectors/sources/github.py
Apmats: Okay, did that. I can't do a set because it's used via offsets for batching later.

connectors/sources/github.py
Apmats: Done!

connectors/sources/github.py
Apmats: Yeah this was already me doing a first pass refactoring of the extremely verbose code genai gave me. I did what you said pretty much now.

connectors/sources/github.py
Apmats: Yeah, refactored and will retest.

connectors/sources/github.py
Apmats: Done.

Apmats:
Addressed review comments from @lorenabalan and also added some typing. Another round of manual testing and a self-review pending - I think I can for sure add some more types.

connectors/sources/github.py
```diff
@@ -807,13 +822,24 @@ async def graphql(self, query, variables=None):
...
-                    error.get("type").lower() == "rate_limited"
+                    error.get("type", "").lower() == "rate_limited"
                     and "api rate limit exceeded" in error.get("message").lower()
                 ):
                     await self._put_to_sleep(resource_type="graphql")
-            msg = f"Error while executing query. Exception: {exception.response.get('errors')}"
+
+            if all(error.get("type") in ignore_errors for error in errors):
+                # All errors are ignored, return just the data part without errors
+                return exception.response.get("data", {})
```
artem-shelkovnikov, line 835: This deserves a debug log

connectors/sources/github.py
```diff
@@ -1002,6 +1028,93 @@ async def get_foreign_repo(self, repo_name):
...
+        Args:
+            repo_names (list): List of fully qualified repository names (owner/repo format) to validate
+            batch_size (int): Number of repos to validate per request (default 15)
+
+        Returns:
+            dict: Dictionary mapping repo_name to repository object (or None if invalid)
+        """
+        results = {}
+
+        # Process repositories in batches to avoid hitting GraphQL complexity limits
```
artem-shelkovnikov, line 1045: What's up with complexity limits? How do we avoid this? Just sending specific numbers of requests in the batch is enough? What if this rule changes on Github side, will we get an error message about it?

connectors/sources/github.py
```diff
@@ -1586,6 +1743,9 @@ async def _fetch_installations(self):
         return self._installations
 
     async def _get_personal_repos(self, user):
```
artem-shelkovnikov, line 1743: I'm actually wondering if these methods are needed now.

Your code collects the data for select repositories, and it can be used in both validation and execution logic

connectors/sources/github.py
Apmats: I think I hit an alias limit while testing this out, but I can't reproduce now. I might be going crazy.
 
Complexity limits I think is about how many nodes you can reach in a single query - currently they are here https://docs.github.com/en/enterprise-cloud@latest/graphql/overview/rate-limits-and-query-limits-for-the-graphql-api#node-limit - I don't understand fully though and I don't know how much of a subject to change they are. 

Hitting them would result in an error that would get raised here and captured in the connector config error I believe (since we only handle rate limit errors and retry after waiting), but I think with the current query structure (which doesn't have nesting) it would require you to have configured a pathologically large number of repos to sync.
But having batching already built in helps prevent any such pathologic case or any change to the downstream query that adds to that complexity.

Another problem with a big number of handling a big number of repos here is the fact that we would hit timeouts probably consistently:
https://docs.github.com/en/enterprise-cloud@latest/graphql/overview/rate-limits-and-query-limits-for-the-graphql-api#timeouts
Testing with 300 repos seems to take me close to timeout territory.

So a) batching will not come into play in realistic scenarios (I can increase to like 50, and most users won't be syncing against more than 50 repos I would guess) and b) if it does come into play it might help us avoid some errors.

What are your thoughts?

connectors/sources/github.py
Apmats: You think? Even if it's the "happy path" for code we write in the future - meaning they'll always see this debug on every request? 

I'll add it.

artem-shelkovnikov: I don't think we need it for the happy path, but we need it for "all errors ignored" - when errors were there, but we ignored some or all

Apmats: I meant, given you're passing the errors you don't really consider errors for your case, hitting them is within the happy path 🤔 . 

Eg. when this is called like in this PR to check for multiple repos' existence.
OR
if you use the same method to verify repo doesn't exist and thus was indeed deleted, again, part of the happy path in that case.

artem-shelkovnikov: I'd look at it from the perspective of SDH or a customer who doesn't look into the connector code base at all.

We roll something out, then there comes the question of "hmmmm, something went wrong - I don't get the data, lemme look at logs" and logs should answer this.

In the context of this PR I want to understand whether we log sufficient amount of info to understand whether the repository is there or not (I don't think the `errors` parameter is passed elsewhere, right?). Now thinking of it, we likely do, because we log a warning saying "repos A, B and C are not available".

If, however, more methods in code start using the method, for debugging/SDH/troubleshooting purpose this can be useful?

So in the end yeah, likely no need to log anything here _yet_.

Apmats: In this case, this always results in an error being raised when we ignore this errors within the context of a GraphQL query, so yes we get visibility.

I don't know why I had to debate this, I'll add the log line. If this ends up generating too many logs it's your/our fault for running in debug mode in a prod-like environment, turn it off! 😅

connectors/sources/github.py
Apmats: This path is still valid because we support wildcards - we don't know the repos to fetch, we fetch all-repos-for-user here for example.

We could probably refactor to do this in a 2 step manner, first find repo names then fetch with the same method as we do in the other case, but maybe selfishly I want to say I'd rather not make this change for this stupid little optimization bigger?
I'll think a bit if we gain/lose something by trying to unify these 2 paths.

connectors/sources/github.py
artem-shelkovnikov: I do agree with the debate though, we will end up reading debug logs.

connectors/sources/github.py
artem-shelkovnikov: Yeah makes sense. The PR is already big enough, adding more to it will make it impossible to merge. I forgot about the wildcards totally, never used it for this connector

connectors/sources/github.py
artem-shelkovnikov: All makes sense, thanks for the explanation

connectors/sources/github.py
```diff
@@ -1371,50 +1528,50 @@ async def _get_repo_object_for_github_app(self, owner, repo_name):
...
-                    self.org_repos[self.configuration["org_name"]][
-                        repo["nameWithOwner"]
-                    ] = repo
-                invalid_repos = list(
-                    set(configured_repos)
-                    - set(self.org_repos[self.configuration["org_name"]].keys())
+                all_repos = configured_repos + foreign_repos
+
+            invalid_repos = []
+            if all_repos:
```
lorenabalan: Do we need the if? If it's empty won't `self.github_client.get_repos_by_fully_qualified_name_batch` return empty dict?

connectors/sources/github.py
```diff
@@ -1371,50 +1528,50 @@ async def _get_repo_object_for_github_app(self, owner, repo_name):
...
+                                    self.user_repos[logged_in_user] = {}
+                                self.user_repos[logged_in_user][repo_name] = repo_data
+                            else:
+                                org_name = self.configuration["org_name"]
+                                if org_name not in self.org_repos:
+                                    self.org_repos[org_name] = {}
+                                self.org_repos[org_name][repo_name] = repo_data
+                    else:
+                        self._logger.debug(f"Detected invalid repository: {repo_name}.")
+                        invalid_repos.append(repo_name)
```
lorenabalan: Can we reduce the nesting by one level by capturing this at the top of the loop

```python
if not repo_data:
    invalid_repos.appen()
    logger.debug...
    continue
```

I like avoiding messing with the loop control flow but I think this is one of the cases where it's welcome, to make the rest of the code easier to reason about.

lorenabalan: Also, optional, these type of snippets
```python
if org_name not in self.org_repos:
    self.org_repos[org_name] = {}
self.org_repos[org_name][repo_name] = repo_data
```

can become

```python
self.org_repos.setdefault(org_name, {})[repo_name] = repo_data
```

but I don't have a strong opinion if you think the current version is more readable.

lorenabalan (approved):
LGTM, only small cosmetic suggestions. Thanks for adding typehints along the way! 🙌

connectors/sources/github.py
Apmats: Indeed. Nice, thanks.

connectors/sources/github.py
Apmats: I like the second suggestion - I prefer the if-else in the first one but not enough to argue back 😁 so applied both.

lorenabalan (dismissed):

Apmats:
@artem-shelkovnikov @lorenabalan Apologies - I'll need another approval post-autoformat (my bad on this, still not used to this) and post-merge.

lorenabalan (approved):
LGTM!

PR #3710, opened 2025-09-22
Update Jira API

lorenabalan:
## Closes https://github.com/elastic/search-team/issues/10914

Jira has removed APIs we rely on - [we will start seeing more 410s](https://community.atlassian.com/forums/Jira-questions/Sunset-schedule-for-Issue-API-CHANGE-2046/qaq-p/3105876). The final-final deadline seems to be [end of October 2025](https://community.atlassian.com/forums/Jira-questions/When-are-JQL-search-endpoints-rest-api-2-search-and-rest-api-3/qaq-p/3029221).

### Key differences
* The new APIs use a different pagination method (cursor-based), while other endpoints use the old way (offset-based). We have to handle both cases in our code.
  * `nextPageToken` is a new query param in order to get results from different pages; however `nextPageToken=null` (first page) is **not** allowed in the GET request. It **must** represent the token of a page _after_ the first page.
* The new API only allows **bounded** JQL, meaning we can't use empty-string as a catch-all for "all issues"
* The new API returns only `id` field by default, unless `fields` query param is specified.
See [documentation](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-jql-get)

## Testing
```
ELASTICSEARCH_VERSION=9.2.0-SNAPSHOT KIBANA_VERSION=9.2.0-SNAPSHOT CONNECTORS_VERSION=9.2.0-SNAPSHOT scripts/stack/run-stack.sh
```

Navigate to localhost:5601 and create Jira connector.

```
make run
```

Run a sync - no more 410s or warnings in the logs, and docs are synced rather than the previous 15.
Screenshot 2025-09-26 at 11 59 49

Screenshot 2025-09-26 at 11 58 29

#### Pre-Review Checklist

## Release Note

The Jira connector now uses the newer `/rest/api/3/search/jql` endpoint


connectors/sources/jira.py
```diff
@@ -283,26 +292,40 @@ async def paginated_api_call(self, url_name, jql=None, **kwargs):
...
                 ):
                     response_json = await response.json()
-                    total = response_json["total"]
                     yield response_json
-                    if start_at + FETCH_SIZE > total or total <= FETCH_SIZE:
-                        return
-                    start_at += FETCH_SIZE
+
+                    # is there another page? if not, we're done, otherwise update cursor/offset.
+                    if is_cursor_based_pagination:
```
artem-shelkovnikov: Would it be hard to split this logic into multiple methods? e.g. `paginated_api_call_with_cursor` and `paginated_api_call_with_offset`, and choose method based on entity that's calling this API

lorenabalan: Nope, not hard - I called this out as personal preference in the PR description but can easily split this method into 2, it'll just have a lot of similar code. DRY vs DAMP etc 😂

artem-shelkovnikov: I'm happy to keep it as is too, we can bundle refactors/rewrites into bigger changes in the connector if we need to

connectors/sources/jira.py
```diff
@@ -270,10 +272,17 @@ async def paginated_api_call(self, url_name, jql=None, **kwargs):
         Yields:
             response: Return api response.
         """
-        start_at = 0
+        is_cursor_based_pagination = url_name == ISSUES
+        pagination_arg = (
+            {"next_page_token": "null"}
```
Apmats: I think this will break - did you test against an actual Jira instance? 

https://jira.atlassian.com/browse/JRACLOUD-94632

```
URLs don't natively support non-string types. When we explicitly pass nextPageToken=null in a search/JQL GET request, it gets converted to {{
{ "nextPageToken": "null" }

}}

In a GET request the user should not be passing any value for nextPageToken if they want to set it to null, this param is anyway optional. Explicitly setting nextPageToken=null in GET means that it will be converted to "null" string. We should be throwing invalid NPT error back to the user in this case. This is expected behaviour.
```

I think we need to conditionally add it to the url if we want to keep using url params.

connectors/sources/jira.py
```diff
@@ -283,26 +292,40 @@ async def paginated_api_call(self, url_name, jql=None, **kwargs):
...
+                    else:  # offset-based pagination
+                        total = response_json["total"]
+                        start_at = pagination_arg["start_at"]
+                        if start_at + FETCH_SIZE > total or total <= FETCH_SIZE:
+                            return
+                        start_at += FETCH_SIZE
+                        pagination_arg["start_at"] = start_at
+
             except Exception as exception:
+                pagination_str = ", ".join(
```
Apmats: Nitpick, but this representation is different than the one generated above 😁  I would extract into an inline function and call it in both cases.

connectors/sources/jira.py
```diff
@@ -890,10 +913,10 @@ async def _get_projects(self):
             )
 
     async def _put_issue(self, issue):
-        """Put specific issue as per the given issue_key in a queue
+        """Put a specific issue as per the given issue_key in a queue
 
         Args:
-            issue (str): Issue key to fetch an issue
+            issue (dict): Issue key to fetch an issue
```
Apmats: Not your change, but this seems to be lying... 
Maybe `issue containing issue key for issue to be put into queue` or something? I'm missing context...

connectors/sources/jira.py
```diff
@@ -270,10 +272,17 @@ async def paginated_api_call(self, url_name, jql=None, **kwargs):
         Yields:
             response: Return api response.
         """
-        start_at = 0
+        is_cursor_based_pagination = url_name == ISSUES
```
Apmats: Super nitpick - I would already go into the effort of maintaining a list of demands-cursor-pagination-urls. Would make it super clear to anyone reading, and also make it easier to identify what we do when we have to transition the next endpoint to this handling.

connectors/sources/jira.py
lorenabalan: > I think this will break - did you test against an actual Jira instance?

Oh no, I was going to before I merged. That's annoying, their documentation led me to believe sth else. I'll update here after testing.

connectors/sources/jira.py
lorenabalan: I thought about that as well and had it originally, but preferred this option purely for performance sake, identity check is slightly quicker than set membership. I figured we're not sure yet if we'll have more cursor-based APIs to rely on, we can update then if it changes.

connectors/sources/jira.py
lorenabalan: Annoyingly it was on purpose because one uses `=` and one `:`. 😂 I can update both log message to use the same symbol though

Apmats: Yeah, I saw they're slightly different too, without a reason - if you're looking through logs through some exact-grep-like-tool you'd want to find both the initial and the error log for the same thing.

connectors/sources/jira.py
lorenabalan: Yeah I need to revisit this, thanks for flagging. Also seems like I missed the [detail](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-jql-get-request-Query%20parameters) around `jql`, empty string is not allowed anymore.

connectors/sources/jira.py
lorenabalan: Yeah docstrings in general need a bit more TLC. Had a go at improving this one in [148118d](https://github.com/elastic/connectors/pull/3710/commits/148118d0b9cd2eeaf47f0d480ad6fdc990e6eed8)

connectors/sources/jira.py
lorenabalan: I've addressed this in subsequent commits! Thanks again for the push, testing this locally seems happy now. ✨

connectors/sources/jira.py
```diff
@@ -958,9 +1018,9 @@ async def _get_issues(self, custom_query=""):
             Dictionary: Jira issue to get indexed
             issue (dict): Issue response to fetch the attachments
         """
-        wildcard_query = ""
-        comma_seperated_projects = '"' + '","'.join(self.jira_client.projects) + '"'
-        projects_query = f"project in ({comma_seperated_projects})"
+        wildcard_query = "key%20IS%20NOT%20EMPTY"
```
artem-shelkovnikov, line 1021: Ha nice!

artem-shelkovnikov (approved):

connectors/sources/jira.py
lorenabalan: This has changed since, and it's been split into 2. :)

PR #3711, opened 2025-09-22
Reorg everything

mattnowzari:
# 📂 Connectors project structure re-organization 🗄️ 
### Closes https://github.com/elastic/search-team/issues/10999
This PR re-organizes the `connectors` project into two directories, each with subdirectories representing packages or applications:

`/library`
- `/connectors_sdk`
- `/connectors_protocol`
- `/connectors_sources`
 
`/app`
- `/connectors_agent`
- `/connectors_service`

## Notes
This PR is big and scary, but I promise it's really not that bad 😄 

It is technically in a working state. You can run `connectors_service` via `elastic-ingest` with the following command and get 'something' that runs:

```shell
python elastic-ingest -c ../../config.yml
```

The initial commit is very rough-and-tumble. Where stuff actually ends up living (Ex. is `preflight_check.py` in `connectors_protocol` or in some `app` subdir?), how stuff is named etc _can and should_ be discussed 💬


library/connectors_sdk/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+    "License :: OSI Approved :: MIT License", # Elastic License here
+    "Operating System :: OS Independent",
+]
+requires-python = ">=3.10,<3.12"
+dependencies = [
+    "aiohttp>=3",
+    "aiofiles==23.2.1",
+    "pymongo==4.6.3",
+    "tzcron==1.0.0",
+    "pympler==1.0.1"
```
artem-shelkovnikov, line 20: These should not be the dependencies of the SDK - `utils.py` should be split into utils specific to each package to avoid that.

Honestly not sure even `aiohttp` or `aiofiles` should be part of dependencies of SDK, and definitely not `pymongo`, `tzcron` or `pympler`

library/connectors_sdk/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+    "pytest"
+]
+
+[project.urls]
+Homepage = "https://github.com/elastic/connectors"
+Issues = "https://github.com/elastic/connectors/issues"
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
artem-shelkovnikov, line 35: Why use `harchling`? Lorena used Setuptools and it works pretty well, no need to package stuff with different build systems

library/connectors_sdk/connectors_sdk/utils.py
```diff
@@ -730,7 +729,7 @@ def hash_id(_id):
     # Collision probability: 1.47*10^-29
     # S105 rule considers this code unsafe, but we're not using it for security-related
     # things, only to generate pseudo-ids for documents
-    return hashlib.md5(_id.encode("utf8")).hexdigest()  # noqa S105
+    return # logger.md5(_id.encode("utf8")).hexdigest()  # noqa S105
```
artem-shelkovnikov, line 732: What happened here?

library/connectors_sdk/connectors_sdk/source.py
```diff
@@ -803,7 +804,7 @@ async def create_temp_file(self, file_extension):
                 temp_filename = async_buffer.name
                 yield async_buffer
         finally:
-            await async_buffer.close()
+            await async_buffer.close() # type: ignore
```
artem-shelkovnikov, line 807: What happened here and above with the `type: ignore`?

library/connectors_sdk/connectors_sdk/content_extraction.py
```diff
@@ -10,7 +10,7 @@
 import aiohttp
 from aiohttp.client_exceptions import ClientConnectionError, ServerTimeoutError
 
-from connectors.logger import logger
```
artem-shelkovnikov, line 13: Why rename `logger` to `connector_logger`?

library/connectors_sdk/connectors_sdk/__init__.py
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+import os
+
+with open(os.path.join(os.path.dirname(__file__), "VERSION")) as f:
+    __version__ = f.read().strip()
```
artem-shelkovnikov, line 9: This implies there's a VERSION file in the sdk folder, so we'll need a VERSION file for each package. I am okay with that, but let's then add the files

.gitignore
```diff
@@ -22,3 +22,4 @@ config.yml
 scripts/stack/connectors-config
 
 .python-version
+rebuild.sh
```
artem-shelkovnikov, line 25: What's that?

app/connectors_service/cli/connectors.py
```diff
@@ -0,0 +1,7 @@
+#!/Users/mattnowzari/repos/search_and_transform/connectors/.venv/bin/python3
```
artem-shelkovnikov, line 1: Why add shebangs here and below?

artem-shelkovnikov: Hmmm looking below I believe this should be in .gitignore

app/connectors_service/pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+authors = [
+    { name = "Search Extract and Transform Team", email = "enterprise-search-ingestion-team@elastic.co" }
+]
+description = "Connectors service in a nutshell"
+readme = "README.md"
+requires-python = ">=3.10,<3.12"
+dependencies = [
+    "library/connectors_sdk",
+    "click>=8.1.0",
+    "asyncio"
```
artem-shelkovnikov, line 13: Also needs dependencies on the connectors and the protocol packages

app/connectors_service/pyproject.toml
```diff
@@ -0,0 +1,34 @@
...
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build]
+dev-mode-dirs = ["src"]
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/connectors_service"]
+sources = ["src"]
```
artem-shelkovnikov, line 34: No need for `hatchling`

app/connectors_service/service_cli.py
```diff
@@ -21,18 +21,17 @@
...
 
-from connectors import __version__
-from connectors.build_info import __build_info__
-from connectors.config import load_config
-from connectors.content_extraction import ContentExtraction
-from connectors.logger import logger, set_logger
-from connectors.preflight_check import PreflightCheck
-from connectors.services import get_services
-from connectors.source import get_source_klass, get_source_klasses
+from connectors_sdk.config import load_config
```
artem-shelkovnikov, line 24: config is not specific to SDK, it's an abstraction for the service and service CLI

library/connectors_protocol/connectors_protocol/es/license.py
```diff
@@ -3,7 +3,7 @@
 # or more contributor license agreements. Licensed under the Elastic License 2.0;
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
-from connectors.protocol import JobType
+from connectors_protocol.protocol.connectors import JobType
```
artem-shelkovnikov, line 6: Can this be simplified to just `connectors_protocol`?

PR #3719, opened 2025-09-23
Improve pagination performance in MySQL connector

lorenabalan:
## Closes https://github.com/elastic/connectors/issues/3495
Started off with [Use late row lookup](https://github.com/elastic/connectors/pull/3719/commits/27dbbdc0c245ea10af29c382c3a345b98aa06c02) as per issue suggestion but found out that there's a more efficient way, by getting rid of OFFSET-based pagination (which I think is actually what was suggested by the user originally). Switched to this based on AI recommendations.

* Updated pagination method in MySQL Connector
* Added a table with multi-column primary key (composite primary key) to ensure the query works for that case as well
* Tested locally with 4 tables in the local db and ingestion works
  * one with 120k rows (2 cols)
  * one with 180k rows (10 cols, mixed types)
  * one with 300k rows (12 cols, mixed types, composite primary key)
  * one with +500k rows (20 *chonky* cols, mixed types, including multiple TEXT/LONGTEXT fields and
large JSON objects for more overhead)

Testing revealed some performance gain on the smaller tables (5-12% time reduction), but the real win is on the last case, where we see large speedup factors (over 3x). See top row in the following 2 screenshots:

Screenshot 2025-09-25 at 19 04 45
Screenshot 2025-09-25 at 18 58 21

#### Pre-Review Checklist

## Release Note

Refactored pagination for MySQL connector, improving performance for big tables by factor of at least 3x

### DIFF
--- a/connectors/sources/mysql.py
+++ b/connectors/sources/mysql.py
@@ -282,7 +282,9 @@ class MySQLClient:
 
             # If we got fewer rows than fetch_size, we've reached the end
             if batch_count < self.fetch_size:
-                self._logger.debug(f"Fetched final batch of {batch_count} rows. Fetch size: {self.fetch_size}.")
+                self._logger.debug(
+                    f"Fetched final batch of {batch_count} rows. Fetch size: {self.fetch_size}."
+                )
                 break
 
     async def _get_table_row_count_for_query(self, query):


connectors/sources/mysql.py
```diff
@@ -229,17 +229,54 @@ async def get_last_update_time(self, table):
...
+            async with self.connection.cursor(aiomysql.cursors.SSCursor) as cursor:
+                if params:
+                    await cursor.execute(query, params)
+                else:
+                    await cursor.execute(query)
+
+                async for row in cursor:
+                    yield row
+                    # Update the cursor to this row's primary key values
+                    # Since we SELECT * and order by primary keys first,
```
Apmats, line 278: Wait, is this true? We order the returned rows, but that doesn't say anything about the order of columns within a row. Or am I misunderstanding something?

artem-shelkovnikov: Yeah I feel if PK columns are not sequential, this code will break unexpectedly. Worth to update functional tests too to have PK columns distributed through the column list, not squeezed sequentially?

Apmats:
Super nice! This approach also would yield performance improvements to all other SQL connectors, right? And also in the `yield_rows_for_query` codepath in the same connector.

Since this is a bit tricky to implement too correctly (at least it would be for me 😅 ) , maybe it makes sense to extract into a reusable utility - even a module like `raw_sql_utils.py`? Then we can have a task to do a refactor for the rest of the connectors to use that.

connectors/sources/mysql.py
Apmats: I think we can do something like
`SELECT {', '.join(f'`{pk}`' for pk in primary_keys)}, *`  in the query above to force the ordering.

artem-shelkovnikov:
> This approach also would yield performance improvements to all other SQL connectors, right?

I don't think so. "SQL" connectors are similar only on the surface. In reality some internal engine implementations differ wildly. I think only MySQL suffers from late row lookup, for example.

connectors/sources/mysql.py
artem-shelkovnikov: Would it make sense to also prefix the PK columns in the query?

E.g.
```
SELECT customer_id as pk_1, order_id as pk_2, * FROM orders
```

lorenabalan: Oooops great catch @Apmats @artem-shelkovnikov!!! 😱 🥇Thank you for the suggestions, addressed in [b72e14d](https://github.com/elastic/connectors/pull/3719/commits/b72e14dbe9041f3b671cb843c777e36ead1977e3). 🙌 
Edit: and then fixed in [4a4f50a](https://github.com/elastic/connectors/pull/3719/commits/4a4f50a77dc233a2a4745015157839b53bb4e29e) lol - `SELECT bla, *` seems to have been wishful thinking

artem-shelkovnikov:
> So it seems that this is beneficial in at least some other vendors too? Or maybe I'm benchmarking wrong? At the very least there's yield_rows_for_query in the same connector which when the query has no order by does the same offset-based run through of the table.

That could be the case too. For now let's merge this PR and add a ticket to try to extrapolate this to other SQL connectors when possible.

My judgement about SQL connectors is that before we've had them all under one "generic SQL DB" connector and it just did not work well. We still have weird leftovers in the code from it and every SQL connector is messy. I'd be happy to separate them meaningfully + do benchmarking of the performance the way you've described.

One additional question lies in - can we support such optimisations for the custom queries that the users provide? I can see it as a non-trivial task, but curious if you see how to solve it in a simple manner.

artem-shelkovnikov:
> Yeah I understand it might be the reality that these need to be quite different at times if we aim to squeeze out extra throughput like we do, so I understand why they're separate and likely to remain so. The only reuse I was advocating for was having some raw-sql-stuff (first of them would be this) that IF a data source wants it can invoke functions from.
I don't even know what further separating meaningfully would mean here, as I only had a brief look at the structure of each one - I don't know if it came across like I have a well formed and firm opinion 😁

Fair, having some util methods shared between connectors could be great, even if it's an SQL statement generator :)

> In the trivial case, meaning without an order_by anywhere in the query, I would say - probably 😅 - trivially. We just add our own order by just how we did into select * from X. Likely doable by reusing this code here in that codepath.
>
> With a query that does have an order by somewhere, it seems trickier (like is it top level? if it's inside a CTE we won't care likely and we can do our optimization still?).

 I vaguely remember trying it out but don't remember the results :/

Approach I tried was to just wrap the customer's query in an alias and query it as a table:

```
SELECT custom_query.* FROM
(
    SELECT * FROM some_table
    WHERE some_property = 'banana'
    ORDER BY some_property
) as custom_query
```

I don't remember results of it though, but since I haven't pushed anything it's likely that I failed 😅

connectors/sources/mysql.py
```diff
@@ -229,17 +229,58 @@ async def get_last_update_time(self, table):
...
+                    # (a,b) > (last_a, last_b)
+                    where_clause = f"WHERE {pk_tuple} > {param_tuple}"
+                    params = last_primary_key_values
+
+            query = f"""
+                SELECT * FROM `{self.database}`.`{table}` 
+                {where_clause} 
+                ORDER BY {primary_keys_str} 
+                LIMIT {self.fetch_size}
+            """
```
artem-shelkovnikov, line 261: I believe adding debug logs for generated queries like this can make debugging problems much easier in case results are unexpected.

This doesn't need to be addressed in this PR as it was never there, but if you feel like adding a log line about it and waiting for CI is worth it, go for it!

lorenabalan: Good idea, added more logging in [2beac6e](https://github.com/elastic/connectors/pull/3719/commits/2beac6e8603f40b8ad277b4de8611c962953cef1).

artem-shelkovnikov (approved):

PR #3721, opened 2025-09-23
Restructure Connectors project

mattnowzari:
# 🗄️ Connectors Project Reorg, Part 1 - creating an `/app` folder
## Part of https://github.com/elastic/search-team/issues/10999

This PR aims to get a general restructure of the `connectors` repository established.

### _This PR_ will focus on establishing an `/app` directory ONLY.
- `/app`
  - `/connectors` # contains all of `connectors` code
  - Makefile, tests, scripts etc

## Checklist of things that should be updated/worked on prior to merging this:
- [x] tests
- [x] ftests
- [x] Buildkite scripts (CI-only for now. I imagine releasing scripts will be broken until we determine _how_ we are going to release connectors post-framework-improvements)
- [x] Dependencies are correct
- [x] All Makefile commands work (at least the ones required for CI)
  - `clean, install, run, lint, autoformat, test, notice, ftest, ftrace, docker-build, docker-push` all work with the top-level Makefile

## Background/more context
The ultimate goal is the following structure:

- `/app`
  - `connectors_service`
  - `connectors_agent`
- `/library`
  - `connectors_sdk`
  - `connectors_protocol`
  - `connectors_sources`
  
but we will spread this work out across several PRs


app/connectors/pyproject_cli.toml
```diff
@@ -0,0 +1,121 @@
...
+]
+classifiers = [
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3 :: Only",
+]
+requires-python = ">=3.10,<3.12"
+dependencies = [
```
mattnowzari, line 21: I haven't gone through the dependency list yet, I copied Lorena's `pyproject.toml` as a starting point

artem-shelkovnikov: Maybe it's worth to reduce the size of the PR and not split CLIs against the code, and just first move all the relevant code in `app/connectors`, merge and then start splitting?

Main concern is that there'll be a lot of dealing with CI/tests and such, and if you do the move and split at the same time it'll be quite tough to make everything green.

mattnowzari: That's fair. You are right it'll be less effort to get CI working again if we just contain connectors code under `/app` dir entirely for now - I'll undo splitting CLI off and focus on getting CI working

.gitignore
```diff
@@ -10,7 +10,7 @@ pyvenv.cfg
 __pycache__
 
 # real configurations
-config.yml
+app/config.yml
```
mattnowzari, line 13: This was weird. Apparently, having just `config.yml` wasn't enough to ignore all of the `config.yml` files in our test fixtures, but now with the changes in this PR all of our `config.yml` files were being dropped from the repo.

This will definitely be getting changed once more stuff gets broken out, so I'm not too worried about it

.buildkite/run_functional_test.sh
```diff
@@ -16,7 +16,7 @@ init_python
 
 BASEDIR=$(realpath $(dirname $0))
 ROOT=$(realpath $BASEDIR/../)
-VENV_ROOT=$ROOT/.venv
+VENV_ROOT=$ROOT/app/.venv
```
mattnowzari, line 19: This was changed as the .venv that gets made now lives under the `app/ ` folder

Makefile
```diff
@@ -1,140 +1,34 @@
-.PHONY: install clean lint autoformat test ftest ftrace install run default-config docker-build docker-run docker-push
+app_dir := app/
```
mattnowzari, line 1: This entire Makefile is a wrapper around the inner Makefile under `app/`. It allows us to call those `make` commands without modifying the commands themselves. Again, this probably won't be like this forever, but good to get us going for now

app/Makefile
```diff
@@ -0,0 +1,140 @@
+.PHONY: install clean lint autoformat test ftest ftrace install run default-config docker-build docker-run docker-push
```
mattnowzari, line 1: This is our original Makefile with no changes to it

.ruff.toml
```diff
@@ -1,31 +0,0 @@
-target-version = "py310"
```
artem-shelkovnikov, line 1: Where did the file go?

NOTICE.txt
```diff
@@ -1602,7 +1602,7 @@ Apache License
```
artem-shelkovnikov: This NOTICE.txt should be deleted, right?

mattnowzari: Yup! This should be removed

.ruff.toml
mattnowzari: Ahh! It must've gotten deleted when I moved everything back under `app/` - I'll put it back

Makefile
lorenabalan: nice workaround!

app/.ruff.toml
```diff
@@ -21,6 +21,7 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 [lint.per-file-ignores]
 "connectors/*" = ["S608"]
 "scripts/verify.py" = [ "EM" ]
+"setup.py" = [ "EM" ]
```
lorenabalan, line 24: I don't think this is intentional right? There's no more `setup.py` file

app/pyproject.toml
```diff
@@ -115,7 +115,7 @@ elastic-agent-connectors = "connectors.agent.cli:main"
 test-connectors = "scripts.testing.cli:main"
 
 [tool.setuptools]
-packages = ["connectors", "scripts"]
+packages = ["connectors"]
```
lorenabalan, line 118: I think I added it in order to make either ftests work locally or the CLI work locally. Can you try doing a `make install` or `pip install -e .` and checking the all the CLI from lines 111-115 still work? If they do then great!

app/pyproject.toml
lorenabalan, line 1: I don't see a `MANIFEST.in` next to this? We need it for packaging `VERSION` next to the code for example.

app/.ruff.toml
mattnowzari: No not intentional, when .ruff.toml accidentally got deleted I think I copied the one in `main` branch to restore it, should have copied from `develop` branch instead - will address this in the next PR!

app/pyproject.toml
mattnowzari: I think it got left behind in the top level directory, I'll move it around!

app/pyproject.toml
mattnowzari: I think I had removed it when I was messing around with the .toml file and there was no `scripts` dir under `app/`, but now that there is I'll add it back!

PR #3737, opened 2025-09-29
Extract an sdk package

artem-shelkovnikov:
This PR extracts the minimal amount of code to become a "connectors_sdk".

The SDK is a separate package, and is a minimal thing required to implement new connectors. Theoretically, this package should be sufficient to extract the connectors, but more code might need to be moved there, which is fine.

Some design decisions:

1. I had to move docker files to the root directory, so that the current process of building docker files would not be broken. Later, when packages will be published, dockerfiles might get moved back to the apps they make available.
2. I also had to update the Makefile in the `app/connectors_service` in such a way, that in installs the dev version of the connectors_sdk package using relative path. I don't know if there's a better way for now
3. Some code got moved around in _specific_ way, I've added comments to specific places that I feel need explanation


app/connectors_service/connectors/utils.py
```diff
@@ -6,32 +6,25 @@
...
+from datetime import datetime, timedelta
 from enum import Enum
-from time import strftime
 
 import dateutil.parser as parser
 import pytz
 import tzcron
-from base64io import Base64IO
 from bs4 import BeautifulSoup
+from connectors_sdk.logger import logger
```
lorenabalan, line 26: I don't think we'd want to reuse the logger from a different application, to me it's more natural for each app to have their own so that it's easier to tell where logs are actually coming from. With L170 in `logger.py` unchanged I think they would always show up as `connectors:INFO...` instead of seeing `connectors_sdk:INFO`, `connectors_service:INFO` etc.

app/connectors_service/connectors/cli/connector.py
```diff
@@ -6,6 +6,9 @@
 import asyncio
 from collections import OrderedDict
 
+from connectors_sdk.source import get_source_klass
```
lorenabalan: Looking at where `get_source_klass` and `get_source_klasses` are used, to me they belong to the connectors service app rather to the sdk?

libs/connectors_sdk/pyproject.toml
```diff
@@ -0,0 +1,51 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "elasticsearch-connectors-sdk"
+description = "Elasticsearch Connectors SDK"
+dynamic = ["version"]
+readme = "README.md"
```
lorenabalan, line 9: There's no README.md at this level, would be nice to add one.

Makefile
```diff
@@ -1,34 +1,53 @@
 app_dir := app/connectors_service
+connectors_sdk_dir := libs/connectors_sdk
+
+VERSION=$(shell cat app/connectors_service/connectors/VERSION)
```
artem-shelkovnikov, line 4: This is a little weird - in the end we'll need a VERSION for each package we release, for now I always assume in this Makefile that its final goal is running an instance of `connectors_service` using all the other packages from source in this repo.

app/connectors_service/Makefile
```diff
@@ -76,13 +73,11 @@ autoformat: .venv/bin/python .venv/bin/ruff .venv/bin/elastic-ingest
 	.venv/bin/ruff check scripts --fix
 	.venv/bin/ruff format scripts
 
-test: .venv/bin/pytest
-	.venv/bin/pytest --cov-report term-missing --cov-fail-under 92 --cov-report html --cov=connectors --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests
+test: .venv/bin/python .venv/bin/elastic-ingest .venv/bin/pytest
+	.venv/bin/pytest --cov-report term-missing --cov-fail-under 90 --cov-report html --cov=connectors --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests
```
artem-shelkovnikov, line 77: I dropped coverage lower as moving the code caused the percentage to drop.

I also feel like code coverage for connectors service is not as important and can be lowered to ~70%

app/connectors_service/Makefile
```diff
@@ -47,6 +43,7 @@ install-agent: .venv/bin/elastic-ingest
 
 
 .venv/bin/elastic-ingest: .venv/bin/python
+	.venv/bin/pip install -e ../../libs/connectors_sdk
```
artem-shelkovnikov, line 46: Yup, install the SDK from source using relative path. Is there a better way?

libs/connectors_sdk/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+    "Programming Language :: Python :: 3 :: Only",
+]
+requires-python = ">=3.10,<3.12"
+dependencies = [
+    "aiofiles==23.2.1",
+    "aiohttp==3.12.14",
+    "base64io==1.0.3",
+    "fastjsonschema==2.16.2",
+    "ecs-logging==2.0.0",
+    "elasticsearch"
```
lorenabalan: I didn't see any `elasticsearch` imports in connectors_sdk, we can remove it. Same for `fastjsonschema`, maybe the dependency reduction was WIP.

app/connectors_service/connectors/config.py
```diff
@@ -212,28 +211,3 @@ def _merge_dicts(hsh1, hsh2):
...
-    class Builder:
-        def __init__(self):
-            self.max_file_size = DEFAULT_MAX_FILE_SIZE
-
-        def with_max_file_size(self, max_file_size):
-            self.max_file_size = max_file_size
-            return self
-
-        def build(self):
-            return DataSourceFrameworkConfig(self.max_file_size)
```
artem-shelkovnikov, line 239: I also dropped this code. I think it brings nothing other than confusion - the only goal of this code is to provide a default value.

We wanted this to be more smart and forwards-compatible, but ended up doing nothing, so I took the liberty to clean this up.

app/connectors_service/connectors/protocol/connectors.py
```diff
@@ -579,67 +540,6 @@ def __init__(self, data):
         super().__init__(default)
 
 
-class Features:
```
artem-shelkovnikov, line 582: It felt weird, but it went to SDK in the end. It feels hard to untie SDK from this logic. Might be for good?

app/connectors_service/connectors/sources/box.py
```diff
@@ -16,17 +16,21 @@
 from aiofiles.os import remove
 from aiofiles.tempfile import NamedTemporaryFile
 from aiohttp.client_exceptions import ClientResponseError
+from connectors_sdk.content_extraction import (
+    TIKA_SUPPORTED_FILETYPES,
```
artem-shelkovnikov, line 20: This got moved from `utils` as it feels it belongs to `content_extraction` module better

app/connectors_service/connectors/utils.py
```diff
@@ -6,32 +6,25 @@
 import asyncio
```
artem-shelkovnikov, line 6: I picked only the functions used in other SDK code and moved them to `connectors_sdk.utils`. If I wouldn't do that, the amount of random dependencies in SDK would be pretty high

libs/connectors_sdk/.ruff.toml
```diff
@@ -0,0 +1,32 @@
+target-version = "py310"
```
artem-shelkovnikov, line 1: Just copied this one from the `app/connectors_service`

libs/connectors_sdk/connectors_sdk/VERSION
```diff
@@ -0,0 +1 @@
+9.2.0
```
artem-shelkovnikov, line 1: Fun thing - SDK will have its own version.

Good thing: it won't probably update as often, we rarely added features to the code that's in SDK. We can follow the release schema from the one that's in the Elasticsearch python client: they have `elasticsearch` package + transport package. Transport package doesn't get updated too often and gets released only when needed. So it's possible to have elasticsearch package == 9.2.0, but transport == 9.0.0 an nothing changed in the transport layer.

libs/connectors_sdk/connectors_sdk/filtering/validation.py
```diff
@@ -135,6 +135,41 @@ def to_dict(self):
         }
 
 
+class Filter(dict):
```
artem-shelkovnikov, line 138: This feels weird in SDK, but there's no easy way to move it out without refactoring

libs/connectors_sdk/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+classifiers = [
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3 :: Only",
+]
+requires-python = ">=3.10,<3.12"
+dependencies = [
+    "aiofiles==23.2.1",
```
artem-shelkovnikov, line 22: I think that's the smallest number of libs I could define here

libs/connectors_sdk/pyproject.toml
artem-shelkovnikov: Lol that was fast. I added it cause of random warning from `pytest` cause it failed to run.

There was a warning exception in `pytest.ini` in the root of the repo and it was mentioning a warning in `elasticsearch`:

```
ignore::elasticsearch.exceptions.GeneralAvailabilityWarning
```

SInce SDK did not have this module. pytest was dying. I added the line to continue and just dropped it in a PR

`fastjsonschema` comes here: https://github.com/elastic/connectors/blob/0954997ec24194c67b459e27abd04ed0df0300a5/libs/connectors_sdk/connectors_sdk/filtering/validation.py#L9

libs/connectors_sdk/pyproject.toml
artem-shelkovnikov: Added a stub README.md for now

app/connectors_service/connectors/cli/connector.py
artem-shelkovnikov: Yeah this one's only for the service. I'll move it.

app/connectors_service/connectors/utils.py
artem-shelkovnikov: Yeah I'll see what I can do. Our logging in connectors is a little weird. I agree we'd need to separate logs into who logged what.

app/connectors_service/connectors/config.py
mattnowzari: Thank you for doing this 🙏  When I was figuring out the configurable queue sizes a few iterations back this piece of code really tripped me up lol

app/connectors_service/connectors/utils.py
mattnowzari: ++ on this - logging is a serious consideration we should be thinking about. I agree with Lorena that we'll need a way of making sure a log's origin application is logged.

I'll think about this a bit - I refactored Crawler's log handling to a facade model a few months back and I feel like we may be able to do something similar here, but at the same time I don't want to over-complicate the solution either. This may end up being a more in-depth `logger` class that can accept custom tags and whatnot 🤷

lorenabalan: If it's cleaner/easier to make the split in a separate PR, let's do that.

PR #3747, opened 2025-10-02
[Connectors Framework Improvements ] Break apart connectors source files into folders

mattnowzari:
# 🚧 Connectors Framework Improvements 🏗️ 
## Convert single source `*.py` files into folders with code broken down into more granular files

The goal with this PR is to convert (in-place) our single-file data source implementations into folders that may contain multiple `*.py` files.

Ex:
Original `MySQL` data source file
```
connectors/sources/mysql.py
```

New source (for example, file names not set in stone yet)
```
../connectors/sources/mysql/
    datasource.py (entry point)
    mysql_utils.py
    mysql_validations.py 
```

## Why?
- Easier understanding of the constituent parts of what a connector 'needs' (`BaseDataSource`, `utils` etc)
  - Great for both human coders and LLM tools


app/connectors_service/connectors/config.py
```diff
@@ -114,18 +114,18 @@ def _default_config():
             "azure_blob_storage": "connectors.sources.azure_blob_storage:AzureBlobStorageDataSource",
             "box": "connectors.sources.box:BoxDataSource",
             "confluence": "connectors.sources.confluence:ConfluenceDataSource",
-            "dir": "connectors.sources.directory:DirectoryDataSource",
-            "dropbox": "connectors.sources.dropbox:DropboxDataSource",
+            "dir": "connectors.sources.directory.datasource:DirectoryDataSource",
+            "dropbox": "connectors.sources.dropbox.datasource:DropboxDataSource",
```
mattnowzari: Right now, I'm calling the entry point to a connector `datasource.py`. This will contain the class that extends `BaseDataSource`. It is a point of commonality between all connectors, so it's a natural contender to be split out.

I'm not married to the name, and naming things is hard, so I am soliciting recommendations for alternative file names that can be applied to all connectors 😄

app/connectors_service/connectors/sources/shared/google/google.py
lorenabalan, line 1: Some of the things from here can actually live next to the sources themselves - the GmailClient for example. 

Also, I wonder if you considered the alternative folder structure below and if there was a particular motivator for `shared`.
```
sources
|_ google
    |_ __init__.py
    |_ google_utils.py
    |_ gmail
        |_ datasource.py ...
    |_ google_drive
    |_ google_cloud_storage
```
I like that with `shared` you can work out the commonalities afterwards and no need to refactor modules to group together, but with the above it's nice that we keep all imports from google packages under one module, which might be easier when we split requirements per connector (e.g. when we want users to `pip install connectors["google"]` in order to use any of the google connectors, similar to [this example](https://github.com/kedro-org/kedro-plugins/blob/main/kedro-datasets/pyproject.toml#L36-L38)).

app/connectors_service/connectors/config.py
lorenabalan: `datasource` or `data_source` sounds good to me! 
I also think all of the connectors can have a `client.py`, for the abstraction that actually does the interaction with the 3rd party.

```
blabla
|_ __init__.py
|_ datasource.py
|_ client.py
|_ utils.py
```

app/connectors_service/connectors/sources/gmail/datasource.py
```diff
@@ -1,107 +1,35 @@
...
+from connectors_sdk.utils import iso_utc
 from connectors_sdk.source import BaseDataSource, ConfigurableFieldValueError
-from connectors_sdk.utils import (
-    iso_utc,
-)
-from fastjsonschema import JsonSchemaValueException
 
-from connectors.access_control import ACCESS_CONTROL, es_access_control_query
-from connectors.sources.google import (
+from connectors.sources.shared.google.google import (
```
lorenabalan: Nit, completely optional: We can add an `__init__.py` in shared/google/ and import everything from `google.py`, to avoid the repetitive import paths. Actually looking at it all now I'm a bit surprised it all works without any `__init__.py` files in `sources` subfolders. 🤔 🤔 🤔 
Can we add one everywhere anw please? All Python dirs should have one.

app/connectors_service/connectors/sources/google_cloud_storage/client.py
```diff
@@ -0,0 +1,116 @@
+#
```
artem-shelkovnikov, line 1: Curious why plural `clients`?

app/connectors_service/connectors/sources/mysql/mysql_utils.py
```diff
@@ -0,0 +1,46 @@
+#
```
artem-shelkovnikov: I think it's worth just calling this file `utils.py`?

app/connectors_service/tests/sources/test_directory.py
```diff
@@ -5,7 +5,7 @@
 #
 import pytest
 
-from connectors.sources.directory import DEFAULT_DIR, DirectoryDataSource
+from app.connectors_service.connectors.sources.directory.datasource import DEFAULT_DIR, DirectoryDataSource
```
artem-shelkovnikov: Why do we need full path here and other places in the PR, but not the rest of the code?

app/connectors_service/connectors/sources/shared/google/google.py
mattnowzari: This is a good suggestion for the Google-related connectors. I wasn't sure how much nesting we wanted to do, so if you think having multiple similar connectors being grouped together under a single file is OK then I'm OK with it too.

I don't think this completely removes the need for a `shared` folder, _unless_ we also do like-for-like grouping in other ways (like, all DB connectors get grouped under their own folder we can package `generic_databse` in with them)

app/connectors_service/connectors/sources/gmail/datasource.py
mattnowzari: This was confusing for me as well, and I'm glad you brought it up - I tried it both with and without an `__init__.py` and it worked both ways 🤔

I can certainly add an `__init__.py` in, esp. because one day you should be able to do something like `import connectors.google.gmail` or something

app/connectors_service/tests/sources/test_directory.py
mattnowzari: We don't :) This was a mistake on my end

app/connectors_service/connectors/sources/google_cloud_storage/client.py
mattnowzari: I think this was a typo/brain fart from me lol

mattnowzari:
> How do you feel about keeping the data sources in the sources directory and moving non-datasource code in the folders?

@artem-shelkovnikov I don't feel great about this approach. I feel it creates an unnecessary level of abstraction when our intention is to extract code out into files yet keep them very self-contained for easy understanding. I get the concern over longer import paths, but I think it's a fair tradeoff here.

> Alternatively we can have:
>  ```
> sources
> - gmail
> -- data_source.py
> -- validator.py
> -- __init__py
> ```

I prefer this approach much more, both for the reasons I mentioned above + what Lorena mentioned in her [comment](https://github.com/elastic/connectors/pull/3747#issuecomment-3362331863)

app/connectors_service/connectors/sources/shared/google/google.py
lorenabalan: Yes a `shared` folder would probably still appear, but I'd like that to not depend on any 3rd party package, only on Python builtin libraries - so that it can be safely imported everywhere without a surprise "missing dependency". Even `generic_database.py` depends on `sqlalchemy` atm. Buuut these are future optimisations, we can keep the code as-is and revisit at a later point. :D

app/connectors_service/tests/sources/test_gmail.py
```diff
@@ -15,13 +15,13 @@
...
-from connectors.sources.gmail import (
+from connectors.sources.gmail.datasource import (
     ACCESS_CONTROL,
-    GMailAdvancedRulesValidator,
     GMailDataSource,
     _message_doc,
 )
-from connectors.sources.google import MessageFields, UserFields
+from connectors.sources.gmail.validator import GMailAdvancedRulesValidator
+from connectors.sources.shared.google.google import MessageFields, UserFields
```
mattnowzari: I don't love `from connectors.sources.shared.google.google`, but I'm willing to live with it for the sake of this PR. I am going to address the `shared` folder in a followup PR:

### Proposal (based on Lorena's suggestion earlier)
Instead of

```
shared/
|_ google/
   |_ google.py
``` 

We can group all of our Google connectors under one folder:

```
google/
|_ gmail/
   |_ datasource.py
   |_ ...etc
|_ google_cloud_storage/
   |_ datasource.py
   |_ ...etc
|_ google_drive/
   |_ datasource.py
   |_ ...etc
|_ shared/
   |_ google.py (can rename to smthn else, like utils etc)
```

app/connectors_service/connectors/config.py
artem-shelkovnikov: I believe now this can be returned back to how it was, e.g. `connectors.sources.directory:DirectoryDataSource`?

app/connectors_service/connectors/sources/mongo/validator.py
```diff
@@ -0,0 +1,76 @@
+#
```
artem-shelkovnikov, line 1: Should probably be just renamed to `utils`?

Also I see you normally move validators to separate files, should probably happen here too

app/connectors_service/connectors/sources/mongo/datasource.py
```diff
@@ -216,7 +146,7 @@ async def ping(self):
         with self.get_client() as client:
             await client.admin.command("ping")
 
-    def remove_temp_file(self, temp_file):
+    def remove_temp_file(self, temp_file):  # type: ignore
```
artem-shelkovnikov: What happened with `# type: ignore` here?

app/connectors_service/connectors/config.py
mattnowzari: Not in the form the source folders are currently in, but if I add an import to `__init__.py` we should be able to remove `.datasource` - I will add this!

app/connectors_service/connectors/sources/directory/__init__.py
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+from connectors.sources.directory.datasource import (
+    DirectoryDataSource as DirectoryDataSource,
+)
```
lorenabalan: Why the `as`? We don't name them if we don't rename the thing we're importing. Same for the other `__init__.py`s.

app/connectors_service/tests/sources/test_mongo.py
```diff
@@ -17,7 +17,8 @@
 from connectors_sdk.source import ConfigurableFieldValueError
 from pymongo.errors import OperationFailure
 
-from connectors.sources.mongo import MongoAdvancedRulesValidator, MongoDataSource
+from connectors.sources.mongo.datasource import MongoDataSource
+from connectors.sources.mongo.validator import MongoAdvancedRulesValidator
```
lorenabalan: Let's have the `__init__.py` files import everything from within the package. That way we'll have shorter import paths and we won't have to make these importpath changes to all files. They can stay as they were as it works whether `mongo` is a file or a folder.
```suggestion
from connectors.sources.mongo import MongoAdvancedRulesValidator, MongoDataSource
```
We also reduce the PR footprint a bit

app/connectors_service/connectors/sources/directory/__init__.py
lorenabalan: Nit, optional: in my experience, for init [it's okay](https://github.com/fastapi/fastapi/blob/master/fastapi/security/__init__.py) to use relative imports

```python
from .datasource import DirectoryDataSource
```

app/connectors_service/connectors/sources/gmail/datasource.py
lorenabalan: Sorry I wasn't super clear, I meant once we have the `__init__.py` (which I see you added, thanks!), we add an import there (`from .google import ...`)  so that this line now becomes `from connectors.sources.shared.google import `
I know the folder structure will change for `google` sources so you don't need to address it now, I was just completing my initial thought for reference.

app/connectors_service/connectors/sources/shared/database/__init__.py
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
```
lorenabalan: extra license
```suggestion

```

lorenabalan (commented):
Generally looks good, some comments on imports. 

I think that a pattern is emerging for a structure like

```
sources
|__ some_source_type
        |__ __init__.py
        |__ datasource.py
        |__ client.py
        |__ validator.py / validation.py
        |__ utils.py / common.py
```

I would make the `__init__.py` the entrypoint for most important imports
```python
from .datasource import SomeSourceTypeDs
from .client import SomeSourceTypeClient
from .validation import SomeSourceTypeAdvancedRulesValidator, SomeSourceTypeSyncRulesValidator
```
so that places **outside** of this folder can reference the objects from "one level up" rather than from individual module (file) level. 
```python
from connectors.sources.some_source_type import SomeSourceTypeDs, SomeSourceTypeClient, ...
```

app/connectors_service/connectors/sources/directory/__init__.py
mattnowzari: > Why the as? 

I'll be honest, I am surprised by this bc this is not something I would write. As it turns out, `ruff` changes `from .datasource import DirectoryDataSource` to `from .datasource import DirectoryDataSource as DirectoryDataSource` all on its own?

C0D00DB0-5BC4-4845-8609-2CB31F736551

My knowledge of Python `import` statements is getting a bit thin here. Any thoughts as to why `ruff` does this?

RE relative imports - thank you for pointing this out, I wanted to do relative imports but I wasn't sure if they were OK for __init__.py files 👍

mattnowzari: Ah - apparently this is expected behavior for "unused" imports? https://docs.astral.sh/ruff/rules/unused-import/

lorenabalan: Ah ofc, I forgot. Imports in `__init__.py` are a bit fiddly for linters. As discussed on Slack, let's try adding a `__all__` to bypass these autoformatting checks.

app/connectors_service/connectors/sources/gmail/__init__.py
```diff
@@ -0,0 +1,12 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
+from .datasource import (
+    GMailDataSource,
+    _message_doc,
```
lorenabalan, line 8: We shouldn't export private functions or variables (stuff that starts with `_`). We could either turn it public (if appropriate) or remove it and make the code that needs it import from exact file. Personal preference would be for 2nd option but will leave it up to you!

app/connectors_service/tests/sources/test_gmail.py
```diff
@@ -157,15 +157,16 @@ class TestGMailDataSource:
     @pytest_asyncio.fixture
     async def patch_gmail_client(self):
         with patch(
-            "connectors.sources.gmail.GMailClient", return_value=AsyncMock()
+            "connectors.sources.gmail.datasource.GMailClient", return_value=AsyncMock()
```
lorenabalan, line 160: this can be back to what it was right?
```suggestion
            "connectors.sources.gmail.GMailClient", return_value=AsyncMock()
```

app/connectors_service/tests/sources/test_google.py
```diff
@@ -104,14 +104,16 @@ class TestGoogleServiceAccountClient:
     @pytest_asyncio.fixture(autouse=True)
     async def patch_service_account_creds(self):
         with patch(
-            "connectors.sources.google.ServiceAccountCreds", return_value=Mock()
+            "connectors.sources.shared.google.google.ServiceAccountCreds",
```
lorenabalan, line 107:
```suggestion
            "connectors.sources.shared.google.ServiceAccountCreds",
```

app/connectors_service/tests/sources/test_gmail.py
```diff
@@ -157,15 +157,16 @@ class TestGMailDataSource:
...
         ) as mock:
             client = mock.return_value
             yield client
 
     @pytest_asyncio.fixture
     async def patch_google_directory_client(self):
         with patch(
-            "connectors.sources.gmail.GoogleDirectoryClient", return_value=AsyncMock()
+            "connectors.sources.gmail.datasource.GoogleDirectoryClient",
```
lorenabalan, line 168:
```suggestion
            "connectors.sources.gmail.GoogleDirectoryClient",
```

app/connectors_service/tests/sources/test_google.py
```diff
@@ -199,7 +201,7 @@ class TestGoogleDirectoryClient:
     @pytest_asyncio.fixture
     async def patch_google_service_account_client(self):
         with patch(
-            "connectors.sources.google.GoogleServiceAccountClient",
+            "connectors.sources.shared.google.google.GoogleServiceAccountClient",
```
lorenabalan, line 204:
```suggestion
            "connectors.sources.shared.google.GoogleServiceAccountClient",
```

app/connectors_service/tests/sources/test_google.py
```diff
@@ -262,7 +264,7 @@ class TestGMailClient:
...
```
lorenabalan, line 267:
```suggestion
            "connectors.sources.shared.google.GoogleServiceAccountClient",
```

app/connectors_service/tests/sources/test_google.py
```diff
@@ -104,14 +104,16 @@ class TestGoogleServiceAccountClient:
...
+            "connectors.sources.shared.google.google.ServiceAccountCreds",
+            return_value=Mock(),
         ) as class_mock:
             yield class_mock
 
     @pytest_asyncio.fixture
     async def patch_aiogoogle(self):
         with patch(
-            "connectors.sources.google.Aiogoogle", return_value=MagicMock()
+            "connectors.sources.shared.google.google.Aiogoogle",
```
lorenabalan, line 115:
```suggestion
            "connectors.sources.shared.google.Aiogoogle",
```

lorenabalan (approved):
Looks good! Final set of import-related comments. 😂 Great work persevering on this!

app/connectors_service/connectors/sources/gmail/__init__.py
mattnowzari: Eek! This is such a code smell I can't believe I glossed over this :( 

It's tricky bc one of the tests requires access to it and I wasn't trying to refactor any testing or functionality if possible. I'll address this in a follow-up PR to make reviewing those particular diffs easier

app/connectors_service/tests/sources/test_gmail.py
mattnowzari: While working on the next round of commits I tried removing `.datasource`and it actually didn't work :( I'm not quite sure why truth be told.

I'm going to see if we can fix/workaround this, and that can be a followup PR if I do get to the bottom of it

lorenabalan: That'll probably be because of how it's imported in the source file that we're testing. If it's imported as `.datasource.` then the test should be the same. Mocking imports in Python is fiddly. It's not super important for now don't spend too much time on it if you're still stuck, I can also have a look tomorrow my time. :)

PR #3748, opened 2025-10-03
Add buildkite pipeline for publishing to PyPI

lorenabalan:
## Related to https://github.com/elastic/connectors-py/issues/10999
* Added new _manual_ pipeline for publishing to PyPI: [.buildkite/pypi-publish-pipeline.yml](https://github.com/elastic/connectors/pull/3748/files#diff-028ba08c5e2e74e83911f11edfc92beb3dd8ebe77b6136d593f8b88bc047f8c0)
* Added new step in main pipeline (that runs on PRs and on daily schedule) to build the packages' [binary & source distributions](https://packaging.python.org/en/latest/specifications/section-distribution-formats/), then `twine check` them; for `connectors_service` it will also install it and test the CLI entrypoints it exposes (e.g. `connectors --help`)
* Changes to `pyproject.toml`
    * Incorporated `pytest`, `pyright`, and `ruff` config into `pyproject.toml` and removed individual files
    * Updated `license` in `pyproject.toml` to avoid deprecation warnings (removal in 2026) - see warnings [here](https://buildkite.com/elastic/connectors/builds/19996#0199aa3a-2539-4dea-81cb-bd1f0a401eb0)
    * Updated `packages` in `pyproject.toml` to avoid "ambiguous packages" warning - see warning [here](https://buildkite.com/elastic/connectors/builds/19996#0199aa3a-2539-4dea-81cb-bd1f0a401eb0)

~❓ I don't see `diff.sh` used anywhere.. can I delete it or am I missing sth?~ nevermind, I was looking for `diff.sh` instead of just `diff`

#### ✍️ Process moving forward
Daily per PR: 
- Validate that packages are built correctly 
- "Publish DRA" part of `pipeline.yml` will contain:
    - building docker image by installing package locally by **building and installing from distribution**
    - build a `.tar.gz` and `.whl` of our packages 

Release day:
- same process as today, a Docker image DRA is chosen as BC and then released
- _additionally,_ trigger manual pipeline to publish to PyPI from appropriate commit; ⚠️ as agreed on 21/10/2025, this will only publish to Test PyPI. 

TODO:
- [x] Update `catalog-info.yml`
- [x] Publish .tar.gz and .whl as DRA
- [x] Ensure Docker images are built with installing code from .whl

#### Pre-Review Checklist

### DIFF
--- a/.buildkite/publish/dra/init_dra_publishing.sh
+++ b/.buildkite/publish/dra/init_dra_publishing.sh
@@ -25,17 +25,18 @@ cd -
 # Download previous step artifacts
 buildkite-agent artifact download '.artifacts/*.tar.gz*' $RELEASE_DIR/dist/ --step build_docker_image_amd64
 buildkite-agent artifact download '.artifacts/*.tar.gz*' $RELEASE_DIR/dist/ --step build_docker_image_arm64
-cp $RELEASE_DIR/dist/.artifacts/* $DRA_ARTIFACTS_DIR
+buildkite-agent artifact download 'app/connectors_service/dist/*.whl' $RELEASE_DIR/ --step test_packages
+buildkite-agent artifact download 'app/connectors_service/dist/*.tar.gz' $RELEASE_DIR/ --step test_packages
+buildkite-agent artifact download 'libs/connectors_sdk/dist/*.whl' $RELEASE_DIR/ --step test_packages
+buildkite-agent artifact download 'libs/connectors_sdk/dist/*.tar.gz' $RELEASE_DIR/ --step test_packages
 
-# Download Python package artifacts
-buildkite-agent artifact download 'app/connectors_service/dist/*.whl' $RELEASE_DIR/ --step 'test_packages'
-buildkite-agent artifact download 'app/connectors_service/dist/*.tar.gz' $RELEASE_DIR/ --step 'test_packages'
-buildkite-agent artifact download 'libs/connectors_sdk/dist/*.whl' $RELEASE_DIR/ --step 'test_packages'
-buildkite-agent artifact download 'libs/connectors_sdk/dist/*.tar.gz' $RELEASE_DIR/ --step 'test_packages'
+
+# Copy previous step artifacts to DRA dir
+cp $RELEASE_DIR/dist/.artifacts/* $DRA_ARTIFACTS_DIR
 cp $RELEASE_DIR/app/connectors_service/dist/* $DRA_ARTIFACTS_DIR
 cp $RELEASE_DIR/libs/connectors_sdk/dist/* $DRA_ARTIFACTS_DIR
 
-# Rename to match DRA expectations (<name>-<version>-<classifier>-<os>-<arch>)
+# Rename docker images to match DRA expectations (<name>-<version>-<classifier>-<os>-<arch>)
 cd $DRA_ARTIFACTS_DIR
 mv $DOCKER_ARTIFACT_KEY-$VERSION-amd64.tar.gz $PROJECT_NAME-$VERSION-docker-image-linux-amd64.tar.gz
 mv $DOCKER_ARTIFACT_KEY-$VERSION-arm64.tar.gz $PROJECT_NAME-$VERSION-docker-image-linux-arm64.tar.gz
@@ -147,13 +148,13 @@ if [[ "${PUBLISH_SNAPSHOT:-}" == "true" ]]; then
   generateDependencyReport $DEPENDENCIES_REPORTS_DIR/$dependencyReportName
 
   echo "-------- Publishing SNAPSHOT DRA Artifacts"
-  cp $DRA_ARTIFACTS_DIR/connectors-${VERSION}.zip $DRA_ARTIFACTS_DIR/connectors-${VERSION}-SNAPSHOT.zip
 
+  # Make *-SNAPSHOT* copies
+  cp $DRA_ARTIFACTS_DIR/connectors-${VERSION}.zip $DRA_ARTIFACTS_DIR/connectors-${VERSION}-SNAPSHOT.zip
   cp $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION.whl $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION-SNAPSHOT.whl
   cp $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION.tar.gz $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION-SNAPSHOT.tar.gz
   cp $DRA_ARTIFACTS_DIR/elasticsearch_connectors_sdk-$VERSION.whl $DRA_ARTIFACTS_DIR/elasticsearch_connectors_sdk-$VERSION-SNAPSHOT.whl
   cp $DRA_ARTIFACTS_DIR/elasticsearch_connectors_sdk-$VERSION.tar.gz $DRA_ARTIFACTS_DIR/elasticsearch_connectors_sdk-$VERSION-SNAPSHOT.tar.gz
-
   cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-amd64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-SNAPSHOT-docker-image-linux-amd64.tar.gz
   cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-arm64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-SNAPSHOT-docker-image-linux-arm64.tar.gz
 
@@ -171,12 +172,11 @@ fi
 if [[ "${PUBLISH_STAGING:-}" == "true" ]]; then
   if [ -n "${VERSION_QUALIFIER:-}" ]; then
     dependencyReportName="dependencies-${VERSION}-${VERSION_QUALIFIER}.csv";
-    zip_artifact_name="connectors-${VERSION}-${VERSION_QUALIFIER}.zip"
-    cp $DRA_ARTIFACTS_DIR/connectors-${VERSION}.zip $DRA_ARTIFACTS_DIR/${zip_artifact_name}
 
+    # Make *-$VERSION_QUALIFIER* copies
+    cp $DRA_ARTIFACTS_DIR/connectors-${VERSION}.zip $DRA_ARTIFACTS_DIR/$PROJECT_NAME-${VERSION}-${VERSION_QUALIFIER}.zip
     cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-amd64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-$VERSION_QUALIFIER-docker-image-linux-amd64.tar.gz
     cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-arm64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-$VERSION_QUALIFIER-docker-image-linux-arm64.tar.gz
-
     cp $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION.whl $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION-$VERSION_QUALIFIER.whl
     cp $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION.tar.gz $DRA_ARTIFACTS_DIR/elasticsearch_connectors-$VERSION-$VERSION_QUALIFIER.tar.gz
     cp $DRA_ARTIFACTS_DIR/elasticsearch_connectors_sdk-$VERSION.whl $DRA_ARTIFACTS_DIR/elasticsearch_connectors_sdk-$VERSION-$VERSION_QUALIFIER.whl


.buildkite/pipeline.yml
```diff
@@ -580,17 +580,45 @@ steps:
               artifact_paths:
                 - "perf8-report-*/**/*"
 
+  - group: ":building_construction: Build Python Packages"
```
lorenabalan: How it shows up in the UI:
Screenshot 2025-10-03 at 12 51 32
...
Screenshot 2025-10-03 at 12 51 37

artem-shelkovnikov:
> ❓ I don't see diff.sh used anywhere.. can I delete it or am I missing sth?

diff.sh is used in here: https://github.com/elastic/connectors/blob/main/.buildkite/pipeline.yml#L84

> ❓ 🧠 What happens if we have a bug in our package? Requesting an emergency stack release in order to release a new version on PyPI seems excessive, so does yanking every buggy package from PyPI. I'm more in favour of having different versions conventions but that will get easily confusing.

We had a {major}.{minor}.{patch}.{release} available at some point, but since we release with stack it could be hard to organise well. We might need more strict QA process for it? I'd open a discussion with our team and potentially the clients team - they can give a good hint about it.

> Re: releasing
For now the process for release is "promoting" the image. So by BC1 we already need to have an image with a real package installed.

Additionally, if we feel like it's too much we can keep the current process for docker images around for now (same as we do now with `make install` and such) and as we polish the process we'll switch to it.

.buildkite/pypi-publish-pipeline.yml
```diff
@@ -0,0 +1,114 @@
...
+              - "${SERVICE_PACKAGE_PATH}"
+              - "${SDK_PACKAGE_PATH}"
+        commands:
+          - "cd {{matrix.package_path}}"
+          - "python{{matrix.python_version}} -m pip install --upgrade build twine"
+          - "python{{matrix.python_version}} -m build"
+          - "ls -lah dist/"
+          - "python{{matrix.python_version}} -m twine check dist/*"
+        artifact_paths:
+          - "{{matrix.package_path}}/dist/*"
```
seanstory: I suggest moving these things to the Makefile. While the regular experience should be to trigger releases with buildkite, I think we'll want to be able to manually do it from our dev envs, or to easily be able to troubleshoot/dry-run locally. Keeping these types of "scripts" in a place where they can be reused will make that cleaner.

lorenabalan: Definitely, agree! I could swear I had done this already but I can't find it in my stash or commit history. :|:|:| It might've disappeared in a bad merge or I will start to seriously worry about my memory. 😅

lorenabalan: Aaah nevermind, I had only made that change to main pipeline, not to the release one. Updated [here](https://github.com/elastic/connectors/pull/3748/commits/d7f7a96d4a9492bad876afe2fb83ff06a8da4fd9)

libs/connectors_sdk/pyproject.toml
mattnowzari, line 1: Will `connectors_sdk` pyproject.toml also consolidate `.ruff.toml`, `pytest.ini` and `pyrightconfig.json` in this PR or a follow-up one?

lorenabalan: I'll do it in a follow-up, I was mainly focusing on the `connectors_service` package as part of this PR.

.buildkite/pipeline.yml
```diff
@@ -597,6 +597,29 @@ steps:
...
+    key: "build_package"
+    steps:
+      - label: ":python: Build Python {{matrix.python_version}} {{matrix.package_path}} package"
+        key: build_python_package
+        <<: *test-agents
+        matrix:
+          setup:
+            python_version:
+              - "3.10"
+              - "3.11"
```
artem-shelkovnikov, line 637: Should we have both, or only 1 version?

.buildkite/pipeline.yml
```diff
@@ -597,6 +597,29 @@ steps:
               artifact_paths:
                 - "perf8-report-*/**/*"
 
+  - group: ":building_construction: Build Python Packages"
```
artem-shelkovnikov: Should we do the build after all steps succeed instead of running in parallel? How long does the step take normally?

.buildkite/pipeline.yml
```diff
@@ -597,6 +597,29 @@ steps:
...
+            python_version:
+              - "3.10"
+              - "3.11"
+            package_path:
+              - "app/connectors_service"
+              - "libs/connectors_sdk"
+        env:
+          PYTHON_VERSION: "{{ matrix.python_version }}"
+          PACKAGE_PATH: "{{ matrix.package_path }}"
+        command: ".buildkite/test_python_packages.sh"
```
artem-shelkovnikov, line 644: Should the step be called test packages then, since it doesn't build?

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -26,6 +26,18 @@ buildkite-agent artifact download '.artifacts/*.tar.gz*' $RELEASE_DIR/dist/ --st
...
+buildkite-agent artifact download 'app/connectors_service/dist/*.whl' $RELEASE_DIR/ --step 'build_python_package'
+buildkite-agent artifact download 'app/connectors_service/dist/*.tar.gz' $RELEASE_DIR/ --step 'build_python_package'
+buildkite-agent artifact download 'libs/connectors_sdk/dist/*.whl' $RELEASE_DIR/ --step 'build_python_package'
+buildkite-agent artifact download 'libs/connectors_sdk/dist/*.tar.gz' $RELEASE_DIR/ --step 'build_python_package'
+
+# Copy Python packages to DRA artifacts directory
+cp $RELEASE_DIR/app/connectors_service/dist/*.whl $DRA_ARTIFACTS_DIR/connectors-service-$VERSION.whl
+cp $RELEASE_DIR/app/connectors_service/dist/*.tar.gz $DRA_ARTIFACTS_DIR/connectors-service-$VERSION.tar.gz
+cp $RELEASE_DIR/libs/connectors_sdk/dist/*.whl $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION.whl
+cp $RELEASE_DIR/libs/connectors_sdk/dist/*.tar.gz $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION.tar.gz
```
artem-shelkovnikov: So this adds the wheels and tarballs to the artefacts that we release, and then we'll change this in the other place that uses them, right?

.buildkite/publish_to_pypi.sh
```diff
@@ -0,0 +1,24 @@
...
+cd "$PACKAGE_PATH"
+python -m pip install --upgrade build twine
+python -m build
+
+export TWINE_USERNAME="__token__"
+
+# upload to test or real PyPI based on TEST_PYPI=1 env var or arg
+if [[ "${1:-}" == "TEST_PYPI=1" ]] || [[ "${TEST_PYPI:-}" =~ ^(1|TRUE|true)$ ]]; then
+  TWINE_PASSWORD=$(vault read -field publishing-api-key secret/ent-search-team/test-pypi-ent-search-dev)
+  export TWINE_PASSWORD
```
artem-shelkovnikov: Curious - why multiline?

.buildkite/pypi-publish-pipeline.yml
```diff
@@ -0,0 +1,92 @@
...
+        key: build_python_package
+        agents:
+          provider: "gcp"
+          machineType: "n1-standard-8"
+          useVault: true
+          image: family/enterprise-search-ubuntu-2204-connectors-py
+        matrix:
+          setup:
+            python_version:
+              - "3.10"
```
artem-shelkovnikov, line 30: Same as above - do we need both versions?

.buildkite/test_python_packages.sh
```diff
@@ -0,0 +1,29 @@
...
+  # Install the connectors_sdk package first
+  LIB_PATH="libs/connectors_sdk"
+  python -m build "$LIB_PATH"
+  python -m pip install "$LIB_PATH"/dist/*.whl
+
+  python -m pip install "$PACKAGE_PATH"/dist/*.whl
+  connectors --help
+  elastic-ingest --help
+#  elastic-agent-connectors --help
+  test-connectors --help
```
artem-shelkovnikov, line 28: That tests that the scripts are present in the local python, right?

Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY --chown=nonroot:nonroot . /app
 
 USER nonroot
 WORKDIR /app
-RUN make clean install
+RUN make clean install-package
```
artem-shelkovnikov, line 11: Would that make the docker command not work any more the way we suggest it?

See https://www.elastic.co/docs/reference/search-connectors/es-connectors-run-from-docker:

```
docker run \
-v "</absolute/path/to>/connectors-config:/config" \
--rm \
--tty -i \
--network host \
docker.elastic.co/integrations/elastic-connectors:9.1.5 \
/app/bin/elastic-ingest \
-c /config/config.yml
```

.buildkite/publish/dra/init_dra_publishing.sh
lorenabalan: I'm not sure what "the other place" is. 😅 
What this PR does is
- on push to target branch (main or version branch), build and test wheels, then publish them as DRAs (my limited understanding was that simply putting them in the right folder was enough??)
- on push to target branch, build Docker image (by building and installing from wheel; it builds again, doesn't download the one from DRA) and publish as DRA

.buildkite/pypi-publish-pipeline.yml
lorenabalan: I think for building it's fine to test all versions we support, to make sure the CLI works in both (`connectors --help` etc).

.buildkite/test_python_packages.sh
lorenabalan: It tests that the CLIs work after installing the connectors_service package. I think we're saying the same thing but with different words.
`elastic-agent-connectors` hangs for me on my local env which is why I commented it out, but would be good to validate it still works for others.

.buildkite/publish_to_pypi.sh
lorenabalan: Ah I had this before in the .yml file to avoid printing out the value of TWINE_PASSWORD, but I guess it's safer here with no `-x` at the top of the file?

Dockerfile
lorenabalan: Mmm I don't remember looking into it, my expectation was that it should be the same. Otherwise wouldn't the ftests fail?

artem-shelkovnikov: Nope, ftests run the service from source, not docker, quite a miss 🤷

.buildkite/test_python_packages.sh
artem-shelkovnikov: Yup `elastic-agent-connectors` is a GRPC service that is untestable like this unfortunately

.buildkite/publish_to_pypi.sh
artem-shelkovnikov: Honestly I'm not super familiar with bash and this, I trust your judgement

Dockerfile
lorenabalan: Thanks for catching this, I think it's actually been broken on `develop`! I've addressed in [ec4d050](https://github.com/elastic/connectors/pull/3748/commits/ec4d0502d335cbbc3ebc42cf11c237ff5cd1013d) for now but this comes back to what I mentioned the other day that it's weird we have different .venv per package and the top-evel one doesn't have everything.

.buildkite/pipeline.yml
lorenabalan: Good point, done!

.buildkite/pipeline.yml
lorenabalan: I think it's alright to keep it for both here, as mentioned in https://github.com/elastic/connectors/pull/3748#discussion_r2455003788

.buildkite/publish_to_pypi.sh
lorenabalan: Claude seems to agree 😅

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: https://github.com/elastic/infra/blob/master/cd/release/release-manager/project-configs/master/connectors.gradle is where our DRA artifacts are defined to be expected. We'll need to update that.

Then, I think @artem-shelkovnikov might be referencing is Elastic Agent's consumption of our current zip artifact. It's used [here](https://github.com/elastic/elastic-agent/blob/9c001b07d322907da9070e4cb0efdd5752aa1549/dev-tools/packaging/packages.yml#L99-L100) (see [original PR](https://github.com/elastic/elastic-agent/pull/5349)). However, I think you and I agree that this doesn't need to be changed right now.

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -139,8 +151,14 @@ if [[ "${PUBLISH_SNAPSHOT:-}" == "true" ]]; then
...
   echo "-------- Publishing SNAPSHOT DRA Artifacts"
   cp $RELEASE_DIR/dist/elasticsearch_connectors-${VERSION}.zip $DRA_ARTIFACTS_DIR/connectors-${VERSION}-SNAPSHOT.zip
+
   cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-amd64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-SNAPSHOT-docker-image-linux-amd64.tar.gz
   cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-arm64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-SNAPSHOT-docker-image-linux-arm64.tar.gz
+
+  cp $DRA_ARTIFACTS_DIR/connectors-service-$VERSION.whl $DRA_ARTIFACTS_DIR/connectors-service-$VERSION-SNAPSHOT.whl
+  cp $DRA_ARTIFACTS_DIR/connectors-service-$VERSION.tar.gz $DRA_ARTIFACTS_DIR/connectors-service-$VERSION-SNAPSHOT.tar.gz
+  cp $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION.whl $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION-SNAPSHOT.whl
+  cp $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION.tar.gz $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION-SNAPSHOT.tar.gz
```
seanstory: This is petty, but can we put these lines up a few lines, wher we're doing the same thing for zip files?
In fact, I notice the zip is coming from `$RELEASE_DIR/dist/`, and these are in `$DRA_ARTIFACTS_DIR/`, is it worth making those consistent, so that future updates are easier to grok and apply?

.buildkite/publish_to_pypi.sh
artem-shelkovnikov: I Don't trust this Claude guy a single bit :D

.buildkite/publish/dra/init_dra_publishing.sh
```diff
@@ -155,9 +173,16 @@ fi
...
+
     zip_artifact_name="connectors-${VERSION}-${VERSION_QUALIFIER}.zip"
+
     cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-amd64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-$VERSION_QUALIFIER-docker-image-linux-amd64.tar.gz
     cp $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-docker-image-linux-arm64.tar.gz $DRA_ARTIFACTS_DIR/$PROJECT_NAME-$VERSION-$VERSION_QUALIFIER-docker-image-linux-arm64.tar.gz
+
+    cp $DRA_ARTIFACTS_DIR/connectors-service-$VERSION.whl $DRA_ARTIFACTS_DIR/connectors-service-$VERSION-$VERSION_QUALIFIER.whl
+    cp $DRA_ARTIFACTS_DIR/connectors-service-$VERSION.tar.gz $DRA_ARTIFACTS_DIR/connectors-service-$VERSION-$VERSION_QUALIFIER.tar.gz
+    cp $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION.whl $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION-$VERSION_QUALIFIER.whl
+    cp $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION.tar.gz $DRA_ARTIFACTS_DIR/connectors-sdk-$VERSION-$VERSION_QUALIFIER.tar.gz
```
seanstory: again, let's handle this `cp` business in the same place we're doing it for the zip file (below, this time)

artem-shelkovnikov (approved):
Let's gooooo 🚀 🌒

catalog-info.yaml
```diff
@@ -245,3 +245,33 @@ spec:
...
+      metadata:
+          name: "connectors-publish-to-pypi"
+      spec:
+        repository: "elastic/connectors"
+        pipeline_file: ".buildkite/publish-to-pypi.yml"
+        provider_settings:
+          trigger_mode: "none"
+        teams:
+          search-extract-and-transform: {}
+          search-productivity-team: {}
```
seanstory, line 275: This is what our other jobs are, but this has bitten us recently - i suggest we add search-eng here too.

.buildkite/pypi-publish-pipeline.yml
```diff
@@ -0,0 +1,92 @@
...
+          setup:
+            package_path:
+              - "${SERVICE_PACKAGE_PATH}"
+              - "${SDK_PACKAGE_PATH}"
+        env:
+          PYTHON_VERSION: "3.11"
+          PACKAGE_PATH: "{{ matrix.package_path }}"
+        commands: "TEST_PYPI=1 ./buildkite/publish_to_pypi.sh"
+
+# FOR WHEN WE'RE READY TO PUBLISH TO PRODUCTION PYPI
```
seanstory, line 64: Any reason to not yolo it? I assume that we can publish with some sort of "experimental" status, and can probably manually remove publishes if we need? But the code is public already, so I don't think there's a security reason to avoid pushing to early.

.buildkite/publish_to_pypi.sh
seanstory: Google tells me that `twine` can take piped passwords, like:
```
echo "your_password" | twine upload -u <username> dist/*
```

this means it's probably safer for us to do something like:

```
vault read -field publishing-api-key secret/ent-search-team/test-pypi-ent-search-dev | python -m twine upload --repository testpypi dist/*
```
because then adding a `-x` doesn't risk us leaking creds to the logs. We've used this pattern before, like see: https://github.com/elastic/connectors/blob/f95251ada485e138108c3d6a295c527f3e7d981c/.buildkite/publish/push-docker.sh#L29-L30

(alternatively/additionally, we should explicitly set a `+x`, see: https://github.com/elastic/connectors/blob/f95251ada485e138108c3d6a295c527f3e7d981c/.buildkite/publish_docker.sh#L20)

.buildkite/publish_to_pypi.sh
```diff
@@ -0,0 +1,24 @@
...
+init_python
+cd "$PACKAGE_PATH"
+python -m pip install --upgrade build twine
+python -m build
+
+export TWINE_USERNAME="__token__"
+
+# upload to test or real PyPI based on TEST_PYPI=1 env var or arg
+if [[ "${1:-}" == "TEST_PYPI=1" ]] || [[ "${TEST_PYPI:-}" =~ ^(1|TRUE|true)$ ]]; then
+  TWINE_PASSWORD=$(vault read -field publishing-api-key secret/ent-search-team/test-pypi-ent-search-dev)
```
seanstory: I'm not confident that buildkite can read secrets from `secret/ent-search-team/`. I think they need to be in `secret/ci` or similar. Is this something you've worked with the dev-prod team on?

.buildkite/test_python_packages.sh
```diff
@@ -0,0 +1,29 @@
...
+
+init_python
+
+python -m pip install --upgrade build twine
+python -m build "$PACKAGE_PATH"
+ls -lah "$PACKAGE_PATH/dist/"
+python -m twine check "$PACKAGE_PATH/dist/*"
+
+# If this is the connectors_service package, test the installation and CLI
+if [[ "$PACKAGE_PATH" == *app/connectors_service* ]]; then
```
seanstory, line 16: Are we missing tests for the SDK package option in the matrix?

app/connectors_service/NOTICE.txt
```diff
@@ -3546,13 +3546,13 @@ Apache Software License
 
 
 elasticsearch-connectors
-9.2.0
-Apache Software License
+9.3.0
+UNKNOWN
```
seanstory, line 3550: This is a red flag, and probably means that the built packages aren't properly advertizing their licenses. This should be Elastic License v2: https://github.com/elastic/connectors/blob/main/LICENSE

app/connectors_service/pyproject.toml
```diff
@@ -114,8 +114,76 @@ connectors = "connectors.connectors_cli:main"
...
+    # aiogoogle inherits on top of RetryableAioHttpSession, which is not recommended by aiohttp
+    "ignore:Inheritance class RetryableAiohttpSession from ClientSession is discouraged:DeprecationWarning",
+    # pytest may generate its own warnings in some situations, such as improper usage or deprecated features.
+    "ignore::pytest.PytestUnraisableExceptionWarning",
+    # this error comes from 8.16 Elasticsearch client: we use some non-GA features in our CI code
+    "ignore::elasticsearch.exceptions.GeneralAvailabilityWarning",
+]
+
+[tool.pyright]
+pythonVersion = "3.10"
```
seanstory: n00b question - why 3.10? shouldn't we be targeting 3.11, since that's the highest we support?

app/connectors_service/pyproject.toml
```diff
@@ -114,8 +114,76 @@ connectors = "connectors.connectors_cli:main"
...
+[tool.ruff.lint]
+select = ["A", "ASYNC", "I", "E", "F", "B", "C4", "T10", "T20", "EM", "ISC", "S", "CPY001"]
+ignore = ["E501", "ISC001"]
+preview = true
+
+# Allow autofix for all enabled rules (when `--fix`) is provided.
+fixable = ["A", "B", "C", "C4", "D", "E", "F", "G", "I", "N", "Q", "S", "T", "W", "ANN", "ARG", "BLE", "COM", "DJ", "DTZ", "EM", "ERA", "EXE", "FBT", "ICN", "INP", "ISC", "NPY", "PD", "PGH", "PIE", "PL", "PT", "PTH", "PYI", "RET", "RSE", "RUF", "SIM", "SLF", "TCH", "TID", "TRY", "UP", "YTT", "T10", "T20"]
+unfixable = []
+
+exclude = [
```
seanstory: Do we need to exclude `.buildkite` too?

.buildkite/test_python_packages.sh
lorenabalan: We don't have any CLI exposed as part of the sdk package, so there's no CLI test for that.

app/connectors_service/pyproject.toml
lorenabalan: ah this was already in pyright json config, probably from a time before we had 3.11? can bump it to newer version though

.buildkite/publish_to_pypi.sh
lorenabalan: Oh I didn't know about that thank you! - maybe that's why [these instructions](https://github.com/elastic/python-elastic-agent-client/blob/main/PUBLISHING.md#getting-credentials) only had manual steps. You're probably right, our other secrets reference `secret/ci/elastic-connectors/...`. What's the name of the owning team / channel? I'll follow up with them.

seanstory: start with asking in `#search-productivity` (Klim's team)

.buildkite/test_python_packages.sh
seanstory: Maybe we could just try running a dumb python script that imports the sdk and prints some classes, or something?

app/connectors_service/pyproject.toml
lorenabalan: Updated in [51d3e46](https://github.com/elastic/connectors/pull/3748/commits/51d3e46e129eae257ea21d57959ee8d7fd5ec00d)

app/connectors_service/pyproject.toml
lorenabalan: I don't think this is needed at all actually, removed the whole block in [51d3e46](https://github.com/elastic/connectors/pull/3748/commits/51d3e46e129eae257ea21d57959ee8d7fd5ec00d)
It shouldn't look at folders that are a) outside the scope of the package and b) don't contain any python anw

app/connectors_service/NOTICE.txt
lorenabalan: Thanks for raising. I'm not sure why this happens, as I specified explicitly `Elastic-2.0` license in pyproject.toml, which is valid as per https://spdx.org/licenses/. 
I've added license file in [6a2d1b2](https://github.com/elastic/connectors/pull/3748/commits/6a2d1b2dcddab4c82929e7b320ece563a6d1eb3d) and that seems to at least include more info.

catalog-info.yaml
lorenabalan: Addressed in [4d185fe](https://github.com/elastic/connectors/pull/3748/commits/4d185fed137d62fecb000346f3a32135d4754aa6)

.buildkite/publish/dra/init_dra_publishing.sh
lorenabalan: I changed the code a bit so that we only work with DRA_ARTIFACTS_DIR for everything - [072855a](https://github.com/elastic/connectors/pull/3748/commits/072855ae06cc062d2b1cc3d11bf451d4a9736e97) - that way the line

```
echo "The artifacts are: $(ls $DRA_ARTIFACTS_DIR)"
```
will actually list all artifacts, including the zip.

I don't fully understand these comments though tbh - we do cp for the Docker images too, why is that "not the same thing"? To my mind it's even closer, because we're copying artifacts we've downloaded from buildkite pipeline run. 🤷

.buildkite/publish/dra/init_dra_publishing.sh
lorenabalan: Don't fully understand this, I tried to tidy up this file in [072855a](https://github.com/elastic/connectors/pull/3748/commits/072855ae06cc062d2b1cc3d11bf451d4a9736e97). Let me know if it makes sense end-to-end.

.buildkite/publish_to_pypi.sh
lorenabalan: Thanks! Added the secrets under the expected path and updated vault paths in [df3f4d6](https://github.com/elastic/connectors/pull/3748/commits/df3f4d68777a38a0d484661785d88aa529662cfa)

.buildkite/publish/dra/init_dra_publishing.sh
seanstory: We probably could put in some work to also treat the docker images the same way. From what I remember:
1. they start out in a different dir, because of the different buildkite step that creates them
2. more than being copied around, they need some specific renaming

.buildkite/pypi-publish-pipeline.yml
lorenabalan: It wasn't about security. We didn't want to push a real package to PyPI and then "abandon it". I think pushing to test pypi is proof enough that the pipeline & process would work, and when/if we decide to come back to this we can publish with each release.

.buildkite/publish/dra/init_dra_publishing.sh
lorenabalan: Raised here https://github.com/elastic/infra/pull/43022

.buildkite/test_python_packages.sh
lorenabalan: Sure, added in [e27a7ae](https://github.com/elastic/connectors/pull/3748/commits/e27a7aee37c78512558e8223ad4902c0c13725ba)

.buildkite/publish/dra/init_dra_publishing.sh
lorenabalan: addressed in [2024377](https://github.com/elastic/connectors/pull/3748/commits/202437798d377cba8f127b3a8ff30647e8d71b28)

.buildkite/publish/dra/init_dra_publishing.sh
lorenabalan: addressed in [2024377](https://github.com/elastic/connectors/pull/3748/commits/202437798d377cba8f127b3a8ff30647e8d71b28) thank you!

PR #3761, opened 2025-10-06
Move _message_doc func into GMailDataSource class

mattnowzari:
# 🚧 Connectors Framework Improvements 🚧 

Move `_message_doc` function inside the `GMailDataSource` class so that we don't have to expose it in gmail's `__init__.py` but can still unit test it in `test_gmail.py`

#### Pre-Review Checklist

## Related Pull Requests
Connectors sources file -> folder work: https://github.com/elastic/connectors/pull/3747


app/connectors_service/connectors/sources/gmail/datasource.py
```diff
@@ -243,14 +243,37 @@ async def get_access_control(self):
 
                 yield self._user_access_control_doc(user, access_control)
 
+    def _message_doc(self, message):
```
lorenabalan: That's a clever way to address it, thank you! ⭐ 

Actually this can be a static method `@staticmethod` because `self` is not actually used inside. That way your test won't need to _instantiate_ a `GMailDataSource` - it can use `GMailDataSource._message_doc(message)` instead.

lorenabalan (approved):

app/connectors_service/connectors/sources/gmail/datasource.py
mattnowzari: Ah that's very neat! I'll implement this, as it makes the test cleaner

PR #3766, opened 2025-10-07
Split Atlassian connectors

lorenabalan:
## Description

Split Jira & Confluence connectors.

## Related Pull Requests

* https://github.com/elastic/connectors/pull/3747


app/connectors_service/connectors/sources/atlassian/__init__.py
```diff
@@ -0,0 +1,5 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+#
```
mattnowzari, line 5: Thoughts on adding `utils.py` and `validator.py` here as top-level imports? In the immediate term it could make their import statements in tests a bit shorter, but it's also not a "must-have" either

app/connectors_service/connectors/sources/atlassian/confluence/client.py
```diff
@@ -0,0 +1,388 @@
...
+class InvalidConfluenceDataSourceTypeError(ValueError):
+    pass
+
+
+class ThrottledError(Exception):
+    """Internal exception class to indicate that request was throttled by the API"""
+
+    pass
+
+
```
erikcurrin-elastic, line 50: These exceptions seem like they could be shared?  Or do we want them specific to a connector?

lorenabalan: Oh yeah good point, I'll move them one level up!

lorenabalan: Actually there's only 2 or 3 in common. I'll keep them next to the respective clients for now, but I think further down the line we can have some generic exceptions in the library that can be used by _all_ clients.

app/connectors_service/connectors/sources/atlassian/__init__.py
lorenabalan: So normally the convention is that what's in `__init__.py` is "public API" and allowed for consumption outside of this folder. I prefer not exposing utils in top-level `__init__` as they should be used only within the `atlassian` package, rather than have them as public API. 
Subpackages of `atlassian` should import from specific module directly (i.e. confluence/client should import from `atlassian.validator` rather than just from `atlassian`). Otherwise we can get ourselves into some circular imports.

PR #3774, opened 2025-10-08
Split out one datasource into libs

lorenabalan:
## Description

Related to https://github.com/elastic/search-team/issues/10999
Create `elasticsearch-connectors-sources` Python package and extract one source (`dir`) from `app/` folder.

TODO: 
- [ ] fix diff script?

### Manual QA
Created a custom connector of `service_type: 'dir'` in the config.yml. Sync worked as expected:
Screenshot 2025-10-10 at 17 14 32

#### Pre-Review Checklist

### DIFF
--- a/libs/connectors_sdk/NOTICE.txt
+++ b/libs/connectors_sdk/NOTICE.txt
@@ -715,7 +715,7 @@ Apache License
 
 attrs
 25.4.0
-UNKNOWN
+MIT
 The MIT License (MIT)
 
 Copyright (c) 2015 Hynek Schlawack and the attrs contributors
@@ -948,7 +948,7 @@ Apache Software License
 
 build
 1.3.0
-UNKNOWN
+MIT
 Copyright © 2019 Filipe Laíns <filipe.lains@gmail.com>
 
 Permission is hereby granted, free of charge, to any person obtaining a
@@ -1221,7 +1221,7 @@ Apache License
 
 idna
 3.11
-UNKNOWN
+BSD-3-Clause
 BSD 3-Clause License
 
 Copyright (c) 2013-2025, Kim Davies and contributors.
@@ -1516,7 +1516,7 @@ THE SOFTWARE.
 
 typing_extensions
 4.15.0
-UNKNOWN
+PSF-2.0
 A. HISTORY OF THE SOFTWARE
 ==========================
 
--- a/libs/connectors_sources/NOTICE.txt
+++ b/libs/connectors_sources/NOTICE.txt
@@ -715,7 +715,7 @@ Apache License
 
 attrs
 25.4.0
-UNKNOWN
+MIT
 The MIT License (MIT)
 
 Copyright (c) 2015 Hynek Schlawack and the attrs contributors
@@ -948,7 +948,7 @@ Apache Software License
 
 build
 1.3.0
-UNKNOWN
+MIT
 Copyright © 2019 Filipe Laíns <filipe.lains@gmail.com>
 
 Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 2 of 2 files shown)


libs/connectors_sources/tests/utils.py
lorenabalan, line 1: This is a duplicate! of tests.sources.support in app/

libs/connectors_sources/ftests/fixture.py
lorenabalan, line 1: This is also a duplicate of app/connectors_service/tests/sources/fixtures/fixture.py :(

libs/connectors_sources/ftests/ftest.sh
```diff
@@ -0,0 +1,381 @@
...
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+ROOT_DIR="$SCRIPT_DIR/.."
+# Use the main connectors_service venv which has all dependencies
+# Path from libs/connectors_sources -> ../../app/connectors_service
+CONNECTORS_SERVICE_DIR="$ROOT_DIR/../../app/connectors_service"
+VIRTUAL_ENV="$CONNECTORS_SERVICE_DIR/.venv"
+PLATFORM='unknown'
+MAX_RSS="200M"
+MAX_DURATION=600
+CONNECTORS_SOURCE_VERSION=$(cat "$ROOT_DIR/connectors_sources/VERSION")
```
lorenabalan, line 20: Not sure if it should at the sources or at the service version

libs/connectors_sources/ftests/ftest.sh
```diff
@@ -0,0 +1,381 @@
...
+
+unamestr=$(uname)
+if [[ "$unamestr" == 'Linux' ]]; then
+   PLATFORM='linux'
+elif [[ "$unamestr" == 'Darwin' ]]; then
+   PLATFORM='darwin'
+fi
+
+
+cd $ROOT_DIR/ftests
```
lorenabalan, line 286: As far as a I remember, this line and the constants at the top are the only differences between this file and the original test.sh from app/.

libs/connectors_sources/ftests/ftest.sh
artem-shelkovnikov: This one's a little awkward. Since functional tests are using the app, might it make sense to move functional tests into the connectors service, or even make it a separate app?

lorenabalan: To me they make more semantic sense if they're next to the datasource they're emulating. I think it's fine to pull the service as a dependency for the ftests (I'll check if I can make some changes to this script), given the sources are mainly only useful in that context anw. 

When a new contributor adds a new connector, they'll have to
* define the implementation
* add unit tests
* add functional tests

To me all of these should be encapsulated in one place/library, rather than modify the tests of two sources.

artem-shelkovnikov: This feels like a circular dependency in a way, though - connectors lib do not really depend on the service, but needs to run it to do the testing. That's the part that I don't really like.

For now the service is pretty stable (nothing's going on), but if we'd continue work there it's gonna become really messy.

I don't really think it's a big problem to have implementation + unit-tests in one place and functional tests in a different place, might be good to open this discussion point to the whole team in Slack?

lorenabalan:
We can revisit this in the future - the only unanswered question is around ftests, otherwise it's pretty much there. However, this is really only good to have if we split _all_ of the datasources out, as otherwise we will end-up in a halfway state. Given existing commitments and priorities, we're pausing this effort here.

PR #3782, opened 2025-10-10
Split GraphQL, Microsoft Teams, ServiceNow connectors

lorenabalan:
## Related PRs
https://github.com/elastic/connectors/pull/3747

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -1880,7 +1880,7 @@ SOFTWARE.
 
 
 azure-core
-1.35.1
+1.36.0
 MIT License
 Copyright (c) Microsoft Corporation.
 


app/connectors_service/connectors/sources/microsoft_teams/client.py
```diff
@@ -0,0 +1,600 @@
...
+]
+FILE_WRITE_CHUNK_SIZE = 1024 * 64  # 64KB default SSD page size
+TOKEN_EXPIRES = 3599
+RETRY_COUNT = 3
+RETRY_SECONDS = 30
+RETRY_INTERVAL = 2
+RUNNING_FTEST = (
+    "RUNNING_FTEST" in os.environ
+)  # Flag to check if a connector is run for ftest or not.
+override_url = os.environ["OVERRIDE_URL"]
```
mattnowzari: I noticed CI had failed bc the "OVERRIDE_URL" key isn't present - prob not in Buildkite's env for Py3.10/11 tests

```
ERROR tests/sources/test_microsoft_teams.py - KeyError: 'OVERRIDE_URL'
```

L39-42 were previously inside the `if` statement on L44 that guards against that key not being defined - maybe we should move them back inside the `if` statement?

lorenabalan: Ah yes this slipped through back-to-back refactoring/moves where I changed my mind. Thanks a lot for both suggestions, addressed them in follow-up commit! ❤️ 🙏

PR #3793, opened 2025-10-20
Fix typo in warning message about token privileges

jlind23:
Simply does what the tin say.


jlind23:
Thanks @lorenabalan for the review and sorry for missing the test while updating the error message 😓

PR #3806, opened 2025-10-22
Fix diff logic

lorenabalan:
## Description
* Updated watched paths to be more precise (`app/connectors_service/` prefix)
* Updated "${DOCKERFILE_FTEST_PATH}" from paths to watch. After https://github.com/elastic/connectors/commit/7cd4a4c68d4eedc89d482caa0c468642d0f4d6f7 it wasn't being set anywhere in `pipeline.yml` so ending up as `""`. According to AI, what ends up happening under the hood is 

> When DOCKERFILE_FTEST_PATH is empty, the grep command becomes:
  echo ".buildkite/pipeline.yml" | grep -q "^"
  An empty pattern with ^ means "match lines that start with... anything" - which matches EVERY line!

* Updated `diff` script according to IDE & AI suggestions

## Testing
* CI for [7359254](https://github.com/elastic/connectors/pull/3806/commits/73592548de9448629ab78b0e5f1e3838126e9eb4) only runs azure blob storage ftest https://buildkite.com/elastic/connectors/builds/20480#019a0c3a-0ca5-496a-99d4-7ed21a1a04c3 
* CI for this PR otherwise doesn't run any ftests

### DIFF
--- a/.backportrc.json
+++ b/.backportrc.json
@@ -3,9 +3,7 @@
     { "name": "main", "checked": true },
     "9.2",
     "9.1",
-    "9.0",
-    "8.19",
-    "8.18"
+    "8.19"
   ],
   "fork": false,
   "targetPRLabels": ["backport"],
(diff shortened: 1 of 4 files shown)


mattnowzari (approved):
Nice!

PR #3812, opened 2025-10-27
README updates for Connectors improvements

mattnowzari:
## README updates in conjunction with Connectors Framework Improvements

Updating the top-level README + adding some basic content to the `app/connectors_service/` and `libs/connectors_sdk` READMEs too

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,9 @@ There are two directories that contain the relevant code for running a connector
 The `app/` directory contains the primary connector service, as well as the individual data source implementations.
 The `libs/` directory is the home of the Connectors SDK, which is where the connectors framework is defined.
 
-The Connectors SDK is a dependency of the connectors application. You can use the Connectors SDK in and of itself in your own projects should you choose.
+The Connectors SDK is a dependency of the connectors application. You can use the Connectors SDK in and of itself to create your own data source connectors without needing to ingest data into Elasticsearch.  
 
-> [!NOTE] For those who are familiar with the monolithic project structure that preceded the current structure
+> [!NOTE] For those who are familiar with the [monolithic project structure](https://github.com/elastic/connectors/tree/1efa53e1ce47db2ae28d1390e5721932c58ce816) that preceded the current structure
 > Functionality has not been affected in any way.
 > The top-level project Makefile provides the same general commands as were previously present, and they will behave the same as before.
 >


README.md
```diff
@@ -33,14 +33,34 @@ Refer to the [Connector API documentation](https://www.elastic.co/docs/api/doc/e
...
 
-In addition to the source code for individual connectors, this repo also contains the connector service code, which is used for tasks like running connectors, and managing scheduling, syncs, and cleanup.
+As mentioned in the section above, the connector service code lives under the `app/` directory. The code here defines tasks such as running connectors and managing scheduling, syncs, and cleanup.
 This is shared code that is not used by individual connectors, but helps to coordinate and run a deployed instance/process.
 
-## Connector framework
+## Connectors SDK and framework
 
-This repo is also the home of the Elastic connector framework. This framework enables developers to build Elastic-supported self-managed connectors.
+The `libs/` directory is home to the Elastic connector SDK. This SDK provides a framework that enables developers to build self-managed connectors.
```
mattnowzari, line 63: Note to reviewer - I removed the phrase "Elastic-supported" bc I don't quite remember where we left off on our support burden for community connectors. If we are still going to take on the maintenance burden of community connectors I'll gladly put it back in

README.md
```diff
@@ -33,14 +33,34 @@ Refer to the [Connector API documentation](https://www.elastic.co/docs/api/doc/e
...
+
+- `app/`
+- `libs/`
+
+The `app/` directory contains the primary connector service, as well as the individual data source implementations.
+The `libs/` directory is the home of the Connectors SDK, which is where the connectors framework is defined.
+
+The Connectors SDK is a dependency of the connectors application. You can use the Connectors SDK in and of itself in your own projects should you choose.
+
+> [!NOTE] For those who are familiar with the monolithic project structure that preceded the current structure
```
mattnowzari: Note to reviewer - made this a [!NOTE] so it grabs the attention of folks who are wondering what's changed but will be easily removable once enough time has passed that the 'old' project structure is largely forgotten

artem-shelkovnikov (commented):
Looks like a good starting point!

libs/connectors_sdk/README.md
```diff
@@ -1 +1,64 @@
...
+
+The Connectors SDK is a framework for writing data connectors. This library is a dependency of the Connectors service found under `app/connectors_service`.
+
+Furthermore, you can use this SDK as a standalone framework to author simple data source connectors without having to ingest data directly into Elasticsearch.
+
+## What's here?
+- A `pyproject.toml` file
+- The connectors framework code
+
+## Simple code example
```
mattnowzari, line 11: Note to reviewer - I'm realizing this is my first time actually writing async Python from scratch, any critiques here to make this suitable as a simple example would be much appreciated

README.md
mattnowzari: > For those who are familiar with the monolithic project structure that preceded the current structure

Something about this sentence I wrote isn't vibing with me, suggestions are most welcome!

libs/connectors_sdk/README.md
artem-shelkovnikov: The example itself doesn't use async in a way in the connector - `async` is good for long-running operations that involve waiting for response for a long time.

Might not be worth to add it here, but if you want to you can probably use `asyncio` to query some websites/public APIs and return their content. That, as I mentioned, might not really be needed.

mattnowzari: Good callout Artem, I've simplified the example to not use `asyncio` and kept the example super simple as I want the focus to be on the importing and basic objects you'd need to get going. I figure if we have time in the future we can flesh out usage examples even more 🤞

README.md
```diff
@@ -33,14 +33,34 @@ Refer to the [Connector API documentation](https://www.elastic.co/docs/api/doc/e
...
+
+There are two directories that contain the relevant code for running a connector:
+
+- `app/`
+- `libs/`
+
+The `app/` directory contains the primary connector service, as well as the individual data source implementations.
+The `libs/` directory is the home of the Connectors SDK, which is where the connectors framework is defined.
+
+The Connectors SDK is a dependency of the connectors application. You can use the Connectors SDK in and of itself in your own projects should you choose.
```
lorenabalan: I think it'd be worth mentioning why they'd use it in their own project - for defining their own connectors. They would still have to merge a 1-line code to make their connector discoverable by the service, but we can cross the bridge when/if a user tries that path.

README.md
lorenabalan: Maybe reference a commit sha of the "previous structure" to make it clear?

PR #3841, opened 2025-11-07
Fix DRA file not containing the whole project

artem-shelkovnikov:
Minor fix to unlock Agent DRA: we need to actually zip the whole project for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ install:
 	cd $(connectors_sdk_dir); make install
 	cd $(app_dir); make install
 
+install-agent: install
+
 install-package:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -11,6 +11,8 @@ install:
 	cd $(connectors_sdk_dir); make install
 	cd $(app_dir); make install
 
+install-agent: install
```
lorenabalan, line 14: Do we need this if they're equivalent?

artem-shelkovnikov: Supposedly not, but in reality it's not gonna do almost anything and I kept it for safety 🤷

PR #3844, opened 2025-11-10
Couple more Agent fixes

artem-shelkovnikov:
See descriptions for each fix.

TLDR: our agent DRA are not being produced + after the project refactoring our agent docker image stopped working, so fixing that while also fixing small issues from the past.

### DIFF
--- a/Dockerfile.wolfi
+++ b/Dockerfile.wolfi
@@ -1,4 +1,4 @@
-FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:b94c277ba6ad5e2a9031efeff84477b62d4d67790628a21d83796c8109ad9fcd
+FROM docker.elastic.co/wolfi/python:3.11-dev@sha256:3b2cbf39767c562dec8665aa0315372b1b178109fb5d04077b7bde34a2feb262
 USER root
 COPY . /app
 WORKDIR /app


Dockerfile.agent
```diff
@@ -1,7 +1,6 @@
 # This file is for internal experimental purposes only.
 # Please do not use this file for any real-world workloads.
-
-FROM docker.elastic.co/elastic-agent/elastic-agent:9.0.0-SNAPSHOT
+FROM docker.elastic.co/elastic-agent/elastic-agent:9.3.0-SNAPSHOT
```
artem-shelkovnikov, line 3: We used an old version that had a bug with the way it's reading env variables, so swapped to latest and it fixed it

Makefile
```diff
@@ -66,3 +66,25 @@ zip: clean
 		$(PACKAGE_NAME_VERSION).zip \
 		./* \
 		-x  *htmlcov*/* *docs*/* Dockerfile* *ruff_cache*/* *pytest_cache*/* *__pycache__*/* *build*/* *egg-info*/*
+
```
artem-shelkovnikov, line 69: Moved from `app/connectors_service/Makefile` - it would have never worked there

Makefile
```diff
@@ -66,3 +66,25 @@ zip: clean
...
+## Agent Docker Zone
+# Only use it for local testing, that's it
+AGENT_ES_HOSTS?='http://host.docker.internal:9200'
```
artem-shelkovnikov, line 72: Had to fix this as it was `127.0.0.1` which actually never worked in docker

app/connectors_service/connectors/agent/config.py
```diff
@@ -71,7 +71,9 @@ def try_update(self, connector_id, service_type, output_unit):
 
         # Auth-related
         if has_hosts and (has_api_key or has_basic_auth):
-            es_creds = {"host": source["hosts"][0]}
+            hosts = source["hosts"] if isinstance(source["hosts"], str) else source["hosts"][0]
```
artem-shelkovnikov: This was not parsing the default config correctly, taking the string `[http://host.docker.internal:9200]` and just taking the first symbol out of it

PR #3868, opened 2025-11-27
[Network Drive] Switch to usage of local connection cache

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/2418

Conceptual problem - `smbprotocol` has a global connection cache [here](https://github.com/jborean93/smbprotocol/blob/master/src/smbclient/_pool.py#L33) (see usages). Hash key is host/port:

```python
connection_key = f"{server.lower()}:{port}"
```

Library is not natively async, that means that it can cause fun stuff when used in asynchronous/threaded environment. In our case one connector configuration can spawn at least 2 connections for the same server: 1 for running `get_docs` method while the sync is running and 1 for `ping` method that gets spawned every 30 seconds by default.

Things get worse if advanced filtering rules are used and several connectors point to the same Network Drive.

What happens is:

1. `get_docs` is called and a connection is established
2. `ping` is triggered, connection is re-used from the cache
3. `ping` finishes and closes the connection. That causes the loss of the connection by `get_docs`
4. `get_docs` sometimes can re-establish it (see [this code](https://github.com/elastic/connectors/blob/main/app/connectors_service/connectors/sources/network_drive/datasource.py#L276)). Sometimes it just fails

This change attempts to address it by providing connection cache that's local to each connector instead of using a global one. This way connector instances would not attempt to close each other's connections if they connect to the same resource.

#### Pre-Review Checklist

## Release Note

Fix the bug with Network Drive connector closing connections to SMB servers prematurely.

### DIFF
--- a/app/connectors_service/connectors/sources/network_drive/datasource.py
+++ b/app/connectors_service/connectors/sources/network_drive/datasource.py
@@ -82,7 +82,13 @@ class NASDataSource(BaseDataSource):
 
     @cached_property
     def smb_connection(self):
-        return SMBSession(self.server_ip, self.username, self.password, self.port, self._connection_cache)
+        return SMBSession(
+            self.server_ip,
+            self.username,
+            self.password,
+            self.port,
+            self._connection_cache,
+        )
 
     @classmethod
     def get_default_configuration(cls):
@@ -204,7 +210,7 @@ class NASDataSource(BaseDataSource):
                             username=self.username,
                             password=self.password,
                             port=self.port,
-                            connection_cache=self._connection_cache
+                            connection_cache=self._connection_cache,
                         ),
                     )
                 )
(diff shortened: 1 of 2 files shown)


app/connectors_service/connectors/sources/network_drive/datasource.py
```diff
@@ -75,13 +75,20 @@ def __init__(self, configuration):
         self.drive_type = self.configuration["drive_type"]
         self.identity_mappings = self.configuration["identity_mappings"]
         self.security_info = SecurityInfo(self.username, self.password, self.server_ip)
+        self._connection_cache = {}
```
lorenabalan, line 78: Aren't we supposed to update this somewhere? 🤔 Isn't this `{}` always, or am I missing sth obvious?

artem-shelkovnikov: We pass it into methods that connect to SMB below. Internally they manage the cache - add records, remove records and such. For example see https://github.com/jborean93/smbprotocol/blob/master/src/smbclient/_pool.py#L407

lorenabalan: Ooooh I see now! Thanks for the link!! I should've expanded more code, hadn't realised it's an existing `smbclient` parameter. LGTM!

lorenabalan (approved):
LGTM

PR #3885, opened 2025-12-09
[MSSQL, PostgreSQL] Resolve case sensitivity mismatch in id_columns transformation

dennis-tismenko:
## Closes https://github.com/elastic/connectors-py/issues/3884

This PR fixes a case sensitivity bug in the PostgreSQL and MSSQL connectors when using `id_columns` in advanced sync rules:
- **PostgreSQL**: Added `.lower()` to the `id_columns` transformation
- **MSSQL**: Added `.lower()` and `sorted()` to match `map_column_names()` behaviour

#### 1. Start MSSQL
```bash
docker run -d --name mssql-test --net elastic \
  -e ACCEPT_EULA=Y -e MSSQL_SA_PASSWORD='Password_123' \
  -p 1433:1433 mcr.microsoft.com/azure-sql-edge:latest
```

#### 2. Create test data
```
docker run --rm --net elastic mcr.microsoft.com/mssql-tools:latest \
  /opt/mssql-tools/bin/sqlcmd -S mssql-test -U sa -P 'Password_123' -Q "
CREATE DATABASE xe;
GO
USE xe;
CREATE LOGIN admin WITH PASSWORD = 'Password_123';
ALTER SERVER ROLE [sysadmin] ADD MEMBER [admin];
GO
USE xe;
CREATE TABLE COAXIS_RV_CALL (
    Call_No INT PRIMARY KEY,
    Call_Date DATETIME,
    Call_Description VARCHAR(255)
);
INSERT INTO COAXIS_RV_CALL (Call_No, Call_Date, Call_Description) VALUES
    (1001, '2025-01-15', 'Server maintenance request'),
    (1002, '2025-01-16', 'Network connectivity issue'),
    (1003, '2025-01-17', 'Database backup failure'),
    (1004, '2025-01-18', 'Application deployment'),
    (1005, '2025-01-19', 'Security patch update');
GO
"
```

#### 3. Create connector in Kibana

| Field | Value |
|-------|-------|
| Host | `host.docker.internal` (or `localhost` if running from source) |
| Port | `1433` |
| Username | `admin` |
| Password | `Password_123` |
| Database | `xe` |
| Schema | `dbo` |

#### 4. Set advanced sync rule
```json
[
  {
    "tables": ["COAXIS_RV_CALL"],
    "query": "SELECT * FROM COAXIS_RV_CALL",
    "id_columns": ["Call_No"]
  }
]
```

#### 5. Run a __Full content sync__
Result: __5__ docs upserted (previously 1).
Screenshot 2026-03-09 at 17 25 54@2x

#### Pre-Review Checklist
- [x] if you added or changed Rich Configurable Fields for a Native Connector, you made a corresponding PR in [Kibana](https://github.com/elastic/kibana/blob/main/packages/kbn-search-connectors/types/native_connectors.ts) — N/A

## Release Note

**PostgreSQL/MSSQL connectors**: Fixed a bug where using `id_columns` in advanced sync rules with mixed-case table or column names caused all documents to receive the same `_id`, resulting in document overwrites and only 1 document being indexed instead of the expected count.

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -2713,7 +2713,7 @@ documentation is licensed as follows:
 
 
 charset-normalizer
-3.4.4
+3.4.5
 MIT
 MIT License
 
@@ -4146,7 +4146,7 @@ Apache Software License
 
 
 google-auth
-2.48.0
+2.49.0
 Apache Software License
                                  Apache License
                            Version 2.0, January 2004


app/connectors_service/connectors/sources/mssql/datasource.py
```diff
@@ -341,13 +341,10 @@ async def get_docs(self, filtering=None):
...
                 id_columns = rule.get("id_columns", [])
 
-                id_columns_str = ""
-                for i, table in enumerate(tables):
-                    if i == 0:
-                        id_columns_str = f"{self.schema}_{table}"
-                    else:
-                        id_columns_str = f"{id_columns_str}_{table}"
-                id_columns = [f"{id_columns_str}_{column}" for column in id_columns]
+                id_columns = [
```
Apmats: I think what needs to happen is that you have to pass both the fetch-from-db side and the configured-from-config-columns through the same transformation.

The other side puts this through `map_column_names` while here and in the postgres connector we have 2 different pieces of code that try to do the same, which makes things a bit more confusing and brittle (as someone might not understand that these are to be fully in sync and make small adjustments).

Based on this, I say we pass the columns here through the `map_column_names` function as well - but I might be missing something.

PR #3893, opened 2025-12-17
Fall back to previous minor, if necessary

seanstory:
Related to https://github.com/elastic/search-team/issues/11615

https://github.com/elastic/connectors/pull/3892's CI is currently failing, like:

```
+ echo 'Failed to get tarball URL for version 9.4.0-SNAPSHOT, trying next...'
+ continue
+ echo 'Error: Failed to fetch any compatible Elasticsearch version'
+ return 1
``` 

This is happening because there's no PATCH before `0` in the `9.4.0` current version to fall back to. But I don't like having to wait for Elasticsearch to bump to a new minor before we can bump our branches every time. 

This change adds an undocumented Env Var that will allow our Ftests to bypass version compatibility checks, and also lets the ftest shell script download a past MINOR version of ES, if it can't find one for the current version AND if the current patch version is `0`. This scenario should only occur when we are all bumping to a new minor. 

#### Pre-Review Checklist

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -2657,7 +2657,7 @@ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
 
 
 cachetools
-6.2.3
+6.2.4
 UNKNOWN
 The MIT License (MIT)
 
@@ -4171,7 +4171,7 @@ Apache Software License
 
 
 google-auth
-2.43.0
+2.45.0
 Apache Software License
                                  Apache License
                            Version 2.0, January 2004


app/connectors_service/tests/ftest.sh
```diff
@@ -68,9 +69,15 @@ function get_fallback_versions {
...
   fallback_versions+=("${major}.${minor}.${patch}-SNAPSHOT")
 
-  # Try only n-1 (previous patch version)
   if [ "$patch" -gt 0 ]; then
+    # Try n-1 (previous patch version)
     fallback_versions+=("${major}.${minor}.$((patch-1))-SNAPSHOT")
+  elif [ "$minor" -gt 0 ]; then
+    # When patch is 0, fall back to previous minor version
+    # Try a few patch versions of the previous minor
+    for prev_patch in 5 4 3 2 1 0; do
```
Apmats, line 78: We do frequently go beyond 5 in patches, but it shouldn't matter much here as we usually don't have x.x.6 before we have x.x+1.0 out?

Apmats (approved):
Nice workaround, and thank you.

app/connectors_service/tests/ftest.sh
seanstory: Yeah exactly. I originally had `9 8 7 6 5 4 3 2 1`, but this seemed excessive, and each one adds a little latency.

PR #3900, opened 2025-12-23
Fix PostgreSQL IP address serialization error

Apmats:
## Closes https://github.com/elastic/connectors-py/issues/3879

Problem:

The PostgreSQL connector fails with  serialization error because asyncpg returns PostgreSQL's INET and CIDR columns as Python ipaddress module objects (IPv4Address, IPv6Address, IPv4Network, etc.) rather than strings. Our serialization layer doesn't handle these types, causing the connector to crash when attempting to convert documents to JSON.

Solution:

The implementation converts these objects to strings before calling the parent serializer, which handles standard types. Recursive serialization handles nested structures (lists/dicts containing IP addresses). Followed the same pattern from the MongoDB connector.

Only added plain tests for these changes, not ftest coverage. You, reviewer!, give me an opinion on this!

#### Pre-Review Checklist


artem-shelkovnikov (commented):
IMO it's really easy to add this to ftest and we should.

Ideally the table(s) in ftest should have all possible data types that require serialisation.

Also, I see their test suite here https://github.com/MagicStack/asyncpg/blob/9b2b027224058e4590fdb9e41b5750a64dab1fce/tests/test_codecs.py that has some more fun data types, do you wanna try them out too? Geospatial types, for example, or UUIDs?

Apmats:
@artem-shelkovnikov yeap working on that now, I knew it was coming. 😁

PR #3906, opened 2026-01-12
[9.0] Fix PostgreSQL IP address serialization error (#3900)

Apmats:
Manually backporting the work from the following PR to 9.0:
 - Fix PostgreSQL IP address serialization error (#3900)
 
 
  I thought doing https://github.com/elastic/connectors/pull/3904 would backport even when merging to something that's not main, but apparently I was wrong, so handling in separate PRs.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1880,7 +1880,7 @@ SOFTWARE.
 
 
 azure-core
-1.37.0
+1.38.0
 MIT License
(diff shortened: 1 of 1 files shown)


artem-shelkovnikov:
Haha I see, I also hit similar problems and indeed, I don't think `backport` can work from anything other than `main`:

https://github.com/elastic/connectors/blob/main/.github/workflows/backport.yml

PR #3940, opened 2026-02-12
Fix multiservice death sequence

Apmats:
## Closes https://github.com/elastic/connectors-py/issues/3939

See issue for details, just implementing suggested fix from over there, don't mind me!

#### Pre-Review Checklist


app/connectors_service/connectors/services/base.py
```diff
@@ -172,11 +172,19 @@ def __init__(self, *services):
 
     async def run(self):
         """Runs every service in a task and wait for all tasks."""
-        tasks = [asyncio.create_task(service.run()) for service in self._services]
+        task_to_service = {}
+        tasks = []
```
lorenabalan: I'm not sure if we need both the dict and the list? Would we have duplicate tasks? If not we can keep the dict and update L182 to pass just the .keys(). Dicts are ordered in insertion ordered so it'll be deterministic.

Apmats: That's perfectly fine to do, no duplicates. I don't think we even really care about the order either, even if it wasn't guaranteed. Applied this.

PR #3941, opened 2026-02-13
Fix some script + make target discrepancies post refactor

Apmats:
## Summary

The monorepo restructure in #3721 moved the connector service into `app/connectors_service/`, but several file path references across Makefiles and stack scripts were not updated. This caused broken behavior for users running connectors from source:

- `make run` from the repo root silently ignored the user's `config.yml`, falling back to `service_cli.py`'s default path inside `app/connectors_service/`. This produced a confusing `Cannot connect to host localhost:9200` error even with a valid cloud config at the repo root.
- The `scripts/stack/` development scripts referenced stale paths for `VERSION`, `config.yml.example`, and the `connectors` CLI binary — all of which moved during the restructure.
- `run-stack.sh --no-connectors` still prompted for the connector configurator.
- The stack scripts themselves were left at the repo root despite being specific to the connector service.

### Makefile wiring for `make run`

- **Root Makefile**: added a `config.yml` target (copies from example if missing) and updated `run` to pass `CONFIG_FILE=$(CURDIR)/config.yml` to the inner Makefile.
- **`app/connectors_service/Makefile`**: added `CONFIG_FILE ?= config.yml` and updated `run` to pass `-c $(CONFIG_FILE)` explicitly. When called from root, it receives the absolute path; when called directly, it defaults to the local `config.yml`.

### Move `scripts/stack/` into `app/connectors_service/scripts/stack/`

Per review feedback, relocated the stack scripts to live alongside the service they belong to. This simplifies all internal path references because `PROJECT_ROOT` (computed as two directories up from the script location) now naturally resolves to `app/connectors_service/` — eliminating the need for `app/connectors_service/` prefixes throughout.

### Script fixes

- **`copy-config.sh`**: removed the intermediary `cp` that created a ghost `config.yml` at `PROJECT_ROOT`; now points directly at `config.yml.example`.
- **`configure-connectors.sh`**: fixed the `connectors` CLI path (`bin/connectors` → `.venv/bin/connectors`).
- **`run-stack.sh`**: moved config copy, configurator prompt, and connector startup inside the `no_connectors == false` block.

### Reference updates

- `.gitignore`: updated `connectors-config` path; added root `config.yml` to ignore list.
- `README.md`, `docs/CLI.md`, stack `README.md`: updated all `scripts/stack/` links and command examples to the new location.

#### Pre-Review Checklist

N/A — no security changes, no new dependencies.

## Related Pull Requests

* #3721 — Restructure Connectors project (introduced the monorepo layout)
* Kibana-side instructions may also need updating to reference `app/connectors_service/config.yml` for the "run from source" path

## Release Note

Fixed broken file path references in Makefiles and stack scripts that were not updated during the monorepo restructure (#3721). Users running connectors from source who were experiencing `Cannot connect to host localhost:9200` errors (despite having a valid `config.yml` at the repo root) will find that `make run` now correctly picks up their configuration. The `scripts/stack/` development scripts have also been relocated to `app/connectors_service/scripts/stack/` and fixed to reference correct post-restructure paths. These fixes will take effect after merging to the corresponding branches and are not strictly tied to a specific release.

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -8520,7 +8520,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 
 xmltodict
-1.0.3
+1.0.4
 UNKNOWN
 Copyright (C) 2012 Martin Blech and individual contributors.
 


scripts/stack/configure-connectors.sh
```diff
@@ -18,7 +18,7 @@ elif which python > /dev/null; then
     PYTHON_EXECUTABLE="python"
 fi
 
-pushd $PROJECT_ROOT
+pushd $PROJECT_ROOT/app/connectors_service
```
lorenabalan: Where's PROJECT_ROOT defined? I thought that was modified to point to app/connectors_service but I might be misremembering!

Apmats: It's under `scripts/stack/set-env.sh` - I didn't think about this subject at all, but you might still have feelings on this since this is largely your baby 😁 - do you think we should refactor where these scripts live too?

lorenabalan: I wouldn't be opposed to it, I'm not sure there was any explicit reason why they were left behind. If they're not used in CI or root-level `Makefile` then it would make sense to me to move them. Looking at them they all seem related to the service itself anw.

Apmats: Alright. Moved and adjusted stuff around that move.

Apmats:
@artem-shelkovnikov the reorg we did was 9.3+ right? 

Edit: 
https://github.com/elastic/connectors/tree/9.2/app/connectors_service


So yes this doesn't need to go before that

lorenabalan (approved):
LGTM!

PR #3951, opened 2026-02-23
Add DRA version bump pipeline

ninalee12:
## Completes Phase 0 and 1 https://github.com/elastic/search-team/issues/13038
Hi Team,

This PR is the first phase of the version-bump automation rollout as mentioned in [#mission-control](https://elastic.slack.com/archives/C0JFN9HJL/p1770388488482519). It introduces a generalized Buildkite pipeline for service teams in the DRA process. The baseline pipeline includes a block step that waits for each team to unblock before polling for DRA artifacts using a buildkite plugin [json-watcher-buildkite-plugin](https://github.com/elastic/json-watcher-buildkite-plugin) for the polling.


.buildkite/version-bump-pipeline.yml
```diff
@@ -0,0 +1,61 @@
+notify:
+  - slack:
+      channels:
+        - "#search-extract-and-transform"
```
artem-shelkovnikov, line 4: Do we want it here, or to [#search-et-alerts](https://elastic.slack.com/archives/C04KMMNAHGB)?

.buildkite/version-bump-pipeline.yml
```diff
@@ -0,0 +1,61 @@
...
+    if: (build.branch == 'main' || build.branch =~ /^[0-9]+\.[0-9x]+\$/) && (build.state == 'passed' || build.state == 'failed')
+  - slack:
+      channels:
+        - "#search-extract-and-transform"
```
artem-shelkovnikov, line 8: Same as above

.buildkite/version-bump-pipeline.yml
```diff
@@ -0,0 +1,61 @@
...
+
+        Ready to fetch DRA artifacts - please unblock when ready.
+        New version: `${NEW_VERSION}`
+        Branch: `${BRANCH}`
+        Workflow: `${WORKFLOW}`
+        ${BUILDKITE_BUILD_URL}
+    if: build.state == "blocked"
+
+steps:
+  # TODO: replace this block step by real version bump logic
```
artem-shelkovnikov, line 21: Just in case, we have 2 version files starting at 9.3

.buildkite/version-bump-pipeline.yml
```diff
@@ -0,0 +1,61 @@
...
+  - label: "Fetch DRA Artifacts"
+    key: fetch-dra-artifacts
+    depends_on: block-get-dra-artifacts
+    agents:
+      image: docker.elastic.co/release-eng/wolfi-build-essential-release-eng:latest
+      cpu: 250m
+      memory: 512Mi
+      ephemeralStorage: 1Gi
+    command:
+      - echo "Starting DRA artifacts retrieval..."
```
artem-shelkovnikov, line 42: Real logic will be here?

.buildkite/version-bump-pipeline.yml
ninalee12: I'm not quite sure what the team's preference is. I went off of the slack channel commented here: https://github.com/elastic/search-team/issues/13038#issuecomment-3954666715

I'm happy to change it if we're all aligned

.buildkite/version-bump-pipeline.yml
ninalee12: The real logic for the version bump will be where the block step currently is. 

In the step you highlighted, it'll be used for polling the artifacts until the become available. Once they're available, your version bump process would be complete. 

The pipeline in this PR is used alongside the centralized pipeline (owned by release-eng).

How it'll work is that the centralized pipeline will be the orchestration piece that will trigger all DRA service team's version bump pipelines.

Version bump logic itself will still be handled by the service teams. There's a lot more details in the [PSI](https://docs.google.com/document/d/1VxKftPLqC3EtkxtjWpkbHqfP0Oen9142Piri-yviTVw/edit?tab=t.0#heading=h.a5xewkvnz7h4) that I won't be able to capture here

Hope this helps!

artem-shelkovnikov: So this PR will not introduce any logic, and you'll do it in follow-up PRs, right?

ninalee12: No, your team is responsible for the version bump logic in the pipeline. We (release-eng) have opened this PR to kick start the version bump pipeline process only. 

We're also in the process of accumulating and creating a FAQ for all version bump related questions. You can find the FAQ here: 
- https://docs.google.com/document/d/14FgV4dMF6Qz0N8yhYXEk147OE4W9UlZ5Ih7iAdlESL0/edit?usp=sharing

If you have any questions that may not be in the FAQ, feel free to ask them here as it may also help other teams that are undergoing the same process, and I'll provide any answers to your questions as best as I can.

Apmats:
Hey @ninalee12,

In the linked PR #3986 I initially went with a flow that directly pushes version bump commits to branches (no PRs involved) -- from what you're saying here I now understand the expectation is that the automation creates PRs with auto-merge, rather than pushing directly, although functionally the end result is the same.

I'm in the process of refactoring to a PR-based flow (not yet pushed, will be in an upcoming commit on #3986). The script will create PRs with the `ci:version-bump` label, enable auto-merge, and wait for them to land. I've also added a fast CI path so version bump PRs skip the full test suite.

However, to make this work we also need `pull_requests: write` added to our `GithubPermissionSet` in `elastic/terrazzo` (`manifests/prod/vault/connectors-permission-set.yml`) -- currently we only have `contents: write`, which is enough for direct pushes but not for creating/merging PRs via the GitHub API.

Can you confirm PRs are the expected approach? If so, I'll open the Terrazzo PR and push the refactored version-bump script.

ninalee12:
Hey ~~@artem-shelkovnikov~~ 
EDIT: sorry pinged wrong person. Meant to ping @Apmats 

I hope you see this in time before the refactor!

I think there are various ways to achieve a version bump, it could be directly as you have mentioned or through a PR + auto-merge bot. So you're not limited to PRs only. 

I've only seen, so far, teams with automation to create PRs follow by a merge. If you already have automation for a direct version bump this could work too. 🤔 Although testing it could be tricky since it'll actually do a version bump, we may need to find some workaround for the test.

It's really up to your team to decide on how you want to achieve the version bump as each team have their own way of bumping. The buildkite plugin only expects to have the newly bumped DRA artifacts in a bucket for it to pass. 

You don't have to follow this approach since it may or may not work for you but here's an example from another team: https://github.com/elastic/apm-server/blob/main/.buildkite/version-bump-pipeline.yml

Apmats:
Thanks for clarifying @ninalee12! I think it makes sense that we go with the PR + auto-merge approach, at least to get to a first review and then if the team disagrees I'll refactor.

I updated #3986 accordingly.

I've also opened elastic/terrazzo#1190 to add `pull_requests: write` to our `GithubPermissionSet`.

 Once that's merged and applied, if we also merge #3986  we'll be able to trigger the pipeline and test both patch and minor flows, I think.

Then we can do the e2e test as well.

jmlrt:
based on https://github.com/elastic/platform-engineering-productivity/issues/917#issuecomment-4388397990 this pr can be closed

PR #3976, opened 2026-03-22
fix: remove unused elasticsearch-connectors-sdk dependency

x4v13r64:
Remove `elasticsearch-connectors-sdk` from `app/connectors_service/pyproject.toml` — this package is not used by the connectors service.


artem-shelkovnikov:
It's not unused. We've merged a temporary fix that provides additional protection against supply-chain attack, will do more afterwards.

PR #3986, opened 2026-04-03
Add DRA version bump automation

Apmats:
Includes and builds on #3951. Replaces the block step with automated version bump logic.

The centralized release-eng pipeline triggers `connectors-version-bump` with `NEW_VERSION`, `BRANCH`, and `WORKFLOW`. For `patch`, the script opens an auto-merging PR bumping both VERSION files on the release branch. For `minor`, it updates `pipeline.yml`'s DRA branch list, creates the new release branch, and bumps main to the next minor -- each step idempotent. The merged commits trigger existing CI which publishes DRA artifacts, and the json-watcher steps confirm completion.

Requires elastic/terrazzo#1190 for `pull_requests: write` permission. Set `DRY_RUN=true` to preview without side effects.

> **Note to reviewer**: we should port this to 9.2, 9.1, and 8.19 unless we explicitly decide to only do this for 9.3 onwards. Those branches have a single VERSION file (pre-monorepo layout) so this can't be backported as-is -- I'll work on bringing a separate PR to 9.2 to be backported.

## Test plan

- [x] Local: shell syntax, pipeline YAML, Python branch-insertion cases, env var validation, DRY_RUN for both workflows
- [ ] Buildkite: DRY_RUN trigger, patch PR (close without merging), minor flow with fake branch, end-to-end with @ninalee12


.buildkite/version-bump-pipeline.yml
```diff
@@ -0,0 +1,65 @@
+notify:
+  - slack:
+      channels:
+        - "#search-et-alerts"
+    if: (build.branch == 'main' || build.branch =~ /^[0-9]+\.[0-9x]+\$/) && (build.state == 'passed' || build.state == 'failed')
```
markjhoy, line 5: Not sure about the last part here... are there other states besides `passed` or `failed`? If not, we can remove the last part of this clause... or, do we want to make sure this runs only on passed states? (I'm not 100% sure myself...)

markjhoy:
Overall - looking pretty good.

Since this script is manually launched from Buildkite, I'm also wondering if it should also bump the DRA version numbers for minor branches [here in the main Buildkite file](https://github.com/elastic/connectors/blob/d6472ee819c9da551db2ad43893dd52a4dceb23a/.buildkite/pipeline.yml#L1100)?

.buildkite/version-bump-pipeline.yml
Apmats: This came from the original PR (#3951 from release-eng) - I didn't touch it. Bu Claude tells me that Buildkite has a bunch of other states beyond passed/failed (running, scheduled, blocked, canceled, skipped etc.), and without that filter we'd get a Slack ping on every state transition.

Apmats:
@markjhoy the `buildkite/publish/version-bump.sh` script updates that exact line in `pipeline.yml` to add the new branch to the DRA condition before creating the release branch (see the Python branch-insertion logic in there). That way I hope DRA publishing works on the new branch from day one instead of needing a manual follow-up.

markjhoy:
> @markjhoy the `buildkite/publish/version-bump.sh` script updates that exact line in `pipeline.yml` to add the new branch to the DRA condition before creating the release branch (see the Python branch-insertion logic in there). That way I hope DRA publishing works on the new branch from day one instead of needing a manual follow-up.

Ah - I missed that! Thanks for pointing it out.

Apmats:
Hi, this is Claude (driving Apostolos's terminal) -- we added an `AUTO_MERGE` env var to `.buildkite/publish/version-bump.sh`. Defaults to `true` so production behavior is unchanged. Setting `AUTO_MERGE=false` creates the PR but skips enabling auto-merge and the merge-wait poll, so PRs stay open for inspection and manual close. Useful for verifying the real PR creation path in Buildkite without landing changes.

Apmats:
@markjhoy if you're okay with the above, you can give me a ✅ and I'll move onto testing.

.buildkite/pipeline.yml
```diff
@@ -44,7 +44,19 @@ env:
   DRA_PYTHON_VERSION: "3.11"
 
 steps:
+  # Version bump PRs only touch VERSION files -- skip the full CI suite.
```
markjhoy: If we're checking the labels for each step below - we don't really need this step to just print out that we're skipping CI... I'd remove this block

markjhoy: The more I look at this - since the `catalog-info.yaml` changes already call `".buildkite/version-bump-pipeline.yml"` instead of using the main pipeline, I don't think it's necessary to change this main pipeline file at all... @seanstory - you probably have a bit more context here with the CI, thoughts?

.buildkite/publish/version-bump.sh
```diff
@@ -0,0 +1,303 @@
...
+  python3 -c '
+import sys
+
+pipeline_yml = sys.argv[1]
+new_branch = sys.argv[2]
+
+with open(pipeline_yml) as f:
+    pipeline = f.read()
+
+dra_marker = "ci:packaging"
```
markjhoy: It _might_ be safer to search for the comment `# Add new maintenance branches here` as well as a comment in the main .yml file telling people to keep the comment as-is...) as who knows what might happen in the future with this script (we might add a step in that meets these conditions 🤷 I tend to err on the side of caution)

.buildkite/publish/version-bump.sh
```diff
@@ -0,0 +1,303 @@
...
+# The list is the `if:` condition on the "Packaging and DRA" group step.
+add_branch_to_dra_list() {
+  local new_branch="$1"
+
+  echo "Adding ${new_branch} to DRA branch list in pipeline.yml"
+
+  python3 -c '
+import sys
+
+pipeline_yml = sys.argv[1]
```
markjhoy: (sorry about the last comment I made here --- just realized this was within a python code block 🤦 ) - ignore the (now deleted) last comment I had

.buildkite/publish/version-bump.sh
```diff
@@ -0,0 +1,303 @@
...
+
+PR_MERGE_TIMEOUT=3600  # 60 minutes
+PR_POLL_INTERVAL=30    # seconds
+
+# Checks if the first VERSION file already contains the given version.
+version_already_bumped() {
+  local version="$1"
+  local current
+  current=$(cat "${VERSION_FILES[0]}" 2>/dev/null || echo "")
+  [[ "${current}" == "${version}" ]]
```
markjhoy: Will it be enough to check to equality here? Or should this check be to ensure that the current version is >= to the version? It's a bit more work to separate / check the version parts as numeric, but might be safer.

.buildkite/publish/version-bump.sh
```diff
@@ -0,0 +1,303 @@
...
+
+# Insert new branch condition before the pull_request.labels clause
+pr_label_marker = "build.pull_request.labels"
+if pr_label_marker not in line:
+    print("Error: could not find pull_request.labels marker in condition", file=sys.stderr)
+    sys.exit(1)
+
+insertion = f"build.branch == {escaped}{new_branch}{escaped} || "
+lines[target_idx] = line.replace(pr_label_marker, insertion + pr_label_marker)
+
```
markjhoy: Would we want to consider removing older branches on this line at some point? IIRC - we only need to keep the last two minors of a major

markjhoy:
> @markjhoy if you're okay with the above, you can give me a ✅ and I'll move onto testing.

I _think_ it's good... I'd like to get a second set of eyes on it (maybe @artem-shelkovnikov ) though, especially around the questions I have about needing to edit the primary `pipeline.yml` file...

.buildkite/pipeline.yml
seanstory: Not sure I've got more context. But I do question if we actually want to skip CI on version bump. I think we probably don't. Our CI is not that long, and if there's something that breaks, I'd rather it break on the branch than on `main`.

Apmats: The only concern and why I did this is that we don't control the timeout set on the Buildkite trigger step that calls us to do this bump. I agree it's better to break on the PR - but adding ~X min per PR could eventually push us past whatever timeout release-eng set on their side, and we'd have to figure out what's happening to adjust.
This matters more for minor bumps, where the script opens two PRs sequentially (one to add the new branch to pipeline.yml's DRA list on main, then a second to bump main to the next minor after cutting the release branch).

I can change this if you guys are sure this is preferable.

seanstory: Did they provide explicit guidance that CI should be skipped and that they have a timeout we need to worry about? If they did, I'll bow to your judgement. But if they haven't explicitly asked us to disable CI, I'd rather not.

markjhoy: > But if they haven't explicitly asked us to disable CI, I'd rather not.

➕ 1 to this... With automated processes, I'm all for more checks and safety

Apmats: Actually dug out the following guidance - in [PSI Version Bump Automation](https://docs.google.com/document/d/1VxKftPLqC3EtkxtjWpkbHqfP0Oen9142Piri-yviTVw/edit?usp=sharing)

```
*Team pipelines should aim to achieve within a maximum of 2–3 hours (including DRA builds), when full automation is in place, to ensure an efficient full version bump process
*Teams should ideally configure their version bump PRs to skip ITs and E2E tests to reduce the duration of the process; version bump PRs only increment the version number (displayed with --version flag), presenting low risk of introducing critical bugs, and if bugs are introduced, DRA artifacts can be published and fixed in a follow-up DRA build to unblock dependent teams
```

Running the tests X times as we do means we might end up pushing close to the suggested 2-3 hours. I know guidance says "ITs and E2E" but I think that's shorthand for "non-instant tests".

@markjhoy I don't think skipping tests gives up any real safety though - DRA publishing is gated on tests passing, it just means we will have to revert from a protected branch which is a hassle.

Updated thoughts? 😁

seanstory: Our builds on main reliably take about 30 min, that that includes a packaging phase that won't run on PRs. We should not be anywhere near the 2-3 hour mark. 

I expect that guidance is more applicable for monorepos like elasticsearch/cloud/kibana that have obscene numbers of tests.

.buildkite/publish/version-bump.sh
```diff
@@ -0,0 +1,303 @@
...
+# Adds a branch to the DRA branch list in pipeline.yml.
+# The list is the `if:` condition on the "Packaging and DRA" group step.
+add_branch_to_dra_list() {
+  local new_branch="$1"
+
+  echo "Adding ${new_branch} to DRA branch list in pipeline.yml"
+
+  python3 -c '
+import sys
```
artem-shelkovnikov: We can just make it a separate script file?

Also, we can theoretically make this whole script python if needed.

.buildkite/publish/version-bump.sh
Apmats: Had claude add a unit test for this then realized that it doesn't actually work because it needs to be discovered. Followed what you suggested. 👍

.buildkite/pipeline.yml
artem-shelkovnikov: Should this be a fully separate pipeline instead that calls non-conditionally all tests/ftests?

.buildkite/publish/version-bump.sh
markjhoy: ++ to this - might make it easier to maintain (at least the inline Python script here - I'm OK with the primary pipeline being in Bash)

.buildkite/pipeline.yml
seanstory: Two separate things.
One pipeline creates the version bump as a PR.

This generated PR runs our normal pipeline. And I maintain should run it _normally_.

Apmats: Alright- undid this.

I got Claude to look at 50 past PR and build runs on buildkite, run a simulation and give me p values for the total run time to run tests 2 or 3 times depending on the flow.


Scenario | Mean | P50 | P90 | P95 | P99
-- | -- | -- | -- | -- | --
Patch (PR + triggered build) | 48.8 | 47.3 | 65.6 | 73.8 | 89.0
Minor (PR1 + PR2 + triggered build) | 71.6 | 69.8 | 94.9 | 104.5 | 120.9

So for now we're mostly safely under 2 hours. 

If our times grow we can look at skipping tests at some point in this process while keeping them in the PR for revert-aversion 😁

.buildkite/publish/version-bump.sh
Apmats: Good call, Claude added a `version_is_downgrade` check here.

.buildkite/publish/version-bump.sh
Apmats: Done. Tests not added because I'm unsure when they would run and how. If you have ideas tell me.

.buildkite/publish/version-bump.sh
Apmats: I don't know if this is a stable thing historically, and if we want to encode it in the code.
I trust you know better, so if you tell me it makes sense I can do it, but I would myself try to contain the changes.

Apmats:
@markjhoy @seanstory @artem-shelkovnikov addressed most things I think - what we're not covering is this part of the RM duty on minor versions:

- Multiple pipeline schedules in catalog-info.yaml
- .backportrc.json
- renovate.json

Worth trying to take care of these as well?

seanstory:
Let's get this in for now, and we can tackle those later if necessary. I don't personally think that the RM duties are too heavy right now, the motivation for this is more about the release team not being blocked by our convenience, and none of those that you list matter to the release team, so they can still wait on our convenience.

Apmats:
@markjhoy @artem-shelkovnikov let me know if anything important remains unadddressed

markjhoy (approved):
👍 I think this is good for now. As @seanstory mentioned, we can address the other items later after we get this initial piece in.

Apmats:
Merging and moving to dry run tests - will coordinate for end to end later.

PR #3995, opened 2026-04-09
Fix SPO incremental sync skipping sites with unchanged timestamps

### LINKED ISSUE #3994: [SharePoint Online] Incremental sync skips sites by timestamp, silently dropping drive item deletions and modifications
## Bug Description

In `get_docs_incrementally`, the `sites()` call uses `check_timestamp=True`, which skips entire sites whose `lastModifiedDateTime` hasn't changed. When a site is skipped, its drive delta links are never consumed — so drive item deletions/modifications are silently lost.

The codebase already handles this pattern one level down:

```python
# Edit operation on a drive_item doesn't update the
# lastModifiedDateTime of the parent site_drive. Therefore, we
# set check_timestamp to False when iterating over site_drives.
async for site_drive in self.site_drives(site, check_timestamp=False):
```

The same problem exists one level up: child-resource changes don't reliably update the parent site's `lastModifiedDateTime` in the Graph API.

### To Reproduce

1. Run a full sync on a SharePoint Online connector.
2. Delete a document in SharePoint Online.
3. Run an incremental sync.
4. The deleted document still exists in the Elasticsearch index.

## Expected behavior

Incremental sync should detect all changes — including deletions — regardless of whether the parent site's timestamp was updated.

## Environment

- **Connector**: SharePoint Online (`sharepoint_online`)
- **Sync mode**: Incremental

## Additional context

This is a silent data consistency bug — no errors or warnings are logged. Deleted documents persist in the index until the next full sync, and stale content is served to users. Affects any deployment relying on incremental sync as steady-state.

kmitul:
## Closes https://github.com/elastic/connectors-py/issues/3994

In `get_docs_incrementally`, the `sites()` call used `check_timestamp=True`, filtering out sites whose `lastModifiedDateTime` hadn't changed. When a site was skipped, its drive delta links were never consumed — silently dropping drive item deletions and modifications.

This is the same class of bug already fixed for `site_drives`:

```python
# Edit operation on a drive_item doesn't update the
# lastModifiedDateTime of the parent site_drive. Therefore, we
# set check_timestamp to False when iterating over site_drives.
```

The fix applies the same pattern one level up: `check_timestamp=False` for `sites()` in incremental sync.

> **Question for maintainers:** `site_lists` also uses `check_timestamp=True` (line 756). If a list item is modified without updating the parent list's `lastModifiedDateTime`, the same class of silent data loss could occur. Should this be addressed here?

#### Pre-Review Checklist
- [x] this PR does NOT contain credentials of any kind, such as API keys or username/passwords
- [ ] Added a label for each target release version

N/A — one-line change to a boolean flag, no new dependencies or security implications.

## Release Note

SharePoint Online incremental sync could silently skip entire sites when their `lastModifiedDateTime` was unchanged, causing drive item deletions and modifications to be missed. Fixed by ensuring all sites are iterated during incremental sync, consistent with the existing behavior for site drives.


Apmats:
Hey, we really appreciate this contribution!
We will try to get to reviewing this eventually - but for now we don't want to merge without fully testing this out as we have experienced some weirdness regarding timestamp propagation in the past ourselves.

In the meantime, and just throwing this out there, we have built a SharePoint Online Connector that can fetch SPO data (although not into an index), that might or might not fit into what you're building - https://www.elastic.co/docs/reference/kibana/connectors-kibana/sharepoint-online-action-type - and we've been very actively iterating on that side of connectors so it might be a bit easier to incorporate feedback if this does end up fitting your use case.

seanstory:
Haven't dug deep into this. But my big concern here is that if we add this, we'll significantly reduce the performance of incremental syncs, by looking at the sites AND site drives of sites that have seen no changes, every time we run an incremental sync. @kmitul have you experimented any to see what the performance impact is?

We have several production customers who have been users of this connector for years, who haven't indicated that they're not seeing their deletions/modifications propagate, which is making me nervous that this is an uncommon situation we'd be solving, with a very common performance hit.

PR #4001, opened 2026-04-30
Support pre-monorepo branches in version-bump script

Apmats:
## Summary

Adds support for pre-9.3 release branches (8.19, 9.0, 9.1, 9.2) in the version-bump script introduced in #3986.

The original PR's description noted that pre-monorepo branches (single `connectors/VERSION` file at a different path) couldn't be supported as-is and would need a separate backport. That turns out not to be necessary: the script always runs from `main`'s checkout (the `connectors-version-bump` Buildkite pipeline uploads `version-bump-pipeline.yml` from `main`, regardless of the target release branch), so a single `case` on `${BRANCH}` in the on-`main` script is sufficient. No backport, no per-branch script copies.

I had originally assumed the script ran from each release branch's checkout, which is why #3986 left this as a follow-up. After verifying how CI actually invokes the script, this turned out to be a one-file change.

## Changes

In `.buildkite/publish/version-bump.sh`, replace the hardcoded `VERSION_FILES` array with a `case` on `${BRANCH}`:

- `8.*`, `9.0`, `9.1`, `9.2` → `connectors/VERSION` (pre-monorepo single-file layout)
- everything else → the two post-monorepo paths (current behavior, unchanged for 9.3+/main)

### DIFF
--- a/.buildkite/publish/version-bump.sh
+++ b/.buildkite/publish/version-bump.sh
@@ -37,10 +37,22 @@ fi
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 PROJECT_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
 
-VERSION_FILES=(
-  "${PROJECT_ROOT}/app/connectors_service/connectors/VERSION"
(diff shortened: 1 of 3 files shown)


lorenabalan (approved):
LGTM!

PR #4006, opened 2026-05-08
Fix DLS query to use .keyword instead of .enum

### LINKED ISSUE #4005: DLS query references non-existent .enum sub-field since 9.0
## Summary

The `DLS_QUERY` template in `access_control.py` references `_allow_access_control.enum`, a keyword sub-field that was created by the custom dynamic mapping template previously defined in `settings.py`. That mapping was removed in PR #3013 (v9.0.0), but the DLS query was not updated to match.

## Impact

For any connector content index created on 9.0+, Elasticsearch's default dynamic mapping maps `_allow_access_control` as `text` with a `.keyword` sub-field -- not `.enum`. The `terms` query against `_allow_access_control.enum` matches zero documents, silently filtering out **all DLS-protected content**. Users see no results for documents that have access control set.

**Existing indices** created before 9.0 are unaffected because the old dynamic template is baked into their index mapping.

## Reproduction

1. Set up any connector with DLS enabled on 9.0+
2. Run a content sync and an access control sync
3. Create an API key using the DLS query from the `.search-acl-filter-*` document
4. Search -- DLS-protected documents are missing from results

## Fix

Change `_allow_access_control.enum` to `_allow_access_control.keyword` in `DLS_QUERY` (`access_control.py`) and the corresponding test.

After deploying the fix, affected customers must re-run an **access control sync** so the corrected query template is written to their `.search-acl-filter-*` documents.

## Workaround

Customers can manually edit the DLS query when creating API key role descriptors, replacing `_allow_access_control.enum` with `_allow_access_control.keyword`.

## References

-  https://github.com/elastic/connectors/pull/3013

seanstory:
## Summary

- Fix DLS query template to use `_allow_access_control.keyword` instead of `_allow_access_control.enum`
- Update corresponding test assertion

The `.enum` sub-field was created by the custom dynamic mapping template in `settings.py`, which was removed in PR #3013 (v9.0.0). Under Elasticsearch's default dynamic mapping, string fields get a `.keyword` sub-field instead.

This caused DLS-protected documents to be silently filtered out for any content index created on 9.0+.

Closes #4005

## Test plan

- [ ] Existing unit test updated and passes
- [ ] Set up a connector with DLS on 9.0+, run content + access control sync, verify DLS queries return expected results

### Release Note

Fixes a bug for Connectors Document Level Security, where the generated query filter used an incorrect subfield

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -1906,7 +1906,7 @@ SOFTWARE.
 
 
 azure-core
-1.40.0
+1.41.0
 MIT
 Copyright (c) Microsoft Corporation.
 
@@ -4146,7 +4146,7 @@ Apache Software License
 
 
 google-auth
-2.50.0
+2.52.0
 Apache Software License
                                  Apache License
                            Version 2.0, January 2004
@@ -8213,7 +8213,7 @@ made under the terms of *both* these licenses.
 
 
 urllib3
-2.6.3
+2.7.0
 MIT
 MIT License
 
--- a/app/connectors_service/connectors/access_control.py
+++ b/app/connectors_service/connectors/access_control.py
@@ -19,7 +19,7 @@ DLS_QUERY = """{
                             },
                             {
                                 "terms": {
-                                    "_allow_access_control.enum": {{#toJson}}access_control{{/toJson}}
+                                    "_allow_access_control.keyword": {{#toJson}}access_control{{/toJson}}
                                 }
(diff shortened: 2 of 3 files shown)


artem-shelkovnikov (approved):

seanstory:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|9.3|[](https://github.com/elastic/connectors/pull/4008)|

### Questions ?

PR #4009, opened 2026-05-12
[Connectors Python] Adjust defaults to be more resilient

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/search-team/issues/14452

The original ask was to add Agentless-specific Elasticsearch defaults so content-heavy connectors (e.g. GMail) wouldn't lose data to bulk timeouts. Performance testing showed the new values are at least as good as the old ones for non-Agentless deployments, so this PR moves the tuning into the global `_default_config` instead of wrapping it inside `ConnectorsAgentConfigurationWrapper` following the agreement reached in https://elastic.slack.com/archives/C01795T48LQ/p1778690756537569 discussion. Self-managed, agent-managed, and Agentless deployments all get the same baseline.

The second issue (#14452) covers the same coupling from a different angle: `queue_refresh_timeout` was previously short enough that the mem-queue circuit breaker could fire before the bulk retry layer (`TransientElasticsearchRetrier.execute_with_retry`, called from `ESManagementClient.bulk_insert`) had finished its retry budget. The new default of 1300s comfortably outlasts the worst-case retry budget under default settings, and the comment / docs / sanity tests make the coupling explicit so future changes don't silently re-introduce the bug. Originally was supposed to be fixed in a **[separate PR](https://github.com/elastic/connectors/pull/4011)**, but with this PR now covering general defaults, it was decided to cover both issues with single PR.

### New defaults

| Setting | Old | New |
|---|---|---|
| `elasticsearch.request_timeout` | 120 | **240** |
| `elasticsearch.max_wait_duration` | 120 | **240** |
| `elasticsearch.bulk.queue_refresh_timeout` | 600 | **1300** |
| `elasticsearch.bulk.chunk_size` | 1000 | **500** |
| `elasticsearch.bulk.chunk_max_mem_size` | 5 | **3** |

`queue_refresh_timeout = 1300` is sized for 5 attempts × 240s `request_timeout` + linear backoff (10 + 20 + 30 + 40s) = 1300s. If any of those upstream values change, `queue_refresh_timeout` must change with them; this is now documented in `config.yml.example` and on the `DEFAULT_BULK_QUEUE_REFRESH_TIMEOUT` constant.

### Changes

- `connectors/config.py`
  - Updated five values in `_default_config()` (table above).
  - Added `DEFAULT_BULK_QUEUE_REFRESH_TIMEOUT = 1300` with a comment explaining how it's derived and why it must move together with `request_timeout` / `max_retries` / `retry_interval`.
- `connectors/es/sink.py`
  - Bulk fallback `queue_refresh_timeout` 60 → 1300 (matches the new default). https://github.com/elastic/connectors/pull/4011#discussion_r3235071864
- `connectors/agent/config.py`
  - Removed the now-redundant Agentless-only `elasticsearch` block from `_default_config` — the global `_default_config` now provides those values.
  - Removed the partial-update guard in `_elasticsearch_config_changed()`; it was only needed to avoid spurious diffs against the wrapper-side ES defaults, which no longer exist.
- `config.yml.example`
  - Updated the commented example values to match the new defaults.
  - Added a note under `queue_refresh_timeout` calling out that it must be raised together with `request_timeout` / `max_retries` / `retry_interval` if a user overrides those.
- `tests/test_config.py`
  - Replaced the Agentless-specific assertions with property-based sanity tests that lock in invariants of the new defaults:
    - `queue_refresh_timeout` outlives the worst-case `TransientElasticsearchRetrier` budget.
    - Bulk chunk / queue / concurrency knobs stay in a sane range.
    - No single retried operation can stall for more than an hour.
    - The error-monitor sliding window can actually observe the consecutive-error threshold.
    - Service polling / heartbeat / cleanup intervals respect their natural ordering.

## Performance Testing

Testing performed on MEDIUM dataset (elasticsearch repo) on local machine with ElasticSearch being set to `ES_JAVA_OPTS=-Xms1g -Xmx1g` (default is `ES_JAVA_OPTS=-Xms1g -Xmx1g`)
>DATA_SIZE=medium MAX_DURATION=1800 REFRESH_RATE=2 make -C app/connectors_service ftest NAME=dir PERF8=yes

BASELINE VS TARGET

Memory Usage
rss rss

CPU
cpu cpu

Threads
threads threads

Disk I/O
disk_io disk_io

Disk Usage
disk disk

Event Loop Lag
loop_lag loop_lag

Task Concurrency
loop_coro loop_coro

TARGET shows overall better performance than BASELINE across the measured metrics, which is the basis for promoting these values from an Agentless-only override to the global default — they are safe (and beneficial) for self-managed and agent-managed deployments too.

#### Pre-Review Checklist

## Release Note

Tune Elasticsearch ingestion defaults to better protect content-heavy connectors against bulk timeouts: increase `elasticsearch.request_timeout` and `elasticsearch.max_wait_duration` to 240s, raise `elasticsearch.bulk.queue_refresh_timeout` to 1300s so the mem-queue circuit breaker outlasts the bulk retry budget, and reduce `elasticsearch.bulk.chunk_size` / `elasticsearch.bulk.chunk_max_mem_size` for steadier per-request memory use. Applies to all deployments (self-managed, agent-managed, and Agentless).


Jan-Kazlouski-elastic:
@coderabbitai full review

artem-shelkovnikov:
Left a question in https://elastic.slack.com/archives/C01795T48LQ/p1778690756537569 - from my perspective looks good, but want a second opinion.

seanstory (approved):
LGTM

Jan-Kazlouski-elastic:
Following the results of the discussion in https://elastic.slack.com/archives/C01795T48LQ/p1778690756537569
Applied new defaults for general configs, not only agentless.

app/connectors_service/connectors/config.py
```diff
@@ -89,8 +96,8 @@ def _default_config():
             "max_retries": DEFAULT_ELASTICSEARCH_MAX_RETRIES,
             "retry_interval": DEFAULT_ELASTICSEARCH_RETRY_INTERVAL,
             "retry_on_timeout": True,
-            "request_timeout": 120,
-            "max_wait_duration": 120,
+            "request_timeout": 240,
+            "max_wait_duration": 240,
             "initial_backoff_duration": 1,
```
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: PR description is fixed now.

Jan-Kazlouski-elastic:
Could you please give this PR another look @seanstory ?
Now it covers general defaults instead of agentless, also has some minor fixes.

app/connectors_service/connectors/config.py
```diff
@@ -68,11 +75,11 @@ def _default_config():
...
                 "queue_max_mem_size": 25,
                 "queue_refresh_interval": 1,
-                "queue_refresh_timeout": 600,
+                "queue_refresh_timeout": DEFAULT_BULK_QUEUE_REFRESH_TIMEOUT,
                 "display_every": 100,
-                "chunk_size": 1000,
+                "chunk_size": 500,
                 "max_concurrency": 5,
-                "chunk_max_mem_size": 5,
+                "chunk_max_mem_size": 3,
```
seanstory: These now disagree with constants we have in utils.py:

```
  DEFAULT_CHUNK_SIZE = 500       # matches new config.py ✓
  DEFAULT_QUEUE_MEM_SIZE = 5     # but config.py now uses 25 for queue_max_mem_size ✗
  DEFAULT_CHUNK_MEM_SIZE = 25    # but config.py now uses 3 for chunk_max_mem_size ✗
```

They weren't coupled before, but now that the values are different, this feels extra problematic. What would you think about defining these constants/defaults either here in `config.py`, or creating a `constants.py` that `utils.py` and `config.py` can both depend upon?

seanstory (approved):
Changes lgtm, but I noticed one thing I'd like to address.

app/connectors_service/connectors/config.py
Jan-Kazlouski-elastic: That makes a lot of sense to define them in single file instead of spreading into two. That is something that I considered as well, but wasn't sure if we'd want to bring more changes into this PR instead of raising a separate one.. Will make the change.

Jan-Kazlouski-elastic: Done.

PR #4011, opened 2026-05-12
[Connectors Python] Align queue_refresh_timeout with Sink's full retry budget

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/search-team/issues/14452

Raises the default `elasticsearch.bulk.queue_refresh_timeout` so that it can never expire before the underlying bulk request itself has exhausted all of its retries. With the previous value (`600s`), the queue circuit-breaker could fire while a legitimate bulk request was still retrying, killing the sync prematurely.

The new default is derived from the worst-case bulk timing:

`5 attempts × 120s (request_timeout) + (10 + 20 + 30 + 40)s backoff = 700s`

### Changes

- `connectors/config.py`
  - New `DEFAULT_BULK_QUEUE_REFRESH_TIMEOUT = 700` constant with a comment explaining the derivation.
  - `elasticsearch.bulk.queue_refresh_timeout` default raised from `600` → `700` (now sourced from the constant).
  - Tiny cleanup: fix `10MB` comment to the correct `10MiB`.
- `config.yml.example`
  - Updated the commented-out example and accompanying explanation to reflect the new `700s` default.
- `tests/test_config.py`
  - New `test_default_config_protects_critical_values` snapshot test that pins every critical production default (`elasticsearch`, `elasticsearch.bulk`, `elasticsearch.bulk.error_monitor`, `service`). Acts as a tripwire so any future change to a production-shaping default has to be made deliberately.

No behavior change for users who have explicitly overridden `queue_refresh_timeout` in `config.yml`.

#### Pre-Review Checklist
- [X] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)
- [X] For bugfixes: backport safely to all minor branches still receiving patch releases

## Release Note

[Optional] Increase the default `elasticsearch.bulk.queue_refresh_timeout` from `600s` to `700s` so the queue circuit-breaker no longer fires before bulk requests have finished their retry cycle, preventing spurious sync failures.


app/connectors_service/tests/test_config.py
```diff
@@ -80,3 +80,56 @@ def test_nest_config_when_root_field_does_exists():
     _nest_configs(config, "test", 50)
 
     assert config["test"] == 50
+
+
+def test_default_config_protects_critical_values():
```
Jan-Kazlouski-elastic: I can see an argument being made that such test:

- wouldn't protect defaults from being infested with new incorrect values 
- requires knowledge from a person making the change
- doesn't bring a lot of value

Perhaps decision of not having such test was made in the past. However I think that it is worth having at least this kind of protection for the critical default values so they are not changed involuntarily.

app/connectors_service/config.yml.example
```diff
@@ -120,10 +120,10 @@
 #
 #
 ##  Maximal interval of time during which MemQueue does not dequeue a single document
-##  For example, if no documents were sent to Elasticsearch within 60 seconds because of
+##  For example, if no documents were sent to Elasticsearch within 700 seconds because of
 ##  Elasticsearch being overloaded, then an error will be raised.
 ##  This mechanism exists to be a circuit-breaker for stuck jobs and stuck Elasticsearch.
-#elasticsearch.bulk.queue_refresh_timeout: 60
```
Jan-Kazlouski-elastic, line 126: Having a value in the example being set to actual default makes more sense than having it set to fallback value. Having `queue_refresh_timeout: 60` with `elasticsearch.request_timeout: 120` next to it doesn't make much sense to me

Jan-Kazlouski-elastic:
@coderabbitai full review

app/connectors_service/tests/test_config.py
artem-shelkovnikov: I honestly don't like this test for the reason you outlined.

I feel more value would be to do what you did - verify that some limits interact correctly - e.g. sink's timeout is bigger than sum of all retries with a timeout and other sanity checks.

I could think of:
1. Sink's timeout as I mentioned
2. Bulk chunk sizes are sane (whatever sane means - at least 250?)
3. Retry timeouts cannot be leading to retry that takes days
4. Error monitor window size is bigger than max_consecutive_errors
5. Anything else?

app/connectors_service/config.yml.example
Jan-Kazlouski-elastic: @artem-shelkovnikov 
Also, with 
>mem_queue_refresh_timeout = options.get("queue_refresh_timeout", 60)

Perhaps it would be good idea to unify the defaults? Including both example, actual default and fallback? Having 700 in one place and 60 in the other feels a bit off.

artem-shelkovnikov: 60 is a mistake, good catch.

app/connectors_service/tests/test_config.py
Jan-Kazlouski-elastic: Done.

app/connectors_service/config.yml.example
Jan-Kazlouski-elastic: Fixed.

app/connectors_service/connectors/config.py
```diff
@@ -11,8 +11,11 @@
 
 DEFAULT_ELASTICSEARCH_MAX_RETRIES = 5
 DEFAULT_ELASTICSEARCH_RETRY_INTERVAL = 10
+# Must outlast the worst-case bulk request: 5 attempts × 120s request_timeout,
+# plus 10 + 20 + 30 + 40s of retry backoff between them = 700s.
+DEFAULT_BULK_QUEUE_REFRESH_TIMEOUT = 700
```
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: Now 60s is replaced with 700s. Fixed.

Jan-Kazlouski-elastic:
@artem-shelkovnikov Thank you for your comments. They are addressed and PR is ready for another round of review.

Jan-Kazlouski-elastic:
Closing this PR in favor of https://github.com/elastic/connectors/pull/4009, because it now touches general defaults not agentless-only ones.

app/connectors_service/connectors/es/sink.py
```diff
@@ -1022,7 +1022,7 @@ async def async_bulk(
         retry_interval = options.get(
             "retry_interval", DEFAULT_ELASTICSEARCH_RETRY_INTERVAL
         )
-        mem_queue_refresh_timeout = options.get("queue_refresh_timeout", 60)
+        mem_queue_refresh_timeout = options.get("queue_refresh_timeout", 700)
         mem_queue_refresh_interval = options.get("queue_refresh_interval", 1)
```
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: Fixed in #4009

app/connectors_service/tests/test_config.py
```diff
@@ -80,3 +80,106 @@ def test_nest_config_when_root_field_does_exists():
...
+
+
+def test_bulk_queue_refresh_timeout_outlasts_worst_case_retry_budget():
+    """The mem-queue refresh timeout must outlive a fully retried bulk request.
+
+    Bulk operations are wrapped with `retryable` using LINEAR_BACKOFF, so the worst-case
+    duration of a single bulk send is `max_retries` request attempts (each capped by
+    `request_timeout`) plus the linear backoff sleeps between them. If
+    `queue_refresh_timeout` is shorter than that budget, the queue gives up on a request
+    that the retry layer is still legitimately working on, and we lose data.
```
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: Fixed in https://github.com/elastic/connectors/pull/4009

app/connectors_service/connectors/config.py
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: Fixed in https://github.com/elastic/connectors/pull/4009

PR #4031, opened 2026-05-19
[Connectors Python] Trim gmail messages to body and minimal headers before indexing

### LINKED ISSUE #1369: [Gmail Connector] Email cleaning
### Problem Description

Currently the connector is extracting the entire raw email and this is in most cases insufficient for correct usage, specially when adding ML pipelines:

image

After some clarification from @seanstory and @danajuratoni I can know that parsing emails is not trivial, but the expectation is see just text, like a web crawler. 

### Proposed Solution

This said we can only offer a best-effort approach to parse the text because of the infinite ways of building an email.

@seanstory proposed using the attachment processor to send the body as .eml file and let tika do the extraction which I like because it aligns with the stack. (Crawlers "ingest full HTML option")

We could make this the default behavior with the option to disable it to support edge cases, and add documentation about how to handle those edge cases.

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/connectors/issues/1369

The Gmail connector currently ships the entire raw RFC 822 message in `_attachment` — every routing/auth header, both `text/plain` and `text/html` alternatives, and every binary part (inline images, PDFs, etc.). This is wasteful on bandwidth, storage, and Tika cycles, and it produces noisy indexed text in `attachment.content` whenever Tika's RFC 822 detection falls back to its `TextParser` (which dumps the raw bytes verbatim — this is the failure mode in the reported issue's screenshot).

This PR trims the message in the connector, before it ever reaches Elasticsearch. Using Python's stdlib `email` module we parse the raw Gmail payload, rebuild a minimal `.eml` containing only the headers and the body part we actually want indexed, and forward that as `_attachment`.

**Kept headers** (matches Tika's default `RFC822Parser` content output, plus `Message-ID` for dedup): `Subject`, `From`, `Reply-To`, `To`, `Cc`, `Bcc`, `Date`, `Message-ID`. Everything else (`Received`, `ARC-*`, `DKIM-Signature`, `Return-Path`, `Authentication-Results`, `List-*`, `X-*`, …) is dropped.

**Body selection**: `text/plain` preferred over `text/html` via `EmailMessage.get_body(preferencelist=("plain", "html"))`. All binary parts (inline images, attachments, signed/encrypted blobs) are dropped — Tika would not produce useful indexed text from them anyway, and the ES attachment processor doesn't run OCR by default.

**New toggle** `include_full_raw_message` (default `False`, order 5) restores the previous full-raw passthrough for edge cases where body extraction misses content.

**Safety**: any exception inside the trim helper returns `None` and the caller falls back to the legacy `base64url_to_base64` raw passthrough with a logged warning. No path can drop a message because of parse failure.

## Smoke testing

Validated against a real Gmail message (`/tmp/sample.eml`, 695 KB — a Google Groups welcome email with two inline PNGs) via two ad-hoc scripts (`scripts/smoke_gmail_trim.py` and `scripts/smoke_gmail_tika.sh`).

**Wire payload (`_attachment`)** — what the connector ships to Elasticsearch:

| | `before.eml` (legacy) | `after.eml` (this branch) |
|---|---:|---:|
| size | 695,427 B | 2,722 B |
| reduction | — | **99.61 %** |

Bytes dropped by the trim, by MIME part:

| part | size | kept? |
|---|---:|:---:|
| `text/plain` (body) | 2,267 B | yes |
| `text/html` (same body, marked-up) | 6,961 B | no |
| `image/png` inline (logo) | 1,434 B | no |
| `image/png` inline (photo) | 488,143 B | no |
| routing/auth/`X-*`/`List-*` headers | ~196 KB | no |

**Tika output (`attachment.content`)** — what ends up in the indexed text after the ES attachment processor runs (apache/tika:latest, content-type auto-detected as `message/rfc822` in both cases here):

- `before.txt`: 2,293 B — HTML-derived flattened text. Loses sender's email address, stripped because it was inside an HTML anchor), collapses paragraph breaks, mangles signature spacing.
- `after.txt`: 2,313 B — proper plain-text rendering. Email addresses preserved inline, line breaks intact, signature block readable.

The bytes are similar because this email happens to hit Tika's happy path; the qualitative formatting win is what's visible in the diff. On emails where Tika misdetects (the reported issue's screenshot is the canonical example), trimming reduces `attachment.content` from tens of KB of `Received:`/`ARC-Seal:`/base64 image blobs down to the body only.

Also covered by unit tests in `tests/sources/test_gmail.py` (44 passing): 7 parametrized fixtures (plain-only, multipart/alternative, html-only, multipart/mixed with PDF attachment, multipart/related with inline image, RFC 2047 encoded subject, Windows-1252 body), DSN with no textual body, malformed-base64 garbage, fallback when extraction returns `None`, toggle-on legacy passthrough, empty/`None` input.

#### Pre-Review Checklist

**Behavior change**: existing Gmail indices will start receiving smaller, body-only `_attachment` payloads after upgrade. Routing/auth headers and binary parts that were previously visible in `attachment.content` will no longer appear. Users who depended on that content can restore it by enabling `include_full_raw_message`.

## Release Note

Gmail connector: by default, only the email body (preferring `text/plain` over `text/html`) and a small set of headers (`Subject`, `From`, `Reply-To`, `To`, `Cc`, `Bcc`, `Date`, `Message-ID`) are indexed; routing/authentication headers and binary attachments are dropped. Reduces `_attachment` payload size by ~90 %+ on typical mail and produces cleaner indexed text. Set the new **Index full raw email (including headers)** toggle in the connector configuration to restore the previous behavior.

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -91,7 +91,7 @@ SOFTWARE.
 
 
 PyMySQL
-1.1.3
+1.2.0
 MIT
 Copyright (c) 2010, 2013 PyMySQL contributors
 
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -1044,7 +1044,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
 aiohappyeyeballs
-2.6.1
+2.6.2
 Python Software Foundation License
 A. HISTORY OF THE SOFTWARE
 ==========================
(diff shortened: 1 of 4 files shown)


app/connectors_service/NOTICE.txt
```diff
@@ -91,7 +91,7 @@ SOFTWARE.
 
 
 PyMySQL
-1.1.3
+1.2.0
```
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: This change is produced by local build. Ignore it.

app/connectors_service/tests/sources/test_gmail.py
```diff
@@ -136,36 +353,207 @@ async def patch_google_directory_client(self):
             client = mock.return_value
             yield client
 
-    # @pytest.mark.asyncio
```
Jan-Kazlouski-elastic, line 139: Not sure why was this commented out. Enabling it doesn't cause test to fail.

app/connectors_service/connectors/sources/gmail/datasource.py
```diff
@@ -29,6 +34,66 @@
...
+_KEPT_HEADERS = (
+    "Subject",
+    "From",
+    "Reply-To",
+    "To",
+    "Cc",
+    "Bcc",
+    "Date",
+    "Message-ID",
+)
```
seanstory, line 47: For consistency, should we be applying similar logic to the Outlook connector, too?

app/connectors_service/connectors/sources/gmail/datasource.py
```diff
@@ -83,10 +148,23 @@ def get_default_configuration(cls):
                 "type": "bool",
                 "value": False,
             },
+            "include_full_raw_message": {
```
seanstory, line 151: This will need a corresponding documentation PR. See https://github.com/elastic/elasticsearch/blob/main/docs/reference/search-connectors/es-connectors-gmail.md

app/connectors_service/connectors/sources/gmail/datasource.py
```diff
@@ -83,10 +148,23 @@ def get_default_configuration(cls):
...
+                "label": "Index full raw email (including headers)",
+                "order": 5,
+                "tooltip": (
+                    "When disabled (default), only the email body is indexed. "
+                    "Enable to keep the full raw message including routing and "
+                    "authentication headers - useful for edge cases where body "
+                    "extraction misses content."
+                ),
+                "type": "bool",
+                "value": False,
```
seanstory, line 162: I'm wondering if we should actually default this to True, to avoid any backwards compatibility issues. I agree that `False` is the better general experience, though.

Have you chatted with @deeptidheer-dotcom about this new behavior?

app/connectors_service/connectors/sources/gmail/datasource.py
Jan-Kazlouski-elastic: We could, but since original issue is related strictly to GMAIL, I believe Outlook change must be done as part of different PR. Will create an issue for it!

app/connectors_service/connectors/sources/gmail/datasource.py
Jan-Kazlouski-elastic: To be honest - I agree that having it disabled by default is a safer option... but considering existing practice of having opt-out configs + issue description having this line in proposal section:

>We could make this the default behavior with the option to disable it to support edge cases, and add documentation about how to handle those edge cases.

I believe we should be ok.

I haven't chatted with @deeptidheer-dotcom .

seanstory: > this line in proposal section

This is gonna be a tricky thing to build an instinct for, but we can't always assume that whatever's in the issue is the actual product requirement. In this instance, a director from our Support organization wrote the issue. It's from his perspective of what he'd like to see. But our Product Managers (Deepti in this case) are the ones who actually make these decisions.

Jan-Kazlouski-elastic: >whatever's in the issue is the actual product requirement

I totally get it and it is wasn't my assumption that it is an actual product _requirement_, just an opinion that is in line with my vision.

app/connectors_service/connectors/sources/gmail/datasource.py
Jan-Kazlouski-elastic: https://github.com/elastic/connectors/issues/4033

Once this PR is merged, I'll update the description. For now it is intentionally minimalistic.

app/connectors_service/connectors/sources/gmail/datasource.py
Jan-Kazlouski-elastic: It was agreed to keep the default value as False in **[slack discussion](https://elastic.slack.com/archives/C01795T48LQ/p1779229489765549?thread_ts=1779211630.480739&cid=C01795T48LQ)**.

app/connectors_service/connectors/sources/gmail/datasource.py
Jan-Kazlouski-elastic: Documentation PR: https://github.com/elastic/elasticsearch/pull/149510

seanstory (approved):
LGTM

PR #4032, opened 2026-05-19
[Connectors Python] Fix PostgreSQL timestamp issue

Jan-Kazlouski-elastic:
## Was created for https://github.com/elastic/sdh-search/issues/1882

After enabling `track_commit_timestamp = on` on the PostgreSQL server (per the [tutorial](https://www.elastic.co/docs/reference/search-connectors/es-postgresql-connector-client-tutorial#es-postgresql-connector-client-tutorial-postgresql-prerequisites)), customers were still seeing the entire table re-indexed on every sync instead of getting incremental behavior (to be confirmed by support person).

### Root cause

The connector runs `SELECT MAX(pg_xact_commit_timestamp(xmin)) FROM <table>` to derive a per-table `_timestamp`. PostgreSQL only records commit timestamps for transactions that commit *after* `track_commit_timestamp` is enabled, so for pre-existing rows `pg_xact_commit_timestamp(xmin)` returns `NULL`. If no DML has hit the table since enabling the setting, `MAX(...)` is `NULL` too — and the connector fell back to `iso_utc()`:

```python
timestamp=last_update_time or iso_utc(),
```

`iso_utc()` returns the current time on every sync, so `existing._timestamp != incoming._timestamp` always, and the sink re-indexes every document on every sync.

### Fix

The connector now distinguishes the three real-world states explicitly, by querying `SHOW track_commit_timestamp` (new helper `is_track_commit_timestamp_enabled` on `PostgreSQLClient`):

| `track_commit_timestamp` | `MAX(pg_xact_commit_timestamp(xmin))` | `_timestamp` emitted | Sync behavior |
|---|---|---|---|
| `on`, recent DML | real timestamp | that timestamp | incremental dedup (unchanged) |
| `on`, no DML since enabling **or** query raises | `NULL` | `None` (stable) | one-time reindex on upgrade, then dedup works |
| `off` | `NULL` | `iso_utc()` (fresh per sync) | full reindex every sync — matches the documented contract |

The `track_commit_timestamp = off` branch is critical: returning a stable `null` there would let the sink skip `null == null` between syncs and **silently drop row updates**, since the database has no change-detection signal at all. The `iso_utc()` fallback is preserved exactly for that case.

Additional fixes carried with this PR (all strict improvements, no behavioral risk):

- Filter `None` out of `last_update_times` before `max()` to avoid a latent `TypeError` in `_yield_docs_custom_query` when one table returns `None` and another a string.
- Promote the exception-path warning to an f-string so the table name is actually interpolated (was logged literally as `{table}`).
- New diagnostic warnings tell operators exactly what state was observed: *"`track_commit_timestamp` is on but no DML since"* vs *"`track_commit_timestamp` is off; reindexing on every sync"*.
- Centralize the per-table timestamp decision in a single helper (`_resolve_table_timestamp`) shared by the standard and advanced-rules paths so the policy can't drift between them.

### Backward compatibility

- Tables where `MAX(pg_xact_commit_timestamp(xmin))` returns a real timestamp are unaffected (identical input/output).
- Customers running with `track_commit_timestamp = off` keep the existing behavior (full reindex every sync, as documented). No regression, no silent data loss.
- Indices polluted by the previous `iso_utc()` churn while `track_commit_timestamp = on` pay a **one-time** reindex on the first post-upgrade sync (existing iso_utc strings overwritten with `null`), then stabilize.
- `_timestamp` is a `date` field in ES; storing `null` is supported and does not require any mapping changes. The framework's read path (`yield_existing_documents_metadata`) already returns `None` for missing/`null` values; the sink's compare-and-skip path uses `==`, so `None == None → skip`.
- No SDK or sink contract is changed (verified against `connectors/es/sink.py` lines 651-677, 778-779, 837-852: the `if TIMESTAMP_FIELD not in doc: doc[TIMESTAMP_FIELD] = iso_utc()` fallback uses `not in`, so a present-with-`None` value is preserved).
- Other connectors (e.g. MSSQL) use a different mechanism for last-update detection and are intentionally out of scope.

### Edge cases verified

Traced explicitly: pre-existing index + setting just enabled + no DML; new row inserted between syncs; row deletion; setting toggled `on → off` between syncs; `off → on` between syncs; `SHOW track_commit_timestamp` query failing or returning unexpected values; mid-sync flips; fresh empty index; multi-table advanced rules with mixed update states.

In every case the connector either produces correct incremental behavior or — when the database can give us no change-detection signal — falls back to "reindex everything" (the documented contract).

#### Pre-Review Checklist
- [x] Added a label for each target release version

Note for reviewers: customers running the buggy code today with `track_commit_timestamp = on` will see a single full re-index on the first sync after upgrade for any table whose `MAX(pg_xact_commit_timestamp(xmin))` returns `NULL`. After that one sync, syncs with no DML will correctly skip every document. Customers running with `track_commit_timestamp = off` see no behavior change.

## Related Pull Requests

_None._

## Release Note

Fix PostgreSQL connector re-indexing all documents on every sync after enabling `track_commit_timestamp = on`. The connector now reads `SHOW track_commit_timestamp` to pick the correct policy: when the setting is on, documents from rows whose commit timestamp is unknown carry a stable null `_timestamp` so the Elasticsearch sink can skip unchanged documents on subsequent syncs; when the setting is off, the connector continues to reindex on every sync as documented (avoiding any silent loss of row updates).

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -91,7 +91,7 @@ SOFTWARE.
 
 
 PyMySQL
-1.1.3
+1.2.0
 MIT
 Copyright (c) 2010, 2013 PyMySQL contributors
 
--- a/app/connectors_service/connectors/sources/postgresql/client.py
+++ b/app/connectors_service/connectors/sources/postgresql/client.py
@@ -95,6 +95,32 @@ class PostgreSQLClient:
             )
         )
 
+    async def is_track_commit_timestamp_enabled(self):
+        """Return True iff `track_commit_timestamp` is `on` on the server.
+
+        When this setting is off, `pg_xact_commit_timestamp(xmin)` returns NULL for
+        every row, so the connector has no signal for detecting row-level changes
+        and must fall back to reindexing on every sync.
+        """
+        try:
+            [value] = await anext(
+                fetch(
+                    cursor_func=partial(
+                        self.get_cursor,
+                        self.queries.track_commit_timestamp_setting(),
+                    ),
+                    fetch_size=1,
+                    retry_count=self.retry_count,
+                )
+            )
+        except Exception as exception:
+            self._logger.warning(
+                f"Could not read 'track_commit_timestamp' setting: {exception}. "
+                "Assuming it is off and reindexing on every sync."
+            )
+            return False
+        return str(value).strip().lower() == "on"
+
     async def get_tables_to_fetch(self, is_filtering=False):
         tables = configured_tables(self.tables)
         if is_wildcard(tables) or is_filtering:
--- a/app/connectors_service/connectors/sources/postgresql/datasource.py
+++ b/app/connectors_service/connectors/sources/postgresql/datasource.py
@@ -25,6 +25,7 @@ from asyncpg.types import (
     Polygon,
 )
 from connectors_sdk.source import BaseDataSource
+from connectors_sdk.utils import iso_utc
 from sqlalchemy.exc import ProgrammingError
 
 from connectors.sources.postgresql.client import PostgreSQLClient
@@ -242,6 +243,36 @@ class PostgreSQLDataSource(BaseDataSource):
         )
         return row
 
+    def _resolve_table_timestamp(self, last_update_times, track_enabled, label):
+        """Pick the per-table `_timestamp` for the documents about to be emitted.
+
+        - If `MAX(pg_xact_commit_timestamp(xmin))` returned a real value, use it.
+        - If it returned NULL but `track_commit_timestamp` is on, return None: this
+          is a stable value, so the sink will dedup unchanged documents on later
+          syncs. Once any DML happens, MAX flips to a real timestamp and the table
+          gets reindexed.
+        - If `track_commit_timestamp` is off, fall back to `iso_utc()`. We have no
+          way to detect row-level changes, so we must reindex on every sync (this
+          matches the behavior the docs promise when the setting is off).
+        """
+        if last_update_times:
+            return max(last_update_times)
+        if track_enabled:
+            self._logger.warning(
+                f"No commit timestamp available for {label}. "
+                "'track_commit_timestamp' is on but no transactions have committed "
+                "since it was enabled - documents will use a stable null timestamp "
+                "until then."
+            )
+            return None
+        self._logger.warning(
+            f"'track_commit_timestamp' is off; cannot detect row-level changes for "
+            f"{label}. All documents will be reindexed on every sync. Enable "
+            "'track_commit_timestamp = on' on the PostgreSQL server to get "
+            "incremental syncs."
+        )
+        return iso_utc()
+
     async def get_primary_key(self, tables):
         self._logger.debug(f"Extracting primary keys for tables: {tables}")
         primary_key_columns = []
@@ -312,6 +343,8 @@ class PostgreSQLDataSource(BaseDataSource):
             )
             return
 
+        track_enabled = await self.postgresql_client.is_track_commit_timestamp_enabled()
+
         last_update_times = []
         for table in tables:
             try:
@@ -326,16 +359,11 @@ class PostgreSQLDataSource(BaseDataSource):
             if last_update_time is not None:
                 last_update_times.append(last_update_time)
 
-        # `MAX(pg_xact_commit_timestamp(xmin))` is NULL when `track_commit_timestamp`
-        # is off or no transactions have committed since enabling it. Keep `_timestamp`
-        # stable (None) - falling back to `iso_utc()` would force a reindex every sync.
-        last_update_time = max(last_update_times) if last_update_times else None
-        if last_update_time is None:
-            self._logger.warning(
-                f"No commit timestamp available for tables: {', '.join(tables)}. "
-                "Ensure 'track_commit_timestamp = on' is set on the PostgreSQL server "
-                "and that data has been committed since enabling it."
-            )
+        last_update_time = self._resolve_table_timestamp(
+            last_update_times=last_update_times,
+            track_enabled=track_enabled,
+            label=f"tables: {', '.join(tables)}",
+        )
 
         async for row in self.yield_rows_for_query(
             primary_key_columns=primary_key_columns, tables=tables, query=query
@@ -358,6 +386,9 @@ class PostgreSQLDataSource(BaseDataSource):
             self._logger.debug(f"Total '{row_count}' rows found in table '{table}'")
             keys, order_by_columns = await self.get_primary_key(tables=[table])
             if keys:
+                track_enabled = (
+                    await self.postgresql_client.is_track_commit_timestamp_enabled()
+                )
                 try:
                     last_update_time = (
                         await self.postgresql_client.get_table_last_update_time(
@@ -369,17 +400,13 @@ class PostgreSQLDataSource(BaseDataSource):
                         f"Unable to fetch last_updated_time for table '{table}'"
                     )
                     last_update_time = None
-                # `MAX(pg_xact_commit_timestamp(xmin))` is NULL when `track_commit_timestamp`
-                # is off or no transactions have committed since enabling it. Keep
-                # `_timestamp` stable (None) - falling back to `iso_utc()` would force a
-                # reindex every sync.
-                if last_update_time is None:
-                    self._logger.warning(
-                        f"No commit timestamp available for table '{table}'. "
-                        "Ensure 'track_commit_timestamp = on' is set on the PostgreSQL "
-                        "server and that data has been committed in this table since "
-                        "enabling it."
-                    )
+                last_update_time = self._resolve_table_timestamp(
+                    last_update_times=[last_update_time]
+                    if last_update_time is not None
+                    else [],
+                    track_enabled=track_enabled,
+                    label=f"table '{table}'",
+                )
                 async for row in self.yield_rows_for_query(
                     primary_key_columns=keys,
                     tables=[table],
--- a/app/connectors_service/connectors/sources/postgresql/datasource.py
+++ b/app/connectors_service/connectors/sources/postgresql/datasource.py
@@ -25,7 +25,6 @@ from asyncpg.types import (
     Polygon,
 )
 from connectors_sdk.source import BaseDataSource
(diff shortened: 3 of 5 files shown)


Jan-Kazlouski-elastic:
## Testing

### Unit tests

`app/connectors_service/tests/sources/test_postgresql.py` pins each of the four `_timestamp` decision states:

- `track_commit_timestamp = on`, `MAX(...)` returns a real value → `_timestamp = MAX` (existing test).
- `track_commit_timestamp = on`, `MAX(...) IS NULL` → `_timestamp = None` (stable across syncs).
- `track_commit_timestamp = on`, `get_table_last_update_time` raises → `_timestamp = None`.
- `track_commit_timestamp = off`, `MAX(...) IS NULL` → `_timestamp = iso_utc()` (full reindex every sync, matches the documented contract).
- Same `None` guarantee for the advanced-rules / custom-query path.

```
$ make test
============================== 52 passed in 0.83s ==============================
```

### End-to-end reproduction against real PostgreSQL

A standalone repro script at `app/connectors_service/scripts/repro_track_commit_timestamp.py` reproduces the customer's exact scenario against a real PostgreSQL 14 server: it flips `track_commit_timestamp` off, loads rows, flips back to on with a server restart, asserts `MAX(pg_xact_commit_timestamp(xmin)) IS NULL`, then runs `PostgreSQLDataSource.get_docs()` twice with no DML between calls and asserts every document's `_timestamp` is identical and equal to `None` across the two runs.

**Setup:**
```bash
docker run -d --name pg-repro -p 9090:5432 \
  -e POSTGRES_USER=admin -e POSTGRES_PASSWORD=Password_123 -e POSTGRES_DB=xe \
  postgres:14.1-alpine
cd app/connectors_service && PYTHONPATH=. .venv/bin/python scripts/repro_track_commit_timestamp.py
```

**With the fix (delivered behavior):**
```
Step 4: verify bug condition (MAX(...) is NULL with track=on)
setup OK: track_commit_timestamp=on, MAX(...)=NULL, 100 rows in 'frozen_tbl'
Step 5: run connector get_docs() twice with no DML between
  sample sync 1: ['xe_public_frozen_tbl_1 -> None', 'xe_public_frozen_tbl_10 -> None']
  sample sync 2: ['xe_public_frozen_tbl_1 -> None', 'xe_public_frozen_tbl_10 -> None']
PASS: 100 docs, every _timestamp is None and stable across two syncs with no DML between them
```

**With the fix temporarily reverted (proves the script catches the bug):**
```
sample sync 1: ["xe_public_frozen_tbl_1 -> '2026-05-19T18:40:12.294298+00:00'", ...]
sample sync 2: ["xe_public_frozen_tbl_1 -> '2026-05-19T18:40:12.347359+00:00'", ...]
FAIL: _timestamp differs between syncs (the bug is present)
```

Each document gets a fresh `iso_utc()` 53 ms apart between the two syncs — exactly the customer-reported "all data re-indexed every sync" symptom.

<details>
<summary>repro_track_commit_timestamp.py</summary>

```
#
# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
#
# ruff: noqa: T201
"""Manual end-to-end repro for the `track_commit_timestamp` bug fix.

What this script proves
-----------------------
The PostgreSQL connector used to assign `_timestamp = iso_utc()` to every document
emitted from a table whose `MAX(pg_xact_commit_timestamp(xmin))` returned NULL,
which made the Elasticsearch sink see every document as updated on every sync and
re-index everything every sync.

This script exactly reproduces the customer's scenario:
  1. Restart PostgreSQL with `track_commit_timestamp = off`.
  2. Load rows (no commit timestamps recorded for them).
  3. Restart PostgreSQL with `track_commit_timestamp = on`.
  4. Confirm `MAX(pg_xact_commit_timestamp(xmin))` returns NULL on the loaded
     table (because the rows pre-date the setting being on).
  5. Run `PostgreSQLDataSource.get_docs()` twice with no DML between calls.
  6. Assert that every document's `_timestamp` is identical between the two
     calls (and is `None`).

Pre-fix the script fails the cross-sync stability assertion.
Post-fix it prints `PASS` and exits 0.

How to run
----------
1. Start the postgres container from the fixture directory:
       cd app/connectors_service/tests/sources/fixtures/postgresql
       ELASTICSEARCH_DRA_DOCKER_IMAGE=hello-world \\
           docker compose up -d postgresql
2. Run this script from the connector_service venv:
       app/connectors_service/.venv/bin/python \\
           app/connectors_service/scripts/repro_track_commit_timestamp.py
"""

import asyncio
import subprocess
import sys
import time

import asyncpg
from connectors.sources.postgresql import PostgreSQLDataSource
from tests.sources.support import create_source

CONN_STR = "postgresql://admin:Password_123@127.0.0.1:9090/xe"
CONTAINER = "pg-repro"
TABLE = "frozen_tbl"
NUM_ROWS = 100


def docker_exec(*args):
    return subprocess.run(
        ["docker", "exec", CONTAINER, *args],
        check=True,
        capture_output=True,
        text=True,
    )


def docker_restart():
    subprocess.run(["docker", "restart", CONTAINER], check=True, capture_output=True)


async def wait_for_ready(timeout=30):
    deadline = time.time() + timeout
    last_err = None
    while time.time() < deadline:
        try:
            db = await asyncpg.connect(CONN_STR)
            await db.fetchval("SELECT 1")
            await db.close()
            return
        except Exception as exc:
            last_err = exc
            await asyncio.sleep(0.5)
    msg = f"PostgreSQL did not become ready within {timeout}s: {last_err}"
    raise RuntimeError(msg)


async def set_track_commit_timestamp(value: str):
    """Set `track_commit_timestamp` to `value` and bounce the server.

    `track_commit_timestamp` is a PGC_POSTMASTER parameter, so it requires a
    full PostgreSQL restart (not just `pg_reload_conf`).
    """
    docker_exec(
        "psql",
        "-U",
        "admin",
        "-d",
        "xe",
        "-c",
        f"ALTER SYSTEM SET track_commit_timestamp = {value};",
    )
    docker_restart()
    await wait_for_ready()
    actual = (
        docker_exec(
            "psql", "-U", "admin", "-d", "xe", "-Atc", "SHOW track_commit_timestamp;"
        )
        .stdout.strip()
    )
    assert actual == value, f"expected track_commit_timestamp={value}, got {actual!r}"


async def load_rows():
    db = await asyncpg.connect(CONN_STR)
    try:
        await db.execute(f'DROP TABLE IF EXISTS "{TABLE}"')
        await db.execute(
            f'CREATE TABLE "{TABLE}" (id INT PRIMARY KEY, val TEXT NOT NULL)'
        )
        await db.executemany(
            f'INSERT INTO "{TABLE}" (id, val) VALUES ($1, $2)',
            [(i, f"row-{i}") for i in range(1, NUM_ROWS + 1)],
        )
    finally:
        await db.close()


async def assert_max_is_null():
    db = await asyncpg.connect(CONN_STR)
    try:
        max_ts = await db.fetchval(
            f'SELECT MAX(pg_xact_commit_timestamp(xmin)) FROM "{TABLE}"'
        )
        track_setting = await db.fetchval("SHOW track_commit_timestamp")
        row_count = await db.fetchval(f'SELECT COUNT(*) FROM "{TABLE}"')
    finally:
        await db.close()
    assert track_setting == "on", (
        f"expected track_commit_timestamp=on, got {track_setting!r}"
    )
    assert max_ts is None, (
        f"expected MAX(pg_xact_commit_timestamp(xmin))=NULL, got {max_ts!r} "
        "- bug condition not reproduced"
    )
    print(
        f"setup OK: track_commit_timestamp={track_setting}, MAX(...)=NULL, "
        f"{row_count} rows in '{TABLE}'"
    )


async def collect_docs():
    """Run the connector's `get_docs` once and return the emitted documents."""
    async with create_source(
        PostgreSQLDataSource,
        host="127.0.0.1",
        port=9090,
        username="admin",
        password="Password_123",
        database="xe",
        tables=TABLE,
        schema="public",
        ssl_enabled=False,
        ssl_ca="",
        retry_count=1,
        fetch_size=100,
    ) as source:
        return [doc async for doc, _ in source.get_docs()]


def fingerprint(docs):
    return sorted(f"{d['_id']} -> {d['_timestamp']!r}" for d in docs)


async def main():
    print("Step 1: restart PG with track_commit_timestamp=off")
    await set_track_commit_timestamp("off")

    print("Step 2: load rows (no commit timestamps recorded)")
    await load_rows()

    print("Step 3: restart PG with track_commit_timestamp=on")
    await set_track_commit_timestamp("on")

    print("Step 4: verify bug condition (MAX(...) is NULL with track=on)")
    await assert_max_is_null()

    print("Step 5: run connector get_docs() twice with no DML between")
    sync1 = await collect_docs()
    sync2 = await collect_docs()

    fp1 = fingerprint(sync1)
    fp2 = fingerprint(sync2)

    print(f"  sync 1: {len(sync1)} docs")
    print(f"  sync 2: {len(sync2)} docs")
    print(f"  sample sync 1: {fp1[:2]}")
    print(f"  sample sync 2: {fp2[:2]}")

    if len(sync1) != NUM_ROWS or len(sync2) != NUM_ROWS:
        print(
            f"FAIL: expected {NUM_ROWS} docs from each sync, "
            f"got sync1={len(sync1)}, sync2={len(sync2)}"
        )
        sys.exit(1)

    if fp1 != fp2:
        diffs = [(a, b) for a, b in zip(fp1, fp2, strict=True) if a != b][:3]
        print("FAIL: _timestamp differs between syncs (the bug is present)")
        print(f"  first diffs: {diffs}")
        sys.exit(1)

    timestamps = {d["_timestamp"] for d in sync1}
    if timestamps != {None}:
        print(
            f"FAIL: expected every _timestamp to be None (since MAX(...) is NULL "
            f"and track_commit_timestamp=on), got distinct values: {timestamps}"
        )
        sys.exit(1)

    print(
        f"PASS: {NUM_ROWS} docs, every _timestamp is None and stable across two "
        "syncs with no DML between them"
    )


if __name__ == "__main__":
    asyncio.run(main())

```

</details>

artem-shelkovnikov:
Why was it closed? @Jan-Kazlouski-elastic

Jan-Kazlouski-elastic:
It was closed based on discussion in slack in which it was decided:
>"Let's not implement incremental sync. Just remove the incorrect info box from the tutorial"

I CCd you in the thread

PR #4039, opened 2026-05-28
Update .backportrc.json

artem-shelkovnikov:
Backporting to 9.4.0 did not work because of `branchLabelMapping` counting `main` as `v9.4.0`.

This fixes it, we'll need to re-backport stuff after this is merged.

### DIFF
--- a/.backportrc.json
+++ b/.backportrc.json
@@ -10,7 +10,7 @@
   "fork": false,
   "targetPRLabels": ["backport"],
   "branchLabelMapping": {
-    "^v9.4.0$": "main",
+    "^v9.5.0$": "main",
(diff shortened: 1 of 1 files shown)


lorenabalan (approved):
Do we need to add a step in the release checklist for minor versions to update this file as well?

artem-shelkovnikov:
We have this in our release template:

> Bump version in .backportrc.json

I will add more to it to make sure there's an example PR + description of why it's important.

artem-shelkovnikov:
PR to update release template: https://github.com/elastic/search-team/pull/14668

Once merged I'll update current release issues

PR #4041, opened 2026-05-29
[Connectors Python] Drop unused `expand=permissions,history` from space query on Confluence Data Center / Server

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/sdh-search/issues/1889

The Confluence connector's `SPACE_QUERY` is hard-coded to `limit=100&expand=permissions,history` and used by both `ConfluenceClient.fetch_spaces` (sync) and `ConfluenceDataSource._remote_validation` (`validate_config`) for every data source type. Those expansions are only consumed on the Confluence Cloud code path:

- On Data Center / Server, space permissions come from the separate `SPACE_PERMISSION` (Extender) endpoint — the `permissions` field on `/rest/api/space` is never read.
- The `history.createdDate` / `history.createdBy` fields are added to the space document only when `data_source_type == confluence_cloud` (see `format_space`).

Sending these expansions on DC/Server is dead weight and also triggers HTTP 500 from Confluence on several DC/Server versions — e.g. [CONFSERVER-99908](https://jira.atlassian.com/browse/CONFSERVER-99908), where `/rest/api/space?expand=permissions` returns `500` instead of `403` when the calling user is not a Confluence administrator. This causes the connector to fail during `validate_config` and never start a sync (confirmed in a customer log: `GET /rest/api/space?limit=1` returns `200`, the very next `GET /rest/api/space?limit=100&expand=permissions,history` returns `500` with an empty body, every cycle, for hours).

### What this PR changes

- Adds `SPACE_QUERY_DATA_CENTER = "limit=100"` in `constants.py`.
- `fetch_spaces` and `_remote_validation` now select `SPACE_QUERY` for `confluence_cloud` and `SPACE_QUERY_DATA_CENTER` otherwise, following the existing `if data_source_type == CONFLUENCE_CLOUD: … else: …` idiom used throughout `client.py` and `datasource.py`.
- Cloud behavior is unchanged.

### Tests

Two parametrized tests pin the query selection across all three data source types (`confluence_cloud`, `confluence_data_center`, `confluence_server`) at both call sites — so any future change that re-merges the queries will fail with a self-explanatory parametrize ID. All 68 tests in `tests/sources/test_confluence.py` pass.

### Side benefits

- Hardens the connector against the wider `/rest/api/space?expand=…` defect family on DC/Server (CONFSERVER-98638, CONFSERVER-96157, etc.), not just CONFSERVER-99908.
- Small efficiency win on DC/Server (Confluence no longer hydrates / serializes permission and history objects that the connector immediately discards).

#### Pre-Review Checklist

None applicable. No security-relevant changes, no new external dependencies.

## Release Note

Fix Confluence connector failing against Confluence Data Center / Server with HTTP 500 from `/rest/api/space?expand=permissions,history` (e.g. CONFSERVER-99908). The connector no longer requests the unused `permissions` and `history` expansions on Data Center / Server.


app/connectors_service/connectors/sources/atlassian/confluence/client.py
```diff
@@ -30,6 +30,7 @@
     SERVER_USER_BATCH,
     SPACE,
     SPACE_QUERY,
```
Jan-Kazlouski-elastic: If we want to have clear separation, we could rename it to `SPACE_QUERY_CLOUD`.
Let me know if that would be a desired approach.

artem-shelkovnikov: Technically `SPACE_QUERY` run for Cloud, right? Based on the code it does.

Let's rename it for clarity

artem-shelkovnikov (approved):

app/connectors_service/connectors/sources/atlassian/confluence/client.py
Jan-Kazlouski-elastic: You are correct. Now naming is `SPACE_QUERY_CLOUD` and `SPACE_QUERY_DATA_CENTER` based on target.

PR #3720, opened 2025-09-23
[ftest] Fail fast when premature EOF, add verbose output

seanstory:
See [this thread](https://elastic.slack.com/archives/C01DJ3DC0AX/p1755538351504819).

We've been seeing issues like 
```
Downloading Docker image tarball from https://artifacts-snapshot.elastic.co/elasticsearch/9.1.3-a2a9be85/downloads/elasticsearch/elasticsearch-9.1.3-SNAPSHOT-docker-image-amd64.tar.gz...
+ curl --retry 3 --retry-connrefused -O https://artifacts-snapshot.elastic.co/elasticsearch/9.1.3-a2a9be85/downloads/elasticsearch/elasticsearch-9.1.3-SNAPSHOT-docker-image-amd64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 93  674M   93  630M    0     0   140M      0  0:00:04  0:00:04 --:--:--  140M
curl: (92) HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
```

while the release team looks into this, I wanted to:
a) increase our verbosity, so that we can get more details from `curl` regarding what's going wrong.
b) fix an issue introduced in https://github.com/elastic/connectors/pull/3651, that caused these EOF issues to not immediately fail the CI job.

#### Pre-Review Checklist
- [x] this PR does NOT contain credentials of any kind, such as API keys or username/passwords (double check `config.yml.example`)
- [x] this PR has a meaningful title
- [x] this PR has a thorough description
- [x] Tested the changes locally
- [x] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)

## Related Pull Requests

* https://github.com/elastic/connectors/pull/3651

PR #4046, opened 2026-06-01
[Python Connectors] [ftest] Stop flooding CI logs with curl TLS verbose output

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/search-team/issues/14142

The ES Docker tarball download in `app/connectors_service/tests/ftest.sh` runs `curl -v` and captures the output into a shell variable. Because the script runs under `set -exu`, bash's `-x` trace echoes the value of every variable assignment — including captured command substitutions — to stderr. For a ~838 MB ES tarball this dumps the entire TLS handshake, server certificate chain, all HTTP headers and a per-second progress meter into every functional-test build's stderr, making `connectors-nightly` logs hard to read. On the older curl in CI runners the per-record `TLSv1.2 (IN), TLS header, Supplemental data (23):` lines also fire ([curl PR #10299](https://github.com/curl/curl/pull/10299) silenced them upstream in newer versions), turning the noise into thousands of additional lines per download.

`-v` was added in #3720 to give more context on download failures, so this change keeps it — but writes the verbose stream to a temp file (`mktemp /tmp/ftest-curl-${NAME}-XXXXXX.log`, matching the existing `/tmp/fips-ftest-config-*` convention in this script) instead of into a shell variable. The file is `cat`'d to stderr only inside the error branch, preserving the same debug info on real failures (DNS, SSL/TLS, premature EOF, HTTP) without leaking the verbose output through `set -x` on every successful run.

Drive-by fixes that fell out of the same change:
- Replaced `local code=$?` with the `|| curl_exit_code=$?` idiom. The previous form silently relied on `set -e` not firing on `var=$(failing_cmd)`, which actually *did* fire — meaning the function's structured error handling (`Error: curl failed ...`, 404 fallback, `failing fast`) was unreachable. With the new idiom, real failures correctly enter the error branch.
- 404 / output checks now read directly from the temp file (`grep -q ... "$curl_log"`, `cat "$curl_log" >&2`).
- Per-return-path `rm -f` of the temp log on the happy path and inside the function's error branches, plus a single `EXIT` trap registered right after `NAME=$1` that sweeps any `/tmp/ftest-curl-${NAME}-*.log` left behind. The trap (rather than a tail-of-script `rm`) is required because the `PERF8 == "yes"` branch calls `exit $STATUS` before reaching the end of the script — a tail-placed sweep would be unreachable on that path and would leak temp files. The trap also covers `set -e` aborts and signal-driven exits, and preserves the script's exit code.

## Testing

End-to-end ftest:
- `make ftest NAME=gitlab` with the connector running at `--debug` log level — full ftest passed (exit 0). Captured stderr: 6,058 lines / 3,196 `DEBUG`-level entries from the connector itself, **0** lines matching `TLS header`, `TLSv1`, `Supplemental data`, or `TLS handshake`.

Direct A/B against the real ES tarball URL (838.5 MB) under the same `set -exu` regime, comparing `origin/main` and the fix branch with verbatim-extracted `download_docker_tarball` bodies:

| Metric | BEFORE (`origin/main`) | AFTER (this PR) |
|---|---:|---:|
| **Successful download — stderr size** | 11,748 B / 97 lines | 1,626 B / 20 lines |
| `TLSv1` / `TLS handshake` lines on success | 11 / 9 | **0 / 0** |
| HTTP request+response headers leaked on success | yes | no |
| Server cert chain leaked on success | yes | no |
| Progress-meter lines leaked on success | 88 | 0 |
| **Failure (closed-port URL) — harness exit code** | 7 (`set -e` aborted before `if [ $code -ne 0 ]`) | 1 (function reached `return 1`) |
| `curl: (7) Failed to connect ...` visible on failure | yes (via `set -x`) | yes (via `cat curl_log`) |
| `Error: curl failed with exit code N` printed on failure | no (unreachable) | yes |
| `Network or download error - failing fast` printed on failure | no (unreachable) | yes |
| Leftover `/tmp/ftest-curl-*` files after success / failure | n/a | 0 / 0 |

Download itself is unchanged: same bytes (879,230,465), same exit code, comparable wall time (~85 s vs ~87 s).

`EXIT` trap validation (separate harness, sentinel temp file pre-created before each run, mirroring the layout of `ftest.sh`):

| Exit path | Tail-of-script `rm` (rejected) | `EXIT` trap (this PR) |
|---|---|---|
| Normal completion (`PERF8=no`) | cleaned | cleaned |
| PERF8 path (`exit $STATUS`) | **leaked** | cleaned |
| `set -e` abort mid-script | leaked | cleaned |
| Exit-code propagation (`exit 42` through trap) | n/a | preserved (`exit=42`) |

`bash -n app/connectors_service/tests/ftest.sh` is clean.

#### Pre-Review Checklist
- [X] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)
- [X] For bugfixes: backport safely to all minor branches still receiving patch releases

None for the functional-test fix itself — test-infrastructure-only change; does not touch any product code, dependencies, configuration, or runtime behavior of the connector.

Note for reviewers: the branch also includes an incidental commit (`781da2a7`) bumping `idna` from 3.16 → 3.17 in `app/connectors_service/NOTICE.txt`, auto-regenerated by the build. Feel free to drop it if you want this PR scoped strictly to the ftest fix.

## Related Pull Requests

* https://github.com/elastic/connectors/pull/3720 — original PR that added `curl -v` and introduced both the noise and the unreachable error-handling branch fixed here.

## Release Note

None — internal test-infrastructure change, no user-facing impact.


app/connectors_service/tests/ftest.sh
```diff
@@ -216,27 +221,36 @@ function download_docker_tarball {
...
   if [ $curl_exit_code -ne 0 ]; then
     echo "Error: curl failed with exit code $curl_exit_code" >&2
-    echo "Curl output: $curl_output" >&2
+    echo "Curl output:" >&2
+    cat "$curl_log" >&2
 
     # Check if it's a 404 (file not found) - these are expected and should allow fallback
-    if echo "$curl_output" | grep -q "404\|Not Found"; then
+    if grep -q "404\|Not Found" "$curl_log"; then
+      rm -f "$curl_log"
```
artem-shelkovnikov, line 241: Why `-f` here and below?

app/connectors_service/tests/ftest.sh
```diff
@@ -216,27 +221,36 @@ function download_docker_tarball {
...
 
-  # Capture curl output and exit code
-  local curl_output
-  curl_output=$(curl -v --http1.1 --retry $CURL_MAX_RETRIES --retry-delay $CURL_RETRY_DELAY --retry-connrefused -O "$tar_url" 2>&1)
-  local curl_exit_code=$?
+  # Send curl -v output to a temp file (not a variable) so `set -x` does not
+  # echo the whole verbose stream on every successful run; we `cat` it to
+  # stderr only in the error branch.
+  local curl_log
+  curl_log=$(mktemp "/tmp/ftest-curl-${NAME}-XXXXXX.log")
```
artem-shelkovnikov: What's `XXXXXX` in the end about? Did you want to also add some suffix like a timestamp there?

Jan-Kazlouski-elastic: `XXXXXX` is `mktemp`'s required template placeholder. `mktemp` replaces the trailing `X`s with random characters and atomically creates a unique file. Timestamp of two runs in the same second could collide, while `mktemp` guarantees uniqueness.

app/connectors_service/tests/ftest.sh
Jan-Kazlouski-elastic: `-f` is for `force`.
It won't error if the file is already gone, and it won't prompt. Both matter because the `EXIT` trap also sweeps `/tmp/ftest-curl-${NAME}-*.log`, so a given path may already be deleted by the time another branch runs — and under `set -e` a plain `rm` returning non-zero on a missing file would abort the script. `-f` keeps the cleanup idempotent and silent regardless of which path got there first.

app/connectors_service/tests/ftest.sh
Jan-Kazlouski-elastic: https://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/mktemp.html

artem-shelkovnikov: Then they should be trailing I believe:

```
artemshelkovnikov-m2@Work-M2 ~ % curl_log=$(mktemp "/tmp/ftest-curl-${NAME}-XXXXXX.log")
artemshelkovnikov-m2@Work-M2 ~ % echo $curl_log
/tmp/ftest-curl--XXXXXX.log
artemshelkovnikov-m2@Work-M2 ~ % curl_log=$(mktemp "/tmp/ftest-curl-${NAME}.log-XXXXXX")
artemshelkovnikov-m2@Work-M2 ~ % echo $curl_log
/tmp/ftest-curl-.log-FbPy1y
```

Jan-Kazlouski-elastic:
### Testing for `b0141679` (trailing-`X` temp-name fix)

Re-ran the full A/B matrix on the current branch; all green.

- **mktemp naming:** new template `…​.log-XXXXXX` has trailing `X`s → 200/200 unique names, 0 literal `XXXXXX`, and the `EXIT` trap glob matches 201/201 produced files (old glob matched 0). Fixes the macOS/BSD `mktemp` case from the review comment (GNU `mktemp` was just leniently masking it).
- **Success noise** (real 51 MB Elastic HTTPS artifact, genuine curl): 88→**19** lines, TLS handshake 11→**0**, headers 10→**0**, cert chain 3→**0**, progress 5→**0**.
- **Old-CI-curl flood** (mock, curl <#10299): 2045→**21** lines, `Supplemental data` 2000→**0**.
- **Failure** (closed port): exit `7`→**`1`**, error branch now reachable (`Error: curl failed…` + `failing fast` printed).
- **EXIT trap:** cleans on normal / `PERF8 exit` / `set -e` abort (tail-`rm` leaked the last two); exit code 42 preserved.
- `bash -n` clean; 0 leftover temp files.

app/connectors_service/tests/ftest.sh
Jan-Kazlouski-elastic: Very well spotted! Missed the `trailing` part in the description. Thank you. Fixed.

Jan-Kazlouski-elastic:
@artem-shelkovnikov 
Your comment was addressed. Could you give this PR another look?

artem-shelkovnikov (approved):
🚀 🌑

PR #4059, opened 2026-06-10
[Jira] Use deprecated v2 issue search API for Server/Data Center

### LINKED ISSUE #3867: Jira and Confluence connectors: ensure that deprecated APIs are still used for older Jira/Confluence Server
## Bug Description
See https://github.com/elastic/connectors/pull/3710, https://developer.atlassian.com/changelog/#CHANGE-2046 and https://developer.atlassian.com/cloud/jira/platform/changelog/#CHANGE-2046.

It seems like the APIs are deprecated for Cloud products and newer Server products, but if used against older versions the new APIs are not available. We need to make sure that when hitting older Server products we still use the deprecated APIs.

### To Reproduce
1. Create a Jira/Confluence server deployment of versions smaller than 10
2. Try to sync content from these instances
3. See error

## Expected behavior
Sync works for older and newer Server deployments

### LINKED ISSUE #4058: [Jira] Data Center sync broken after v3 API migration: ISSUES endpoint not compatible with DC
## Bug Description

After the migration in #3710 to use `rest/api/3/search/jql` for fetching issues, the Jira connector is broken for Jira Data Center deployments. Jira Data Center pre-v10 does not expose `rest/api/3` endpoints and only supports `rest/api/2/search`. The v3 endpoint also uses cursor-based pagination (`nextPageToken`), which is not available in v2. As a result, any sync against an affected Jira Data Center instance will fail when attempting to fetch issues.

The existing pattern for handling this divergence already exists in the codebase (e.g. `USERS` vs `USERS_FOR_DATA_CENTER`), but was not applied to the issues endpoint.

## Fix

The immediate fix is a safe fallback: non-Cloud instances use `rest/api/2/search?jql={jql}&maxResults={max_results}&startAt={start_at}` with offset-based pagination, consistent with the existing DC-specific patterns in the connector. This is correct for all Server/DC versions since v2 remains available even on newer versions.

A follow-up optimisation could use version detection via `rest/api/2/serverInfo` to enable cursor-based pagination on DC/Server v10+ instances that support v3. See #3867 for the broader context on API version divergence across Server versions.

## References

- #3710 (introduced the regression)
- #3867 (broader context on API version divergence for Server/DC)

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/sdh-search/issues/1896

PR #3710 migrated the Jira issues endpoint to the new cursor-based `rest/api/3/search/jql` endpoint. That endpoint is not available on Jira Server/Data Center pre-v10, so syncs against those instances fail when fetching issues (see also #4058).

This change makes the connector pick the issue search API based on the configured data source type, following the existing `USERS` vs `USERS_FOR_DATA_CENTER` divergence pattern:

- **Jira Cloud**: keeps using the cursor-based `rest/api/3/search/jql` endpoint (`nextPageToken` pagination).
- **Jira Server / Data Center**: falls back to the deprecated, offset-based `rest/api/2/search` endpoint (`startAt` pagination), which is available on all Server/DC versions.

Both paths now apply the same `key IS NOT EMPTY` catch-all when no JQL is provided, so behavior is consistent across data source types and `jql=None` is never formatted into the URL as the literal string `"None"`.

Confluence is intentionally out of scope: it uses stable v1 REST APIs (`rest/api/search`, `rest/api/content/search`, `rest/api/space`) that remain available on older Server/DC versions and are unaffected by this class of deprecation.

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors/pull/3710 (introduced the regression)

## Release Note

The Jira connector now falls back to the deprecated `rest/api/2/search` endpoint for Jira Server and Data Center deployments, fixing syncs against versions older than v10 that do not support the newer `rest/api/3/search/jql` endpoint.

### DIFF
--- a/app/connectors_service/connectors/sources/atlassian/jira/client.py
+++ b/app/connectors_service/connectors/sources/atlassian/jira/client.py
@@ -11,6 +11,7 @@ from connectors_sdk.logger import logger
 
 from connectors.sources.atlassian.jira.constants import (
     ALL_FIELDS,
+    ALL_ISSUES_JQL,
     DEFAULT_RETRY_SECONDS,
     FETCH_SIZE,
     ISSUE_DATA,
@@ -213,10 +214,6 @@ class JiraClient:
                 await self._handle_client_errors(url=url, exception=exception)
 
     async def _paginated_api_call_cursor_based(self, url_name, jql=None, **kwargs):
-        if not jql and url_name == ISSUES:
-            # only bound jql allowed, so using "key IS NOT EMPTY" as a catch-all for "all issues"
-            jql = "key%20IS%20NOT%20EMPTY"
-
         url = parse.urljoin(
             self.host_url,
             URLS[url_name].format(
@@ -259,11 +256,6 @@ class JiraClient:
                 break
 
     async def _paginated_api_call_offset_based(self, url_name, jql=None, **kwargs):
-        if not jql and url_name == ISSUES_FOR_SERVER:
-            # match cursor-based behaviour: use "key IS NOT EMPTY" as a catch-all for
-            # "all issues" so jql=None isn't formatted as the literal string "None"
-            jql = "key%20IS%20NOT%20EMPTY"
-
         start_at = 0
         self._logger.info(
             f"Started pagination for the API endpoint: {URLS[url_name]} to host: {self.host_url} with the parameters -> startAt: 0, maxResults: {FETCH_SIZE} and jql query: {jql}"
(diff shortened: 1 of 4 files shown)


app/connectors_service/connectors/sources/atlassian/jira/client.py
```diff
@@ -258,6 +259,11 @@ async def _paginated_api_call_cursor_based(self, url_name, jql=None, **kwargs):
                 break
 
     async def _paginated_api_call_offset_based(self, url_name, jql=None, **kwargs):
+        if not jql and url_name == ISSUES_FOR_SERVER:
+            # match cursor-based behaviour: use "key IS NOT EMPTY" as a catch-all for
+            # "all issues" so jql=None isn't formatted as the literal string "None"
+            jql = "key%20IS%20NOT%20EMPTY"
```
Apmats: I don't think it matters because this is already "legacy" code but we could move this to a point shared between all places we do this, so that each method doesn't need to know how to construct this magic thingy.

Apmats (approved):
Approved with a nitpick comment.

app/connectors_service/connectors/sources/atlassian/jira/client.py
Jan-Kazlouski-elastic: Agreed. Fixed.

Issue #2933, opened 2024-11-01
Simplify Teams connector design to not run a queue inside of the connector


artem-shelkovnikov:
### Problem Description

Teams connector is syncing data using Graph API: https://github.com/elastic/connectors/blob/main/connectors/sources/microsoft_teams.py#L249.

Graph API has tough[ throttling policies](https://learn.microsoft.com/en-us/graph/throttling) and in most cases there's no point implementing high concurrency when loading data from it. Teams connector does concurrent requests to Graph API:

```
class MicrosoftTeamsDataSource(BaseDataSource):
    def __init__(self, configuration):
        self.queue = MemQueue(maxmemsize=QUEUE_MEM_SIZE, refresh_timeout=120)
        ...

    async def get_docs(self):
        ...
        async for tabs in self.client.get_user_chat_tabs(chat_id=chat["id"]):
            for tab in tabs:
                await self.queue.put(
                    (
                        self.formatter.format_doc(
                            item=tab,
                            document_type=self.schema.chat_tabs,
                            document={
                                "type": UserEndpointName.TABS.value,
                                "url": tab.get("configuration", {}).get("websiteUrl"),
                                "_timestamp": chat["lastUpdatedDateTime"],
                                "members": members,
                            },
                        ),
                        None,
                    )
                )
        ...
```

 which:

1. Makes code more complex
2. Makes errors much less readable due to the way they are handled (in async queue rather than in regular async methods)
3. Does not add any parallelism even - see the code snippet, it puts the formatter statements into the queue, so data is loaded in a regular way

### Proposed Solution
Get rid of usage of MemQueue from Microsoft Teams connector

### Alternatives
None

### Additional Context
This issue was created after seeing a user reporting problems and their log statements from the connector were absolutely unreadable:

```
[FMWK][15:53:04][ERROR] Exception found for task Task-471: 'NoneType' object has no attribute 'get'
NoneType: None
```

Investigation has shown that it happens because the queue is masking the exceptions potentially

PR #4062, opened 2026-06-10
[Jira] Stop connector process memory from growing across full syncs

### LINKED ISSUE #3914: Connectors integration steadily increases in memory consumption over time
## Bug Description

In an agentless deployment, the connectors process memory usage increases after each full sync and does not return to baseline, eventually leading to an OOM kill.

### To Reproduce
I would assume that the steps to reproduce are to deploy a Jira connector with an agentless agent.
But there has been an attempt to reproduce this by @artem-shelkovnikov without success. So it might be related to the Jira Cloud instance used.

## Expected behavior

Connector process memory usage should return to a steady baseline after a full sync completes, without continuous growth across syncs leading to OOM.

## Screenshots

Here is the memory usage of the entire agentless pod over a period of 7 days. Each bump happens when a full sync is triggered. The full sync on the pink arrow resulted in an OOM kill.

Image

Here is the same period of time, showing the memory usage of each process in the pod:

Image

## Environment

- agentless container image: `docker.elastic.co/cloud-release/elastic-agent-service:9.2.3`

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/connectors/issues/3914

In agentless deployments the connector is a long-lived process that runs many
full syncs in a row. Process RSS climbed after Jira full syncs and could
eventually lead to an OOM kill.

Two changes reduce the per-sync memory footprint and remove a failure mode that
could strand work in memory:

1. **Stop retaining the full issue-search payload** (`JiraClient`).
   `get_issues_for_jql` now yields only the issue key; the full issue is already
   re-fetched per key, so yielding the whole `fields=*all` search object only
   pinned large payloads on the fetcher queue and inflated peak memory.

2. **Never strand the consumer on a producer error** (`JiraDataSource`). The
   producers (`_get_projects`, `_get_issues`, `_put_issue`) now emit the
   `FINISHED` sentinel from a `finally` block so the consumer's task counter
   always drains and the sync can't hang. Errors still propagate and stay as
   visible as before.

**How it was verified:** unit tests cover both changes — issue listing now
yields keys only, and `FINISHED` is emitted (and errors still propagate) even
when a producer fails.

#### Pre-Review Checklist

- None.

## Related Pull Requests

None.

## Release Note

Reduced the Jira connector's memory usage during full syncs and fixed a case
where a failed fetch could leave a sync waiting indefinitely, both of which
contributed to growing memory consumption in agentless deployments.


artem-shelkovnikov (commented):
We also have a couple issues that ask to remove usage of MemQueue from the connectors, like this one, for example: https://github.com/elastic/connectors/issues/2933

I think in the end this will also reduce the amount of bugs related to parallelism and not releasing the resources.

I have no concerns about changes in this PR, but wanted to ask - have you done benchmarks verifying that queue fill-up along with stuck syncs is actually the source of this problem?

Jan-Kazlouski-elastic:
The issue was claimed to be not reproduceable in its description and during the work on the fix it wasn't benchmarked enough to claim queue fill-up + stuck syncs is the root cause. Original incident was caused by hundreds of syncs taking more and more memory over the week. OOM was not hit locally so this is more of a low-risk hardening rather than a proven root-cause fix.

`MemQueue` is bounded (5 MB) and fetchers are capped at concurrency 5, so the queue itself can't really explain multi-GB growth. The issue-key change mainly trims peak memory, and the `FINISHED`-in-`finally` change fixes a real hang/stranding bug. But we don't have confirmation that it actually fired in this incident.

Removing `MemQueue` (#2933) sure does look like a more durable fix for this class of bugs. It could be too early to say this PR would close the issue.

We could introduce the usage of the `malloc_trim(0)`. Python/glibc frees objects back to the allocator, but glibc does not always return the arenas to the OS, so RSS stays inflated. The sync cleanup would call a best-effort `trim_memory()` (glibc `malloc_trim(0)`). It is a no-op on non-glibc platforms and never raises. According to local tests `trim_memory` reliably returned ~23 MiB/sync to the OS. 

@artem-shelkovnikov

artem-shelkovnikov (approved):
Let's try it, this change does not look like it will hurt the connector.

I also don't think manual memory trimming is a good way forward for us, theoretically GC should be smart enough to do this

PR #4065, opened 2026-06-11
Fix Outlook connector crash on localized Exchange servers (contacts folder)

### LINKED ISSUE #4064: Outlook connector fails to sync contacts on localized Exchange servers
Reported via https://github.com/elastic/sdh-search/issues/1898.

On non-English Exchange servers, default folder names are localized. The connector currently accesses the Contacts folder via name-based path navigation (`account.root / "Top of Information Store" / "Contacts"`), which raises `ErrorFolderNotFound` and crashes the sync on localized deployments.

**Fix:** Try the existing path first; fall back to `account.contacts` (a distinguished folder ID, locale-agnostic); warn and skip if both fail.

Apmats:
## Problem

On non-English Exchange servers, default folder names are localized. The connector resolved folders by their English display names (e.g. `account.root / "Top of Information Store" / "Contacts"`), which raises `ErrorFolderNotFound` on localized deployments. For the Contacts folder this was unhandled and crashed the sync.

## Fix

Resolve folders by their locale-agnostic **distinguished folder IDs** instead of display names:

- **Contacts:** `get_contacts` now uses `account.contacts` (backed by the `"contacts"` distinguished folder ID). If it still can't be resolved, it logs a warning and skips instead of crashing the sync.
- **Archive:** the parent is now resolved via `account.msg_folder_root` (the `"msgfolderroot"` ID, i.e. "Top of Information Store") instead of its display name. The bare `except Exception` was also tightened to `except ErrorFolderNotFound`.

## Testing

- Unit tests cover both folders: successful ID-based resolution, and graceful skip when the folder is missing.
- No functional (ftest) coverage: Outlook has no emulated backend, and the issue only reproduces on a non-English Exchange server. Manual verification is via the shared M365 tenant with the mailbox locale changed (admin access required).

## Known limitation: Archive leaf folder

The `"Archive"` leaf folder must still be resolved by name, because Microsoft never assigned it a distinguished folder ID in EWS — there is no locale-agnostic way to reach it. On localized servers it is skipped (existing behavior, now via the specific `ErrorFolderNotFound` catch). A full fix would require exposing the archive folder name as a connector config option so users can supply their localized name.

### DIFF
--- a/app/connectors_service/connectors/sources/outlook/client.py
+++ b/app/connectors_service/connectors/sources/outlook/client.py
@@ -379,11 +379,10 @@ class OutlookClient:
                 f"Fetching {mail_type['folder']} mails for {account.primary_smtp_address}"
             )
             if mail_type["folder"] == "archive":
-                # If 'Archive' folder is not present, skipping the iteration
+                # msg_folder_root is locale-agnostic; the "Archive" leaf has no
+                # distinguished ID, so resolve it by name and skip if absent.
                 try:
-                    folder_object = (
-                        account.root / "Top of Information Store" / "Archive"
-                    )
+                    folder_object = account.msg_folder_root / "Archive"
                 except ErrorFolderNotFound:
                     continue
             else:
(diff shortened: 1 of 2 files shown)


Jan-Kazlouski-elastic (approved):

artem-shelkovnikov:
Oh wow, good stuff!

PR #4075, opened 2026-06-15
fix(cli): allow help without config

### LINKED ISSUE #3700: `connectors login` required before `connectors <command group> --help`, but error message doesn't say so
## Bug Description

When installing the `connectors` CLI locally from `main`, if you don't run `login` first, none of the commands with `--help` will work and the error message is confusing.

### To Reproduce
1. Go to `connectors` repo in a new virtual environment (` python -m venv .venv `)
2. Install CLIs locally (`make install` or `pip install -e .`)
3. Try `--help` on one of the `connectors` commands like `connectors job --help`
4. See error `.cli/config.yml was not found. Make sure that the config is either present at the default location (.cli/config.yml) or it's passed via the '-c' or '--config' option.`

## Expected behavior

`--help` commands should ideally work without `login`, but if not possible we should prompt the user to run `connectors login` first.

Optionally, as @seanstory suggested, it would be nice if we can default to calling `login` in the background with default credentials, such as `elastic` & `changeme` in order make the UX a bit smoother.

## Screenshots

Image

## Environment

- OS: MacOS 15.6.1
- Version: 9.2.0

nanookclaw:
## Summary

Fixes the CLI help path so command-group and subcommand help can render before a local `.cli/config.yml` exists. The root CLI now stores a lazy config wrapper and only loads the config when a real command callback reads from it, which keeps help-only invocations from failing before Click can print usage.

Normal command execution without a config still fails, but the message now points users at the expected first step: `connectors login`.

## Testing

- `python3 -m py_compile app/connectors_service/connectors/connectors_cli.py app/connectors_service/tests/test_connectors_cli.py`
- `git diff --check`
- Stubbed Click-routing smoke check for `job --help`, `job list --help`, `job list <connector_id>`, and a `--help` token used as an option value

I also tried `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest app/connectors_service/tests/test_connectors_cli.py -q`, but local pytest could not start because this environment is missing the `elasticsearch` package required by the repo warning filters.

Closes #3700


artem-shelkovnikov:
@nanookclaw thanks for the contribution, the change looks okay to me. Can you sign the CLA so that we could continue with CI and review?

nanookclaw:
Completed the Elastic Contributor Agreement on 2026-06-15 as Nanook Claw using the commit author email `nanook@agentmail.to`, but the `CLA` status context is still failing on commit `46b21d2e`.

I do not want to churn the PR while that external status is unreconciled. If there is a specific recheck path or GitHub/email mapping I should update, I can do that; otherwise I will wait for the CLA sync/recheck before expecting CI/review to proceed.

artem-shelkovnikov:
buildkite test this

nanookclaw:
Rebased `fix/cli-help-without-config` onto current `main` and force-pushed head `df88c86d`.

Local verification:
- `python3 -m py_compile app/connectors_service/connectors/connectors_cli.py app/connectors_service/tests/test_connectors_cli.py`
- `git diff --check`

`PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest app/connectors_service/tests/test_connectors_cli.py -q` still cannot collect in this local environment because the pytest warning filters import the missing `elasticsearch` package. Buildkite should now be testing the current branch.

app/connectors_service/connectors/connectors_cli.py
```diff
@@ -45,6 +45,25 @@ def load_config(ctx, config):
...
+    def __getitem__(self, key):
+        if self.value is None:
+            try:
+                self.value = load_config(self.ctx, self.config)
+            except FileNotFoundError as e:
+                click.echo(
+                    f"{e} Run `connectors login` first, or make sure that the config is either present at the default location ({CONFIG_FILE_PATH}) or it's passed via the '-c' or '--config' option."
+                )
+                self.ctx.exit(1)
+
```
Jan-Kazlouski-elastic, line 63:
```suggestion
        if self.value is None:
            msg = f"{CONFIG_FILE_PATH} is empty or invalid. Run `connectors login` to (re)create it."
            raise click.ClickException(msg)

```

Jan-Kazlouski-elastic (commented):
make lint fails on this branch due to a type error introduced by LazyConfig:
>connectors/connectors_cli.py:64:16 - error: Object of type "None" is not subscriptable (reportOptionalSubscript)

Adding a guard fixes the lint error and turns that crash into a helpful message. 
Apart from that - great job.
@artem-shelkovnikov , I believe after the fix is applied - this could be approved.

app/connectors_service/connectors/connectors_cli.py
artem-shelkovnikov: Let's close this one and leave another PR that tries to address the same problem?

Jan-Kazlouski-elastic: This one actually fixes it, I tested it. The other one doesn't work.
@artem-shelkovnikov

artem-shelkovnikov: I'm happy to merge the one that addresses the issue - if it's this one then let's merge this one :)

nanookclaw:
Addressed the lazy-config review feedback in `4f01401`.

`LazyConfig` now treats an empty or YAML-null config as invalid, prints a clear `connectors login` message, and exits with status 1 instead of attempting to subscript `None`. Missing-file handling, help commands, and valid configurations are unchanged. Added parameterized coverage for both empty and `null` config files.

Verification on the pushed tree:
- `ruff check` passed
- `ruff format --check` passed
- `python3 -m compileall` passed
- `git diff --check` passed
- dependency-independent lazy-config regression harness passed for empty/null-like, valid, and missing configurations

The focused pytest file could not start in this environment because the project `elasticsearch` dependency is unavailable, including after disabling third-party plugin autoload.

Jan-Kazlouski-elastic:
Greetings @nanookclaw 

Thanks for the update, the empty/null config guard addresses the runtime half of the Jul 2 feedback.

Before merge:
1. **`make lint` still fails** — pyright `reportOptionalSubscript` at `return self.value[key]` (line 70). Using `raise click.ClickException(...)` instead of `click.echo` + `ctx.exit(1)` should fix both runtime and typecheck (see Jul 2 suggestion).
2. **Please merge current main into feature branch** — currently is conflicting and ~58 commits behind; CI hasn't run on `4f01401`.
3. **Invalid but non-null config** — a file containing `{}` still raises `KeyError: 'elasticsearch'` with a traceback; consider treating missing required keys like empty/null.

Core `--help` behavior looks good locally. Happy to approve after lint + Buildkite are green.

Issue #2931, opened 2024-10-30
Outlook Connector Crashes when an SMTP address does not have a mailbox associated


josephschultz-expedient:
## Bug Description
While deploying the Outlook Connector for Exchange Online, I found that the process will crash about half way through the sync operation with this error:

`TransportError: No valid version headers found in response (ErrorNonExistentMailbox('The SMTP address has no mailbox associated with it.'))`

This is a small dev environment with only about 15 users and the only SMTP objects without mailboxes would be a Microsoft 365 Group Email.

This environment was configured as per the [Elastic documentation](https://www.elastic.co/guide/en/enterprise-search/current/connectors-outlook.html). The only caveat would be that I had to add the Microsoft Graph `User.Read.All` permission in order to retrieve the user list. Otherwise it returned a 403 error

`aiohttp.client_exceptions.ClientResponseError: 403, message='Forbidden', url='https://graph.microsoft.com/v1.0/users?$top=999'`

### To Reproduce
1. Verify that the Exchange Online environment has at least one Microsoft 365 Group Email created
2. When creating the new App in Azure, assign both the Exchange Online `full_access_as_app` and Microsoft Graph `User.Read.All` permissions
3. Deploy as per the documentation

## Expected behavior
Unless there is a different approach of retrieving the SMTP or user listing, the ideal behavior would be skipping SMTP addresses without mailboxes associated.

## Environment
- Elasticsearch 8.15.1
- 8.15.1 Elastic Connector running within a docker 26.0 environment
- Office365 / Exchange Online

## Error Logs
```
[FMWK][18:26:01][ERROR] [Connector id: VOzQ1JIB5RYrKcWA6hjz, index name: client-outlook-connector, Sync job id: zdas3pIBsfHVd-ovtKc9] Extractor failed with an error: No valid version headers found in response (ErrorNonExistentMailbox('The SMTP address has no mailbox associated with it.'))
[FMWK][18:26:01][CRITICAL] [Connector id: VOzQ1JIB5RYrKcWA6hjz, index name: client-outlook-connector, Sync job id: zdas3pIBsfHVd-ovtKc9] Document extractor failed
Traceback (most recent call last):
  File "/app/lib/python3.10/site-packages/exchangelib/version.py", line 202, in guess
    list(ConvertId(protocol=protocol).call([AlternateId(id="DUMMY", format=EWS_ID, mailbox="DUMMY")], ENTRY_ID))
[...]

PR #4085, opened 2026-06-19
fix(outlook): skip mailbox-less accounts and stop SSL misconfig from aborting sync

Jan-Kazlouski-elastic:
## Relates to https://github.com/elastic/sdh-search/issues/1898

Fixes two independent failure modes in the Outlook Server connector that caused an entire sync job to abort when a single account or the SSL configuration was misconfigured. The guiding principle of this branch: **per-account problems skip that account and continue; connection-wide problems fail the sync loudly** (so they cannot silently empty the index).

**Issue A — `ErrorNonExistentMailbox` kills the whole sync**

When iterating over AD user accounts, `get_mails()` accesses `account.inbox`, which lazily resolves `account.root` via an EWS `GetFolder` call. For AD users whose SMTP address has no associated Exchange mailbox (a valid LDAP entry but no mailbox provisioned), `exchangelib` raises `ErrorNonExistentMailbox`. With no exception handling around the per-account block, this propagated up and terminated the sync, losing all remaining accounts. This is the still-open part of #2931.

**Fix:** the per-account block in `get_docs()` is wrapped in `try/except ErrorNonExistentMailbox`. A missing mailbox is specific to a single account, so the account is skipped with a warning and the sync continues with the remaining accounts.

**Issue B — `NO_CERTIFICATE_OR_CRL_FOUND` SSL crash**

When `ssl_enabled=True` but no certificate is provided, `ssl_ca` is set to `""`. The old code wrote that empty string to the cert file and still selected `RootCAAdapter`. When urllib3 later called `context.load_verify_locations()` on the empty file it raised `NO_CERTIFICATE_OR_CRL_FOUND`, aborting the sync. The error occurs inside the HTTP layer (not inside `cert_verify()`), so `RootCAAdapter`'s own `try/except` never caught it.

**Fix:** in `ExchangeUsers.get_user_accounts()`, the cert file is written and `RootCAAdapter` is selected only when `ssl_ca` is actually populated. When SSL is enabled but no certificate is supplied, the connector falls back to `NoVerifyHTTPAdapter` and logs a clear warning instead of crashing.

### Why connection-wide SSL errors are NOT skipped per account

An earlier iteration of this branch also caught a custom `SSLFailed` exception inside the per-account loop to "skip" accounts on SSL problems. That approach was removed because it was both ineffective and dangerous:

- **Ineffective:** `SSLFailed` is only raised inside `RootCAAdapter.cert_verify`, and `requests`' `cert_verify` does not actually load the certificate — it only records the CA path. A genuinely bad/expired cert fails later, during the TLS handshake, where `exchangelib` catches the underlying `requests.exceptions.SSLError` and re-raises it as `exchangelib.errors.TransportError` (and explicitly does **not** retry it). So `SSLFailed` was never raised in practice and the `except SSLFailed` clause was dead code.
- **Dangerous:** an SSL/connection failure affects *every* account, not one. Silently skipping all accounts would produce an empty but "successful" sync, which the framework interprets as "all documents deleted" — wiping previously indexed data.

Therefore only `ErrorNonExistentMailbox` (genuinely per-account) is skipped. Connection-wide failures — TLS errors surfaced as `TransportError`, and any other unexpected error — propagate and abort the sync loudly, so the misconfiguration is surfaced to the operator and existing indexed data is preserved.

#### Pre-Review Checklist

- [x] Security-related changes (encryption, TLS, SSRF, etc) — the SSL cert fallback behaviour changes: `ssl_enabled=True` with no cert now falls back to `NoVerifyHTTPAdapter` (unverified connections) instead of crashing. A genuinely invalid/expired certificate now fails the sync loudly rather than being silently skipped. Reviewers should confirm the no-cert fallback posture is desired.

## Related Pull Requests

* Partially addresses #2931

## Release Note

**Outlook Server connector**: sync jobs no longer abort when an Active Directory user has a valid SMTP address but no associated Exchange mailbox (`ErrorNonExistentMailbox`). The affected account is now skipped with a warning and the sync continues. Additionally, configuring SSL without providing a certificate no longer crashes with `NO_CERTIFICATE_OR_CRL_FOUND`; the connector falls back to unverified connections and logs a clear warning. Genuine certificate/connection errors still fail the sync loudly rather than silently emptying the index.

### DIFF
--- a/app/connectors_service/connectors/sources/outlook/client.py
+++ b/app/connectors_service/connectors/sources/outlook/client.py
@@ -200,16 +200,10 @@ class ExchangeUsers:
             yield user
 
     async def get_user_accounts(self):
-        if self.ssl_enabled and self.ssl_ca:
-            await ManageCertificate().store_certificate(certificate=self.ssl_ca)
-            BaseProtocol.HTTP_ADAPTER_CLS = RootCAAdapter
-        else:
-            if self.ssl_enabled and not self.ssl_ca:
-                logger.warning(
-                    "SSL is enabled but no certificate was provided; "
-                    "falling back to unverified SSL connections."
-                )
-            BaseProtocol.HTTP_ADAPTER_CLS = NoVerifyHTTPAdapter
+        await ManageCertificate().store_certificate(certificate=self.ssl_ca)
+        BaseProtocol.HTTP_ADAPTER_CLS = (
+            RootCAAdapter if self.ssl_enabled else NoVerifyHTTPAdapter
+        )
 
         credentials = Credentials(
             username=self.user,
--- a/app/connectors_service/connectors/sources/outlook/client.py
+++ b/app/connectors_service/connectors/sources/outlook/client.py
@@ -200,10 +200,16 @@ class ExchangeUsers:
             yield user
 
     async def get_user_accounts(self):
-        await ManageCertificate().store_certificate(certificate=self.ssl_ca)
-        BaseProtocol.HTTP_ADAPTER_CLS = (
-            RootCAAdapter if self.ssl_enabled else NoVerifyHTTPAdapter
-        )
+        if self.ssl_enabled and self.ssl_ca:
+            await ManageCertificate().store_certificate(certificate=self.ssl_ca)
+            BaseProtocol.HTTP_ADAPTER_CLS = RootCAAdapter
+        else:
+            if self.ssl_enabled and not self.ssl_ca:
+                logger.warning(
+                    "SSL is enabled but no certificate was provided; "
+                    "falling back to unverified SSL connections."
+                )
+            BaseProtocol.HTTP_ADAPTER_CLS = NoVerifyHTTPAdapter
 
         credentials = Credentials(
             username=self.user,
--- a/app/connectors_service/connectors/sources/outlook/datasource.py
+++ b/app/connectors_service/connectors/sources/outlook/datasource.py
@@ -4,10 +4,11 @@
 # you may not use this file except in compliance with the Elastic License 2.0.
 #
 
+import ssl
 from copy import copy
 from functools import cached_property, partial
 
-from connectors_sdk.source import BaseDataSource
+from connectors_sdk.source import BaseDataSource, ConfigurableFieldValueError
 from connectors_sdk.utils import (
     hash_id,
     iso_utc,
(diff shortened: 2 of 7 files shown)


app/connectors_service/connectors/sources/outlook/datasource.py
```diff
@@ -562,24 +563,34 @@ async def get_docs(self, filtering=None):
...
 
-            async for child_calendar in self._fetch_child_calendars(
-                account=account, timezone=timezone
-            ):
-                yield child_calendar
+                async for child_calendar in self._fetch_child_calendars(
+                    account=account, timezone=timezone
+                ):
+                    yield child_calendar
+            except ErrorNonExistentMailbox:
```
artem-shelkovnikov, line 623: Would non-existent mailbox mean that there are no contacts, tasks, calendars and so on?

Jan-Kazlouski-elastic: Yes. In exchangelib, `inbox`, `contacts`, `tasks`, and `calendar` all resolve through the same `account.root` (`Root.get_distinguished`) which is what raises `ErrorNonExistentMailbox`. So a missing mailbox means none of them exist. 
It's an account-level condition, not folder-level, so skipping the whole account is correct and nothing syncable is dropped.

artem-shelkovnikov (approved):
LGTM!

Example reason of mailbox-less account: https://help.bittitan.com/hc/en-us/articles/115008114287-No-mailbox-with-such-GUID-ErrorNonExistentMailbox

app/connectors_service/connectors/sources/outlook/client.py
```diff
@@ -200,10 +200,16 @@ async def get_users(self):
...
-        await ManageCertificate().store_certificate(certificate=self.ssl_ca)
-        BaseProtocol.HTTP_ADAPTER_CLS = (
-            RootCAAdapter if self.ssl_enabled else NoVerifyHTTPAdapter
-        )
+        if self.ssl_enabled and self.ssl_ca:
+            await ManageCertificate().store_certificate(certificate=self.ssl_ca)
+            BaseProtocol.HTTP_ADAPTER_CLS = RootCAAdapter
+        else:
+            if self.ssl_enabled and not self.ssl_ca:
+                logger.warning(
```
Jan-Kazlouski-elastic: Hi @artem-shelkovnikov 
Last minute check regarding this unverified SSL connection fallback. It happens here if SSL is enabled, but there is no certificate. Customer gets a warning and the operation proceeds unverified. We might want to throw an error here to be safe. Warning is easy to miss and fallback might give customer false sense of security.

Jan-Kazlouski-elastic: Current approach fixes the reported issue, but I just thought that it might be a better solution to be a bit more restrictive here.

artem-shelkovnikov: True so. Let's revert this code?

Why would the user want to run the connector with SSL enabled but without a certificate?

artem-shelkovnikov (changes requested):
As Jan mentioned, there's a problem with SSL logic in the connector

app/connectors_service/connectors/sources/outlook/client.py
Jan-Kazlouski-elastic: Reverted this and moved check to `validate_config` method in order to have a clean understandable error message.

Jan-Kazlouski-elastic:
Fix is applied, @artem-shelkovnikov 
Could you give this PR another look?

app/connectors_service/tests/test_utils.py
```diff
@@ -847,6 +847,12 @@ def test_evaluate_timedelta():
     assert expected_response == "2023-02-19T14:25:05.158843"
```
artem-shelkovnikov, line 848: The way I read the tests below it looks like the change is not breaking existing logic, only fixing a bug when multiline PEM certificate is provided, is this correct?

Is there a chance this would break existing customers' setup for other connectors?

Jan-Kazlouski-elastic: Single-line certs/keys still take the original reflow path unchanged; the new branch only triggers on already-multiline input, which the old `replace(" ", "\n")` corrupted (e.g. real `.pem` with a trailing newline → broken `-----END` marker → `NO_CERTIFICATE_OR_CRL_FOUND`).
No regression for other connectors: GCS only calls it for newline-free keys, and Redis/MSSQL/Mongo/PostgreSQL/`ssl_context` get byte-identical output for single-line input and a *correct* result for multiline input the old code would have broken.
Verified by reverting just this function and re-running: unchanged-behavior tests pass on old & new, only the bug-fix cases fail on old — and the Redis/MSSQL/Mongo/PostgreSQL/GCS suites (154 tests) stay green.

artem-shelkovnikov (approved):
🚀 🌓

PR #4108, opened 2026-06-26
PoC - expose former native mode publicly

artem-shelkovnikov:
This PR allows running connectors service in what was formly called "native" mode. Names of the modes are arbitrary.

Practically that means that now it's possible to run connectors service without specifying which connectors it will handle, it will attempt to handle all connectors of specified `service.service_types`.

For example old ways of configuring connectors works as expected:

```
connectors:
-
  connector_id: Iah2A58B3JFXscmguRMh
  service_type: outlook
  api_key: <API_KEY>
elasticsearch:
  host: https://localhost:9200
  api_key: <API_KEY>
```

This would make it so that connectors service is handling connector `Iah2A58B3JFXscmguRMh`

However, it's now possible to do this:

```
elasticsearch:
  host: https://localhost:9200
  api_key: <API_KEY>

service:
  worker_mode: true
  service_types: ['mongodb', 'mysql']
```

This will run the service to handle all connectors of types `mongodb` and `mysql` that have `is_native: false` ONLY. When `worker_mode: true` is set, `service_types` need to be provided - they are not set by default!

This change is done in a manner so that there would be no interactions with agentless connectors that work only against connectors that are `is_native: true`. This also means that this change will not get backported to 8.19

Prior to this change you would need to do this (see our [ent-search repo code](https://github.com/elastic/ent-search/blob/main/dist/shared/cloud/coservices/python-connectors.yml#L37-L66)):

```
elasticsearch:
  host: https://localhost:9200
  api_key: <API_KEY>

_force_allow_native: true # In this PR I've changed this to `mode` to be more descriptive, I like enum better than bool
native_service_types: ['mongodb', 'mysql']
```

#### Pre-Review Checklist

## Release Note

Introduced settings `service.worker_mode` and `service.service_types` that allow running the service in a worker mode: instead of handling specific connectors it will attempt to handle all self-hosted connectors of specified `service.service_types`


app/connectors_service/connectors/preflight_check.py
```diff
@@ -232,7 +240,7 @@ def _validate_configuration(self):
                 }
             )
 
-        if not configured_connectors and not force_allowed_native:
+        if not configured_connectors:
```
artem-shelkovnikov, line 241: Optimised the condition above (line 210)

app/connectors_service/tests/test_preflight_check.py
```diff
@@ -288,25 +292,18 @@ async def test_index_exist_transient_error(mock_responses):
 
 @pytest.mark.asyncio
 @patch("connectors.preflight_check.logger")
-async def test_native_config_is_warned(patched_logger, mock_responses):
+async def test_native_config_is_not_warned(patched_logger, mock_responses):
```
artem-shelkovnikov, line 326: Flipped the other way around - now officially supported

app/connectors_service/connectors/protocol/connectors.py
```diff
@@ -231,8 +231,8 @@ async def supported_connectors(self, native_service_types=None, connector_ids=No
             "bool": {
                 "must_not": [{"term": {"deleted": True}}],
                 "filter": [
-                    {"term": {"is_native": True}},
                     {"terms": {"service_type": native_service_types}},
+                    {"term": {"is_native": False}},
```
artem-shelkovnikov, line 235: We don't want the service to run against `is_native` connectors when `service.run_all_service_types: true`. Now agentless works based on this flag, so it would cause weird interactions. Not running against `is_native` connectors is a safe bet that things will work well.

app/connectors_service/connectors/services/job_cleanup.py
```diff
@@ -21,7 +21,7 @@ class JobCleanUpService(BaseService):
     def __init__(self, config):
         super().__init__(config, "job_cleanup_service")
         self.idling = int(self.service_config.get("job_cleanup_interval", 60 * 5))
-        self.native_service_types = self.config.get("native_service_types", []) or []
+        self.native_service_types = self.service_config.get("service_types", []) or []
```
artem-shelkovnikov, line 24: I left the "native_` lingo because I found no better naming

erikcurrin-elastic (approved):
Looks good

app/connectors_service/connectors/preflight_check.py
```diff
@@ -202,15 +202,21 @@ async def _check_system_indices_with_retries(self):
...
 
     def _validate_configuration(self):
-        # "Native" mode
-        configured_native_types = "native_service_types" in self.config
-        force_allowed_native = self.config.get("_force_allow_native", False)
-        if configured_native_types and not force_allowed_native:
-            logger.warning(
-                "The configuration 'native_service_types' has been deprecated. Please remove this configuration."
-            )
+        configured_native_types = "service_types" in self.config["service"]
```
Jan-Kazlouski-elastic: Because we've added 
```
"service": {
  "worker_mode": False,
  "service_types": [],
```
to `config.py` 
`configured_native_types` is going to be `True` by default. Even for non-worker nodes. Because it checks the presence of the key.

artem-shelkovnikov: good catch!

app/connectors_service/connectors/preflight_check.py
```diff
@@ -202,15 +202,21 @@ async def _check_system_indices_with_retries(self):
...
+        if worker_mode:
+            if not configured_native_types:
+                logger.error(
+                    "Service is running with 'service.worker_mode: true' with no 'service.service_types' configured"
+                )
+                return False
+            return True
+
+        # else not running all service types
+        if configured_native_types:
```
Jan-Kazlouski-elastic: WIth `configured_native_types` being `True` by default based on https://github.com/elastic/connectors/pull/4108/changes#r3535248514, this warning will be written all the time. 

That seems like an issue to me.

app/connectors_service/connectors/preflight_check.py
Jan-Kazlouski-elastic: Based on https://github.com/elastic/connectors/pull/4108/changes#r3535248514, this becomes dead code, because `configured_native_types` is always True.

app/connectors_service/connectors/preflight_check.py
```diff
@@ -202,15 +202,21 @@ async def _check_system_indices_with_retries(self):
...
-        force_allowed_native = self.config.get("_force_allow_native", False)
-        if configured_native_types and not force_allowed_native:
-            logger.warning(
-                "The configuration 'native_service_types' has been deprecated. Please remove this configuration."
-            )
+        configured_native_types = "service_types" in self.config["service"]
+        worker_mode = self.config["service"].get("worker_mode", False)
+
+        if worker_mode:
+            if not configured_native_types:
```
Jan-Kazlouski-elastic: Here this would be problematic as well. Because the `service_types` will be empty. but we'll return True.

app/connectors_service/tests/test_preflight_check.py
Jan-Kazlouski-elastic: I believe this test is using obsolete keys now. Doesn't test anything.

Jan-Kazlouski-elastic: I would try searching for a text `native_service_types` and `_force_allow_native` to see if any more tests are dependant on these keys.

app/connectors_service/connectors/services/job_execution.py
```diff
@@ -99,7 +99,7 @@ async def _run(self):
         self.connector_index = ConnectorIndex(self.es_config)
         self.sync_job_index = SyncJobIndex(self.es_config)
 
-        native_service_types = self.config.get("native_service_types", []) or []
+        native_service_types = self.config["service"].get("service_types", []) or []
```
Jan-Kazlouski-elastic: We're using `self.service_config.get("service_types")` which is I believe to be a cleaner approach here: https://github.com/elastic/connectors/pull/4108/changes#diff-229ed60791ecd1f0312fc7d9b77589d892707560dde83e10d630c644beec1f03R24
and 
`self.config["service"].get("service_types", []` here. Any reason for them to be different?

app/connectors_service/connectors/preflight_check.py
artem-shelkovnikov: Agreed, fixed

app/connectors_service/tests/test_preflight_check.py
artem-shelkovnikov: Yup, it was only in this place, good catch

app/connectors_service/connectors/services/job_execution.py
artem-shelkovnikov: Yup changed for consistency

app/connectors_service/config.yml.example
```diff
@@ -220,6 +220,25 @@
...
+##    This means that the api_key in elasticsearch.api_key should have access to all indices that are used by connectors that
+##    are handled by this service. You won't be able to have an api_key per connector any more.
+#service.worker_mode: false
+#
+#
+##  A collection of service types that a service running with "service.worker_mode: true" config will attempt to handle.
+##    This allows partitioning the load allowing to spread different connector types between different
+##    physical instance that can be bigger or smaller depending on the load and hardware requirements.
+##    If `service.worker_mode` is `false` then this setting is ignored.
+##    If no values are provided, then all known service types from "sources" config are used
```
Jan-Kazlouski-elastic: >If no values are provided, then all known service types from "sources" config are used

I don't think this is true.

Based on https://github.com/elastic/connectors/pull/4108/changes#diff-e3dddd7257ae72cc5381efc269efa7dab96de61890295602492b2269f6707b3eR209-R213
we fail loudly.

app/connectors_service/tests/test_preflight_check.py
Jan-Kazlouski-elastic: Could you check whether or not this assertion is a dead code?

app/connectors_service/connectors/services/job_scheduling.py
```diff
@@ -153,13 +153,13 @@ async def _run(self):
         self.connector_index = ConnectorIndex(self.es_config)
         self.sync_job_index = SyncJobIndex(self.es_config)
 
-        native_service_types = self.config.get("native_service_types", []) or []
+        native_service_types = self.config["service"].get("service_types", []) or []
```
Jan-Kazlouski-elastic: Could we use `self.service_config` here as well?

Jan-Kazlouski-elastic (commented):
Looking good. Left a few comments worth looking into

app/connectors_service/connectors/services/job_scheduling.py
artem-shelkovnikov:
```suggestion
        native_service_types = self.service_config.get("service_types", []) or []
```

app/connectors_service/config.yml.example
artem-shelkovnikov:
```suggestion
##    If `service.worker_mode` is `false` then you should explicitly provide service types you want the service to handle
```

app/connectors_service/tests/test_preflight_check.py
artem-shelkovnikov: Funnily it's not a dead code, but does not produce any errors.

You can literally do:

```
patched_logger.lalalalaal()
```

And this will do nothing. I've fixed the test to actually do assert (tested by inverted assert) and it properly works, funnily this line never worked 🤷

app/connectors_service/config.yml.example
Jan-Kazlouski-elastic: https://github.com/elastic/connectors/pull/4146
@artem-shelkovnikov

PR #4109, opened 2026-06-27
fix: allow CLI group help without config

### LINKED ISSUE #3700: `connectors login` required before `connectors <command group> --help`, but error message doesn't say so

Saithej2k:
Summary
- Skip config loading for nested CLI help requests such as `connectors job --help`.
- Update the missing-config message for real commands to point users at `connectors login`.

Testing
- `git diff --check`

Closes #3700

### DIFF
--- a/app/connectors_service/connectors/connectors_cli.py
+++ b/app/connectors_service/connectors/connectors_cli.py
@@ -45,6 +45,21 @@ def load_config(ctx, config):
         raise FileNotFoundError(msg)
 
 
+def is_help_request(ctx):
+    return any(arg in ctx.help_option_names for arg in ctx.args)
+
+
+def ensure_config(ctx):
+    ctx.ensure_object(dict)
+    try:
+        ctx.obj["config"] = load_config(ctx, ctx.obj.get("config_file"))
+    except FileNotFoundError as e:
+        click.echo(
+            f"{e} Run `connectors login` first, or pass a config via the '-c' or '--config' option."
+        )
+        ctx.exit(1)
+
+
 # Main group
 @click.group(
     invoke_without_command=True,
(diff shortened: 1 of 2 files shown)


Jan-Kazlouski-elastic:
Hello @Saithej2k 
Thank you for your contribution. Could you please sign the Contributor License Agreement so that we could continue with CI and review?

Jan-Kazlouski-elastic:
buildkite test this

Saithej2k:
buildkite build this

Saithej2k:
buildkite build this

Jan-Kazlouski-elastic (changes requested):
Thanks for the contribution, @Saithej2k! Unfortunately this only fixes `<group> --help` — nested help like `connectors job start --help` still fails, since the group callback runs `ensure_config` before Click processes the leaf command's `--help`. Also, `is_help_request()` is effectively dead code (`ctx.args` is empty in the parent callback, so it always returns `False`).

Saithej2k:
Thanks for the review. I addressed both points in #4150: configuration now loads only when a leaf command callback runs, and the tests cover every nested help route, including job start --help. I opened the corrected change as a focused follow-up because this PR was already closed.

Issue #2393, opened 2024-04-16
Sync job gets marked as Succeeded even if there was a error that ended the job prematurely


gcasar:
## Bug Description
After an internal connector error that prematurely ends a Full sync the Sync job proceeds, reconciles the index by **deleting documents** previously in the system and is marked as **Sync complete**

### To Reproduce

This behavior is hard to reproduce reliably because it is 1) Connector specific and 2) Relies on the third party error. There are at least two bugs that surface this problem.

Please note that the example I'm giving below is not the only place this happens.
```
06:59:51.107
enterprise_search.connectors
[enterprise_search.connectors][warning] Skipping data for type content from https://stackoverflow.atlassian.net/wiki/rest/api/content/search?next=true&cursor=_f_NTA%3D_sa_WyJcdDM2NjM5OTMwIFI2KF8wZUpoJWhYdDA8L2w8a2NcIiBjcCJd&expand=children.attachment,history.lastUpdated,body.storage,space,space.permissions,restrictions.read.restrictions.user,restrictions.read.restrictions.group&limit=50&start=285750&cql=space+in+('LTDD','BATD','MADCOW')+AND+type=page. Exception: Response payload is not completed.
06:59:53.811
enterprise_search.connectors
[enterprise_search.connectors][info] Sync ended with status completed -- created: 0 | updated: 285753 | deleted: 23748 (took 14338 seconds)
```

## Expected behavior
When an exception occurs that forces a Full sync to prematurely end 
the Sync job should be marked as Failed
and the existing documents in the index should not get Deleted.

As a user of connector
I expect this problem to be handled in a shared way
so that uncaught exceptions in connectors don't get silently discarded
giving a false sense of completed syncs.

## Screenshots
<img width="770" alt="image" src="https://github.com/elastic/connectors/assets/2477157/0e7dff93-97d4-4699-8081-e8c1d04383e2">
Please note that this is an older image, but we are still able to reproduce this in 8.13.2 where the errors reported above are from.

## Environment

- Elastic Cloud, version 8.13.2, native Confluence cloud and Github cloud connectors

## Additional context

The second example is related and for Github:
```
06:39:29.252
enterprise_search.connectors
[enterprise_search.connectors][warning] Something went wrong while fetching the issues. Exception: 403, message='Forbidden', url=URL('https://api.github.com/graphql')
  File "/usr/share/enterprise-search/lib/python3.10/site-packages/connectors/sources/github.py", line 1542, in _fetch_issues
[...]

PR #4119, opened 2026-07-01
[GitHub] Propagate fetch errors instead of swallowing them

### LINKED ISSUE #2395: github swallows errors even if fetching issues fails

Jan-Kazlouski-elastic:
## Summary

Fixes the GitHub connector silently swallowing errors during sync. Previously each fetch method wrapped its whole body in `try/except Exception: log warning`, so a sync could "succeed" while syncing nothing (e.g. if no issues could be fetched at all). This implements the two-tier error handling requested in the issue, removes the broad exception handling, and hardens null-handling of GraphQL responses.

Part of #2393, closes #2395.

### Two-tier error handling (`datasource.py`)
- **Page/API-level failures propagate** out of `_fetch_repos` / `_fetch_pull_requests` / `_fetch_issues` / `_fetch_files` / `_fetch_files_by_path`, so the sync fails when data genuinely can't be fetched.
- **Per-document enrichment failures are WARN-logged and skipped** via narrow guards that catch only `gidgethub.GitHubException`, so a single bad issue/PR/file no longer aborts the whole sync.

### Transient rate limiting
- `_put_to_sleep` now only sleeps; its callers re-raise a dedicated `RateLimitingError`, which is not in `skipped_exceptions`, so the existing `@retryable` layer retries it.

### No more bare `except Exception`
- Replaced across the `github/` package with `gidgethub.GitHubException` (plus the connector's own `UnauthorizedException` / `ForbiddenException` in log-and-reraise spots like `ping` and `_get_invalid_repos_for_personal_access_token`).

### Null-safe GraphQL response handling
- GitHub can return `null` for connection fields (`reviews`, `comments`, `labels`, `assignees`, `reviewRequests`, `tree`, edge nodes). The old `x.get("key", {}).get(...)` pattern crashes with `AttributeError`/`TypeError` when the key is present but its value is `null` (`.get()` only applies the default when the key is absent). Switched the fetch/prepare helpers to the `... or {}` / `... or []` idiom so a single such document can no longer abort the sync.

### Tests
- Added coverage for propagation vs. per-item skip for each fetch method, `RateLimitingError` propagation, and regression tests for issues/PRs with null connection fields.
- Fixed a latent test-isolation bug where shared mock fixtures were mutated across tests and the resulting error was hidden by the old catch-all (now `deepcopy`).

## Acceptance criteria

- [x] GitHub connector fails syncs if no issues can be synced (page-level errors propagate)
- [x] GitHub connector attempts retries for errors that look transient (`@retryable` + `RateLimitingError`)
- [x] Errors with limited scope (single documents / edge cases) are moved past with WARN logging

## Test plan

- `pytest tests/sources/test_github.py` — 121 passed
- Passes across multiple random orderings (`-p randomly` seeds 1/2/42/100/777)
- `ruff check` clean on changed files

## Notes for reviewers

- The per-item guard wraps `_fetch_remaining_fields`, which itself makes paginated calls. The design assumes systemic backend failures surface on the initial page fetch (which propagates and fails the sync). If a systemic error only manifested during per-item enrichment, it could be WARN-skipped for every document without failing the sync. Happy to add per-repo "synced at least one" tracking if we want a hard guarantee.
- `raise Exception(msg)` remains for non-rate-limited GraphQL `QueryError`s in `graphql()`; left as-is as a possible small follow-up.

### DIFF
--- a/app/connectors_service/connectors/sources/github/datasource.py
+++ b/app/connectors_service/connectors/sources/github/datasource.py
@@ -7,6 +7,7 @@
 from collections import defaultdict
 from functools import partial
 
+import gidgethub
 from connectors_sdk.source import BaseDataSource, ConfigurableFieldValueError
 from connectors_sdk.utils import nested_get_from_dict
 
@@ -395,7 +396,11 @@ class GitHubDataSource(BaseDataSource):
                         org_name = self.configuration["org_name"]
                         self.org_repos.setdefault(org_name, {})[repo_name] = repo_data
             return invalid_repos
-        except Exception as exception:
+        except (
+            gidgethub.GitHubException,
+            UnauthorizedException,
+            ForbiddenException,
+        ) as exception:
             self._logger.exception(
                 f"Error while checking for inaccessible repositories. Exception: {exception}.",
                 exc_info=True,
@@ -498,7 +503,7 @@ class GitHubDataSource(BaseDataSource):
         try:
             await self.github_client.ping()
             self._logger.debug("Successfully connected to GitHub.")
-        except Exception:
+        except (gidgethub.GitHubException, UnauthorizedException, ForbiddenException):
             self._logger.exception("Error while connecting to GitHub.")
             raise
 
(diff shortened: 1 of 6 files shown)


app/connectors_service/connectors/sources/github/datasource.py
```diff
@@ -772,55 +769,59 @@ async def _fetch_pull_requests(
...
+                for field in ["comments", "labels", "assignees"]:
+                    await self._fetch_remaining_fields(
+                        type_obj=issue,
+                        object_type=ObjectType.ISSUE.value.lower(),
+                        owner=owner,
+                        repo=repo,
+                        field_type=field,
+                    )
+                    issue.pop(field)
+            except gidgethub.GitHubException as exception:
```
Jan-Kazlouski-elastic, line 819: @artem-shelkovnikov 
Something worth flagging is that this will now catch only GitHubException and write a warning for them without failing the whole sync, but any non-GitHubException will now fail the whole sync. 
I agree that we shouldn't be catching General Exception, but that is a risk.

artem-shelkovnikov: Did you do any testing in regards how errors are raised? I remember this connector was tricky because _I believe_ under the hood the github library uses `aiohttp` and with certain usage of the library it raises `aiohttp` errors instead of its own ones.

I _think_ we've fixed it at some point and only `gidgethub` errors are now exposed, but might be worth checking - like just configuring connector with wrong config, then with wrong repo name and see if the error is correctly propagated.

What do you think?

Jan-Kazlouski-elastic: Tested end-to-end against a local mock and live `api.github.com`.

`gidgethub` only maps **HTTP status codes** to `GitHubException`s; pre-response
failures (DNS, refused, TLS, timeout) come from `aiohttp` and aren't
`GitHubException`s.

- Bad token (local + real api) → `UnauthorizedException` ✅
- Wrong repo → 404 → `gidgethub.BadRequest`, propagates ✅
- Page-level 500 → `gidgethub.GitHubBroken`, propagates ✅
- Per-item 500 → WARN + skipped ✅
- Dead port → `aiohttp.ClientConnectorError` (not a GitHubException)

So errors propagate correctly. Transport errors dodge the per-item
guard, but `@retryable` (3×) absorbs blips and a persistent outage *should* fail
the sync. Extending guard to aiohttp errors would
reintroduce #2395 (skip every doc, sync "succeeds" empty).

artem-shelkovnikov (approved):

Issue #1155, opened 2023-06-30
Unexpected text extraction failures report non-actionable errors


artem-shelkovnikov:
## Bug Description
When unexpected text extraction error happens, sometimes no error is reported and it's not possible to act on it

### To Reproduce
1. Create a Sharepoint Online connector that will sync all site collections (*)
2. Start a sync and observe logs
3. See error:
```
[FMWK][12:58:32][ERROR] Text extraction unexpectedly failed. Error:
[FMWK][12:58:44][ERROR] Text extraction unexpectedly failed. Error:
```

## Expected behavior
Error is communicated clearly.

seanstory:
@navarone-feekery to confirm if this is still an issue

navarone-feekery:
I confirmed that this is still occurring, I'll investigate.

navarone-feekery:
Actually, I was mistaken -- this doesn't happen anymore.
A similar error was happening, and it also happens when not using local text extraction. When I run a full sync on the enterprise search sites, I occasionally get `Exception found for task Task-5802:` with no actionable information.

artem-shelkovnikov:
@navarone-feekery it can probably be closed now then - there's another issue for the empty "Exception found for task" exception: https://github.com/elastic/connectors-python/issues/789

navarone-feekery:
I did one more sync on the full enterprise search collection and couldn't reproduce (just to quadruple check) and it isn't occurring anymore, so I'll close this.

PR #4123, opened 2026-07-02
fix(outlook): harden sync against missing Exchange field values

Jan-Kazlouski-elastic:
## Summary
- Handle nullable mail, calendar, contact, and attachment fields so a single malformed Exchange item no longer aborts the sync (fixes the customer `'NoneType' object has no attribute 'email_address'` failure on `mail.sender`).
- Skip optional mail folders (inbox/sent/junk/archive) when absent instead of failing the whole account.
- Harden on-prem DLS access-control sync for list-valued LDAP `mail` attributes and malformed DNs.

Relates to https://github.com/elastic/sdh-search/issues/1898

## Test plan
- [x] `pytest tests/sources/test_outlook.py` (57 tests)

### DIFF
--- a/app/connectors_service/connectors/sources/outlook/client.py
+++ b/app/connectors_service/connectors/sources/outlook/client.py
@@ -405,15 +405,19 @@ class OutlookClient:
             self._logger.debug(
                 f"Fetching {mail_type['folder']} mails for {account.primary_smtp_address}"
             )
-            if mail_type["folder"] == "archive":
-                # msg_folder_root is locale-agnostic; the "Archive" leaf has no
-                # distinguished ID, so resolve it by name and skip if absent.
-                try:
+            try:
+                if mail_type["folder"] == "archive":
+                    # msg_folder_root is locale-agnostic; the "Archive" leaf has no
+                    # distinguished ID, so resolve it by name and skip if absent.
                     folder_object = account.msg_folder_root / "Archive"
-                except ErrorFolderNotFound:
-                    continue
-            else:
-                folder_object = getattr(account, mail_type["folder"])
+                else:
+                    folder_object = getattr(account, mail_type["folder"])
+            except ErrorFolderNotFound:
+                self._logger.warning(
+                    f"Could not resolve {mail_type['folder']} folder for "
+                    f"{account.primary_smtp_address}, skipping."
+                )
+                continue
 
             for mail in await asyncio.to_thread(folder_object.all().only, *MAIL_FIELDS):
                 yield mail, mail_type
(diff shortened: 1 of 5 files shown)


app/connectors_service/connectors/sources/outlook/client.py
```diff
@@ -405,15 +405,19 @@ async def get_mails(self, account):
...
-            else:
-                folder_object = getattr(account, mail_type["folder"])
+                else:
+                    folder_object = getattr(account, mail_type["folder"])
+            except ErrorFolderNotFound:
+                self._logger.warning(
+                    f"Could not resolve {mail_type['folder']} folder for "
+                    f"{account.primary_smtp_address}, skipping."
+                )
+                continue
```
artem-shelkovnikov, line 420: Is this one somehow related to https://github.com/ecederstrand/exchangelib/issues/1155?

app/connectors_service/connectors/sources/outlook/datasource.py
```diff
@@ -89,7 +103,9 @@ def calendar_doc_formatter(self, calendar, child_calendar, timezone):
                     "attendees": [
                         attendee.mailbox.email_address
                         for attendee in (calendar.required_attendees or [])
-                        if attendee.mailbox.email_address
+                        if attendee
```
artem-shelkovnikov, line 106: Can `attendee` be null?

It's a curiosity question here, would be weird but not unexpected 🤷

artem-shelkovnikov (approved):

app/connectors_service/connectors/sources/outlook/datasource.py
Jan-Kazlouski-elastic: IRL - no, i don't think so.
exchangelib populates `required_attendees` with `Attendee` objects, not `None`. The real guards are `attendee.mailbox` & `.email_address` ones.

`if attendee` is for symmetry and extra safety.

app/connectors_service/connectors/sources/outlook/client.py
Jan-Kazlouski-elastic: Could be related, yes. But I'm not positive.
In #4065 Archive was already handled because its leaf has no distinguished ID and is resolved by name, which breaks on localized servers. inbox/sent/junk resolve via distinguished IDs and I have no field reports of them being missing. Wrapping them here is strictly precautionary. [#1155](https://github.com/ecederstrand/exchangelib/issues/1155) was closed as a transient Microsoft-side bug, so this isn't a direct fix for it; the trade-off is just a warning + skipped folder instead of aborting the whole account.

PR #4133, opened 2026-07-07
[8.19] fix(outlook): harden sync against missing Exchange field values (#4123)

Jan-Kazlouski-elastic:
Backports #4123 to `8.19`.

## Summary
- Handle nullable mail, calendar, contact, and attachment fields so a single malformed Exchange item no longer aborts the sync (fixes the customer `'NoneType' object has no attribute 'email_address'` failure on `mail.sender`).
- Skip optional mail folders (inbox/sent/junk/archive) when absent instead of failing the whole account.
- Harden on-prem DLS access-control sync for list-valued LDAP `mail` attributes and malformed DNs.

## Backport notes
- The 8.19 branch still uses the monolithic `connectors/sources/outlook.py` layout (rather than the `outlook/` package split into `client.py`/`datasource.py` on `main`), so the changes were manually adapted to that single module. Behaviour is identical to the original PR.
- The incidental `NOTICE.txt` dependency bump from the original PR (unrelated `charset-normalizer` version) is intentionally excluded.

## Test plan
- [x] `make clean install autoformat lint test PYTHON=python3.11` (the only failures are unrelated `fail-slow` timing flakes in s3 / google_drive / google_cloud_storage / kibana; all tests otherwise pass)
- [x] `pytest tests/sources/test_outlook.py` (57 passed)

### DIFF
--- a/connectors/sources/outlook.py
+++ b/connectors/sources/outlook.py
@@ -489,6 +489,14 @@ class Office365Users:
 class OutlookDocFormatter:
     """Format Outlook object documents to Elasticsearch document"""
 
+    @staticmethod
+    def _calendar_meeting_type(calendar):
+        if calendar.type == "Single":
+            return "Single"
+        if calendar.recurrence and calendar.recurrence.pattern:
+            return f"Recurring {calendar.recurrence.pattern}"
+        return calendar.type
+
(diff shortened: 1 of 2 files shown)


erikcurrin-elastic (approved):
Simple safe-guards, look good

PR #4137, opened 2026-07-07
[Connectors Python] Lower default max file download size from 10MiB to 8MiB

Jan-Kazlouski-elastic:
Lowers the default `service.max_file_download_size` from 10MiB (10485760) to 8MiB (8388608).

Aligns the connectors default with the new 8MiB `ingest.attachment.max_field_size` cap that Elasticsearch Serverless enforces on the Ingest Attachment Processor.

Changes:
- `connectors/config.py`: `DEFAULT_MAX_FILE_SIZE` → 8388608
- `config.yml.example`: `service.max_file_download_size` → 8388608

## Release Note

Lowered the default maximum file download size (`service.max_file_download_size`) from 10MiB to 8MiB.

### DIFF
--- a/app/connectors_service/NOTICE.txt
+++ b/app/connectors_service/NOTICE.txt
@@ -4146,7 +4146,7 @@ Apache Software License
 
 
 google-auth
-2.55.1
+2.55.2
 Apache Software License
                                  Apache License
                            Version 2.0, January 2004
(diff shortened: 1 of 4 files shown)


seanstory:
Please add a release notes section to the PR description, and the associated gh label. Otherwise, LGTM

erikcurrin-elastic (approved):
looks good, but we need a doc change also

seanstory:
Oh, and we probably shouldn't backport this, as it's not strictly a bugfix.

PR #4147, opened 2026-07-09
fix(outlook): dispatch Contacts by item type and harden folder/field assumptions

Jan-Kazlouski-elastic:

Follow-up to #4123. The last month has been a series of "the next malformed item aborts the whole sync" reports, most recently `'DistributionList' object has no attribute 'email_addresses'`. This PR fixes that class of failure with explicit, visible handling, and audits the connector for the remaining assumptions of the same shape.

## Root cause

The Contacts folder returns two item models — `Contact` and `DistributionList` — but the formatter assumed every item was a `Contact` and read per-contact fields off it, so a single contact group aborted the entire sync. The same "unhandled shape aborts everything" gap existed elsewhere: Calendar/Tasks folders that shared and resource mailboxes can legitimately lack, birthday events without a start, and an optional LDAP `type` key.

## Changes

**Contacts formatted by item type.** `_fetch_contacts` now dispatches on type: a `Contact` goes through the strict `contact_doc_formatter` and a contact group through a dedicated `distribution_list_doc_formatter` that indexes it by name and its members' email addresses (typed `"Distribution List"`). The folder query fetches the union of both models' fields (incl. `members`). The Contacts folder contractually returns only these two models, so any other type raises a `TypeError` rather than being silently skipped — an unknown type is a broken assumption we want surfaced, not a whole category of items quietly dropped. (Thanks @artem-shelkovnikov for the review — explicit type handling over defensive `getattr`, and failing on unexpected types.)

**Note on approach.** An earlier revision of this PR wrapped all formatting in a broad `_format_item` guard that caught item-shape errors (`AttributeError`/`KeyError`/`ValueError`/`TypeError`) and skipped the item. That was dropped: swallowing those errors around every item can turn a systematic bug into a "successful" sync that returns 0 items and deletes previously indexed documents. Instead, known shapes are handled explicitly (type dispatch + targeted field guards) and genuine errors still fail the sync loudly — consistent with the principle established in #4085/#4123.

**Remaining assumptions found in the audit, now fixed:**

* **Calendar & Tasks folders** are now skipped on `ErrorFolderNotFound`, mirroring the existing Contacts/mail-folder handling. Shared and resource mailboxes can legitimately lack these folders; previously that aborted the sync (same class as #4065).
* **Birthdays calendar branch** called `.split("T")` on the formatted datetime, which is `None` when `calendar.start` is missing → `AttributeError`. Now guarded.
* **Optional LDAP `type` key** is read with `.get(...)` instead of `user["type"]`.

## Test plan

* `pytest tests/sources/test_outlook.py` (65 tests), incl. coverage for:
   * contact / `DistributionList` type dispatch and the group formatter
   * unexpected Contacts item type → raises `TypeError`
   * `ErrorFolderNotFound` skips for Calendar / child calendars / Tasks
   * birthday without a start
   * missing LDAP `type` key
* `ruff check` / `ruff format --check`


app/connectors_service/connectors/sources/outlook/datasource.py
```diff
@@ -145,6 +145,9 @@ def task_doc_formatter(self, task, timezone):
         }
 
     def contact_doc_formatter(self, contact, timezone):
+        # The Contacts folder can also return contact groups (DistributionList),
+        # which lack the per-contact fields below. Read them defensively so a
+        # single group no longer aborts the whole sync with an AttributeError.
```
artem-shelkovnikov: Technically this change says different:

If an object returned by `contacts` does not have some attributes, skip them.

I'm wondering if there are other types of objects that can come here, and will these objects also have `display_name`?

Also I'm wondering if we should instead check the object type and return formatted documents based on type of incoming object - it will make it easier to comprehend why some objects do not have all fields and which fields are supported per object.

Jan-Kazlouski-elastic: Good call, agreed on both points — I've reworked this to dispatch on item type instead.

To your questions: a Contacts-folder query in exchangelib returns exactly two item models — `Contacts.supported_item_models == (Contact, DistributionList)` — and both inherit `display_name` (plus `id`/`last_modified_time`) from the base `Item`, so the shared fields are always safe. `DistributionList` just doesn't have the per-contact fields (`email_addresses`, `phone_numbers`, `company_name`, `birthday`); instead it has `members`.

Changes in a9fb2124:
- Reverted `contact_doc_formatter` to strict `Contact` access (no more `getattr`), so it documents the real Contact schema.
- Added a dedicated `distribution_list_doc_formatter` that indexes the group by name and its members' email addresses, typed as `"Distribution List"`.
- Dispatch by `isinstance` in `_fetch_contacts`, and the query now fetches the union of both models' fields (incl. `members`).

The general per-item `_format_item` guard stays as defense-in-depth for genuinely unexpected shapes, but the contact path is now explicit about which fields belong to which type.

app/connectors_service/connectors/sources/outlook/client.py
```diff
@@ -423,20 +423,42 @@ async def get_mails(self, account):
                 yield mail, mail_type
 
     async def get_calendars(self, account):
-        for calendar in await asyncio.to_thread(
-            account.calendar.all().only, *CALENDAR_FIELDS
-        ):
+        # Skip instead of aborting if the mailbox has no Calendar folder.
+        try:
+            folder = account.calendar
```
artem-shelkovnikov: Sanity check - this does not produce network call, `folder.all()` does, right?

app/connectors_service/connectors/sources/outlook/datasource.py
```diff
@@ -537,10 +601,22 @@ async def _fetch_mails(self, account, timezone):
...
-                timezone=timezone,
-            )
+            # Contacts folder holds both contacts and groups; each has its own shape.
+            if isinstance(contact, DistributionList):
+                formatter = partial(
+                    self.doc_formatter.distribution_list_doc_formatter,
+                    distribution_list=contact,
+                    timezone=timezone,
+                )
+            else:
```
artem-shelkovnikov, line 572: I personally like doing explicit type checks and warn/raise on the case when the type is unexpected.

app/connectors_service/connectors/sources/outlook/datasource.py
```diff
@@ -488,6 +513,31 @@ async def download_func(self, content):
         """
         yield content
 
+    def _format_item(self, formatter, item, account):
```
artem-shelkovnikov: This actually has tendency to actually produce syncs that are successful but return 0 items. I'm not really sure we should do this.

We've tried an approach with this class: https://github.com/elastic/connectors/blob/0b1727c00854985e55c71503735f4d3cb03143cc/app/connectors_service/connectors/utils.py#L827

Error Monitor swallows errors up to a certain threshold and then raises an exception if too many errors happen. It also has its downsides.

artem-shelkovnikov: Continuing the train of thought from above:

I feel this method attempt to help not raise when formatting objects of unexpected types. E.g. you have an API returning types `contact` and `distribution_list` and you format them well, but then for some niche case the customer gets an error because an object of type `broadcast_group` (imaginary name) is returned and the method breaks.

So instead each formatter method can do type checks and warn or error if type is unexpected.

Warn or error? Not sure. It's all about expectation setting to me.

Warns are good because the sync does not crash
Errors are good because they tell what went wrong immediately, so you will know that something is not being synced.

Because of that Error Monitor will not actually help, but will just hide the problem.

All in all I think it's best to do type checks in formatter methods and warn on unexpected types. I am open to a discussion here, though

app/connectors_service/connectors/sources/outlook/client.py
Jan-Kazlouski-elastic: Hmm, actually it's the other way around:
`account.calendar` makes the network call. It is the same as `self.root.get_default_folder(Calendar)`, which calls `GetFolder` and raises `ErrorFolderNotFound`.

`folder.all()` on the other hand is a lazy `QuerySet`. No network call until it gets iterated, and no `ErrorFolderNotFound`. So we should be good here.

app/connectors_service/connectors/sources/outlook/datasource.py
Jan-Kazlouski-elastic: Thanks. Fixed. Now we're raising instead of silently skipping.

app/connectors_service/connectors/sources/outlook/datasource.py
Jan-Kazlouski-elastic: Done.

app/connectors_service/connectors/sources/outlook/datasource.py
Jan-Kazlouski-elastic: Update: 

Removed `_format_item` entirely rather than keeping it as defense-in-depth.

app/connectors_service/connectors/sources/outlook/client.py
artem-shelkovnikov: Okay, then there is a problem here. `exchangelib` is not async, and the call to `account.calendar` is gonna block the whole process until it resolves. Then we'd need to wrap it into `asyncio.to_thread` or something similar

artem-shelkovnikov: This then will have to be done to the methods that do network calls below unfortunately

Jan-Kazlouski-elastic: Yep. You're right. Done.

artem-shelkovnikov (approved):
lgtm

PR #3476, opened 2025-06-17
Add support for DatetimeMS in MongoDB

artem-shelkovnikov:
## Closes https://github.com/elastic/connectors/issues/3444

This PR adds support for `datetime` values that cannot be handled by python.

Before this change instances of `datetime` that are outside of `[datetime.min; datetime.max]` range will raise an error:

```
year 643385 is out of range (Consider Using CodecOptions(datetime_conversion=DATETIME_AUTO) or MongoClient(datetime_conversion='DATETIME_AUTO')). See: https://pymongo.readthedocs.io/en/stable/examples/datetimes.html#handling-out-of-range-datetimes
```

This PR makes use of `DatetimeConversion.DATETIME_AUTO` conversion setting.

In practice it will mean that:

- `datetime` objects that can be Python datetimes will be parsed as such and sent to Elasticsearch as datetimes
- If driver cannot parse the object as Python `datetime` it will be serlaised as a `long`

This, though, can be problematic due to internal Elasticsearch type conversion (something that is already a problem with MongoDB connector to some extent):

Elasticsearch allows conversion of  `long` to `datetime`, but does not allow the opposite.

Demonstration:

```
PUT connector-mongodb-f1e8/_doc/zdO8fZcBkReHFeyYyxGV
{
    "address": """3821 Jennifer Key
Victoriaville, MD 12105""",
    "birthdate": "1976-11-08",
    "unique_id": "46292e6c-5227-4715-879c-1932ccae9062",
    "some_small_datetime": -4611686018427388000, # this field is a long
    "some_zero_datetime": "1970-01-01T00:00:00", # this field is a datetime
    "name": "Earl Phillips",
    "comment": "Four tax per.",
    "id": "6851474d557891815b63f1e3",
    "time": "12:23:15",
    "some_large_datetime": 4611686018427388000,
    "fun_field": 4611686018427388000
}

OK

{
    "address": """3821 Jennifer Key
Victoriaville, MD 12105""",
    "birthdate": "1976-11-08",
    "unique_id": "46292e6c-5227-4715-879c-1932ccae9062",
    "some_small_datetime": -4611686018427388000, # this field is a long
    "some_zero_datetime": -4611686018427388000, # this field is a datetime
    "name": "Earl Phillips",
    "comment": "Four tax per.",
    "id": "6851474d557891815b63f1e3",
    "time": "12:23:15",
    "some_large_datetime": 4611686018427388000,
    "fun_field": 4611686018427388000
}

OK


{
[...]

Issue #3444, opened 2025-06-02
MongoDB connector - add support for DatetimeMS objects


mattnowzari:
### Problem Description

Currently, the MongoDB connector uses the stdlib `datetime` library to handle datetime objects from the Mongo client. We should expand on this handling to include handling of `datetimeMS` objects, which is an object that can represent dates that would otherwise be out-of-range for the stdlib `datetime` library.

See [this](https://pymongo.readthedocs.io/en/stable/examples/datetimes.html#handling-out-of-range-datetimes) article here about handling out-of-range datetimes with PyMongo.

### Proposed Solution
We should enhance the object serialization of the MongoDB connector to handle`datetimeMS` objects and not just `datetime` objects when dealing with dates.

https://github.com/elastic/connectors/blob/ae3f04a2a98f2f3bbea5ada32b65f2a130ef4ac3/connectors/sources/mongo.py#L226-L246

artem-shelkovnikov:
Update: a PR is there, but we haven't heard anything from the customer. We can merge it if we're confident enough it's a solid approach in the PR.

PR #4148, opened 2026-07-09
[MongoDB] Handle out-of-range BSON datetimes to prevent sync failures

Jan-Kazlouski-elastic:
## Summary

Documents containing dates outside the Python `datetime` range (years 1–9999) cause pymongo to raise `InvalidBSON` while decoding, which aborts the entire MongoDB sync:

```
year 643385 is out of range (Consider Using CodecOptions(datetime_conversion=DATETIME_AUTO) ...)
```

This PR configures the MongoDB client with `DatetimeConversion.DATETIME_CLAMP`, so out-of-range values are clamped to `datetime.min` / `datetime.max` and remain valid dates. They then serialize to normal ISO date strings, exactly like in-range dates.

## Why DATETIME_CLAMP (and not DATETIME_AUTO)

The earlier (closed) PR #3476 used `DATETIME_AUTO`, which keeps in-range values as dates but converts out-of-range values to `long` (epoch millis). That makes a single field hold **mixed** types (ISO string vs. long). Because Elasticsearch infers a field's type from the first document it sees, a later document of the other type is then rejected — turning the hard crash into an order-dependent ingestion failure.

`DATETIME_CLAMP` avoids this entirely: every date serializes as a date string, so there is one consistent mapping, no crash, and no breaking change for existing users. The trade-off is that extreme dates lose their exact value (clamped to the datetime bounds) — an acceptable outcome for what are almost always sentinel/garbage dates.

## Context

This revives and supersedes the work from the earlier (closed) PR #3476, which was never merged and predates the connector package refactor (it targeted the old single-file `connectors/sources/mongo.py`). This PR ports the fix to the current `connectors/sources/mongo/` package layout, switches the strategy to clamping, and adds unit test coverage.

- Original PR: #3476
- Original issue: #3444
- Support case: elastic/sdh-search#1924

## Changes

- `get_client()`: set `datetime_conversion=DatetimeConversion.DATETIME_CLAMP`.
- `serialize()`: no special date handling needed — clamped values are ordinary `datetime` objects and go through the existing `datetime -> isoformat()` branch.
- Added unit tests: in-range and clamped-bound datetimes serialize to ISO strings, a client-config test asserting `DATETIME_CLAMP`, and an end-to-end test that encodes an out-of-range BSON date, decodes it with the client codec, and confirms it becomes a clamped `datetime` serialized as an ISO string (never a `DatetimeMS` or a number).

## Test plan

- [x] `make clean install autoformat lint test PYTHON=python3.11` (all MongoDB tests pass; unrelated pre-existing flaky failures only)
- [ ] Verify against a MongoDB collection containing out-of-range dates

## Release Note

MongoDB connector: set default `datetime_conversion` to `DatetimeConversion.DATETIME_CLAMP` so out-of-range `datetime` values from MongoDB are clamped to valid dates instead of aborting the sync. See https://www.mongodb.com/docs/languages/python/pymongo-driver/current/data-formats/dates-and-times/#handling-out-of-range-datetimes for additional information.


artem-shelkovnikov:
Just checking - how is it related to the support case?

Jan-Kazlouski-elastic:
> Just checking - how is it related to the support case?

Not directly. Looking into the support case just triggered agent to look into related issues and this was found. I decided it would be a good easy drive by fix.

erikcurrin-elastic (approved):

Jan-Kazlouski-elastic:
As discussed offline with @artem-shelkovnikov , it was decided to make the logic of handling the out of bounds dates configurable.
Will push the change shortly.

Jan-Kazlouski-elastic:
@erikcurrin-elastic thanks for the approval. We've had a chat with @artem-shelkovnikov today and decided it would be a good thing to have this behavior configurable. 
New `datetime_conversion` config controls how MongoDB dates outside the supported range (years 1–9999) are handled. Defaults to `DATETIME` (legacy: raise/fail the sync). Opt-in alternatives: `DATETIME_CLAMP` (clamp to min/max date), `DATETIME_AUTO` (out-of-range dates → epoch ms), `DATETIME_MS` (all dates → epoch ms).

erikcurrin-elastic (approved):
The configuration option is definitely much better.  Good test coverage on what threw errors and no longer does

artem-shelkovnikov (approved):
LGTM! Imo this needs to be backported as far as possible

PR #4150, opened 2026-07-10
fix: allow CLI help without config

Saithej2k:
## Summary

- Defer configuration loading until a leaf command is invoked.
- Allow top-level, group, and nested command help without a local config file.
- Cover every connector, index, and job help route.

## Context

Follow-up to #4109. This addresses the review finding that parent group callbacks loaded config before Click processed a leaf command help option. The unused parent-context help check is removed.

## Testing

- 53 CLI tests passed
- Ruff passed for the changed source and test files
- git diff --check

### DIFF
--- a/app/connectors_service/connectors/connectors_cli.py
+++ b/app/connectors_service/connectors/connectors_cli.py
@@ -14,6 +14,7 @@ executes the `main` function of this module, which starts the service.
 import asyncio
 import json
 import os
+from functools import wraps
 
 import click
 import yaml
@@ -32,17 +33,45 @@ from connectors.es import DEFAULT_LANGUAGE
 __all__ = ["main"]
 
 
-def load_config(ctx, config):
+def load_config(config):
     if config:
         return yaml.safe_load(config)
     elif os.path.isfile(CONFIG_FILE_PATH):
         with open(CONFIG_FILE_PATH, "r") as f:
             return yaml.safe_load(f.read())
-    elif ctx.invoked_subcommand == "login":
-        pass
     else:
         msg = f"{CONFIG_FILE_PATH} was not found."
         raise FileNotFoundError(msg)
+
+
+def ensure_config(ctx):
+    ctx.ensure_object(dict)
+    if "config" in ctx.obj:
+        return
+
+    try:
+        ctx.obj["config"] = load_config(ctx.obj.get("config_file"))
+    except FileNotFoundError as e:
+        click.echo(
+            f"{e} Make sure that the config is either present at the default location ({CONFIG_FILE_PATH}) or it's passed via the '-c' or '--config' option."
+        )
+        ctx.exit(1)
+
+
+def eager_config_check(ctx, _param, value):
+    if not ctx.resilient_parsing:
+        ensure_config(ctx)
+    return value
+
+
+def requires_config(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        ctx = click.get_current_context()
+        ensure_config(ctx)
+        return func(ctx.obj, *args, **kwargs)
+
+    return wrapper
 
 
 # Main group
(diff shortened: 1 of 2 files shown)


app/connectors_service/connectors/connectors_cli.py
```diff
@@ -45,6 +46,27 @@ def load_config(ctx, config):
         raise FileNotFoundError(msg)
 
 
+def ensure_config(ctx):
+    ctx.ensure_object(dict)
+    try:
+        ctx.obj["config"] = load_config(ctx, ctx.obj.get("config_file"))
+    except FileNotFoundError as e:
+        click.echo(
+            f"{e} Run `connectors login` first, or pass a config via the '-c' or '--config' option."
```
Jan-Kazlouski-elastic: The actual fix in this PR is the lazy config loading. It makes `--help` work without a config. The reworded not-found message, however, affects the real-command path (e.g. `connectors job list <id>`), where `--help` isn't involved.

Login isn't strictly required to have a config — -c/--config or a file at the default location work without ever logging in (login just writes that file). Leading the error with "Run connectors login first" narrows the meaning a bit. Could we keep the original wording to keep the diff focused on the bug?

CC @artem-shelkovnikov

app/connectors_service/connectors/connectors_cli.py
Jan-Kazlouski-elastic: I believe this can be dropped now?

app/connectors_service/tests/test_connectors_cli.py
```diff
@@ -128,6 +128,45 @@ def test_connector_help_page():
     assert "Commands:" in result.output
 
 
+@pytest.mark.parametrize(
```
Jan-Kazlouski-elastic, line 131: Great job on covering the logic. But the new tests cover that help works without a config. Could we also add the other half — a real command without a config should exit non-zero and report the missing file?

app/connectors_service/connectors/connectors_cli.py
```diff
@@ -235,7 +250,7 @@ def interactive_service_type_prompt():
     prompt=f"{click.style('?', fg='green')} Connector name",
     help="Connector name",
 )
-@click.pass_obj
+@requires_config
 def create(
```
Jan-Kazlouski-elastic, line 269: `create` is the only config-requiring command with `prompt=` options, and Click resolves prompts during parsing. Before `requires_config` runs. So without a config the user answers every prompt (index name, service type menu, …) and only then hits the "not found" error; on `main` - it failed fast.

Suggestion: run the check before the prompts via a hidden pre-check option, and make `ensure_config` idempotent so the later `requires_config` call is a no-op.

Jan-Kazlouski-elastic (commented):
Great job @Saithej2k
Changes look good.
A few small suggestions to take care of and we'll be good to merge it!
Would you please address these?

app/connectors_service/connectors/connectors_cli.py
Saithej2k: Restored the original missing-config wording and kept the error path focused in 0384135f.

app/connectors_service/connectors/connectors_cli.py
Saithej2k: Removed the now-unnecessary login special case from `load_config` in 0384135f.

app/connectors_service/tests/test_connectors_cli.py
Saithej2k: Added regression coverage verifying that a real command exits non-zero and reports the missing config file.

app/connectors_service/connectors/connectors_cli.py
Saithej2k: Added a hidden eager pre-check to `connector create` and made `ensure_config` idempotent, so missing config fails before prompts and the command callback does not reload it.

Jan-Kazlouski-elastic: eager config pre-check runs before --help is processed and it re introduces the bug we're fixing.

Jan-Kazlouski-elastic:
buildkite test this

PR #4158, opened 2026-07-13
fix(outlook): skip unexpected item types and harden folder/account error handling

Jan-Kazlouski-elastic:

Follow-up to #4147. That PR fixed the Contacts folder's single-item-model assumption; this one applies the same "one malformed item/folder must not abort the whole sync" principle to every remaining folder-iteration and error-handling site the audit surfaced, most recently the `'CalendarItem' object has no attribute 'sender'` / localized-`Archive` class of reports.

## Root cause

The connector iterated Exchange folders assuming each folder returns exactly one item model, that every distinguished folder resolves, and that the only recoverable per-folder fault is `ErrorFolderNotFound`. Real mailboxes violate all three: a Calendar/Tasks/mail folder can contain stray item types, a child folder under Calendar may not be a calendar folder (so the `CALENDAR_FIELDS` projection raises `ValueError`), folder resolution can fail with `ErrorManagedFolderNotFound`, an inaccessible mailbox raises `ErrorAccessDenied`, an attachment can be an embedded `ItemAttachment` (no `content`) or lack a `size`, and the timezone fallback was a bare string that the datetime formatter can never use. Each of these aborted the entire sync (or mis-typed an item) instead of being handled visibly.

## Changes

**Type allowlists on every folder iteration.** `_fetch_mails`, `_fetch_tasks`, `_enqueue_calendars`, and `_fetch_contacts` now route/skip by type: mail folders accept `Message`/`MeetingRequest`/`MeetingResponse`/`MeetingCancellation`, Tasks accept `Task`, Calendars accept `CalendarItem`, and Contacts route `Contact` vs `DistributionList`. A stray type is logged and skipped rather than crashing on a missing field. Contacts now skip-and-warn on an unexpected type (rather than raising `TypeError` as in #4147) for consistency with the other folders — the warning keeps it visible, and a single odd item no longer aborts the sync.

**Broadened per-folder skips.** `get_mails`/`get_calendars`/`get_child_calendars`/`get_tasks`/`get_contacts` skip on `FOLDER_SKIP_ERRORS` (`ErrorFolderNotFound` + `ErrorManagedFolderNotFound`) instead of only `ErrorFolderNotFound`.

**Child-calendar projection guard.** A non-calendar child folder under Calendar can't be projected with `CALENDAR_FIELDS` and raises `ValueError`; that child is now skipped (`CHILD_CALENDAR_SKIP_ERRORS`) while healthy sibling calendars still sync.

**Account-level access errors.** `get_docs` now skips an account on `ErrorAccessDenied` in addition to `ErrorNonExistentMailbox`. Connection-wide `TransportError`s (e.g. TLS failures) still propagate — silently swallowing them across all accounts would yield a doc-less "successful" sync that deletes previously indexed documents (the principle from #4085/#4123).

**Attachment guards.** `get_content` only extracts `FileAttachment` content (embedded `ItemAttachment` has no `content`) and guards the optional `size`, which may be `None`.

**Timezone fallback.** Replaced the dead `DEFAULT_TIMEZONE = "UTC"` string fallback with the real `exchangelib.UTC` (`EWSTimeZone`), so the `account.default_timezone or UTC` path actually formats datetimes if a mailbox ever lacks a default timezone.

## Test plan

* `pytest tests/sources/test_outlook.py` (92 tests), incl. coverage for:
  * `FOLDER_SKIP_ERRORS` (`ErrorFolderNotFound` + `ErrorManagedFolderNotFound`) skips across mails / calendars / child calendars / tasks / contacts
  * child-calendar `ValueError` / managed-folder skip, with a healthy sibling still syncing
  * account skip on `ErrorAccessDenied`, and connection-wide errors still re-raised
  * every type in `MAIL_ITEM_TYPES` accepted, and stray non-mail / non-task / non-calendar / unexpected-contact items skipped-and-warned
  * `get_content` skips `ItemAttachment` and `None` / `0` size
  * UTC timezone fallback when `default_timezone` is missing
* `ruff check` / `ruff format --check`
* 97% line coverage on `client.py` and `datasource.py`; every changed line exercised

## Release Note

Harden the Outlook connector so unexpected Exchange item types, unresolvable or inaccessible folders, embedded/size-less attachments, and missing mailbox timezones are skipped with a warning instead of aborting the sync.


artem-shelkovnikov (approved):
LGTM, sorry for the delay, overlooked the ping

PR #4253, opened 2026-07-21
Return back snyk scanning

artem-shelkovnikov:
## Closes https://github.com/elastic/search-team/issues/15280

This PR returns Snyk scanning back to the table. How it's supposed to work:

1. New `make freeze` action freezes the content of virtual environment for the package into a `requirements.txt` that can be scanned. We clean up entities that are not supported by general pip snyk scanner, such as git references or relative file references
2. We enable snyk scanning on nightlies on `main`. `./snyk monitor` should actually report scan issues in the Snyk UI
3. We package a requirements.json file into docker images

We need to merge this PR and see what Snyk actually reports as some stuff is hard to confirm (namely Docker image scanning)

### DIFF
--- a/.buildkite/nightly_steps.yml
+++ b/.buildkite/nightly_steps.yml
@@ -48,7 +48,7 @@ steps:
     command: ".buildkite/run_snyk.sh"
     env:
       PYTHON_VERSION: "3.11"
-    timeout_in_minutes: 5
+    timeout_in_minutes: 15
 
   - label: "🔨 [Python {{ matrix }}] MySQL"
     <<: *retries
--- a/.buildkite/nightly_steps.yml
+++ b/.buildkite/nightly_steps.yml
@@ -1,4 +1,11 @@
 definitions:
+  steps:
+    - step: &test-agents
+        agents:
+          provider: "gcp"
+          machineType: "n1-standard-8"
+          useVault: true
+          image: family/enterprise-search-ubuntu-2204-connectors-py
   retries: &retries
     retry:
       automatic:
@@ -31,6 +38,18 @@ env:
   DOCKERFILE_PATH: "Dockerfile.wolfi"
 
 steps:
+
+  - label: ":snyk: Snyk scan"
+    if: 'build.branch == "main"'
+    <<: *test-agents
+    <<: *retries
+    key: "snyk"
+    soft_fail: true
+    command: ".buildkite/run_snyk.sh"
+    env:
+      PYTHON_VERSION: "3.11"
+    timeout_in_minutes: 5
+
   - label: "🔨 [Python {{ matrix }}] MySQL"
     <<: *retries
     <<: *ftest_defaults
--- a/.buildkite/run_snyk.sh
+++ b/.buildkite/run_snyk.sh
@@ -5,6 +5,8 @@ set -euo pipefail
 
 source .buildkite/shared.sh
 
+BASEDIR=$(realpath $(dirname $0))
+ROOT=$(realpath $BASEDIR/../)
 APP_ROOT=$ROOT/app/connectors_service
 SDK_ROOT=$ROOT/libs/connectors_sdk
 
--- a/.buildkite/run_snyk.sh
+++ b/.buildkite/run_snyk.sh
@@ -23,9 +23,10 @@ curl -sL --retry-max-time 60 --retry 3 --retry-delay 5 https://static.snyk.io/cl
 chmod +x ./snyk
 
 echo "--- Initializing venv for the test ---"
-python3 -m .snyk-venv
+python3 -m /tmp/.snyk-venv
+
 echo "--- Installing dependencies ---"
-$ROOT/.snyk-venv/bin/pip install \
+/tmp/.snyk-venv/bin/pip install \
         -r $SDK_ROOT/requirements.txt \
         -r $APP_ROOT/requirements.txt
 
--- a/.buildkite/run_snyk.sh
+++ b/.buildkite/run_snyk.sh
@@ -24,8 +24,10 @@ chmod +x ./snyk
 
 echo "--- Initializing venv for the test ---"
 python3 -m .snyk-venv
-$ROOT/.snyk-venv/bin/pip install --quiet -r $SDK_ROOT/requirements.txt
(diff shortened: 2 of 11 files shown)


.buildkite/nightly_steps.yml
```diff
@@ -31,6 +38,18 @@ env:
   DOCKERFILE_PATH: "Dockerfile.wolfi"
 
 steps:
+
+  - label: ":snyk: Snyk scan"
+    if: 'build.branch == "main"'
```
artem-shelkovnikov, line 43: We've always reported Snyk issues only on `main`, so keeping it here. We can do on branches, but need to do testing on how it's actually gonna look in Snyk

.buildkite/run_snyk.sh
```diff
@@ -0,0 +1,39 @@
...
+APP_ROOT=$ROOT/app/connectors_service
+SDK_ROOT=$ROOT/libs/connectors_sdk
+
+
+init_python
+
+make clean install freeze
+
+echo "--- Logging into snyk...---"
+export SNYK_TOKEN=$(vault read -field=value secret/ci/elastic-connectors/SNYK_TOKEN)
```
artem-shelkovnikov, line 19: Anyone in `search-extract-and-transform` should have access here

.buildkite/run_snyk.sh
```diff
@@ -0,0 +1,39 @@
...
+
+echo "--- Downloading snyk..."
+curl -sL --retry-max-time 60 --retry 3 --retry-delay 5 https://static.snyk.io/cli/latest/snyk-linux -o snyk
+chmod +x ./snyk
+
+echo "--- Initializing venv for the test ---"
```
artem-shelkovnikov, line 25: Snyk uses its own pip so we really need to have a clean environment where we install our requirements to to do the scanning.

Running against our app/lib will cause in errors because Snyk's pip cannot resolve relative file paths (we use connectors_sdk as a relative path dependency)

erikcurrin-elastic (approved):
This is awesome.

app/connectors_service/Makefile
```diff
@@ -82,6 +82,15 @@ autoformat: .venv/bin/python .venv/bin/ruff .venv/bin/elastic-ingest
...
 	.venv/bin/pytest --cov-report term-missing --cov-fail-under 90 --cov-report html --cov=connectors --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests
 
+freeze: .venv/bin/python
+	# It looks a little insane, but hear me out.
+	# Snyk CLI does not use local PIP, it runs its own copy of it and does not really respect some things that we have in pyproject.toml
+	# For example relative file paths are not allowed, or evaluation of env variables are not possible while allowed.
+	# So to make things work we need to convert our requirements to requirements.txt file so that snyk could run its scanning against it
+	# Pipdeptree is able to actually produce requirements.txt file that contains both direct and transitive dependencies of the service
+	# So when included to a docker artefact, for example, it should give accurate description of what's actually installed into the .venv
+	.venv/bin/pip install --quiet pipdeptree
```
Jan-Kazlouski-elastic: Wouldn't installing **`pipdeptree`** include it into the shipped Docker image as well? Introducing dev-time dependencies into runtime, and also being reported in requirements.txt, making **`snyk`** reporting any CVE in **`pipdeptree`**?
Keep me honest here.

.buildkite/nightly_steps.yml
```diff
@@ -31,6 +38,18 @@ env:
...
+    <<: *test-agents
+    <<: *retries
+    key: "snyk"
+    soft_fail: true
+    command: ".buildkite/run_snyk.sh"
+    env:
+      PYTHON_VERSION: "3.11"
+    timeout_in_minutes: 5
```
Jan-Kazlouski-elastic: Isn't 5 minutes way too little for the amount of work that run_snyk.sh does? It might be just enough, I just want to be sure this 5 minutes value was evaluated and we won't be hitting timeouts.

libs/connectors_sdk/Makefile
```diff
@@ -43,6 +43,11 @@ lint: install .venv/bin/ruff typecheck
 test: .venv/bin/pytest
 	$(VENV_DIR)/pytest --cov-report term-missing --cov-fail-under 80 --cov-report html --cov=connectors_sdk --fail-slow=$(SLOW_TEST_THRESHOLD) -sv tests
 
+freeze: .venv/bin/python
+	.venv/bin/pip freeze | \
+		grep -v '^-e' \
+		> requirements.txt
```
Jan-Kazlouski-elastic, line 58: It feels a bit fragile. Could we leverage the same approach as in main `Makefile`? with `pipdeptree exclude`?

Jan-Kazlouski-elastic (commented):
Left a few comments. 

Also I might be confused, but perhaps we could do something similar to `trivy image` run? `snyk container monitor` seemingly does the same check for the whole image, if I understand correctly. Keep in mind I may be missing something

.buildkite/nightly_steps.yml
artem-shelkovnikov: Good call, bumped to 15 minutes. It typically takes less than 3 minutes to execute, but having a bit of overhead is good.

app/connectors_service/Makefile
artem-shelkovnikov: Yeah, good call, I've added removal of `pipdeptree` on the next step of `make freeze`

libs/connectors_sdk/Makefile
artem-shelkovnikov: Yup, replaced with usage of `pipdeptree` from the other makefile

artem-shelkovnikov:
@Jan-Kazlouski-elastic the situation is that Snyk is our main source of issues in Github - there's automation that's tracking those CVEs, creating issues and such. Trivy doesn't do it.

Also Snyk is responsible for RM release blocks when our SLOs are violated, not Trivy.

My current understanding is `trivy` is used for docker images and is _also_ a good source of CVEs. Snyk finds some, trivy finds some other.

Jan-Kazlouski-elastic (approved):
LGTM!
My previous review-wide comment is more of a general direction, rather than comment that would keep us from merging. Good job!

artem-shelkovnikov:
Removed 8.19 as a target because project structure was different back then and requirements.txt were present

PR #4255, opened 2026-07-22
Fix/4139 sharepoint license error spam 

falscherwiener1-svg:
## What

Only check the Elasticsearch license for access control sync scheduling when the user has actually enabled access control sync — not just because the connector supports DLS.

Closes #4139

## Why

The scheduling loop calls `document_level_security_enabled()` which returns `True` for any connector that *supports* DLS (e.g. SharePoint Online). On a Basic license this logs an ERROR every ~30s:

> `Minimum required Elasticsearch license: 'platinum'. Actual license: 'basic'. Skipping access control sync scheduling...`

This is misleading when the user never enabled access control sync. The error should only appear when someone explicitly enables it without having the right license.

## Fix

Added `connector.access_control_sync_scheduling.get("enabled", False)` as an additional condition before entering the license check.

## Checklist

- [x] Unit tests pass
- [ ] Changelog not needed (fix not yet released)

PR #4256, opened 2026-07-22
Fix/3994 sharepoint incremental sync skip sites

### LINKED ISSUE #3994: [SharePoint Online] Incremental sync skips sites by timestamp, silently dropping drive item deletions and modifications

falscherwiener1-svg:
## What

Set `check_timestamp=False` on the `sites()` call in `get_docs_incrementally` so that all sites are visited during incremental sync.

Closes #3994

## Why

Drive item deletions and modifications do not reliably update the parent site's `lastModifiedDateTime` in the Graph API. With `check_timestamp=True`, entire sites are skipped — deleting drive items and modifications are silently lost until the next full sync.

## How

The codebase already uses this pattern for `site_drives()` (line 705):
```python
# Edit operation on a drive_item doesn't update the
# lastModifiedDateTime of the parent site_drive. Therefore, we
# set check_timestamp to False when iterating over site_drives.

Checklist
[ X ] Unit tests pass

### CODE REVIEW

**Comment by erikcurrin-elastic** [comment]
This PR incorporates the previous two (4254 and 4255).  I assume you only want review of this one?

File: app/connectors_service/connectors/es/sink.py [hunk 1]
```diff
@@ -264,6 +264,113 @@ def _map_id_to_op(self, operations):
...
+            data = item[action]
+            status = data.get("status", 200)
+            if status in TRANSIENT_STATUS_CODES:
+                doc_id = data.get("_id")
+                if doc_id in ops_by_id:
+                    failed_ops.extend(ops_by_id[doc_id])
+
+        return failed_ops
+
+    async def _retry_transient_item_failures(self, operations, pipeline_name):
```
erikcurrin-elastic, line 327: loses first-call successes from the return value. After a retry, res contains only the retried items. _batch_bulk then calls _process_bulk_response(res, ids_to_ops) where ids_to_ops maps all original operations, but res only has the retried subset. Items that succeeded on the first call won't be tallied in stats. Was that intentional?

app/connectors_service/connectors/services/job_scheduling.py
```diff
@@ -127,7 +127,10 @@ async def _schedule(self, connector):
         finally:
             await data_source.close()
 
-        if connector.features.document_level_security_enabled():
+        if (
+            connector.features.document_level_security_enabled()
+            and connector.access_control_sync_scheduling.get("enabled", False)
```
erikcurrin-elastic, line 132: Guard against access_control_sync_scheduling not being set

app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
```diff
@@ -683,10 +683,13 @@ async def get_docs_incrementally(self, sync_cursor, filtering=None):
...
             yield site_collection, None, OP_INDEX
 
+            # Edit operation on a drive_item doesn't update the
+            # lastModifiedDateTime of the parent site. Therefore, we
+            # set check_timestamp to False when iterating over sites.
             async for site in self.sites(
                 site_collection["siteCollection"]["hostname"],
                 self.configuration["site_collections"],
-                check_timestamp=True,
+                check_timestamp=False,
```
erikcurrin-elastic, line 692: Do we have to worry about site_lists and site_list_items also?

app/connectors_service/connectors/es/sink.py
falscherwiener1-svg: Good catch, that was not intentional. Items that succeeded on the first call were silently dropped from the returned response, which meant `_process_bulk_response` and `_populate_stats` never saw them.

I have updated `_retry_transient_item_failures` to track successful items across all calls and return a merged response containing both the originally successful items and the final retry outcome. See commit 564533c.

app/connectors_service/connectors/services/job_scheduling.py
falscherwiener1-svg: Yep, that is the intent. This guards against the case where DLS is enabled but access control sync scheduling has not been explicitly turned on. The `.get("enabled", False)` defaults to off, so the license check is only hit when the feature is actually configured.

app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
falscherwiener1-svg: Good question. `site_lists` and `site_list_items` do not need this change, the issue is specific to the drive item path.

Drive items are fetched via delta links (Graph API delta queries, see `self.client.drive_items()` at line 717). The parent site/site_drive are only traversal containers to reach the drives. Since editing a drive_item does not update the parent `lastModifiedDateTime`, we must skip the timestamp check to always reach the delta-enabled drives.

Lists, list items, and pages do not use delta queries, they are fetched directly and rely on `lastModifiedDateTime` comparison against `self.last_sync_time()`. Modifying a list item does update the list item own `lastModifiedDateTime`, and creating/updating a list does update the list `lastModifiedDateTime`. So `check_timestamp=True` is correct for those.

The same reasoning already existed for `site_drives()` (line 705 comment), this PR just applies the same logic one level up to `sites()`.

falscherwiener1-svg:
@erikcurrin-elastic Yes, that is correct. PRs #4254 and #4255 have been closed in favor of this single PR which incorporates all changes. Please review only this one (#4256).

I have also addressed the retry feedback in `sink.py` — see commit 564533c and the inline reply.

PR #4261, opened 2026-07-22
Backport freezing to 8.19

artem-shelkovnikov:
Manual backport of https://github.com/elastic/connectors/pull/4253

Since we don't do any scanning of deps on `main` we only needed a mainfest in our docker image.

This change in essence just makes it so that docker images emit `requirements.txt` during build that are just a snapshot of all current dependencies. Then Snyk will be able to see it (it's pretty silly and requires `requirements.txt` or `requirements-*.txt` or `requirements_*.txt` to be present to scan).

Fun side problem: our dependencies differ per architecture. The approach in this PR works in a way that _current_ docker image dependencies are frozen, so `requirements.txt` will be different per CPU architecture, but it's okay.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ scripts/stack/connectors-config
 *credentials*
 *secret*
 kubeconfig
+
+# This file is actually not real, it's derived and should only be present in the docker images as a "lock" file declaring versions that are used
+requirements.txt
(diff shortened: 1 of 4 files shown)


Jan-Kazlouski-elastic (approved):
lgtm

PR #4287, opened 2026-07-28
fix(outlook): skip unrecognised EWS elements instead of aborting the sync

Jan-Kazlouski-elastic:
## Part of https://github.com/elastic/sdh-search/issues/1898

Follow-up to #4158. The reporter's Exchange server aborts every sync with:

```
ValueError: Item type {http://schemas.microsoft.com/exchange/services/2006/types}EndTimeZone was unexpected in a BaseFolder folder
```

### Root cause

`CALENDAR_FIELDS` asks for `start` and `end`, and exchangelib silently appends two more fields to that projection — `calendar:StartTimeZone` and `calendar:EndTimeZone` (`BaseFolder.normalize_fields`). The server then returns the `EndTimeZone` element as a sibling of the `CalendarItem` it belongs to, inside the response's item container, rather than nested within it.

exchangelib treats every element of that container as an item, looks its tag up in `BaseFolder.ITEM_MODEL_MAP`, and raises `ValueError` when it is absent. This is the same failure shape as ecederstrand/exchangelib#877 (`Booking` items on servers with Microsoft Bookings installed).

Two things make it fatal rather than merely noisy:

* The exception is raised while `list(folder.all().only(*CALENDAR_FIELDS))` is materialised in `get_calendars`, so the per-folder type allowlists added in #4158 never see any items — the entire folder is lost, not one element.
* `get_docs` skips an account only on `ErrorNonExistentMailbox` / `ErrorAccessDenied`, so a bare `ValueError` propagates and aborts the whole sync.

### Change

`BaseFolder.ITEM_MODEL_MAP` is replaced with a `dict` subclass whose `__missing__` returns the generic `Item` and logs a warning naming the exact tag — once per tag, since a single folder can hold many copies of the same stray element. Extending that map is the extension point exchangelib's maintainer recommends for this class of problem.

The degraded `Item` then reaches the existing allowlists from #4158, which skip it with a warning while healthy siblings keep syncing. All three exchangelib call sites (`GetItem`, `FindItem`, `SyncFolderItems`) resolve tags through this one map, so mails, contacts, tasks and calendars are covered by the single change.

Genuine errors still fail loudly. Only the unmappable element is dropped, so this cannot turn a systematic failure into a doc-less "successful" sync that deletes previously indexed documents — the principle established in #4085 / #4123 / #4147.

### Testing

* `pytest tests/sources/test_outlook.py` — 103 tests, including the four added here: known tags still resolve to their own models, an unknown tag degrades to `Item` and warns once, the reporter's exact response shape parses into `[CalendarItem, Item]` instead of raising, and the degraded item is skipped by the calendar allowlist.
* Full service suite: 2302 passed, total coverage 92.19% (gate 90%). `client.py` and `datasource.py` at 97%; every line added here is exercised.
* `ruff check` / `ruff format --check` clean on `connectors` and `tests`; `pyright tests` clean, `pyright connectors` unchanged at the 13 pre-existing errors present on `main`.
* Before/after check against the reporter's response shape through exchangelib's real parsing path: `ValueError` as reported without the change, `['CalendarItem', 'Item']` plus the new warning with it.
* No E2E coverage: Outlook has no ftest fixture (no emulated backend), and the failure only reproduces on the reporter's server. Verification is the parsing check above against the exact XML shape from the SDH.

#### Pre-Review Checklist

- [x] `BaseFolder.ITEM_MODEL_MAP` is mutated process-wide at import time. Outlook is the only consumer of exchangelib, and this mirrors the existing process-global `BaseProtocol.HTTP_ADAPTER_CLS` assignment from #4094. A test asserts that every model exchangelib already knows still resolves to itself, so the map only ever adds a fallback.

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4158 — per-folder item type allowlists this change feeds into
* https://github.com/elastic/connectors/pull/4147 — Contacts item type dispatch
* https://github.com/elastic/connectors/pull/4123 — nullable field hardening
* https://github.com/elastic/connectors/pull/4085 — per-account skips vs. connection-wide failures
* https://github.com/elastic/connectors/pull/4065 — localized Exchange folder names

The 8.19 backport will need to be done by hand: that branch still has the single-file `connectors/sources/outlook.py` layout, so the cherry-pick conflicts (as it did for #4158).

## Release Note

Fixed an issue where the Outlook connector aborted an entire sync when an Exchange server returned an element it did not recognise as an item, such as a stray `EndTimeZone` alongside a calendar item. Such elements are now skipped with a warning and the rest of the mailbox continues to sync.

### DIFF
--- a/app/connectors_service/connectors/sources/outlook/client.py
+++ b/app/connectors_service/connectors/sources/outlook/client.py
@@ -22,7 +22,8 @@ from exchangelib import (
     OAuth2Credentials,
 )
 from exchangelib.errors import ErrorFolderNotFound, ErrorManagedFolderNotFound
-from exchangelib.folders import Calendar, Messages
+from exchangelib.folders import BaseFolder, Calendar, Messages
+from exchangelib.items import Item
 from exchangelib.protocol import BaseProtocol, NoVerifyHTTPAdapter
 from ldap3 import SAFE_SYNC, Connection, Server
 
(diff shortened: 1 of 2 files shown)


app/connectors_service/connectors/sources/outlook/client.py
```diff
@@ -53,6 +54,37 @@
 FOLDER_SKIP_ERRORS = (ErrorFolderNotFound, ErrorManagedFolderNotFound)
 
 
+class UnknownItemTagTolerantMap(dict):
+    """Item model map that degrades unrecognised EWS tags to the generic `Item`.
```
erikcurrin-elastic: Nit: This seems like it's verbose.

erikcurrin-elastic (approved):
Approved with a nit

app/connectors_service/connectors/sources/outlook/client.py
Jan-Kazlouski-elastic: Agreed.

PR #4291, opened 2026-07-28
fix(outlook): prefix content document identities so DLS can match them

### LINKED ISSUE #4290: [Outlook] Document Level Security filters out every document
## Summary

With DLS enabled, the Outlook connector indexes access-control documents and content documents in two different dialects, so the DLS terms query can never match. The owner of a mailbox retrieves **none** of their own documents.

Access-control documents grant **prefixed** identities, built with `prefix_identity`:

```
["user_id:60e8a9c1-...", "name:Alex Wilber", "email:alex.wilber@gmail.com"]
```

Every content document is decorated with the **raw** address instead:

```
["alex.wilber@gmail.com"]
```

The intersection is empty, and DLS filters `_allow_access_control` with a `terms` query, so nothing matches.

## Verification

Confirmed end to end against a real Elasticsearch 8.19.0, using documents produced by the connector itself (no hand-written queries): the access-control document and the content documents were generated by driving `get_access_control()` and `get_docs()`, the index was created with the framework's own mappings, and the mustache template stored on the access-control document was rendered by Elasticsearch and executed.

| Document | main / 9.6 | 8.19 |
| --- | --- | --- |
| as the connector writes it today | **HIDDEN** | **HIDDEN** |
| same document, `email:` prefix applied | VISIBLE | VISIBLE |
| document with no `_allow_access_control` | VISIBLE | VISIBLE |

The second row is the control: the identical document becomes visible as soon as the prefix is added, which rules out the mappings, the template and the query and isolates the prefix as the sole cause.

Note that the two versions differ in ways that have to be paired correctly when reproducing this — `main` queries `_allow_access_control.keyword` and relies on dynamic mapping, while 8.19 queries `_allow_access_control.enum` and applies `Mappings.default_text_fields_mappings`. Each is internally consistent, and both fail for the same reason.

## Scope

Pre-existing since DLS was added to the connector, and present unchanged in 8.19, so no recent change introduced it. All five `_decorate_with_access_control` call sites are affected (mails, attachments, contacts, tasks, calendars).

## Why CI did not catch it

The Outlook suite asserts the access-control side (`assert len(acl) == 1`) but never asserts `_allow_access_control` on content documents, so the two sides are never compared. Connectors such as Jira, Confluence and OneDrive assert both sides with explicit expected strings, which is what would have caught this.

## Fix

Prefix the identities on content documents so they match what the access-control document grants. Casing needs handling at the same time: a `terms` query is case-sensitive, and a mailbox's primary SMTP address is not guaranteed to match the directory's casing for the same person.

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/connectors/issues/4290

With DLS enabled, the Outlook connector writes its two sides in different dialects. Access control documents grant **prefixed** identities, while every content document is decorated with the **raw** SMTP address:

```
ACL doc grants  : ["user_id:60e8a9c1-...", "name:Alex Wilber", "email:alex.wilber@gmail.com"]
content doc has : ["alex.wilber@gmail.com"]
overlap         : NONE
```

DLS filters `_allow_access_control` with a `terms` query, so the intersection being empty means the owner of a mailbox retrieves **none** of their own documents. This is pre-existing since DLS was added and affects all five `_decorate_with_access_control` call sites (mails, attachments, contacts, tasks, calendars).

This routes the mailbox address through `_prefix_email` at every call site. Casing is normalised inside that helper rather than at the call sites, because both sides of the comparison already share it, so the invariant holds by construction instead of by convention.

Two smaller things came along with it, both load-bearing rather than drive-by:

- **Casing.** A `terms` query is case-sensitive, and a mailbox's primary SMTP address is not guaranteed to match the casing the directory reports for the same person. Prefixing alone would leave that as a second way to silently match nothing.
- **`None` identities.** A mailbox with no address prefixes to `None`, and `_decorate_with_access_control` would previously store that null in the terms list. `es_access_control_query` already filters `None` on the access control side; this makes the content side agree.

### What this does not fix

Access control documents take the address from the directory (Graph's `mail` on cloud, the LDAP `mail` attribute on server), while content documents take it from the mailbox (`primary_smtp_address`). Those are normally the same address, and after this change they also agree on casing, but they are not guaranteed to be equal: a tenant using aliases or proxy addresses can have a directory `mail` that differs from the mailbox's primary SMTP address, and for those users the terms query still will not match. Covering that properly means granting every proxy address a mailbox answers to, which is a larger change and deserves its own issue. Recording it here so it is a known limitation rather than a surprise.

## Upgrade note

The incorrect values are stored on the content documents themselves, not only in the query template, so deploying this is not sufficient on its own. Affected deployments need a **full content sync** to rewrite `_allow_access_control` on every document; an access control sync alone will not do it. This differs from #4005, where only the stored query was wrong and re-running the access control sync was enough.

Worth knowing alongside that: documents indexed before DLS was switched on carry no `_allow_access_control` field, and the `must_not exists` clause in `DLS_QUERY` makes those visible to everyone. That is existing framework behaviour rather than anything this PR changes, but it does mean enabling DLS never retroactively protects already-indexed content.

## Verification

Beyond unit tests, I confirmed the bug and the fix end to end against a real Elasticsearch 8.19.0. Both documents were produced by driving the connector itself rather than written by hand, the index was created with the framework's own mappings, and the mustache template stored on the access control document was rendered by Elasticsearch and then executed.

Before:

| Document | main / 9.6 | 8.19 |
| --- | --- | --- |
| as the connector writes it today | **HIDDEN** | **HIDDEN** |
| same document, `email:` prefix applied | VISIBLE | VISIBLE |
| document with no `_allow_access_control` | VISIBLE | VISIBLE |

After, on this branch, the first row is `VISIBLE` in both configurations.

One thing worth knowing if you reproduce this: the two versions have to be paired correctly or you will get a misleading result. `main` queries `_allow_access_control.keyword` and applies no mappings, so dynamic mapping supplies `.keyword`; 8.19 queries `_allow_access_control.enum` and applies `Mappings.default_text_fields_mappings`, whose dynamic template supplies `.enum`. Each is internally consistent, and both fail for the same reason, which is what makes the prefix the sole cause.

#### Pre-Review Checklist

**Tests.** Six new tests. The three `test_content_documents_are_reachable_*` ones assert the invariant directly — that the identities on content documents intersect the identities the access control document grants — for the cloud path, the server/LDAP path, and mismatched address casing. This is the dual-side assertion the Outlook suite was missing: it asserted the access control side with `assert len(acl) == 1` and never looked at `_allow_access_control` on content documents, which is why this survived. Jira, Confluence and OneDrive already assert both sides.

I checked the new tests are load-bearing by reverting only the source change and re-running them:

```
FAILED test_content_documents_are_reachable_by_their_owner
FAILED test_content_documents_are_reachable_by_their_owner_on_server
FAILED test_content_documents_are_reachable_whatever_the_address_casing
FAILED test_decorate_with_access_control_drops_unknown_identities
FAILED test_prefix_email_normalises_casing[Alex.Wilber@Example.COM-...]
```

Full suite on this branch: `2301 passed`, coverage 92.19%.

**Local run.** `make autoformat lint test PYTHON=python3.11`. Ruff is clean. `typecheck` reports 13 pyright `unknown import symbol` errors in `kibana.py`, `service_cli.py`, `agent/`, `cli/`, `es/`, `protocol/` and `services/` — all pre-existing on `main` and none in files this PR touches.

This changes which documents a user can retrieve under Document Level Security, so it deserves a careful read. The change only ever *narrows* nothing and *widens* access from "nobody sees anything" to "the mailbox owner sees their own mailbox" — content documents still carry exactly one identity, the mailbox they came from, which is the same value as before with a prefix applied. No document becomes visible to anyone other than its own mailbox owner.

Worth flagging explicitly: documents with no `_allow_access_control` field are visible to everyone by design (the `must_not exists` clause in `DLS_QUERY`). That is unchanged here, and Outlook always sets the field when DLS is on.

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4287 — the other bugfix from the same investigation

## Release Note

Fixed Document Level Security for the Outlook connector, where content documents were indexed with identities that did not match the ones granted by the access control documents, so no synced document was retrievable by the user who owned it.

### DIFF
--- a/app/connectors_service/connectors/sources/outlook/datasource.py
+++ b/app/connectors_service/connectors/sources/outlook/datasource.py
@@ -51,6 +51,16 @@ CALENDAR_ITEM_TYPES = (CalendarItem,)
 TASK_ITEM_TYPES = (Task,)
 
 
+def _mailbox_access_control(account):
+    """Identities allowed to read everything in a mailbox.
+
+    These have to stay in the same dialect as the identities granted by
+    `_user_access_control_doc`: DLS matches the two with a terms query, so an
+    address stored here unprefixed is invisible to every user.
+    """
+    return [_prefix_email(account.primary_smtp_address)]
+
+
 class OutlookDocFormatter:
     """Format Outlook object documents to Elasticsearch document"""
 
@@ -465,8 +475,9 @@ class OutlookDataSource(BaseDataSource):
 
     def _decorate_with_access_control(self, document, access_control):
         if self._dls_enabled():
+            identities = document.get(ACCESS_CONTROL, []) + access_control
             document[ACCESS_CONTROL] = list(
-                set(document.get(ACCESS_CONTROL, []) + access_control)
+                {identity for identity in identities if identity is not None}
             )
         return document
 
@@ -553,7 +564,7 @@ class OutlookDataSource(BaseDataSource):
             )
             yield (
                 self._decorate_with_access_control(
-                    document, [account.primary_smtp_address]
+                    document, _mailbox_access_control(account)
                 ),
                 partial(
                     self.get_content, attachment=copy(attachment), timezone=timezone
@@ -577,7 +588,7 @@ class OutlookDataSource(BaseDataSource):
             )
             yield (
                 self._decorate_with_access_control(
-                    document, [account.primary_smtp_address]
+                    document, _mailbox_access_control(account)
                 ),
                 None,
             )
@@ -616,7 +627,7 @@ class OutlookDataSource(BaseDataSource):
                 continue
             yield (
                 self._decorate_with_access_control(
-                    document, [account.primary_smtp_address]
+                    document, _mailbox_access_control(account)
                 ),
                 None,
             )
@@ -637,7 +648,7 @@ class OutlookDataSource(BaseDataSource):
             )
             yield (
                 self._decorate_with_access_control(
-                    document, [account.primary_smtp_address]
+                    document, _mailbox_access_control(account)
                 ),
                 None,
             )
@@ -694,7 +705,7 @@ class OutlookDataSource(BaseDataSource):
         )
         yield (
             self._decorate_with_access_control(
-                document, [account.primary_smtp_address]
+                document, _mailbox_access_control(account)
             ),
             None,
         )
--- a/app/connectors_service/connectors/sources/outlook/utils.py
+++ b/app/connectors_service/connectors/sources/outlook/utils.py
@@ -33,7 +33,11 @@ def ews_format_to_datetime(source_datetime, timezone):
 
 
 def _prefix_email(email):
-    return prefix_identity("email", email)
+    # Access control documents and content documents both route addresses
+    # through here. DLS compares them with a case-sensitive terms query, and a
+    # mailbox's primary SMTP address does not always match the directory's
+    # casing, so normalise in the one place both sides share.
+    return prefix_identity("email", email.lower() if email else email)
 
 
 def _prefix_display_name(user):
--- a/app/connectors_service/tests/sources/test_outlook.py
+++ b/app/connectors_service/tests/sources/test_outlook.py
@@ -37,6 +37,7 @@ from exchangelib.items import (
 )
 from exchangelib.protocol import BaseProtocol, NoVerifyHTTPAdapter
 
+from connectors.access_control import ACCESS_CONTROL
 from connectors.sources.outlook import OutlookDataSource
 from connectors.sources.outlook.client import (
     ExchangeUsers,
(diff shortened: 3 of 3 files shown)


erikcurrin-elastic (approved):

Jan-Kazlouski-elastic:
### E2E confirmation

Verified end-to-end against a real Outlook Cloud mailbox + local ES/Kibana (trial license) on this branch:

1. Full content sync → docs indexed with `_allow_access_control: ["email:<mailbox>"]` (prefixed)
2. Access control sync → ACL doc grants include the same `email:<mailbox>` identity
3. DLS `terms` filter as the mailbox owner → hits (owner’s mail visible)
4. Same filter as an unrelated identity → 0 hits
5. Filter using the pre-fix raw address only → 0 hits

So the content/ACL identity dialects match and Document Level Security works for the mailbox owner after a full content sync + access control sync.

PR #4297, opened 2026-07-29
fix(confluence): index inherited page restrictions for DLS (#4095)

### LINKED ISSUE #4095: Confluence Data Center Connector does not index inherited page restrictions
## Bug Description

When a Confluence page has no explicit page-level restrictions set ("No restrictions"), but inherits view restrictions from a parent page, the connector indexes the document with an incorrect or overly broad _allow_access_control value (eq. space permissions). The inherited ACL from the parent page hierarchy is not resolved and applied to the child page document.

This is a Document Level Security gap: users who should not have access to a restricted child page may be able to retrieve it via Elasticsearch queries, because the indexed ACL does not reflect the effective (inherited) permission.

### To Reproduce

1. In Confluence, create a parent page and restrict it to a specific set of groups/users ("Viewing and editing restricted").
  Example — Parent page: "Parent Page" (ID: <page-id-A>)
  Restrictions: group-admin, group-power-user, group-standard-user, user:[user@example.com].  (mailto:user@example.com)
  Everyone else: Has no access

2. Create a child page under that parent and leave it at the default "No restrictions" page-level setting.
Confluence UI confirms "There are inherited view restrictions" from the parent page.

3. Trigger a full sync of the Confluence Connector.

4. Query the indexed document for the child page in Elasticsearch and inspect the _allow_access_control field.

The _allow_access_control field of the child page document reflects space-level or license-level group membership instead of the effective inherited restrictions. For example, it includes a broader group such as "group:group-all-licensed-users" which should NOT have access to this page hierarchy.
Observation from Elasticsearch:
- Parent page (ID: <page-id-A>) — _allow_access_control correctly reflects the 4 restricted principals.
- Child page with "No restrictions" (page-level) — _allow_access_control does NOT reflect the inherited restrictions; instead it contains a broader permission set including groups that are explicitly excluded from the parent page.

## Expected behavior

The _allow_access_control field of the child page document should contain the resolved effective permissions — i.e., the inherited restrictions from the parent page:
```
["group:group-admin", "group:group-power-user", "group:group-standard-user", "user:[user@example.com](mailto:user@example.com)"]
```

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/connectors/issues/4095

When a Confluence page has no page-level restrictions of its own but inherits view restrictions from ancestors — or has its own restrictions that must also satisfy parent restrictions — the connector previously either ignored inherited ACL or used only the nearest restricted ancestor / the child's list alone, then fell back to broad space-level permissions. That over-granted access in `_allow_access_control` (a Document Level Security gap).

Confluence's documented view model requires users to satisfy **both** parent and child view restrictions ([CONF57](https://confluence.atlassian.com/spaces/CONF57/pages/701434765/Page+Restrictions)). This change resolves the **effective** read ACL as the intersection of:

1. The child's explicit read restrictions (when non-empty), and
2. Every ancestor's explicit read restrictions (full chain, not nearest-only).

Behavior details:

- Fetch each ancestor via `rest/api/content/{id}/restriction/byOperation/read` (Confluence does not expose inherited restrictions inline; expanding `ancestors.restrictions...` on `content/search` 404s the whole batch on unreadable drafts — CONFCLOUD-77618).
- **200 + empty** → no restriction layer on that content; keep walking.
- **404** → skip that ancestor; keep walking.
- **403 / 401 / 5xx / other** → fail closed: index `_allow_access_control = []` and **do not** fall back to space ACL (sync continues).
- **No restriction layers** after a successful walk → fall back to space VIEWSPACE permissions (unchanged Cloud / Server / DC paths).
- Always runs this resolver even when the child has its own restrictions (child list alone is not the effective ACL).
- Attachments continue to reuse the page's resolved ACL.

Applies to Confluence Cloud, Server, and Data Center. Advanced Sync Rules / `search_by_query` ACL is unchanged (out of scope).

#### Pre-Review Checklist

## Release Note

Confluence connector Document Level Security now indexes effective page view restrictions by intersecting the child's and all ancestors' explicit read restrictions, instead of falling back to broader space-level permissions when a child page has no restrictions of its own (or only its own list). Ambiguous ancestor restriction fetches fail closed without over-granting via space ACL.

### DIFF
--- a/app/connectors_service/connectors/sources/atlassian/confluence/datasource.py
+++ b/app/connectors_service/connectors/sources/atlassian/confluence/datasource.py
@@ -464,6 +464,44 @@ class ConfluenceDataSource(BaseDataSource):
 
         return identities
 
+    async def _resolve_inherited_restrictions(self, ancestors, extract_identities):
+        """Resolve the effective view restrictions inherited from ancestor pages.
+
+        A Confluence page with no restrictions of its own ("No restrictions")
+        still inherits the view restrictions of its ancestors. The REST API does
+        not expose these inherited restrictions directly, so we walk the ancestor
+        chain from the closest parent up to the root and return the identities of
+        the nearest ancestor that has explicit read restrictions.
+
+        Args:
+            ancestors (list): Ancestors of the document, ordered from the root to
+                the immediate parent (as returned by the Confluence API).
+            extract_identities (callable): Function that extracts the set of
+                access-control identities from a restrictions payload.
+
+        Returns:
+            set: Identities inherited from the nearest restricted ancestor, or an
+                empty set when no ancestor has explicit read restrictions.
+        """
+        if not self._dls_enabled():
+            return set()
+
+        for ancestor in reversed(ancestors or []):
+            ancestor_id = ancestor.get("id")
+            if not ancestor_id:
+                continue
+
+            restrictions = await self.confluence_client.fetch_content_restrictions(
+                content_id=ancestor_id
+            )
+            identities = extract_identities(
+                response=restrictions.get("restrictions", {})
+            )
+            if identities:
+                return identities
+
+        return set()
+
     async def close(self):
         """Closes unclosed client session"""
         await self.confluence_client.close_session()
@@ -547,6 +585,7 @@ class ConfluenceDataSource(BaseDataSource):
             Integer: Number of attachments in a page/blogpost
             List: List of permissions attached to document
             Dictionary: Dictionary of restrictions attached to document
+            List: List of ancestors (with their ids) of the document
         """
         async for (
             document,
@@ -592,6 +631,7 @@ class ConfluenceDataSource(BaseDataSource):
                 document.get("restrictions", {})
                 .get("read", {})
                 .get("restrictions", {}),
+                document.get("ancestors", []),
             )
 
     async def fetch_attachments(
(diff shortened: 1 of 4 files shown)


app/connectors_service/connectors/sources/atlassian/confluence/datasource.py
```diff
@@ -464,6 +479,52 @@ def _extract_identities_for_datacenter(self, response):
...
+        if child_identities:
+            layers.append(child_identities)
+
+        for ancestor in ancestors or []:
+            ancestor_id = ancestor.get("id")
+            if not ancestor_id:
+                continue
+
+            try:
+                restrictions = await self.confluence_client.fetch_content_restrictions(
```
erikcurrin-elastic, line 506: Does this introduce another permission the auth must have?  If it does, we should document it.

Jan-Kazlouski-elastic: According to their docs: https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-content-restrictions/
it doesn't require anything on top of what we already have. We're allowed to get the restrictions for the child as part of the content during the search https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-content/#api-wiki-rest-api-content-search-get, it's the same as getting restrictions for the parents https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-content-restrictions/#api-wiki-rest-api-content-id-restriction-byoperation-get, but we focus on restrictions alone. Which is the same level of permission.
The wording is a bit confusing in their API reference, but I think it is safe to say that we don't have to get another permission for that. No admin/write ones for sure.

erikcurrin-elastic (approved):

PR #4341, opened 2026-08-12
chore: migrate DRA publish to elastic/dra-prep plugin

navyau09:
## Summary

Replaces the legacy Release Manager Docker step with the [`elastic/dra-prep-buildkite-plugin`](https://github.com/elastic/dra-prep-buildkite-plugin) (v0.1.5) for DRA artifact publishing.

**Changes:**
- Deleted `init_dra_publishing.sh` and `publish-daily-release-artifact.sh` (Release Manager scripts)
- Added `.buildkite/publish/dra/upload-dra-pipeline.sh` — generator script that resolves `STACK_VERSION` from the `VERSION` file and emits dra-prep + trigger steps dynamically
- Added `.buildkite/publish/dra/stage_artifacts.sh` — stages Docker tarballs (amd64/arm64), Python packages, and zip into `artifacts/` for the plugin
- Updated `pipeline.yml` to replace the old publish step with the generator approach

**How it works:**
1. The "Publish DRA Items" step runs `upload-dra-pipeline.sh | buildkite-agent pipeline upload`
2. The generator resolves `STACK_VERSION` at runtime and emits the DRA Prep + Trigger steps
3. The DRA Prep step runs `stage_artifacts.sh` to stage artifacts, then the `dra-prep` plugin classifies them, generates the manifest, and uploads to GCS
4. The Trigger step fires `unified-release-dra-processing` with the manifest coordinates

**Notes:**
- Snapshot workflow runs on all active branches; staging only on version branches (not `main`)
- Dependency CSV generation was removed — the GCP agent doesn't have `python3-venv` and this artifact was previously generated inside the Release Manager container; can be added back if needed

## Test plan

- [x] Triggered test build on `gh-dra-prep-migration` branch — DRA Prep step passed, manifest uploaded to GCS, `unified-release-dra-processing` triggered successfully, Docker images pushed and signed
https://buildkite.com/elastic/connectors/builds/24371

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Jan-Kazlouski-elastic (commented):
Hello @navyau09 

### 1. Snapshot `.tar.gz` files are missing the `-SNAPSHOT` suffix

In `stage_artifacts.sh`, the rename logic only matches `-{VERSION}-` (e.g. `-9.6.0-py3-...`), so it works for `.whl` files but not for sdists like `elasticsearch_connectors-9.6.0.tar.gz`.

The old `init_dra_publishing.sh` handled these with explicit copies. Worth fixing before merge — snapshot manifests may end up with inconsistently named Python packages.

### 2. Workflow gating is broader than before

Previously, snapshot/staging publishing was limited to specific branches (`main`, `8.x`, version branches). With `ci:packaging` on a feature-branch PR, the packaging group would run but nothing was published to DRA.

The new `upload-dra-pipeline.sh` always emits snapshot steps and emits staging on any non-`main` branch. That means a feature-branch PR with `ci:packaging` could now publish to DRA, which wasn't the case before. Suggest restoring the old branch regex checks.

### 3. Silent failures can produce incomplete manifests

`|| true` on artifact downloads and `mv` calls means missing Docker/Python artifacts won't fail the step. The final check only verifies that `artifacts/` is non-empty — and `make clean zip` always produces at least the zip.

This already bit us during development (wrong `DOCKER_ARTIFACT_KEY` → no docker images in the manifest, step still passed). Suggest failing loudly when expected artifacts are missing.

### 4. Question: `dependencies.csv`

The PR removes dependency CSV generation (noted as due to missing `python3-venv` on the GCP agent). Is that acceptable for release/compliance, or should we restore it (e.g. on an agent image that supports it, or via dractl)?

### 5. Question: hardcoded constants

`stage_artifacts.sh` hardcodes `PROJECT_NAME="connectors"` instead of sourcing `publish-common.sh` (which defaults to `elastic-connectors`). Docker tarball names changed accordingly (`connectors-...` vs `elastic-connectors-...`). The test build passed, so this may be fine — but is that intentional, and should we source the shared config to avoid drift?

navyau09:
Hey @Jan-Kazlouski-elastic Thanks for the feedback.

**1. Snapshot .tar.gz files are missing the -SNAPSHOT suffix**
Fixed. The rename now handles both cases: -{VERSION}- for `.whl` files and `-{VERSION}` for dists like `elasticsearch_connectors-9.6.0.tar.gz.`

**2. Workflow gating is broader than before**
Fixed. upload-dra-pipeline.sh now matches the old branch regex: snapshot only on main and version branches ([0-9]+\.[0-9x]+), staging only on version branches. Feature branch PRs with `ci:packaging` will run the packaging group but won't publish to DRA.

**3. Silent failures can produce incomplete manifests**
Fixed. Removed all || true from artifact downloads and mv calls. Missing Docker images or Python packages will now fail the step loudly.

**4. Question: dependencies.csv**
In all other migrated repos (`ems-server`, `cloud-defend`, `endpoint-dev`), the CSV is generated in the build step and uploaded as a Buildkite artifact, then downloaded in `stage_artifacts.sh`. I have created a [draft PR](https://github.com/elastic/connectors/pull/4344) following the same pattern, that adds the building python packages and csv dependencies as part of `.buildkite/pipeline.yml`. Please take a look

**5. Question: hardcoded constants**
PROJECT_NAME=`connectors` is intentional. It is the DRA product ID registered in unified-release, not the Docker image name. `DOCKER_ARTIFACT_KEY="elastic-connectors-docker"` mirrors the default in `publish-common.sh` (elastic-connectors + -docker). We can't source `publish-common.sh` directly due to side effects (requires CURDIR, calls git-setup.sh, creates build.yaml). Since we removed || true from the `mv`, a mismatch would now fail the build.

Jan-Kazlouski-elastic (approved):
Thank you for the fixes, @navyau09 
Approved!

PR #4344, opened 2026-08-13
buildkite: build Python packages and generate dependency CSV in dedicated step

navyau09:
## Summary

- Adds a `build_python_packages` step inside the existing "Packaging and DRA" group that builds whl/sdist packages and generates `dependencies.csv` using `pip-licenses` (via `make install-package` + `make deps-csv` in `app/connectors_service/`)
- `stage_artifacts.sh` now downloads packages and CSV from this step via `buildkite-agent artifact download --step build_python_packages`, and copies the CSV into `artifacts/` with the DRA-expected filename (`connectors-{VERSION}{SUFFIX}-dependencies.csv`)
- Fixes branch gating in `upload-dra-pipeline.sh`: snapshot now only runs on `main` and version branches (not feature branches), staging uses a strict version branch regex

## Why

The dependency CSV could not be generated in `stage_artifacts.sh` because that step runs on a bare GCP agent (no Python venv). Moving it to a dedicated step that uses the existing `enterprise-search-ubuntu-2204-connectors-py` image (which has `pyenv` and all required tooling) follows the same pattern used by ems-server, cloud-defend, and endpoint-dev.

## Test plan

- [ ] Trigger a build with `ci:packaging` label and confirm `build_python_packages` step passes and uploads artifacts
- [ ] Confirm `stage_artifacts.sh` downloads packages and CSV correctly and they appear in `artifacts/`
- [ ] Confirm DRA prep step succeeds and manifest includes the CSV

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Jan-Kazlouski-elastic (changes requested):
Thanks for the follow-up — `build_python_packages` is the right approach. One blocking issue in `stage_artifacts.sh` prevents Python packages and the CSV from reaching the DRA manifest.

### 1. `make clean zip` wipes downloaded artifacts (blocking)

Python packages and `dependencies.csv` are downloaded into `dist/`, then `make clean zip` runs. Root `zip: clean` deletes `dist/` in both subprojects, so the staging loop finds nothing (`[[ -f "$f" ]] || continue` skips silently). Only docker + zip end up in `artifacts/`.

Seen in the #4341 validation build — staged list was only:
```
connectors-9.6.0-SNAPSHOT-docker-image-linux-amd64.tar.gz
connectors-9.6.0-SNAPSHOT-docker-image-linux-arm64.tar.gz
connectors-9.6.0-SNAPSHOT.zip
```

**Fix:** run `make clean zip` **before** Python artifact downloads.

### 2. Silent CSV copy failure

```bash
cp ... dependencies.csv ... 2>/dev/null || true
```

Remove `2>/dev/null || true` — fail loudly if CSV is missing.

### 3. Test plan

After the fix, confirm **"Staged artifacts:"** includes: 2× docker, 1× zip, 2× whl, 2× sdist, 1× `connectors-{VERSION}-SNAPSHOT-dependencies.csv`.

Happy to re-review once updated.

Jan-Kazlouski-elastic (approved):
Thank you for the fixes @navyau09 
The PR looks good to be merged.

Jan-Kazlouski-elastic:
Hello @navyau09 
With #4344 merged, staging works and all 8 artifacts appear in **`artifacts/`**, but **`dractl`** prep fails classifying **`.whl`** files. Looks like the connectors product config in **`dractl`** doesn't support wheels yet — can we add that, or confirm which artifacts connectors should publish via **`dractl`**?

Failing buildkite job:
https://buildkite.com/elastic/connectors/builds/24514

navyau09:
> Hello @navyau09 With #4344 merged, staging works and all 8 artifacts appear in **`artifacts/`**, but **`dractl`** prep fails classifying **`.whl`** files. Looks like the connectors product config in **`dractl`** doesn't support wheels yet — can we add that, or confirm which artifacts connectors should publish via **`dractl`**?
> 
> Failing buildkite job: https://buildkite.com/elastic/connectors/builds/24514

Hey @Jan-Kazlouski-elastic Thanks for flagging this. Before adding `.whl` support to `dractl`, can you confirm whether connectors actually needs to publish Python wheels via DRA? My understanding is that `.whl` files do not publish via Elastic artifact registry. If yes, I create a follow up PR to remove the `whl/sdist` staging from `stage_artifacts.sh` and only publish the Docker tarballs, the zip, and the dependencies CSV via DRA.

Jan-Kazlouski-elastic:
Good thinking, @navyau09 
That would make more sense, I just assumed you wanted them in DRA. Let's drop `.whl` and Python sdist `.tar.gz` from `stage_artifacts.sh` and keep only Docker tarballs, the zip, and the dependencies CSV. Current DRA report matches that set, so we're not losing anything that's actually published today. Go ahead with the follow-up PR, I'll review it and we'll unblock the `main` pipeline without waiting on `dractl` changes.

navyau09:
Hey @Jan-Kazlouski-elastic, I noticed we have only added the backport label for `9.6.0`. Should we also add the labels for the other release versions?

Jan-Kazlouski-elastic:
Hey @navyau09 
After the full set of:

- https://github.com/elastic/connectors/pull/4341
- https://github.com/elastic/connectors/pull/4344
- https://github.com/elastic/connectors/pull/4371
- https://github.com/elastic/connectors/pull/4373

is confirmed to be working - I'll handle backports to 9.5, 9.4, 8.19 versions manually as single clean commits and link them accordingly. No further action would be required from your side. I will reach out if more trouble arise.

Issue #4311, opened 2026-07-30
Long syncs can be killed as idle after ES ConnectionTimeout / refresh storm


Jan-Kazlouski-elastic:
## Summary

After a long-running content sync (reported: ~7h, ~583k Outlook documents), the connector service hit `ConnectionTimeout` on `indices.refresh` while polling `.elastic-connectors*`, then `job_cleanup_service` marked the in-progress job as idle/ERROR, and the sync runner aborted with `ConnectorJobNotRunningError`.

Reported on https://github.com/elastic/sdh-search/issues/1898#issuecomment-5132828764

## Failure chain

1. During an active sync, `check_job()` reloads the connector and job **every 1s**, and each reload goes through `fetch_response_by_id` which calls **`indices.refresh`** before `get`.
2. Scheduling / job-execution services also call `get_all_docs` → `indices.refresh` every ~30s.
3. Under bulk ingest load, those refreshes time out (`ConnectionTimeout`).
4. The background `update_ingestion_stats` task has **no error handling** — a single failed stats/heartbeat update kills the task, so `last_seen` stops advancing.
5. After `IDLE_JOBS_THRESHOLD` (5 minutes), `job_cleanup_service` marks the job ERROR (`The job has not seen any update for some time.`).
6. `check_job()` sees ERROR and raises `ConnectorJobNotRunningError`.

This is **not** Outlook/EWS-specific. Any long sync under ES pressure can hit it.

## Likely fixes

- Stop refreshing before `get` (Get API is realtime) and stop/avoid refreshing on every `get_all_docs` poll — related to #3155.
- Harden `update_ingestion_stats` so a transient ES error does not permanently stop the heartbeat.
- (Follow-up) Make `IDLE_JOBS_THRESHOLD` configurable — draft #3375.

## Related

- #3155 (refresh on system indices)
- #2298 / #3375 (idle timeout)
- #2925 (ConnectorJobNotRunningError at scale)

Issue #3155, opened 2025-01-27
Connectors cannot call refresh on `.elastic-connectors` after System Index conversion


seanstory:
## Bug Description
When launching a fresh agentless connector on 9.0.0-SNAPSHOT, I'm hitting:

```
[elastic_agent.connectors][error] Error while checking existence of connector 'f641ab2c-8002-41d5-959d-943425d217a2': AuthorizationException(403, 'security_exception', 'action [indices:admin/refresh] is unauthorized for API key id [W-SfqJQBlxUzDiujVIuP] of user [elastic/fleet-server] on restricted indices [.elastic-connectors], this action is granted by the index privileges [maintenance,manage,all]')
```

Jedr suggests that we should be able to remove these refresh checks, since the Connectors APIs are doing these for us under the hood. 

### To Reproduce
1. Create a 9.0.0 deployment
2. create a Confluence Cloud connector from the integrations tab
3. navigate to the Agent UI
4. click on the Agent and view its connectors logs
5. see the above stack trace

## Expected behavior
Connectors should have sufficient privileges do do whatever they try to do, and should only do the things they have privileges for.

## Environment
9.0.0-SNAPSHOT

jedrazb:
Not in scope for Feb 2025 anymore, since we want to migrate to Connector APIs fully

PR #4345, opened 2026-08-13
fix: keep long syncs alive through ES refresh storm (#4311)

Jan-Kazlouski-elastic:
## Relates to https://github.com/elastic/connectors/issues/4311

Long-running syncs could be killed as idle after Elasticsearch `ConnectionTimeout` on `indices.refresh` while the connector service polled `.elastic-connectors*` system indices. During an active sync, `check_job()` reloaded connector and job documents every second; each reload called `indices.refresh` before `GET`. Background scheduling, execution, and cleanup services also refreshed before every `get_all_docs` search (~30s). Under bulk-ingest load this produced a refresh storm. When refresh timed out, the `update_ingestion_stats` heartbeat task had no error recovery — a single failed stats update stopped `last_seen` from advancing. After `IDLE_JOBS_THRESHOLD` (5 minutes), `job_cleanup_service` marked the job ERROR and `check_job()` raised `ConnectorJobNotRunningError` even though bulk ingest was still running.

Reported via https://github.com/elastic/sdh-search/issues/1898#issuecomment-5132828764 (Outlook customer, ~7h / ~583k documents). This is connectors-framework-wide, not Outlook-specific. Also addresses unnecessary refresh on system indices ([#3155](https://github.com/elastic/connectors/issues/3155)).

**Changes:**
- Remove `indices.refresh` before `GET` in `fetch_response_by_id` and before search in `get_all_docs` on system indices (GET is realtime; search NRT is sufficient for 30s pollers and 5-minute idle detection).
- Make `update_ingestion_stats` resilient to transient ES/transport errors (`ApiError`, `elastic_transport.ConnectionError`, `ProtocolError`): log a warning with `exc_info` and retry on the next interval instead of permanently stopping the heartbeat.
- Retry `check_job()` through the same retriable errors in the active-sync loop so a transient status reload failure does not abort the sync.
- Increase `JOB_CHECK_INTERVAL` from 1s to 5s to reduce polling load during long syncs (cancel detection remains within 5s).

**Tests:** updated `test_index.py` refresh expectations; added heartbeat retry tests (metadata failure, reload failure) and `test_sync_job_runner_survives_transient_check_job_errors`.

#### Pre-Review Checklist

## Related Pull Requests

* Customer report: https://github.com/elastic/sdh-search/issues/1898#issuecomment-5132828764
* Related: https://github.com/elastic/connectors/issues/3155

## Release Note

Long-running connector syncs are no longer falsely marked as idle when Elasticsearch is temporarily slow or refresh calls time out. The connector service stops forcing index refresh on every job status check, keeps the ingestion heartbeat alive through transient errors, and retries job status checks during active syncs.


app/connectors_service/connectors/es/index.py
```diff
@@ -174,11 +174,6 @@ async def fetch_by_id(self, doc_id):
 
     async def fetch_response_by_id(self, doc_id):
         try:
-            if not self.serverless:
-                await self._retrier.execute_with_retry(
-                    partial(self.client.indices.refresh, index=self.index_name)
-                )
-
```
Jan-Kazlouski-elastic, line 181: Removed `indices.refresh` before `GET`. The Get API is realtime and is not affected by the index refresh interval — a separate refresh on every reload was redundant and expensive (during sync this path ran every few seconds per connector + job).

ES docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-get.html
("By default, this API is realtime and is not affected by the refresh rate of the index.")

app/connectors_service/connectors/sync_job_runner.py
```diff
@@ -30,15 +34,22 @@
...
     INDEXED_DOCUMENT_VOLUME,
+    ProtocolError,
 )
 from connectors.utils import truncate_id
 
 UTF_8 = "utf-8"
 
 JOB_REPORTING_INTERVAL = 10
-JOB_CHECK_INTERVAL = 1
+JOB_CHECK_INTERVAL = 5
```
Jan-Kazlouski-elastic, line 44: Reduced frequency from 1s to 5s to cut system-index GET load during long syncs (2 GETs per tick: connector + job). Worst-case cancel detection delay rises from ~1s to ~5s, which is still well within the 5-minute idle threshold. This was part of the customer failure under ES pressure (~7h sync, ~583k docs).

app/connectors_service/connectors/sync_job_runner.py
```diff
@@ -30,15 +34,22 @@
...
 ES_ID_SIZE_LIMIT = 512
 
+_RETRIABLE_INGESTION_STATS_ERRORS = (
+    ApiError,
+    ElasticTransportConnectionError,
+    ProtocolError,
+)
+
```
Jan-Kazlouski-elastic, line 52: Shared tuple for heartbeat + check_job retry paths. We only swallow known transient ES/transport failures:
- `ApiError` — ES HTTP errors (timeouts, 5xx)
- `elastic_transport.ConnectionError` — connection-level failures (customer saw `ConnectionTimeout` here)
- `ProtocolError` — connectors wrapper around failed `update_metadata` / API calls

Programming errors (`KeyError`, etc.) still propagate. `asyncio.CancelledError` is not caught (subclass of `BaseException`).

app/connectors_service/connectors/sync_job_runner.py
```diff
@@ -189,7 +200,20 @@ async def execute(self):
             )
 
             while not self.sync_orchestrator.done():
-                await self.check_job()
+                try:
```
Jan-Kazlouski-elastic, line 203: Previously one failed heartbeat tick permanently killed the background task, so `last_seen` stopped updating and `job_cleanup_service` marked the job idle after 5 minutes — even while bulk ingest was still running.

The loop retries every 10s with no outer attempt cap; the safety net remains `IDLE_JOBS_THRESHOLD` (5 min without a successful `last_seen` update). `exc_info=True` for RCA in support cases.

app/connectors_service/connectors/sync_job_runner.py
```diff
@@ -481,16 +505,22 @@ async def update_ingestion_stats(self, interval):
...
-
-            result = self.sync_orchestrator.ingestion_stats()
-            ingestion_stats = {
-                INDEXED_DOCUMENT_COUNT: result.get(INDEXED_DOCUMENT_COUNT, 0),
-                INDEXED_DOCUMENT_VOLUME: result.get(INDEXED_DOCUMENT_VOLUME, 0),
-                DELETED_DOCUMENT_COUNT: result.get(DELETED_DOCUMENT_COUNT, 0),
-            }
-            await self.sync_job.update_metadata(ingestion_stats=ingestion_stats)
+            try:
+                if not await self.reload_sync_job():
```
Jan-Kazlouski-elastic, line 509: Same class of failure as the heartbeat

Jan-Kazlouski-elastic (commented):
A few explanations for the changes, since they are affecting every connector

app/connectors_service/connectors/sync_job_runner.py
```diff
@@ -189,7 +200,20 @@ async def execute(self):
...
-                await self.check_job()
+                try:
+                    await self.check_job()
+                except (
+                    ConnectorJobCanceledError,
+                    ConnectorJobNotRunningError,
+                    ConnectorNotFoundError,
+                    ConnectorJobNotFoundError,
+                ):
+                    raise
```
erikcurrin-elastic, line 211: Do we have a limit to how many times we will retry?

erikcurrin-elastic (approved):

app/connectors_service/connectors/sync_job_runner.py
Jan-Kazlouski-elastic: There is **no per-attempt limit in the loop** and it is **intentional**:
Retry is every 5s within `check_job` and every 10s in `heartbeat` until sync ends or ES recovers. The bug was that a single transient failure permanently stopped updates. The backstop is `IDLE_JOBS_THRESHOLD` (5 min), meaning if `last_seen` still isn’t advancing, `job_cleanup_service` marks the job ERROR.
There is no reason for us to give up on retries and wait for up to 5 minutes to mark the job as an ERROR since there is no possibility of `last_seen` being updated after we stopped trying.

PR #4392, opened 2026-08-27
fix(servicenow): add compact role-based DLS to reduce ACL memory

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/chat-program/issues/49

Part of https://github.com/elastic/chat-program/issues/47.
ServiceNow DLS expands every table read role into individual `user_id:` / `user:` / `email:` entries on each content document (and attachment). For large tenants — especially when a table ACL includes the `public` role — that can mean hundreds of thousands of ACL entries per document and OOM Elasticsearch during `_reindex`.

This PR adds compact DLS, **opt-in** via `expand_role_members=false` (default `true` preserves existing behavior). When compact mode is enabled:

- Content documents store compact `role_id:<sys_id>` tokens instead of every role member
- Identity docs include role memberships from `sys_user_has_role` so DLS term overlap still works
- Tables with a `public` read role omit `_allow_access_control` (world-readable via existing DLS `must_not exists`)
- Tables where read roles cannot be resolved omit `_allow_access_control` (same as public tables; world-readable via existing DLS `must_not exists`)

Also fixes advanced sync rules to resolve ACL per table instead of stamping an empty ACL on every batch.

Changing the setting requires a full content sync and access control sync.

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4396 — SharePoint Online compact site-group DLS (same tracking issue)
* Kibana `native_connectors.ts` RCF for `expand_role_members` — to be opened separately

## Release Note

ServiceNow document-level security can store compact role tokens on content documents instead of expanding every role member onto each record, reducing memory use for large tenants. Disable **Expand role members** to enable compact mode. The default preserves the previous per-user expansion behavior. Changing the setting requires a full content sync and access control sync.


app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -443,7 +590,7 @@ async def _fetch_access_controls(self, table_name):
                 user_roles.append(acl.get("sys_user_role", {}).get("value"))
 
             for role in user_roles:
-                if roles.get(role).lower() == "public":
+                if roles.get(role).lower() == PUBLIC_ROLE_NAME:
```
erikcurrin-elastic: This probably needs a null guard.  If role was deleted then roles.get(role) will throw an error

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -227,19 +260,47 @@ async def _fetch_users_by_roles(self, role):
...
+        return user_roles
+
     async def get_access_control(self):
         if not self._dls_enabled():
             self._logger.warning("DLS is not enabled. Skipping")
             return
 
+        self._logger.info(
+            "Enriching identity docs with role memberships from sys_user_has_role"
+        )
```
erikcurrin-elastic: Do we want to do this in legacy mode? The doc will add a role_ids field

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -227,19 +260,47 @@ async def _fetch_users_by_roles(self, role):
...
-        if self._dls_enabled():
-            document[ACCESS_CONTROL] = list(
-                set(document.get(ACCESS_CONTROL, []) + access_control)
-            )
+        if not self._dls_enabled():
+            return document
+        if access_control is None:
+            # Public table: omit field so DLS must_not exists grants access.
+            return document
+        document[ACCESS_CONTROL] = list(
```
erikcurrin-elastic: If we do not order these, it will reindex the doc even if nothing changed, because set is random order

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -411,7 +472,93 @@ async def _fetch_table_data(self, batched_apis, table_access_control):
...
+                self._logger.info(
+                    f"Compact DLS: table {table_name} has public read role — "
+                    "omitting _allow_access_control on content documents"
+                )
+                return None
+        return self._finalize_compact_access_control(table_name, role_sys_ids)
+
+    def _finalize_compact_access_control(self, table_name, role_sys_ids):
+        compact_acl = list({_prefix_role_id(role_id) for role_id in role_sys_ids})
+        if not compact_acl:
```
erikcurrin-elastic, line 587: If a table has no read ACL entries, I believe it is public.  This will make it inaccessible even when the array is empty

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -411,7 +472,93 @@ async def _fetch_table_data(self, batched_apis, table_access_control):
...
+                else:
+                    self._logger.debug(
+                        f"Role '{role_name}' not found while building compact ACL "
+                        f"for table {table_name}"
+                    )
+            return self._finalize_compact_access_control(table_name, role_sys_ids)
+
+        roles_by_sys_id = await self._role_sys_id_to_name_map()
+        role_sys_ids = await self._table_read_role_sys_ids(table_name)
+        for role_sys_id in role_sys_ids:
```
erikcurrin-elastic, line 565: If the role was deleted, I think it locks everyone out.  Filter those?

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -227,19 +260,47 @@ async def _fetch_users_by_roles(self, role):
         ):
             yield user
 
+    async def _fetch_user_roles_map(self):
+        """Build a map of user sys_id -> set of role sys_ids from sys_user_has_role."""
```
erikcurrin-elastic: This may OOM.  We may want to page this sys_user_has_role table

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -411,7 +472,93 @@ async def _fetch_table_data(self, batched_apis, table_access_control):
         finally:
             await self.queue.put(EndSignal.RECORD)
 
-    async def _fetch_access_controls(self, table_name):
+    async def _role_name_to_sys_id_map(self):
+        roles = {}
```
erikcurrin-elastic: We call this for every table.  If the tenant has a lot of roles, we can hit rate limits. Can we cache any of it?

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -411,7 +472,93 @@ async def _fetch_table_data(self, batched_apis, table_access_control):
...
+                if role_sys_id:
+                    role_sys_ids.append(role_sys_id)
+                else:
+                    self._logger.debug(
+                        f"Role '{role_name}' not found while building compact ACL "
+                        f"for table {table_name}"
+                    )
+            return self._finalize_compact_access_control(table_name, role_sys_ids)
+
+        roles_by_sys_id = await self._role_sys_id_to_name_map()
```
erikcurrin-elastic: if _table_read_role_sys_ids() returns an empty array, we completely ignore this.

app/connectors_service/connectors/sources/servicenow/datasource.py
```diff
@@ -411,7 +472,93 @@ async def _fetch_table_data(self, batched_apis, table_access_control):
         finally:
             await self.queue.put(EndSignal.RECORD)
 
-    async def _fetch_access_controls(self, table_name):
+    async def _role_name_to_sys_id_map(self):
```
erikcurrin-elastic: nit... duplicate of _role_sys_id_to_name_map

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Fixed. We now skip unknown/deleted role sys_ids with a warning instead of calling `.lower()` on `None`.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Agreed. Role enrichment is gated to compact mode only; legacy identity docs stay unchanged.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Fixed. ACL lists are sorted for stable ordering and to avoid unnecessary reindexes.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Updated compact path to omit `_allow_access_control` when no roles resolve (aligned with your read). 
Note: this proposal slightly diverges from legacy’s empty-list behavior. We haven't validated real SN semantics end-to-end yet. We'll try to validate this E2E before merging and keep both legacy and compact modes in best shape possible.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Fixed. Orphan ACL role references are filtered out against `sys_user_role` before stamping tokens.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: I'd push back on this comment for now.
Fetches are already paginated; we only hold an O(assignments) map (~tens of MB), not per-doc expansion. But I'd be happy to revisit streaming if ACL sync memory shows up in prod.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Fixed. `sys_user_role` is cached once per sync via `_get_roles_maps()`.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Same as for your fourth comment - empty compact ACL now returns `None` (omit field). 
Legacy divergence and lack of live SN validation.
Will try to validate during E2E testing.

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: Well spotted. I consolidated into a single `_get_roles_maps()` helper.

app/connectors_service/connectors/sources/servicenow/datasource.py
erikcurrin-elastic: I think you misunderstand.  If role_sys_ids is empty, we never need role_by_sys_id, as the for loop will never execute.  It's minor

erikcurrin-elastic (approved):
Needs testing, but the code looks good to me

app/connectors_service/connectors/sources/servicenow/datasource.py
Jan-Kazlouski-elastic: You're right. I misunderstood initially. Pushed the fix now. Should be ok.

PR #4390, opened 2026-08-26
[8.19] buildkite: migrate DRA publish to dractl

Jan-Kazlouski-elastic:
Migrate DRA publishing from Release Manager to `dractl` via `elastic/dra-prep`.

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4341
* https://github.com/elastic/connectors/pull/4344
* https://github.com/elastic/connectors/pull/4371
* https://github.com/elastic/connectors/pull/4373
* https://github.com/elastic/connectors/pull/4376

#### Pre-Review Checklist
- [x] Tested the changes locally

## Test plan

- [x] `make clean install autoformat lint test PYTHON=python3.11`

PR #4401, opened 2026-08-31
buildkite: inline init_python in build-python-packages.sh for 8.19

jmlrt:
The backport of #4390 ("migrate DRA publish to dractl") introduced a `build-python-packages.sh` that sources `.buildkite/shared.sh`, but `shared.sh` only exists on `main` — it was added in early 2025 and never backported to 8.19.

This has been breaking the 8.19 DRA build since Aug 27, causing the `Build Python packages` step to fail with:

```
.buildkite/publish/build-python-packages.sh: line 6: .buildkite/shared.sh: No such file or directory
```

Because `Publish DRA Items` is gated on that step, connectors never publishes its DRA artifacts, so `unified-release-snapshot` can't find the expected container images in `docker.elastic.co/dra-snapshot`.

## Fix

The script only uses `init_python` from `shared.sh`, which is two lines (`source ~/.bash_profile` + `pyenv global $PYTHON_VERSION`). Inline them directly and drop the `shared.sh` dependency.

Assisted-by: Claude (claude-sonnet-4-6)

### DIFF
--- a/.buildkite/publish/build-python-packages.sh
+++ b/.buildkite/publish/build-python-packages.sh
@@ -3,10 +3,11 @@
 # Outputs are uploaded as Buildkite artifacts by the pipeline step.
 set -euo pipefail
 
-source .buildkite/shared.sh
-
 PYTHON_VERSION="${DRA_PYTHON_VERSION:?DRA_PYTHON_VERSION is required}"
-init_python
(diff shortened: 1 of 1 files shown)


Jan-Kazlouski-elastic (approved):
LGTM

PR #4402, opened 2026-08-31
buildkite: add missing DRA_PYTHON_VERSION env var to 8.19 pipeline

jmlrt:
Follow-up to #4401.

The backport of "migrate DRA publish to dractl" (#4390) added a step that references `DRA_PYTHON_VERSION` but didn't include the global env definition present on `main` (`DRA_PYTHON_VERSION: "3.11"`). Without it, `build-python-packages.sh` exits immediately with:

```
.buildkite/publish/build-python-packages.sh: line 6: DRA_PYTHON_VERSION: DRA_PYTHON_VERSION is required
```

One-line fix: add `DRA_PYTHON_VERSION: "3.11"` to the global `env:` block in `pipeline.yml`, matching what `main` has at line 44.

Assisted-by: Claude (claude-sonnet-4-6)

PR #4403, opened 2026-08-31
buildkite: use connectors agent for DRA Prep step on 8.19

jmlrt:
The `DRA Prep (snapshot/staging)` step in the backported DRA pipeline runs `make clean sdist` to build the source distribution artifact. `make sdist` depends on `.venv/bin/python` — it calls `python3 -m venv .venv` as a Make prerequisite. The `core-ubuntu-2204` agent has Python 3.10 but without the `python3.10-venv` package, so venv creation fails immediately:

```
python3 -m venv .venv
The virtual environment was not created successfully because ensurepip is not available.
[...]
apt install python3.10-venv
```

The `enterprise-search-ubuntu-2204-connectors-py` agent already has pyenv with Python 3.11 and full venv support — it's the same agent used by the `build_python_packages` step in this pipeline. We also activate pyenv in `stage_artifacts.sh` so the `python3` command used by `make sdist` resolves to the pyenv-managed Python rather than the system one.

Part of the follow-up from the incomplete backport in #4390: previous fixes in #4401 and #4402 unblocked earlier failures in the pipeline chain; this unblocks the DRA Prep step itself.

See build 24739 for the failure: https://buildkite.com/elastic/connectors/builds/24739

Assisted-by: Claude Code (claude-sonnet-4-6)

### DIFF
--- a/.buildkite/publish/dra/stage_artifacts.sh
+++ b/.buildkite/publish/dra/stage_artifacts.sh
@@ -15,6 +15,11 @@ elif [[ "${WORKFLOW}" != "staging" ]]; then
   exit 1
 fi
 
+# core-ubuntu-2204 has Python 3.10 without python3-venv; use pyenv from the
+# connectors agent image so make sdist can create the .venv prerequisite
+source ~/.bash_profile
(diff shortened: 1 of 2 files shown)


Jan-Kazlouski-elastic (approved):
lgtm

PR #4410, opened 2026-09-01
fix: exclude write-only allow ACEs from network drive DLS read access (#1963) (#2875)

### LINKED ISSUE #1963: [DLS][Network Drive] Document is visible if Group has GRANT access but user has an explicit DENY

### LINKED ISSUE #2875: Request for Aligning Access Control in Enterprise Search Network Drive with Windows File Server
### Problem Description

We have identified two instances where the document access permissions in Elastic Search Applications differ from Windows behavior for network drive documents. In Elastic Search Applications, we use the Network Drive Connector Client for Access Control sync, and users log in with roles that have document-level access permissions based on the information in the "search-acl-filter-" prefixed index.

## Issue 1: Write-only Permissions

**Windows Behavior**: Users with only write permissions for a document cannot retrieve or view the contents of that document.

**Elastic Search Applications Behavior**: The document appears in search results for these users.

**Probable Cause**: The condition at line 860 in the [network_drive.py](https://github.com/elastic/connectors/blob/8.15/connectors/sources/network_drive.py#L860) file lacks a check for read permissions. It doesn't exclude documents from the allow_permissions list when the user lacks read access.

## Issue 2: Deny Permissions Override

**Windows Behavior**: If a user belongs to a group with read permissions for a document but is explicitly denied read access, they cannot open the file in File Explorer or view its contents using PowerShell's `type` command. This is because deny permissions always take precedence over allow permissions.

**Elastic Search Applications Behavior**: The document still appears in search results for these users.

**Probable Cause**: The deduplication process at line 714 in the [network_drive.py](https://github.com/elastic/connectors/blob/main/connectors/sources/network_drive.py#L714) file mixes user IDs and group IDs in the same list. After deduplication, even if the denied user's ID is removed, the allowed group ID remains. This results in behavior opposite to Windows' "deny permissions take precedence" rule.

## Requested Changes

1. Modify the access control logic to check for read permissions before adding documents to the allow_permissions list.
2. Implement a mechanism to ensure that deny permissions take precedence over allow permissions, even when the allow permission comes from a group membership.

These changes would align the behavior of Enterprise Search's network drive document access control more closely with that of the Windows file server.

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/connectors/issues/2875

On Windows, write permission does not grant read access. Network Drive DLS was treating write-only allow ACEs as read access, so users who could only write to a file still saw it in search results.

This change excludes `ACCESS_MASK_ALLOWED_WRITE_PERMISSION` from `allow_permissions`, so only read-capable allow ACEs grant document visibility. Deny-write ACE handling is unchanged.

Deny precedence (group allow + user deny) was already fixed on `main`; this PR addresses the remaining write-only gap reported in both issues.

#### Pre-Review Checklist

## Release Note

Network Drive document-level security no longer grants read access for users who only have write permission on a file, matching Windows file server behavior.

### DIFF
--- a/app/connectors_service/connectors/sources/network_drive/datasource.py
+++ b/app/connectors_service/connectors/sources/network_drive/datasource.py
@@ -609,10 +609,14 @@ class NASDataSource(BaseDataSource):
             else:
                 # Else the SID corresponds to a user, hence we use it directly.
                 permissions = [_prefix_sid(sid)]
+            # A write-only allow ACE grants write but not read access, so it must
+            # not make the document visible in search (Windows: write does not
+            # imply read). A deny-write ACE only removes write access and leaves
+            # read intact, so it is treated as an allow to preserve read access.
             if (
                 ace_type == ACCESS_ALLOWED_TYPE
-                or mask == ACCESS_MASK_DENIED_WRITE_PERMISSION
-            ):
+                and mask != ACCESS_MASK_ALLOWED_WRITE_PERMISSION
+            ) or mask == ACCESS_MASK_DENIED_WRITE_PERMISSION:
                 allow_permissions.extend(permissions)
 
             if (
--- a/app/connectors_service/connectors/sources/network_drive/datasource.py
+++ b/app/connectors_service/connectors/sources/network_drive/datasource.py
@@ -609,10 +609,7 @@ class NASDataSource(BaseDataSource):
             else:
                 # Else the SID corresponds to a user, hence we use it directly.
                 permissions = [_prefix_sid(sid)]
-            # A write-only allow ACE grants write but not read access, so it must
-            # not make the document visible in search (Windows: write does not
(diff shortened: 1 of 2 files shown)


erikcurrin-elastic (approved):
Write only?  Really?

PR #4416, opened 2026-09-01
fix(fips): MD5 will be blocked in FIPS environment without flag

### LINKED ISSUE #4415: fix(fips): hashlib.md5 call in hash_id crashes on FIPS-enabled systems
## Summary

`libs/connectors_sdk/connectors_sdk/utils.py:62` calls `hashlib.md5()` without `usedforsecurity=False`. On a FIPS-enabled host, OpenSSL blocks MD5 unconditionally — regardless of intent — and raises:

```
ValueError: [digital envelope routines] unsupported
```

This crashes every connector that generates document IDs (i.e., all of them) when running in the FIPS Docker image or on a FIPS-enabled RHEL/Fedora host.

## Fix

Pass `usedforsecurity=False` to `hashlib.md5()`. This flag:
- Is supported since Python 3.9 (the FIPS image uses Python 3.11)
- Tells OpenSSL the call is non-cryptographic, bypassing the FIPS block
- Does **not** change the hash algorithm or output — existing document IDs are unaffected

A regression test is included that simulates FIPS enforcement via monkeypatching so this cannot silently regress.

## Affected components

- `libs/connectors_sdk/connectors_sdk/utils.py` — `hash_id()` — shared by all connectors

erikcurrin-elastic:
On a FIPS-enabled host, OpenSSL blocks hashlib.md5() unless usedforsecurity=False is passed. This flag is a no-op on non-FIPS systems and does not change the hash output, so existing document IDs are unaffected.

Adds a test that simulates FIPS enforcement via monkeypatching to prevent regression.

#### Pre-Review Checklist

### DIFF
--- a/libs/connectors_sdk/connectors_sdk/utils.py
+++ b/libs/connectors_sdk/connectors_sdk/utils.py
@@ -58,7 +58,8 @@ def get_file_extension(filename):
 def hash_id(_id):
     # Collision probability: 1.47*10^-29
     # S105 rule considers this code unsafe, but we're not using it for security-related
-    # things, only to generate pseudo-ids for documents
+    # things, only to generate pseudo-ids for documents. useforsecurity=False tells the md5 method that
+    # we are NOT using this for security, otherwise FIPS would block this
     return hashlib.md5(_id.encode("utf8"), usedforsecurity=False).hexdigest()  # noqa S105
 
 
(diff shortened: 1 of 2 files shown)


artem-shelkovnikov (approved):
LGTM. Might be worth leaving a comment why it's there? (e.g. to not force re-indexing of the data while providing FIPS compliance + it's not about security, just about having a unique IDs for docs)

Jan-Kazlouski-elastic (commented):
Changes look good, but CI is failing on lint. I can run an autoformat and push if you want. Also I believe this has to be backported. Keep me honest here. @erikcurrin-elastic

Jan-Kazlouski-elastic (approved):
pushed a small fix, added labels

PR #4396, opened 2026-08-28
fix(sharepoint): add compact site-group DLS mode to reduce ACL memory

Jan-Kazlouski-elastic:
## Closes https://github.com/elastic/chat-program/issues/48

Part of https://github.com/elastic/chat-program/issues/47.
SharePoint Online DLS expands every site group member onto each document's `_allow_access_control` array. Large site groups can produce multi-megabyte ACLs per document and OOM Elasticsearch during `_reindex`.

This PR adds `expand_site_group_members` (default `true` to preserve existing behavior). When disabled (compact mode):

- Content documents store a compact `site_group:<site_id>:<group_id>` token instead of every site group member
- Direct user, Entra group, and site user permissions are unchanged on the document
- ACL sync builds a site-group membership index and enriches identity docs so DLS term overlap still resolves access (including nested Entra groups and EEEU guest exclusion)
- If a compact token cannot be written safely (e.g. missing `site_id`), the connector falls back to expanding site group members and logs a warning

Changing the setting requires a full content sync and access control sync.

#### Pre-Review Checklist
- [x] this PR does NOT contain credentials of any kind, such as API keys or username/passwords (double check `config.yml.example`)
- [x] this PR has a meaningful title
- [x] this PR links to all relevant github issues that it fixes or partially addresses
- [x] if there is no GH issue, please create it. Each PR should have a link to an issue
- [x] this PR has a thorough description
- [x] Covered the changes with automated tests
- [x] Tested the changes locally
- [ ] Added a label for each target release version (example: `v7.13.2`, `v7.14.0`, `v8.0.0`)
- [ ] For bugfixes: backport safely to all minor branches still receiving patch releases
- [ ] Considered corresponding documentation changes
- [ ] Contributed any configuration settings changes to the configuration reference
- [x] if you added or changed Rich Configurable Fields for a Native Connector, you made a corresponding PR in [Kibana](https://github.com/elastic/kibana/blob/main/packages/kbn-search-connectors/types/native_connectors.ts)

#### Changes Requiring Extra Attention

- [x] Security-related changes (encryption, TLS, SSRF, etc)
- [ ] New external service dependencies added.

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4392 — ServiceNow compact role DLS (same tracking issue)
* https://github.com/elastic/kibana/pull/288070 — Kibana RCF for `expand_site_group_members`

## Release Note

[...]

PR #4437, opened 2026-09-03
fix: restrict unpublished SharePoint Online pages to owners/editors (#3645)

### LINKED ISSUE #3645: [SPO] Unpublished pages appear in search results
## Bug Description
With DLS enabled, unpublished Sharepoint Online pages are still returned in search queries to all users that have permissions to view the page.

Unpublished pages are ingested and added to the ES index as they are still considered accessible to both owners and editors of the document. However, the Sharepoint Online API doesn't appear to be updating their ACLs for that page when it goes into an unpublished state, so our SPO connector will continue to treat these pages as accessible to anyone with view permissions.

Also, we should consider exposing a field representing the `published` state of a page (if the Sharepoint API allows for it).

### To Reproduce
1. Add a page to a site in Sharepoint Online and publish it.
2. Within Sharepoint Online, navigate to "Manage access" for the created page and add some users/groups with "View access" 
3. Sync content with DLS. Verify that the `site_page` created was ingested and `_allow_access_control` contains the users with view access.
4. Unpublish the page in Sharepoint Online.
5. Perform a full content sync.
6. Document is still indexed (expected) and `_allow_access_control` is unchanged (bug).

## Expected behavior
The `_allow_access_control` field should only contain owners and editors of the page, implying that users performing searches on the index without ownership/edit access of the unpublished page will not see it in their search results

Jan-Kazlouski-elastic:
## Summary
With DLS enabled, unpublished SharePoint Online pages were still returned to users with view access. SharePoint does not update a page's ACLs when it is unpublished, so the connector kept the old view-only permissions on the indexed document.

This detects draft vs published state from the page version string and restricts `_allow_access_control` on unpublished pages to owners/editors only. A `published` field is also exposed on `site_page` documents.

Closes #3645

## Changes
- Detect published state from `OData__UIVersionString` (major version = published, minor version = draft)
- For unpublished pages, restrict ACLs to owners/editors:
  - unique per-page permissions: only edit-or-higher role members
  - inherited site permissions: use a site editors ACL set instead of the full site ACL
- Add `require_edit_access` to `_get_access_control_from_role_assignment`
- Add `EDIT_ITEM_MASK` / `EDIT_ROLE_TYPES` constants
- Add `published` field on `site_page` documents
- Debug log when an unpublished page ACL is restricted

## Testing
- Unit tests: version parsing, edit-only role filtering, unpublished/published ACL behavior (unique + inherited permissions). `209 passed` in `test_sharepoint_online.py`.
- E2E: not run locally (requires live SharePoint Online tenant with DLS).

## Notes for reviewers
- Published detection defaults to `published=True` when version info is missing, to avoid over-restricting access.
- **Merge note with #4396:** both PRs touch `datasource.py` and `test_sharepoint_online.py`. A test merge shows one mechanical conflict in `_get_access_control_from_role_assignment` (this PR adds `require_edit_access`, #4396 adds `site_id`). Tests auto-merge. Resolution is to keep both optional parameters and pass both from `site_pages`.
- Consider backport labels if needed.

### DIFF
--- a/app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
+++ b/app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
@@ -59,14 +59,7 @@ from connectors.utils import html_to_text, iterable_batches_generator
 
 
 def _is_page_published(version_string):
-    """Determine whether a site page is published based on its version string.
-
-    SharePoint uses major.minor versioning for site pages. Publishing a page
-    creates a major version (the minor component is ``0``, e.g. ``"3.0"``),
-    while a draft/unpublished page carries a minor version (e.g. ``"3.1"`` or
-    ``"0.1"``). When the version cannot be determined we assume the page is
-    published to avoid over-restricting access.
-    """
+    """True when the page version is a major release (e.g. "3.0")."""
     if not version_string:
         return True
 
@@ -339,9 +332,7 @@ class SharepointOnlineDataSource(BaseDataSource):
                 [
                   "user":spo-admin"
                 ]
-            - list: subset of the first list, applying only to members granted
-                edit (or higher) access to the site, plus all site-admins. Used
-                to restrict access to unpublished pages to owners/editors.
+            - list: site members with edit-or-higher access, plus site-admins
                 [
                     "user:spo-admin",
                     "user:spo-editor"
@@ -399,7 +390,6 @@ class SharepointOnlineDataSource(BaseDataSource):
             )
             raise PermissionsMissing(msg) from e
 
-        # Owners/admins always retain access, so they are considered editors too.
         editors_access_control |= site_admins_access_control
 
         return (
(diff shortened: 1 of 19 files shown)


app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
```diff
@@ -56,6 +58,22 @@
...
+        return True
+
+    _, _, minor = str(version_string).partition(".")
+
+    if not minor:
+        return True
+
+    try:
+        return int(minor) == 0
+    except ValueError:
```
erikcurrin-elastic, line 73: Why True?

Jan-Kazlouski-elastic: Good catch
I wanted it to be less restrictive initially, but after thinking about it more, changing this to `False` makes more sense. If we can't parse the version, we shouldn't assume the page is published. Same as with `_grants_access`.

erikcurrin-elastic (approved):

PR #4439, opened 2026-09-03
Feature(sharepoint-online): concurrent ACL sync user processing

### LINKED ISSUE #4435: SharePoint Online: speed up ACL sync via concurrent user processing
### Problem Description
The get_access_control() method in the SharePoint Online connector processes users sequentially — each user is fully awaited before the next one begins. For every user who belongs to more than 20 groups (the Microsoft Graph API $expand limit), an additional paginated Graph API call is issued to fetch their full group membership. In enterprise tenants with thousands of users — many of whom are service accounts or admins belonging to dozens of groups — this creates a long chain of serial network round trips, making the ACL sync significantly slower than it needs to be.

### Proposed Solution
Replace the sequential async for / await process_user(user) loop in get_access_control() with a bounded concurrent task pool using the existing ConcurrentTasks utility (connectors.utils). Results are collected into an asyncio.Queue and yielded as tasks complete, keeping throughput high without unbounded memory growth.

Key details:

Concurrency bounded at DEFAULT_PARALLEL_CONNECTION_COUNT (10) — matches the existing connection pool size used elsewhere in the connector
The deduplication guard (already_seen_ids) is unchanged — it runs before a task is submitted, so no duplicate ACL documents are produced
Results begin yielding immediately as tasks complete rather than waiting for the full user list to be processed

### Alternatives
1. /users/delta incremental endpoint — Microsoft Graph supports a delta query for users that returns only changed users since a previous sync token. This would reduce subsequent ACL syncs from O(N all users) to O(Δ changed users). Higher impact for repeat syncs but requires cursor management and a schema change to the sync state. Worth pursuing as a follow-up.

2. $batch API for group membership fallback — For users who exceed the 20-group $expand limit, the current code already issues one extra paged API call per user. These could be batched into a single Graph $batch request (up to 20 sub-requests), similar to how drive_items_permissions_batch() works. This is complementary to concurrent processing and could be a follow-up optimisation.

3. Increase $top page size — The client already requests $top=999 but the Graph API caps responses at ~100 users per page regardless. No gain available here.

### Additional Context
The bottleneck is most severe in enterprise tenants (10 000+ users, many with large group memberships).

ConcurrentTasks is already used for similar fan-out patterns in the Box, Microsoft Teams, Jira, and Confluence connectors — no new infrastructure is introduced.

MacBlu:
## Closes https://github.com/elastic/connectors/issues/4435

Process users concurrently in get_access_control using ConcurrentTasks (max_concurrency=DEFAULT_PARALLEL_CONNECTION_COUNT=10) instead of serial awaits. Results are collected via asyncio.Queue and yielded as tasks complete, so output begins immediately without waiting for the full user list to finish.

For users who belong to more than 20 groups (the Graph API $expand limit), the existing per-user paged fallback now runs in parallel rather than sequentially, significantly reducing ACL sync wall-clock time in large enterprise tenants.

Two new unit tests cover deduplication correctness and full yield under concurrency.

Benchmark: 241k records, 34h 39m -> 3h 58m (8.7x speedup) with concurrency=10.

#### Pre-Review Checklist
- [X ] this PR does NOT contain credentials of any kind, such as API keys or username/passwords (double check `config.yml.example`)
- [X] Covered the changes with automated tests


app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
```diff
@@ -30,8 +30,7 @@
 )
 from connectors.sources.sharepoint.sharepoint_online.constants import (
     CURSOR_SITE_DRIVE_KEY,
-    EDIT_ITEM_MASK,
-    EDIT_ROLE_TYPES,
```
Jan-Kazlouski-elastic, line 34: `EDIT_ITEM_MASK` and `EDIT_ROLE_TYPES` were removed from this import in the merge with `main` but are still referenced below.
Please restore both imports alongside `DEFAULT_PARALLEL_CONNECTION_COUNT`.

.gitignore
```diff
@@ -37,3 +37,4 @@ libs/connectors_sdk/requirements.txt
 *credentials*
 *secret*
 kubeconfig
+Dockerfile.codeengine
```
Jan-Kazlouski-elastic: Please revert this line. Ignoring `Dockerfile.codeengine` is out of scope for ACL concurrency and shouldn’t ship with this PR.

app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
Jan-Kazlouski-elastic, line 573: Please guard dedup with an `asyncio.Lock` around `_already_seen` / `update_already_seen` in `process_user` (check + mark under the lock). Today there’s no `await` between check and mark, but workers share this set — a lock makes the intent explicit and avoids surprises if someone refactors later.

app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
```diff
@@ -515,25 +518,41 @@ def update_already_seen(*ids):
...
+        try:
+            async for user in self.client.active_users_with_groups():
+                await task_pool.put(lambda u=user: process_user(u))
+                # Drain any completed results without blocking
+                while not results.empty():
+                    yield results.get_nowait()
+
+            # Wait for all in-flight tasks to finish, propagating any worker errors
+            await task_pool.join(raise_on_error=True)
+        finally:
```
Jan-Kazlouski-elastic, line 628: On early exit or error before `join`, `cancel()` alone doesn’t wait for in-flight Graph work. Consider `cancel()` then `await task_pool.join()` (with exceptions handled) so tasks don’t keep running after the generator closes.

app/connectors_service/connectors/sources/sharepoint/sharepoint_online/datasource.py
```diff
@@ -515,25 +518,41 @@ def update_already_seen(*ids):
...
+            # Cancel any tasks still in flight if the generator is closed early
+            # or an exception is raised during enumeration
+            task_pool.cancel()
+
+        # Drain remaining results after all tasks have completed
```
Jan-Kazlouski-elastic: If `join(raise_on_error=True)` raises, this drain never runs and ACL docs already on `results` are dropped. Fail-fast is fine; worth a one-line comment in code that a failed sync may leave a partially updated ACL index (under-indexing for completed users).

app/connectors_service/tests/sources/test_sharepoint_online.py
```diff
@@ -3993,6 +3994,90 @@ async def test_get_access_control_with_dls_enabled_and_fetch_all_users(
 
             assert len(user_access_control_docs) == 2
 
+    @pytest.mark.asyncio
+    async def test_get_access_control_deduplicates_users(self, patch_sharepoint_client):
```
Jan-Kazlouski-elastic, line 4028: Optional: this doesn’t assert dedup under overlap (two tasks in flight for the same user). Not blocking if we add the `already_seen_ids` lock, but a concurrent duplicate test would match the new execution model.

app/connectors_service/tests/sources/test_sharepoint_online.py
```diff
@@ -3993,6 +3994,90 @@ async def test_get_access_control_with_dls_enabled_and_fetch_all_users(
...
+                return docs
+
+            # Start collection as a background task then release the gate once
+            # enough tasks have had a chance to pile up.
+            # Multiple yields are needed: each await task_pool.put() in the
+            # producer loop requires one yield, and each task needs one yield to
+            # reach the await release.wait() suspension point.
+            collect_task = asyncio.create_task(_collect())
+            for _ in range(DEFAULT_PARALLEL_CONNECTION_COUNT * 2):
+                await asyncio.sleep(0)
```
Jan-Kazlouski-elastic: Optional: timing here may be fragile on slow CI. Consider gating on “N tasks entered” via an `Event` instead of a fixed number of `sleep(0)` iterations.

Jan-Kazlouski-elastic (changes requested):
Thanks for the benchmarks and for addressing the first review round in da5223f (pool lifecycle, `join(raise_on_error=True)`, stronger concurrency test).

**Blocking before merge**
- Restore `EDIT_ITEM_MASK` and `EDIT_ROLE_TYPES` in the constants import (dropped in the merge with `main`). They are still used in `_get_access_control_from_role_assignment` (~L1222); without them, edit-access DLS paths raise `NameError`.
- Revert the `.gitignore` entry for `Dockerfile.codeengine` (unrelated to this change).
- Protect `already_seen_ids` with an `asyncio.Lock` around check + update in `process_user` so dedup stays safe if this code is refactored later.

Happy to approve once the blockers and alignment are addressed.

MacBlu:
@Jan-Kazlouski-elastic I believe I have everything merged.  

**Validation Status**

make lint: **Passed (0 errors, 0 warnings)**

pytest tests/sources/test_sharepoint_online.py: **231/231 passed**

PR #4441, opened 2026-09-04
fix(outlook): retry LDAP user searches with fresh connections

Jan-Kazlouski-elastic:
## Problem

On large on-prem Exchange deployments the Outlook Server connector enumerates users from Active Directory via LDAP before opening each mailbox over EWS. `ExchangeUsers` kept a single cached `ldap3` connection for both the normal-user and admin-user searches.

`get_user_accounts()` consumes users lazily: it yields each normal user, the sync processes that mailbox (often for hours), then continues. The admin-user LDAP search can therefore run days after the first bind. AD or intermediate network gear often drops the idle TCP session, and the next search fails with errors such as:

```
UsersFetchFailed: Something went wrong while fetching users. Error: socket sending error[Errno 104] Connection reset by peer
```

Unlike the Office365 Graph path, LDAP had no retry layer. A single transient failure aborted the entire sync. A customer on SDH #1898 hit this after indexing 4.2M+ documents over ~45 hours.

## Fix

- Replace the cached LDAP connection with a fresh `Connection` per search.
- Add `_ldap_search()` with the existing `@retryable` settings (`RETRIES`, `RETRY_INTERVAL`, exponential backoff) and `skipped_exceptions=[UsersFetchFailed]` so empty LDAP results are not retried.
- Unbind after each search attempt.
- Refactor `_fetch_normal_users` / `_fetch_admin_users` to call the shared helper.

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4406 — skip accounts with non-primary SMTP addresses
* https://github.com/elastic/connectors/pull/4078 — skip AD users without a valid `mail` attribute
* https://github.com/elastic/connectors/pull/4094 — per-account EWS error handling and TLS hardening

## Release Note

**Outlook Server connector**: LDAP user enumeration now opens a fresh Active Directory connection per search and retries transient network failures instead of aborting long syncs when the admin-user query runs on a stale connection.

### DIFF
--- a/app/connectors_service/connectors/sources/outlook/client.py
+++ b/app/connectors_service/connectors/sources/outlook/client.py
@@ -145,8 +145,7 @@ class ExchangeUsers:
         self.ssl_enabled = ssl_enabled
         self.ssl_ca = ssl_ca
 
-    @cached_property
-    def _create_connection(self):
+    def _create_ldap_connection(self):
         return Connection(
             server=self.ad_server,
             user=self.user,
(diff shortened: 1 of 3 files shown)


app/connectors_service/connectors/sources/outlook/client.py
```diff
@@ -155,47 +154,48 @@ def _create_connection(self):
...
 
-            if not has_value_for_normal_users:
+            if not has_value:
                 msg = "Error while fetching users from Exchange Active Directory."
                 raise UsersFetchFailed(msg)
 
-            for user in response:
-                yield user
+            return response
+        finally:
```
erikcurrin-elastic, line 177: might want to catch the except here, otherwise a throw on bind will likely not retry

```
finally:
    try:
        connection.unbind()
    except Exception:
        pass
```

Jan-Kazlouski-elastic: Good catch. Fixed.

app/connectors_service/tests/sources/test_outlook.py
```diff
@@ -732,6 +732,65 @@ async def test_fetch_admin_users(mock_connection):
...
+        exchange_server="exchange.example.com",
+        user="user",
+        password="password",
+        ssl_enabled=False,
+        ssl_ca=None,
+    )
+    response = exchange_users._ldap_search("search_query", "filter")
+    assert list(response) == ["user@example.com"]
+    assert mock_connection.call_count == 2
+    assert mock_connection_instance.search.call_count == 2
```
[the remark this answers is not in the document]
Jan-Kazlouski-elastic: Fair enough. Added the check.

app/connectors_service/connectors/sources/outlook/client.py
```diff
@@ -145,8 +145,7 @@ def __init__(
         self.ssl_enabled = ssl_enabled
         self.ssl_ca = ssl_ca
 
-    @cached_property
-    def _create_connection(self):
+    def _create_ldap_connection(self):
```
artem-shelkovnikov, line 148: I'd personally inline this into the method that does LDAP search just in case so that nobody would think of returning back `@cached_property` and causing trouble

artem-shelkovnikov (approved):
LGTM, left non-blocking comment

PR #4455, opened 2026-09-07
test: mock real I/O in unit tests to stop fail-slow flakes

Jan-Kazlouski-elastic:
Several unit tests passed assertions but intermittently failed CI under the default 1s `pytest-fail-slow` threshold because they still hit real I/O:

- Oracle/MSSQL tests patched `Engine.connect` but still called real `create_engine`
- GraphQL `close()` accessed the cached `session` property, creating an `aiohttp.ClientSession` even when none was used
- Some GitLab `get_docs` integration tests left `get_work_items_group` unmocked, triggering real HTTP for epic fetches

This PR mocks `create_engine` via a shared test helper, guards GraphQL session teardown, stubs epic fetches where not under test, and adds a regression test for GraphQL close behavior.

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4184
* https://github.com/elastic/connectors/pull/4442

### DIFF
--- a/app/connectors_service/connectors/sources/graphql/client.py
+++ b/app/connectors_service/connectors/sources/graphql/client.py
@@ -257,10 +257,8 @@ class GraphQLClient:
 
     async def close(self):
         self._sleeps.cancel()
-        session = self.__dict__.get("session")
-        if session is not None:
-            await session.close()
-            del self.session
+        await self.session.close()
+        del self.session
 
     async def ping(self):
         await self.make_request(graphql_query=PING_QUERY)
--- a/app/connectors_service/connectors/sources/graphql/client.py
+++ b/app/connectors_service/connectors/sources/graphql/client.py
@@ -257,8 +257,10 @@ class GraphQLClient:
 
     async def close(self):
         self._sleeps.cancel()
-        await self.session.close()
-        del self.session
+        session = self.__dict__.get("session")
+        if session is not None:
+            await session.close()
+            del self.session
 
     async def ping(self):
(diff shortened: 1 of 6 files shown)


app/connectors_service/connectors/sources/graphql/client.py
```diff
@@ -257,8 +257,10 @@ async def post(self, graphql_query):
 
     async def close(self):
         self._sleeps.cancel()
-        await self.session.close()
-        del self.session
+        session = self.__dict__.get("session")
```
artem-shelkovnikov: That doesn't look right. Why?

Jan-Kazlouski-elastic: Hi @artem-shelkovnikov 

`__dict__.get("session")` approach was done to work around how `@cached_property` behaves. Reading `self.session` in `close()` always creates an aiohttp session, even when the test never made a request. Our tests always call `close()` on teardown, so that was triggering real session creation and flaky fail-slow failures.

The idea was to guard `close()` in production so it **only** shuts down a session if one was **already** opened. That works in practice, but I agree that it looks weird and we probably wouldn't want this in production code as part of `test fix only` PR.

I've switched to a different fix instead: `create_graphql_source` now mocks `aiohttp.ClientSession`, so teardown stays fast without changing connector code. The production `close()` is back to what it was on main.

We might want to clean up session handling in the GraphQL client itself (with explicit `_session = None` like Confluence/Jira) a as a follow up.

artem-shelkovnikov (approved):

Issue #4084, opened 2026-06-19
ssl.verification_mode from Elastic Agent output config is ignored when running under agent


rubenruizdegauna:
## Bug Description
When running Agentless connectors-py **in dev** , setting `ssl.verification_mode: none` on the Elasticsearch output has no effect. Connectors-py always performs full certificate verification regardless of the agent output configuration.

Output configuration in elastic-agent:
```yaml
outputs:
    es-default-output-internal:
        type: elasticsearch
        ssl:
            verification_mode: none
```

Error I got:
```
timestamp:  2026-06-19T08:46:32.487Z
  component:  connectors-py-es-default-output-internal
  logger:     agent-client-py

  error: TlsError
    Cannot connect to host b33b8f3eeb404bd794ae3498e86e17b5.es.eu-west-1.aws.internal.dev.elastic.cloud:443
    ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED]
    certificate verify failed: self-signed certificate (_ssl.c:1016)')]

  stack trace:
    elastic_agent_client/service/checkin.py:101        receive_checkins
    elastic_agent_client/service/checkin.py:137        apply_expected
    connectors/agent/protocol.py:162                   apply_from_client
    connectors/agent/connector_record_manager.py:53    ensure_connector_records_exist
    connectors/protocol/connectors.py:184              connector_exists
    connectors/es/index.py:50                          connector_get
    connectors/es/client.py:254                        execute_with_retry
    connectors/es/index.py:31                          connector_get
    elasticsearch/_async/client/_base.py:271           perform_request
    elasticsearch/_async/client/_base.py:315           _perform_request
    elastic_transport/_async_transport.py:264          perform_request
    elastic_transport/_node/_http_aiohttp.py:228       perform_request  ← raises
```

It would be nice if it could fetch the ssl configuration and pass it to the connector so it can be tested in the dev environment.

rubenruizdegauna:
Not sure if bug or enhancement, so I just chose one :)

Jan-Kazlouski-elastic:
Hello @rubenruizdegauna 

Thank you for your patience. We're happy to announce that this issue is fixed in main and the fix was backported to 9.5, 9.4 and 8.19 release lines. In order to receive the fixed version - you can 

- run from the latest source 
- build a docker image with the fix from the latest version of the appropriate branch
- wait for the snapshot image
- Use 9.4.6 released version

[...]

PR #4456, opened 2026-09-07
fix: read agent ssl config from protobuf Struct (#4084)

Jan-Kazlouski-elastic:
## Follows up on https://github.com/elastic/connectors/issues/4084

#4391 started honoring `ssl.verification_mode` from Elastic Agent output config, but
read the nested `ssl` block with dict-style `.get()`. Agent delivers output config as
protobuf `Struct` objects, so check-in crashed with `AttributeError` whenever an `ssl`
block was present (reported in devenv after #4391).

Read `verification_mode` via `fields` and bracket access, matching how other agent
config values are handled. Update SSL tests to use protobuf `Struct` sources like Agent
sends instead of plain dict mocks.

#### Pre-Review Checklist

## Related Pull Requests

* https://github.com/elastic/connectors/pull/4391

## Release Note

Fixes a regression where connectors running under Elastic Agent crashed on check-in when
the Elasticsearch output policy included an `ssl` block. Agent SSL settings are now
read correctly from the protobuf config payload.

### DIFF
--- a/app/connectors_service/connectors/agent/config.py
+++ b/app/connectors_service/connectors/agent/config.py
@@ -120,8 +120,9 @@ class ConnectorsAgentConfigurationWrapper:
         # ESClient ignores verify_certs unless ssl is enabled.
         ssl_config = {"ssl": True}
 
-        verification_mode = ssl_source.get("verification_mode")
-        if verification_mode is not None:
+        verification_mode_field = ssl_source.fields.get("verification_mode")
+        if verification_mode_field is not None:
+            verification_mode = ssl_source["verification_mode"]
             verify_certs = verification_mode != "none"
             logger.debug(
                 f"Found ssl.verification_mode '{verification_mode}', "
--- a/app/connectors_service/tests/agent/test_agent_config.py
+++ b/app/connectors_service/tests/agent/test_agent_config.py
@@ -7,6 +7,8 @@ import warnings
 from unittest.mock import MagicMock, Mock
 
 from elastic_transport import SecurityWarning
+from google.protobuf import json_format
+from google.protobuf.struct_pb2 import Struct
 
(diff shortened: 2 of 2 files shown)


app/connectors_service/connectors/agent/config.py
```diff
@@ -120,8 +120,9 @@ def _extract_ssl_config(self, source):
         # ESClient ignores verify_certs unless ssl is enabled.
         ssl_config = {"ssl": True}
 
-        verification_mode = ssl_source.get("verification_mode")
-        if verification_mode is not None:
+        verification_mode_field = ssl_source.fields.get("verification_mode")
+        if verification_mode_field is not None:
+            verification_mode = ssl_source["verification_mode"]
```
lorenabalan, line 125: isn't this equivalent to
```suggestion
        verification_mode = ssl_source.fields.get("verification_mode", {}).get("verification_mode")
        if verification_mode is not None:
```
and that way we protect from KeyError on second `verification_mode = ssl_source["verification_mode"]`

Jan-Kazlouski-elastic: Thanks @lorenabalan 
`fields.get("verification_mode")` returns a protobuf `Value`, not a dict, so chaining `.get()` on it would hit the same `AttributeError` we’re fixing; the two-line `fields.get` + `ssl_source["verification_mode"]` pattern matches how Agent config is read elsewhere and is safe once the field check passes.

Jan-Kazlouski-elastic: I also tried your suggestion with throw-away unit test and caught AttributeError, so I'm sure here.

lorenabalan: Ah okay sorry I misread the description. LGTM

lorenabalan (approved):

